PR #34, opened 2019-04-12
PD-147: Icon Component

DesignerDave:
This pull request implements the Icon component, a component that dynamically renders an inline SVG based on the `glyph` prop passed in.

Screen Shot 2019-04-12 at 4 15 19 PM

The initial set of icons is fairly limited compared to what we will need in the future to have this component widely adopted, but implements all of the required logic to easily implement new and modify existing glyphs with little to no development work.

Changes:
- Implemented a new Icon component.
- Updated the Webpack configuration with a loader scoped to apply only to the Icon component.
- Implemented a `.svgrrc.js` to configure the loader behavior, and component output.
- Tweaked the storybook styling to better present grids of things.


webpack.config.js
```diff
@@ -50,6 +50,14 @@ module.exports = function(env = 'production') {
           exclude: /node_modules/,
         },
 
+        {
+          test: /\.svg(\?v=\d+\.\d+\.\d+)?$/,
+          include: [
+            path.resolve(__dirname, "packages/Icon")
```
hswolff: Going to whitelist which folders will use `svgr`? We don't want all svgs to use svgr?

packages/Icon/src/createIconComponent.js
```diff
@@ -0,0 +1,59 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+export default glyphs => {
+  return class Icon extends PureComponent {
```
hswolff: Are we making this a class only for PureComponent usage as we're unsure if we can use React.memo?

We need to get clear minimum requirements for browsers and React. I'm still uncertain about these things, unless you know.

packages/Icon/src/createIconComponent.js
```diff
@@ -0,0 +1,59 @@
...
+        PropTypes.oneOf(['default', 'large', 'xlarge']),
+        PropTypes.number,
+      ]),
+    };
+
+    static defaultProps = {
+      size: 'default',
+    };
+
+    get humanReadableTitle() {
```
hswolff: Neat, but not necessary. Can we stick with functions as that is what people reading the code expect to see?

hswolff: This can also just be a plain function that is written in the module.

Ditto the other function.

webpack.config.js
DesignerDave: Not necessarily. Initially I was gonna specify the configuration here which we won't want to apply to all SVGs (given the icon-specific template). Since we aren't, I could justify removing this for the time being.

packages/Icon/src/createIconComponent.js
DesignerDave: It's a class for organizational reasons mostly. Static `displayName`, `propTypes`, and `defaultProps` are nice, and it also has a few getters that would be a bit messier in a functional component in my opinion. It's a class component less by necessity than because it's nicer to read.

Support for Memo is an issue, but given that these components will rarely re-render, this was not written as so for performance reasons.

packages/Icon/src/createIconComponent.js
DesignerDave: Fair enough – I rather like this (especially for the other getter), but I get your point that it may not be clear to those reading that this is where the class property is coming from (though wouldn't they look for `humanReadableTitle` if they needed to find it either way?).

I can make these either a class method, or do this in the render method once we decide on class vs functional component in here.

.svgrrc.js
```diff
@@ -0,0 +1,53 @@
+module.exports = {
```
hswolff: Can we add any links to documentation for this file? Impossible to find it on their website.

packages/Icon/src/createIconComponent.js
hswolff: When I see a value of `this.something` I assume it's an instance property, not a getter. Which may just be me, but I think that's the default way people read classes.

Additionally having it as an instance getter can make people believe that it relies on state when in actuality it's a pure function with no reliance on state.

I think having this as either a static method, directly in render, or as a standalone function in the module (unexported) are all fine.

webpack.config.js
hswolff: Hmm....

Yeah I'm leaning towards removing it such that the assumption is this behavior is true for all usages. i.e. You assume this rule is applied universally from the start, not that it starts as an exceptional opt-in behavior. 

If we need to customize our svg consumption then we can edit the webpack config at that point.

.svgrrc.js
DesignerDave: Yeah, the docs are pretty bad.

The default template is here (with no documentation on it's parameters): https://www.smooth-code.com/open-source/svgr/docs/cli/#use-a-specific-template

And the config file docs are here: https://www.smooth-code.com/open-source/svgr/docs/configuration-files/

I'll update with these links.

packages/Icon/src/createIconComponent.js
DesignerDave: They aren't pure, since they do have a reliance on props, but point taken.

I'll update them to not be getters.

webpack.config.js
DesignerDave: Fair enough!

packages/Icon/src/createIconComponent.js
DesignerDave: So, I updated this, but didn't do a static method, standalone function, or directly in the render method.

I went with an instance method, since I wanted to avoid passing props around too much, and wanted to keep the render method more minimal.

If you're still not cool with this, I'll refactor this to return a functional component instead, and host all the logic there.

DesignerDave:
RFAL @hswolff

.svgrrc.js
```diff
@@ -0,0 +1,58 @@
...
+			};
+
+			export default ${componentName};
+		`;
+  },
+  jsx: {
+    babelConfig: {
+      plugins: [
+        [
+          '@svgr/babel-plugin-replace-jsx-attribute-value',
```
hswolff: Where's documentation for this? So opaque what this is doing. May want to add some comments.

packages/Icon/src/Icon.spec.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { render, cleanup } from 'react-testing-library';
+import Icon from '.';
+
+afterAll(cleanup);
+
+// Test here
```
hswolff: Let's make one placeholder test that just asserts something minimal. Not sure what tbh....if you can't think of one either then I guess we can skip :\

packages/Icon/src/createIconComponent.js
```diff
@@ -0,0 +1,59 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+export default glyphs => {
+  return class Icon extends PureComponent {
+    static displayName = 'Icon';
+    static propTypes = {
```
hswolff: nit, line break between expressions

packages/Icon/src/createIconComponent.js
```diff
@@ -0,0 +1,59 @@
...
+        PropTypes.oneOf(['default', 'large', 'xlarge']),
+        PropTypes.number,
+      ]),
+    };
+
+    static defaultProps = {
+      size: 'default',
+    };
+
+    getHumanReadableTitle() {
```
hswolff: So when I called these 'pure functions' I meant that they don't rely on the state of the component instance. These can be written as functions that take an input and give an output - by which I mean they don't need to know anything about the instance state, and as such shouldn't be instance methods.

Having these as instance methods implies to the reader that there is a reliance on internal state, which isn't true.

My desire is these be standalone functions or static methods.

or...

Just inline them in the render function.

packages/Icon/src/Icon.spec.js
DesignerDave: I will add tests as a part of this PR – this was just a placeholder for the moment, and wasn't intending for this to go into master as it is.

.svgrrc.js
DesignerDave: Done

.svgrrc.js
DesignerDave: Added comment

DesignerDave:
RFAL @hswolff 

Note: Added a few tests for `createIconComponent`, and added some documentation.

packages/Icon/README.md
```diff
@@ -0,0 +1,84 @@
...
+	<div>
+		<Icon glyph='MyCustomGlyph' />
+	</div>
+);
+```
+
+We also export the default icon set for you! If you want to include our glyphs with your custom glyphs, you can do something like this:
+
+```js
+import { createIconComponent, glyphs } from '@leafygreen-ui/Icon;
```
hswolff: Missing trailing `"`.

packages/Icon/src/createIconComponent.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default glyphs => {
```
hswolff: nit: IMO this will be easier to read + find if it was made a function declaration with a name. i.e. `export default function createIconComponent(glyphs) { .. }`. Such that if someone is grepping the code base then they'd find this function. It also highlights that this is _the thing_ that is being exported. As an anonymous function it gets kinda lost in the other noise of the function.

hswolff: Thought: what if we allowed this argument to also accept any valid React Node, such that someone could do `createIconComponent(() => <svg />)` and get back an `Icon` which has a fixed value for its `glyph` prop? May not be necessary now so no need to bloat, but wanted to raise it as a question.

DesignerDave: Funny you mention that, I was actually talking with a friend of mine about that pattern over the weekend. Really wasn't something I had thought about much prior, but I love it as a pattern going forward!

DesignerDave: Done

packages/Icon/README.md
DesignerDave: Fixed

PR #36, opened 2019-04-16
Initial Typescript Tooling Support

brookescarlett:
Configured the tooling to enable Typescript in Leafygreen :)


package.json
```diff
@@ -72,8 +73,16 @@
...
+    "typescript": "^3.4.3",
     "url-loader": "1.1.2",
     "webpack": "4.26.0",
     "webpack-cli": "3.1.2"
+  },
+  "dependencies": {
+    "@types/jest": "^24.0.11",
+    "@types/node": "^11.13.4",
+    "@types/react": "^16.8.13",
+    "@types/react-dom": "^16.8.4"
```
DesignerDave: These can probably be `devDependencies`.

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist/",
+    "sourceMap": true,
+    "module": "commonjs",
```
DesignerDave: This should probably be `es6`. @hswolff that sound right?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist/",
```
hswolff: We can omit as we're not actually having TS output any files.

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "sourceMap": true,
```
hswolff: Ditto above.

tsconfig.json
hswolff: This is how TypeScript transpiles your code. Since we're not transpiling we can probably omit, or what we did in Cloud is have this, and target be `esnext` to effectively turn off transpiling.

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "module": "commonjs",
+    "target": "es6",
+    "jsx": "preserve", 
+  },
+  "allowJs": false,
```
hswolff: This value up till the 'include' (excluding the include prop) are all child properties of `compilerOptions`.

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+  "noEmit": true,
+  "pretty": true,
+  "strictNullChecks": true,
```
hswolff: If we can we should just opt-in to `strict: true` which turns on all strict typing from the get go.

tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
```
hswolff: Let's also add `allowSyntheticDefaultImports`. Can see [what it means here](https://www.typescriptlang.org/docs/handbook/compiler-options.html) but basically makes TypeScript behave more like Babel when doing imports. 

i.e. you can do `import React from 'react'` instead of `import * as React from 'react'`.

hswolff: Also need `"moduleResolution": "node". [More details here](https://www.typescriptlang.org/docs/handbook/module-resolution.html)

brookescarlett: Got it!

tsconfig.json
jetpacmonkey: We probably actually _want_ TS to output definition files, right? ([TS docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html))

jetpacmonkey:
If I'm understanding this right, this PR will only add internal consistency through types, and won't expose them to consumers of the library.

brookescarlett:
Based on our conversation @jetpacmonkey, it sounds like we don't need to expose anything until we have Typescript components -- then it'll fall on the individual packages to expose their types to the consumers of the library. This PR is just the initial tooling support for Typescript

PR #41, opened 2019-04-22
PD-227: Popover Component

brookescarlett:
Popover Component 
- Positions content relative to a refEl (a reference passed as a prop titled refEl)
- The content can either be portaled or positioned absolutely, relative to a surrounding parent
- There are 12 options for positioning, justify: start, middle, end and align: top, bottom, left, right
- Spec file includes some basic tests for the component, as well as one unit test (unit tests were harder to write, because lots of the methods depend on props) -- that said, very open to feedback on how to write more tests to provide better coverage. Additionally, tested in MMS
Screen Shot 2019-04-22 at 3 14 37 PM


packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+  }
+
+  // Determines the justification to render based on an order of justification fallbacks
+  // Returns the first justification that doesn't collide with the window,
+  // defaulting to the justify prop if all justifications fail.
+  getJustification(alignment) {
+    const { justify } = this.props;
+    let justifications = {};
+
+    if (alignment === 'top' || alignment === 'bottom') {
```
DesignerDave: Could write this differently to make it more readable:
```JS
switch (alignment) {
  case 'top':
  case 'bottom': {
      justifications = {
        start: ['left', 'right', 'center'],
        middle: ['center-horizontal', 'right', 'left'],
        end: ['right', 'left', 'center'],
      };
  }

  case 'left':
  case 'right': {
      justifications = {
        start: ['top', 'bottom', 'center'],
        middle: ['center-vertical', 'bottom', 'top'],
        end: ['bottom', 'top', 'center'],
      };
  }`
}

File: packages/Popover/src/Popover.js [hunk 2]
```diff
@@ -0,0 +1,500 @@
...
+  }
+
+  // Check if vertical position collides with edge of window
+  checkVerticalWindowCollision(top) {
+    const contentRefPos = this.getRefPosition(this.contentRef);
+    const tooTall = top + contentRefPos.height > this.state.windowHeight;
+
+    return !(top < 0 || tooTall);
+  }
+  ////////////////////////////////////////////////////////////////////////////////////////////
```
DesignerDave: I'd rather we don't put comment breaks like this in

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+        case 'bottom':
+          return referenceElPos.top + referenceElPos.height;
+      }
+    }
+
+    if (justification) {
+      switch (justification) {
+        case 'top':
+          return referenceElPos.top;
+        case 'bottom':
```
DesignerDave: Nit: line break between these case blocks, especially given we aren't using `{}` here.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+    }
+  }
+
+  calcWithoutPortalTop({ alignment, justification }) {
+    const contentRefPos = this.getRefPosition(this.contentRef);
+
+    if (alignment) {
+      switch (alignment) {
+        case 'top':
+          return `${0 - contentRefPos.height}px`;
```
DesignerDave: As above, line breaks for these as well please 😁

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+      className,
+      withoutPortal,
+      refEl,
+      rest,
+    } = this.props;
+    const { position } = this.state;
+
+    const Root = withoutPortal ? Fragment : Portal;
+
+    const style = isActive
```
DesignerDave: I'd love to see this broken up ->
```js
const activeStyle = {
  transform: 'translate3d(0, 0, 0) scale(1)',
  opacity: 1,
  pointerEvents: 'initial',
}

const style = { ...position, ...(isActive && activeStyle) }
```

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+
+    return `${x} ${y}`;
+  }
+
+  ////////////////////////////////////////////////////////////////////////////////////////////
+
+  render() {
+    const {
+      children,
+      isActive,
```
DesignerDave: Been thinking a bit about this a bit, and I _think_ that calling this `active` would be better.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+      className,
+      withoutPortal,
+      refEl,
```
DesignerDave: Any thoughts around the name of this? `refEl` doesn't seem very descriptive, but I'm not sure I have a better alternative...

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+      rest,
```
DesignerDave: `...rest`

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+          pointerEvents: 'initial',
+        }
+      : position;
+
+    return (
+      <Root>
+        <div
+          {...rest}
+          ref={this.contentRef}
+          className={ccClassName(rootPopoverStyle, className)}
```
DesignerDave: So... I _think_ we're gonna use `cx` from emotion instead. Mind giving that a shot? Should be used the same in this case.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+    getUpdatePosition: PropTypes.func,
+  };
+
+  static defaultProps = {
+    align: 'top',
+    justify: 'start',
+  };
+
+  // I know you don't normally like constructors, but it is a clean way to group state/ref declaration
+  // Otherwise wasn't sure where to put the ref declaration in a pattern-y way
```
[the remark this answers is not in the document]
brookescarlett: Fixed

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+  componentDidMount() {
+    const { getUpdatePosition } = this.props;
+
+    if (getUpdatePosition) {
+      getUpdatePosition(this.updatePosition.bind(this, true));
+    }
+
+    const referenceElPos = this.getRefPosition(this.props.refEl);
+
+    this.setState({ hasMounted: true, referenceElPos });
```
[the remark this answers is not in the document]
brookescarlett: I know it's an artificial way but I was doing a lot of research into how to handle this, and I'll find the ticket but the React team suggested this approach in one instance. Definitely open to alternatives though. 

As far as `referenceElPos` goes, this is here because I don't want to call this.getRefPosition in all of the other methods, so I'm just calling it once, here, and leveraging it all around the class methods.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+      hasMounted: false,
+    };
+
+    this.contentRef = React.createRef();
+  }
+
+  componentDidMount() {
+    const { getUpdatePosition } = this.props;
+
+    if (getUpdatePosition) {
```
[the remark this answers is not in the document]
brookescarlett: Will do 👍

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+    if (!this.state.position || force || isStale) {
+      this.setState({
+        position: this.calculatePosition(),
+        lastPositionUpdate: now,
+      });
+    }
+  }
+
+  // Gets top offset, left offset, width and height dimensions for a node
+  getRefPosition({ current }) {
```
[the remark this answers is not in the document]
brookescarlett: Done!

packages/Popover/src/Popover.js
brookescarlett: Removed! Was just trying to make this component a little more readable 😬

packages/Popover/src/Popover.js
brookescarlett: 👍

packages/Popover/src/Popover.js
brookescarlett: 👍

packages/Popover/src/Popover.js
brookescarlett: 👍

packages/Popover/src/Popover.js
brookescarlett: We thought this worked because it both states that it takes a reference and is another element, but I can keep brainstorming for sure

packages/Popover/src/Popover.js
brookescarlett: Ah thank you

packages/Popover/src/Popover.js
brookescarlett: Yep, got it!

packages/Popover/src/Popover.js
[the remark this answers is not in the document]
brookescarlett: We have `current` here so that we can access the `current` key of the ref when it's passed in

packages/lib/src/index.js
```diff
@@ -2,4 +2,24 @@ import emotion from './emotion';
 
 const ccClassName = (...args) => args.join(' ');
 
-export default { ccClassName, emotion };
+// Returns a function, that, as long as it continues to be invoked, will not
+// be triggered. The function will be called after it stops being called for
+// N milliseconds. If `immediate` is passed, trigger the function on the
+// leading edge, instead of the trailing.
+function debounce(func, wait, immediate) {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+      windowWidth: window.innerWidth,
+    });
+  };
+
+  // updatePosition updates the popover position in state if an update is needed
+  // You can pass force to force a recalculation of position
+  updatePosition(force) {
+    const { lastPositionUpdate } = this.state;
+    const now = Date.now();
+    const isStale = !lastPositionUpdate || now - lastPositionUpdate > 1000; // 10 seconds
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+      getUpdatePosition(this.updatePosition.bind(this, true));
+    }
+
+    const referenceElPos = this.getRefPosition(this.props.refEl);
+
+    this.setState({ hasMounted: true, referenceElPos });
+
+    window.addEventListener('resize', this.updateWindowDimensions);
+
+    setTimeout(this.updatePosition.bind(this, true), 500);
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/Popover/src/Popover.js
brookescarlett: 👍

brookescarlett:
Rebased with master, but pulled from my remote and so ended up merging (duplicating commits)

Created a new branch, cherry picked, and that's why there's only one commit that has been force pushed and why some of the comments may be out of order/removed

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+
+  // If refEl is not a supplied prop, determines an appropriate refEl, rather than error
+  getParentRef() {
+    const { withoutPortal } = this.props;
+
+    if (withoutPortal) {
+      return this.getRefPosition({
+        current: this.contentRef.current.parentNode,
+      });
+    } else {
```
DesignerDave: Let's practice happy-path coding here.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+      return this.getRefPosition({
+        current: this.createRefEl.current.parentNode,
+      });
+    }
+  }
+
+  // Sets referenceElPos in state based on our determination of an appropriate reference element
+  setRefEl() {
```
DesignerDave: I think we could merge this with `getParentRef` pretty elegantly.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,500 @@
...
+  getRefPosition({ current }) {
+    if (!current) {
+      return {};
+    }
+
+    const { top, bottom, left, right } = current.getBoundingClientRect();
+    const height = current.offsetHeight;
+    const width = current.offsetWidth;
+
+    return { top, bottom, left, right, height, width };
```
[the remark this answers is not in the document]
DesignerDave: Bump!

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+
+  // Gets top offset, left offset, width and height dimensions for a node
+  getRefPosition({ current }) {
+    if (!current) {
+      return {};
+    }
+
+    const { top, bottom, left, right } = current.getBoundingClientRect();
+    const height = current.offsetHeight;
+    const width = current.offsetWidth;
```
DesignerDave: I think we can do these as:
```JS
const { offsetHeight: height, offsetWidth: width } = current
```

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+      }
+    }
+  }
+
+  // Returns the 'left' position in pixels for a valid alignment or justification.
+  calcLeft({ alignment, justification }) {
+    const { referenceElPos } = this.state;
+    const contentRefPos = this.getRefPosition(this.contentRef);
+
+    if (alignment) {
```
DesignerDave: We don't need these conditionals before the switches. It will work just fine without. Same for the justification switch.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+            contentRefPos.width / 2
+          );
+      }
+    }
+  }
+
+  calcWithoutPortalLeft({ alignment, justification }) {
+    const contentRefPos = this.getRefPosition(this.contentRef);
+
+    if (contentRefPos && alignment) {
```
DesignerDave: We don't need to check for alignment here. Same for justification below.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+    const {
+      children,
+      active,
+      className,
+      withoutPortal,
+      refEl,
+      ...rest
+    } = this.props;
+
+    // const { position } = this.state;
```
DesignerDave: Remove comment please

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+
+    const Root = withoutPortal ? Fragment : Portal;
+
+    const activeStyle = {
+      transform: 'translate3d(0, 0, 0) scale(1)',
+      opacity: 1,
+      pointerEvents: 'initial',
+    };
+
+    const style = { ...position, ...(active && activeStyle) };
```
DesignerDave: Can we move this conditional to the declaration for activeStyle?

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+
+    return (
+      <>
+        <div ref={this.createRefEl} className={divRef} />
+        <Root>
+          <div
+            {...rest}
+            ref={this.contentRef}
+            className={cx(className, rootPopoverStyle)}
```
DesignerDave: To note: with `cx`, order matters, and the last param will take the highest precedence. Let's be sure that `className` always comes last.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+    }
+  }
+
+  // Checks that a justification will not cause the popover to collide with the window.
+  checkJustification(justification) {
+    const { withoutPortal } = this.props;
+    const top = withoutPortal
+      ? this.calcWithoutPortalTop({ justification })
+      : this.calcTop({ justification });
+    const left = withoutPortal
```
DesignerDave: All this ternary block-style formatting is hard to read without spaces in between. Could we be sure to add trailing spaces after blocks, even these?

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+
+    return validAlignment;
+  }
+
+  // Determines the justification to render based on an order of justification fallbacks
+  // Returns the first justification that doesn't collide with the window,
+  // defaulting to the justify prop if all justifications fail.
+  getJustification(alignment) {
+    const { justify } = this.props;
+    let justifications = {};
```
DesignerDave: I don't think we need to supply a value here.

packages/Popover/src/Popover.js
```diff
@@ -0,0 +1,517 @@
...
+export default class Popover extends Component {
+  static displayName = 'Popover';
+
+  static propTypes = {
+    children: PropTypes.node,
+    active: PropTypes.bool,
+    className: PropTypes.string,
+    align: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    justify: PropTypes.oneOf(['start', 'middle', 'end']),
+    refEl: PropTypes.object,
```
DesignerDave: How do we handle if the element that the ref is created on is unmounted?

packages/Popover/README.md
```diff
@@ -0,0 +1,82 @@
+# Popover
```
DesignerDave: Oh, please document the parent ref API with portals. E.g. if the popover is the child of the element they want it to be relative to, they don't need to pass a ref.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,619 @@
...
+  | 'right'
+  | 'center-vertical'
+  | 'center-horizontal';
+
+type AbstractPosition = {
+  alignment?: Alignment;
+  justification?: Justification;
+};
+
+type Props = {
```
brookescarlett: nit and probably not a rule, but should we group the required/optional props when we type props?

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,619 @@
...
+    className: PropTypes.string,
+    align: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    justify: PropTypes.oneOf(['start', 'middle', 'end']),
+    refEl: PropTypes.object,
+    withoutPortal: PropTypes.bool,
+    getUpdatePosition: PropTypes.func,
+  };
+
+  static defaultProps: Props = {
+    align: 'top',
```
brookescarlett: do we want this to be bottom or top?

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,619 @@
...
+    const newReferenceElement = (() => {
+      if (passedRef && passedRef.current) {
+        return passedRef.current;
+      }
+
+      if (this.placeholderRef && this.placeholderRef.current) {
+        const parent = this.placeholderRef.current.parentNode;
+
+        if (parent && parent instanceof HTMLElement) {
+          return parent;
```
brookescarlett: 👍

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,619 @@
...
+  // Returns the first alignment that doesn't collide with the window,
+  // defaulting to the align prop if all alignments fail.
+  getAlignment() {
+    const { align } = this.props;
+
+    const alignments: {
+      top: Array<Alignment>;
+      bottom: Array<Alignment>;
+      left: Array<Alignment>;
+      right: Array<Alignment>;
```
brookescarlett: 💯

packages/Popover/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import Popover from './Popover';
```
brookescarlett: I did this but,  `export default from './Popover'`

DesignerDave: Yeah but that doesn't work in Typescript since we don't have the babel plugin for it.

packages/Popover/src/Popover.tsx
DesignerDave: 😄

packages/Popover/src/Popover.tsx
DesignerDave: You're right, probably bottom.

packages/Popover/src/index.ts
brookescarlett: Ooh

packages/Popover/src/Popover.tsx
DesignerDave: Ehh I'm not sure it matters. People will see the inline annotations and errors from not having the right type, so I don't think the order matters as much.

packages/Popover/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+**Default:** `false`
+
+Determines whether the Popover is active or inactive
+
+### align
+
+**Type:** `string`
+
+**Default:** `top`
```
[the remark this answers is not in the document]
DesignerDave: Yep, you're right.

packages/Popover/README.md
```diff
@@ -0,0 +1,86 @@
...
+<div>Popover!</div>
+```
+
+### refEl
+
+**Type:** `object`
+
+**Default:** `null`
+
+Required. Ref to the element to which the popover component should be positioned against
```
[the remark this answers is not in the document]
DesignerDave: 👍

packages/Popover/src/Popover.tsx
brookescarlett: Kk! SGTM

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,625 @@
...
+type AbstractPosition = {
+  alignment?: Alignment;
+  justification?: Justification;
+};
+
+type Props = {
+  children?: ReactNode;
+  active?: boolean;
+  className?: string;
+  align?: Alignment;
```
jetpacmonkey: If it's in `defaultProps`, then it should be marked as required. Defaults are applied before the types are checked.

As far as TS is concerned, for a component `Comp` with `defaultProps = { foo: 'bar' }`, `<Comp />` is equivalent to `<Comp foo="bar" />`.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,625 @@
...
+  | 'right'
+  | 'center-vertical'
+  | 'center-horizontal';
+
+type AbstractPosition = {
+  alignment?: Alignment;
+  justification?: Justification;
+};
+
+type Props = {
```
jetpacmonkey: I generally use `interface` for props, although for the most part they seem to be interchangeable. @hswolff do you remember what the main reasons were for that convention?

webpack.config.js
```diff
@@ -24,8 +24,19 @@ module.exports = function(env = 'production') {
           'create-emotion',
           'polished',
           'prop-types',
-          '@leafygreen-ui/lib',
-          '@leafygreen-ui/theme',
+          ...[
```
hswolff: This is a list of all our current packages? 

This file is ran inside of a node context, meaning we have access to all node APIs.

As such we can make a function that reads all the directories inside of the packages/ directory and returns the list of their names, such that this array is generated off of the directory content and not something we have to manually sync. 

And even better...you could read all `package.json` files and take their `name` value such that there's no deriving of package names, we're directly referencing them.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,624 @@
...
+    active: PropTypes.bool,
+    className: PropTypes.string,
+    align: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    justify: PropTypes.oneOf(['start', 'middle', 'end']),
+    refEl: PropTypes.object,
+    withoutPortal: PropTypes.bool,
+    getUpdatePosition: PropTypes.func,
+  };
+
+  static defaultProps: DefaultProps = {
```
hswolff: TS should be able to infer this without you being explicit.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,624 @@
...
+  static displayName = 'Popover';
+
+  static propTypes = {
+    children: PropTypes.node,
+    active: PropTypes.bool,
+    className: PropTypes.string,
+    align: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    justify: PropTypes.oneOf(['start', 'middle', 'end']),
+    refEl: PropTypes.object,
+    withoutPortal: PropTypes.bool,
```
hswolff: This is inverted logic which makes it a little hard to parse. Usually better to have the prop name align with the value being passed in.

As is now if I do `<Popover withoutPortal={true} />` that doesn't use a portal.

Compare to `<Popover withPortal={true} />, that does use a portal. 

i.e. the value of true affirms what the prop describes, as opposed to refuting it.

We can have `asPortal` or `withPortal` or `usePortal` and have the default value be true.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,624 @@
...
+}
+
+export default class Popover extends Component<Props, State> {
+  static displayName = 'Popover';
+
+  static propTypes = {
+    children: PropTypes.node,
+    active: PropTypes.bool,
+    className: PropTypes.string,
+    align: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
```
hswolff: This should use a string enum and then `Object.keys(Align)`. Ditto below.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,624 @@
...
+    getUpdatePosition: PropTypes.func,
+  };
+
+  static defaultProps: DefaultProps = {
+    align: 'bottom',
+    justify: 'start',
+    active: false,
+  };
+
+  state = {
```
hswolff: Due to a limitation of TypeScript we need to explicit declare this as the State type: `state: State = {...}`.

The reason is that TypeScript can't correctly associate the generic type you passed in and assign it here. If you used a constructor and created `this.state` in there instead then TS could do that, but not with this shorthand syntax. So fine just to add the extra type annotation.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,624 @@
...
+export default class Popover extends Component<Props, State> {
+  static displayName = 'Popover';
+
+  static propTypes = {
+    children: PropTypes.node,
+    active: PropTypes.bool,
+    className: PropTypes.string,
+    align: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    justify: PropTypes.oneOf(['start', 'middle', 'end']),
+    refEl: PropTypes.object,
```
hswolff: `PropTypes.shape({ current: PropTypes.instanceOf(Element) })` from https://stackoverflow.com/a/51127130

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,626 @@
...
+      if (!referenceElPos) {
+        this.setState({
+          referenceElPos: this.getRefPosition(referenceElement),
+        });
+      }
+
+      return;
+    }
+
+    const newReferenceElement = (() => {
```
hswolff: Any need for this to be an IIFE?

packages/Popover/src/Popover.tsx
hswolff: Are these `ReadonlyArray<Alignment>`? i.e. they can't be mutated?

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,626 @@
...
+    const { justify } = this.props;
+
+    let justifications: {
+      start: Array<Justification>;
+      middle: Array<Justification>;
+      end: Array<Justification>;
+    };
+
+    switch (alignment) {
+      case 'left':
```
hswolff: Can then use Enum here when it's defined.

packages/Popover/src/Popover.tsx
jetpacmonkey: That's only kind of true. TS will allow you to have an invalid `defaultProps` object, but will complain if you try to use the component with that invalid default prop applied.

```tsx
class Comp extends Component<{ foo: string }> {
  static defaultProps = {
    foo: 3, // no error here!
  }
}

//...

<Comp /> // Will complain here about "foo" being a number
```

packages/Popover/src/Popover.tsx
brookescarlett: That makes sense -- just updated to `usePortal` which defaults to `true` and reversed all logic where we check for the presence of the prop, to checking for the absence/falsey-ness

packages/Popover/src/Popover.tsx
brookescarlett: 👍

packages/Popover/src/Popover.tsx
brookescarlett: 👍

packages/Popover/src/Popover.tsx
brookescarlett: do you mean doing Align.left, based on the enum? @hswolff

packages/Popover/src/Popover.tsx
DesignerDave: Originally, this was written as

```JS
  let newReferenceElement = null

  // Various conditionals and such to give it a value
```

So no, it doesn't strictly *need* to be an IIFE, but writing it as one lets us declare the variable as the result of a block expression. This is why I really want JS `do`.

DesignerDave: We reuse the variable, so making it strictly a function means we would have to call it twice in the function, which seems like a performance no-no since it is dealing with DOM nodes.

packages/Popover/src/Popover.tsx
brookescarlett: 👍

packages/Popover/src/Popover.tsx
brookescarlett: 👍

packages/Popover/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Content that will appear inside of the Popver component
+
+### withoutPortal
+
+**Type:** `boolean`
+
+**Default:** `false`
+
+Will position the contents of the Popover component absolutely to a parent element, which should be position `relative.` The popover contents will appear linearly in the DOM, rather than being portaled to the end.
```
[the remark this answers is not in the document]
DesignerDave: How about something simpler:
> Will position Popover's children relative to its parent without using a Portal. 
> **NOTE:** The parent element should be CSS position `relative`, `fixed`, or `absolute` if using this option.

packages/Popover/README.md
```diff
@@ -0,0 +1,106 @@
...
+</button>
+
+<div align="bottom" justify="start" class="leafygreen-ui-1t5dnko">
+  <div class="leafygreen-ui-ii2v5b">Popover content</div>
+</div>
+```
+
+## Simple Use Case
+
+The popover component will be automatically positioned relative to its neartest parent, which has position `absolute` specified.
```
DesignerDave: Typo - "neartest".

Also, not entirely true. It will default to being positioned relative to its direct parent, regardless of its positioning. The only time it matters what the parent's position is, is when `usePortal: false`. In that case, it will be relative to its nearest descendant that has `position: absolute | relative | fixed`.

jetpacmonkey: nearest ancestor, not descendant :)

brookescarlett: Fixed 👍

DesignerDave: @jetpacmonkey Yep! My bad

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,226 @@
...
+  // We use this to force a re-render on window resize
+  const [windowUpdateVal, setWindowUpdateVal] = useState(true);
+
+  useEffect(() => {
+    const forceUpdate = debounce(
+      () => setWindowUpdateVal(!windowUpdateVal),
+      100,
+    );
+
+    window.addEventListener('resize', forceUpdate);
```
hswolff: How about...we extract this logic into a custom hook (located within this file) so if we need to reuse it we're ready to.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,226 @@
...
+    }),
+  );
+
+  const activeStyle = css`
+    transform: translate3d(0, 0, 0) scale(1);
+    opacity: 1;
+    position: ${usePortal ? '' : 'absolute'};
+  `;
+
+  const contentStyle = cx(
```
hswolff: style nit, I'd inline this on the component.

packages/Popover/src/Popover.spec.js
```diff
@@ -0,0 +1,76 @@
...
+  const { container, unmount } = render(
+    <>
+      <button ref={triggerRef}>Trigger Element</button>
+      <Popover justify="start" align="top" active refEl={triggerRef}>
+        Content to appear inside of Popover component
+      </Popover>
+    </>,
+  );
+
+  test('Should show popover when trigger is clicked', () => {
```
hswolff: we're capitalizing sentences in tests?

packages/Popover/src/positionUtils.ts
```diff
@@ -0,0 +1,563 @@
+import { Align, Justify } from './Popover';
```
hswolff: We should definitely get some unit tests made for this file, only testing the two exported functions.

brookescarlett: will do!

packages/Popover/src/Popover.tsx
brookescarlett: Got it!

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,226 @@
...
+  active,
+  className,
+  usePortal,
+  spacing,
+  align,
+  justify,
+  refEl,
+  ...rest
+}: PopoverProps): ReactElement {
+  const placeholderRef: React.RefObject<HTMLDivElement> = useRef(null);
```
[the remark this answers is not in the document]
brookescarlett: Typescript throws an error when undefined

packages/Popover/src/Popover.spec.js
```diff
@@ -0,0 +1,121 @@
...
+
+    expect(container.innerHTML.includes('Popover in DOM')).toBe(true);
+  });
+
+  test('removes Popover instance on unmount', () => {
+    unmount();
+    expect(container.innerHTML).toBe('');
+  });
+
+  describe('unit tests', () => {
```
hswolff: I'd even make a new file positionUtils.spec.js so we can have them all consolidated.

We should have tests that:

- Top, left, bottom, right viewport collisions. Both when a position is within that boundary and conflicting with it. Such that if we have a popover that's aligned at Top we want to have a test that tests when Top is fine, when we have to move to Align.Bottom, and then to Align.Middle.
  - Need to test absolute and relative positioning for every permutation.

The goal being that we have courage that should we decide to refactor the code we can do so without worry of any regressions, so we need to write tests for any edge cases.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+  function calcResize() {
+    debounce(setWindowUpdateVal(getWindowSize()));
+  }
+
+  useEffect(() => {
+    window.addEventListener('resize', calcResize);
+
+    return () => window.removeEventListener('resize', calcResize);
+  });
```
hswolff: We want this to only run on initial mount, not on every re-render. So we need to add a second argument of an empty array to skip additional renders. Docs: https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,256 @@
...
+}
+
+function getWindowSize() {
+  return {
+    width: window.innerWidth,
+    height: window.innerHeight,
+  };
+}
+
+function useWindowResize() {
```
hswolff: We don't really care how this happens, just the values we want to use. In reality we can name this `useViewportSize` as that's the values we want to use.

packages/Popover/src/Popover.spec.js
brookescarlett: Makes sense, will start working on it 👍

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const [windowSize, setWindowUpdateVal] = useState(getWindowSize());
```
jetpacmonkey: [micro-optimization] This could probably be `useState(getWindowSize)` to have it lazily initialize and not call the function on every render.

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+  function calcResize() {
+    debounce(setWindowUpdateVal(getWindowSize()));
+  }
```
jetpacmonkey: Shouldn't this be `const calcResize = debounce(() => setWindowUpdateVal(getWindowSize()))`?

jetpacmonkey: Oh also, since this event listener is being added/removed inside the `useEffect` call, why not define it inside there as well to keep it as tightly scoped as possible?

packages/Popover/src/Popover.tsx
brookescarlett: 👍

brookescarlett:
@hswolff RFAL: the tests are not complete yet, definitely want to clean them up more, but want to make sure that this is what you had in mind and that I'm thinking about it the right way. Let me know if this is on the right track, and I will add a few more and make the file a lot DRY-er

packages/Popover/src/positionUtils.spec.js
```diff
@@ -0,0 +1,274 @@
...
+};
+const CONTENTELPOS = {
+  bottom: 48.8636360168457,
+  height: 61,
+  left: 0,
+  right: 123.21591186523438,
+  top: 0,
+  width: 154,
+};
+global.window.innerWidth = 1024;
```
hswolff: calculatePosition takes in `windowHeight` and `windowWidth` as arguments. It has default values but we can explicitly set it when calling the function. That was done so we can easily test it.

packages/Popover/src/positionUtils.spec.js
```diff
@@ -0,0 +1,274 @@
...
+global.window.innerHeight = 768;
+
+describe('unit tests', () => {
```
hswolff: Should be name of component we're testing, `positionUtils`

packages/Popover/src/positionUtils.spec.js
```diff
@@ -0,0 +1,274 @@
+import { Align, Justify } from './Popover';
+import { calculatePosition, getElementPosition } from './positionUtils';
+
+const SPACING = 10;
+const CENTERED_REFELPOS = {
```
hswolff: General nit, can we use whole numbers for all values? Is there a reason to use decimals? Will whole numbers allow us to still test things as require?

hswolff: Also perhaps we can restructure this:

```
const refElPos = {
  centered: {},
  topLeft: {}
}
```

Also generally I try and reserve all uppercase for string constants. These are just objects that we're using, so no need to uppercase them.

packages/Popover/src/positionUtils.spec.js
```diff
@@ -0,0 +1,274 @@
...
+  top: 0,
+  width: 154,
+};
+global.window.innerWidth = 1024;
+global.window.innerHeight = 768;
+
+describe('unit tests', () => {
+  describe('calculatePosition', () => {
+    test('TOP: it aligns content to top, when the space is available and the prop is set', () => {
+      const pos = calculatePosition({
```
hswolff: Can we add an explicit type response to calculatePosition? I forgot to include that, and it'd be good to have that be explicitly typed as we do rely on it heavily.

packages/Popover/src/positionUtils.spec.js
```diff
@@ -0,0 +1,274 @@
...
+  right: 123.21591186523438,
+  top: 0,
+  width: 154,
+};
+global.window.innerWidth = 1024;
+global.window.innerHeight = 768;
+
+describe('unit tests', () => {
+  describe('calculatePosition', () => {
+    test('TOP: it aligns content to top, when the space is available and the prop is set', () => {
```
hswolff: We can leverage describe to make logical sections so the tests read with groupings.

i.e. the test output looks like:

```
positionUtils
   calculatePosition
      Align.Top
         when space is available
         when space is not available
```

etc.

packages/Popover/src/positionUtils.spec.js
```diff
@@ -0,0 +1,274 @@
...
+      const pos = calculatePosition({
+        useRelativePositioning: false,
+        spacing: SPACING,
+        align: Align.Top,
+        justify: Justify.Start,
+        referenceElPos: CENTERED_REFELPOS,
+        contentElPos: CONTENTELPOS,
+      });
+
+      expect(pos.top).toBe(303.78692626953125);
```
hswolff: This is exactly what we need to assert.

Could also use https://jestjs.io/docs/en/expect.html#expectobjectcontainingobject to make the writing of the assertions easier. 

(Little confused if that is the right assertion or this one is https://jestjs.io/docs/en/expect.html#tomatchobjectobject )

packages/Popover/src/positionUtils.spec.js
brookescarlett: The reason that I'm using decimals is because I'm testing in the browser as well to make sure these values appear not just in the calculations but also IRL and in the browser there end up being decimals. I can just do pure calculations without double-checking the output elsewhere, if you think that's cleaner and that the double-check is unnecessary.

packages/Popover/src/positionUtils.spec.js
brookescarlett: 👍

packages/Popover/src/positionUtils.spec.js
brookescarlett: 👍

packages/Popover/src/positionUtils.spec.js
brookescarlett: working on doing this now

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,258 @@
...
+  };
+}
+
+function useViewportSize() {
+  const [windowSize, setWindowUpdateVal] = useState(getWindowSize);
+
+  useEffect(() => {
+    const calcResize = debounce(() => setWindowUpdateVal(getWindowSize()), 100);
+
+    // function calcResize() {
```
hswolff: rm

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,258 @@
...
+  transition: transform 150ms ease-in-out, opacity 150ms ease-in-out;
+  position: absolute;
+  pointer-events: none;
+  opacity: 0;
+`;
+
+/**
+ * Options to determine the alignment of the popover relative to
+ * the other component
+ * ---
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/Popover/src/Popover.tsx
```diff
@@ -0,0 +1,258 @@
...
+  useRef,
+  useMemo,
+  Fragment,
+  ReactNode,
+  RefObject,
+  ReactElement,
+} from 'react';
+import PropTypes from 'prop-types';
+import Portal from '@leafygreen-ui/portal';
+import debounce from 'lodash/lodash';
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/Popover/src/Popover.tsx
brookescarlett: 👍

packages/Popover/src/Popover.tsx
```diff
@@ -171,14 +148,43 @@ function Popover({
...
+
+  const lastTimeRefElMutated = useMutationObserver(
+    referenceElement,
+    mutationOptions,
+    () => {
+      return Date.now();
+    },
+  );
+
+  const lastTimeContentlMutated = useMutationObserver(
```
jetpacmonkey: Was this supposed to be `ContentEl`?

packages/Popover/src/hooks.ts
```diff
@@ -0,0 +1,50 @@
...
+
+function getWindowSize() {
+  return {
+    width: window.innerWidth,
+    height: window.innerHeight,
+  };
+}
+
+export function useViewportSize() {
+  const [windowSize, setWindowUpdateVal] = useState(getWindowSize);
```
hswolff: Rename variables to align with name of function?

packages/Popover/src/positionUtils.spec.js
```diff
@@ -0,0 +1,1050 @@
...
+
+          expect(pos.top).toBe(45);
+          expect(pos.left).toBe(15);
+          expect(pos.transformOrigin).toBe('left top');
+          expect(pos.transform).toBe('translate3d(-5px, 0, 0) scale(0.8)');
+        });
+      });
+    });
+
+    describe('when reference element is in the ceneter', () => {
```
hswolff: typo, center

hswolff: Let's also do some 'only works' tests when useRelativePositioning is true. Can just do it for centered, don't need to replicate everywhere.

packages/Popover/src/Popover.story.js
```diff
@@ -0,0 +1,43 @@
...
+        usePortal={boolean('usePortal', true)}
+        spacing={number('spacing', 10)}
+      >
+        <div className={popoverStyle}>Popover content</div>
+      </Popover>
+    </button>
+  );
+}
+
+storiesOf('Popover', module).add('Default', () => <Example />);
```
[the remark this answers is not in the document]
brookescarlett: Were there knobs that you wanted me to expose in connection with this?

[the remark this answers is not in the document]
hswolff: I can't think of any that are applicable. If you think of any definitely add. Can browse what types of knobs are available.

packages/Popover/src/Popover.story.js
```diff
@@ -68,4 +68,41 @@ function Example() {
...
+          select(
+            'Reference Element Position',
+            ['center', 'top', 'right', 'bottom', 'left'],
+            'center',
+          )
+        ],
+      )}
+    >
+      {text('Button Content', 'Popover')}
+      <Popover
```
hswolff: Let's show Popover doesn't have to be a child of the reference element, to really drive home the flexibility of the advanced usage.

DesignerDave:
Good job!

PR #42, opened 2019-04-22
PD-220: Migrate Checkbox to not use interpolated classNames

hswolff:
https://jira.mongodb.org/browse/PD-220

### DIFF
--- a/packages/lib/src/index.js
+++ b/packages/lib/src/index.js
@@ -1,6 +1,6 @@
 import emotion from './emotion';
 
-const ccClassName = (...args) => args.join(' ');
+export const ccClassName = (...args) => args.join(' ');
 
 /**
  * Utility for making it easier to couple a React Component to a css selector.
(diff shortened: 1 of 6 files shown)


packages/lib/src/index.js
```diff
@@ -2,4 +2,25 @@ import emotion from './emotion';
...
+  const prefix = 'data-leafygreen-ui';
+  return {
+    prop: {
+      [prefix]: name,
+    },
+    selector: `[${prefix}="${name}"]`,
+  };
+}
+
+export default { emotion, ccClassName, createDataProp };
```
DesignerDave: Had a thought:

Aren't the following two expressions identical in this case?
```
export default { emotion, ccClassName, createDataProp };
```
vs
```
export { emotion, ccClassName, createDataProp };
```

I know this was here before, but should we really be using `export default` here?

PR #44, opened 2019-04-24
Changed export pattern for RadioGroup and RadioBoxGroup

brookescarlett:
- Changed export pattern for RadioGroup and RadioBox Group
- Can now be imported as: 
  - import RadioBox, { RadioBoxGroup } from '@leafygreen-ui/RadioBoxGroup'
  - import Radio, { RadioGroup } from '@leafygreen-ui/RadioGroup'
- Updated README to reflect pattern change 
- Tested in MMS

### DIFF
--- a/packages/RadioGroup/src/index.js
+++ b/packages/RadioGroup/src/index.js
@@ -1,4 +1,2 @@
-import RadioGroup from './RadioGroup';
-import Radio from './Radio';
-
-export { RadioGroup, Radio };
+export { default as Radio } from './Radio';
+export { default as RadioGroup } from './RadioGroup';
--- a/packages/RadioGroup/src/index.js
+++ b/packages/RadioGroup/src/index.js
@@ -1,5 +1,4 @@
 import RadioGroup from './RadioGroup';
 import Radio from './Radio';
 
-export { RadioGroup }
-export default Radio;
+export { RadioGroup, Radio };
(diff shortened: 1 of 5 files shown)


DesignerDave:
@hswolff I was suggesting this pattern.

The reason I think this pattern may be best is that, you can actually use `Radio` in isolation from `RadioGroup` (though we do some heavy-lifting for the consumer in `RadioGroup`). The reverse is not really true though. You need to use `Radio` in order to use `RadioGroup`.

That said, if you strongly think these should remain as named exports, then we can go with that.

hswolff:
Hmmm...that's a good argument. I think the main reason I prefer keeping named exports is that really both of these components can be used in isolation. Also most of the time we'll be using both components, so remembering which is default and which is named seems like overhead. And one last point is that....the default export isn't even the name of the package, which is even more confusing.

DesignerDave:
All fair points. Sounds good to me - let's keep them named exports if that sounds alright to you @brookescarlett

packages/RadioGroup/src/index.js
```diff
@@ -1,4 +1,4 @@
 import RadioGroup from './RadioGroup';
 import Radio from './Radio';
```
DesignerDave: Might we consider doing the following instead? ```JS
export { default as Radio } from './Radio';
```

brookescarlett: @hswolff thoughts?

PR #45, opened 2019-04-25
PD-220: Refactor toggle component

DesignerDave:
This PR refactors the Toggle so that it no longer depends on interpolating Emotion-generated classNames.


DesignerDave:
@hswolff First attempt at this one. The added layer of styles from `size`s made the pattern you implemented for `Checkbox` a bit cumbersome for this. This implementation does something similar to how we might do the enormous amount of conditional styling, if we were to use `styled`.

Also to note: Prettier did some weird stuff with many of the interpolated ternary expressions. Any recommendations on how to improve prettier's formatting here? It's bothering me quite a bit, and certainly makes the code harder to read.

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,352 @@
...
+  top: -2px;
+  bottom: -2px;
+  left: -2px;
+  right: -2px;
+  border: 2px solid rgba(67, 177, 229, 0.25);
+  border-radius: 50px;
+  opacity: 0;
+  transform: scale(0.8);
+
+  ${toggleInput.selector}:focus ~ & {
```
hswolff: I think we have a different way of thinking about styles.

I would have instead inverted this. Made a `createDataProp('focus-state')` and inside of the toggleInput styles done `&:focus ~ ${focusState.selector}`. i.e. keep the ampersands on the left hand side. Trying to explain my thinking more, but wanted to get this written out for now.

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,352 @@
...
+  line-height: 11px;
+  text-transform: uppercase;
+  font-weight: bold;
+  font-size: 11px;
+  color: white;
+  user-select: none;
+`;
+
+const onLabelStyle = css`
+  ${labelStyleBase};
```
hswolff: Why not just have this as part of `cx` on the components? Compose the styles on the element as opposed to making an 'extended' style.

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,352 @@
...
+  left: 9px;
+  color: #bbebff;
+
+  ${toggleInput.selector}:hover ~ ${toggleGroove.selector} > &,
```
hswolff: ditto here, I would have done within the toggleInputStyle variable: `&:hover ~ ${toggleGroove.selector} > ${onLabel.selector}`.

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,352 @@
...
+        opacity: 1;
+      }
+    `
+        : ''
+    }
+  `;
+
+  const variants = {
+    default: css`
+      ${disabled
```
hswolff: We have these variables in state...all of these styles can be discrete css`` calls that we customize on the component itself. 

Right now I find this very hard to piece together what the underlying style is. 

We could have 

```
const variants = {
  default: ({ disabled, checked }) => {
    if (disabled) return css``
  }
}
```

Similar to what's in Checkbox.js.

packages/Toggle/src/Toggle.js
```diff
@@ -51,20 +396,23 @@ export default class Toggle extends PureComponent {
...
     return (
       <label
-        className={ccClassName(
-          className,
-          style.container,
-          style.containerSizes[size],
-        )}
+        className={cx(statefulStyles.container, className)}
         htmlFor={this.checkboxId}
-        disabled={disabled}
```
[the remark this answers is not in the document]
DesignerDave: Correct, this was only there for a selector that isn't in use.

packages/Toggle/src/Toggle.js
DesignerDave: The way I think about styles is to apply styles for the same element as close as possible. The problem with `&:focus ~ ${focusState.selector}` is that it's not styling the element whose selector you're nested in. Rather than keeping the ampersand to the left, `${toggleInput.selector}:focus ~ &` keeps related styles closer together. It's a very conscious method I've used for a while now that helps readability IMO.

packages/Toggle/src/Toggle.js
DesignerDave: Responded above

packages/Toggle/src/Toggle.js
DesignerDave: Yep, done!

packages/Toggle/src/Toggle.js
DesignerDave: I'm not so sure about this. I will say that the ternary expressions made the styles hard to read. I like having the styles for each element in one place. Take a look at the updates I made. I find it a bit more reasonable to parse through.

If that's still not working, I'll take this for a spin. I'm just worried about having so many places where we will need to pass the same 3 parameters in (`checked, disabled, variant`) and (`checked, disabled, size`) since there's a lot of styles that depend on at least two states.

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,352 @@
...
+  }
+`;
+
+const offLabelStyle = css`
+  ${labelStyleBase};
+  right: 6px;
+  color: #9fa1a2;
+`;
+
+const getStatefulContainerStyles = ({ size, disabled }) => {
```
hswolff: Stateful _seems_ redundant but I can be convinced otherwise. Could just be `getContainerStyles`? For the others as well?

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,352 @@
...
+  const baseStyle = css`
```
hswolff: Why not just put this inline at the bottom of this function?

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,354 @@
...
+      border-radius: 50px;
+      background-color: ${colors.mongodb.blue};
+      box-shadow: inset 0 0 10px rgba(0, 0, 0, 0.1);
+      opacity: 0;
+      transform: scale(0.85);
+    }
+  `;
+
+  if (disabled) {
+    baseStyle += `
```
hswolff: I don't love the string concatenation, but I can't think of a better alternative. We could use object syntax, but that gets super awkward when we deal with pseudo selectors - although maybe it's not that bad.

packages/Toggle/src/Toggle.js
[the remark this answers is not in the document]
DesignerDave: I don't believe it's good semantic HTML to put disabled on the label wrapping the input. We set disabled on the actual checkbox underlying the component already.

packages/Toggle/src/Toggle.js
DesignerDave: I think I prefer concatenation over writing styles as objects still. Part of that might be my resentment of camelCase CSS though, and I definitely don't wanna have to write and read styles as:
```CSS
const styles = {
  'background-color': colors.green,
}
```
to have them be kebab-case.

packages/Toggle/src/Toggle.js
DesignerDave: Done!

packages/Toggle/src/Toggle.js
DesignerDave: Makes sense to me. Done!

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,354 @@
...
+  }
+`;
+
+const offLabelStyle = css`
+  right: 6px;
+  color: #9fa1a2;
+`;
+
+const getStatefulContainerStyles = ({ size, disabled }) => {
+  const baseStyle = `
```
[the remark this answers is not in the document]
DesignerDave: This comment is a duplicate I believe, but it's done regardless!

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,357 @@
...
+      width: 40px;
+    `,
+
+    xsmall: css`
+      height: 14px;
+      width: 26px;
+    `,
+  };
+
+  return cx(
```
hswolff: I love this.

packages/Toggle/src/Toggle.js
hswolff: Yeah, I agree.

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,357 @@
...
+        opacity: 1;
+      }
+    `;
+  }
+
+  const variants = {
+    default: (() => {
+      let variantStyle = '';
+
+      if (disabled) {
```
hswolff: You can just directly return css() with the styles in there, can also be first line in the file.

packages/Toggle/src/Toggle.js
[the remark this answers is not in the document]
hswolff: Ah didn't notice that, got it.

packages/Toggle/src/Toggle.js
DesignerDave: Ahh, good point. Was converting stuff to practice happy-path better, and that would take it further in that direction. Love it.

packages/Toggle/src/Toggle.js
DesignerDave: 😁

DesignerDave: Realized we can use this pattern a lot more. Check out the rest of the file now.

packages/Toggle/src/Toggle.js
DesignerDave: Done and it feels good.

packages/Toggle/src/Toggle.js
```diff
@@ -1,7 +1,370 @@
...
+      border-radius: 50px;
+      background-color: ${colors.mongodb.blue};
+      box-shadow: inset 0 0 10px rgba(0, 0, 0, 0.1);
+      opacity: 0;
+      transform: scale(0.85);
+    }
+  `;
+
+  if (disabled) {
+    return cx(
```
hswolff: What I love about using cx is that it clearly shows which styles will take precedence, as last one in wins.

DesignerDave:
🍾

PR #47, opened 2019-04-25
PD-190: Adds pre-push git hook to repository

DesignerDave:
This PR does a few things:
- Upgrades Prettier.
- Updates README
- Adds Husky, and configuration for it in `package.json` for easy git hooks.
- Adds a `prettier:lint` script.
- Updates `npm run lint` to both run `eslint` and `prettier:lint`.
- Adds prepublish script that should run prior to publishing a package.

🎉Happy committing


package.json
```diff
@@ -32,11 +32,14 @@
     "start": "start-storybook -p 9001 -c storybook -s ./storybook/static",
     "build": "lerna run --parallel build -- -- --env production",
     "build:storybook": "build-storybook -c storybook -o ./storybook/public -s ./storybook/static",
+    "setup": "git config core.hooksPath .githooks && npm run init",
```
brookescarlett: 💯

package.json
```diff
@@ -32,11 +32,14 @@
...
     "init": "npm install && lerna bootstrap && npm run build",
     "link-all-packages": "lerna exec --ignore @leafygreen-ui/lib --ignore @leafygreen-ui/theme -- npm link ",
     "link": "./scripts/link.sh",
-    "lint": "eslint \"packages/**/*.js\"",
+    "lint": "npm run prettier:lint && npm run eslint",
+    "eslint": "eslint \"packages/**/*.js\"",
     "prettier": "prettier --write \"**/*.js\"",
+    "prettier:lint": "prettylint \"**/*.{js,ts}\" --config ./.prettierrc",
```
hswolff: [Prettier has this built in](https://prettier.io/docs/en/cli.html#list-different).

.githooks/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
hswolff: nit, let's put into `scripts/githooks`

.githooks/pre-push
```diff
@@ -0,0 +1,4 @@
...
+
+npm run lint
```
hswolff: We should have a `npm run ci` script which is what the CI will run when we want to validate the new code is merge-able. 

Optionally, we should have this script do that instead.

package.json
hswolff: nit, I'd rename to `setup-hooks` or `install-hooks` or `setup:hooks` or `install:hooks`. And document that we have git hooks available.

Alternatively we can use [husky](https://github.com/typicode/husky) instead to manage git hooks. Very popular and widely used library for managing git hooks, should make maintaining them easier.

DesignerDave: If we add more complexity to the hooks, then would be a good time to explore using husky. Until then, the pre-push script is two lines and not worth a library.

I'm gonna add an `npm run setup:hooks`, and call that as part of `npm run setup`.

package.json
DesignerDave: That's what I thought too, but prettier kept telling me that it didn't recognize `--check`. Spent a bit of time on this yesterday. `prettier --help` doesn't list it as an option either.

.githooks/pre-push
DesignerDave: I think in the future, they'll want to be separate. We may want to implement different tests like visual diffing that would be impractical as part of a git hook. The git hook's job is to stop people from wasting time waiting for the CI to run when something simple is gonna fail.

Probably gonna leave this to you as part of a different set of changes (if you intend to be the one setting up the CI pipeline).

.githooks/pre-push
DesignerDave: Done!

DesignerDave:
RFAL @hswolff

package.json
hswolff: I think it may be due to our version of Prettier. `--check-different` works and it's what I'd recommend we use too so we can see what files aren't formatted correctly. 

```
prettier --list-different "**/*.{js,ts,tsx}"
```

package.json
```diff
@@ -32,11 +32,15 @@
     "start": "start-storybook -p 9001 -c storybook -s ./storybook/static",
     "build": "lerna run --parallel build -- -- --env production",
     "build:storybook": "build-storybook -c storybook -o ./storybook/public -s ./storybook/static",
+    "setup": "npm run setup:hooks && npm run init",
```
hswolff: I don't want to have multiple 'setup' or 'init' scripts. It gets confusing knowing which one to run when you're setting up the repo locally.

Additionally there's nothing that requires a user to install a hook. I believe when you `npm install` and husky is there it installs the hooks for you by default, so there's no avoiding the hooks.

On that note, I think having us run all of lint and all of tests before pushing is a little heavy. It's going to make for an annoying dev experience when everytime I push I have to wait for tests to run. As the codebase gets bigger that'll become slower and slower. Also should be something that is handled by the CI (coming soon!).

However I do think we should have a precommit hook when running publish. We should run `npm run build` before publish to make sure we publish the right assets.

DesignerDave: Fair enough. Given how init is used, I could almost see renaming it to `npm run install`, though that could get confusing with `npm install`. Maybe `npm run bootstrap`? As for the hooks setup script, do you believe that installing them should be entirely standalone then? Thoughts?

Git hooks do need to be installed in some fashion since the `.git/hooks/` directory that is used by default doesn't get committed. If we use husky, yes that would be automated. If you strongly think that we should add it, then we can. It just feels bad since this current implementation is 2-3 lines of code without a library.

Sure we could totally omit the tests when we have a CI process in place. Right now, I think the tests are fine since they currently don't take very long, but we can remove as soon as that's automated.

As for publish, yeah we should flesh out the script to do so more to add protections and automation around building. Made a ticket to do that, since it's outside the scope of this PR: https://jira.mongodb.org/browse/PD-239.

package.json
DesignerDave: As of this PR we should be on the latest. I don't think `--check-different` returns status code `1` on any mismatch like `prettylint` does.

Good point on adding `tsx`.

DesignerDave: As far as I'm aware, I don't think that `--check-different` is an option, is it?

hswolff: It doesn't return 0 so pretty sure it works. We use it on Cloud and it's been fine.

package.json
```diff
@@ -35,13 +35,21 @@
     "init": "npm install && lerna bootstrap && npm run build",
     "link-all-packages": "lerna exec --ignore @leafygreen-ui/lib --ignore @leafygreen-ui/theme -- npm link ",
     "link": "./scripts/link.sh",
-    "lint": "eslint \"packages/**/*.js\"",
-    "prettier": "prettier --write \"**/*.js\"",
+    "lint": "npm run prettier:lint && npm run eslint",
+    "eslint": "eslint \"**/*.{js,ts,tsx}\"",
+    "prettier": "prettier --write \"**/*.{js,ts,tsx}\"",
+    "prettier:lint": "prettier --check \"**/*.{js,ts,tsx}\"",
```
hswolff: nit, let's align with the flag and have it be `prettier:check`. Also should probably have it be `--list-different` so we can see what files are not formatted correctly.

hswolff: Also should probably add `.md` as a file to use prettier on as well.

PR #49, opened 2019-04-26
Setup CI

hswolff:
## ✍️ Proposed changes

Adds CI integration with CircleCI!

🎟 *Jira ticket:* [Name of ticket](https://my.jira/ticket)

## 🛠 Types of changes

## ✅ Checklist

- [ ] Tests pass locally with my changes
- [ ] I ran `npm run prettier` before committing my changes.

## 💬 Further comments

I can configure CircleCI to show up in GitHub two ways.

The current method shows CircleCI as one check in the list of checks:
image

Conversely I can configure CircleCI to show up as multiple checks that do independent things, similar to what the Emotion repo does:

image

I kind of like the second one better, but in the CircleCI UI it shows up as discrete jobs, making the CircleCI UI a little bit harder to read.

We can keep it as is for now, and always migrate to the other method later if we want.

### DIFF
--- a/packages/Portal/src/Portal.spec.js
+++ b/packages/Portal/src/Portal.spec.js
@@ -44,7 +44,7 @@ describe('packages/Portal', () => {
   });
 
   test('does not move Portaled content when the container props is changed, and logs an error to the console', () => {
-    const { container, rerender } = render(
+    const { rerender } = render(
       <div>
         <Portal>Moving Portaled Content</Portal>
       </div>,
--- a/packages/Portal/src/Portal.story.js
+++ b/packages/Portal/src/Portal.story.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import Portal from './Portal';
 import { storiesOf } from '@storybook/react';
 import { emotion } from '@leafygreen-ui/lib';
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  test:
+  build:
     docker:
       - image: circleci/node:10.15
     steps:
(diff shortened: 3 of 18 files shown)


brookescarlett:
This is so exciting! I kind of like the stepped-approach to the checks as well. That way, if something goes wrong, it's easy to locate where the error is. Additionally, I think it's helpful for people less familiar with the repo to see exactly what processes CircleCI is running, just MO 💯 Either way excited to have this tooling enabled.

gssbzn:
Should this add junit support to test so circleci can give us a better reports of failing tests?
jest may need to add a reporter but eslint is just `--format junit -o test_results/eslint.xml` and then add the reference to it on the config file
```yml
- store_test_results:
     path: test_results
```

hswolff:
Was thinking about that but hadn't seen any benefits. Found this [doc page on CircleCI about setting it up](https://circleci.com/docs/2.0/collect-test-data/#jest) and it indeed looks pretty neat!

I'll add, and intentionally fail some things to see what it looks like :)

gssbzn:
You get a nice digest of failures like this
Screenshot 2019-04-26 at 15 55 19

packages/Portal/src/Portal.spec.js
```diff
@@ -44,7 +44,7 @@ describe('packages/Portal', () => {
   });
 
   test('does not move Portaled content when the container props is changed, and logs an error to the console', () => {
-    const { container, rerender } = render(
+    const { rerender } = render(
```
DesignerDave: This change is included in my hooks branch FYI.

packages/Portal/src/Portal.story.js
```diff
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
```
DesignerDave: Same with this.

DesignerDave:
Noted a couple changes that I have in my PR, but otherwise looks good from my end! Since they're identical to the ones on my branch, I _think_ there should be no merge issues.

hswolff:
Woaaaah
image

PR #50, opened 2019-04-26
Configure Jest for TypeScript tests

DesignerDave:
## ✍️ Proposed changes

Previously, tests written in `ts` or `tsx` wouldn't run. This fixes that issue.

## 🛠 Types of changes

## ✅ Checklist

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ N/A
- [ ] ~~I have added necessary documentation (if appropriate)~~ N/A

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -118,7 +118,7 @@ module.exports = {
   // restoreMocks: false,
 
   // The root directory that Jest should scan for tests and modules within
-  // rootDir: "null",
+  // rootDir: null,
 
   // A list of paths to directories that Jest should use to search for files in
   // roots: [
--- a/jest.config.js
+++ b/jest.config.js
@@ -118,7 +118,7 @@ module.exports = {
   // restoreMocks: false,
 
   // The root directory that Jest should scan for tests and modules within
-  // rootDir: "./packages/",
+  // rootDir: "null",
 
   // A list of paths to directories that Jest should use to search for files in
   // roots: [
--- a/jest.config.js
+++ b/jest.config.js
@@ -63,12 +63,23 @@ module.exports = {
   // ],
 
   // An array of file extensions your modules use
-  // moduleFileExtensions: [
-  //   "js",
-  //   "json",
-  //   "jsx",
-  //   "node"
-  // ],
+  moduleFileExtensions: [
+    'js',
+    'jsx',
+    'ts',
+    'tsx',
+    'json',
+    'node',
+    'png',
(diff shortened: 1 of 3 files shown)


jest.config.js
```diff
@@ -107,7 +118,7 @@ module.exports = {
   // restoreMocks: false,
 
   // The root directory that Jest should scan for tests and modules within
-  // rootDir: null,
+  // rootDir: "./packages/",
```
hswolff: o_O this is commented out.

DesignerDave: Supposed to be – I was playing with the option, but we don't need it.

DesignerDave: That said, I can certainly put the comment back at the default, or remove all comments. I didn't mean to change this line.

DesignerDave:
@hswolff CI was giving me the following error:
```
npm WARN prepare removing existing node_modules/ before installation
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

I thought that the issue might've been something I can resolve so I did run `npm i`, but now I think that it may be because of lerna-hoisted packages. Thoughts? It was failing prior to the `package-lock.json` change as well.

jest.config.js
DesignerDave: Update: it's now back to the default value in the comment. There's an argument that we should consider removing the unused options in this file, but I find it's nice to have inline documentation on the defaults for each. I could, however, be easily swayed.

hswolff: Yeah I'm fine either way. No real preference. Could remove and always look up the options later.

hswolff:
I had this happen too and managed to fix it by re-generating package-lock. Maybe let's add emotion and create-emotion to the root repo? Or can try `rm -rf node_modules package-lock.json && npm install` which seemed to fix it last time for me.

DesignerDave:
I'm gonna add the dependencies to the root. I don't think it's a great flow if we have to remove `package-lock` and install before every push.

package.json
```diff
@@ -96,6 +96,9 @@
     "typescript": "^3.4.3",
     "url-loader": "1.1.2",
     "webpack": "4.26.0",
-    "webpack-cli": "3.1.2"
+    "webpack-cli": "3.1.2",
+    "create-emotion": "^10.0.7",
```
hswolff: should be alphabetical. if you do it via npm cli it does it for you.

PR #58, opened 2019-05-01
PD-220: Refactor RadioBox and RadioBoxGroup to eliminate console errors

DesignerDave:
## ✍️ Proposed changes

Refactors RadioBox and RadioBoxGroup to use cx and JS state rather than className interpolation. Also makes RadioBox a functional component, and addresses a stylistic issue with it.

🎟 [Fix Emotion 10 console errors](https://jira.mongodb.org/browse/PD-220)
🎟 [Emit .d.ts files when building TypeScript files](https://jira.mongodb.org/browse/PD-250)

## 🛠 Types of changes

## ✅ Checklist

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [ ] ~~I have added necessary documentation (if appropriate)~~


packages/RadioBoxGroup/src/RadioBox.js
```diff
@@ -1,72 +1,213 @@
...
+  children: PropTypes.node,
+};
+
+RadioBox.defaultProps = {
+  onChange: () => {},
+  disabled: false,
+  className: '',
+};
+
+export default RadioBox;
```
hswolff: nit, inline on same line as function declaration

PR #62, opened 2019-05-06
PD-250: Generate .d.ts files in build script

DesignerDave:
## ✍️ Proposed changes

This PR adds an `emit-types` script, and also adds it to our `build` script so that TypeScript definitions are emitted to the `dist` directory.

🎟 **Jira ticket:** [Emit .d.ts files when building TypeScript files](https://jira.mongodb.org/browse/PD-250)

## 🛠 Types of changes

## ✅ Checklist

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~

## 💬 Further comments

To note:
- Packages written in TypeScript need a `types` field in their `package.json` file for this to add any benefit.
- Those packages also need their own `emit-types` script for lerna to run.

_See `lib` for examples of both of these notes._


package.json
```diff
@@ -42,7 +42,8 @@
     "prettier": "prettier --write \"**/*.{js,ts,tsx,md}\"",
     "prettier:check": "prettier --check \"**/*.{js,ts,tsx,md}\"",
     "test": "jest --env jsdom",
-    "types": "tsc --project tsconfig.json",
+    "emit-types": "lerna run emit-types",
+    "types": "tsc --project tsconfig.json --noEmit",
```
hswolff: How about we update these two to `types:check` and `types:emit`. Or `types` can become `ts`. Up to you.

packages/lib/package.json
```diff
@@ -3,8 +3,10 @@
   "version": "2.1.4",
   "description": "leafyGreen UI Kit Internal Library",
   "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
   "scripts": {
-    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js"
+    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js",
+    "emit-types": "tsc --project ../../tsconfig.json --declaration --declarationDir dist --emitDeclarationOnly"
```
hswolff: Can these args just go into the tsconfig.json file? So we don't have to duplicate them across projects?

README.md
```diff
@@ -47,6 +47,13 @@ The script does several things in order:
 
 After the script completes, you can make changes directly to the component in your local `leafygreen-ui` repository. Once you do this, run `npm run build` in the root of the `leafygreen-ui` repository and the changes will be visible on your running application.
 
+## Notes when creating a new component
```
hswolff: > Creating New Component

package.json
```diff
@@ -42,7 +42,8 @@
     "prettier": "prettier --write \"**/*.{js,ts,tsx,md}\"",
     "prettier:check": "prettier --check \"**/*.{js,ts,tsx,md}\"",
     "test": "jest --env jsdom",
-    "types": "tsc --project tsconfig.json",
+    "emit-types": "lerna run emit-types",
```
hswolff: This can be done in parallel with building components right? I think it's nice to have it as its own method too, but we can optimize the actual build step.

DesignerDave: Given we use `CleanWebpackPlugin`, the types should probably run after, not in parallel, to avoid a race condition issue.

hswolff: That plugin causes so many troubles. We should have a separate command of `clean` that runs `npx lerna clean` and removing the `dist` folders.

packages/lib/package.json
DesignerDave: That's probably a good idea – what are your thoughts on extending the base configuration in a new definition file-specific tsconfig for this?

hswolff: Oh hell yeah that sounds great. Maybe not duplicated across every individual package, but one other package-tsconfig.json that can be solely referenced.

package.json
DesignerDave: It's been fine since we made our packages externals, and I rather like deferring to webpack cleaning up before emitting bundles. Ideally, I think webpack would deal with outputting the definition files as well.

Is it actually a problem for us to have these two commands run synchronously? It's not part of the development workflow, only the distribution flow.

A clean command seems nice and we may want one, but I think that doing `npm run clean && npm-run-all --parallel build ts:emit` would likely take longer overall to run despite the two build commands running in parallel. Could test that theory, of course.

hswolff: `npm run build` is part of the developer setup workflow. That's the main reason I raised the objection.

This could be part of `prepublishOnly` hook instead. I'd be fine to have it run sync after build to keep the two commands separate.

DesignerDave: That's true. If we only have this run on prepublish though, we don't get the benefits of the definition files when we're working with packages internally like lib.

packages/lib/package.json
DesignerDave: Took it as far as I could for the time being. Less duplication when we add this to other packages.

package.json
DesignerDave: Did `ts:check` and `ts:emit`.

README.md
DesignerDave: Done.

README.md
```diff
@@ -47,6 +47,13 @@ The script does several things in order:
 
 After the script completes, you can make changes directly to the component in your local `leafygreen-ui` repository. Once you do this, run `npm run build` in the root of the `leafygreen-ui` repository and the changes will be visible on your running application.
 
+## Creating New Component
+
+### package.json
+
+1. Each of the packages needs it's own `build`, and `emit-types` (TypeScript packages only) script so that lerna can build the package correctly.
```
hswolff: Update rule name.

DesignerDave: Whoops, fixed.

PR #63, opened 2019-05-07
Use package-scoped tsconfig to support individual TS output locations

DesignerDave:
## ✍️ Proposed changes

I ran into an issue where if there were multiple TS packages, the `ts:emit` script would output the following in each package:

```
packages/Package1/dist/
|__ Package1/
     |__ index.d.ts
|__ Package2/
     |__ index.d.ts
```

To repair this, I'm proposing that we introduce a `tsconfig.json` file within each package that looks like the following:
```
{
  "extends": "../../package.tsconfig.json",
  "compilerOptions": {
    "declarationDir": "dist"
  },
  "include": [
    "./**/*"
  ]
}
```

Then, the `ts:emit` script within `package.json` can be simplified to:
```
"ts:emit": "tsc --project ./tsconfig.json"
```

Although adding a repetitive configuration file isn't DRY, I'm having trouble coming up with an approach that's cleaner. All non-package specific options are contained within `package.tsconfig.json`, and this has the side effect of simplifying the emit script.

## 🛠 Types of changes

- [x] Tooling


DesignerDave:
We could if we would rather duplicate the two rules that are still in that config file: ```
  "declaration": true,
  "emitDeclarationOnly": true
```

Alternately, if we anticipate that we won't want to use the TS compiler for the foreseeable future, then we can probably move those rules to the base tsconfig.json.

hswolff:
Oh I got confused that we're extending that file...yeah I'm fine to move that into the main one for now. Seems un-necessarily broken up into multiple files right now.

DesignerDave:
Found a good reason to not merge the two - `ts:check` runs with the `noEmit` option which is incompatible with `emitDeclarationOnly`. We would have to change that script to explicitly set that to `false` to run properly. I think it's probably better as-is.

PR #68, opened 2019-05-21
fix(badge): make all badges same height

brookescarlett:
## ✍️ Proposed changes

Make all badges same height regardless of outline

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/Badge/src/Badge.js
+++ b/packages/Badge/src/Badge.js
@@ -12,7 +12,7 @@ export const baseStyle = css`
   font-size: 11px;
   line-height: 20px;
   border-radius: 50px;
-  height: 22px;
+  height: 20px;
   padding-left: 9px;
   padding-right: 9px;
   text-transform: uppercase;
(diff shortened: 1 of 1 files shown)


packages/Badge/src/Badge.js
```diff
@@ -11,7 +11,7 @@ export const baseStyle = css`
   font-size: 11px;
   line-height: 20px;
   border-radius: 10px;
-  padding: 0px 9px;
+  padding: 1px 9px;
```
DesignerDave: I don't love the approach of changing the size of all other badges to work. I'll suggest a different approach: specifying height directly. Badges aren't meant for wrapping text, so we can directly specify a `22px` height for this element. Let's also take this opportunity to make the component a little cleaner and more resilient.

Even though it's more text,

```
padding-left: 9px;
padding-right: 9px;
```
is better than
``` CSS
padding: 0 9px;
```
because the intent here is not to have any padding, which is the default.

Additionally, for pill-shapes, I always overshoot the desired border-radius. When you have an element that's `20px` tall, `border-radius: 10px;` is rendered the same as `border-radius: 1000px;`. For the outlined badge, the height was actually `22px`, so the border radius specified wasn't enough, and the pill shape wasn't rendered properly. I'd use something like `border-radius: 50px` instead of trying to specify exactly half the height.

PR #72, opened 2019-05-24
fix(popover): correctly calculate rect values on first render

hswolff:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/Popover/src/Popover.tsx
+++ b/packages/Popover/src/Popover.tsx
@@ -150,10 +150,8 @@ function Popover({
   adjustOnMutation,
   ...rest
 }: PopoverProps): ReactElement {
-  const [placeholderNode, setPlaceholderNode] = useElementNode<
-    HTMLDivElement
-  >();
-  const [contentNode, setContentNode] = useElementNode<HTMLDivElement>();
+  const [placeholderNode, setPlaceholderNode] = useElementNode();
+  const [contentNode, setContentNode] = useElementNode();
 
   let referenceElement: HTMLElement | null = null;
 
--- a/packages/Popover/src/Popover.tsx
+++ b/packages/Popover/src/Popover.tsx
@@ -1,5 +1,4 @@
 import React, {
-  useRef,
   useMemo,
   Fragment,
   ReactNode,
@@ -10,7 +9,7 @@ import PropTypes from 'prop-types';
 import Portal from '@leafygreen-ui/portal';
 import { emotion } from '@leafygreen-ui/lib';
 import { calculatePosition, getElementPosition } from './positionUtils';
-import { useViewportSize, useMutationObserver } from './hooks';
+import { useViewportSize, useMutationObserver, useElementNode } from './hooks';
 
 const { css, cx } = emotion;
 
@@ -151,15 +150,18 @@ function Popover({
   adjustOnMutation,
   ...rest
 }: PopoverProps): ReactElement {
-  const placeholderRef: React.RefObject<HTMLDivElement> = useRef(null);
-  const contentRef: React.RefObject<HTMLDivElement> = useRef(null);
+  const [
+    placeholderNode,
+    setPlaceholderNode,
+  ] = useElementNode<HTMLDivElement | null>();
+  const [contentNode, setContentNode] = useElementNode<HTMLDivElement>();
 
   let referenceElement: HTMLElement | null = null;
 
   if (refEl && refEl.current) {
     referenceElement = refEl.current;
-  } else if (placeholderRef.current) {
-    const parent = placeholderRef.current.parentNode;
+  } else if (placeholderNode) {
+    const parent = placeholderNode.parentNode;
 
     if (parent && parent instanceof HTMLElement) {
       referenceElement = parent;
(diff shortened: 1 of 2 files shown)


brookescarlett:
💯

packages/Popover/src/hooks.ts
```diff
@@ -56,3 +56,18 @@ export function useMutationObserver<Value>(
...
+] {
+  const [element, setElement] = useState<ElementType | null>(null);
+
+  const elementRefCallback = useCallback(node => {
+    if (node !== null) {
+      setElement(node as ElementType);
+    }
+  }, []);
+
+  return [element, elementRefCallback];
```
DesignerDave: It would really be sweet if we could offer an API similar to `useRef`... Haven't thought through how that would look from the implementation side without actually using `useRef` in here though.

packages/Popover/src/Popover.tsx
```diff
@@ -151,15 +150,18 @@ function Popover({
   adjustOnMutation,
   ...rest
 }: PopoverProps): ReactElement {
-  const placeholderRef: React.RefObject<HTMLDivElement> = useRef(null);
-  const contentRef: React.RefObject<HTMLDivElement> = useRef(null);
+  const [
+    placeholderNode,
+    setPlaceholderNode,
+  ] = useElementNode<HTMLDivElement | null>();
+  const [contentNode, setContentNode] = useElementNode<HTMLDivElement>();
```
DesignerDave: Boy I love this new hook.

packages/Popover/src/hooks.ts
```diff
@@ -56,3 +56,18 @@ export function useMutationObserver<Value>(
 
   return value;
 }
+
+export function useElementNode<ElementType = HTMLElement>(): [
```
DesignerDave: I'm wondering what the value of having the person using the hook pass a type provides. What was the thought behind this?

It doesn't appear that we treat any element type differently, so couldn't we omit this?

packages/Popover/src/hooks.ts
```diff
@@ -56,3 +56,18 @@ export function useMutationObserver<Value>(
...
+  ElementType | null,
+  (node: ElementType | null) => void
+] {
+  const [element, setElement] = useState<ElementType | null>(null);
+
+  const elementRefCallback = useCallback(node => {
+    if (node !== null) {
```
DesignerDave: Any reason not to do
``` js
if (node) {
```
instead?

packages/Popover/src/Popover.tsx
```diff
@@ -10,7 +9,7 @@ import PropTypes from 'prop-types';
 import Portal from '@leafygreen-ui/portal';
 import { emotion } from '@leafygreen-ui/lib';
 import { calculatePosition, getElementPosition } from './positionUtils';
-import { useViewportSize, useMutationObserver } from './hooks';
+import { useViewportSize, useMutationObserver, useElementNode } from './hooks';
```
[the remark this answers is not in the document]
DesignerDave: Ooooor `@leafygreen-ui/hooks` 😱

DesignerDave:
Left comments, but this addresses the problem perfectly!

packages/Popover/src/hooks.ts
hswolff: Copied from [the react docs FAQ](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node). Don't really care either way.

packages/Popover/src/Popover.tsx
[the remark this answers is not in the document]
hswolff: Ticket created: https://jira.mongodb.org/browse/PD-274

packages/Popover/src/hooks.ts
```diff
@@ -56,3 +56,18 @@ export function useMutationObserver<Value>(
...
   return value;
 }
+
+export function useElementNode<ElementType = HTMLElement>(): [
+  ElementType | null,
+  (node: ElementType | null) => void
+] {
+  const [element, setElement] = useState<ElementType | null>(null);
+
+  const elementRefCallback = useCallback(node => {
```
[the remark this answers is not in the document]
hswolff: useCallback is being memoized via the second arg. I'm a little confused. Also most of this is [inspired from the React docs](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node).

packages/Popover/src/hooks.ts
hswolff: If you wanted to be more exact about the type you have the ability to do so.

packages/Popover/src/Popover.tsx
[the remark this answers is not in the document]
DesignerDave: 😄

packages/Popover/src/hooks.ts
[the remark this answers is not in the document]
DesignerDave: 🤦‍♂ 

I don't know why, but I was thinking of the empty array as saying, "this should fire every time", rather than only ever firing once. Ignore this, I was the one confused 😅

packages/Popover/src/hooks.ts
DesignerDave: I don't personally see a reason not to do `if (node) {` and it's a bit cleaner.

packages/Popover/src/hooks.ts
DesignerDave: I seem to have a case of the Fridays today. Completely missed the point of this in that we use `ElementType` to type the returned value. 👍

PR #74, opened 2019-05-30
Chore: export popover props and clean object defaults

brookescarlett:
## ✍️ Proposed changes

Export Popover props in Popover.tsx and index.ts to be used in Menu
Also removed defaultProps in favor of object defaults

## 🛠 Types of changes

- [ x] Bugfix (non-breaking change which fixes an issue)

## ✅ Checklist

### DIFF
--- a/packages/Popover/src/Popover.tsx
+++ b/packages/Popover/src/Popover.tsx
@@ -59,7 +59,7 @@ export enum Justify {
   End = 'end',
 }
 
-interface PopoverProps {
+export interface PopoverProps {
   /**
    * Content that will appear inside of the popover component.
    */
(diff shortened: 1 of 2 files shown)


packages/Popover/src/Popover.tsx
```diff
@@ -139,15 +139,15 @@ interface PopoverProps {
  * @param props.adjustOnMutation Should the Popover auto adjust its content when the DOM changes (using MutationObserver).
  */
 function Popover({
-  children,
-  active,
+  children = undefined,
```
hswolff: This is undefined by default (due to JS), seems un-necessary to include.

brookescarlett: Kk removing!

PR #75, opened 2019-05-30
feat(hooks): created hooks package, to make custom hooks accesible

brookescarlett:
## ✍️ Proposed changes

Created a package to contain and export our custom hooks, so that they can be accessible around leafygreen. Additionally, added the package as a dependency of Popover, and removed existing hooks file from the Popover package. Now Popover's hooks come from our new hooks package:) 

🎟 _Jira ticket:_ [PD-274](https://jira.mongodb.org/browse/PD-274)

## 🛠 Types of changes

## ✅ Checklist


packages/hooks/src/hooks.ts
```diff
@@ -68,3 +68,23 @@ export function useElementNode<ElementType = HTMLElement>() {
 
   return [element, elementRefCallback] as const;
 }
+
+export function useDocumentEventListener(
```
DesignerDave: Could we add TSDoc annotation to these since they're soon to be externally consumable? Would also be nice for our own development 😬

Also, we should probably have tests for these hooks.

packages/hooks/src/hooks.ts
```diff
@@ -68,3 +68,23 @@ export function useElementNode<ElementType = HTMLElement>() {
...
+  enabled: boolean,
+  type: string,
+  eventCallback: (e) => void,
```
DesignerDave: Nit: `e => void`

packages/hooks/src/hooks.ts
```diff
@@ -68,3 +68,23 @@ export function useElementNode<ElementType = HTMLElement>() {
 
   return [element, elementRefCallback] as const;
 }
+
+export function useDocumentEventListener(
+  enabled: boolean,
```
DesignerDave: Do we need to make this a required prop? I'd think this would default to `true`.

packages/hooks/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './hooks';
```
DesignerDave: If we're doing this, personally I'd rather we define the hooks in the index file.

packages/hooks/src/hooks.ts
DesignerDave: Also, could we actually make the type for this an `EventCallback` or something?

packages/hooks/src/index.ts
brookescarlett: There are some functions that we just use internally, so thought it might be cleaner to break it up like this, but can use named exports or just combine -- up to you!

DesignerDave: I think that 1 of 2 approaches would make sense –
1. Each hook gets it's own file, and we export them on a per-file basis from `index.ts`.
2. All hooks are defined in and exported from `index.ts`.

Right now, it doesn't feel like anything's truly broken up, rather this index file just feels unnecessary at the moment.

packages/hooks/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@leafygreen-ui/hooks",
+  "version": "0.9.0",
```
hswolff: I think having this as 1.0.0 is fine, no?

packages/hooks/src/index.ts
hswolff: I'm a fan of 1 hook per file. Makes git diffs easy.

packages/hooks/src/hooks.ts
hswolff: May want to use this: https://github.com/mpeyper/react-hooks-testing-library

packages/hooks/src/hooks.ts
hswolff: +1

packages/hooks/package.json
brookescarlett: lerna version bumps to 1.0.0 when we release, so @DesignerDave was in favor of starting off at 0.9.0, so the first release is 1.0.0 instead of 1.0.1

hswolff: Ah, didn't know that. Neat! TIL.

packages/hooks/src/hooks.ts
brookescarlett: `EventListener` 👍

packages/hooks/src/hooks.ts
brookescarlett: Done!

packages/hooks/src/index.ts
brookescarlett: Done!

packages/hooks/package.json
DesignerDave: Learnings from using Lerna publish hah

packages/hooks/src/hooks.ts
brookescarlett: Added TSDoc annotation! Hopping over to menu but will work on adding tests shortly :)

packages/hooks/src/useDocumentListener.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect } from 'react';
+
+/**
+ * Hook to subcribe to an event listener.
```
hswolff:
```suggestion
 * Hook to subscribe to an event listener.
```

packages/hooks/src/useDocumentListener.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param eventCallback Event listener callback function.
+ * @param enabled Determines whether or not the hook should run, defaults to true.
+ * @param options Optional parameter to specify options passed to the eventListener.
+ * @param dependencies Optional array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
+ */
+export default function useDocumentEventListener(
+  type: string,
+  eventCallback: EventListener,
+  enabled: boolean = true,
+  options?: object,
```
hswolff: Let's move this up in order.

Such that the default API of this mirrors the addEventListener API.

packages/hooks/src/useDocumentListener.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param type Represents the event type to listen for.
+ * @param eventCallback Event listener callback function.
+ * @param enabled Determines whether or not the hook should run, defaults to true.
+ * @param options Optional parameter to specify options passed to the eventListener.
+ * @param dependencies Optional array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
+ */
+export default function useDocumentEventListener(
+  type: string,
+  eventCallback: EventListener,
+  enabled: boolean = true,
```
hswolff: I can't decide if this should be the last argument, or second to last. I feel like last as it's most likely that you won't need to set this to false.

packages/hooks/src/useElementNode.ts
```diff
@@ -0,0 +1,16 @@
+import { useState, useCallback } from 'react';
+
+/**
+ * Hook to subscribe to changes in a ref element.
```
hswolff: Let's include an example of how to use this in these comments, cuz this hook is a little hard to understand by just looking at it.

packages/hooks/src/useDocumentListener.ts
brookescarlett: 👍

packages/hooks/src/useDocumentListener.ts
brookescarlett: 👍

packages/hooks/src/useDocumentListener.ts
brookescarlett: Moved to last arg!

packages/hooks/src/useElementNode.ts
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+      act(() => {
+        document.body.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+      });
+
+      expect(eventCallback).toHaveBeenCalledTimes(0);
+    });
+  });
+
+  // describe('useMutationObserver', () => {})
```
hswolff: remove? or have it be `describe.skip` so we track that it's missing?

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+  describe('useElementNode', () => {
```
hswolff: Can make a simple function component to test that it gets a reference to the element and that the function is re-ran.

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,62 @@
...
+        document.body.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+      });
+
+      expect(eventCallback).toHaveBeenCalledTimes(1);
+    });
+
+    test('event callback should not fire when enabled is false', () => {
+      const eventCallback = jest.fn();
+
+      renderHook(() => useDocumentListener('click', eventCallback, false));
```
hswolff: May need to update this now that the order has changed.

packages/hooks/src/useElementNode.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Example:
+ * ```
+ * const Example = () => {
```
hswolff: I think examples you don't need to include the `*`. Hover over the function to see how it's rendered.

packages/hooks/src/hooks.spec.js
brookescarlett: ^ yes did that

packages/hooks/src/useElementNode.ts
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
brookescarlett: 👍

packages/hooks/src/useDocumentListener.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Hook to subscribe to an event listener.
+ * @param type Represents the event type to listen for.
+ * @param eventCallback Event listener callback function.
+ * @param enabled Determines whether or not the hook should run, defaults to true.
+ * @param options Optional parameter to specify options passed to the eventListener.
+ * @param dependencies Optional array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
+ */
+export default function useDocumentEventListener(
```
hswolff:
```suggestion
export default function useDocumentListener(
```

Or align everything else with `useDocumentEventListener`.

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      });
+
+      expect(eventCallback).toHaveBeenCalledTimes(1);
+    });
+
+    test('event callback should not fire when enabled is false', () => {
+      const eventCallback = jest.fn();
+
+      renderHook(() =>
+        useDocumentListener({ type: 'click', eventCallback, enabled: false }),
```
hswolff: Whoops.

brookescarlett: 👍

packages/hooks/src/useDocumentListener.ts
brookescarlett: 👍

brookescarlett:
Not finished yet, need to figure out why debounce was breaking the useViewportSize tests

packages/hooks/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "description": "LeafyGreen UI Kit Custom Hooks",
+  "main": "./dist/index.js",
+  "scripts": {
```
DesignerDave: Don't forget to add the following script:
```
"ts:emit": "tsc --project ./tsconfig.json"
```

brookescarlett: 👍

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+
+      act(() => {
+        document.body.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+      });
+
+      expect(eventCallback).toHaveBeenCalledTimes(0);
+    });
+  });
+
+  describe.skip('useMutationObserver', () => {}); //eslint-disable-line jest/no-disabled-tests
```
DesignerDave: 🤔

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+
+  describe('useElementNode', () => {
+    let count = 1;
```
DesignerDave: Not sure it matters here, but could use `useState` instead...

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+      const [refEl, setRefEl] = useElementNode();
+      useEffect(() => {
+        count += 1;
+      }, [refEl]);
+      return <div ref={setRefEl} />;
+    }
+
+    render(<TestUseElementNode />);
+
+    test('it gets called twice', () => {
```
DesignerDave: Maybe I'm missing something, but it appears it only is called once (if this is passing). `count` starts at 1 (which is probably an anti-pattern).

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+      expect(count).toEqual(2);
+    });
+
+    const { result } = renderHook(() => useElementNode());
```
DesignerDave:
```js
const { result } = renderHook(useElementNode);
```

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+
+    test('returns an array where the first element refers to the ref node', () => {
```
DesignerDave: You're testing that the first element of the array is `null`, not a node.

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+  describe('useViewportSize', () => {
+    const { result, rerender } = renderHook(() => useViewportSize());
+
+    test('responds to updates in window width', () => {
+      act(() => {
+        window.width = 200;
+        window.dispatchEvent(new Event('resize'));
+        rerender();
+      });
+      setTimeout(() => {
```
DesignerDave: Is this timeout for the debounce? I'd make a comment indicating that.

Better yet, create a variable that helps explain the magical number here, and alsu use it below.

```js
const debounceTimeoutLength = 100
```

Or something similar.

DesignerDave: Also, should have a space above the timeout block (eslint should catch this).

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+        expect(result.current.width).toBe(200);
+      }, 100);
+
+      act(() => {
```
DesignerDave: given the amount of timeouts in here, this test feels brittle. I'd do one of two things with this test.

1. Rewrite as an async function with async / await to ensure proper call order.
2. Write this `act` block and the following `setTimeout` within the first `setTimeout` to ensure proper call order.

Both of these approaches also have the added benefit of allowing you to have the same `setTimeout` delay in both instances.

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+        window.dispatchEvent(new Event('resize'));
+        rerender();
+      });
+
+      setTimeout(() => {
+        expect(result.current.width).toBe(1024);
+      }, 200);
+    });
+
+    test('responds to updates in window height', () => {
```
DesignerDave: This test isn't testing anything that the last one isn't. By testing both width and height, this second test really is only testing that the window resize event is working.

That said, if you can break the hook in a way that fails one of these tests and not the other, let me know!

packages/hooks/src/useDocumentEventListener.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (!enabled) {
+      return;
+    }
+
+    document.addEventListener(type, eventCallback, options);
+
+    return () => {
+      document.removeEventListener(type, eventCallback);
+    };
+  }, dependencies);
```
DesignerDave: I think `enabled` should always be included as a dependency here.

```
}, [enabled, ...dependencies]);
```

packages/hooks/src/useDocumentEventListener.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param eventCallback Event listener callback function.
+ * @param enabled Determines whether or not the hook should run, defaults to true.
+ * @param options Optional parameter to specify options passed to the eventListener.
+ * @param dependencies Optional array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
+ */
+export default function useDocumentEventListener(
+  type: string,
+  eventCallback: EventListener,
+  options?: object,
+  dependencies?: Array<any>,
```
DesignerDave: We should probably default this to an empty array. This hook should only fire once really unless dependencies are explicitly set.

packages/hooks/src/useDocumentEventListener.ts
```diff
@@ -0,0 +1,29 @@
...
+  enabled: boolean = true,
+) {
+  useEffect(() => {
+    if (!enabled) {
+      return;
+    }
+
+    document.addEventListener(type, eventCallback, options);
```
DesignerDave: Some logic here to prevent setting multiple identical listeners would be a good idea.

packages/hooks/src/useElementNode.ts
DesignerDave: As a note, let's include an example by default in our tsdoc annotations.

packages/hooks/src/useMutationObserver.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Hook to subscribe to changes on the DOM.
+ * @param target HTML element that is subscribed to DOM changes.
+ * @param options Object with information about what DOM changes to subscribe to.
+ * @param callback Callback function to execute inside of MutationObserver instance.
+ * @param enabled Determines whether or not the hook should run, defaults to true.
+ */
+export default function useMutationObserver<Value>(
+  target: HTMLElement | null,
+  options: MutationObserverInit,
```
DesignerDave: Options should probably be optional, and come after `callback`.

DesignerDave: *Unless it's required for initialization

packages/hooks/src/useMutationObserver.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const observer = new MutationObserver((...args) => {
+      setValue(callback(...args));
+    });
+
+    if (target) {
+      observer.observe(target, options);
+    }
+
+    return () => observer.disconnect();
```
DesignerDave:
```js
return observer.disconnect
```

packages/hooks/src/hooks.spec.js
hswolff: async act is coming in 16.9.0 https://github.com/facebook/react/pull/14853

packages/hooks/src/useDocumentEventListener.ts
hswolff: That's a departure from the defaults of `useEffect` which seems like a bad thing to splinter from.

packages/hooks/src/useMutationObserver.ts
hswolff: I'd want to verify that `disconnect` is bound to the observer instance.

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,120 @@
...
+
+describe('packages/hooks', () => {
+  describe('useDocumentEventListener', () => {
+    test('event callback should fire when enabled is true', () => {
+      const eventCallback = jest.fn();
+
+      renderHook(() => useDocumentEventListener('click', eventCallback));
+
+      act(() => {
+        document.body.dispatchEvent(new MouseEvent('click', { bubbles: true }));
```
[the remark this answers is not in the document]
brookescarlett: Got this from the React test docs: https://reactjs.org/docs/test-utils.html#act

packages/hooks/src/hooks.spec.js
brookescarlett: useEffect to make sure it's capturing the lifecycle update

packages/hooks/src/hooks.spec.js
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
brookescarlett: I understand what you're saying, but felt like stronger coverage to write both. I can remove the second test block if you prefer though.

packages/hooks/src/useDocumentEventListener.ts
brookescarlett: I think we are safe but let me know: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Multiple_identical_event_listeners

packages/hooks/src/useMutationObserver.ts
brookescarlett: based on the docs it does look like options is optional, but @hswolff wrote this one so want to make sure he agrees

hswolff: Based on the docs it's optional but when you run it in the browser you get an error without the object present.

I think I also did it this way to force the user of this function to think about what they wanted to observe, so you didn't get the default behavior which may surprise a user.

image

packages/hooks/src/hooks.spec.js
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
brookescarlett: Very challenging to test without access to the DOM but open to ideas if you have any! Documented this way to show that it wasn't an oversight

packages/hooks/src/useDocumentEventListener.ts
brookescarlett: solved inspired by: https://usehooks.com/useEventListener/

packages/hooks/src/useDocumentEventListener.ts
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
DesignerDave: Yeah this can be tough – I'm not entirely sure we can't test this, but I'm fine with keeping this as-is for now if you add some additional comments around why this is skipped.

packages/hooks/src/hooks.spec.js
DesignerDave: It's additional maintenance burden – since you're directly dispatching the resize event in both, that's what triggers the hook. I don't see any added benefit in having both in here.

packages/hooks/src/useDocumentEventListener.ts
DesignerDave: Hmm... perhaps. I feel that in the majority of cases, we only will want to have this fire on mount, but perhaps having that be an explicit decision is okay.

packages/hooks/src/hooks.spec.js
```diff
@@ -45,35 +45,32 @@ describe('packages/hooks', () => {
...
 
   describe('useElementNode', () => {
-    let count = 1;
+    let count = 0;
 
     function TestUseElementNode() {
       const [refEl, setRefEl] = useElementNode();
       useEffect(() => {
-        count += 1;
+        ++count;
```
hswolff: I'd prefer we keep it as `count += 1` for readability. Prefix or postfix increments always confuse me.

packages/hooks/src/useDocumentEventListener.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    const callback = e => memoizedEventCallback.current(e);
+
+    document.addEventListener(type, callback, options);
+
+    return () => {
+      document.removeEventListener(type, eventCallback);
+    };
+  }, [enabled, ...(dependencies ? dependencies : [])]);
```
DesignerDave: Thinking about this, we might want
```js
...(dependencies ? dependencies : [])
```
to be
```js
...(dependencies ? dependencies : [type, eventCallback, options])
```

Also, see Harry's note in the comment on the type for `dependencies` as the conversation feels relevant.

packages/hooks/src/useMutationObserver.ts
DesignerDave: Okay, if it errors without an object present, I buy this! Thanks for the note Harry 👍

packages/hooks/src/useMutationObserver.ts
DesignerDave: Curious why this is relevant here? Having this being invoked within an arrow function shouldn't affect whether or not the method is bound to the instance, right?

Maybe I'm missing something here, but I was under the impression this change wouldn't fundamentally change how the function operates in this case. Are you afraid that it's setting something attached to the function scope that you want to persist when this is called?

packages/hooks/src/useDocumentEventListener.ts
hswolff: I'm so torn on this. Looking at [this hook that Brooke linked to](https://usehooks.com/useEventListener/) its default behavior is probably what you want 99% of the time. In fact I don't know when we wouldn't want this behavior.

i.e. re-run the event listeners if the 'type' or `enabled` field changes. We can't really do it on options as that'd be a new object every time, and may be tricky to memoize. Might have to do more useRef trickery. 

Wonder if the type signature for this should be `useDocumentEventListener(type, eventCallback, options, { dependencies, enabled, element = document })`. Such that the API is the same as `addEventListener` but you can configure it more in the last arg.

packages/hooks/src/useMutationObserver.ts
hswolff: Who is calling `disconnect`? I think it'd be React? 

It's the same issue as event handlers.

If you do `addEventListener('click', observer.disconnect)` then disconnect is being called not in the context of `observer`. From the event listener instead.

packages/hooks/src/useDocumentEventListener.ts
brookescarlett: If it is present, is it not fair to expect an array? Not defaulting to an empty array in the absence of the presence of the arg

packages/hooks/src/useMutationObserver.ts
DesignerDave: That's really annoying it's not bound to the instance... :/ Alright, fine as is I suppose then.

packages/hooks/src/useDocumentEventListener.ts
DesignerDave: Perhaps `element` isn't the best name for that option, but I see your point. Hmm... we're deviating a bit from the argument pattern used for memoized hooks with including `dependencies`, but perhaps that breaks the expectation it's following a similar signature and allows us to have different defaults.

brookescarlett: The thing about including dependencies is that we could assume them, but where we're invoking this hook we're deviating from what would be the assumed dependencies so we'd be hurting ourselves and also limiting future flexibility in that way.

brookescarlett: I like the signature proposed by @hswolff, and I don't think element is that problematic because if it's not document being passed, it'll be another element that a listener can be attached to.

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,97 @@
...
+      setTimeout(() => {
+        expect(count).toBe(2);
+      }, 10);
+    });
+  });
+
+  describe('useViewportSize', () => {
+    const { result, rerender } = renderHook(() => useViewportSize());
+    const debounceTimeoutLength = 100;
+    const secondDebounceTimeoutLength = 200;
```
DesignerDave: Shouldn't this just equal `debounceTimeoutLength * 2`? The less magic numbers, the better.

Should be fine to inline that where it's used as well.

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,97 @@
...
+      }, [refEl]);
+      return <div ref={setRefEl} />;
+    }
+
+    test('it gets called twice', () => {
+      act(() => {
+        render(<TestUseElementNode />);
+      });
+
+      // without setTimeout, test ran before final render
```
DesignerDave: This feels wrong – the docs don't seem to imply this executes asynchronously. timeouts with magic numbers (like `10`) are brittle, and may lead to false negatives down the line...

packages/hooks/src/hooks.spec.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  describe('useViewportSize', () => {
+    const { result, rerender } = renderHook(() => useViewportSize());
+    const debounceTimeoutLength = 100;
+    const secondDebounceTimeoutLength = 200;
+
+    test('responds to updates in window width', () => {
+      act(() => {
+        window.width = 200;
```
DesignerDave: We reuse these values in several places. Could we set this and `1024` to variables please?

packages/hooks/src/useDocumentEventListener.ts
DesignerDave: I'm game.

packages/hooks/src/hooks.spec.js
DesignerDave: +1

brookescarlett: 👍

packages/hooks/src/hooks.spec.js
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
brookescarlett: I think because it renders twice it takes longer, I'm not sure how to properly accommodate this in the tests though.

packages/hooks/src/hooks.spec.js
brookescarlett: 👍

packages/hooks/src/hooks.spec.js
DesignerDave: I'd do some research – I think it's important we address this properly to make sure our tests are reliable.

packages/hooks/src/useDocumentEventListener.ts
brookescarlett: Is there a reason to have the last arguments grouped in an object? Should we turn it into one object? Or can we just add the one extra argument (document) at the end? Right now I have it as five individual arguments, but curious re: thoughts on this.

brookescarlett: Just realized the latest version of this does not work w/ the Menu component. Once Menu is in going to rebase and play with it there

hswolff: Also we need to rename this to `useEventListener` now as it can listen to non-document elements.

My thinking is the common way you'll use this is like: `useEventListener('resize', handler)`. If I want to use advanced features then I can opt-in via:

```
useEventListener('resize', handler, { element: window, enabled: false ))
```

I'd actually be fine having the argument signature be `(type, handler, options)` and move all advanced options to the hash at the end. Cuz type and handler are required and everything else is optional.

packages/hooks/src/useDocumentEventListener.ts
```diff
@@ -28,10 +30,13 @@ export default function useDocumentEventListener(
...
-    document.addEventListener(type, callback, options);
+    element.addEventListener(type, callback, options);
 
     return () => {
       document.removeEventListener(type, eventCallback);
     };
-  }, [enabled, ...(dependencies ? dependencies : [])]);
+  }, [
+    enabled,
+    ...(dependencies ? dependencies : [type, eventCallback, options]),
```
hswolff: Let's have dependencies in the argument just be default to this: `useEventListener(type, callback, dependencies = [type, eventCallback, options])`.

brookescarlett: 👍

packages/hooks/src/useDocumentEventListener.ts
brookescarlett: I actually don't think that we should spread enabled into the dependencies. I get that it's intuitive that you'd want to listen for changes to that value, but I think that if we're letting users pass in their own array of dependencies, we shouldn't add anything to potentially break that. (Speaking from experience, this broke ours). I'd be open to including it as a default in dependencies, however, if a user passes nothing in for dependencies.

packages/Menu/src/Menu.tsx
```diff
@@ -82,32 +82,35 @@ function Menu({
...
-    'click',
-    nativeToggleEventHandler,
-    { once: true },
-    [isActive, trigger],
+  useEventListener('click', nativeToggleEventHandler, {
+    options: { once: true },
+    dependencies: [isActive, trigger],
     enabled,
-  );
+    element: document,
```
hswolff: Is this needed? Is it not the default?

packages/hooks/src/useEventListener.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param optional.dependencies Array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
+ * @param optional.enabled Determines whether or not the useEffect hook should run.
+ * @param optional.element Value to be passed as target of event handler, will default to document.
+ */
+export default function useEventListener(
+  type: string,
+  eventCallback: (e) => void,
+  {
+    options = undefined,
+    dependencies = [type, eventCallback],
```
hswolff: Default actually can just be `[enabled, type]` since we're using that `memoizedEventCallback` hack to not worry about `eventCallback` changing.

packages/Menu/src/Menu.tsx
brookescarlett: 👍

packages/hooks/src/useEventListener.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!enabled) {
+      return;
+    }
+
+    const callback = e => memoizedEventCallback.current(e);
+
+    element.addEventListener(type, callback, options);
+
+    return () => {
+      document.removeEventListener(type, eventCallback);
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/hooks/src/useEventListener.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param type Represents the event type to listen for.
+ * @param eventCallback Event listener callback function.
+ * @param optional Optional third argument passed to function with implementation specifications
+ * @param optional.options Parameter to specify options passed to the eventListener.
+ * @param optional.dependencies Array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
+ * @param optional.enabled Determines whether or not the useEffect hook should run.
+ * @param optional.element Value to be passed as target of event handler, will default to document.
+ */
+export default function useEventListener(
+  type: string,
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/Menu/src/Menu.tsx
```diff
@@ -82,32 +82,35 @@ function Menu({
     setActiveState(current => !current);
   }, []);
 
-  const nativeToggleEventHandler: EventListener = useCallback((e: Event) => {
-    e.stopImmediatePropagation();
-    setActiveState(current => !current);
-  }, []);
+  const nativeToggleEventHandler: EventListener = useCallback(
```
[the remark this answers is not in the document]
brookescarlett: This always should have been for closing, and I just realized it. The only thing that can trigger the opening of the menu is the trigger button itself, everything else should close the menu. Will rename now.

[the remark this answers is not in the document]
brookescarlett: 👍

packages/Menu/src/Menu.tsx
```diff
@@ -82,32 +82,35 @@ function Menu({
...
+    (e: Event) => {
+      e.stopImmediatePropagation();
+      setActiveState(false);
+    },
+    [isActive],
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/hooks/src/useEventListener.ts
brookescarlett: 👍

brookescarlett:
@hswolff im dumb, thank you for catching, just fixed

PR #77, opened 2019-06-04
Fix: added option to pass an id to toggle component

brookescarlett:
## ✍️ Proposed changes

Added option to pass an id of type number or string to toggle component. Tested in storybook passing an id as a prop and not passing an id as a prop and both options worked, also test suite still passing 👍 

🎟 _Jira ticket:_ [PD-307](https://jira.mongodb.org/browse/PD-307)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/Checkbox/README.md
+++ b/packages/Checkbox/README.md
@@ -104,7 +104,7 @@ Determines whether the text will be bold or not.
 
 **Default:** randomly generated string
 
-Adds an id to the input element and corresponding label element.
+Adds an ID only to the input, and it's used elsewhere for accessibility props.
 
 ### onChange
 
--- a/packages/Checkbox/README.md
+++ b/packages/Checkbox/README.md
@@ -98,6 +98,14 @@ Adds a className to the outermost element.
 
 Determines whether the text will be bold or not.
 
+### id
+
+**Type:** `string` || `number`
+
+**Default:** randomly generated string
+
+Adds an id to the input element and corresponding label element.
+
 ### onChange
 
 **Type:** `function`
--- a/packages/Toggle/src/Toggle.js
+++ b/packages/Toggle/src/Toggle.js
@@ -377,6 +377,7 @@ export default class Toggle extends PureComponent {
     className: PropTypes.string,
     onChange: PropTypes.func,
     name: PropTypes.string,
+    id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
   };
 
   static defaultProps = {
@@ -411,6 +412,7 @@ export default class Toggle extends PureComponent {
       disabled,
       size,
       variant,
+      id,
       ...rest
     } = this.props;
 
@@ -424,12 +426,12 @@ export default class Toggle extends PureComponent {
     return (
       <label
         className={cx(statefulStyles.container, className)}
-        htmlFor={this.checkboxId}
+        htmlFor={id || this.checkboxId}
       >
         <input
           {...toggleInput.prop}
           {...rest}
-          id={this.checkboxId}
+          id={id || this.checkboxId}
           className={inputStyle}
           type="checkbox"
           role="checkbox"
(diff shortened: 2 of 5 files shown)


packages/Toggle/src/Toggle.js
```diff
@@ -424,12 +426,12 @@ export default class Toggle extends PureComponent {
     return (
       <label
         className={cx(statefulStyles.container, className)}
-        htmlFor={this.checkboxId}
+        htmlFor={id || this.checkboxId}
```
hswolff: We missed `const labelId = `${this.checkboxId}-label`;` on line 181.

What if we turn `checkboxId` to a function and just include this branching inside there?

brookescarlett: I don't see what you are referring to on 181, but is this what you had in mind for `checkboxId`

`checkboxId = this.props.id || `toggle-${Math.floor(Math.random() * 10000000)}`;`

or

```const getCheckboxId = () => { 
  return this.props.id || `toggle-${Math.floor(Math.random() * 10000000)}`;`
}```

Comment (hswolff) [reply packages/Toggle/src/Toggle.js]: `const labelId = `${this.checkboxId}-label`;` line. That should use the passed in ID right?

It would have to be a function so it always reflects the latest value.

Comment (brookescarlett) [reply packages/Toggle/src/Toggle.js]: Oh that's on the Checkbox component! I was only updating Toggle... probably makes sense to update both though

Comment (DesignerDave) [reply packages/Toggle/src/Toggle.js]: @hswolff Could you clarify what you mean by "It would have to be a function so it always reflects the latest value."?

Wouldn't simply doing something like this before returning the JSX, and using this value be sufficient?
```js
const {id} = this.props
const checkboxId = id || this.checkboxId
```

hswolff: Having one location to get an ID prevents the whackamole of finding every usage of `this.checkbox` and prefixing it with `this.props.id`. It's not as much of an issue for Toggle, but is more so with Checkbox that has more usages of it.

DesignerDave: Even though checkbox is still a class, we don't use the IDs outside of the render method, and a const should be just fine in that case.

brookescarlett: @hswolff we fixed so that it's defined in only one location

packages/Toggle/src/Toggle.js
```diff
@@ -377,6 +377,7 @@ export default class Toggle extends PureComponent {
     className: PropTypes.string,
     onChange: PropTypes.func,
     name: PropTypes.string,
+    id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
DesignerDave: I personally think only permitting strings would be fine. Using numbers is a pretty bad practice with IDs anyways.

packages/Checkbox/README.md
```diff
@@ -98,6 +98,14 @@ Adds a className to the outermost element.
 
 Determines whether the text will be bold or not.
 
+### id
+
+**Type:** `string` || `number`
+
+**Default:** randomly generated string
+
+Adds an id to the input element and corresponding label element.
```
DesignerDave: Let's be more accurate here – It adds an ID only to the input, and it's used elsewhere for accessibility props.

packages/Toggle/src/Toggle.js
hswolff: I had the same thought, but in practice this may be more of a nuisance than necessary, as React does the conversion internally anyways.

packages/Checkbox/README.md
brookescarlett: 👍

packages/Toggle/src/Toggle.js
brookescarlett: Should we accept both or just the string?

PR #78, opened 2019-06-04
PD299: Port <Button> to TypeScript

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD299: Port `<Button>` to TypeScript](https://jira.mongodb.org/browse/PD-299)

## 🛠 Types of changes

## ✅ Checklist


packages/Button/src/Button.spec.tsx
```diff
@@ -16,8 +16,20 @@ describe('packages/Button', () => {
     </Button>,
   );
 
+  function isElement(el: Node): el is HTMLElement {
+    return el.nodeType === Node.ELEMENT_NODE;
+  }
+
+  function isButton(el: Node): el is HTMLButtonElement {
+    return isElement(el) && el.tagName.toLowerCase() === 'button';
+  }
```
jetpacmonkey: These helper typeguards could probably be extracted somewhere, but I'm not sure where that should be

packages/Button/src/Button.tsx
```diff
@@ -3,10 +3,24 @@ import PropTypes from 'prop-types';
...
+
+/** Helper type to extract an HTML element's valid props */
+type HTMLElementProps<
+  Element extends keyof JSX.IntrinsicElements
+> = JSX.IntrinsicElements[Element] extends React.DetailedHTMLProps<
+  infer Props,
+  any
+>
+  ? Props
+  : never;
```
jetpacmonkey: Another helper that could be extracted for reuse to... somewhere?

packages/Button/src/Button.tsx
```diff
@@ -267,61 +281,124 @@ const baseStyle = css`
...
 }
+
+/*
+NOTE(JeT):
+When this is defined inside the actual component, React will try to infer prop types from it,
+merging them together with ButtonProps (see LibraryManagedAttributes and MergePropTypes in @types/react).
+Unfortunately, this merging uses keyof, which appears to drop the [key: string] index signature, meaning TS won't 
+allow us to pass unrecognized props down to custom components when we're using the `as` prop. This workaround avoids
+the attempt at merging, while still getting runtime type-checking for non-TS consumers of the library.
+*/
```
jetpacmonkey: I want to draw special attention to this comment, because it was a weird confusing situation that I'd like to make sure I'm explaining well. Of special note, the two `@types/react` definitions I refer to:
[LibraryManagedAttributes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/932108dd26567dd5cfbdcc7b80b2c37e39af36b9/types/react/index.d.ts#L2818) and [MergePropTypes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/932108dd26567dd5cfbdcc7b80b2c37e39af36b9/types/react/index.d.ts#L2772)

packages/Button/src/Button.tsx
```diff
@@ -267,61 +281,124 @@ const baseStyle = css`
...
+  as: React.ElementType<any>;
+  [key: string]: any;
+};
+
+type ButtonProps =
+  | LinkButtonProps
+  | ButtonButtonProps
+  | CustomElementButtonProps;
+
+function hasCustomRoot(props: ButtonProps): props is CustomElementButtonProps {
```
hswolff: Little confused by this naming. What do you mean by root? Is it just `isCustomElement` ?

packages/Button/src/Button.tsx
```diff
@@ -267,61 +281,124 @@ const baseStyle = css`
...
+    if (hasLinkRoot(this.props)) {
+      const rest: HTMLElementProps<'a'> = omit(this.props, omitProps);
+      return (
+        <a {...rest} {...commonProps}>
+          {children}
+        </a>
+      );
+    }
+
+    const rest: HTMLElementProps<'button'> = omit(this.props, omitProps);
```
hswolff: We have to split this line up into 3 places for type reasons? Boo. We can't just cast it as appropriate? Seems silly to dupe it.

packages/Button/src/Button.tsx
```diff
@@ -267,61 +281,124 @@ const baseStyle = css`
...
-        className={cx(
-          baseStyle,
-          buttonSizes[size],
-          buttonVariants[variant],
-          className,
-        )}
-        disabled={disabled}
-        aria-disabled={disabled}
-      >
+      <button type="button" {...rest} {...commonProps}>
```
hswolff: Could we have: `const { type = 'button' } = this.props` when we first desctructure?

jetpacmonkey: I did it here because links/custom components might not have a `type` field

packages/Button/src/Button.tsx
jetpacmonkey: The existing code had called the component variable `Root` so I was just going off of that. `usesCustomElement` could maybe be a good name? `isCustomElement` would suggest to me that the argument passed into it is being checked, when instead it's the `props` object.

packages/Button/src/Button.spec.tsx
DesignerDave: So I haven't seen this `is` syntax yet. If I'm understanding correctly, the `(type): type is someType` syntax is interpreted as: "if type is someType, this function will return true".

If so, this is sweet.

DesignerDave: Also, we can probably break these out into `lib` at a later time (I think we talked a bit about this in the channel?).

packages/Button/src/Button.tsx
jetpacmonkey: Oh, I can do `{...rest as HTMLElementProps<'button'>}`. That seems nicer, I guess. I'll do that.

packages/Button/src/Button.spec.tsx
```diff
@@ -16,8 +16,20 @@ describe('packages/Button', () => {
...
+
   const renderedButton = renderedComponent.container.firstChild;
 
+  if (renderedButton == null || !isElement(renderedButton)) {
```
DesignerDave: There a reason why we aren't simply checking this?

```js
if (!isElement(renderedButton)) {
```

`isElement` should return false here if `renderedButton` is nullish. Does TS complain that `renderedButton` could be nullish and doesn't match the `Node` type?

packages/Button/src/Button.story.tsx
```diff
@@ -0,0 +1,145 @@
...
+        ) as any
+      }
+      className={buttonClass}
+    >
+      {text('Children', 'Button')}
+    </Button>
+  ))
+  .add('Info', () => (
+    <Button
+      size={select('Size', Object.values(Size) as Array<Size>, Size.Normal)}
```
DesignerDave: Wooo enums! 😁

packages/Button/src/Button.tsx
```diff
@@ -3,11 +3,36 @@ import PropTypes from 'prop-types';
...
-const buttonVariants = {
-  default: css`
+export enum Variant {
+  Default = 'default',
+  Primary = 'primary',
+  Info = 'info',
+  Danger = 'danger',
+  Dark = 'dark',
+}
+export enum Size {
```
DesignerDave: Small nit – Line break above here.

Eslint is supposed to auto-correct this sort of thing, but I don't think we have it properly configured for TS.

packages/Button/src/Button.tsx
```diff
@@ -267,61 +292,126 @@ const baseStyle = css`
...
+  | ButtonButtonProps
+  | CustomElementButtonProps;
+
+function usesCustomElement(
+  props: ButtonProps,
+): props is CustomElementButtonProps {
+  return (props as any).as != null;
+}
+
+function hasLinkRoot(
```
hswolff:
```suggestion
function usesLinkElement(
```

packages/Button/src/Button.spec.tsx
jetpacmonkey: Yup, and yes! [Docs for User-Defined Type Guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)

packages/Button/src/Button.tsx
hswolff: Good comment!

packages/Button/src/Button.tsx
jetpacmonkey: ah shoot, yeah good call

packages/Button/src/Button.tsx
DesignerDave: Buttons should not always be `type="button"`. `type="submit"` is a good example where we might not want to explicitly set this to "button".

packages/Button/src/Button.tsx
```diff
@@ -267,61 +292,124 @@ const baseStyle = css`
...
-      href,
-      ...rest
-    } = this.props;
+      'aria-disabled': disabled,
+    };
 
-    let Root = href ? 'a' : 'button';
+    if (hasCustomRoot(this.props)) {
+      const Root = this.props.as;
+      const rest = omit(this.props, omitProps);
```
DesignerDave: I might be missing something... how come we can't just use object `rest` syntax?

packages/Button/src/Button.tsx
```diff
@@ -267,61 +292,124 @@ const baseStyle = css`
...
+/*
+NOTE(JeT):
+When this is defined inside the actual component, React will try to infer prop types from it,
+merging them together with ButtonProps (see LibraryManagedAttributes and MergePropTypes in @types/react).
+Unfortunately, this merging uses keyof, which appears to drop the [key: string] index signature, meaning TS won't 
+allow us to pass unrecognized props down to custom components when we're using the `as` prop. This workaround avoids
+the attempt at merging, while still getting runtime type-checking for non-TS consumers of the library.
+*/
+// @ts-ignore
+Button.propTypes = {
```
DesignerDave: Why the `@ts-ignore`?

packages/Button/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import Button, { Size, Variant } from './Button';
+export default Button;
```
DesignerDave: Tiny nit: Prefer having the default export last.

packages/Button/src/Button.tsx
jetpacmonkey: Order matters with spread props. I tried to call this out in the comment right above this line, but if a consumer passes in a `type` prop, it will be part of `rest`, and will override my `type="button"` here.

jetpacmonkey: I'll add a test to make sure `type` is respected, if that helps :)

DesignerDave: 🤦‍♂ Right, my bad! Alternatively, could we not set `type: 'button'` in default props?

packages/Button/src/Button.spec.tsx
jetpacmonkey: If I had defined `isElement` as taking `Node | null` then that'd be fine. I guess that's probably a more useful typeguard defined that way, I'll change it.

packages/Button/src/Button.tsx
jetpacmonkey: TS complains that `propTypes` doesn't exist on `Button` otherwise. I'm only adding it here (as opposed to defining it as a `static`) to avoid TS seeing it (see comment above this line for my best understanding of what's going on there), so ignoring the type error seems like it shouldn't be an issue.

DesignerDave: We've had issues before with setting properties on components after they're exported. Might the error you're seeing be resolved by exporting after setting the `propTypes`?

packages/Button/src/Button.tsx
jetpacmonkey: If I try to pull off the `as` prop while destructuring, it will complain because `as` doesn't exist on all of the union members that make up `ButtonProps`. So that means if I use object rest syntax with the destructuring, `rest` is going to contain `as`. This was the best pattern I could come up with to get around that 🤷‍♂

packages/Button/src/Button.tsx
jetpacmonkey: side-note, I wish I could've just committed this from inside GitHub's UI, but alas... GitHub doesn't have TypeScript's understanding of renames...

packages/Button/src/Button.tsx
jetpacmonkey: That's how it was before, actually. It was probably fine that way, but it would still be passing along `type="button"` for `<a>` tags, which we probably don't want? I found out yesterday that `type` is a valid property on `<a>` tags, meant to be used to hint to the browser the MIME type of the linked URI, so we probably don't want to set it to `"button"`, although I guess it most likely wouldn't hurt anything.

packages/Button/src/Button.tsx
DesignerDave: Not sure this is better than doing something like this prior to returning the JSX:
```
delete rest.as;
```

Thoughts @hswolff?

packages/Button/src/Button.tsx
DesignerDave: Good point – type does have a very different meaning for `<a>` tags. On a tangential note, should we be setting `role="button"` when rendering as `<a>`?

packages/Button/src/Button.tsx
jetpacmonkey: Nope, same error. TS sees that when I defined the `Button` class I didn't have any mention of a `propTypes` static member, so I can't assign anything to it. Although... I could probably make it work by doing `static propTypes: any = { ...` and get around the issue with React inferring things that way... That feels like a better solution, I'm gonna do that.

packages/Button/src/Button.tsx
DesignerDave: Question: Does the HTMLElementProps type allow for `data-*` and `aria-*` props?

packages/Button/src/Button.tsx
jetpacmonkey: Done. That feels much less intrusive to me.

DesignerDave: 🍻

packages/Button/src/Button.tsx
jetpacmonkey: I'd expect it to, although I don't know exactly how. But as far as I understand, it's accessing the same types as the core JSX parser with a `<div>` or whatever

PR #79, opened 2019-06-04
PD-302: Port <Portal> to TypeScript

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-302](https://jira.mongodb.org/browse/PD-302)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/Portal/src/Portal.spec.tsx
+++ b/packages/Portal/src/Portal.spec.tsx
@@ -14,8 +14,8 @@ describe('packages/Portal', () => {
     </div>,
   );
 
-  function isElement(el: Node): el is HTMLElement {
-    return el.nodeType === Node.ELEMENT_NODE;
+  function isElement(el: Node | null): el is HTMLElement {
+    return el != null && el.nodeType === Node.ELEMENT_NODE;
   }
 
   function getByIdOrThrow(id: string) {
@@ -31,11 +31,11 @@ describe('packages/Portal', () => {
   test(`appends portal content to document body`, () => {
     const { firstChild, lastChild } = document.body;
 
-    if (firstChild == null || !isElement(firstChild)) {
+    if (!isElement(firstChild)) {
       throw new Error('Could not find firstChild element');
     }
 
-    if (lastChild == null || !isElement(lastChild)) {
+    if (!isElement(lastChild)) {
       throw new Error('Could not find lastChild element');
     }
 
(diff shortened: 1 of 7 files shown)


packages/Portal/src/Portal.spec.tsx
```diff
@@ -14,12 +14,37 @@ describe('packages/Portal', () => {
...
+
+    return el;
+  }
+
   test(`appends portal content to document body`, () => {
-    expect(document.body.firstChild.outerHTML).toBe(
+    const firstChild = document.body.firstChild;
+    const lastChild = document.body.lastChild;
+
+    if (firstChild == null || !isElement(firstChild)) {
```
hswolff: Ditto other PR - why throws and not assertions?

packages/Portal/src/Portal.spec.tsx
```diff
@@ -14,12 +14,37 @@ describe('packages/Portal', () => {
...
+    if (el == null) {
+      throw new Error(`Could not find element: ${id}`);
+    }
+
+    return el;
+  }
+
   test(`appends portal content to document body`, () => {
-    expect(document.body.firstChild.outerHTML).toBe(
+    const firstChild = document.body.firstChild;
```
hswolff:
```suggestion
    const { firstChild, lastChild } = document.body;
```

😱

jetpacmonkey: Yeesss

packages/Portal/src/Portal.spec.tsx
jetpacmonkey: Ditto other PR response: TypeScript type narrowing

packages/Portal/src/Portal.spec.tsx
```diff
@@ -14,12 +14,36 @@ describe('packages/Portal', () => {
     </div>,
   );
 
+  function isElement(el: Node): el is HTMLElement {
```
DesignerDave: Will be nice once we move this to `lib`.

jetpacmonkey: I guess I should make the same `| null` improvement here that I did in the other branch. Definitely looking forward to having a shared util :)

DesignerDave: Indeed

PR #80, opened 2019-06-04
PD-298: Port <Badge> to TypeScript

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-298](https://jira.mongodb.org/browse/PD-298)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/Badge/src/Badge.spec.tsx
+++ b/packages/Badge/src/Badge.spec.tsx
@@ -9,6 +9,10 @@ describe('packages/Badge', () => {
   const className = 'test-pill-class';
   const child = 'Bubble Pill';
 
+  function isElement(node: Node): node is Element {
+    return node.nodeType === Node.ELEMENT_NODE;
+  }
+
   const { container } = render(
     <Badge className={className} onClick={onClick}>
       {child}
@@ -16,6 +20,11 @@ describe('packages/Badge', () => {
   );
 
   const badge = container.firstChild;
+
+  if (badge == null || !isElement(badge)) {
+    throw new Error('Badge element not found');
+  }
+
   test(`renders "${className}" in the button's classList`, () => {
     expect(badge.classList.contains(className)).toBe(true);
   });
--- a/packages/Badge/src/Badge.story.tsx
+++ b/packages/Badge/src/Badge.story.tsx
@@ -2,10 +2,11 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { text, select } from '@storybook/addon-knobs';
 import Badge from './Badge';
+import { Variant } from '.';
 
 storiesOf('Badge', module).add('Default', () => (
   <Badge
-    variant={select(
+    variant={select<Variant>(
       'Variant',
       [
         'default',
--- a/packages/Badge/src/Badge.story.tsx
+++ b/packages/Badge/src/Badge.story.tsx
@@ -6,19 +6,10 @@ import { Variant } from '.';
 
 storiesOf('Badge', module).add('Default', () => (
   <Badge
-    variant={select<Variant>(
+    variant={select(
       'Variant',
-      [
-        'default',
-        'danger',
-        'warning',
-        'darkBlue',
-        'lightBlue',
-        'primary',
-        'outline',
-        'dark',
-      ],
-      'darkBlue',
+      Object.values(Variant) as Array<Variant>,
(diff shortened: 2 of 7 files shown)


packages/Badge/src/Badge.spec.tsx
```diff
@@ -9,13 +9,22 @@ describe('packages/Badge', () => {
...
+
   const { container } = render(
     <Badge className={className} onClick={onClick}>
       {child}
     </Badge>,
   );
 
   const badge = container.firstChild;
+
+  if (badge == null || !isElement(badge)) {
```
hswolff: Can this just be a `test()`? Or is this needed for TS?

packages/Badge/src/Badge.spec.tsx
```diff
@@ -9,13 +9,22 @@ describe('packages/Badge', () => {
   const className = 'test-pill-class';
   const child = 'Bubble Pill';
 
+  function isElement(node: Node): node is Element {
```
hswolff: We should probably put these type of functions inside `lib` package? Maybe make a ticket to "Move all type utils to  `@leafygreen-ui/lib/types`"?

packages/Badge/src/Badge.tsx
```diff
@@ -5,6 +5,16 @@ import { ccClassName, emotion } from '@leafygreen-ui/lib';
 
 const { css } = emotion;
 
+export type Variant =
```
hswolff: What if this was a string enum instead?

packages/Badge/src/Badge.spec.tsx
jetpacmonkey: Needed for TS type narrowing (discussed in Slack)

packages/Badge/src/Badge.spec.tsx
jetpacmonkey: Works for me

packages/Badge/src/Badge.tsx
jetpacmonkey: I don't really have a preference between unions of strings and enums. This component was already using strings for variants rather than an object, so I went with a union, but if we want to standardize on one pattern I can change it

packages/Badge/src/Badge.spec.tsx
jetpacmonkey: So we want a separate ticket, as opposed to doing it as part of this one, then?

packages/Badge/src/Badge.tsx
jetpacmonkey: Talked with @DesignerDave in Slack, sounds like having the enum in the emitted js file is enough of a selling point to pick string enums

packages/Badge/src/Badge.spec.tsx
hswolff: Up to you.

jetpacmonkey: Filed [PD-313](https://jira.mongodb.org/browse/PD-313)

packages/Badge/src/Badge.story.tsx
```diff
@@ -2,22 +2,14 @@ import React from 'react';
...
-        'danger',
-        'warning',
-        'darkBlue',
-        'lightBlue',
-        'primary',
-        'outline',
-        'dark',
-      ],
-      'darkBlue',
+      Object.values(Variant) as Array<Variant>,
```
DesignerDave: Interesting... Does TS not infer `as Array<Variant>`?

packages/Badge/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import Badge, { Variant } from './Badge';
+export default Badge;
```
DesignerDave: Nit: Prefer default exports last.

packages/Badge/src/Badge.story.tsx
jetpacmonkey: Nope, explanation [here](https://github.com/microsoft/TypeScript/pull/12253#issuecomment-263132208) for why `Object.keys`, `Object.values`, and `Object.entries` doesn't infer quite as much as you might think. Although it'd be nice if they could make an exception for enums...

DesignerDave: Shame, thanks for the explanation!

PR #82, opened 2019-06-06
PD-300: Port <Checkbox> to TypeScript

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-300: Port `<Checkbox>` Component to TypeScript](https://jira.mongodb.org/browse/PD-300)

## 🛠 Types of changes

## ✅ Checklist


packages/Checkbox/src/Checkbox.spec.tsx
```diff
@@ -70,9 +87,18 @@ describe('packages/Checkbox', () => {
...
+    }
+
     const uncontrolledCheckbox = uncontrolledContainer.children[0];
 
-    fireEvent.click(uncontrolledContainer.firstChild);
+    if (!isInput(uncontrolledCheckbox)) {
+      throw new Error('Could not find checkbox input element');
+    }
+
+    fireEvent.click(uncontrolledCheckbox);
```
jetpacmonkey: Could someone confirm for me that this is doing the right thing? `.firstChild` should be the same as `.children[0]` afaik, but the tests say "when the label is clicked", not "when the checkbox is clicked" so I wanna make sure I'm not missing something.

packages/Checkbox/src/Checkbox.tsx
```diff
@@ -138,19 +156,26 @@ export default class Checkbox extends PureComponent {
   state = { checked: false };
 
   componentDidMount() {
+    if (this.inputRef.current == null) {
```
hswolff: This should just be for TS sake, right? I wonder if we should do `this.inputRef.current!.indeterminate`. If we start getting errors then we should switch away from `React.createRef` and use a function instead.

packages/Checkbox/src/Checkbox.tsx
```diff
@@ -160,12 +185,14 @@ export default class Checkbox extends PureComponent {
     // For Microsoft Edge and IE, when checkbox is indeterminate, change event does not fire when clicked.
     // Explicitly call onChange for this case
     if (this.inputRef.current && this.inputRef.current.indeterminate) {
-      this.onChange(e);
+      // Event bubbling means that TS can't make assumptions about the event's target.
```
hswolff: Seems better to override TypeScript than modify the event we're sending to the handler.

packages/Checkbox/src/Checkbox.story.tsx
```diff
@@ -18,7 +27,7 @@ class Control extends PureComponent {
 
   state = { checked: false };
 
-  onChange = e => {
+  onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
jetpacmonkey: I think I just default to annotating params/returns for functions, rather than annotating the whole function as a single value. They should be essentially equivalent, outside of `ChangeEventHandler` mandating that the function doesn't have a return value.

packages/Checkbox/src/Checkbox.tsx
jetpacmonkey: so `this.onChange(e as unknown as React.ChangeEvent<HTMLInputElement>)`? That doesn't feel better to me, but... 🤷‍♂

hswolff: Can you omit `as unknown`? I think that's uglier but more 'correct' as far as the application code is concerned.

packages/Checkbox/src/Checkbox.story.tsx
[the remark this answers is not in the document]
DesignerDave: Wouldn't that make `ChangeEventHandler` the preferred way to type handlers like this?

packages/Checkbox/src/Checkbox.tsx
jetpacmonkey: without the `as unknown` it complains

> Conversion of type 'MouseEvent<HTMLInputElement, MouseEvent>' to type 'ChangeEvent<HTMLInputElement>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

packages/Checkbox/src/Checkbox.story.tsx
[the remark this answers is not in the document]
jetpacmonkey: I don't think it matters. I would say that the provided handler would be the way to go for defining a prop, except that I'd prefer it have a return type of `unknown` instead of `void` (since afaik any return values will just be ignored). For something that's used internally, I don't see any real reason to prefer one pattern over the other.

packages/Checkbox/src/Checkbox.tsx
jetpacmonkey: I'm not 100% sure I agree about it being "more correct". Change events are guaranteed to have the input as the event's `target` (according to [this comment](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/12239#issuecomment-256308813) at least), but click events don't have that guarantee.

I think I might've stumbled upon a more technically-correct solution, actually:

```tsx
onClick = (e: React.MouseEvent<HTMLInputElement> & React.ChangeEvent<HTMLInputElement>) => {
  ...
```

jetpacmonkey: Or more precisely...

```ts
e: React.MouseEvent<HTMLInputElement> & { target: HTMLInputElement }
```

packages/Checkbox/src/Checkbox.story.tsx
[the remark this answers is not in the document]
DesignerDave: My point here I suppose is: should this be the pattern we use? Best to keep things uniform if possible.

PR #83, opened 2019-06-06
PD-316: Add new UI palette to leafygreen-ui

DesignerDave:
## ✍️ Proposed changes

This PR adds the new UI palette to leafygreen-ui, and removes the entry for the old theme's storybook in preparation for removal.

🎟 _Jira ticket:_ [Create new Palette component and deprecate theme](https://jira.mongodb.org/browse/PD-316)

## 🛠 Types of changes

## ✅ Checklist


packages/palette/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@leafygreen-ui/palette",
+  "version": "0.9.0",
+  "description": "LeafyGreen UI Palettes",
+	"main": "./dist/index.js",
+	"types": "./dist/index.d.ts",
+  "scripts": {
+		"build": "../../node_modules/.bin/webpack --config ../../webpack.config.js",
```
hswolff: This isn't formatted according to prettier.

Let's update `prettier` and `prettier:check` to include `json` as an extension it checks to prevent future regressions.

packages/palette/package.json
```diff
@@ -0,0 +1,18 @@
...
+		"ts:emit": "tsc --project ./tsconfig.json"
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/lib": "^2.1.7"
```
hswolff: What do we need lib for?

packages/palette/src/_ui-colors.less
```diff
@@ -0,0 +1,34 @@
+@ui-color--black: #061621;
```
hswolff: Want to get your arms dirty with some node scripting?

We can extend the `build` command for this package to include `&& generateStyleSheets.js`

In which `generateStyleSheets.js` is a homegrown node script that will read in `uiColors.ts` (may be a little tricky to get this to work with TS, may be fine to use [ts-node](https://github.com/TypeStrong/ts-node)) and using a plain template output the .less and .sass files for us. 

Additionally when consuming this from LESS I believe someone will have to do: `@import "node_modules/@leafygreen-ui/palette/src/_ui-colors.less`.

I wonder if we should have this generated and put into `styles/ui-colors.less` and have that folder git ignored as well. So the import path looks a little clearer. I'm also fine with `dist/ui-colors.less`. Not sure we need to do the underscore prefix as we're supplying it as a module, but not sure what consumers would prefer.

packages/palette/src/uiColors.ts
```diff
@@ -0,0 +1,54 @@
...
+  dark2: string;
+  dark1?: string;
+  base: string;
+  light1?: string;
+  light2: string;
+  light3: string;
+}
+
+export const gray: HueRange = {
+  dark3: '#21313C',
```
hswolff: Alternative API:

`uiColors.dark[0]` so it's actually an array of values. 

I don't actually like it more than what you have but wanted to brainstorm aloud.

packages/palette/README.md
```diff
@@ -0,0 +1,71 @@
+# Palette
+
+![npm (scoped)](https://img.shields.io/npm/v/@leafygreen-ui/palette.svg)
+
+This package contains the colors for the following palettes:
+
+- **UI**: MongoDB-branded, general-use UI palette.
+- _Chart (coming soon)_: Color palette for accessible charting colors.
```
hswolff: What do you envision the API for these will be?

We'll have `import { uiColors, chartColors, syntaxColors } from '@leafygreen-ui/palette';` ?

Seems redundant to have colors appended to every value. 

I'm not sure there's anything better.

Could do `import palette from '.'` and then have `palette.ui` etc. But then you will always include all colors. Which may be ok as it's small?

Thoughts?

packages/palette/package.json
DesignerDave: Emotion and such is used for the Storybook entry.

packages/palette/src/uiColors.ts
DesignerDave: These names directly map to the color palette framework that we built that designers will also be using. Keeping these names will make sure nothing is lost in communication.

packages/palette/src/_ui-colors.less
DesignerDave: So... I was thinking about that but given the infrequency of changing these, I actually think that it would be less maintenance burden to simply write these separately. That, and transforming the names correctly would be a PITA and more brittle than this.

We could definitely copy the file during the build though.

packages/palette/README.md
DesignerDave: I thought a bunch about this. I _think_ this is the best solution to avoid everyone's imports looking like:

```js
import { ui as uiColors } from '@leafygreen-ui/palette';
```

I think it's important to keep the three palettes separate, and usually you will only be importing `uiColors`. `syntaxColors` will be used mostly internally in leafygreen-ui, and chart colors is pretty niche in where it should be used as well.

packages/palette/package.json
DesignerDave: Fixed and ran prettier!

packages/palette/README.md
hswolff: Makes sense to me!

packages/palette/package.json
hswolff: This looks still wrong to me in the diff....But it passes the checks...but if I check it out locally and run format in my editor it does change it.

packages/palette/package.json
hswolff: So devDependency?  But aren't those in the root `package.json` so no need to have it explicitly here?

packages/palette/src/palette.story.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import styled from 'react-emotion';
```
hswolff: We should at least upgrade to the latest version of emotion and use that instead. Such that this would be `@emotion/styled`.

packages/palette/package.json
DesignerDave: That's fair enough. I actually will need to add it to the root `package.json` since it's not there yet.

packages/palette/package.json
DesignerDave: 👍

packages/palette/src/palette.story.tsx
DesignerDave: Whoops, you're right!

packages/palette/src/_ui-colors.less
DesignerDave: Omitting Less for now, as per our Slack convo.

packages/palette/src/uiColors.ts
```diff
@@ -0,0 +1,54 @@
+export const white = '#FFFFFF';
+export const black = '#061621';
```
hswolff: Doesn't look so black to me. ;) (this is a joke)

PR #84, opened 2019-06-07
PD-301: Port <Icon> to TypeScript

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-301: Port <Icon> Component to TypeScript](https://jira.mongodb.org/browse/PD-301)

## 🛠 Types of changes

## ✅ Checklist


packages/Icon/src/index.ts
```diff
@@ -2,4 +2,5 @@ import createIconComponent from './createIconComponent';
 import glyphs from './glyphs';
 
 export { glyphs, createIconComponent };
+export { Size, IconProps } from './createIconComponent';
```
hswolff: Why exporting `IconProps`?

packages/Icon/src/Icon.spec.tsx
```diff
@@ -6,12 +6,20 @@ afterAll(cleanup);
 
 const text = 'Hello world';
 
+// eslint-disable-next-line react/display-name
+const MyGlyph: React.FC<React.SVGProps<SVGSVGElement>> = () => (
```
DesignerDave: Testing this component was interesting...

I don't know if this will cause problems, but maybe the fake glyph should render an SVG if we're typing it as returning such...

typings/images.d.ts
```diff
@@ -13,7 +13,17 @@ declare module '*.gif' {
   export = value;
 }
 
+declare namespace SVGR {
```
DesignerDave: Not sure this belongs in `images.d.ts`...

jetpacmonkey: I'll take a look at this again. I had it in a separate file and for some reason I couldn't get it to work, but I feel like I must have been overlooking something obvious...

packages/Icon/src/Icon.spec.tsx
jetpacmonkey: Technically I'm pretty sure that all it's claiming here is that it can accept the same props as an SVG element, and that if any event handlers are called `event.target` will be an SVG. None of that is inaccurate, exactly...

packages/Icon/src/index.ts
jetpacmonkey: You know... I'm sure I had a reason for this when I wrote it, but I can't think why it'd be any different than any other component now...

jetpacmonkey: Oh, well I found the reason pretty quick when I went to make it private. The `export default createIconComponent(glyphs)` line throws an error when `IconProps` is private:

> Default export of the module has or is using private name 'IconProps'

I suppose I could export it from `Icon.tsx` leave it out from `index.ts`, though.

jetpacmonkey: Explanation here: https://github.com/Microsoft/TypeScript/issues/6307#issuecomment-168321011

typings/images.d.ts
jetpacmonkey: I think I must have named the file `svgr.ts` instead of `svgr.d.ts` in my first try, because it all seemed to work beautifully when I tried again. Pushed up a fix.

DesignerDave: 👍

packages/Icon/src/Icon.spec.tsx
DesignerDave: Fair enough!

PR #85, opened 2019-06-08
PD-305: Port <Toggle> to TypeScript

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-305: Port <Toggle> to TypeScript](https://jira.mongodb.org/browse/PD-305)

## 🛠 Types of changes

## ✅ Checklist


packages/Checkbox/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "LeafyGreen UI Kit Checkbox",
   "main": "./dist/index.js",
   "scripts": {
-    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js"
+    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js",
+    "ts:emit": "tsc --project ./tsconfig.json"
```
jetpacmonkey: Drive-by fix adding a line that I missed in #82

jetpacmonkey: Also... Is there some way we could lint for this?

packages/Toggle/src/Toggle.tsx
```diff
@@ -40,19 +51,26 @@ const focusStateStyle = css`
   }
 `;
 
-const getContainerStyles = ({ size, disabled }) => {
-  const sizeStyle = {
-    default: css`
+interface StateForStyle {
```
jetpacmonkey: I don't love the name `StateForStyle`. Any suggestions on a better one?

packages/Checkbox/package.json
hswolff: Would be neat to make an eslint rule for this. Or just a custom script to validate that we have this field present in a package.json file. But yeah agree, annoying we have to manually catch this.

packages/Toggle/src/Toggle.tsx
hswolff: Meh, it works for me. Can't think of anything better either.

packages/Checkbox/package.json
DesignerDave: Even better – re-explore removing these redundant scripts. They're used to work around lerna, but it would be great to revisit this to remove the `build`/`ts:emit` scripts, as well as the package-specific `tsconfig.json`.

packages/Toggle/src/Toggle.tsx
DesignerDave: Possibly `StatefulStyleProps`? Not sure that's better?

packages/Toggle/src/Toggle.tsx
```diff
@@ -403,12 +444,13 @@ export default class Toggle extends PureComponent {
     }
   };
   render() {
-    const checkboxId = this.props.id || this.checkboxId;
+    const checkboxId = this.props.id
+      ? this.props.id.toString()
```
DesignerDave: Is the ID here not converted to a string when rendered in HTML? Could be wrong but I don't think this is necessary.

jetpacmonkey: I think you're right, but TS complained, so 🤷‍♂

DesignerDave: I'm not in love for altering the output JS for the sake of TS complaining. It would be good to know specifically why it thinks that it can't be a number.

PR #86, opened 2019-06-11
PD-318: Upgrade to @testing-library/react

brookescarlett:
## ✍️ Proposed changes

Updated testing library used, as react-testing-library is now deprecated. All existing tests still passing.

🎟 _Jira ticket:_ [PD-318](https://jira.mongodb.org/browse/PD-318)

## 🛠 Types of changes

## ✅ Checklist

Ran existing tests to make sure that they still pass

## Comments
@jetpacmonkey the Button test suite would not pass unless I explicitly added size and variant props. Know these tests weren't failing before/the ts:linter wasn't complaining so didn't want to change anything about the component without raising to you first.


packages/Button/src/Button.tsx
```diff
@@ -305,9 +305,7 @@ interface LinkButtonProps extends HTMLElementProps<'a'>, SharedButtonProps {
   href: string;
 }
 
-interface ButtonButtonProps
-  extends HTMLElementProps<'button'>,
-    SharedButtonProps {
+interface ButtonButtonProps extends SharedButtonProps, SharedButtonProps {
```
hswolff: Whoops.

packages/Button/src/Button.tsx
```diff
@@ -407,7 +405,7 @@ export default class Button extends Component<ButtonProps> {
     return (
       <button
         type="button"
-        {...rest as HTMLElementProps<'button'>}
+        {...(rest as HTMLElementProps<'button'>)} //eslint-disable-line
```
hswolff: Not seeing this disable being required locally. Hoping we can remove.

packages/Popover/package-lock.json
```diff
@@ -1,4 +1,5 @@
 {
+<<<<<<< HEAD
```
hswolff: Whoops.

packages/Button/src/Button.spec.tsx
```diff
@@ -116,7 +137,12 @@ describe('packages/Button', () => {
 
   test(`renders component inside of a React Element/HTML tag based on as prop, even when "href" is set`, () => {
     const { container } = render(
-      <Button as="div" href="http://mongodb.design">
+      <Button
+        as="div"
+        href="http://mongodb.design"
+        variant={Variant.Default}
+        size={Size.Normal}
```
jetpacmonkey: These props should definitely not be required, they're in `Button.defaultProps`.

packages/Button/src/Button.tsx
jetpacmonkey: Why is this extending `SharedButtonProps` twice now?

packages/Button/src/Button.tsx
```diff
@@ -397,7 +395,7 @@ export default class Button extends Component<ButtonProps> {
 
     if (usesLinkElement(this.props)) {
       return (
-        <a {...rest as HTMLElementProps<'a'>} {...commonProps}>
+        <a {...(rest as HTMLElementProps<'a'>)} {...commonProps}>
```
jetpacmonkey: Weird... Prettier _removed_ the parens for me when I had originally put them in.

packages/Button/src/Button.tsx
jetpacmonkey: What was eslint complaining about here?

packages/Popover/package-lock.json
brookescarlett: 👍

packages/Button/src/Button.tsx
brookescarlett: 👍

packages/Button/src/Button.spec.tsx
brookescarlett: That's the only way they passed ts:check

packages/Button/src/Button.tsx
brookescarlett: Prettier kept adding them for me -- couldn't take them out!

packages/Button/src/Button.tsx
brookescarlett: Fixed that, I was messing around with  the props trying to figure out what was making ts:check fail, but it turns out it was just adding size/variant to the test files.

packages/Button/src/Button.tsx
brookescarlett: Was adding the () and was just testing if that was what was breaking the test suite... tried everything before adding size/variant to the test files

packages/Button/src/Button.spec.tsx
brookescarlett: I know! But ts:check would not pass without it.

jetpacmonkey: That sounds like it'd be a problem for other people using TS, they shouldn't be forced to include these props. I think we should be able to just convert `<Button>` to a function component, using destructuring defaults instead of defaultProps, actually. There isn't even any state. That should simplify these types. I'll push a commit to your branch, if you don't mind 😁 .

brookescarlett: Sounds good to me :)

jetpacmonkey: Pushed! Not sure what exactly was causing `defaultProps` to stop working suddenly, but a function component makes more sense for `<Button>` regardless of that, so I'm not going to dive too deep down that rabbit hole.

packages/Button/src/Button.tsx
```diff
@@ -333,85 +320,81 @@ function usesLinkElement(
...
+      {...(rest as HTMLElementProps<'button'>)}
+      {...commonProps}
+    >
+      {children}
+    </button>
+  );
 }
+
+/*
+NOTE(JeT):
```
DesignerDave: Is this still the case with the functional component?

packages/Button/src/Button.tsx
```diff
@@ -333,85 +320,81 @@ function usesLinkElement(
...
-      </button>
+      </a>
     );
   }
+
+  // NOTE(JeT): The button's `type` will be overridden if it is in the passed-in props
+  return (
+    <button
+      type="button"
+      {...(rest as HTMLElementProps<'button'>)}
```
DesignerDave: Does `HTMLElementProps` include event handler props like `onClick`?

DesignerDave: @jetpacmonkey

packages/Button/src/Button.tsx
DesignerDave: @jetpacmonkey

jetpacmonkey: I don't know of any reason why it wouldn't be the case, although I didn't re-check

packages/Button/src/Button.tsx
jetpacmonkey: It should, yes. For this specific line it's just a type-cast that doesn't actually affect anything, but `HTMLElementProps<'button'>` is equivalent to `React.ButtonHTMLAttributes<HTMLButtonElement>` (it's pulling it out of this definition [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fa812c285d13280dc12c285d4538f209bbe1e12b/types/react/index.d.ts#L2514))

DesignerDave: Just double-checking 👍

packages/Button/src/Button.tsx
DesignerDave: So in my limited TS experience, I've noticed there's a difference with type inference between functional and class components. I just tested removing `as any` and it seems fine.

I ran `npm run ts:check`, `npm run test`, and successfully built the component. I can't find a reason to keep it around, as everything appears to work fine.

jetpacmonkey: Very weird... well in that case, I guess we can get rid of it!

PR #87, opened 2019-06-11
PD-317: Migrate from NPM to Yarn

brookescarlett:
## ✍️ Proposed changes

Migrating from using NPM to Yarn 1.16.0.
- Updated package.json
  - Created a yarn workspace
  - Changed internal commands to use yarn instead of NPM
- Updated README
- Updated link script

🎟 _Jira ticket:_ [PD-317](https://jira.mongodb.org/browse/PD-317)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

This branch is dependent upon PD-318, which is why those commits currently appear in the git diff.


README.md
```diff
@@ -12,17 +12,17 @@ A set of CSS styles and React components built with design in mind.
 
    via [nodejs installer](https://nodejs.org/en/)
 
-2. npm >= 6 installed.
+2. yarn >= 1.16.0 installed.
 
 3. Install dependencies and link packages.
 
-   `npm run init`
+   `yarn run init`
```
hswolff: I think `yarn init` will also just work.

README.md
```diff
@@ -33,7 +33,7 @@ This will allow you to make changes to your local repository of `leafygreen-ui`
 To do this, clone this repository and navigate to the root directory (where `package.json` is located), then run the following:
 
 ```
-npm run link -- ${PATH_TO_APPLICATION}
+yarn run link -- ${PATH_TO_APPLICATION}
```
hswolff: Can we test this please?

README.md
```diff
@@ -94,15 +94,15 @@ We follow conventional commits with our commit messages.
...
+yarn build
 ```
 
 2. Create a new branch, and make sure to push it. This is where package version updates, and git tags will be pushed from.
 
 3. Publish all packages to NPM using Lerna. In the future, this will also include automatic changelog creation.
 
 ```
-npm run release
+yarn run release
```
hswolff: `yarn release` ?

README.md
```diff
@@ -111,8 +111,8 @@ npm run release
 
 You can deploy a static build of our Storybook site to gh-pages.
 
-1. First be sure you've built a static version of Storybook: `npm run build:storybook`
-2. Then deploy to gh-pages: `npm run release:site`
+1. First be sure you've built a static version of Storybook: `yarn run build:storybook`
```
hswolff: Omit `run` here and next lien.

package.json
```diff
@@ -30,28 +30,28 @@
...
   "scripts": {
     "start": "start-storybook -p 9001 -c storybook -s ./storybook/static",
-    "build": "lerna run --parallel build -- -- --env production && npm run ts:emit",
+    "build": "lerna run --parallel build -- -- --env production && yarn ts:emit",
     "build:storybook": "build-storybook -c storybook -o ./storybook/public -s ./storybook/static",
-    "format": "npm run prettier && npm run eslint:format",
-    "init": "npm install && lerna bootstrap && npm run build",
-    "link-all-packages": "lerna exec --ignore @leafygreen-ui/lib --ignore @leafygreen-ui/theme -- npm link ",
+    "format": "yarn prettier && yarn eslint:format",
+    "init": "yarn && lerna bootstrap && yarn build",
```
hswolff: I think we can omit bootstrap now. If you test it, it should be fine.

package.json
```diff
@@ -74,6 +74,7 @@
     "@storybook/theming": "^5.0.11",
     "@svgr/babel-plugin-replace-jsx-attribute-value": "^4.2.0",
     "@svgr/webpack": "^4.2.0",
+    "@testing-library/react": "^8.0.1",
```
hswolff: This can be omitted.

packages/Badge/src/Badge.spec.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, cleanup } from 'react-testing-library';
+import { render, cleanup } from '@testing-library/react';
```
hswolff: Need to back these changes out when the other PR goes through.

README.md
brookescarlett: `yarn init` generates a package.json file: https://yarnpkg.com/lang/en/docs/cli/init/

README.md
brookescarlett: Tested this and it works! A `package-lock.json` is created in the process though, and trying to figure out where that's happening and how to prevent.

README.md
brookescarlett: 👍

packages/Badge/src/Badge.spec.tsx
brookescarlett: Will do!

package.json
brookescarlett: 👍 you're right

README.md
brookescarlett: Update: fixed this!

package.json
```diff
@@ -30,28 +30,28 @@
...
     "start": "start-storybook -p 9001 -c storybook -s ./storybook/static",
-    "build": "lerna run --parallel build -- -- --env production && npm run ts:emit",
+    "build": "lerna run --parallel build -- -- --env production && yarn ts:emit",
     "build:storybook": "build-storybook -c storybook -o ./storybook/public -s ./storybook/static",
-    "format": "npm run prettier && npm run eslint:format",
-    "init": "npm install && lerna bootstrap && npm run build",
-    "link-all-packages": "lerna exec --ignore @leafygreen-ui/lib --ignore @leafygreen-ui/theme -- npm link ",
+    "format": "yarn prettier && yarn eslint:format",
+    "init": "yarn && yarn build",
+    "link-all-packages": "lerna exec --ignore @leafygreen-ui/lib --ignore @leafygreen-ui/theme -- yarn link ",
```
DesignerDave: Nit: Remove space at the end.

brookescarlett: 👍

.circleci/config.yml
```diff
@@ -9,23 +9,22 @@ jobs:
...
+      - run: yarn install
+      - run: yarn build
       - run:
           name: prettier
-          command: npm run prettier:check
+          command: yarn prettier:check
       - run:
           name: lint
-          command: npm run eslint -- --format junit -o reports/junit/js-lint-results.xml
+          command: yarn eslint -- --format junit -o reports/junit/js-lint-results.xml
```
gssbzn: yarn doesn't need the extra `--` you can just do `yarn eslint --format junit...`

.circleci/config.yml
```diff
@@ -9,23 +9,22 @@ jobs:
           keys:
             - node-cache-{{ checksum "package-lock.json" }}
```
gssbzn: you need to update this to make the cache work again recommendation to also add a counter to evict the cache in case is needed,
`yarn-v1-{{ checksum "yarn.lock" }}`

.circleci/config.yml
```diff
@@ -9,23 +9,22 @@ jobs:
...
             - node-cache
```
gssbzn: I think it's safe to remove this key

.circleci/config.yml
```diff
@@ -9,23 +9,22 @@ jobs:
...
           name: tests
-          command: npm run test -- --no-cache --ci --runInBand --reporters=default --reporters=jest-junit
+          command: yarn test -- --no-cache --ci --runInBand --reporters=default --reporters=jest-junit
           environment:
             JEST_JUNIT_OUTPUT: 'reports/junit/js-test-results.xml'
       - run:
           name: types
-          command: npm run ts:check
+          command: yarn ts:check
       - save_cache:
```
gssbzn: you may need to update this with the same key as above, also I'm not sure this may work as expected, please check circleci is actually using the cache, before I have to manually set the yarn cache folder (`yarn install --cache-folder ~/.cache/yarn`) and save that folder here but I haven't use yarn in a while

.circleci/config.yml
hswolff: can just change to yarn.lock. main reason I had the prefix be `node-cache` to be package manager agnostic.

.circleci/config.yml
hswolff: It's a fallback key, recommended best practice from other circleci configs I've seen.

.circleci/config.yml
hswolff: Def need to update to use the same key.

.circleci/config.yml
gssbzn: yeah, name is just a suggestion, like I said I would still add the `v1`, It has happened before to me that we had to expire the cache because of reasons and increment that number is the easiest thing

.circleci/config.yml
gssbzn: there's no save step with this key if you want the fallback there should be something to fallback to, so we either remove it or add a save step with it

.circleci/config.yml
brookescarlett: 👍

.circleci/config.yml
brookescarlett: 👍

package.json
```diff
@@ -30,28 +30,28 @@
   },
   "scripts": {
```
hswolff: While we're here we should fix the warnings yarn throws up:

```
warning " > @emotion/styled@10.0.12" has unmet peer dependency "@emotion/core@^10.0.0".
warning "@emotion/styled > @emotion/styled-base@10.0.13" has unmet peer dependency "@emotion/core@^10.0.0".
warning " > react-emotion@9.0.2" has incorrect peer dependency "emotion@^9.0.2".
```

We can remove react-emotion as we don't use that version anymore, and we can add @emotion/core to the root package.json as it's a peerDep.

package.json
```diff
@@ -30,28 +30,28 @@
   },
```
hswolff: I think we should remove the `contributors` array as (hopefully) it'll grow too large that it'd be better to just track the GitHub page for that. cc @DesignerDave if you're ok with that.

package.json
```diff
@@ -30,28 +30,28 @@
...
   "scripts": {
     "start": "start-storybook -p 9001 -c storybook -s ./storybook/static",
-    "build": "lerna run --parallel build -- -- --env production && npm run ts:emit",
+    "build": "lerna run --parallel build -- -- --env production && yarn ts:emit",
```
hswolff: Let's update `lerna.json` so that it uses `yarn` when issuing `run` commands. Docs: https://github.com/lerna/lerna/tree/master/commands/run#--npm-client-client

Then we can update this to: 
```suggestion
    "build": "lerna run --parallel build -- --env production && yarn ts:emit",
```

Still need that other `--` to pass it to webpack tho.

.circleci/config.yml
```diff
@@ -7,27 +7,25 @@ jobs:
...
-          command: npm run prettier:check
+          command: yarn prettier:check
       - run:
           name: lint
-          command: npm run eslint -- --format junit -o reports/junit/js-lint-results.xml
+          command: yarn eslint --format junit -o reports/junit/js-lint-results.xml
       - run:
           name: tests
-          command: npm run test -- --no-cache --ci --runInBand --reporters=default --reporters=jest-junit
+          command: yarn test -- --no-cache --ci --runInBand --reporters=default --reporters=jest-junit
```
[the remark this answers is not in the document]
brookescarlett: 👍

package.json
brookescarlett: 👍 , if you're okay with that @DesignerDave

package.json
brookescarlett: 👍

package.json
brookescarlett: 👍

package.json
```diff
@@ -30,30 +20,33 @@
...
   },
   "husky": {
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-push": "npm run lint"
+      "pre-push": "yarn lint"
     }
   },
+  "peerDependencies": {
+    "@emotion/core": "^10.0.10"
```
hswolff: The warning was saying that this IS a peerDependency. i.e. we need to install it ourselves in devDependencies.

brookescarlett: 👍

package.json
DesignerDave: :+1:

PR #89, opened 2019-06-11
PD-304: Port <RadioGroup> to TypeScript

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-304: Port <RadioGroup> to TypeScript](https://jira.mongodb.org/browse/PD-304)

## 🛠 Types of changes

## ✅ Checklist


packages/RadioGroup/src/Variant.ts
```diff
@@ -0,0 +1,6 @@
+enum Variant {
```
DesignerDave: Few thoughts here – I ran into a case where I needed to do something similar for Syntax.

1. Some packages also have `Size` which is treated similarly. Should we name this file to something more generic for other enums (or whatever we decide on), and make `Variant` a named export to extend this pattern across packages?
2. Haven't we been naming these enums `Variants`?
3. We should probably figure out the right pattern for these, since it sounds like we probably aren't going forward with enums for this case.

packages/RadioGroup/src/RadioGroup.tsx
```diff
@@ -20,7 +22,34 @@ const baseStyle = css`
   padding: 5px;
 `;
 
-export default class RadioGroup extends PureComponent {
+function isRadioElement(
+  element: React.ReactNode,
+): element is React.ReactElement<RadioProps, typeof Radio> {
+  return (
```
DesignerDave: Boy I want optional chaining in JS... Is there any reason we can't do:

```js
return element && element.type && element.type.displayName === 'Radio'
```

Should be easier to read and accomplish the same thing.

One day we can make it something like...
```js
return element?.type?.displayName === 'Radio'
```

I can dream on that last part 😁

jetpacmonkey: TS wouldn't like that. If `element` is `null` you'd be returning `null` when it wants a boolean, and if it's a string or some other non-object thing that's a valid React node, then `.type` doesn't exist on it.

DesignerDave: Right, so it would have to be something like:

```js
return !!(element && element.type && element.type.displayName === 'Radio')
```

Good point, I defer to you then 🤷‍♂

jetpacmonkey: And that would only solve the "null" issue, you'd still have to convince it that there really is a `type` key on there :)

DesignerDave: I tested it out, it also solves for returning undefined (when it's a string, or Array for instance).

packages/RadioGroup/src/Variant.ts
jetpacmonkey: I don't _think_ I've been using the plural. In general I try not to, the type is defining what a valid single value is, so the singular feels more accurate.

packages/RadioGroup/src/RadioGroup.tsx
jetpacmonkey: Yeah, TS seems to be generally of the opinion that having all unknown properties on an object come back as `undefined` was a mistake.

packages/RadioGroup/src/Variant.ts
DesignerDave: I think you're right – my bad.

PR #91, opened 2019-06-11
PD-321: Convert enums to const objects

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-321: Convert enums to const objects](https://jira.mongodb.org/browse/PD-321)

## 🛠 Types of changes

## ✅ Checklist


hswolff:
I'm fine keeping it inside a file. If it's shared amidst a complex package then having it extracted to its own file makes sense to avoid circular dependencies.

jetpacmonkey:
There aren't any in this PR that are shared between files in a single component, are there? In general, I'm not opposed to extracting, though.

DesignerDave:
Aside from Popover, I don't think so.

jetpacmonkey:
Ah, right, I keep on forgetting about Popover

jetpacmonkey:
Alright this should be ready to go

packages/Popover/src/Popover.story.tsx
```diff
@@ -55,8 +55,16 @@ function DefaultExample() {
     >
       {text('Button Content', 'Popover')}
       <Popover
-        align={select('Align', ['top', 'right', 'bottom', 'left'], 'top')}
-        justify={select('justify', ['start', 'middle', 'end'], 'start')}
+        align={select(
+          'Align',
+          ['top', 'right', 'bottom', 'left'] as const,
```
hswolff: Can we import the object and do Object.values? Ditto everywhere else?

DesignerDave: ⬆️

packages/Popover/src/Popover.spec.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import { render, cleanup } from '@testing-library/react';
-import Popover, { Align, Justify } from './Popover';
+import Popover from './Popover';
+import Align from './Align';
+import Justify from './Justify';
```
DesignerDave: Why not one unified `types` file? Feels like a lot to have a separate file for every "enum-like" type we have.

jetpacmonkey: @hswolff do you have a preference here?

hswolff: Not really. They're not exactly types tho, cuz they're also objects. But no preference.

DesignerDave: Another idea would be `props.ts` – export the props and the specific prop types like these and such.

I know we extend Popover's props elsewhere too, so there's that.

DesignerDave: I suppose my point is that it would be nice to have a uniform file across packages that exports these, rather than `Align.ts`, `Justify.ts`, `Variant.ts`, etc.

hswolff: Yeah, I agree with that sentiment. Not sure of what that is yet. We can think on it tho, but not block this PR with that answer.

DesignerDave: Fair enough – onward march!

PR #93, opened 2019-06-13
PD-324: Enable noImplicitAny flag

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-324: Enable noImplicitAny flag](https://jira.mongodb.org/browse/PD-324)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/hooks/src/useEventListener.ts
+++ b/packages/hooks/src/useEventListener.ts
@@ -1,7 +1,7 @@
 import { useEffect, useRef } from 'react';
 
 interface UseEventOptions {
-  options?: object;
+  options?: EventListenerOptions;
   dependencies?: Array<any>;
   enabled?: boolean;
   element?: Document | HTMLElement;
@@ -17,17 +17,21 @@ interface UseEventOptions {
  * @param optional.dependencies Array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
  * @param optional.element Value to be passed as target of event handler, will default to document.
  */
-export default function useEventListener(
-  type: keyof GlobalEventHandlersEventMap,
-  eventCallback: (e) => void,
+export default function useEventListener<
+  Type extends keyof GlobalEventHandlersEventMap
+>(
+  type: Type,
+  eventCallback: (e: GlobalEventHandlersEventMap[Type]) => void,
   {
-    options = undefined,
+    options,
     enabled = true,
     dependencies = [enabled, type],
     element = document,
   }: UseEventOptions = {},
 ) {
-  const memoizedEventCallback = useRef(e => {}); //eslint-disable-line @typescript-eslint/no-unused-vars
+  const memoizedEventCallback: React.MutableRefObject<
+    (e: GlobalEventHandlersEventMap[Type]) => void
+  > = useRef(() => {});
 
   useEffect(() => {
     memoizedEventCallback.current = eventCallback;
@@ -38,12 +42,18 @@ export default function useEventListener(
       return;
     }
 
-    const callback = e => memoizedEventCallback.current(e);
+    const callback = (e: GlobalEventHandlersEventMap[Type]) =>
+      memoizedEventCallback.current(e);
 
-    element.addEventListener(type, callback, options);
+    // NOTE(JeT): I'm pretty sure there should be a way to avoid this type assertion, but I couldn't figure it out.
+    element.addEventListener(type, callback as EventListener, options);
 
     return () => {
-      document.removeEventListener(type, eventCallback, options);
+      element.removeEventListener(
+        type,
+        eventCallback as EventListener,
+        options,
+      );
     };
   }, dependencies);
 }
--- a/packages/hooks/src/useEventListener.ts
+++ b/packages/hooks/src/useEventListener.ts
@@ -1,7 +1,7 @@
 import { useEffect, useRef } from 'react';
 
 interface UseEventOptions {
-  options?: EventListenerOptions;
+  options?: AddEventListenerOptions;
   dependencies?: Array<any>;
   enabled?: boolean;
   element?: Document | HTMLElement;
--- a/packages/palette/src/palette.spec.ts
+++ b/packages/palette/src/palette.spec.ts
@@ -10,11 +10,32 @@ interface HueRange {
   light3: string;
 }
 
+const requiredKeys = new Set(['dark3', 'base', 'light3']);
+const optionalKeys = new Set(['dark2', 'dark1', 'light1', 'light2']);
+
 const allColors: Array<string | HueRange> = Object.values(uiColors);
 describe('packages/palette', function() {
-  // eslint-disable-next-line no-console
-  console.log(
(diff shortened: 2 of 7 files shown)


packages/hooks/src/useEventListener.ts
```diff
@@ -17,17 +17,21 @@ interface UseEventOptions {
...
  * @param optional.element Value to be passed as target of event handler, will default to document.
  */
-export default function useEventListener(
-  type: keyof GlobalEventHandlersEventMap,
-  eventCallback: (e) => void,
+export default function useEventListener<
+  Type extends keyof GlobalEventHandlersEventMap
+>(
+  type: Type,
+  eventCallback: (e: GlobalEventHandlersEventMap[Type]) => void,
```
DesignerDave: If we type this as an event listener, rather than typing the event parameter, can we omit typing it's usage on line 54?

packages/hooks/src/useEventListener.ts
```diff
@@ -38,12 +42,18 @@ export default function useEventListener(
       return;
     }
 
-    const callback = e => memoizedEventCallback.current(e);
+    const callback = (e: GlobalEventHandlersEventMap[Type]) =>
```
DesignerDave: Same comment as above – if we type this differently, can we omit typing where it's used on line 49?

packages/palette/src/palette.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  dark3: string;
+  dark2?: string;
+  dark1?: string;
+  base: string;
+  light1?: string;
+  light2?: string;
+  light3: string;
+}
+
+const allColors: Array<string | HueRange> = Object.values(uiColors);
```
DesignerDave: If we export one named object with `as const` rather than have a bunch of named exports from `uiColors`, can the type of `allColors` be inferred here?

packages/palette/src/palette.story.tsx
```diff
@@ -26,13 +32,13 @@ const ColorBlock = styled('div')`
...
-    color: ${props => readableColor(lighten(0.2, props['data-color']))};
-    background-color: ${props => lighten(0.2, props['data-color'])};
+    color: ${props => readableColor(lighten(0.2, props.color))};
+    background-color: ${props => lighten(0.2, props.color)};
     border-radius: 4px;
   }
 
   &:after {
-    content: attr(data-name);
+    content: "${props => props.name}";
```
DesignerDave: Why is this JS solution preferable to the baked-in CSS function?

packages/palette/src/uiColors.ts
```diff
@@ -1,17 +1,7 @@
 export const white = '#FFFFFF';
 export const black = '#061621';
 
-interface HueRange {
```
DesignerDave: Why are we removing this?

packages/hooks/src/useEventListener.ts
```diff
@@ -1,7 +1,7 @@
 import { useEffect, useRef } from 'react';
 
 interface UseEventOptions {
-  options?: object;
+  options?: EventListenerOptions;
```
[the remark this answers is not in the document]
jetpacmonkey: No, but I wish it would. This was just a case where I found the existing definition for this type, so I figured I'd narrow it to what can actually be passed to the event listener.

packages/hooks/src/useEventListener.ts
jetpacmonkey: We could, but then we would lose the information that, say, a `click` handler is going to get a `MouseEvent`, with `clientX` and `clientY` parameters. It's weird, because `addEventListener` is supposed to do this same mapping, but it seems like it's not following along quite right, something to do with `element` being `Document | Element` rather than just one or the other I think.

packages/palette/src/palette.story.tsx
jetpacmonkey: 🤷‍♂ this was we can give it a type

packages/palette/src/uiColors.ts
jetpacmonkey: Having a declared type on the variables takes precedence over `as const`. This way, in the storybook (for instance) we don't need to include a fallback to `transparent` because it knows what keys are defined. I added the `palette.spec.ts` file with the thinking that if we go outside of the defined `HueRange` type it would throw a type error there, so we still get that consistency.

jetpacmonkey: That test doesn't protect us from adding a key outside of `HueRange`, btw, but it does at least ensure that we didn't remove any of the required hues, or set it to a number or something.

packages/palette/src/palette.spec.ts
jetpacmonkey: As mentioned in the other comment, this line is explicitly typed on purpose, to try and make sure that we're exporting a consistent shape from `uiColors`, while still allowing `as const` to do its thing.

packages/hooks/src/useEventListener.ts
```diff
@@ -17,17 +17,21 @@ interface UseEventOptions {
  * @param optional.dependencies Array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
  * @param optional.element Value to be passed as target of event handler, will default to document.
  */
-export default function useEventListener(
-  type: keyof GlobalEventHandlersEventMap,
-  eventCallback: (e) => void,
+export default function useEventListener<
+  Type extends keyof GlobalEventHandlersEventMap
+>(
+  type: Type,
```
hswolff: What's the benefit of making this generic?

packages/hooks/src/useEventListener.ts
```diff
@@ -38,12 +42,18 @@ export default function useEventListener(
...
+    const callback = (e: GlobalEventHandlersEventMap[Type]) =>
+      memoizedEventCallback.current(e);
 
-    element.addEventListener(type, callback, options);
+    // NOTE(JeT): I'm pretty sure there should be a way to avoid this type assertion, but I couldn't figure it out.
+    element.addEventListener(type, callback as EventListener, options);
 
     return () => {
-      document.removeEventListener(type, eventCallback, options);
+      element.removeEventListener(
```
hswolff: Whoooops. Nice catch.

packages/hooks/src/useEventListener.ts
jetpacmonkey: It allows TS to infer that if I pass in, say, `'click'`, then the event handler is going to be passed a `MouseEvent` rather than just a generic `Event`. So it'll have `clientX` and `clientY` and `button` and all that stuff on it. The built-in types for `addEventListener` are already doing the same thing, so I'm just kinda stealing from that.

packages/palette/src/uiColors.ts
jetpacmonkey: Actually... maybe it'd be better to just do a regular test instead of trying to get TS to do it for us

jetpacmonkey: Alright I did that

PR #97, opened 2019-06-18
PD-313: Extract common type utils and util types to lib

jetpacmonkey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-313](https://jira.mongodb.org/browse/PD-313)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/badge/src/Badge.spec.tsx
+++ b/packages/badge/src/Badge.spec.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { render, cleanup } from '@testing-library/react';
+import { typeIs } from '@leafygreen-ui/lib';
 import Badge from './Badge';
 
 afterAll(cleanup);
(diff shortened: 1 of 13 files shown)


packages/lib/src/index.ts
```diff
@@ -11,7 +13,7 @@
  */
 export function createDataProp(name: string) {
   // ts:emit throws an error if this is not typed. Usually this can be inferred.
-  const prefix: string = 'data-leafygreen-ui';
+  const prefix = 'data-leafygreen-ui' as string;
```
jetpacmonkey: This change was just to make my editor stop auto-"fixing" it. There's probably a better way...

DesignerDave: If you're setting a `const` to a literal, I don't think you need to type it because it's then inferred.

packages/badge/src/Badge.spec.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { render, cleanup } from '@testing-library/react';
+import { typeIs } from '@leafygreen-ui/lib';
```
DesignerDave: This approach turned out clean!

packages/lib/src/index.ts
jetpacmonkey: Btw, the inferred type of this `const` is not actually `string`. It's the string literal type `'data-leafygreen-ui'`. If it was just `string`, TS wouldn't have any unexported type to worry about.

PR #99, opened 2019-06-20
PD-310: Implement Modal Component

brookescarlett:
## ✍️ Proposed changes

Created modal component. The tests aren't perfect yet, so still working on those (not sure why they only pass asynchronously), but would love to get feedback on the initial implementation in the meantime. 

To note: all styles were ported from MMS.

🎟 _Jira ticket:_ [PD-310](https://jira.mongodb.org/browse/PD-310)

## 🛠 Types of changes

## ✅ Checklist


packages/modal/src/Modal.story.tsx
```diff
@@ -0,0 +1,25 @@
...
+function Example() {
+  const { active, setActive } = useModalState();
+
+  return (
+    <>
+      <button onClick={() => setActive(active => !active)}>Open Modal</button>
+      <Modal
+        active={active}
+        setActive={setActive}
+        usePortal={boolean('usePortal', true)}
```
DesignerDave: Hmm... I suppose that we can expose this but I feel a lot of people may shoot themselves in the foot by setting false here.

@hswolff Is there any reason we'd really want to expose this for modal?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
+import React, { Fragment, useCallback, useRef, SetStateAction } from 'react';
+import PropTypes from 'prop-types';
+import Portal from '@leafygreen-ui/portal';
+import Icon, { Size } from '@leafygreen-ui/icon';
+import { useEventListener } from '@leafygreen-ui/hooks';
+import { css, cx } from '@leafygreen-ui/emotion';
+
+export const ModalSize = {
+  XXSmall: 'xxsmall',
+  XSmaill: 'xsmall',
```
DesignerDave: Typo

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
+import React, { Fragment, useCallback, useRef, SetStateAction } from 'react';
+import PropTypes from 'prop-types';
+import Portal from '@leafygreen-ui/portal';
+import Icon, { Size } from '@leafygreen-ui/icon';
+import { useEventListener } from '@leafygreen-ui/hooks';
+import { css, cx } from '@leafygreen-ui/emotion';
+
+export const ModalSize = {
+  XXSmall: 'xxsmall',
```
DesignerDave: We need to resolve this before merging, but I think Fred and I need to talk about having this many sizes included. Much harder to take something away than to add it, and I'm not sure we need more than, say, 3 sizes.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  Small: 'small',
+  Normal: 'normal',
+  Large: 'large',
+  XLarge: 'xlarge',
+} as const;
+
+export type ModalSize = typeof ModalSize[keyof typeof ModalSize];
+
+const overlayStyle = css`
+  animation: fade-in 500ms cubic-bezier(0.165, 0.84, 0.44, 1);
```
DesignerDave: Heh getting fancy with the timing functions! I'll get back to ya on this, but usually I don't like using the CSS `cubic-bezier` function because it only supports two points on the curve, where you need 3+ to make a bounce feel natural. May want to consider using a JS animation for this, but I'll get back to ya!

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  -webkit-animation: fade-in 500ms cubic-bezier(0.165, 0.84, 0.44, 1);
+  background-color: #000;
```
DesignerDave: Let's start pulling from the palette library please 😁

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  useEventListener('keydown', handleEscape, {
+    options: { once: true },
+  });
+
+  const Root = usePortal ? Portal : Fragment;
+
+  return active ? (
+    <Root>
+      <div {...rest} className={overlayStyle} onClick={handleDocumentClick}>
+        <div
```
DesignerDave: If we make the modal and the overlay siblings instead of nesting the modal content, we can eliminate the `contains` check in `handleDocumentClick`, and probably remove the event handler from this element.

DesignerDave: Also we could remove all of the flex rules for the overlay!

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  active = false,
+  usePortal = true,
+  size = ModalSize.Normal,
+  setActive,
+  children,
+  title,
+  onRequestClose,
+  className,
+  ...rest
+}: ModalProps) {
```
DesignerDave: We should handle disabling scrolling on the body, and allow scrolling of modal content. That's typical behavior for modals that I don't think we're handling here.

The first portion might be best extracted to a hook.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+`;
+
+const modalContentStyle = css`
+  margin: 18px 40px;
+  max-width: 1270px;
+  min-width: 944px;
+  padding: 30px;
+  align-self: flex-start;
+  text-align: left;
+  color: black;
```
DesignerDave: Pull from palette please

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  justify-content: center;
+`;
+
+const modalContentStyle = css`
+  margin: 18px 40px;
+  max-width: 1270px;
+  min-width: 944px;
+  padding: 30px;
+  align-self: flex-start;
+  text-align: left;
```
DesignerDave: This should be the default for the property.

packages/modal/src/Modal.tsx
DesignerDave: This needs to be done for all the colors here I believe.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  color: black;
+  position: relative;
+  background-color: #fff;
+  background-clip: padding-box;
+  border: 1px solid rgba(0, 0, 0, 0.1);
+  border-radius: 3px;
+  outline: 0;
```
DesignerDave: Why?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  margin: 18px 40px;
+  max-width: 1270px;
+  min-width: 944px;
+  padding: 30px;
+  align-self: flex-start;
+  text-align: left;
+  color: black;
+  position: relative;
+  background-color: #fff;
+  background-clip: padding-box;
```
DesignerDave: 🤔 I doubt we need to use this if we structure the modal right.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  border: 1px solid rgba(0, 0, 0, 0.1);
+  border-radius: 3px;
+  outline: 0;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+`;
+
+const modalBodyStyle = css`
+  padding: 18px 0;
```
DesignerDave: We use 18px in several places, presumably intentionally. If it's intentional that they're the same, we should extract this to a variable.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+    max-width: inherit;
+    width: 1270px;
+  `,
+};
+
+const closeButton = css`
+  font-weight: 200;
+  color: #464c4f;
+  position: relative;
+  float: right;
```
DesignerDave: Yeah, I hate using floats pretty much anywhere except to make text wrap around stuff.

We should use absolute positioning for this button. I'd also remove anything that has to do with this being text since we're using the icon component.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+  top: -18px;
+  right: -10px;
+  margin-top: -2px;
+  cursor: pointer;
+`;
```
DesignerDave: No hover state? I can help with that.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+const titleStyle = css`
+  margin: 0;
+`;
+
+const EscapeKey = 27;
```
DesignerDave: Let's camelCase this since it's not a class or type. Second, let's move this to where it's used. Lastly, I'd rename to `escapeKeyCode` since that's more accurate.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+   */
+  title?: string;
+
+  /**
+   * Callback to change the active state of the Modal.
+   *
+   */
+  setActive?: (
+    active: boolean,
+  ) => void | React.Dispatch<SetStateAction<boolean>>;
```
DesignerDave: Sweet.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+ * @param props.children Content to appear inside of Modal container.
+ * @param props.title Title for the Modal, will appear inside header tags.
+ * @param props.onRequestClose Callback invoked when Modal is closed.
+ * @param props.className className applied to overlay div.
+ *
+ */
+function Modal({
+  active = false,
+  usePortal = true,
+  size = ModalSize.Normal,
```
DesignerDave: Maybe `Default` rather than `Normal`

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    if (!contentRef.current.contains(e.target as Node)) {
+      handleClose();
+    }
+  };
+
+  useEventListener('keydown', handleEscape, {
+    options: { once: true },
+  });
```
DesignerDave: `useModalState` should handle all of these event listeners, since that's the point of having it.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  const Root = usePortal ? Portal : Fragment;
+
+  return active ? (
```
DesignerDave: Happy path. Instead of a ternary within the return statement –

```js
if (!active) {
  return null;
}

return (...jsx);
```

packages/modal/src/Modal.tsx
DesignerDave: Alternatively, maybe we don't need it?

```js
const { active, setActive } = useState(false)

return <Modal active={active} setActive={setActive}>{ children }</Modal>
```

In that case, `setActive` is optional, and handling the ref is much easier. Thinking about all this more, the `useModalState` hook use would have to look like:

```js
const { active, setActive, setRef } = useModalState(false)

return <Modal active={active} setActive={setActive} setRef={setRef}>{ children }</Modal>
```

Thoughts @hswolff? I'm starting to lean towards the former option here, now remembering that we would have to have a ref internally in the hook to prevent click events within the modal from closing it.

DesignerDave: I think if we were considering the `useModal` hook I proposed before, it would be a different story because the hook could set that ref itself without developer intervention:

```js
const { active, setActive, Modal } = useModalHook(false);

return <Modal>{ children }</Modal>;
```

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,281 @@
...
+}
+
+Modal.displayName = 'Modal';
+
+Modal.propTypes = {
+  active: PropTypes.bool,
+  usePortal: PropTypes.bool,
+  size: PropTypes.string,
+  children: PropTypes.node,
+  onRequestClose: PropTypes.func,
```
DesignerDave: What if this was `onWillClose`, and if the consumer returns false, it prevents the modal from closing?

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: Makes sense there's also been a lot of talk about this with the designers lately

packages/modal/src/Modal.tsx
brookescarlett: Sounds good LMK

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: I'm learning 🤓

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: ported from MMS, removed.

packages/modal/src/Modal.tsx
brookescarlett: Again, ported from MMS but happy to abstract into a variable if you think I should.

packages/modal/src/Modal.tsx
brookescarlett: Help with that!

packages/modal/src/Modal.tsx
brookescarlett: Had it cased like this and in this position per @hswolff's feedback on the Menu Component. Happy to go with whatever you both prefer

packages/modal/src/Modal.tsx
brookescarlett: Thought this was our convention from before. Will update now!

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: That can't be done with vanilla CSS?

packages/modal/src/Modal.tsx
brookescarlett: Removed!

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+import 'jest-dom/extend-expect';
+import Modal from './Modal';
+
+afterAll(cleanup);
+
+describe('packages/Modal', () => {
+  const modalId = 'modal';
+  const modalTitle = 'Modal Title';
+  const modalContent = 'Modal Content';
+  // const onRequestClose = jest.fn();
```
DesignerDave: Remove commented code please

packages/modal/README.md
```diff
@@ -0,0 +1,129 @@
...
+  );
+}
+```
+
+**Output HTML**
+
+```html
+<button>Open Modal</button>
+<div class="leafygreen-ui-np1ghg">
+  <div class="leafygreen-ui-zbhk5h" tabindex="-1">
```
DesignerDave: This DOM doesn't look up-to-date with your changes.

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+    );
+
+    test('it does not display the modal', () => {
+      const modal = getByTestId(modalId);
+      const title = getByText(modalTitle);
+      const content = getByText(modalContent);
+
+      !expect(modal).toBeVisible();
+      !expect(title).toBeInTheDocument();
+      !expect(content).toBeInTheDocument();
```
DesignerDave: I think you can do something along the lines of: `expect(content).toBeNull()`. Not in love with testing the negative like this.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,268 @@
...
+  animation: fade-in 500ms cubic-bezier(0.165, 0.84, 0.44, 1);
+  -webkit-animation: fade-in 500ms cubic-bezier(0.165, 0.84, 0.44, 1);
+  background-color: ${uiColors.black};
+  opacity: 0.6;
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  overflow: hidden;
```
DesignerDave: 🤔 I don't think we should need this.

packages/modal/src/Modal.tsx
DesignerDave: It's... complicated. For one, we will need to set a CSS property on the document to disable scrolling, but I believe in some browsers you need to additionally set and unset a property on an actual DOM node to disable scrolling.

We do need some way to set the CSS.

packages/modal/src/Modal.tsx
DesignerDave: @hswolff Could you elaborate on why PascalCase is better here?

packages/modal/src/Modal.tsx
DesignerDave: Do the values seem linked at all? If so, let's do that.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,268 @@
...
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  background-color: ${uiColors.white};
+  border: 1px solid ${uiColors.gray.light3};
+  border-radius: 3px;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+  margin-right: auto;
+  margin-left: auto;
```
DesignerDave: We declare a different margin on the first line of this style declaration. We should remove these two rules, and update the first margin declaration to: `margin: 18px auto;`

packages/modal/src/Modal.story.tsx
DesignerDave: I think we should just remove the implementation around `usePortal`.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,268 @@
...
+          size={Size.Large}
+          onClick={handleClose}
+          className={closeButton}
+          data-dismiss="modal"
+          aria-hidden="true"
+        />
+
+        {title && (
+          <header>
+            <h3 className={titleStyle}>{title}</h3>
```
DesignerDave: I think we should remove these for a couple reasons:

1. We haven't gotten to defining this pattern well yet. I think that it's fine to remove this until we have a well established pattern.
2. Philosophically, these components should be individually minimal, and this sort of thing should be composed of multiple components rather than being built in.

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: This is coming from MMS, thoughts on this @hswolff?

packages/modal/src/Modal.tsx
hswolff: I don't remember making a case for either one, and if I did I disagree with my former self. My main desire was to have it be a named variable for clear referencing.

There is a future I see where we move this to `lib` and have a `const KeyboardKeys = { Escape: 27 }` object for uniform referencing.

packages/modal/src/Modal.tsx
hswolff: Agreed with @DesignerDave, let's omit from this primitive. It's too prescriptive to belong in the base Modal component.

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/lib": "^3.0.0",
+    "@leafygreen-ui/portal": "^1.1.0",
+    "@leafygreen-ui/palette": "^1.0.0",
+    "@leafygreen-ui/hooks": "^1.0.0",
+    "@leafygreen-ui/icon": "^1.0.4",
+    "lodash": "^4.17.11"
```
DesignerDave: I don't think we're using this.

packages/modal/package.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "gitHead": "53385cfed5ecf3101d098d26dedc6a40b27c402a"
```
DesignerDave: This can be removed until lerna adds it.

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const modalContent = 'Modal Content';
+
+  describe('when rendered with content', () => {
+    const { getByText, getByTestId } = render(
+      <Modal active data-testid={modalId}>
+        <h4>{modalContent}</h4>
+      </Modal>,
+    );
+
+    test('it appears on DOM when active prop is set', () => {
```
DesignerDave: Nit: `it's rendered in the DOM when...`

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      const modal = getByTestId(modalId);
+      expect(modal).toBeVisible();
```
DesignerDave: I don't think this is accurate to the test name – if this tests visibility, that's different than testing if it's mounted.

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  describe('when rendered with active equal to false', () => {
+    const { container } = render(
+      <Modal active={false} data-testid={modalId}>
+        <h4>{modalContent}</h4>
+      </Modal>,
+    );
+
+    test('it does not display the modal', () => {
```
DesignerDave: Also not a great test title – consider `it is not rendered within the DOM`

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+          modalShouldClose={() => true}
+        >
+          <h4>test content</h4>
+        </Modal>,
+      );
+
+      const modal = getByTestId('trueModalShouldClose');
+      window.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+
+      !expect(modal).toBeVisible();
```
DesignerDave: Is there something like `expect(modal).toNotBeVisible();`? I don't like the pattern of `!expect()`.

packages/modal/package.json
brookescarlett: 👍

packages/modal/package.json
brookescarlett: 👍

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.spec.tsx
brookescarlett: that's why i used the word 'appears' i'm testing that it's visible in the DOM

packages/modal/src/Modal.spec.tsx
brookescarlett: there is not unfortunately

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,259 @@
+import React, { useCallback, SetStateAction, useRef } from 'react';
+import PropTypes from 'prop-types';
+import Portal from '@leafygreen-ui/portal';
+import { Global } from '@emotion/core';
```
hswolff: Can't use this, must use injectGlobal like @DesignerDave's using in his Syntax component.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,259 @@
...
+  XLarge: 'xlarge',
+} as const;
+
+export type ModalSize = typeof ModalSize[keyof typeof ModalSize];
+
+const defaultSpacing = 18;
+
+const scrollContainer = css`
+  animation: fade-in 500ms cubic-bezier(0.165, 0.84, 0.44, 1);
+  -webkit-animation: fade-in 500ms cubic-bezier(0.165, 0.84, 0.44, 1);
```
hswolff: This should be handled by stylis via emotion I believe?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,259 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: flex-start;
+  pointer-events: initial;
+`;
+
+const modalContentStyle = css`
+  margin: ${defaultSpacing}px auto;
+  max-width: 1270px;
+  min-width: 944px;
```
hswolff: This component....we may want to be responsive. So it can work on mobile devices. On mms we're using [facepaint](https://github.com/emotion-js/facepaint) to success.

packages/modal/src/Modal.tsx
hswolff: Should that be handled by the Icon component by default?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,259 @@
...
+            glyph="X"
+            fill={'#5D6C74'}
+            size={Size.Large}
+            onClick={handleClose}
+            className={closeButton}
+            data-dismiss="modal"
+            aria-hidden="true"
+          />
+
+          <div className={modalBodyStyle}>{children}</div>
```
hswolff: I feel there will be a day when someone wants to set the className on this div.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,259 @@
...
+   */
+  setActive?: (
+    active: boolean,
+  ) => void | React.Dispatch<SetStateAction<boolean>>;
+
+  /**
+   * Callback to determine whether or not Modal should close when user tries to close it.
+   *
+   */
+  modalShouldClose?: () => boolean;
```
hswolff: Maybe in a future release if people desire, we should include an object to the function as a reason for why we're asking if the Modal should close. i.e. `{reason: 'escapeKey Pressed' }`, such that a user can make smarter decisions in this function.

hswolff: Also a rename to `shouldClose`? or `canClose`?

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: should we expose a modalBodyClassName prop? @DesignerDave

packages/modal/src/Modal.tsx
brookescarlett: I can rename to shouldClose, I know we were attempting to imitate the componentShouldUpdate name, but shouldClose works for me as well if you're good with it @DesignerDave

packages/modal/src/Modal.tsx
brookescarlett: Is there a standard for the number of/places where we use breakpoints?

brookescarlett: Happy to come up with some for this use case, but was wondering if there were certain breakpoints that we use as benchmarks and then back into styles from there?

packages/modal/src/Modal.tsx
brookescarlett: updated the animation to normal `ease-in-out` because cubic-bezier has a limited number of points cc: @DesignerDave

packages/modal/src/Modal.tsx
brookescarlett: updated such that we no longer need this prop. className is being spread properly on the container.

packages/modal/src/Modal.spec.tsx
jetpacmonkey: Can't you do `expect(modal).not.toBeVisible()`? It looks like the [examples for `jest-dom`](https://github.com/testing-library/jest-dom) follow that pattern

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,253 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: flex-start;
+  pointer-events: initial;
+`;
+
+const modalContentStyle = css`
+  margin: ${defaultSpacing}px auto;
+  max-width: 1270px;
+  min-width: 944px;
```
jetpacmonkey: Do we actually need `min-width` and `max-width` defined here? It looks like they're being overridden by each of the size-specific styles below.

packages/modal/src/Modal.tsx
jetpacmonkey: It gets even messier with mobile :(

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,253 @@
...
+    if (!contentRef.current.contains(e.target as Node)) {
+      handleClose();
+    }
+  };
+
+  const handleEscape = useCallback((e: KeyboardEvent) => {
+    if (e.keyCode === escapeKey) {
+      handleClose();
+    }
+  }, []);
```
jetpacmonkey: If the consuming component passes in a new `setActive` function for some reason, this callback is going to be stale.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,253 @@
...
+  );
+}
+
+Modal.displayName = 'Modal';
+
+Modal.propTypes = {
+  active: PropTypes.bool,
+  size: PropTypes.string,
+  children: PropTypes.node,
+  shouldClose: PropTypes.func,
```
jetpacmonkey: Should this be `setActive`?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,253 @@
...
+  -webkit-animation: fade-in 250ms ease-in-out;
+  background-color: ${uiColors.black};
+  opacity: 0.6;
+  overflow-y: auto;
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  display: flex;
```
jetpacmonkey: This combination of `display: flex`, `position: fixed`, and `overflow-y: auto` caused me issues in Firefox with modals that were long enough to require scrolling (I linked you the private repo PR in Slack)

brookescarlett: Fixed! Do you mind taking a look now?

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: I think I'm missing what you are getting at here and with the comment below?

packages/modal/src/Modal.tsx
jetpacmonkey: Hm. Soo.... my fix apparently broke a bunch of other stuff in the other repo. I'll let you know when I track down what happened.

packages/modal/src/Modal.tsx
jetpacmonkey: The comment below is unrelated (just pointing out that the name of the prop doesn't seem to match what it's called in prop types).

What I'm saying here is related to [this problem](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) outlined in the hooks FAQ.

brookescarlett: We used this hook with the Menu component, but also seems like it's okay because we aren't using information from the component's scope. The reason we're using the [] is just to ensure it's only called once. Let me know if this makes sense -- definitely open to updating as needed, just not sure I understand the problem yet.

brookescarlett:
@DesignerDave and @fredtruman what do we think about the sizing of modals? Are we going to expose all of these sizes or should we pick a few? And from there, how should we handle breakpoints?

packages/modal/src/Modal.tsx
brookescarlett: Got it!! THANK YOU

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,261 @@
...
+  XLarge: 'xlarge',
+} as const;
+
+export type ModalSize = typeof ModalSize[keyof typeof ModalSize];
+
+const defaultSpacing = 18;
+
+const mainContainer = css`
+  animation: fade-in 250ms ease-in-out;
+  -webkit-animation: fade-in 250ms ease-in-out;
```
hswolff: I think we can remove this and Emotion will do this for us.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,261 @@
...
+  border: 1px solid ${uiColors.gray.light3};
+  border-radius: 3px;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+  pointer-events: all;
+  height: 200vh;
+`;
+
+const modalSizes: { readonly [K in ModalSize]: string } = {
+  xxsmall: css`
+    min-width: inherit;
```
hswolff: Can we just upstream this to `modalContentStyle`?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,261 @@
...
+const modalContentStyle = css`
+  margin: ${defaultSpacing}px auto;
+  padding: 30px;
+  color: ${uiColors.gray.dark3};
+  background-color: ${uiColors.white};
+  border: 1px solid ${uiColors.gray.light3};
+  border-radius: 3px;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+  pointer-events: all;
+  height: 200vh;
```
hswolff: We're setting a fixed height for the modal?

packages/modal/src/Modal.tsx
hswolff: Bump.

packages/modal/src/Modal.tsx
hswolff: Bump.

packages/modal/src/Modal.tsx
hswolff: So do we need to append `setActive` here? I'm unclear if we do.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,261 @@
...
+    if (e.keyCode === escapeKey) {
+      handleClose();
+    }
+  }, []);
+
+  useEventListener('keydown', handleEscape, {
+    options: { once: true },
+  });
+
+  injectGlobal(`
```
hswolff: If this is ran multiple times does it inject this value multiple times?

Also how is this removed when the component unmounts?

packages/modal/src/Modal.tsx
brookescarlett: we need handleClose there and setActive in handleClose, both have been added

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: No sorry that was for testing the scroll!!

packages/modal/src/Modal.tsx
DesignerDave: Sorry – didn't see this.

No, I think that should be handled here. I don't see a reason that `Icon` should handle cursor display.

packages/modal/src/Modal.tsx
brookescarlett: we actually do not need to be using this anymore so I removed

packages/modal/src/Modal.tsx
brookescarlett: Added breakpoints based on the small, medium and large screen sizes defined in MMS. The only size that was effected was the large size. Let me know what you think @hswolff!

packages/modal/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+Determines active state of `Modal` component
+
+### setActive
+
+**Type:** `function`
+
+**Default:** `(bool) => {}`
+
+Callback to set active state of Modal component. `setActive` optionally accepts a boolean value, which will determine the active state of `Modal` component.
```
jetpacmonkey: "optionally accepts a boolean value"? Is it optional?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+const modalContentStyle = css`
+  margin: ${defaultSpacing}px auto;
+  padding: 36px;
+  color: ${uiColors.gray.dark3};
+  background-color: ${uiColors.white};
+  border: 1px solid ${uiColors.gray.light3};
+  border-radius: 3px;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+  pointer-events: all;
```
jetpacmonkey: This `pointer-events: all` is probably not necessary anymore

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+    width: 400px;
+  `,
+
+  default: css`
+    width: 720px;
+  `,
+
+  large: css`
+    ${mq({
+      width: ['720px', '720px', '960px'],
```
jetpacmonkey: I wonder if we could get a similar effect with `max-width: 90vw;` or something along those lines?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+      <div {...rest} onClick={handleDocumentClick} className={mainContainer}>
+        <div className={scrollContainer}>
+          <div
+            className={cx(modalContentStyle, modalSizes[size], className)}
+            tabIndex={-1}
+            ref={contentRef}
+          >
+            <Icon
+              glyph="X"
+              fill={'#5D6C74'}
```
jetpacmonkey: What is this color?

packages/modal/src/Modal.tsx
jetpacmonkey: Oh I was wrong, `shouldClose` is a real thing. But also we should add a propType for `setActive`

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const contentRef = useRef<HTMLDivElement>(null);
+
+  const handleClose = useCallback(() => {
+    // Don't close modal if shouldClose returns false or no setActive callback is passed
+    if (!setActive || (shouldClose && !shouldClose())) {
+      return;
+    }
+
+    setActive(false);
+  }, [setActive]);
```
jetpacmonkey: I think we should probably have `shouldClose` in this dependencies array, too

packages/modal/README.md
brookescarlett: Thank you, good call

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: We were thinking about leaving 3 fixed styles for now, and then adding a `full` size down the line that is responsive against the viewport

packages/modal/src/Modal.tsx
brookescarlett: From the new UI color palette

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+    setActive(false);
+  }, [setActive, shouldClose]);
+
+  const handleDocumentClick = (e: React.SyntheticEvent) => {
+    if (!contentRef.current) {
+      return;
+    }
+
+    if (!contentRef.current.contains(e.target as Node)) {
```
jetpacmonkey: Just came across this in the mms version. What happens if there's content inside a modal that uses a Portal? For instance, a dropdown. The target isn't going to be contained inside the modal in the DOM.

brookescarlett: Fixed! Good catch 😏

packages/modal/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/lib": "^3.0.0",
+    "@leafygreen-ui/portal": "^1.1.0",
+    "@leafygreen-ui/palette": "^1.0.0",
+    "@leafygreen-ui/hooks": "^1.0.0",
+    "@leafygreen-ui/icon": "^1.0.4"
```
DesignerDave: Amazing to see our work starting to mesh. Is this the most leafygreen-ui components that we've used as dependencies to a component we've built?

packages/modal/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js", 
+    "ts:emit": "tsc --project ./tsconfig.json"
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
```
DesignerDave: These should be peerDependencies.

packages/modal/src/Modal.spec.tsx
DesignerDave: My confusion comes down to the test name I think. Things are `rendered` in the DOM, and `visible` on the page. This test doesn't test specifics about the DOM.

We should probably test both whether it's rendered, and then its visibility.

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <h4>{modalContent}</h4>
+      </Modal>,
+    );
+
+    test('it appears on DOM when active prop is set', () => {
+      const modal = getByTestId(modalId);
+      expect(modal).toBeVisible();
+    });
+
+    test('it renders the children as expected', () => {
```
DesignerDave: I like dynamic test names. We could name this test:
```
`it renders ${modalContent}` within the modal
```

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      const content = getByText(modalContent);
+      expect(content).toBeInTheDocument();
```
DesignerDave: We should test that it renders within the modal rather than within the document.

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <Modal
+          active={true}
+          data-testid="falseshouldClose"
+          shouldClose={() => false}
+        >
+          <h4>test content</h4>
+        </Modal>,
+      );
+
+      const modal = getByTestId('falseshouldClose');
```
DesignerDave: Nice separation of component structure from test logic 👌

packages/modal/src/Modal.spec.tsx
DesignerDave: Bump

packages/modal/src/Modal.story.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { storiesOf } from '@storybook/react';
+import { select } from '@storybook/addon-knobs';
+import Modal, { ModalSize } from './index';
+
+function Example() {
+  const [active, setActive] = useState(false);
+
+  return (
+    <>
+      <button onClick={() => setActive(active => !active)}>Open Modal</button>
```
DesignerDave: Why not:
```
<button onClick={() => setActive(!active)}>Open Modal</button>
```

`active` is already available in scope without having to use a callback with `setActive`.

packages/modal/src/Modal.tsx
DesignerDave: You should either: Use the value straight from the palette (preferred) _or_ remove the wrapping `{}` for the property as the property can accept a string just fine without that.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const handleClose = useCallback(() => {
+    // Don't close modal if shouldClose returns false or no setActive callback is passed
+    if (!setActive || (shouldClose && !shouldClose())) {
+      return;
+    }
+
+    setActive(false);
+  }, [setActive, shouldClose]);
+
+  const handleDocumentClick = (e: React.SyntheticEvent) => {
```
DesignerDave: We should rename this to reflect that it's not an actual document click.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+  background-color: ${uiColors.white};
+  border: 1px solid ${uiColors.gray.light3};
+  border-radius: 3px;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+  min-width: inherit;
+  max-width: inherit;
+  position: relative;
+  pointer-events: all;
+  z-index: 1000;
+  transition: all 200ms ease-in-out;
```
DesignerDave: Nit: I prefer transitions first in CSS declarations.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+  color: ${uiColors.gray.dark3};
+  background-color: ${uiColors.white};
+  border: 1px solid ${uiColors.gray.light3};
+  border-radius: 3px;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+  min-width: inherit;
+  max-width: inherit;
+  position: relative;
+  pointer-events: all;
+  z-index: 1000;
```
DesignerDave: I really don't like arbitrary z-indexes. Can you explain why `1000` is needed, or why we need this at all?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+  `@media only screen and (max-width: ${small})`,
+  `@media only screen and (min-width: ${medium})`,
+  `@media only screen and (min-width: ${large})`,
+]);
+
+const defaultSpacing = 18;
+
+const fadein = keyframes`
+  from {
+    margin-top: 50px;
```
DesignerDave: Never animate margin, since it's not hardware accelerated and looks jumpy. Use `translate3d` for best performance.

DesignerDave: Additionally, the translate transition should not be on the backdrop, only on the content.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+  animation: fade-in 250ms ease-in-out;
+  background-color: ${transparentize(0.4, uiColors.black)};
+  overflow-y: auto;
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  z-index: 1000;
+  animation: ${fadein} 300ms ease-in-out;
```
DesignerDave: Let's double the speed here (150ms).

packages/modal/src/Modal.tsx
DesignerDave: Additionally, having `z-index: 1000` in 3 places here is a red flag. I'm certain we don't need all instances of this.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+    margin-top: 50px;
+    opacity: 0;
+  }
+  to {
+    margin-top: 0;
+    opacity: 1;
+  }
+`;
+
+const mainContainer = css`
```
DesignerDave: Maybe call this "backdrop"? We have a lot of "container"s.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+const modalContentStyle = css`
+  margin: ${defaultSpacing}px auto;
+  padding: 36px;
+  color: ${uiColors.gray.dark3};
+  background-color: ${uiColors.white};
+  border: 1px solid ${uiColors.gray.light3};
+  border-radius: 3px;
+  box-shadow: 0 5px 15px rgba(0, 0, 0, 0.3);
+  min-width: inherit;
```
DesignerDave: I might be missing something, but where are we inheriting from?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+  justify-content: center;
+  z-index: 1000;
+`;
+
+const modalContentStyle = css`
+  margin: ${defaultSpacing}px auto;
+  padding: 36px;
+  color: ${uiColors.gray.dark3};
+  background-color: ${uiColors.white};
+  border: 1px solid ${uiColors.gray.light3};
```
DesignerDave: Not a fan of borders on modals, and this one doesn't add anything. Let's get rid of this.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+    opacity: 0;
+  }
+  to {
+    margin-top: 0;
+    opacity: 1;
+  }
+`;
+
+const mainContainer = css`
+  animation: fade-in 250ms ease-in-out;
```
DesignerDave: Let's have a consistent timing on our animations.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,258 @@
...
+    options: { once: true },
+  });
+
+  return (
+    <Portal>
+      <div {...rest} onClick={handleDocumentClick} className={mainContainer}>
+        <div className={scrollContainer}>
+          <div
+            className={cx(modalContentStyle, modalSizes[size], className)}
+            tabIndex={-1}
```
DesignerDave: Generally I prefer static property values to be declared as strings to mirror HTML.

DesignerDave: Additionally, the modal still appears to be focusable when testing it.

packages/modal/src/index.ts
```diff
@@ -0,0 +1,4 @@
+import Modal from './Modal';
+export { ModalSize } from './Modal';
```
DesignerDave: Do we need to export this?

package.json
```diff
@@ -120,5 +121,8 @@
     "packages": [
       "packages/*"
     ]
+  },
+  "dependencies": {
```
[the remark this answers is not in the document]
brookescarlett: All of these? That's not how we've done it for other packages in the past, but happy to fix here and moving forward

packages/modal/package.json
brookescarlett: <3

package.json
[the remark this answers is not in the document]
brookescarlett: devDependency, that's my bad.

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: This is here because when I tested the component in MMS, the side and top navs were not hidden. They have z-indexes of 998 (possibly 997), so I did this to ensure that the backdrop would conceal the side/left nav in MMS. I can add a comment to indicate that

packages/modal/src/Modal.tsx
brookescarlett: Got it!

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: Got it

packages/modal/src/index.ts
brookescarlett: Nope!

brookescarlett: Fixed

packages/modal/src/Modal.tsx
brookescarlett: Can you explain what you mean by static property values being declared as strings, sorry!

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: What would you like to rename to? It is listening for clicks on the document and then deciding whether or not to close the modal

packages/modal/package.json
DesignerDave: devDependencies*

brookescarlett: `Could not find a declaration file for module '@leafygreen-ui/icon'` when peerDeps

packages/modal/src/Modal.tsx
DesignerDave: Details related to a specific app's implementation should be left to them. if there's not a way currently for them to implement this, we should add a way.

packages/modal/package.json
brookescarlett: Got it with the dev

packages/modal/src/Modal.tsx
DesignerDave: This value will always be `-1`, so we can simply say `tabIndex='-1'` for the property instead of indicating it needs to be evaluated as JS with `{-1}`

packages/modal/src/Modal.tsx
brookescarlett: so expose a zIndex prop?

packages/modal/package.json
DesignerDave: Yeah, sorry that was my bad.

packages/modal/src/Modal.tsx
DesignerDave: It's not listening to clicks on the document though. The onClick handler is on the backdrop.

packages/modal/src/Modal.tsx
brookescarlett: Typescript complains and says that tabIndex should be a number or undefined when you pass a string

packages/modal/src/Modal.tsx
brookescarlett: It is listening for clicks on the entire document, if you click on the modal content itself it'll still hear the click

brookescarlett: handleBackdropClick

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.story.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: @DesignerDave ^

packages/modal/src/Modal.tsx
brookescarlett: Fixed!

packages/modal/src/Modal.tsx
DesignerDave: Only if this needs to be applied to a different element than className.

packages/modal/src/Modal.tsx
DesignerDave: Ahh okay. 👍

packages/modal/src/Modal.tsx
brookescarlett: So omit the zIndex 1000 and let MMS handle that with className on their own?

brookescarlett:
@jetpacmonkey will tackle as soon as goes in: https://jira.mongodb.org/browse/PD-336

packages/lib/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@leafygreen-ui/emotion": "^1.0.2",
+    "facepaint": "^1.2.1",
```
hswolff: I wouldn't normally want to include this by default in every package however:

- This package itself is very small
- We'll definitely be using it in other components

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,249 @@
...
+export const ModalSize = {
+  Small: 'small',
+  Default: 'default',
+  Large: 'large',
+} as const;
+
+export type ModalSize = typeof ModalSize[keyof typeof ModalSize];
+
+// breakpoints for different screen sizes
+const small = '767px'; // mobile screens, from 0px - 767px
```
hswolff: Are these going to be global breakpoints we'll be using in other libraries? Should we extract to `lib` or some other package?

Ditto with the facepaint instance?

packages/modal/src/Modal.tsx
hswolff: No hover?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,249 @@
...
+   * Content that will appear inside of the Modal component.
+   */
+  children: React.ReactNode;
+
+  /**
+   * Determines the active state of the modal
+   *
+   * default: `false`
+   */
+  active: boolean;
```
hswolff: Curious why we landed on active to control if the modal is open/visible/shown (other words I've seen used for this)?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,249 @@
...
+}: ModalProps) {
+  if (!active) {
+    return null;
+  }
+
+  const scrollContainerRef = useRef<HTMLDivElement>(null);
+
+  const handleClose = useCallback(() => {
+    // Don't close modal if shouldClose returns false or no setActive callback is passed
+    if (!setActive || (shouldClose && !shouldClose())) {
```
hswolff: Can we set default arguments of noop functions for setActive and shouldClose and then the code path becomes more straightforward?

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,249 @@
...
+      return;
+    }
+
+    setActive(false);
+  }, [setActive, shouldClose]);
```
hswolff: If a user doesn't wrap these functions with `useCallback` they're always going to be different on every run, which may reduce the value of using `useCallback` to begin with. May be fine to skip `useCallback` here...or keep it. Not sure what the real difference is going to be.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,249 @@
...
+    (e: KeyboardEvent) => {
+      if (e.keyCode === escapeKey) {
+        handleClose();
+      }
+    },
+    [handleClose],
+  );
+
+  useEventListener('keydown', handleEscape, {
+    options: { once: true },
```
hswolff: Do we want once here? What if a user types in 'f'? Won't that remove this event listener?

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+      const modal = getByTestId(modalId);
+      expect(modal).toBeVisible();
+    });
+
+    test('it is rendered to the DOM when the active prop is set', () => {
+      const modal = getByTestId(modalId);
+      expect(modal).toBeInTheDocument();
+    });
+
+    test(`it renders the ${modalContent} as expected`, () => {
```
hswolff: Kinda weird to interpolate the actual value. we just want to make sure it renders our content, not what the content is necessarily. That's up to the assertion.

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+import 'jest-dom/extend-expect';
+import Modal from './Modal';
+
+afterAll(cleanup);
+
+describe('packages/Modal', () => {
+  const modalId = 'modal';
+  const modalContent = 'Modal Content';
+
+  describe('when rendered with content', () => {
```
hswolff:
```suggestion
  describe('when rendered with content and set as active', () => {
```

Then we can remove the redundant 'when active props is set` in the child test blocks.

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import 'jest-dom/extend-expect';
+import Modal from './Modal';
+
+afterAll(cleanup);
+
+describe('packages/Modal', () => {
```
hswolff: Can we add a test for escape key? Both after a user types in one character and then the esc key?

packages/lib/package.json
brookescarlett: Kk, should I leave or move to Modal alone?

packages/modal/src/Modal.tsx
brookescarlett: I think it might require a little more discussion? I ported these specific breakpoints from MMS, but can move to lib or somewhere else if we feel that it belongs there now? Let me know what you think

packages/modal/src/Modal.tsx
brookescarlett: @DesignerDave  and I actually had a discussion about this: active has been used in other components, so we did it for consistency.

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: You're correct. Removed!

packages/modal/src/Modal.spec.tsx
brookescarlett: 👍

brookescarlett: Removed the once only, so not sure we need both but happy to include

packages/modal/src/Modal.tsx
brookescarlett: I think we did it because in order for the useCallback in handleEscape to work correctly we'd have to include setActive in the list of dependencies, but it made more semantic sense to break it out the way that we did. (i.e. handleEscape includes handleClose and handleClose includes setActive) cc: @jetpacmonkey

packages/modal/src/Modal.tsx
DesignerDave: I think he was referring to the vendor prefixed CSS rule.

packages/modal/src/Modal.tsx
DesignerDave: As a general rule, we should avoid adding props like that, especially before they're asked for.

packages/modal/src/Modal.tsx
DesignerDave: Yeah, seems fine. I also like the "exit code" idea

packages/modal/src/Modal.tsx
DesignerDave: Let's transition it over 100ms to dark gray 2.

brookescarlett: 👍

packages/modal/src/Modal.tsx
jetpacmonkey: Yeah the chain of dependencies was easier to follow this way. Having `setActive` as a dependency in a function that doesn't directly reference `setActive` seemed likely to be confusing.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,251 @@
...
+  active = false,
+  size = ModalSize.Default,
+  setActive = () => {},
+  shouldClose = () => true,
+  children,
+  className,
+  ...rest
+}: ModalProps) {
+  if (!active) {
+    return null;
```
DesignerDave: Good catch in the conversation in the team channel – this prevents animating the modal out.

Also a side-note: we aren't animating the backdrop in at all – it just flashes onto the page.

DesignerDave: Should be an interesting challenge...

DesignerDave: From @jetpacmonkey:
> I’d say, persist the modal on the screen while having it fade out, and after the animation ends render null

brookescarlett:
Fixed the animation so now it fades both in and out, cc: @DesignerDave :)

packages/modal/src/Modal.tsx
brookescarlett: Solved 👍

packages/modal/src/Modal.spec.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const modalContent = 'Modal Content';
+
+  describe('when rendered with content and set as active', () => {
+    const { container, getByText, getByTestId } = render(
+      <Modal active data-testid={modalId}>
+        <h4>{modalContent}</h4>
+      </Modal>,
+    );
+
+    test('it appears on page when active prop is set', () => {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,264 @@
...
+              <div
+                className={cx(
+                  modalContentStyle,
+                  modalSizes[size],
+                  {
+                    [visibleModalContentStyle]: state === 'entered',
+                  },
+                  className,
+                )}
+                tabIndex={-1}
```
DesignerDave: Let's get rid of this. It's actually making the modal highlight as if it were focused when it's clicked.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+const closeButton = css`
+  color: ${uiColors.gray.dark1};
+  position: absolute;
+  cursor: pointer;
+  right: 12px;
+  top: 12px;
+
+  &:hover {
+    transition: color 100ms ${uiColors.gray.dark2};
```
DesignerDave: So the transition property doesn't actually accept values.

Anyways, I realized that I missed a huge reason to have icons use currentColor – the fill for the close button can be transitioned with CSS pseudo-classes. I'll make a ticket for this, and in the meantime, let's remove the hover selector, and the color property until this is added.

DesignerDave: https://jira.mongodb.org/browse/PD-355

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,264 @@
...
+  shouldClose = () => true,
+  children,
+  className,
+  ...rest
+}: ModalProps) {
+  const scrollContainerRef = useRef<HTMLDivElement>(null);
+
+  const handleClose = useCallback(() => {
+    // Don't close modal if shouldClose returns false or no setOpen callback is passed
+    if (!setOpen || !shouldClose()) {
```
DesignerDave: We could probably write this like:
```js
const handleClose = useCallback(() => {
  if (setOpen && shouldClose()) {
    setOpen(false);
  }
}, [setOpen, shouldClose]);
```

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,264 @@
...
+      return;
+    }
+
+    setOpen(false);
+  }, [setOpen, shouldClose]);
+
+  const handleBackdropClick = (e: React.SyntheticEvent) => {
```
DesignerDave: Could probably write this like:
```js
const handleBackdropClick = (e: React.SyntheticEvent) => {
  if (scrollContainerRef.current && e.target === scrollContainerRef.current) {
    handleClose();
  }
};
```

packages/modal/README.md
```diff
@@ -0,0 +1,104 @@
...
+      <button onClick={() => setOpen(open => !open)}>Open Modal</button>
+      <Modal open={open} setOpen={setOpen}>
+        Modal Content goes here
+      </Modal>
+    </>
+  );
+}
+```
+
+**Output HTML**
```
DesignerDave: The output here is out-of-date.

In the future, we may wanna hold out on adding this until we finalize the component itself.

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: happy path 💯

packages/modal/src/Modal.tsx
brookescarlett: 👍

packages/modal/README.md
brookescarlett: Yep, got it!

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,252 @@
...
+        handleClose();
+      }
+    },
+    [handleClose],
+  );
+
+  useEventListener('keydown', handleEscape);
+
+  return (
+    <Transition in={open} timeout={500} mountOnEnter unmountOnExit>
```
hswolff: I'm not seeing any transition when running this locally.

packages/modal/src/Modal.tsx
```diff
@@ -0,0 +1,252 @@
+import React, { useCallback, SetStateAction, useRef } from 'react';
```
hswolff: When the content of the modal is larger than the viewport the modal content does not scroll. This is a tricky problem, may want to pair with @jetpacmonkey on this.

packages/modal/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js",
+    "ts:emit": "tsc --project ./tsconfig.json"
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/modal/src/Modal.tsx
brookescarlett: I'm seeing it locally... I can send a video to the channel!

packages/modal/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+### size
+
+**Type:** `string`
+
+**Default:** `default`
+
+Determines `Modal` size. Options: `['small', 'default', 'large']`
+
+### modalShouldClose
```
hswolff: This should be shouldClose.

packages/modal/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+### children
+
+**Type:** `node`
+
+**Default:** `null`
+
+Children that will be rendered inside `Modal` component.
+
+### modalShouldClose
```
hswolff: It's now duped!

packages/modal/README.md
hswolff: And default is `()=> true`.

packages/modal/README.md
brookescarlett: 👍

packages/modal/README.md
brookescarlett: 👍

PR #100, opened 2019-06-20
PD-330: Correctly apply disabled styles on a Button when used as an anchor element

gssbzn:
## ✍️ Proposed changes

The disabled styles weren't being correctly applied to the component if the root was an anchor, this PR looks into solving this

🎟 _Jira ticket:_ [PD-330](https://jira.mongodb.org/browse/PD-330)

## 🛠 Types of changes

## ✅ Checklist


packages/button/src/Button.tsx
```diff
@@ -302,6 +306,8 @@ export default function Button(props: ButtonProps) {
       baseStyle,
       buttonSizes[size],
       buttonVariants[variant],
+      { [disabledStyle]: disabled },
+      { [disabledAnchor]: disabled && usesLinkElement(props) },
       className,
     ),
     disabled,
```
DesignerDave: We should remove `disabled` from `commonProps`, and apply it only to the button.

packages/button/src/Button.tsx
```diff
@@ -238,14 +238,18 @@ const baseStyle = css`
...
+const disabledStyle = css`
+  color: ${colors.gray[3]};
+  border-color: ${colors.gray[5]};
+  background-color: ${colors.gray[7]};
+  background-image: none;
+  box-shadow: none;
+`;
+
+const disabledAnchor = css`
+  pointer-events: none;
```
DesignerDave: Is there any reason this rule shouldn't be a part of `disabledStyle`? I don't see a reason to not have this be a part of both the button and anchor tag styles.

packages/button/src/Button.tsx
gssbzn: Should we? It may be tricky when someone uses this with a custom root that supports `disabled`, a checkbox or radio button for example

packages/button/src/Button.tsx
gssbzn: I don't know to be honest, felt unnecessary for other cases but I'll merge them

packages/button/src/Button.tsx
DesignerDave: Maybe we could only exclude it from the link then, since we know it's an unsupported attribute.

Side-note: I was playing with the branch, and noticed that a recent fix for the button style in Firefox seems to have broken the styling of the button when rendered as an anchor tag. The text will show as being misaligned in that case.

Adding the following rules to the button styles seems to fix it:
```js
display: inline-flex;
align-items: center;
```

packages/button/src/Button.tsx
DesignerDave: 👍

packages/button/src/Button.tsx
gssbzn: Added this and did some cleanup on the styles as well, there were some duplication/overriding from the base to the variation/sizes
Also added a text decoration none when hovering, links would show underlined

packages/button/src/Button.tsx
gssbzn: done

gssbzn:
someone may need to enable circleci for forks
Screenshot 2019-06-21 at 11 49 03

hswolff:
Updated.

PR #102, opened 2019-06-24
PD-309: Card Component

brookescarlett:
## ✍️ Proposed changes

Created a card component and wrote tests to test the functionality. To note: this component is largely just a display component.

🎟 _Jira ticket:_ [PD-309](https://jira.mongodb.org/browse/PD-309)

## 🛠 Types of changes

## ✅ Checklist

- [ x] I have added tests that prove my fix is effective or that my feature works


packages/card/README.md
```diff
@@ -0,0 +1,50 @@
...
+  className='card-styles'
+>
+  This is my card component
+</Card>
+```
+
+**Output HTML**
+
+```HTML
+ <section class="leafygreen-ui-1xf4bcx">This is a card component</section>
```
DesignerDave: Maybe `<article>` is a better semantic tag to use here. @hswolff Do you have a preference?

packages/card/src/Card.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+
+const containerStyle = css`
+  background-color: white;
```
DesignerDave: Could ya add a transition to the border and box-shadow?

packages/card/src/Card.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+`;
+
+interface CardProps {
+  children: React.ReactNode;
+  as?: React.ElementType<any>;
+  className?: string;
+}
+
+function Card({ children, as = 'div', className, ...rest }: CardProps) {
```
DesignerDave: The default for `as` of `div` is inconsistent with your README (additionally we might want to default to something more semantic like `<article>`.

packages/card/src/Card.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const Root = as;
+  return (
```
DesignerDave: Line break above please

packages/card/README.md
brookescarlett: It's defaulting to div, section was just an example of using the as prop.

packages/card/src/Card.tsx
brookescarlett: what kind of transition?

packages/card/src/Card.tsx
brookescarlett:
```<Card
  as='section'
  className='card-styles'
>
  This is my card component
</Card>```

Comment (brookescarlett) [reply packages/card/src/Card.tsx]: just showing how an as work, can default to whatever you feel makes the most sense.

File: packages/card/src/Card.tsx (hunk 4 above)
Comment (brookescarlett) [reply packages/card/src/Card.tsx]: 👍

File: packages/card/README.md (hunk 1 above)
Comment (hswolff) [reply packages/card/README.md]: I don't have a strong preference.

I always feel weird using article as it's so semantically tied to news sites, which doesn't apply to most web apps.

Not sure which description on mdn applies best.

File: packages/card/src/Card.tsx [hunk 5]
```diff
@@ -0,0 +1,42 @@
...
+
+  &:hover {
+    border: 1px solid ${uiColors.gray.light2};
+    box-shadow: 0px 3px 6px -2px ${uiColors.gray.base};
+  }
+`;
+
+interface CardProps {
+  children: React.ReactNode;
+  as?: React.ElementType<any>;
```
hswolff: I think TS has all the available DOM elements that we can include as the generic value.

brookescarlett: But what if someone wants to use as a <Link> element or something like that

brookescarlett: Hi @hswolff! Want to merge this in, but first wanted to make sure you were okay with leaving as-is. If you disagree, I'll update the type!

hswolff: Couldn't we have TS support both? 

Also what does it look like to have a Link as a card? That seems very odd.

PR #107, opened 2019-07-01
PD-333: Fix new glyphs not having fill prop applied correctly

DesignerDave:
## ✍️ Proposed changes

This PR sets the fill for the newer glyphs to `#000000` and adds tests to verify that this is the case with all new glyphs in the future. In the future, we should find a way to do this more resiliently (such as an implementation with `currentColor`, but this should make us less error-prone in the meantime.

🎟 _Jira ticket:_ [Fix new glyphs not having fill prop applied correctly](https://jira.mongodb.org/browse/PD-333)

## 🛠 Types of changes

## ✅ Checklist


package.json
```diff
@@ -118,5 +118,5 @@
     "packages": [
       "packages/*"
     ]
-  }
+  }
```
hswolff: y u add white space

packages/icon/src/Icon.spec.tsx
```diff
@@ -1,20 +1,71 @@
...
 import { typeIs } from '@leafygreen-ui/lib';
 import { createIconComponent } from '.';
 
 afterAll(cleanup);
 
+describe('packages/Icon/glyphs/', () => {
+  const glyphPaths = fs
+    .readdirSync(path.resolve(__dirname, './glyphs'))
+    /**
+     * './glyphs/' contains an index.ts file, so we filter out
```
hswolff: Can we also match that this list aligns with the exported object given from `index.ts`?

packages/icon/src/Icon.spec.tsx
```diff
@@ -1,20 +1,71 @@
...
+     * anything that's not an svg.
+     */
+    .filter(path => /.*\.svg$/.test(path));
+
+  glyphPaths.forEach(glyphPath => {
+    describe(`${glyphPath}`, () => {
+      const { svg } = require(path.resolve(__dirname, `./glyphs/${glyphPath}`));
+
+      if (!svg) {
+        return;
```
hswolff: Should we fail instead? This shouldn't ever be the case. `expect(svg).not.toBeNull()`?

packages/icon/src/Icon.spec.tsx
```diff
@@ -1,20 +1,71 @@
...
+        Object.keys(obj).forEach(key => {
+          const currentValue = obj[key];
+
+          if (typeof currentValue === 'object') {
+            processGlyphObject(currentValue);
+          } else {
+            if (key === 'fill') {
+              const validFills = ['#000', '#000000', 'none'];
+
+              if (!validFills.includes(currentValue)) {
```
hswolff: Could this not be the test itself? Such that if it fails the test shows what value it currently is? Maybe it doesn't matter?

package.json
DesignerDave: Hmm our format script should've fixed this...

packages/icon/src/Icon.spec.tsx
DesignerDave: Actually, I think we can get rid of this altogether. We already explicitly filter out non-svg files from `glyphPaths`. I'm just gonna remove this.

packages/icon/src/Icon.spec.tsx
DesignerDave: I did something that _almost_ does what you're referring to. Unfortunately you won't see the current value, as I couldn't find a way to do something like `expect(currentValue).toBeWithin(validFills)`. If you have a better idea, let me know, but the new implementation is an improvement.

packages/icon/src/Icon.spec.tsx
DesignerDave: Done!

PR #109, opened 2019-07-09
PD-311: Implement Tabs Component

brookescarlett:
## ✍️ Proposed changes

Creating a PR for the Tabs component. It was created with these guidelines in mind: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html. 

To note, I have not finished testing the component, but as Skunkworks is approaching quickly I wanted to start getting feedback on the component.

Screen Shot 2019-07-09 at 10 45 04 AM
(Non-focused)

Screen Shot 2019-07-09 at 10 45 10 AM
(Focused state)

🎟 _Jira ticket:_ [PD-311](https://jira.mongodb.org/browse/PD-311)

## 🛠 Types of changes

## ✅ Checklist


packages/tabs/src/TabTitle.tsx
```diff
@@ -0,0 +1,80 @@
...
+      className={className}
+      id={id}
+      role="tab"
+      data-tab-id={dataTabId}
+      onClick={onClick}
+      onKeyDown={onKeyDown}
+      aria-controls={ariaControls}
+      aria-selected={active}
+      aria-disabled={disabled}
+      tabIndex={active ? 0 : -1}
```
DesignerDave: I _think_ it should be fine to do `tabIndex={!active && -1}`.

I always prefer not setting the prop at all if possible.

packages/tabs/src/TabTitle.tsx
```diff
@@ -0,0 +1,80 @@
...
+      onBlur={() =>
```
DesignerDave: I'd prefer defining these using `useCallback` rather than inlining the functions.

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,285 @@
...
+import Tab, { TabProps } from './Tab';
+import TabTitle from './TabTitle';
+
+const listStyle = css`
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  list-style: none;
+  padding: 0px;
+  margin: 20px 0px 5px 0;
```
DesignerDave: Prefer not setting a margin at all for values that would be `0px`. Use `margin-top` and `margin-bottom` instead.

DesignerDave: Unless those margins are set by default, of course.

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,285 @@
...
+import { uiColors } from '@leafygreen-ui/palette';
+import Tab, { TabProps } from './Tab';
+import TabTitle from './TabTitle';
+
+const listStyle = css`
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  list-style: none;
+  padding: 0px;
```
DesignerDave: Why?

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,285 @@
+import React, { useState, useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+import Tab, { TabProps } from './Tab';
+import TabTitle from './TabTitle';
+
+const listStyle = css`
+  display: flex;
+  flex-direction: row;
```
DesignerDave: This is the default for `flex-direction`.

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,285 @@
+import React, { useState, useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+import Tab, { TabProps } from './Tab';
+import TabTitle from './TabTitle';
+
+const listStyle = css`
+  display: flex;
```
DesignerDave: Looking at the flex properties on this, it looks a lot like this doesn't need to use flexbox.

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,285 @@
...
+ * @param props.className Classname applied to Tabs container.
+ */
+function Tabs({ children, onChange, selected, className, ...rest }: TabsProps) {
+  const [activeTab, setActiveTab] = useState();
+  const [currentIndex, setCurrentIndex] = useState();
+  const [focusedState, setFocusedState] = useState([] as Array<string>);
+  const listRef = useRef<HTMLUListElement>(null);
+
+  useEffect(() => {
+    const currIdx = React.Children.map(
```
DesignerDave: Could we spell out `currentIndex` instead?

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        return [...acc, index];
+      },
+      [] as Array<number>,
+    );
+
+    const enabledCurrentIndex = enabledIndexes.indexOf(currentIndex);
+
+    let idx: number;
```
DesignerDave: Let's spell out `index`.

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,285 @@
...
+        if (child.props.disabled) {
+          return acc;
+        }
+
+        return [...acc, index];
+      },
+      [] as Array<number>,
+    );
+
+    const enabledCurrentIndex = enabledIndexes.indexOf(currentIndex);
```
DesignerDave: This can just be `currentIndex`, right?

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,285 @@
...
+      case 'ArrowUp':
+        idx =
+          (enabledCurrentIndex - 1 + enabledIndexes.length) %
+          enabledIndexes.length;
+        setActiveTab(children[enabledIndexes[idx]].props.value);
+        break;
+    }
+  }
+
+  function isTab(
```
DesignerDave: This can be declared outside the component render function.

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  return (
+    <Root
+      {...rest}
+      disabled={disabled}
+      aria-disabled={disabled}
+      aria-selected={active}
+      aria-controls={`tab-${index}`}
+      data-tab-id={value}
+      className={className}
```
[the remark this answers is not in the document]
brookescarlett: rest handle className?

packages/tabs/src/Tabs.tsx
brookescarlett: No, because we have to make sure it's iterating through only the enabled tabs, not all of the tabs.

packages/tabs/package.json
```diff
@@ -0,0 +1,20 @@
...
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/lib": "^3.0.0",
+    "@leafygreen-ui/theme": "^1.2.0"
+  },
+  "gitHead": "53385cfed5ecf3101d098d26dedc6a40b27c402a"
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/TabTitle.tsx
brookescarlett: Typescript doesn't allow that, must be a number or undefined. Boolean is not a valid type.

packages/tabs/src/TabTitle.tsx
```diff
@@ -0,0 +1,80 @@
...
+  onKeyDown?: React.KeyboardEventHandler;
+  ariaControls?: string;
+  setFocusedState: React.Dispatch<SetStateAction<Array<string>>>;
+}
+
+function TabTitle({
+  active,
+  children,
+  disabled,
+  className,
```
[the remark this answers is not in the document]
brookescarlett: using active and disabled to set aria-active and aria-disabled, but className can be handled through rest.

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: you're right 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: Why to padding?

packages/tabs/src/Tab.tsx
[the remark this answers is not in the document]
DesignerDave: Yeah, I don't see a reason not here to unless you do.

packages/tabs/src/TabTitle.tsx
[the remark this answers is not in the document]
DesignerDave: `id` could as well I believe.

packages/tabs/src/TabTitle.tsx
DesignerDave: 🤷‍♂ Fair enough

packages/tabs/src/Tabs.tsx
DesignerDave: Yeah, is this needed?

brookescarlett: 👍

brookescarlett: Yeah it is needed, this is what it looks like without: 
Screen Shot 2019-07-11 at 9 25 53 AM
Screen Shot 2019-07-11 at 9 25 55 AM

packages/tabs/src/TabTitle.tsx
brookescarlett: 👍

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,115 @@
...
+  // default is not an HTML property
+  delete rest.default;
+
+  return (
+    <Root
+      {...rest}
+      disabled={disabled}
+      aria-disabled={disabled}
+      aria-selected={active}
+      aria-controls={`tab-${index}`}
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/Tab.tsx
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const Root = as;
+
+  // default is not an HTML property
+  delete rest.default;
+
+  return (
+    <Root
+      {...rest}
+      disabled={disabled}
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/TabTitle.tsx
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/TabTitle.tsx
brookescarlett: 👍

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,115 @@
...
+import PropTypes from 'prop-types';
+
+export interface TabProps {
+  /**
+   * Value supplied to the data-tab-id attribute. Used to determine what Tab is active.
+   */
+  value: string;
+
+  /**
+   * Set internally. Used to determine if the Tab is active.
```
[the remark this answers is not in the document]
brookescarlett: We're setting it from the Tabs component, whether or not the component is controlled. Consumers should never interact with the prop. Maybe I should remove from readme/tsdocs? 
@hswolff, thoughts?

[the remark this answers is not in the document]
DesignerDave: If they shouldn't interface with it, it probably shouldn't be documented as if they should.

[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,312 @@
...
+
+const listTitle = css`
+  display: inline-block;
+  background-color: ${uiColors.white};
+  color: ${uiColors.gray.base};
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 1;
+  text-transform: capitalize;
+  padding: 3px 24px;
```
DesignerDave: So I found a UX problem with these tabs:
https://drive.google.com/file/d/1lb-aZYBgtQJvQXGi6uuHPxia-XlX4X7h/view?usp=sharing

You'll notice that from a few pixels above the bottom border to the bottom of said border, the tab isn't actually clickable. We should make it so that that region is also clickable.

packages/tabs/src/TabTitle.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function TabTitle({
+  active,
+  children,
+  disabled,
+  dataTabId,
+  onClick,
+  onKeyDown,
+  ariaControls,
```
DesignerDave: What's the reason for explicitly declaring this?

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,312 @@
...
+`;
+
+const listTitle = css`
+  display: inline-block;
+  background-color: ${uiColors.white};
+  color: ${uiColors.gray.base};
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 1;
+  text-transform: capitalize;
```
DesignerDave: This doesn't seem to be doing anything?

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,312 @@
...
+import TabTitle from './TabTitle';
+
+const listStyle = css`
+  list-style: none;
+  padding: 0px;
+  margin-top: 20px;
+  margin-bottom: 5px;
+`;
+
+const listTitle = css`
```
DesignerDave: Shouldn't these styles be baked into the tab title component instead of be in this file?

packages/tabs/src/TabTitle.tsx
brookescarlett: https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-controls
From the W3C docs, a use case of aria-controls: "A tab controls the display of its associated tab panel."
Aria-controls denote a relationship, or when the presence of one thing is controlled by something else

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,316 @@
...
+      },
+      [] as Array<number>,
+    );
+
+    const enabledCurrentIndex = enabledIndexes.indexOf(currentIndex);
+
+    let index: number;
+
+    switch (e.key) {
+      case 'ArrowRight':
```
hswolff: We want arrow up and arrow down to have the same behavior? Doesn't seem that's according to [the a11y guidelines](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html).

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,114 @@
...
+export interface TabProps {
+  /**
+   * Value supplied to the data-tab-id attribute. Used to determine what Tab is active.
+   */
+  value: string;
+
+  /**
+   * Content that will appear as the title in the Tab list.
+   */
+  tabTitle: string;
```
hswolff: Can this just be `title`? Reads redundant when written out: `<Tab tabTitle="Blah" />`.

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,114 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export interface TabProps {
+  /**
+   * Value supplied to the data-tab-id attribute. Used to determine what Tab is active.
+   */
+  value: string;
```
hswolff: This is an implementation detail right? Can we make this optional? Do we really need to force the user to set this?

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,316 @@
...
+  // using an object here in order to alert consumer if duplicate values are used
+  const tabObject = childrenArray.reduce(
+    (acc, child, index) => {
+      if (!isTab(child)) {
+        return { ...acc, [index]: child };
+      }
+
+      const childValue = child.props.value;
+
+      if (Object.keys(acc).includes(childValue)) {
```
hswolff:
```suggestion
      if (acc[childValue] != null) {
```

And then for the initial value for accumulator we can do:

```
{} as { [key: string]: React.ReactElement },
```

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,316 @@
...
+    }
+  }
+
+  function calcStyle() {
+    if (
+      !tabListRef ||
+      !tabListRef.current ||
+      typeof currentIndex !== 'number'
+    ) {
+      return css``;
```
hswolff:
```suggestion
      return null;
```

Let's make it a little more lightweight.

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    const tabListChildren: Array<Element> = Array.from(
+      tabListRef.current.children,
+    );
+
+    let computedX = 0;
+
+    for (let i = 0; i < currentIndex || 0; i++) {
```
hswolff: Why do we need `|| 0`?

packages/tabs/src/TabTitle.tsx
```diff
@@ -0,0 +1,154 @@
...
+    color: ${uiColors.blue.base};
+    outline: none;
+  }
+
+  &:hover:not(:focus) {
+    color: ${uiColors.gray.dark3};
+  }
+`;
+
+interface SharedTabTitleProps {
```
hswolff: I'm wondering if this should extend from `TabProps` as that's the public interface and this has to in many ways mirror what public props people are using.

packages/tabs/src/TabTitle.tsx
```diff
@@ -0,0 +1,154 @@
...
+    ariaControls,
+    setFocusedState,
+    as,
+    ...rest
+  } = props;
+
+  const titleRef = useRef<HTMLElement>(null);
+
+  useEffect(() => {
+    if (titleRef.current && active && !disabled) {
```
hswolff:
```suggestion
    if (!disabled && active && titleRef.current) {
```

Micro optimization. The way you have it now we'll have to do a property lookup every iteration. 

Here we're just checking the state of the variable.

This is the smallest of perf optimizations that it's almost silly to worry about but it's easy to implement which is why I'm pointing it out.

Also disabled being false is the most likely case so checking for that first should also help speed things up.

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,114 @@
...
+      aria-selected={active}
+      aria-controls={ariaControl}
+      data-tab-id={value}
+      role="tabpanel"
+    >
+      {children}
+    </div>
+  );
+
+  return renderTab();
```
hswolff: Looks like a holdover from a refactor. This doesn't need to be an inline function right?

packages/tabs/src/TabTitle.tsx
```diff
@@ -0,0 +1,154 @@
...
+    setFocusedState((curr: Array<string>) =>
+      curr.filter(el => dataTabId !== el),
+    );
+  }, [setFocusedState]);
+
+  const onFocus = useCallback(() => {
+    setFocusedState((curr: Array<string>) => [...curr, dataTabId]);
+  }, [setFocusedState]);
+
+  const renderTabTitle = (Root: React.ElementType<any> = 'button') => (
```
hswolff: Trying to understand, how does someone set this to a different value? via `<Tab as={Link} />`?

packages/tabs/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@leafygreen-ui/tabs",
+  "version": "0.9.0",
+  "description": "leafyGreen UI Kit Tabs",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js",
+    "ts:emit": "tsc --project ./tsconfig.json"
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,316 @@
...
+  className,
+  as = 'button',
+  ...rest
+}: TabsProps) {
+  const [activeTab, setActiveTab] = useState();
+  const [currentIndex, setCurrentIndex] = useState();
+  const [focusedState, setFocusedState] = useState([] as Array<string>);
+  const tabListRef = useRef<HTMLDivElement>(null);
+
+  const childrenArray = React.Children.toArray(children) as Array<
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,316 @@
...
+import Tab, { TabProps } from './Tab';
+import TabTitle from './TabTitle';
+import omit from 'lodash/omit';
+
+const borderHeight = 3;
+
+const listStyle = css`
+  list-style: none;
+  padding: 0px;
+  margin-top: 20px;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: Removed!

packages/tabs/src/Tab.tsx
brookescarlett: 👍

packages/tabs/src/Tab.tsx
brookescarlett: Currently, It's how we're managing what Tab should be active, but happy to discuss alternatives to this approach

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/TabTitle.tsx
brookescarlett: 👍 Makes sense

packages/tabs/src/Tab.tsx
brookescarlett: Yep great catch thank you

packages/tabs/src/TabTitle.tsx
brookescarlett: yes

brookescarlett: `<Tabs as={Link} />`

packages/tabs/src/Tabs.tsx
brookescarlett: You're right

packages/tabs/src/TabTitle.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,316 @@
...
+  const [activeTab, setActiveTab] = useState();
+  const [currentIndex, setCurrentIndex] = useState();
+  const [focusedState, setFocusedState] = useState([] as Array<string>);
+  const tabListRef = useRef<HTMLDivElement>(null);
+
+  const childrenArray = React.Children.toArray(children) as Array<
+    React.ReactElement
+  >;
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
brookescarlett: When it's an inline computation or wrapped in a `useMemo` it causes too many rerenders

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,313 @@
...
+        if (child.props.default) {
+          return [...acc, index];
+        }
+
+        return acc;
+      },
+      [] as Array<number>,
+    );
+
+    setCurrentIndex(currIdx[0]);
```
jetpacmonkey: This looks like it's the only place you're setting this state. Why not just do `const currentIndex = ...`? What's the benefit to using state?

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        if (child.props.default) {
+          return [...acc, index];
+        }
+
+        return acc;
+      },
+      [] as Array<number>,
+    );
```
jetpacmonkey: This looks like it's building up an array of indexes, but you're only ever using one. Would this work the same way?

```ts
const currrentIndex = childrenArray.findIndex(child => child && (
  [activeTab, selected].includes(child.props.value) || child.props.default
));
```

packages/tabs/src/Tabs.tsx
brookescarlett: If we don't use state, currentIndex doesn't update as the 'active' tab changes, unless I'm missing something

jetpacmonkey: Sure it does, `active` changes, so the component re-renders, and the value gets calculated again. The value of `currentIndex` is completely derived from other values, so we shouldn't be storing it separately. Unless _I'm_ the one missing something 😁

brookescarlett: I'm testing it out now, and when I inline currentIndex it remains at 0 while activeTab updates

brookescarlett: Wait never mind!! You are right.. but it's just when I try and use the below that it breaks

packages/tabs/src/Tabs.tsx
brookescarlett: ^ this breaks it

jetpacmonkey: Oh my logic was off. What about...

```ts
const currentIndex = childrenArray.findIndex(child => {
  if (!child) {
    return false;
  }

  if (activeTab || selected) {
    return [activeTab, selected].includes(child.props.value);
  }
  return child.props.default;
});
```

brookescarlett: 🎆

brookescarlett: Got it :)

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,297 @@
...
+ */
+function Tabs({
+  children,
+  setSelected,
+  selected,
+  className,
+  as = 'button',
+  ...rest
+}: TabsProps) {
+  const isControlled = typeof selected === 'number';
```
hswolff: I like that we're being strict here, however it has the potentional to confuse people if they give a string number value.

It'd be nice if we had dev only warnings (i.e. they only show up when not in production mode). May be a thing to add as a TODO for the future.

However in the interest in not getting in our own way I think we should just keep this as is. We could add a console.warn if you give a string value here, but we don't have a way to turn that off in production yet.

Mostly wanted to raise awareness here, I ultimately think it's good as is.

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    return child.props.default;
+  });
+
+  function handleChange(
+    e: React.SyntheticEvent<Element, MouseEvent>,
+    index: number,
+  ) {
+    if (!isControlled && !selected) {
```
hswolff: Seems redundant to also check !selected, no? I like being thorough but this seems unnecessary.

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,297 @@
...
+          return acc;
+        }
+
+        return [...acc, index];
+      },
+      [] as Array<number>,
+    );
+
+    const enabledCurrentIndex = enabledIndexes.indexOf(
+      isControlled ? (selected as number) : activeIndex,
```
hswolff: Alternatively you can do `selected!` to tell TS that you know this value isn't undefined here. 

[Docs here](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator)

packages/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,297 @@
...
+  children,
+  setSelected,
+  selected,
+  className,
+  as = 'button',
+  ...rest
+}: TabsProps) {
+  const isControlled = typeof selected === 'number';
+  const childrenArray = React.Children.toArray(children);
+  const [activeIndex, setActiveIndex] = useState(
```
hswolff: Seems like a relic to call this activeIndex - should be selectedIndex.

Outside of that, this is to hold the controlled value right? I'm wondering if we can handle all that behavior forking up at the top of the function. Perhaps alias the selected prop to `controlledSelected` and rename this to `uncontrolledSelected`.

Then decide which one is the source of truth at the beginning of the function and assign that to `selected` so all code can reference that one value.

Same thing for the update function.

If you want to pair on this, I'm happy to as well. I hope it makes the code cleaner, but wouldn't be surprised if it doesn't as well.

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export interface TabProps {
+  /**
+   * Content that will appear as the title in the Tab list.
+   */
+  name: string | React.ReactNode;
```
hswolff: ReactNode includes string. No need to doubly include it.

packages/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Destination when name is rendered as `a` tag.
+   */
+  href?: string;
+
+  /**
+   * Destination when name is rendered as `Link` tag.
+   */
+  to?: string;
+
```
hswolff: We should rename active to selected to align with the Tabs prop.

packages/tabs/src/TabTitle.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+function usesLinkElement(
+  props: LinkTabTitleProps | ButtonTabTitleProps,
+): props is LinkTabTitleProps {
+  return props.href != null;
+}
+
+function TabTitle(props: TabTitleProps) {
+  const {
+    active,
```
hswolff: Ditto rename to selected.

packages/tabs/src/Tabs.tsx
brookescarlett: Makes sense. The reason that we are being so strict is because we can't check truthy/falsyness in terms of the presence of selected, so we have to check if it is a type of number. Agreed that we should work on adding dev warnings in the future.

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tab.tsx
brookescarlett: 👍

packages/tabs/src/TabTitle.tsx
brookescarlett: 👍

packages/tabs/src/Tab.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.spec.tsx
```diff
@@ -117,6 +116,7 @@ describe('packages/Tabs', () => {
       setTimeout(() => {
         const firstContent = getByText('First Content');
         expect(firstContent).toBeInTheDOM();
+        expect(container.querySelectorAll('[role="tabpanel"]').length).toBe(1);
```
hswolff: Are these assertions being called? How does jest know to wait for the setTimeout?

brookescarlett: You're right, removed!

PR #111, opened 2019-07-19
Use TypeScript Projects

hswolff:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


packages/checkbox/tsconfig.json
```diff
@@ -3,5 +3,16 @@
   "compilerOptions": {
     "declarationDir": "dist"
   },
-  "include": ["src/**/*", "../../typings/*"]
+  "include": ["src/**/*", "../../typings/*"],
+  "references": [
```
DesignerDave: Doing this for every package feels like a lot. 🤔

DesignerDave: Also a shame that the references array doesn't also accept path strings rather than objects.

DesignerDave: Could we have the references array only in `package.tsconfig.json`? That could be a whole lot cleaner if so.

hswolff: Yeah I agree on all counts. I'm not happy about any of it.

AFAICT the way references works is it essentially tells tsc what the dependencies are so it references them and builds them before building the current project. 

We can definitely play with it a little more but this was me getting things into a working state. We can now try optimizing it.

hswolff:
I've verified that this change doesn't make a material affect for our TypeScript type definitions.

To test I:

- Checked out master and built our `.d.ts` files
- Copied the packages folder (including the `dist` folder) into a new directory 
- Initialized a new git repo in that folder and committed all the code
- Checked out this branch and re-generated all `.d.ts` files
- Copied it over to the other folder
- Used git to diff the changes.

The only changes I can tell are of two types:

1. Different order of string unions:

image

2. Replacing `string | number` with `React.ReactText` which is a type alias for the same thing.

image


So tl;dr: this doesn't require a major breaking change and we can just use this new architecture in our next release without worry.

brookescarlett:
LGTM :)

PR #119, opened 2019-08-01
PD-214: SSO Menu

brookescarlett:
## ✍️ Proposed changes

Built SSOMenu using our base Menu component. Will need to work through some styling issues, as the border appears fainter in MMS than in Storybook (per images below). 

Wrote tests to test the component as well as tested the functionality in MMS.

🎟 _Jira ticket:_ [PD-214](https://jira.mongodb.org/browse/PD-214)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

To note: made some minor changes to the Menu component in order to support work in SSO Menu. Can extract and move to separate PR if need be!

Screen Shot 2019-08-01 at 2 04 12 PM
Screen Shot 2019-08-01 at 2 04 18 PM

Screen Shot 2019-08-01 at 11 14 34 AM


packages/menu/src/MenuItem.tsx
```diff
@@ -34,6 +33,16 @@ const containerStyle = css`
   }
 `;
 
+const linkStyle = css`
+  text-decoration: none;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,278 @@
...
+  },
+];
+
+const escapeKey = 27;
+
+interface SSOMenuProps {
+  /**
+   * Object that contains information about the active user. {name: 'string', email: 'string'}
+   */
+  userInfo: { name: string; email: string };
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,278 @@
...
+      handleClose();
+    }
+  };
+
+  useEventListener('click', handleBackdropClick);
+
+  useEventListener('keydown', handleEscape);
+
+  return (
+    <button
```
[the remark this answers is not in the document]
brookescarlett: Only because the styles were not aligned at all, but can use our component if you think that we should!!

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,278 @@
...
+      ref={triggerRef}
+      type="button"
+      onClick={() => setActive(!active)}
```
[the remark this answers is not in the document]
brookescarlett: Makes sense 👍

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+const menuItemPadding = css`
+  padding-left: 20px;
+`;
+
+const logoutContainerHeight = css`
+  height: 46px;
+`;
+
+const menuItems: Array<{
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,278 @@
...
+interface SSOMenuProps {
+  /**
+   * Object that contains information about the active user. {name: 'string', email: 'string'}
+   */
+  userInfo: { name: string; email: string };
+
+  /**
+   * MongoDB product that is currently active: ['atlas', 'university', 'support'].
+   */
+  activeProduct: string;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,278 @@
...
+        />
+      )}
+
+      <Menu active={active} align="bottom" justify="end" refEl={triggerRef}>
+        <MenuGroup className={accountMenuGroupStyle}>
+          <h3 className={cx(nameStyle, truncate)}>{name}</h3>
+          <p className={descriptionStyle}>{email}</p>
+          <Button
+            size="small"
+            onClick={onAccountClick}
```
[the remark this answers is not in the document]
brookescarlett: Updated to be an `a` tag for Account. Logout can be as well, as soon as I have an href URL to pass to the logout MenuItem!

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,250 @@
...
+  onAccountClick,
+}: SSOMenuProps) {
+  const triggerRef = useRef<HTMLButtonElement>(null);
+
+  return (
+    <Menu
+      trigger={
+        <button ref={triggerRef} type="button" className={menuButtonStyle}>
+          <span className={menuNameStyle}>{name}</span>
+          <Icon glyph="CaretDown" className={iconStyle} />
```
hswolff: Should we change this to CaretUp when it's opened?

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,250 @@
...
+        </button>
+      }
+    >
+      <MenuGroup className={accountMenuGroupStyle}>
+        <h3 className={cx(nameStyle, truncate)}>{name}</h3>
+        <p className={descriptionStyle}>{email}</p>
+        <Button
+          size="small"
+          onClick={onAccountClick}
```
hswolff: I think this will always redirect to account.mongodb.com, so we should add that `href` as well. Not sure we need the onClick handler.

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,250 @@
...
+            href={el.href}
+          >
+            <p className={menuItemTextStyle}>{el.displayName}</p>
+            <p className={descriptionStyle}>{el.description}</p>
+          </MenuItem>
+        ))}
+      </MenuGroup>
+      <MenuItem
+        active={false}
+        onClick={onLogout}
```
hswolff: We may want this to just be a URL as well.

To clarify I think the onClick event is good, I think we should also support defining a logout url.

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,250 @@
...
+      </MenuItem>
+    </Menu>
+  );
+}
+
+SSOMenu.displayName = 'SSOMenu';
+
+SSOMenu.propTypes = {
+  user: PropTypes.object,
+  activeProduct: PropTypes.oneOf(['atlas', 'support', 'university']),
```
hswolff: We should generate this from `menuItems`.

packages/sso-menu/src/SSOMenu.tsx
hswolff: I see we're doing that with the `.lg-active`.

How much more work is it to just make this a fully controlled component? Seems like a rule we're breaking that doesn't provide enough value to warrant it.

packages/sso-menu/src/SSOMenu.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+  /**
+   * MongoDB product that is currently active: ['atlas', 'university', 'support'].
+   */
+  activeProduct: ActiveProduct;
+
+  /**
+   * Callback invoked when user logs out.
+   */
+  onLogout?: React.MouseEventHandler;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/sso-menu/src/SSOMenu.tsx
brookescarlett: 👍

packages/sso-menu/src/SSOMenu.tsx
brookescarlett: Sounds good, do we know what that URL would be or not yet?

packages/sso-menu/src/SSOMenu.tsx
brookescarlett: Sounds good!

packages/sso-menu/src/SSOMenu.tsx
brookescarlett: It felt like we were rewriting logic we already wrote about handling backdrop clicks and using escape to close the menu. Can definitely make this a controlled component if need be, but this felt like a good way to make use of the code we've already written. (That's what was driving the decision here)

hswolff: I think I thought Menu behaved like Modal. Where it calls the setClosed function for these events.

Probably worthwhile to port that functionality over to Menu in a separate PR. If we use Menu as a controlled component and then add that functionality to Menu it should 'just work' the way we want.

PR #126, opened 2019-08-08
fix(icon): PD-355 Use currentcolor to set glyph fills

brookescarlett:
## ✍️ Proposed changes

Updated Icon component such that fill styling now uses currentColor, and thus CSS color rules can be used to style glyphs. 

🎟 _Jira ticket:_ [PD-355](https://jira.mongodb.org/browse/PD-355)

## 🛠 Types of changes

- [ x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## ✅ Checklist

### DIFF
--- a/.svgrrc.js
+++ b/.svgrrc.js
@@ -3,8 +3,6 @@ module.exports = {
   titleProp: true,
   svgProps: {
     role: 'img',
-    height: '{props.size}',
-    width: '{props.size}',
     viewBox: '0 0 16 16',
   },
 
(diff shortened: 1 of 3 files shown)


.svgrrc.js
```diff
@@ -22,9 +20,21 @@ module.exports = {
...
     return template.ast`
 			${imports}
-			import PropTypes from 'prop-types';
+      import PropTypes from 'prop-types';
+      import { css, cx } from '@leafygreen-ui/emotion';
 			
-			const ${componentName} = (${props}) => ${jsx};
+			const ${componentName} = (${props}) => {
+        // Setting className to manually enable fill prop to overwrite the color when the prop is set
+        props.className = cx(props.className, {[css\`color: \${ props.fill }\`]: props.fill});
```
hswolff: I'm confused....how are these props used? Like standard React Props? We're overwriting them?

What does the resulting code look like with this change?

brookescarlett: Yes these props are used like standard react props. This is how we are making sure that if a consumer passes the component a fill prop, it's being prioritized over css color rules. Then, if no fill prop is passed, we're just taking the className that they passed and determining color that way (if they have set CSS color rules). 

Not sure that I'm clear on the question that you are asking though

hswolff: So we're modifying the props being passed in?

What does the resulting code look like? 

In a typical react component you wouldn't modify props this way, is my point. 

```
function Foo(props) {
   props.blah = 'blah';
}
```

brookescarlett: So the consumer is interacting with the Icon component, which wraps the SVGComponent. We're changing the way that the SVGComponent reads the props passed to the Icon component.

PR #129, opened 2019-08-14
PD-375: Update Menu Component API

brookescarlett:
## ✍️ Proposed changes

Changing the Menu component's API to mirror the `open`, `shouldOpen` pattern in the Modal component's API. 

This change also exposed a bug in the useEventListener hook, where the placement of the `enabled` conditional wasn't allowing the hook to perform properly. Where placed before, we were not creating an event listener when enabled was set. This PR also fixes the problem.

🎟 _Jira ticket:_ [PD-375](https://jira.mongodb.org/browse/PD-375)

## 🛠 Types of changes

- [x ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## ✅ Checklist


packages/menu/src/Menu.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState, useCallback, EventHandler } from 'react';
+import React, { useState, useRef, SetStateAction } from 'react';
```
DesignerDave: We have precedent to use React types like `React.SetStateAction`, rather than using import destructuring.

packages/menu/src/Menu.tsx
```diff
@@ -35,87 +52,114 @@ interface MenuProps extends Omit<PopoverProps, 'spacing'> {
...
   trigger,
   ...rest
 }: MenuProps) {
-  const [isActive, setActiveState] = useState(false);
+  const [uncontrolledOpen, setUncontrolledOpen] = useState(false);
+
+  const popoverRef: React.RefObject<HTMLDivElement> = useRef(null);
+
+  const handleClose = () => {
+    if (setOpen && shouldClose()) {
```
DesignerDave: Maybe a bit of a nitpick, but:
```js
if (shouldClose()) {
  if (setOpen) {
    setOpen(false);
  } else {
    setUncontrolledOpen(false);
  }
}

File: packages/menu/src/Menu.tsx [hunk 3]
```diff
@@ -35,87 +52,114 @@ interface MenuProps extends Omit<PopoverProps, 'spacing'> {
...
+      setUncontrolledOpen(false);
+    }
+  };
+
+  const handleBackdropClick = (e: MouseEvent) => {
+    const popoverReference = popoverRef && popoverRef.current;
+
+    if (
+      popoverReference &&
+      !popoverReference.contains(e.target as HTMLElement)
```
DesignerDave: Do we need to do `e.target as HTMLElement` here?

packages/menu/src/Menu.tsx
```diff
@@ -35,87 +52,114 @@ interface MenuProps extends Omit<PopoverProps, 'spacing'> {
...
+
+  useEventListener('keydown', handleEscape, {
+    enabled,
+  });
 
   const popoverContent = (
     <Popover
       key="popover"
-      active={trigger ? isActive : active}
+      active={setOpen ? open : uncontrolledOpen}
```
DesignerDave: Shouldn't we be able to do `open || uncontrolledOpen` here?

packages/menu/src/Menu.tsx
```diff
@@ -35,87 +52,114 @@ interface MenuProps extends Omit<PopoverProps, 'spacing'> {
...
-    options: { once: true },
-    dependencies: [isActive, trigger],
-    enabled,
-  });
-
   if (trigger) {
     if (typeof trigger === 'function') {
       return trigger({
-        onClick: syntheticToggleEventHandler,
+        onClick: () => setUncontrolledOpen(open => !open),
```
DesignerDave: Hmm... shouldn't this use `uncontrolledOpen` instead of `open`?

packages/menu/src/Menu.tsx
```diff
@@ -35,87 +52,114 @@ interface MenuProps extends Omit<PopoverProps, 'spacing'> {
...
         children: popoverContent,
       });
     }
 
     return React.cloneElement(trigger, {
-      onClick: syntheticToggleEventHandler,
+      onClick: (e: React.MouseEvent) => {
+        setUncontrolledOpen(open => !open);
```
DesignerDave: Ditto here

packages/menu/src/Menu.tsx
```diff
@@ -35,87 +52,114 @@ interface MenuProps extends Omit<PopoverProps, 'spacing'> {
...
+
+        if (trigger.props && trigger.props.onClick) {
```
DesignerDave: Shouldn't `trigger.props` always exist?

packages/hooks/src/useEventListener.ts
```diff
@@ -38,20 +38,28 @@ export default function useEventListener<
...
 
   useEffect(() => {
-    if (!enabled) {
-      return;
-    }
-
     const callback = (e: GlobalEventHandlersEventMap[Type]) =>
       memoizedEventCallback.current(e);
 
+    const modifiedEventCallback = (e: any) => {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/menu/src/Menu.tsx

packages/menu/src/Menu.tsx

packages/menu/src/Menu.tsx
brookescarlett: Yes otherwise typescript says: `Argument of type 'EventTarget | null' is not assignable to parameter of type 'Node | null'.`

DesignerDave: Fair enough!

packages/menu/src/Menu.tsx

packages/menu/src/Menu.tsx
brookescarlett: Yep.... great catch

packages/menu/src/Menu.tsx

packages/menu/src/Menu.tsx

packages/hooks/src/useEventListener.ts
```diff
@@ -38,22 +38,19 @@ export default function useEventListener<
   }, [eventCallback]);
 
   useEffect(() => {
-    if (!enabled) {
-      return;
-    }
+    const callback = (e: GlobalEventHandlersEventMap[Type]) => {
```
hswolff: Can we add a test to cover what bug this uncovered? I still don't fully understand it but I think if we had a test that covers this situation would help me.

packages/menu/src/Menu.tsx
```diff
@@ -12,14 +12,33 @@ const rootMenuStyle = css`
   background-color: ${colors.mongodb.white};
 `;
 
-const EscapeKey = 27;
-
-interface MenuProps extends Omit<PopoverProps, 'spacing'> {
+interface MenuProps extends Omit<PopoverProps, 'spacing, active'> {
```
hswolff:
```suggestion
interface MenuProps extends Omit<PopoverProps, 'spacing' | 'active'> {
```
I had to look this up heh, https://stackoverflow.com/questions/48215950/exclude-property-from-type

packages/menu/src/Menu.tsx
```diff
@@ -35,87 +54,115 @@ interface MenuProps extends Omit<PopoverProps, 'spacing'> {
...
-  active,
+  open = false,
+  shouldClose = () => true,
+  setOpen,
   children,
   className,
   refEl,
   trigger,
   ...rest
 }: MenuProps) {
```
hswolff: We should consider this component 'controlled' if the `open` prop is set. Have it be undefined by default and then change behavior based on if the value is given.

This is similar to how the input component behaves.

hswolff: Curious to see how this change changes the code.

packages/menu/src/Menu.tsx

packages/menu/src/Menu.tsx
brookescarlett: Updated so that it does not default to false, but unless I'm missing something, it doesn't change much about the code at all

PR #137, opened 2019-09-04
PD-383: Add yarn clean script to workspace

brookescarlett:
## ✍️ Proposed changes

Added a `yarn clean` script to global package.json. This script does three things: 
* Runs a clean script that I added to local package.json files, this clean script removes the dist folder of every package (using rimraf)
* Removes the `node_modules` folder in local packages
* Removes `node_modules` folder in workspace (using rimraf)

Additionally, I removed the logic that cleaned the dist folder from the webpack config file.

🎟 _Jira ticket:_ [PD-383](https://jira.mongodb.org/browse/PD-383)

## 🛠 Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "start": "start-storybook -p 9001 -c storybook -s ./storybook/static",
     "build": "lerna run --parallel build -- --env production",
     "build:storybook": "build-storybook -c storybook -o ./storybook/public -s ./storybook/static",
+    "clean": "lerna run clean --parallel && lerna clean --yes && rimraf node_modules",
     "format": "yarn prettier && yarn eslint:format",
     "init": "yarn && yarn build && yarn ts --verbose --force",
     "link-all-packages": "lerna exec --ignore @leafygreen-ui/lib --ignore @leafygreen-ui/theme -- yarn link",
@@ -119,5 +120,8 @@
     "packages": [
       "packages/*"
     ]
+  },
+  "dependencies": {
+    "rimraf": "^3.0.0"
   }
 }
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -119,5 +120,8 @@
     "packages": [
       "packages/*"
     ]
+  },
+  "dependencies": {
+    "rimraf": "^3.0.0"
```
hswolff: devDependencies

package.json
```diff
@@ -22,6 +22,7 @@
     "start": "start-storybook -p 9001 -c storybook -s ./storybook/static",
     "build": "lerna run --parallel build -- --env production",
     "build:storybook": "build-storybook -c storybook -o ./storybook/public -s ./storybook/static",
+    "clean": "lerna run clean --parallel && lerna clean --yes && rimraf node_modules",
```
hswolff: Rather than have to modify each sub package I think having this root clean command do all cleaning would be fine.

I believe `rimraf` takes a glob pattern such that we could do : `rimraf packages/*/{dist,node_modules} node_modules` and that should clean up everything.

package.json
brookescarlett: 👍

package.json
brookescarlett: 👍

webpack.config.js
```diff
@@ -1,6 +1,5 @@
 const path = require('path');
 const webpack = require('webpack');
-const CleanWebpackPlugin = require('clean-webpack-plugin');
```
hswolff: Should remove from package.json.

PR #149, opened 2019-09-23
PD-392: Make RadioGroup accessible

brookescarlett:
## ✍️ Proposed changes

Added group role to RadioGroup container div and supplied aria-label to ensure component is accessible

🎟 _Jira ticket:_ [PD-392](https://jira.mongodb.org/browse/PD-392)

## 🛠 Types of changes

### DIFF
--- a/packages/radio-group/src/RadioGroup.tsx
+++ b/packages/radio-group/src/RadioGroup.tsx
@@ -164,7 +164,7 @@ export default class RadioGroup extends PureComponent<
           className,
         )}
         role="group"
-        aria-label={this.defaultName}
+        aria-label={name}
       >
         {renderChildren}
       </div>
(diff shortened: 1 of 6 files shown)


packages/radio-group/src/RadioGroup.tsx
```diff
@@ -163,6 +163,8 @@ export default class RadioGroup extends PureComponent<
           `,
           className,
         )}
+        role="group"
+        aria-label={this.defaultName}
```
DesignerDave: Shouldn't this be `name`, not `this.defaultName` here? We set the default for `name` to `this.defaultName`.

hswolff:
I think it's non-breaking as its additive.

brookescarlett:
@hswolff The reason this could be a breaking change is because if consumers were previously setting the `role` or `aria-label` attributes, we are now overwriting what a consumer was expecting as the value there.

PR #151, opened 2019-09-23
PD-380: Introduce 'title' and 'desc' props to Icon component

gssbzn:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Introduce 'title' and 'desc' props to Icon component
](https://jira.mongodb.org/browse/PD-380)

## 🛠 Types of changes

## ✅ Checklist


packages/icon/src/createIconComponent.tsx
```diff
@@ -27,24 +27,23 @@ const sizeMap: { [S in Size]: number } = {
...
-    const humanReadableTitle = `${(glyph as string).replace(
-      /([A-Z][a-z])/g,
-      ' $1',
-    )} Icon`;
-
     return (
       <SVGComponent
         {...rest}
-        title={humanReadableTitle}
+        title={rest.title ? rest.title : humanReadableTitle(glyph as string)}
```
brookescarlett: This could be `rest.title || humanReadableTitle(glyph as string)`

gssbzn: Can it? I went this way because I remember `className` throwing warning about this format and they rather go for a ternary but I'm not sure about svg

gssbzn:
@brookescarlett sorry for the half cooked PR was late for me and just wanted to push this and discuss tests as I could not figure out the tests for `Icon`
About the `desc` it looks like SVGR doesn't support `desc` out of the box and we may need to define a custom template which may be too much for our use case

brookescarlett:
@gssbzn I see what you're saying about SVGR not supporting `desc` out of the box. I think for now incorporating `title` is a great step toward making the component more accessible, and perhaps can redefine with a custom template when that's more in scope.

.github/pull_request_template.md
```diff
@@ -31,6 +31,7 @@ Put an `x` in the boxes that apply. You can also fill these out after creating t
 
 - [ ] I have added tests that prove my fix is effective or that my feature works
 - [ ] I have added necessary documentation (if appropriate)
+- [ ] I have run `yarn changeset` and documented my changes
```
brookescarlett: Thank you for adding this!

packages/icon/src/createIconComponent.tsx
brookescarlett: I made the change locally and am not seeing any warnings on my end, but let me know if you are seeing otherwise. Open to other format if that's the case.

gssbzn:
@brookescarlett RFAL

PR #152, opened 2019-09-23
PD-295: Make Menu accessible

brookescarlett:
## ✍️ Proposed changes

Removed role='menu' from MenuGroup component, based on accessibility guidelines. And removed associated test (which was testing that MenuGroup had that role).

🎟 _Jira ticket:_ [PD-295](https://jira.mongodb.org/browse/PD-295)


hswolff:
I could argue this for major or for patch.

For all intents and purposes I think having this be patch is fine. At the worst it should only affect tests, and if you're relying on internal structure we have no contract on that structure.

PR #162, opened 2019-09-26
PD-293: Adding a11y linter to leafyGreen

brookescarlett:
## ✍️ Proposed changes

Bringing a11y linter into leafyGreen

To note: there are seven errors raised by this linter, but will address those in separate PRs and have disabled the lint rules in the interim. Doing this because I believe that there will be different levels of version-bumping the effected components, so did not want to batch all of these changes.

🎟 _Jira ticket:_ [PD-293](https://jira.mongodb.org/browse/PD-293)

## 🛠 Types of changes

## ✅ Checklist

~[ ] I have added tests that prove my fix is effective or that my feature works~
~[ ] I have added necessary documentation (if appropriate)~

### DIFF
--- a/packages/checkbox/src/Checkbox.tsx
+++ b/packages/checkbox/src/Checkbox.tsx
@@ -285,6 +285,7 @@ export default class Checkbox extends PureComponent<
         style={style}
         htmlFor={checkboxId}
       >
+        {/* eslint-disable-next-line jsx-a11y/no-redundant-roles */}
         <input
           {...rest}
           id={checkboxId}
--- a/packages/checkbox/src/Checkbox.tsx
+++ b/packages/checkbox/src/Checkbox.tsx
@@ -285,14 +285,12 @@ export default class Checkbox extends PureComponent<
         style={style}
         htmlFor={checkboxId}
       >
-        {/* eslint-disable-next-line jsx-a11y/no-redundant-roles */}
         <input
           {...rest}
           id={checkboxId}
           ref={this.inputRef}
           className={inputStyle}
           type="checkbox"
-          role="checkbox"
           name={name}
           disabled={disabled}
           checked={checked}
--- a/packages/modal/src/Modal.tsx
+++ b/packages/modal/src/Modal.tsx
@@ -200,9 +200,9 @@ function Modal({
     <Transition in={open} timeout={500} mountOnEnter unmountOnExit>
       {(state: string) => (
         <Portal>
-          {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,  jsx-a11y/no-static-element-interactions */}
           <div
             {...rest}
+            role="none"
             onClick={handleBackdropClick}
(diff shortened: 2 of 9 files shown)


packages/modal/src/Modal.tsx
```diff
@@ -209,6 +210,7 @@ function Modal({
           >
             <div className={scrollContainer} ref={scrollContainerRef}>
               <div
+                // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
```
DesignerDave: Let's leave a "why" comment above this.

packages/modal/src/Modal.tsx
```diff
@@ -200,6 +200,7 @@ function Modal({
     <Transition in={open} timeout={500} mountOnEnter unmountOnExit>
       {(state: string) => (
         <Portal>
+          {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,  jsx-a11y/no-static-element-interactions */}
```
DesignerDave: Can you explain what this is for / leave a comment in the file? I think as a general standard, any time we disable a linting rule, especially accessibility-related ones, we should have a comment explaining why.

packages/checkbox/src/Checkbox.tsx
```diff
@@ -285,6 +285,7 @@ export default class Checkbox extends PureComponent<
         style={style}
         htmlFor={checkboxId}
       >
+        {/* eslint-disable-next-line jsx-a11y/no-redundant-roles */}
```
DesignerDave: See my comment below.

packages/modal/src/Modal.tsx
brookescarlett: Was only disabling for now, and was planning on making separate PRs to each component to make them comply with new linter rules. Just wanted to break it up, so it wasn't one large PR. Let me know if you'd prefer if I just make the updates right now, and can just include multiple changesets in this PR.

packages/modal/src/Modal.tsx
```diff
@@ -202,14 +202,17 @@ function Modal({
         <Portal>
           <div
             {...rest}
+            role="none"
```
DesignerDave: What's the difference between this, and not setting `role` at all?

packages/modal/src/Modal.tsx
```diff
@@ -226,7 +229,7 @@ function Modal({
                   onClick={handleClose}
                   className={closeButton}
                   data-dismiss="modal"
-                  aria-hidden="true"
+                  tabIndex={0}
```
DesignerDave: We seem to be setting tabIndex twice?

brookescarlett: -1 on the container, 0 on the "x" close button

packages/modal/src/Modal.tsx
brookescarlett: Added a comment!

PR #166, opened 2019-10-04
PD-90: Implement Tooltip Component

brookescarlett:
## ✍️ Proposed changes

Initial implementation of Tooltip component. This component abides by [WAI-Aria Tooltip Practices](https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip).

A couple of notes to call out:
- Made some edits to the Popover component, to expose the final alignment/justification of the element relative to the trigger. We did this because the final positions were needed to create the notch/triangle that appears in the Tooltip.
- I'm going to create useBackdropClick and useEscape hooks while this PR is in review, because we use the exact same logic in 3+ components.

Screen Shot 2019-10-04 at 4 13 04 PM
Screen Shot 2019-10-04 at 4 13 13 PM

🎟 _Jira ticket:_ [PD-90](https://jira.mongodb.org/browse/PD-90)

## 🛠 Types of changes

## ✅ Checklist

* I will run `yarn changeset` once the component is more finalized


packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+  const open = isControlled ? controlledOpen : uncontrolledOpen;
+  const setOpen = isControlled ? controlledSetOpen : uncontrolledSetOpen;
+
+  const triggerRef = useRef<HTMLElement>(null);
+  const tooltipRef = useRef<HTMLDivElement>(null);
+
+  const tooltipId =
+    id || `tooltip-${Math.floor(Math.random() * Math.floor(10))}`;
+
+  const triggerEventMap = (triggerType: TriggerEvent, triggerProps?: any) => {
```
DesignerDave: I wouldn't call this an event map, given it's a function. What about `mapToTriggerEvent`?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+    if (triggerType === 'hover') {
+      return {
+        onMouseEnter: debounce(() => {
```
DesignerDave: Why are we debouncing these?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+      if (triggerProps.onClick) {
+        return {
+          onClick: () => {
+            triggerProps.onClick();
+            setOpen && setOpen(!open);
+          },
+        };
+      } else {
+        return {
+          onClick: () => setOpen && setOpen(!open),
```
DesignerDave: Is there a case where `setOpen` wouldn't be defined?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+        };
+      }
+    }
+  };
+
+  const handleClose = () => {
+    if (!shouldClose || shouldClose()) {
+      setOpen && setOpen(false);
```
DesignerDave: Ditto here – will setOpen ever be undefined?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+    if (
+      triggerEvent === 'click' &&
+      tooltipReference &&
+      !tooltipReference.contains(e.target as HTMLElement)
+    ) {
+      handleClose();
+    }
+  };
+
+  const enabled = open;
```
DesignerDave: Why not just use `open`?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+    <Popover
+      key="popover"
+      active={open}
+      align={align}
+      justify={justify}
+      usePortal={true}
+      adjustOnMutation={true}
+      spacing={12}
+      setAlignment={setAlignment}
+      setJustification={setJustification as any}
```
DesignerDave: Why `as any`?

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const borderOffset = 'calc(100% - 1px)';
+
+  const containerSize = 15;
+  const notchSize = 8;
+  const containerOffsetX = (triggerRect.width - containerSize) / 2;
+  const containerOffsetY = (triggerRect.height - containerSize) / 2;
+  const notchOverlap = -notchSize / 2;
+  const containerStyleObj: any = {};
+  const notchStyleObj: any = {};
```
DesignerDave: We should type these more accurately: 

Something like this I think:
```
const notchStyleObj: {K: string} = {};
```

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+      switch (justification) {
+        case 'left':
+          containerStyleObj.left = `${containerOffsetX}px`;
+          break;
+
+        case 'center-horizontal':
+          containerStyleObj.left = '0px';
+          containerStyleObj.right = '0px';
+          containerStyleObj.margin = 'auto';
```
DesignerDave: We should be able to set this style more outside a conditional for this element.

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const notchStyleObj: any = {};
+
+  switch (alignment) {
+    case 'top':
+    case 'bottom':
+      notchStyleObj.left = 0;
+      notchStyleObj.right = 0;
+
+      if (alignment === 'top') {
+        containerStyleObj.top = `${borderOffset}`;
```
DesignerDave: ``${borderOffset}px``?

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,105 @@
...
+        notchStyleObj.top = `${notchOverlap}px`;
+      } else {
+        containerStyleObj.bottom = `${borderOffset}`;
```
DesignerDave: Ditto

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    case 'left':
+    case 'right':
+      notchStyleObj.top = 0;
+      notchStyleObj.bottom = 0;
+
+      if (alignment === 'left') {
+        notchStyleObj.left = `${notchOverlap}px`;
+
+        // right align is .8px off, factoring that in here to correct offset
```
DesignerDave: Hmm... There has to be a better way. That discrepancy is likely to change based on other variables, so this is brittle.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+    >
+      <div
+        {...rest}
+        role="tooltip"
+        id={tooltipId}
```
DesignerDave: What's this being used for?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+}: TooltipProps) {
+  const isControlled = !!controlledSetOpen;
+  const [uncontrolledOpen, uncontrolledSetOpen] = useState(false);
+  const open = isControlled ? controlledOpen : uncontrolledOpen;
+  const setOpen = isControlled ? controlledSetOpen : uncontrolledSetOpen;
+
+  const triggerRef = useRef<HTMLElement>(null);
+  const tooltipRef = useRef<HTMLDivElement>(null);
+
+  const tooltipId =
```
DesignerDave: This ID will change every render. This needs to be set once on render, and not change otherwise.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+  Light: 'light',
+  Dark: 'dark',
+} as const;
+
+type Variant = typeof Variant[keyof typeof Variant];
+
+const baseStyles = css`
+  font-size: 14px;
+  line-height: 20px;
+  padding: 14px 16px 15px 16px;
```
DesignerDave: This... seems wrong. Probably should be `14px 16px` or just `16px`.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+import PropTypes from 'prop-types';
+import Popover, { PopoverProps, Align, Justify } from '@leafygreen-ui/popover';
+import { useEventListener } from '@leafygreen-ui/hooks';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+import { transparentize } from 'polished';
+import debounce from 'lodash/debounce';
+import { trianglePosition } from './tooltipUtils';
+
+export const TriggerEvent = {
```
DesignerDave: Might this be a bit overkill? JS relies on the plain strings (e.g. `'click'`) when creating event listeners.

packages/tooltip/src/Tooltip.story.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Align, Justify } from '@leafygreen-ui/popover';
+import Tooltip, { TriggerEvent, Variant } from '.';
+
+function ControlledTooltip() {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <Tooltip
+      open={open}
+      setOpen={setOpen}
```
DesignerDave: I might be missing something, but the button isn't actually firing the `setOpen` event.

packages/tooltip/src/Tooltip.tsx
brookescarlett: So that if a user quickly hovers over a tooltip trigger on accident, they aren't seeing the tooltip. Making sure the hover is intentional

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: I know Typescript was yelling at me though. I thought it was because we had defined setOpen in the interface as optional, so it was questioning me. But when I renamed setOpen I encountered the same issue. Could i use `setOpen!(!open)` to clean this up?

packages/tooltip/src/Tooltip.tsx
brookescarlett: See above

packages/tooltip/src/Tooltip.tsx
brookescarlett: Idk I like the syntax better in the object, but that's not a good enough reason I suppose.

packages/tooltip/src/Tooltip.tsx
brookescarlett: Was erroring because of differences in justification and justify.

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: Not sure what you mean sorry

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: calc~

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: calc~!

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: Agreed, just not sure what that .8 is from yet and this was my best attempt for now.

packages/tooltip/src/Tooltip.tsx
brookescarlett: Accessibility. Matches `aria-describedby`

packages/tooltip/src/Tooltip.tsx
brookescarlett: Based on the invision spec! Lmk which direction to take and will apply changes

packages/tooltip/src/Tooltip.tsx
brookescarlett: Perhaps 🤷‍♀

brookescarlett: It is clean though, and I use in multiple places.

packages/tooltip/src/Tooltip.story.tsx
brookescarlett: yeah the button is, but pass it to Tooltip so we can handle closing on escape and backdrop clicks :)

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: fixed!

packages/popover/src/Popover.tsx
```diff
@@ -93,6 +93,14 @@ export interface PopoverProps {
    * default: false
    */
   adjustOnMutation?: boolean;
+
+  setAlignment?: (
+    alignment: Align,
+  ) => void | React.Dispatch<React.SetStateAction<string>>;
```
DesignerDave: Can we use `Align` instead of `string` as the type parameter to `React.SetStateAction`?

packages/popover/src/positionUtils.ts
```diff
@@ -40,7 +40,11 @@ export function calculatePosition({
   contentElPos = defaultElementPosition,
   windowHeight = window.innerHeight,
   windowWidth = window.innerWidth,
-}: CalculatePosition) {
+}: CalculatePosition): {
+  alignment: Align;
+  justification: string;
```
DesignerDave: I feel like we should have an enum for `Justification`. I know why this is currently string, but it might be worth creating one.

packages/tooltip/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+If controlling the component, pass state handling function to setOpen prop. This will keep the consuming application's state in-sync with LeafyGreen's state, while the `Tooltip` component responds to events such as backdrop clicks and a user pressing the Escape key.
+
+### shouldClose
+
+**Type:** `function`
+
+**Default:** `() => true`
+
+Callback to determine whether or not `Tooltip` should close when user tries to close it.
```
DesignerDave: `Callback that should return a boolean that determines...`

packages/tooltip/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+### align
+
+**Type:** `top`, `bottom`, `left`, `right`
+
+**Default:** `top`
+
+Determines the alignment of the `Tooltip` component relative to the element passed to the `trigger` prop.
```
DesignerDave: Should include that it will be relative to the parent if no trigger is passed.

packages/tooltip/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+**Default:** `start`
+
+Determines the justification of the `Tooltip` component (based on the alignment) relative tto the element passed to the `trigger` prop.
+
+### trigger
+
+**Type:** `HTMLElement` or `ReactNode`
+
+An `HTMLElement` or `ReactNode` against which the `Tooltip` will be positioned, and what will be used to trigger the opening and closing of the `Tooltip` component.
```
DesignerDave: *if the Tooltip is uncontrolled.

packages/tooltip/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+### triggerEvent
+
+**Type:** `hover` or `click`
+
+**Default:** `hover`
+
+Event that triggers opening/closing of `Tooltip` component
```
DesignerDave: I'd say "DOM Event" rather than just "Event" here.

packages/tooltip/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+**Default:** `light`
+
+Determines color of `Tooltip` component
+
+### id
+
+**Type:** `string`
+
+id applied to `Tooltip` component
```
DesignerDave: Either "ID" or `id` in code tags so that this is easier to read correctly.

packages/tooltip/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+Event that triggers opening/closing of `Tooltip` component
+
+### variant
+
+**Type:** `light` or `dark`
+
+**Default:** `light`
+
+Determines color of `Tooltip` component
```
DesignerDave: "Determines the color variant of the `Tooltip` component."

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+  const triggerRef = useRef<HTMLElement>(null);
+  const tooltipRef = useRef<HTMLDivElement>(null);
+
+  const tooltipId = useMemo(
+    () => id || `tooltip-${Math.floor(Math.random() * Math.floor(10))}`,
+    [id],
+  );
+
+  const mapToTriggerEvent = (triggerType: TriggerEvent, triggerProps?: any) => {
+    if (triggerType === 'hover') {
```
DesignerDave: Why not use the `TriggerEvent` enum here?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+      return {
+        onMouseEnter: debounce(() => {
+          setOpen && setOpen(!open);
+        }, 35),
+        onMouseLeave: debounce(() => {
```
DesignerDave: Nit:
```
onMouseLeave: debounce(handleClose, 35),
```

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+      };
+    } else {
+      if (triggerProps.onClick) {
+        return {
+          onClick: () => {
+            triggerProps.onClick();
+            setOpen && setOpen(!open);
+          },
+        };
+      } else {
```
DesignerDave: Happy path here please

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+        onMouseEnter: debounce(() => {
+          setOpen && setOpen(!open);
+        }, 35),
+        onMouseLeave: debounce(() => {
+          handleClose();
+        }, 35),
+        onFocus: () => setOpen && setOpen(true),
+        onBlur: handleClose,
+      };
+    } else {
```
DesignerDave: Happy path please

packages/tooltip/src/Tooltip.tsx
DesignerDave: Hmm... there's something fishy here.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+  };
+
+  const handleClose = () => {
+    if (!shouldClose || shouldClose()) {
+      setOpen && setOpen(false);
+    }
+  };
+
+  const handleEscape = (e: KeyboardEvent) => {
+    e.stopImmediatePropagation();
```
DesignerDave: Do we need to stop propagation if we don't call `handleClose`?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+  // And to make sure we're responding to not just the desired alignment, but the alignment
+  // As a result of our calcPosition() function
+  const [alignment, setAlignment] = useState(align);
+  const [justification, setJustification] = useState(justify);
+
+  const triangleStyle = trianglePosition(
+    alignment,
+    justification,
+    triggerRect,
+  ) as any;
```
DesignerDave: Why `as any`?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+  const tooltip = (
+    <Popover
+      key="popover"
```
DesignerDave: Why do we need a key?

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,116 @@
...
+    right?: string;
+    top?: string;
+    bottom?: string;
+    margin?: string;
+  } = {};
+  const notchStyleObj: {
+    left?: string | number;
+    right?: string | number;
+    top?: string | number;
+    bottom?: string | number;
```
DesignerDave: Why not just always return a string?

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const containerOffsetX = (triggerRect.width - containerSize) / 2;
+  const containerOffsetY = (triggerRect.height - containerSize) / 2;
+  const notchOverlap = -notchSize / 2;
+  const containerStyleObj: {
+    left?: string;
+    right?: string;
+    top?: string;
+    bottom?: string;
+    margin?: string;
+  } = {};
```
DesignerDave: There's gotta be a better way to type this...

packages/tooltip/src/tooltipUtils.tsx
DesignerDave: We can probably unconditionally set margin to `auto`.

packages/tooltip/src/tooltipUtils.tsx
DesignerDave: If you change a variable like notchSize, does this discrepancy change?

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/popover/src/Popover.tsx
brookescarlett: 👍

packages/tooltip/README.md
brookescarlett: 👍

packages/tooltip/README.md
brookescarlett: 👍

packages/tooltip/README.md
brookescarlett: 👍

packages/tooltip/README.md
brookescarlett: 👍

packages/tooltip/README.md
brookescarlett: Yep makes sense

packages/tooltip/README.md
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: smart

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
DesignerDave: Occasionally for various reasons, those can be slightly inaccurate when indicating the exact value (e.g. when centering an object, sometimes it will snap to the nearest pixel value rather than use sub-pixels). Let's run with `14px 16px` as it looks like that's the intent.

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: Will be handled by hook -- can fix in other PR

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: nope...

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: 👍

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: 👍

packages/popover/src/positionUtils.ts
brookescarlett: Done

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,291 @@
...
+    box-shadow: 0px 2px 4px ${transparentize(0.85, uiColors.black)};
+  `,
+};
+
+interface TooltipProps
+  extends Omit<PopoverProps, 'active' | 'spacing' | 'refEl' | 'usePortal'> {
+  /**
+   * A slot for the element used to trigger the `Tooltip`.
+   */
+  trigger: React.ReactElement | Function;
```
hswolff: We should document what args the function is called with.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,291 @@
...
+    </Popover>
+  );
+
+  const sharedTriggerProps = {
+    ref: triggerRef,
+    'aria-describedby': tooltipId,
+  };
+
+  if (trigger) {
+    if (typeof trigger === 'function') {
```
hswolff: So....if you do `typeof class Foo {}` it returns function. Not sure if we should even worry about it.

The reason React long ago had you extend React.Component was so it could [add a flag on the prototype](https://github.com/facebook/react/blob/master/packages/react/src/ReactBaseClasses.js#L31) so it could sniff out whether it was a class for just this reason.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,291 @@
...
+  trigger,
+  variant = Variant.Light,
+  triggerEvent = TriggerEvent.Hover,
+  align = 'top',
+  justify = 'start',
+  id,
+  shouldClose,
+  ...rest
+}: TooltipProps) {
+  const isControlled = !!controlledSetOpen;
```
hswolff: Not a fan of the double negative. Hurts my brain.

Can we do `typeof controlledSetOpen === 'function'` to say what we actually want?

hswolff: Also I think control is handled by whether `open` is defined, i.e. `isControlled = typeof open === 'boolean'`. 

If you set a setOpen function and then don't set open then it's controlled but nothing is defining what it's controlled behavior is.

If you set open and omit setOpen then the value is set but then just is never updated.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+  const triggerRef = useRef<HTMLElement>(null);
+  const tooltipRef = useRef<HTMLDivElement>(null);
+
+  const tooltipId = useMemo(
+    () => id || `tooltip-${Math.floor(Math.random() * Math.floor(10))}`,
+    [id],
+  );
+
+  const mapToTriggerEvent = (triggerType: TriggerEvent, triggerProps?: any) => {
```
hswolff: Seems unnecessary to pass in the triggerEvent. We can just use the outer functions value right? As it's the same.

hswolff: Function name is a little confusing to me as well. This is creating trigger props right?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,291 @@
...
+        onFocus: () => setOpen(true),
+        onBlur: handleClose,
+      };
+    }
+
+    if (triggerProps.onClick) {
+      return {
+        onClick: () => {
+          triggerProps.onClick();
+          setOpen(!open);
```
hswolff: Feels like this should go first so we clear up our state first. Otherwise the triggerProps.onClick could perhaps remove the Tooltip from the DOM and then we'd get an error when we try to call a state update change on this component.

hswolff: Also whenever toggling a boolean value with a setState function it's always best to use the function form to ensure you always get the current value and flip its value.

```jsx
setOpen(current => !current);
```

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,291 @@
...
+      };
+    }
+
+    return {
+      onClick: () => setOpen(!open),
+    };
+  };
+
+  const handleClose = () => {
+    if (!shouldClose || shouldClose()) {
```
hswolff: Why `!shouldClose`? Are we just checking if it exists and is a function? May be better to do that.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,291 @@
...
+      return {
+        onClick: () => {
+          triggerProps.onClick();
+          setOpen(!open);
+        },
+      };
+    }
+
+    return {
+      onClick: () => setOpen(!open),
```
hswolff: So here's a fun bug....

If I set the trigger to 'click' and then open the tooltip, if I try and click on anything inside the tooltip content then this function is triggered and the tooltip closes.

Makes it impossible to interact with content inside the tooltip, which I imagine we want to provide.

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,106 @@
...
+      break;
+  }
+
+  return {
+    containerStyle: css`
+      position: absolute;
+      width: ${containerSize}px;
+      height: ${containerSize}px;
+      overflow: hidden;
+      containerStyleObj.margin = 'auto';
```
hswolff: :eyes:

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,291 @@
...
+  };
+
+  useEventListener('click', handleBackdropClick, { enabled: open });
+
+  const triggerRect =
+    triggerRef &&
+    triggerRef.current &&
+    triggerRef.current.getBoundingClientRect();
+
+  // We are doing this to get the final alignment and justification from Popover
```
hswolff: I'm wondering if the better way to do this is by having Popover accept a function as a child and we pass in final values into the function.

Thoughts? cc @DesignerDave

packages/tooltip/src/Tooltip.spec.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const buttonText = 'trigger button';
+
+  describe('when uncontrolled', () => {
+    const triggerEvent = 'click';
+
+    const { getByText, getByTestId } = render(
+      <>
+        <div>backdrop content</div>
+        <Tooltip
+          trigger={<button>{buttonText}</button>}
```
hswolff: Can we get a test for when the trigger is a function.

And maybe one where it's a class? Not sure if we want to worry about this right now.

packages/tooltip/src/Tooltip.spec.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const button = getByText(buttonText);
+    const tooltip = getByTestId('uncontrolled-tooltip');
+    const backdrop = getByText('backdrop content');
+
+    test(`renders a button to the DOM with ${buttonText}`, () => {
+      expect(button).toBeInTheDocument();
+      expect(button.tagName.toLowerCase()).toBe('button');
+    });
+
+    test(`${triggerEvent} triggers opening and closing of tooltip`, () => {
```
hswolff: We need a test that clicking on the content inside of the tooltip content does not cause the tooltip to hide.

packages/tooltip/src/Tooltip.tsx
DesignerDave: Huh. I really like that idea actually, and I don't think it would take all that much time! We're definitely gonna run up against this again...

Did you envision the function's signature being `(alignment, justification) => {}` or `({alignment, justification}) => {}`? The latter seems less clean, but more extensible.

hswolff: Generally I prefer the latter. Extensible and self documenting.

packages/tooltip/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
```
hswolff: We need to update the root `tsconfig.json` to include this package too.

packages/tooltip/src/Tooltip.tsx
DesignerDave: Fair enough! @brookescarlett How does this sound?

```js
<Popover>
  {({alignment, justification}) => (
    <>
      <div>My alignment is {alignment}</div>
      <div>My justification is {justification}</div>
    </>
  )}
</Popover>
```

packages/popover/src/Justification.ts
```diff
@@ -0,0 +1,14 @@
+// We transform 'middle' into 'center-vertical' or 'center-horizontal' for internal use,
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/popover/src/Justification.ts
```diff
@@ -0,0 +1,14 @@
...
+const Justification = {
+  Top: 'top',
+  Bottom: 'bottom',
+  Left: 'left',
+  Right: 'right',
+  CenterVertical: 'center-vertical',
+  CenterHorizontal: 'center-horizontal',
+} as const;
+
+type Justification = typeof Justification[keyof typeof Justification];
```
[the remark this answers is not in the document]
brookescarlett: Followed pattern on Align.ts and Justify.ts

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: Changed the name to `createTriggerProps` but confused about not passing triggerEvent. That's needed to determine what event handlers we attach

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: thank you 👍

packages/tooltip/tsconfig.json
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: It works 🔥This is sick

packages/tooltip/src/Tooltip.tsx
brookescarlett: Fixed, good call out 👍

packages/tooltip/src/Tooltip.spec.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: You're right that using a class component breaks things. I'm trying to check if trigger is an instanceof React.Component, but that's not working for me either. Having trouble figuring out how to access that flag ^

packages/tooltip/src/Tooltip.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: Fixed this ish, but when I'm testing now if I pass {Button} to the trigger prop, React complains: `Cannot call a class as a function`

brookescarlett: (So could be a good thing, just natively not supported, no one will try and use this way... Also supplied better direction/example in docs now re proper usage)

packages/tooltip/src/Tooltip.spec.tsx
brookescarlett: 👍

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const notchSize = 8;
+  const containerOffsetX = (triggerRect.width - containerSize) / 2;
+  const containerOffsetY = (triggerRect.height - containerSize) / 2;
+  const notchOverlap = -notchSize / 2;
+
+  type Styles = 'left' | 'right' | 'top' | 'bottom' | 'margin';
+  const notchStyleObj: { [K in Styles]?: string } = {};
+  const containerStyleObj: { [K in Styles]?: string } = {};
+
+  switch (alignment) {
```
DesignerDave: Might just be a preference of mine, but rather than nesting switch statements, I'd prefer we use a good ol' conditional for the outer switch here.

packages/popover/src/Popover.tsx
```diff
@@ -35,11 +34,17 @@ const mutationOptions = {
   subtree: true,
 };
 
+interface FunctionParameters {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+**Type:** `boolean`
+
+**Default:** `false`
+
+Controls the component, and determines whether or not the `Tooltip` will appear open or closed.
+
+### setOpen
+
+**Type:** `function`
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+Callback that should return a boolean that determines whether or not the `Tooltip` should close when a user tries to close it.
+
+### align
+
+**Type:** `top`, `bottom`, `left`, `right`
+
+**Default:** `top`
+
+Determines the alignment of the `Tooltip` component relative to the element passed to the `trigger` prop. If no `trigger` is passed, the Tooltip will be positioned against its nearest parent element.
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+**Default:** `start`
+
+Determines the justification of the `Tooltip` component (based on the alignment) relative to the element passed to the `trigger` prop. If no `trigger` is passed, the Tooltip will be positioned against its nearest parent element.
+
+### trigger
+
+**Type:** `HTMLElement` or `ReactNode`
+
+**Signature:** ({children, ...rest}) => (<button {...rest}>trigger {children}</button>)
```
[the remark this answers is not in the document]
brookescarlett: If passing a function we need to be explicit about how to do it, but can find a better way to word this.

packages/tooltip/src/Tooltip.spec.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+interface ButtonTestProps {
+  [key: string]: any;
+}
+
+describe('packages/Tooltip', () => {
+  const buttonText = 'trigger button';
+
+  describe('when uncontrolled', () => {
+    const triggerEvent = 'click';
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/Tooltip.spec.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <Tooltip
+          trigger={<button onClick={onClick}>{buttonText}</button>}
+          triggerEvent={triggerEvent}
+        >
+          <div data-testid="uncontrolled-tooltip">Uncontrolled tooltip!</div>
+        </Tooltip>
+      </>,
+    );
+
+    const button = getByText(buttonText);
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/Tooltip.spec.tsx
```diff
@@ -0,0 +1,182 @@
...
+      expect(button.tagName.toLowerCase()).toBe('button');
+    });
+
+    test(`${triggerEvent} triggers opening and closing of tooltip`, () => {
+      fireEvent.click(button);
+      expect(tooltip).toBeVisible();
+      expect(onClick).toHaveBeenCalledTimes(1);
+
+      fireEvent.click(button);
+      expect(tooltip).not.toBeVisible();
```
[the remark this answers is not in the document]
brookescarlett: Was more so just to test that the eventHandler fires, once proven was less concerned about the number of times its called. Can update if you like though.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+  /**
+   * Callback to change the open state of the `Tooltip`.
+   */
+  setOpen?: (
+    open: boolean,
+  ) => void | React.Dispatch<React.SetStateAction<boolean>>;
+
+  /**
+   * Whether the `Tooltip` will be `light` or `dark`.
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+  const tooltipRef = useRef<HTMLDivElement>(null);
+
+  const tooltipId = useMemo(
+    () => id || `tooltip-${Math.floor(Math.random() * Math.floor(10))}`,
+    [id],
+  );
+
+  const createTriggerProps = (
+    triggerType: TriggerEvent,
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,300 @@
...
+        onMouseLeave: debounce(handleClose, 35),
+        onFocus: () => setOpen(true),
+        onBlur: handleClose,
+      };
+    }
+
+    if (triggerProps && triggerProps.onClick) {
+      return {
+        onClick: (e: MouseEvent) => {
+          if (e.target !== tooltipRef.current) {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,300 @@
...
+      justify={justify}
+      usePortal={true}
+      adjustOnMutation={true}
+      spacing={12}
+    >
+      {({
+        alignment,
+        justification,
+        referenceElPos,
+      }: {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+  useHandleEscape(handleClose);
+
+  const handleBackdropClick = (e: MouseEvent) => {
+    const tooltipReference = tooltipRef && tooltipRef.current;
+
+    if (
+      triggerEvent === 'click' &&
+      tooltipReference &&
+      !tooltipReference.contains(e.target as HTMLElement)
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/tooltipUtils.tsx
brookescarlett: Talked offline, leaving as-is.

packages/popover/src/Popover.tsx
```diff
@@ -15,7 +9,12 @@ import {
 } from '@leafygreen-ui/hooks';
 import Align from './Align';
 import Justify from './Justify';
-import { calculatePosition, getElementPosition } from './positionUtils';
+import Justification from './Justification';
+import {
+  calculatePosition,
+  getElementPosition,
+  ElementPosition,
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/popover/src/Popover.tsx
```diff
@@ -15,7 +9,12 @@ import {
 } from '@leafygreen-ui/hooks';
 import Align from './Align';
 import Justify from './Justify';
-import { calculatePosition, getElementPosition } from './positionUtils';
+import Justification from './Justification';
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,307 @@
...
+        // ensure that we don't close the tooltip when content inside tooltip is clicked
+        if (e.target !== tooltipRef.current) {
+          setOpen((curr: boolean) => !curr);
+        }
+      },
+    };
+  };
+
+  const handleClose = () => {
+    if (typeof shouldClose !== 'function' || shouldClose()) {
```
hswolff: This logic is wrong, right?

```suggestion
    if (typeof shouldClose === 'function' && shouldClose()) {
```

Before we change this code can we write a test that covers this case?

brookescarlett: We're checking that shouldClose either doesn't exist so we can go ahead and close the tooltip automatically, or that shouldClose() returns true and we can still proceed with closing the tooltip. Happy to write a test though :)

packages/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,307 @@
...
+  id,
+  shouldClose,
+  ...rest
+}: TooltipProps) {
+  const isControlled = typeof controlledOpen === 'boolean';
+  const [uncontrolledOpen, uncontrolledSetOpen] = useState(false);
+  const open = isControlled ? controlledOpen : uncontrolledOpen;
+  // typescript is not recognizing isControlled checks that controlledSetOpen exists
+  const setOpen = isControlled
+    ? (controlledSetOpen as Function)
```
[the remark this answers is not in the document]
brookescarlett: WOAH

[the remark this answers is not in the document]
brookescarlett: 💥

packages/tooltip/src/Tooltip.tsx
hswolff: If `shouldClose` is set to a non-function, like a string, then the logic breaks.

packages/popover/src/Popover.tsx
```diff
@@ -36,11 +34,17 @@ const mutationOptions = {
...
+  justification: Justify | Justification;
+  referenceElPos: ElementPosition;
+}
+
 export interface PopoverProps {
   /**
    * Content that will appear inside of the popover component.
    */
-  children: ReactNode;
+  children: ReactNode | ((Object: FunctionParameters) => ReactNode);
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/popover/src/types.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * Should the Popover auto adjust its content when the DOM changes (using MutationObserver).
+   *
+   * default: false
+   */
+  adjustOnMutation?: boolean;
+}
+
+export { Align, Justification, Justify };
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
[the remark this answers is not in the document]
brookescarlett: 👍 Changed the logic per a previous comment, but still using the TooltipProps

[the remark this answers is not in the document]
brookescarlett: But Typescript doesn't like using TooltipProps as a value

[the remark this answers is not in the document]
hswolff: ? What do you mean? I saw some usages of the resulting variable that TS didn't like as we use it as `setState(value)` and `setState((val) => newVal)` that we didn't define.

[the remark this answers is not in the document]
brookescarlett: `const isControlled = typeof controlledOpen === TooltipProps['open']`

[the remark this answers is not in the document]
brookescarlett: Doing it this way now per a discussion we had (perhaps in person, or further up in the thread)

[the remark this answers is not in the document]
brookescarlett: But TS throwing: `TooltipProps only refers to a type, but is being used as a value here`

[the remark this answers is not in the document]
hswolff: Uh I meant to only replace that one line. Not the entire expression.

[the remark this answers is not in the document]
brookescarlett: Sorry was looking at the wrong thing. I see now.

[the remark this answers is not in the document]
brookescarlett: Because in the type definition of setOpen could be undefined, TS is complaining in all usages that it can't invoke an object that is possibly undefined.

packages/tooltip/src/Tooltip.tsx
brookescarlett: We're defaulting to handling the open/close behavior ourselves. The shouldClose() is just an optional hook into this by allowing a consumer to pass a callback function that can return true/false to determine whether or not something should close when we think it should. Happy to add a test, but I don't think the logic is broken.

packages/popover/src/types.ts
```diff
@@ -0,0 +1,123 @@
...
+ */
+const Justify = {
+  Start: 'start',
+  Middle: 'middle',
+  End: 'end',
+} as const;
+
+type Justify = typeof Justify[keyof typeof Justify];
+
+export { Align, Justification, Justify };
```
hswolff: Oh sorry, I meant `export const Justify` etc.

packages/tooltip/src/Tooltip.tsx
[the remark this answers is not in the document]
hswolff: This works:

```js
  const setOpen = isControlled && controlledSetOpen
    ? controlledSetOpen
    : uncontrolledSetOpen;
```

packages/popover/src/types.ts
brookescarlett: We're using Align, Justify and Justification as both a type and value so the export { Align } signature allows us to do that. Exported individually after the declarations though.

matt-d-rat:
Hey LeafyGreen team! I wanted to ask whether you have an ETA on when this PR will be merged and this package released for consumption?

The reason I ask is because this sprint I am working on porting the `Pipeline` component from MongoDB Charts to LeafyGreen (see [PD-381](https://jira.mongodb.org/browse/PD-381) / [CHARTS-2864](https://jira.mongodb.org/browse/CHARTS-2864) for details), and one of the functional requirements of this component is to show a tooltip when the counter is hovered over to display details about the full pipeline.

Currently in Charts we use `react-tooltip` for our tooltips, but I would prefer to leverage a LeafyGreen package, and the timing of this PR seems to be perfect.

I have another week left in our sprint to get this work finished and a PR raised.

brookescarlett:
Hi @matt-d-rat I believe we'll be able to get this released this week (hopefully today)! Will keep you posted and let you know as soon as it's been published to NPM.

PR #167, opened 2019-10-07
PD-409: Add useEscapeHandler hook

brookescarlett:
## ✍️ Proposed changes

- Added useEscapeHandler hook to hooks folder
- Replaced existing instances of useEventListener (where listening for EscapeKey) to using new hook

🎟 _Jira ticket:_ [PD-409](https://jira.mongodb.org/browse/PD-409)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/hooks/src/useHandleEscape.ts
+++ b/packages/hooks/src/useHandleEscape.ts
@@ -3,9 +3,8 @@ import useEventListener, { UseEventOptions } from './useEventListener';
 const escapeKeyCode = 27;
 
 const handleEscape = (e: KeyboardEvent, callback: () => void) => {
-  e.stopImmediatePropagation();
-
   if (e.keyCode === escapeKeyCode) {
+    e.stopImmediatePropagation();
     callback();
   }
 };
--- a/packages/hooks/src/useHandleEscape.ts
+++ b/packages/hooks/src/useHandleEscape.ts
@@ -1,0 +1,26 @@
+import useEventListener, { UseEventOptions } from './useEventListener';
+
+const escapeKeyCode = 27;
+
+const handleEscape = (e: KeyboardEvent, callback: () => void) => {
+  e.stopImmediatePropagation();
+
+  if (e.keyCode === escapeKeyCode) {
+    callback();
+  }
+};
+
+/**
+ * Hook to subscribe to an escape-key-press.
+ * @param callback Callback function to be executed when Escape key is pressed.
+ * @param options Object to refine when callback is invoked.
+ * @param options.options Parameter to specify options passed to the eventListener.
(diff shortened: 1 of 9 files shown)


packages/hooks/src/useHandleEscape.ts
```diff
@@ -0,0 +1,25 @@
+import useEventListener, { UseEventOptions } from './useEventListener';
+
+const escapeKeyCode = 27;
+
+const handleEscape = (e: KeyboardEvent, callback: () => void) => {
+  if (e.keyCode === escapeKeyCode) {
+    e.stopImmediatePropagation();
```
hswolff: Why do we want this? This will stop any other keydown listeners from firing, curious why we want this?

packages/hooks/src/useHandleEscape.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Hook to subscribe to an escape-key-press.
+ * @param callback Callback function to be executed when Escape key is pressed.
+ * @param options Object to refine when callback is invoked.
+ * @param options.options Parameter to specify options passed to the eventListener.
+ * @param options.enabled Determines whether or not the useEffect hook should run.
+ * @param options.dependencies Array to be passed to useEffect hook, such that the hook will only run if the array's values have changed.
+ * @param options.element Value to be passed as target of event handler, will default to document.
+ */
+const useHandleEscape = (callback: () => void, options?: UseEventOptions) => {
```
hswolff: nit, I know I'm late, but what do we think of the name `useEscapeKey`?

packages/hooks/src/useHandleEscape.ts
hswolff: Perhaps it should be handled on a per usage basis instead of globally?

I guess I'm looking for the thinking behind this as removing it may be tricky.

brookescarlett: We always had `e.stopImmediateProp()` when using the generic version in Modal, Menu and Tooltip components. We can definitely think about removing or finding a way to handle per usage?

packages/hooks/src/useHandleEscape.ts
brookescarlett: Happy to rename, thoughts @DesignerDave

packages/hooks/src/useHandleEscape.ts
hswolff: Perhaps we should call [stopPropagation](https://stackoverflow.com/a/5299841) instead? ImmediatePropagation seems a little heavy and could conflict with other application code.

brookescarlett: Updates to useEscape: https://github.com/mongodb/leafygreen-ui/pull/168

PR #168, opened 2019-10-08
PD-409: useEscapeKey

brookescarlett:
- Renamed useEscapeHandler to useEscapeKey
- Updated instances in use and changeset to reflect new name
- Removed `immediateProp` in favor of `stopProp`

### DIFF
--- a/packages/hooks/src/useEscapeKey.ts
+++ b/packages/hooks/src/useEscapeKey.ts
@@ -4,7 +4,7 @@ const escapeKeyCode = 27;
 
 const handleEscape = (e: KeyboardEvent, callback: () => void) => {
   if (e.keyCode === escapeKeyCode) {
-    e.stopPropagation();
+    e.stopImmediatePropagation();
     callback();
   }
 };
--- a/packages/hooks/src/useEscapeKey.ts
+++ b/packages/hooks/src/useEscapeKey.ts
@@ -4,7 +4,7 @@ const escapeKeyCode = 27;
 
 const handleEscape = (e: KeyboardEvent, callback: () => void) => {
   if (e.keyCode === escapeKeyCode) {
-    e.stopImmediatePropagation();
+    e.stopPropagation();
     callback();
   }
 };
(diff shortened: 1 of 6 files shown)


packages/hooks/src/useEscapeKey.ts
```diff
@@ -4,7 +4,7 @@ const escapeKeyCode = 27;
 
 const handleEscape = (e: KeyboardEvent, callback: () => void) => {
   if (e.keyCode === escapeKeyCode) {
-    e.stopImmediatePropagation();
+    e.stopPropagation();
```
DesignerDave: Hmm... What was the original reason for it being `stopImmediatePropagation`?

If we were to have a Menu open within a modal, and the user opens the menu, pressing escape to close it would close both the menu and the modal with `stopPropagation`. I'm wondering if this was the original reason `stopImmediatePropagation was used...`

brookescarlett: IIRC: we used `stopImmediateProp()` because we were using one callback to handle both mouse events and keyboard events, so we had a conditional at one point and used both (stop and stop immediate). Now that we've broken this out, I think we're okay to only use `stopProp()`. But that may only have been part of it, thinking about the point you raise above @DesignerDave.

hswolff: Yeah I'm really torn on this cuz I don't fully understand all the possible permutations.

My main theoretical worry with `stopImmediatePropagation` is that it can affect application code more concretely.

DesignerDave: In this specific case, it would only affect escape key presses, which is desirable here in my mind. I'm under the impression that this affects the ability to use multiple leafygreen components together, and we may regret this decision.

brookescarlett: To note: Where we were using it already, we were using `stopImmediatePropagation()` so we wouldn't be adding any new behavior by leaving as-is.

PR #170, opened 2019-10-08
PD-359: Develop shareable way to determine if a focus state should be shown

DesignerDave:
## ✍️ Proposed changes

This PR addresses several related tickets and issues around focus state display.

1. It adds an `InteractionProvider` component that currently implements a context provider for whether a user is navigating the page with their keyboard or mouse.
2. It adds a hook called `useShowFocus` that easily gets the value of the context provider.
3. It implements the logic needed to correctly utilize the state in `MongoMenu` and `RadioBoxGroup`, while also gracefully degrading when the provider is not implemented.
4. The focus state for `RadioBoxGroup` previously didn't work at all. This fixes that, and also implements (what I'm calling) the "Interaction Ring" motif that we used for the active state, for the focus and hover states.
5. A stacking context was not previously being created in the context of RadioBox. The z-index used for displaying the interaction ring should no longer be applied within a global context.

🎟 [Develop shareable way to determine if a focus state should be shown](https://jira.mongodb.org/browse/PD-359)
🎟 [Implement improved focus state in Radio Box Group](https://jira.mongodb.org/browse/PD-403)

## 🛠 Types of changes

## ✅ Checklist


packages/interaction-context/src/InteractionContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+  | undefined);
+
+interface InteractionContextProps {
+  children: React.ReactNode;
+  initialStates?: {
+    usingKeyboard?: boolean;
+  };
+}
+
+const InteractionContext = ({
```
hswolff: This should be `InteractionProvider`. The context contains the provider and the consumer.

packages/menu/src/Menu.story.js
```diff
@@ -1,56 +1,60 @@
...
-          </MenuItem>
-          <MenuItem description="I am a description">
-            Menu Item With Description
-          </MenuItem>
-          <MenuItem href="http://mongodb.design">I am a link!</MenuItem>
-        </MenuGroup>
-        <MenuItem>Left out of the MenuGroup</MenuItem>
-      </Menu>
-    </button>
+    <InteractionContext>
```
hswolff: If we keep on this trend of having global context providers we may want to have one `LeafygreenProvider` in the future that wraps them all up for us. Don't need it now, but just an idea.

packages/interaction-context/src/InteractionContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+import PropTypes from 'prop-types';
+import { useEventListener } from '@leafygreen-ui/hooks';
+
+export const UsingKeyboardContext = createContext(undefined as
+  | boolean
+  | undefined);
+
+interface InteractionContextProps {
+  children: React.ReactNode;
+  initialStates?: {
```
hswolff: nit

```suggestion
  initialState?: {
```

nit

packages/interaction-context/src/useShowFocus.ts
```diff
@@ -0,0 +1,13 @@
+import { useContext } from 'react';
+import { UsingKeyboardContext } from './InteractionContext';
+
+export default function useShowFocus() {
+  const usingKeyboard = useContext(UsingKeyboardContext);
+
+  // If there's no context provider available, we show focus states as usual.
+  if (usingKeyboard === null || usingKeyboard === undefined) {
```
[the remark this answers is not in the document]
DesignerDave: I think a while back, we decided not to use this shorthand? It was a while ago though, but it does pain me to write this.

packages/interaction-context/src/useShowFocus.ts
```diff
@@ -0,0 +1,13 @@
+import { useContext } from 'react';
+import { UsingKeyboardContext } from './InteractionContext';
+
+export default function useShowFocus() {
+  const usingKeyboard = useContext(UsingKeyboardContext);
```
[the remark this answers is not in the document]
DesignerDave: So this is the weird part – it returns null if the context doesn't exist, and undefined when it does (since that's the initial value we set for the context in our call to `createContext`). We consider this below when setting the value the hook returns.

packages/menu/src/Menu.story.js
DesignerDave: That's the idea for this, and why we didn't just export `UsingKeyboardContext.Provider` instead. We can add context providers in the InteractionContext component. If you have better ideas for this name, let me know (aside from renaming to `InteractionProvider`).

DesignerDave: I know @brookescarlett wasn't a fan of `LeafyGreenProvider` initially

packages/interaction-context/src/InteractionContext.tsx
DesignerDave: Fair enough!

packages/menu/src/Menu.story.js
brookescarlett: I wasn't at first, because I didn't think of it as a wrapper. I support having one wrapper called LeafyGreen provider.

packages/interaction-context/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import InteractionContext, { UsingKeyboardContext } from './InteractionContext';
+import useShowFocus from './useShowFocus';
+
+export { UsingKeyboardContext, useShowFocus };
```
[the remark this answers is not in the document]
DesignerDave: Added a use-case for exporting this. Let me know if you still think we should build an abstraction on top of the context for the new setter.

packages/menu/src/Menu.story.js
DesignerDave: Done.

packages/interaction-context/src/InteractionContext.tsx
DesignerDave: There were more places to change this, so I did so manually. Either way, done.

packages/interaction-context/src/InteractionContext.tsx
DesignerDave: Changed the component to `LeafyGreenProvider`

packages/leafygreen-provider/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import LeafyGreenProvider, { UsingKeyboardContext } from './LeafyGreenProvider';
+import useShowFocus from './useShowFocus';
+
+export { UsingKeyboardContext, useShowFocus };
```
hswolff: Can we not export the context publicly? That shouldn't be externally accessible. We don't want people to directly interface with the context outside of our control.

packages/leafygreen-provider/src/LeafyGreenProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const NavigationKeyCodes = {
+  tab: 9,
+  leftArrow: 37,
+  upArrow: 38,
+  rightArrow: 39,
+  downArrow: 40,
+} as const;
+
+type NavigationKeyCodes = typeof NavigationKeyCodes[keyof typeof NavigationKeyCodes];
```
hswolff: Are we using this type anywhere? Do we need to define it?

packages/leafygreen-provider/src/LeafyGreenProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+interface LeafyGreenProviderProps {
+  children: React.ReactNode;
+  initialState?: {
+    usingKeyboard?: boolean;
+  };
+}
+
+const LeafyGreenProvider = ({
```
hswolff: Ideally this is a very simple component. What you have here is the former `KeyboardContextProvider`. To prepare for future context's we should still have `KeyboardContextProvider` and have this component merely use that provider. Such that it becomes trivial to add or remove contexts in the future. 

I'd even argue to make UsingKeyboard related code its own file.

packages/leafygreen-provider/src/LeafyGreenProvider.tsx
```diff
@@ -0,0 +1,72 @@
...
+      // All are used to manage focus in a keyboard navigation context.
+      if (Object.values<number>(NavigationKeyCodes).includes(keyCode)) {
+        setUsingKeyboard(true);
+      }
+    },
+    { enabled: !usingKeyboard },
+  );
+
+  return (
+    <UsingKeyboardContext.Provider value={{ usingKeyboard, setUsingKeyboard }}>
```
hswolff: Having an inline object here will cause un-necessary re-renders as its a new object instance on every render. The value object should be memoized.

packages/leafygreen-provider/src/LeafyGreenProvider.tsx
DesignerDave: We only use it in this file and the tests. Will remove the type.

packages/leafygreen-provider/src/UsingKeyboardProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+  leftArrow: 37,
+  upArrow: 38,
+  rightArrow: 39,
+  downArrow: 40,
+} as const;
+
+export function useShowFocus(): boolean {
+  const { usingKeyboard } = useContext(UsingKeyboardContext);
+
+  if (usingKeyboard === null || usingKeyboard === undefined) {
```
brookescarlett: Can use `!= null` based on recent discussion

packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -9,58 +9,61 @@ import {
...
 
-const buttonReset = css`
+const wrapperStyle = css`
   appearance: none;
   background: none;
   border: 0px;
   position: relative;
   padding: 0px;
 
+  &::-moz-focus-inner {
```
brookescarlett: Should this be handled in button now as well?

packages/radio-box-group/package.json
```diff
@@ -13,7 +13,8 @@
   },
   "dependencies": {
     "@leafygreen-ui/lib": "^3.0.6",
```
brookescarlett: Let's make sure all of these deps are updated before publishing.

packages/mongo-menu/package.json
```diff
@@ -16,6 +16,7 @@
     "@leafygreen-ui/menu": "4.0.1",
```
brookescarlett: Make sure that all deps are up to date

packages/menu/package.json
```diff
@@ -15,7 +15,8 @@
     "@leafygreen-ui/hooks": "^1.1.0",
     "@leafygreen-ui/lib": "^3.2.0",
     "@leafygreen-ui/popover": "^2.0.0",
-    "@leafygreen-ui/palette": "^1.1.0"
+    "@leafygreen-ui/palette": "^1.1.0",
```
brookescarlett: Make sure all deps are up to date

packages/leafygreen-provider/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js"
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/lib": "^3.0.6",
```
brookescarlett: update deps

DesignerDave:
> This is epic. A couple of questions/nits spread throughout but excited to have this!
> 
> Just one thing: Is there a reason why certain components were selected over others to demo the new provider? Will we be making tickets to update the rest of the components in the repo as well?

So, partly this was to test different aspects of the API. `Menu` was a great place to do this because of the manual focus when the menu is opened. On the other hand, RadioBoxGroup specifically was entirely broken, and didn't show focus states pretty much at all. This ticket was a good time to fix that.

We will have tickets to update the rest, where appropriate.

packages/leafygreen-provider/src/LeafyGreenProvider.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import UsingKeyboardProvider from './UsingKeyboardProvider';
+
+interface LeafyGreenProviderProps {
+  children?: React.ReactNode;
+}
+
+function LeafyGreenProvider({ children }: LeafyGreenProviderProps) {
+  return <UsingKeyboardProvider>{children}</UsingKeyboardProvider>;
```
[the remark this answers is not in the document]
DesignerDave: Yes, say we add some random provider. It would look like:
```js
return (
  <UsingKeyboardProvider>
    <NewProvider>
      {children}
    </NewProvider>
  </UsingKeyboardProvider>
```

As far as common provider design patterns, I don't think there's another great way to do this. We could have one combined state provider, but that would probably cause a few more renders than would be ideal, and doesn't let us isolate the functionality as much.

packages/leafygreen-provider/src/UsingKeyboardProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+  setUsingKeyboard?: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+export const UsingKeyboardContext = createContext<UsingKeyboardState>({
+  usingKeyboard: undefined,
+  setUsingKeyboard: undefined,
+});
+
+// All keys here are used to manage focus through keyboard interaction.
+export const NavigationKeyCodes: { readonly [k: string]: number } = {
```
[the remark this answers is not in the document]
DesignerDave: I don't think we can use what's in lib as-is. The enum in lib includes keys that aren't used to manage focus. `Enter` and `Space` in particular take action on a focused element, but wouldn't be used to set focus.

AFAIK, yes. These should be the only keys used to set focus on an element.

packages/mongo-menu/src/MongoMenu.tsx
DesignerDave: Maybe once we convert button to use `useShowFocus`.

packages/leafygreen-provider/src/UsingKeyboardProvider.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+interface UsingKeyboardProviderProps {
+  children?: React.ReactNode;
+}
+
+function UsingKeyboardProvider({ children }: UsingKeyboardProviderProps) {
+  const [usingKeyboard, setUsingKeyboard] = useState(false);
+
+  useEventListener('mousedown', () => setUsingKeyboard(false), {
```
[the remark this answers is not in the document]
DesignerDave: I suppose we don't need to enable / disable them, but I'm not sure what case you see where it's more expensive to register them when needed rather than having them be persistent. They should only swap on user interaction, so the actual registering of the events should have a negligible effect on performance.

That said, I'm not convinced it will have a large impact either way, and am happy to leave both event listeners always enabled if you strongly feel this will be a performance bottleneck.

packages/leafygreen-provider/src/UsingKeyboardProvider.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { createContext, useContext, useState, useMemo } from 'react';
+import PropTypes from 'prop-types';
+import { useEventListener } from '@leafygreen-ui/hooks';
+
+interface UsingKeyboardState {
+  usingKeyboard?: boolean;
+  setUsingKeyboard: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+export const UsingKeyboardContext = createContext<UsingKeyboardState>({
```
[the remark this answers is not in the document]
DesignerDave: Great point – definitely missed the point that we can handle graceful degradation straight from here. Will change this up.

packages/leafygreen-provider/src/UsingKeyboardProvider.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+afterAll(cleanup);
+
+describe('packages/leafygreen-provider/UsingKeyboardProvider', () => {
+  const childTestID = 'using-keyboard-provider';
+  const buttonTestId = 'test-button';
+
+  function TestContextComponent() {
+    const { usingKeyboard, setUsingKeyboard = () => {} } = useContext(
+      UsingKeyboardContext,
```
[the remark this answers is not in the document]
DesignerDave: I was attempting to create a separation of concerns when it comes to testing the provider. Given how a break in this component might affect a consuming application, I feel this is a case where we should be testing the component in a more granular fashion.

Perhaps I'm overly paranoid, but I feel we are marginally more likely to catch if the API is returning the wrong value from here, than if we only test it in the hook.

packages/leafygreen-provider/src/LeafyGreenProvider.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import UsingKeyboardProvider from './UsingKeyboardProvider';
+
+interface LeafyGreenProviderProps {
+  children?: React.ReactNode;
```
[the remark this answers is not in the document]
DesignerDave: Sure haha

packages/mongo-menu/src/MongoMenu.story.tsx
```diff
@@ -1,18 +1,21 @@
...
-      name: text('name', 'Alex Smith'),
-      email: text('email', 'alex.smith@mongodb.com'),
-    }}
-    activeProduct={select(
-      'activeProduct',
-      Object.values(Product) as Array<Size>,
-      Product.Atlas,
-    )}
-  />
+  <LeafyGreenProvider initialStates={{ usingKeyboard: false }}>
```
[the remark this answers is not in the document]
DesignerDave: Nope! Good catch

packages/leafygreen-provider/src/UsingKeyboardProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+import PropTypes from 'prop-types';
+import { useEventListener } from '@leafygreen-ui/hooks';
+
+interface UsingKeyboardState {
+  usingKeyboard: boolean;
+  setUsingKeyboard: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+export const UsingKeyboardContext = createContext<UsingKeyboardState>({
+  usingKeyboard: true,
```
hswolff: This should probably default to false, so we don't show a focus state by default everywhere?

packages/leafygreen-provider/src/UsingKeyboardProvider.tsx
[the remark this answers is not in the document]
hswolff: Meh, let's keep it as is and see how it goes. I have no proof either way and not really concerned to measure it now.

packages/leafygreen-provider/src/useUsingKeyboardContext.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { render, cleanup, fireEvent } from '@testing-library/react';
+import UsingKeyboardProvider, {
+  useUsingKeyboardContext,
+} from './UsingKeyboardProvider';
+
+afterAll(cleanup);
+
+const genId = () => Math.round(Math.random() * 1000000).toString();
+
+describe('useUsingKeyboardContext', () => {
```
hswolff: Feels like this should be a child of the `UsingKeyboardProvider.spec.tsx` test file to keep our conventions consistent. One src file, one test file.

packages/leafygreen-provider/src/UsingKeyboardProvider.tsx
DesignerDave: This is being overly cautious, but this is intentional – the fallback should always be to show focus states first. Once a user clicks anywhere on the page, focus states won't be shown, so this should be a non-issue.

hswolff: Wouldn't we show focus states everywhere then? Wouldn't that cause issues?

If they click anywhere that'll only affect css styles tho right? Not those controlled by this provider?

DesignerDave: Focus states are still shown with `:focus` CSS pseudo-classes. Where this provider is used, we're simply adding or removing those pseudo-classes rather than toggling focus states on everything.

hswolff: Ah ok, got it. Perfect!

hswolff: Ah ok, got it. Perfect!

packages/leafygreen-provider/src/useUsingKeyboardContext.spec.tsx
DesignerDave: Makes sense! Done ✅

PR #173, opened 2019-10-14
PD-411: add accountURL prop

brookescarlett:
## ✍️ Proposed changes

Add accountURL prop to MongoMenu so that consuming application can decide where this button links out to.

🎟 _Jira ticket:_ [PD-411](https://jira.mongodb.org/browse/PD-411)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/mongo-menu/README.md
+++ b/packages/mongo-menu/README.md
@@ -50,7 +50,7 @@ import MongoMenu from '@leafygreen-ui/mongo-menu';
         <h3 class="leafygreen-ui-i2e27j">Alex Smith</h3>
         <p class="leafygreen-ui-1jozpc5">alex.smith@mongodb.com</p>
         <a
-          href="https://cloud.mongodb.com/account/profile"
+          href="https://cloud.mongodb.com/v2#/account"
           class="leafygreen-ui-1q7ajb0"
           aria-disabled="false"
         >
@@ -116,7 +116,7 @@ Object that contains information about the active user. {name: 'string', email: 
 
 **Type:** `string`
 
-**Default:** `https://cloud.mongodb.com/account/profile`
+**Default:** `https://cloud.mongodb.com/v2#/account`
 
 URL passed to MongoDB Account button.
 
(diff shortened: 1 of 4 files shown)


packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -206,6 +211,7 @@ interface MongoMenuProps {
  */
 function MongoMenu({
   user: { name, email },
+  accountURL = 'https://cloud.mongodb.com/account/profile',
```
hswolff: This URL doesn't yet exist, and it also might not be the final location.

The current location is `https://cloud.mongodb.com/v2#/account`, so perhaps we default to that until we have the actual working new URL?

brookescarlett: 👍

packages/mongo-menu/README.md
```diff
@@ -50,7 +50,7 @@ import MongoMenu from '@leafygreen-ui/mongo-menu';
         <h3 class="leafygreen-ui-i2e27j">Alex Smith</h3>
         <p class="leafygreen-ui-1jozpc5">alex.smith@mongodb.com</p>
         <a
-          href="https://account.mongodb.com"
+          href="https://cloud.mongodb.com/account/profile"
```
hswolff: Need to update here.

brookescarlett: Updated README 👍

packages/mongo-menu/src/MongoMenu.tsx
DesignerDave: Ahh, @Faeylayn was saying that this was where it was gonna live until it was moved to `account.mongodb.com`. Just wanna double-check this.

Faeylayn: We can agree that defaulting the url at currently to v2#/account makes sense, currently the account/profile is only going to exist in a placeholder setup for a while now. This just means we'll need to have another release to change this to the URL we will be using while the account App is a part of MMS and then another change for when the account app is separated from MMS. Currently the App is being built where it will exist at https://cloud.mongodb.com/account/profile and then will move to the account.mongodb.com at some point in the far flung future

brookescarlett: 👍

PR #175, opened 2019-10-28
PD-402: Implement Accessible Keyboard Nav for Menu

brookescarlett:
## ✍️ Proposed changes

- Implement accessible keyboard navigation for Menu component: [guidelines here](https://www.w3.org/TR/wai-aria-practices-1.1/#menu)

🎟 _Jira ticket:_ [PD-402](https://jira.mongodb.org/browse/PD-402)

## 🛠 Types of changes

## ✅ Checklist


packages/menu/src/Menu.tsx
```diff
@@ -71,25 +88,64 @@ function Menu({
...
+  ): Array<React.ReactElement> => {
+    return React.Children.map(childset, child => {
+      if (isMenuItemElement(child)) {
+        menuitems.push(child);
+        return React.cloneElement(child, {
+          ref: (ref: HTMLElement) => refs.push(ref),
+        });
+      }
+
+      if (typeof child.props.children === 'object') {
```
hswolff: What does this check mean? What does it mean that children is an object? May need to add a comment.

packages/menu/src/Menu.tsx
```diff
@@ -71,25 +88,64 @@ function Menu({
...
+        });
+      }
+
+      return child;
+    });
+  };
+
+  const updatedChildren = updateChildren(children);
+
+  const enabledIndexes = menuitems.reduce(
```
hswolff: Menu accepts children that aren't MenuItem's right?

What happens to this functionality if that's the case?

packages/menu/src/Menu.tsx
```diff
@@ -104,14 +160,64 @@ function Menu({
     }
   };
 
-  const enabled = open || uncontrolledOpen;
+  const enabled = open;
```
hswolff: Be better to just use the `open` variable directly and not use this `enabled` variable.

packages/menu/src/Menu.tsx
brookescarlett: No longer using this, relying on refs now

packages/menu/src/FocusableMenuItem.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+export interface FocusableMenuItemProps {
+  children: React.ReactElement;
+}
+
+const FocusableMenuItem = React.forwardRef(
```
hswolff: Can we add some inline docs on what this component's purpose is?

packages/menu/src/Menu.story.js
```diff
@@ -30,27 +30,24 @@ function Uncontrolled() {
...
-        </MenuGroup>
-        <MenuItem>Left out of the MenuGroup</MenuItem>
-      </Menu>
-    </button>
+    <Menu
+      align={select('Align', Object.values(Align), Align.Bottom)}
+      justify={select('Justify', Object.values(Justify), Justify.Start)}
+      open={open}
+      setOpen={setOpen}
+      trigger={<button>trigger</button>}
```
hswolff: We updated this example to use this format? Is the previous version still possible if people desire, where the trigger is placed external of the component?

packages/menu/src/Menu.tsx
```diff
@@ -5,15 +5,47 @@ import { useEventListener, useEscapeKey } from '@leafygreen-ui/hooks';
...
+  margin-block-end: 0px;
+  padding-inline-start: 0px;
 `;
 
-interface MenuProps extends Omit<PopoverProps, 'active' | 'spacing'> {
+function isMenuItemElement(
+  element: React.ReactNode,
+): element is React.ReactElement<MenuItemProps, typeof MenuItem> {
+  return (
+    element != null &&
```
hswolff: Let's abstract these common checks? Or rather seems like this built in [React method isValidElement](https://reactjs.org/docs/react-api.html#isvalidelement) does exactly what we need!

[Source reference which is very similar!](https://github.com/facebook/react/blob/d3622d0f977def825123f1d5f4cef19888b1eaf1/packages/react/src/ReactElement.js#L507)

.changeset/sixty-pianos-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/button': major
+---
+
+Component now composed of React.forwardRef()
```
hswolff: nit: Wrapping component with React.forwardRef to provide direct access to the underlying element.

brookescarlett: 👍

packages/menu/src/Menu.tsx
brookescarlett: Can't use isValidElement, because we want to be more specific than just is a React Component. But can definitely abstract this and put in `lib`.

packages/menu/src/FocusableMenuItem.tsx
brookescarlett: 👍

packages/menu/src/Menu.story.js
brookescarlett: Yes and reset back to initial state 😄

packages/menu/src/Menu.tsx
```diff
@@ -104,48 +133,111 @@ function Menu({
...
+
+    return false;
+  }
+
+  function handleKeyDown(e: KeyboardEvent) {
+    const filteredRefs = refs.filter(ref => ref !== null);
+    let refToFocus: HTMLElement;
+
+    switch (e.key) {
+      case 'ArrowDown':
```
DesignerDave: For maximum browser compatibility, we should stick to keyCode, rather than key.

packages/menu/src/MenuItem.tsx
```diff
@@ -20,6 +19,10 @@ const containerStyle = css`
   text-decoration: none;
   position: relative;
   transition: background-color 150ms ease-in-out;
+  border: none;
+  margin: unset;
+  width: 100%;
+  font-family: 'Akzidenz';
```
DesignerDave: We should always set font-family fallbacks. Can we make the font-family list here consistent with our other uses?

```css
font-family: Akzidenz, 'Helvetica Neue', Helvetica, Arial, sans-serif;
```

As an aside, we should think about how to share design tokens like this in the future.

packages/menu/src/MenuSeparator.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { css } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+
+const borderStyle = css`
+  border-top: 1px solid ${uiColors.gray.light1};
+`;
+
+function MenuSeparator() {
+  return <li role="separator" className={borderStyle}></li>;
```
DesignerDave: This can be self-terminating.

packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -237,31 +253,33 @@ function MongoMenu({
...
-          <p className={descriptionStyle}>{email}</p>
+        <div className={topPadding}></div>
+        <h3 className={cx(nameStyle, truncate, horizontalPadding)}>{name}</h3>
+        <p className={cx(horizontalPadding, descriptionStyle)}>{email}</p>
+        <FocusableMenuItem>
           <Button
             size="small"
             href={accountURL}
-            className={accountButtonStyle}
+            className={cx(accountButtonStyle)}
```
DesignerDave: Not composing any styles here, so `cx` isn't needed.

packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -237,31 +253,33 @@ function MongoMenu({
...
-              active={el.slug === activeProduct}
-              href={el.href}
-              description={el.description}
-            >
-              {el.displayName}
-            </MenuItem>
-          ))}
-        </MenuGroup>
+        </FocusableMenuItem>
+        <div className={bottomPadding}></div>
```
DesignerDave: These shouldn't be their own divs. We should just use margin here.

PR #183, opened 2019-11-05
PD-430: Use keyCode rather than Key for Tab keyboard navigation

brookescarlett:
## ✍️ Proposed changes

- Use keyCode rather than key in Tabs keyboard navigation

🎟 _Jira ticket:_ [PD-430](https://jira.mongodb.org/browse/PD-430)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/tabs/src/Tabs.tsx
+++ b/packages/tabs/src/Tabs.tsx
@@ -47,16 +47,6 @@ const focusedStyle = css`
   background-color: ${uiColors.blue.base};
 `;
 
-// function isTab(
-//   element: React.ReactNode,
-// ): element is React.ReactElement<TabProps, typeof Tab> {
-//   if (!(element != null && typeof element === 'object' && 'type' in element)) {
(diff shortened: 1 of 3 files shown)


packages/tabs/src/Tabs.tsx
```diff
@@ -158,14 +153,14 @@ function Tabs({
 
     const enabledCurrentIndex = enabledIndexes.indexOf(selected!);
 
-    switch (e.key) {
-      case 'ArrowRight':
+    switch (e.keyCode) {
+      case keyMap['ArrowRight']:
```
hswolff: Why are we doing a string lookup?

brookescarlett: I thought it made it more readable, can eliminate if you think that makes more sense.

brookescarlett: 🤔 Never knew that, thank you!

PR #184, opened 2019-11-06
Release Branch

brookescarlett:
Update changesets

### DIFF
--- a/packages/mongo-menu/CHANGELOG.md
+++ b/packages/mongo-menu/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Major Changes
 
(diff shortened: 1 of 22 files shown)


packages/mongo-menu/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # @leafygreen-ui/mongo-menu
 
+## 2.0.0
+
+### Major Changes
+
+- f6b6b7a: Content that appears in Popover is no longer present on DOM by default
```
DesignerDave: Could we change this to something like this (and update other instances of it)?
```
Children of Popover are no longer rendered to the DOM when the Popover is closed.
```

PR #191, opened 2019-11-08
PD-381: Port MongoDB Charts Pipeline summary component to LeafyGreen (CHARTS-2864)

matt-d-rat:
## ✍️ Proposed changes

Hey LeafyGreen team! This is my first PR to LeafyGreen, and also the first time I have ever used TypeScript and Emotion so go easy on me 😄.

This PR ports the Pipeline summary component from MongoDB Charts to LeafyGreen. The original implementation of this component can be found [here](https://github.com/10gen/charts/tree/master/packages/charts-frontend/src/components/pipeline). The port involved converting the original implementation from JavaScript to TypeScript, as well as porting the tests from Mocha / Enzyme to Jest / Testing Library.

Additionally I fixed a TypeScript build error in `MenuItem` that was preventing me from successfully building the project ~~(the `title` prop wasn't defined on the interface)~~ (removed the deprecated `title` prop from the `propTypes`).

I have also upgraded `jest-dom` to version 4, the breaking change of which required me to update any imports of the library from `jest-dom` to `@testing-library/jest-dom` in order to fix a number of build errors I was getting in other packages.

🎟 _Jira ticket:_ [PD-381](https://jira.mongodb.org/browse/PD-381) / [CHARTS-2864](https://jira.mongodb.org/browse/CHARTS-2864)

## 🛠 Types of changes

## ✅ Checklist

### Screencast
pipeline

Some things to call out about the implementation:

- In order to test the behaviour of the `Stage` component I had to mock out the browser `IntersectionObserver` API. I did this via the implementation found in `pipeline/src/mocks/IntersectionObserver.ts` though I am not sure if this is the best place for this to live?
- The style definitions for the `Stage` and `Counter` are essentially identical, so I tried to convert them the best way that I could from `Less` to `Emotion` via helper functions. Not sure if this is the best approach or whether there are alternatives?
- I was unable to find a way to test the behaviour of the `useMutationObserver` used by the `Pipeline` component. I attempted to use the `waitForDomChange` from `@testing-library/dom`, but the tests always timed out. Not sure if this is a problem with the test that I was writing, or with the `@testing-library/dom` dependency itself.


packages/pipeline/README.md
```diff
@@ -0,0 +1,91 @@
...
+    />
+  </div>
+</div>
+```
+
+## Properties
+
+### Pipeline
+
+| Prop        | Type     | Description                                                                                                                                                                                                | Default     |
```
brookescarlett: Matched our other readmes to this format, I dig it!

packages/pipeline/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js"
+  },
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
```
brookescarlett: Please make sure to use the latest versions of the @leafygreen-ui components

packages/pipeline/src/Stage.tsx
```diff
@@ -0,0 +1,208 @@
...
+}
+
+interface StageProps {
+  /**
+   * Content that will appear inside of the Pipeline component.
+   **/
+  children: ReactNode;
+
+  /**
+   * Classname applied to Pipeline content container.
```
brookescarlett: Think you mean className applied to the Stage content container, ditto with above (children prop)

packages/menu/src/MenuItem.tsx
```diff
@@ -158,6 +158,11 @@ export interface MenuItemProps {
    */
   children?: React.ReactNode;
 
+  /**
```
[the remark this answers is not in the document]
matt-d-rat: Okay cool, I figured it might be one or the other, and so decided to air on the side of caution. I will remove this prop entirely. Thanks for clarifying.

packages/pipeline/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@leafygreen-ui/pipeline",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
matt-d-rat: Makes sense, I will change to `0.9.0` for now.

packages/pipeline/package.json
matt-d-rat: Yeah I think you guys cut a new release shortly after I had raised the PR. I will update to use the latest version of all dependencies.

packages/pipeline/src/Stage.tsx
matt-d-rat: Whoops, copy/paste typo. Will update.

packages/menu/src/MenuItem.tsx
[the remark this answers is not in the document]
matt-d-rat: Resolved in 1bcbfd1

packages/pipeline/package.json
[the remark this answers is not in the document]
matt-d-rat: Resolved in 090a6f0

packages/pipeline/package.json
matt-d-rat: Resolved in 090a6f0

packages/pipeline/src/Stage.tsx
matt-d-rat: Resolved in 4fb3d4b

packages/pipeline/src/Stage.tsx
```diff
@@ -0,0 +1,208 @@
+import React, { forwardRef, ReactNode, Ref, ReactElement } from 'react';
```
DesignerDave: Screen Shot 2019-11-18 at 4 58 34 PM

I managed to hit this issue when moving from the `xsmall` size where all stages are visible to the `small` size at which a size would be collapsed. It's my understanding that the squared-off extra stage (or so it seems) shouldn't be there. 

Not sure where exactly someone would hit this issue in practice, but to be safe it's probably worth addressing.

packages/pipeline/src/utils/isElementOverflowed.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Helper function to check whether a DOM element is overflowed.
+ * @param element the element to check
+ * @returns boolean true if the element is overflowing, false if not.
+ */
+export function isElementOverflowed(element: HTMLElement): boolean {
```
DesignerDave: We should probably just do:
```
export default function isElementOverflowed...
```

packages/pipeline/src/Pipeline.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  /**
+   * Alter the rendered size of the component
+   */
+  size: Size;
+
+  /**
+   * Alter the visual apperance of the component
+   */
+  variant: Variant;
```
DesignerDave: It looks like we don't actually support multiple variants at the moment. If that's the case as it appears it is, it would significantly clean up the component to not support the prop at all.

packages/pipeline/src/Pipeline.story.tsx
```diff
@@ -0,0 +1,51 @@
...
+    'Variant',
+    Object.values(Variant) as Array<Variant>,
+    Variant.Default,
+  );
+
+  return (
+    <div className={containerStyle}>
+      <Pipeline size={size} variant={variant}>
+        {stages.map((stage, index) => (
+          <Stage key={`${stage}-${index}`}>{stage}</Stage>
```
DesignerDave: We prefer wrapping in a Fragment rather than using `key` props.

packages/pipeline/src/Stage.tsx
DesignerDave: Realized it's somewhat difficult to hit this. You can hit this with the resizable container set to `474px`.

packages/pipeline/src/utils/getPipelineCounterTooltip.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export function getPipelineCounterTooltip(children: ReactNode): string {
+  const stages = flow(
+    onlyText,
+    split('$'),
+    compact,
+    map(str => `$${str}`),
+  )(children);
+
+  const formattedStages = flatMap(stages, (value, index, array) =>
```
DesignerDave: I think our build supports flatMap as an array method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap).

DesignerDave: If not, for tree-shaking purposes, we should import flatMap like so:
```
import flatMap from 'lodash/flatMap'
```

packages/pipeline/src/utils/getPipelineCounterTooltip.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * A utility function which takes the React.children rendered by the Pipeline component
+ * and returns the string which represents the full pipeline, for usage with the tooltip.
+ *
+ * @param children - the children rendered by the pipeline component
+ * @returns string - the tooltip text
+ */
+export function getPipelineCounterTooltip(children: ReactNode): string {
+  const stages = flow(
```
DesignerDave: Perhaps I'm missing something, so let me know if I am. Couldn't this be written as this?
```
const stages = children.map(onlyText)
```

As a note, we do tend to prefer native methods to lodash when possible for bundle size, and readability reasons.

packages/pipeline/src/utils/isStageElement.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Size, Variant } from '../styles';
+
+describe('packages/pipeline/utils/isStageElement', () => {
+  test('returns "true" if the element is a "Stage" element', () => {
+    const props = {
+      size: Size.XSmall,
+      variant: Variant.Default,
+      children: '$legitimate',
+    };
+    const element = React.createElement(Stage, props);
```
DesignerDave: You should be able to make this file `.tsx` and write this as JSX rather than using `createElement`.

packages/pipeline/src/Stage.tsx
```diff
@@ -0,0 +1,208 @@
...
+    ref: Ref<HTMLLIElement>,
+  ): ReactElement => {
+    // Effects
+    const [setRef, isVisible] = useInView({
+      threshold,
+      root: intersectionNode,
+    });
+
+    const styles = getStatefulStyles({
+      size: size || Size.XSmall,
```
DesignerDave: Rather than setting default values here, we should be setting default values in the function parameters.

```js
({size = Size.XSmall}) => {...}
```

packages/pipeline/src/Pipeline.story.tsx
matt-d-rat: This won't work due to [this code](https://github.com/mongodb/leafygreen-ui/blob/e8b91b5178f4406066a39edbd3c3a4e96b120df3/packages/pipeline/src/Pipeline.tsx#L206-L217) which transforms the child nodes into `Stage` components.


Essentially the Fragments will be converted to `Stage` components themselves, and if thet contain `Stage` components, the rendering will be incorrect.

Given the `key` prop is to simply prevent React warnings in this context, I am of the opinion that this is not a problem.

packages/pipeline/src/utils/getPipelineCounterTooltip.ts
matt-d-rat: Fair enough, I opted to use a functional programming style here, but if that is not aligned with the norm on LeafyGreen I can change.

packages/pipeline/src/Counter.tsx
```diff
@@ -0,0 +1,151 @@
...
+ * <Counter />
+ * ```
+ * @param props.className Classname applied to Counter content container.
+ */
+const Counter = forwardRef(
+  (
+    { className, children, size, variant, ...rest }: CounterProps,
+    ref: Ref<HTMLDivElement>,
+  ): ReactElement => {
+    const styles = getStatefulStyles({ size, variant });
```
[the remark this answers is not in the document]
matt-d-rat: Resolved by a962ae4

packages/pipeline/src/Pipeline.tsx
```diff
@@ -0,0 +1,275 @@
...
+    `,
+  );
+
+const getPipelineStyle = ({ size, variant }: StateForStyles): string => {
+  const { minWidth } = layout[size];
+
+  return cx(
+    getRootStyle({ size, variant }),
+    css`
+      flex: 1 1 100%;
```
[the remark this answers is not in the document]
matt-d-rat: Resolved by c677f78

packages/pipeline/src/Counter.tsx
```diff
@@ -0,0 +1,151 @@
...
+        data-leafygreen-ui="pipeline-counter"
+        className={cx(styles.base, className)}
+        ref={ref}
+      >
+        {/* Children will be the tooltip provided by the Pipeline component */}
+        {children}
+
+        <div
+          data-testid="pipeline-counter-chevron"
+          data-leafygreen-ui="pipeline-counter-chevron"
```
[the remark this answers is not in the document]
matt-d-rat: Resolved by 6aa095b

packages/pipeline/src/Counter.tsx
```diff
@@ -0,0 +1,151 @@
...
+    { className, children, size, variant, ...rest }: CounterProps,
+    ref: Ref<HTMLDivElement>,
+  ): ReactElement => {
+    const styles = getStatefulStyles({ size, variant });
+
+    return (
+      <div
+        {...rest}
+        data-testid="pipeline-counter"
+        data-leafygreen-ui="pipeline-counter"
```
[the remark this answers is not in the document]
matt-d-rat: Resolved by 6aa095b

packages/pipeline/src/utils/isElementOverflowed.ts
matt-d-rat: Resolved by 3049f71

packages/pipeline/src/utils/getPipelineCounterTooltip.ts
matt-d-rat: Resolved by ~~fc2aade~~ 52053b1

packages/pipeline/src/utils/getPipelineCounterTooltip.ts
matt-d-rat: Resolved by ~~fc2aade~~ 52053b1

packages/pipeline/src/utils/isStageElement.spec.ts
matt-d-rat: Resolved by ba8c9a9

packages/pipeline/src/Stage.tsx
matt-d-rat: Resolved by b5f3c0f

packages/pipeline/src/Pipeline.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  const boxShadow = hasHiddenStages
+    ? // With counter
+      { innerColor: uiColors.white, outerColor: secondary.backgroundColor }
+    : // No counter
+      { innerColor: 'transparent', outerColor: 'transparent' };
+
+  return css`
+    .leafygreen-ui-pipeline-stage--last-visible
+      > [data-leafygreen-ui='pipeline-stage-chevron']::before {
```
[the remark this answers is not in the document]
matt-d-rat: Resolved by 6aa095b

packages/pipeline/src/Pipeline.tsx
matt-d-rat: I would prefer the style implementation as is, as this defines the foundations for adding further variants should we need to, and helps keep the visual colours organised.

What I can do is remove the prop definition from the `propTypes` so that this cannot be changed at this point if that is more desirable?

packages/pipeline/src/Counter.tsx
```diff
@@ -0,0 +1,151 @@
...
+  variant: PropTypes.oneOf(Object.values(Variant)).isRequired,
+};
+
+Counter.defaultProps = {
+  className: '',
+};
+
+export default Counter;
+export {
+  Counter,
```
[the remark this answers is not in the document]
matt-d-rat: Resolved by a962ae4

packages/pipeline/src/Counter.tsx
```diff
@@ -0,0 +1,151 @@
...
+  layout,
+  colors,
+} from './styles';
+
+interface StateForStyles {
+  size: Size;
+  variant: Variant;
+}
+
+interface CounterProps {
```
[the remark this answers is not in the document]
matt-d-rat: Resolved by a962ae4

packages/pipeline/src/Stage.tsx
matt-d-rat: Nice catch, I wasn't firing the side-effects of the mutation observer for changes to the `childList`. You could also generate the same behaviour by removing childNodes.

I have fixed this in 92c2b15, so you shouldn't run into this anymore in either case:

pipeline-observer-fix

packages/pipeline/src/mocks/IntersectionObserver.ts
```diff
@@ -0,0 +1,91 @@
+import { act } from 'react-dom/test-utils';
+import { IntersectionOptions } from 'react-intersection-observer';
+
+const observerMap = new Map();
+const instanceMap = new Map();
+
+function IntersectionObserver(
```
hswolff: Curious why we wouldn't use a [polyfill](https://github.com/w3c/IntersectionObserver/tree/master/polyfill) for this?

packages/pipeline/src/Counter.tsx
```diff
@@ -0,0 +1,158 @@
...
+Counter.displayName = 'Counter';
+
+Counter.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  size: PropTypes.oneOf(Object.values(Size)).isRequired,
+  variant: PropTypes.oneOf(Object.values(Variant)).isRequired,
+};
+
+Counter.defaultProps = {
```
hswolff: We're trying to not use `defaultProps` as the React core team has mentioned they intend to deprecate them in the future in favor of destructuring defaults. Instead we're leveraging ES6 so we could set the default value in the function arguemnt.

Ditto for all other defaultProps usages.

matt-d-rat: Not a problem, I wasn't sure whether the team followed this pattern or not. I think it would be valuable to add this to a set of design patterns to be added to the contributing guidelines. I suggested adding this to the contributing guidelines that I submitted to the LeafyGreen team.

Have you seen this document? I can't recall if you were on the shared invites?

packages/pipeline/src/mocks/IntersectionObserver.ts
matt-d-rat: I'm sure we can use a polyfill, I am not that familiar with how Jest works / is setup to include a polyfill. I would appreciate some guidance in how to do this.

packages/pipeline/src/Counter.tsx
matt-d-rat: Resolved by ed26972

jest.config.js
```diff
@@ -46,5 +46,9 @@ module.exports = {
     '.svg': '<rootDir>/__mocks__/svgTransformer.js',
   },
 
+  // Ignore transforming node_modules except for:
+  // 1. `react-children-utilities`
+  transformIgnorePatterns: ['/node_modules/(?!(react-children-utilities)/)'],
```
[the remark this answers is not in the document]
matt-d-rat: Without it I get the following error when running the tests:

```bash
Jest encountered an unexpected token
```
<details>
  <summary>Error Details</summary>
  <p>
    
Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/matt/Documents/projects/mongodb/leafygreen-ui/node_modules/react-children-utilities/react-children-utilities.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import { Children } from 'react';
                                                                                                    ^

    SyntaxError: Unexpected token {

      1 | import React, { ReactNode } from 'react';
    > 2 | import { onlyText } from 'react-children-utilities';
        | ^
      3 | import flatMap from 'lodash/flatMap';
      4 | 
      5 | /**

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:403:17)
      at Object.<anonymous> (packages/pipeline/src/utils/getPipelineCounterTooltip.ts:2:1)
  </p>
</details>

I think the third-party package is not setup correctly with the `module` syntax. Both their `module` and `main` definitions in `package.json` point to the same `react-children-utilities.js` file, which for whatever reason is not playing nicely with the build.

Adding this Jest configuration resolves the issue.

packages/pipeline/src/mocks/IntersectionObserver.ts
hswolff: We have `scripts/jestSetup.js` which uses the [setupFiles](https://jestjs.io/docs/en/configuration#setupfiles-array) configuration of Jest for configuring the environment of tests.

jest.config.js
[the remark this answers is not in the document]
hswolff: Damn, that's an amazing error message.

packages/pipeline/src/Stage.tsx
```diff
@@ -0,0 +1,207 @@
...
+      size,
+      variant,
+    });
+
+    return (
+      <li
+        {...rest}
+        {...stageAttr.prop}
+        ref={mergeRefs([setRef, ref])}
+        data-testid="pipeline-stage"
```
hswolff: Curious if we should add something about this to our convention guidelines too.

packages/pipeline/src/Counter.tsx
hswolff: I saw the PR get merged with it? I think it's great!

And yes this would be great to add to that list. Feel free to include that in this PR if you like.

packages/pipeline/src/Stage.spec.tsx
```diff
@@ -0,0 +1,116 @@
...
+    expect(thresholds).toContain(0.5);
+    expect(thresholds).toContain(0.75);
+  });
+
+  describe('when it is NOT intersecting its parent', () => {
+    test('should set the "data-stage-visible" attribute to "false"', () => {
+      const { element } = renderStage();
+
+      mockIsIntersecting(element, false);
+      expect(element.getAttribute('data-stage-visible')).toBe('false');
```
hswolff: Generally preferable to test the actual UI that's rendered on the page but I can understand the need here.

packages/pipeline/src/Counter.tsx
```diff
@@ -0,0 +1,154 @@
...
+  Size,
+  Variant,
+  layout,
+  colors,
+  counterAttr,
+  counterChevronAttr,
+} from './styles';
+
+interface StateForStyles {
+  size: Size | undefined;
```
hswolff:
```suggestion
  size?: Size;
```

Is equivalent to what you have here. Generally preferable to use the question mark suffix. Ditto for variant field.

packages/pipeline/src/utils/isElementOverflowed.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Helper function to check whether a DOM element is overflowed.
```
hswolff: What do you mean by overflowed? It extends beyond the viewport? `isElementOverflownViewport`? Not sure I like that any better.

Also FWIW I can easily see a feature where this function and some other functions get upstreamed to the lib package for use with other components. We'll have to keep our eye out for when we need to start using them elsewhere.  Very exciting addition!

packages/pipeline/src/utils/isElementOverflowed.spec.ts
```diff
@@ -0,0 +1,62 @@
+import isElementOverflowed from './isElementOverflowed';
+
+const createMockElement = ({
```
hswolff: Lovely tests!

packages/pipeline/src/Pipeline.tsx
```diff
@@ -0,0 +1,285 @@
...
+  forwardRef,
+  createRef,
+  useState,
+  Ref,
+  ReactElement,
+  ReactNode,
+} from 'react';
+
+import PropTypes from 'prop-types';
+import { findLast } from 'lodash';
```
hswolff:
```suggestion
import findLast from 'lodash/findLast';
```

We really need to install that babel plugin so we don't have to worry about this...

packages/pipeline/src/Pipeline.tsx
```diff
@@ -0,0 +1,285 @@
...
+  Size,
+  Variant,
+  layout,
+  colors,
+  pipelineAttr,
+  pipelineStages,
+  stageChevronAttr,
+} from './styles';
+
+export interface StateForStyles {
```
hswolff: Doesn't seem like we need to export this?

packages/pipeline/src/Stage.tsx
```diff
@@ -0,0 +1,207 @@
...
+  getChevronStyle,
+  Size,
+  Variant,
+  layout,
+  colors,
+  stageAttr,
+  stageChevronAttr,
+} from './styles';
+
+export interface StateForStyles {
```
hswolff: Doesn't seem like we need to export this?

packages/pipeline/src/Pipeline.tsx
```diff
@@ -0,0 +1,285 @@
...
+  hasHiddenStages: boolean;
+  size: Size;
+  variant: Variant;
+}
+
+export interface PipelineProps {
```
hswolff: Don't need to export this either. Nothing is directly using the interface, it's just being used by the component. 

If we had another component that wanted to extend the interface or directly access it then we'd export the interface, however we don't need that right now.

packages/pipeline/src/Pipeline.tsx
```diff
@@ -0,0 +1,285 @@
...
+      pipelineNode,
+      {
+        childList: true,
+        subtree: true,
+        attributes: true,
+      },
+      observeChanges,
+    );
+
+    const _children = React.Children.map(children, child => {
```
hswolff: Is there a more semantic name for this variable?

packages/pipeline/src/index.ts
```diff
@@ -0,0 +1,8 @@
+import Pipeline from './Pipeline';
+import Stage from './Stage';
+import Counter from './Counter';
+import { Size, Variant } from './styles';
+
+export default Pipeline;
+
+export { Pipeline, Stage, Counter, Size, Variant };
```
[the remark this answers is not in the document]
matt-d-rat: Sounds reasonable. I have opted for named export instead of default as the ideal usage would be to use the `Pipeline` component with the `Stage` component as children, so you would need a named export the later anyway.

Resolved by 45a95ac.

packages/pipeline/src/Stage.tsx
```diff
@@ -0,0 +1,207 @@
...
+  size?: Size;
+
+  /**
+   * Either a single number or an array of numbers which indicate at what percentage of the target's visibility
+   * the observer's callback should be executed.
+   */
+  threshold?: number | Array<number>;
+
+  /**
+   * Alter the visual apperance of the component. Inheritted from the parent Pipeline component.
```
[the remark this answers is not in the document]
matt-d-rat: Whoops. I have installed `cspell` now to avoid my fat finger mistakes in the future, thanks for the recommendation.

Resolved by 08819d8.

packages/pipeline/src/Stage.tsx
matt-d-rat: I think that might be a good idea. 👍

packages/pipeline/src/Stage.spec.tsx
matt-d-rat: I completely agree, in this case the behaviour is difficult / I couldn't find a way to simulate in `jsdom`, so I opted for this approach over not testing at all.

packages/pipeline/src/Counter.tsx
matt-d-rat: Resolved by 275f61c.

packages/pipeline/src/Pipeline.tsx
matt-d-rat: Resolved by 022bad9

packages/pipeline/src/Stage.tsx
matt-d-rat: Resolved by 9f54a5e

packages/pipeline/src/Pipeline.tsx
matt-d-rat: Resolved by 98f6cdd

packages/pipeline/src/Pipeline.tsx
matt-d-rat: Resolved by ffa74d1

packages/pipeline/src/Pipeline.tsx
matt-d-rat: Resolved by 9f54a5e

packages/pipeline/src/utils/isElementOverflowed.ts
matt-d-rat: Yeah naming things is hard. What I mean by overflowed, is really the past tense of `overflow` – which in a CSS context means the content is clipped when an element's it is too big to fit in a specified area.

I'm not a fan of `Overflown`, specifically the flown part, as I think that would be closer to the past tense of "fly".

🤷‍♂ Not sure what else to suggest.

matt-d-rat:
@hswolff @brookescarlett @DesignerDave The build is failing for this branch due to:

```
Exit code 137 - Out of memory
```

Not sure how to resolve as I don't believe I have the appropriate permissions in CircleCI project to increase the memory of the containers. The build and tests pass when I run them locally, could you please assist me in resolving this issue? 

See for more info: https://support.circleci.com/hc/en-us/articles/115014359648-Exit-code-137-Out-of-memory

hswolff:
Try merging in master. We updated the node version we use on circleCI which seems to help with the memory usage.

packages/pipeline/src/mocks/IntersectionObserver.ts
matt-d-rat: This is related to the way in which I am testing the Stage component. I use this as a mock to test the functionality for determining the visibility of the `Stage` component relative to its parent.

<details open>
<summary>I had attempted to test this in an alternative manner:</summary>
<p>

```ts
describe('when it is NOT intersecting its parent', () => {
    test('should set the "data-stage-visible" attribute to "false"', async () => {
      const { element } = renderStage();

      // Set the starting sizes of the element under test and the parentElement container
      element.setAttribute('style', 'width: 50px; margin-left: -100px;');
      parentElement.setAttribute('style', 'width: 100px;');

      // Observe changes to the element
      const promise = waitForDomChange({
        container: element,
        mutationObserverOptions: {
          attributes: true,
          childList: false,
          subtree: false,
        },
      });

      // Trigger an intersection change on the element and await the result
      element.setAttribute('style', 'width: 100px; margin-left: 0px;');
      await promise;

      // Make assertion on element visibility attribute
      expect(element.getAttribute('data-stage-visible')).toBe('true');
    });
  });

  describe('when it is intersecting its parent', () => {
    test('should set the "data-stage-visible" attribute to "false"', async () => {
      const { element } = renderStage();

      // Set the starting sizes of the element under test and the parentElement container
      element.setAttribute('style', 'width: 50px; margin-left: 0px;');
      parentElement.setAttribute('style', 'width: 100px;');

      // Observe changes to the element
      const promise = waitForDomChange({
        container: element,
        mutationObserverOptions: {
          attributes: true,
          childList: false,
          subtree: false,
        },
      });

      // Trigger an intersection change on the element and await the result
      element.setAttribute('style', 'width: 50px; margin-left: -100px;');
      await promise;

      // Make assertion on element visibility attribute
      expect(element.getAttribute('data-stage-visible')).toBe('false');
    });
  });
```

</p>
</details>

But I get a test failure no matter what I do, so something is clearly not write with the way in which this test is written:

```
packages/pipeline/Stage › when it is NOT intersecting its parent › should set the "data-stage-visible" attribute to "false"

Expected: "true"
Received: "false"
```

If this is a blocker for merging, then I would really appreciate some help as I am all out of ideas how to best solve this.

matt-d-rat:
> Try merging in master. We updated the node version we use on circleCI which seems to help with the memory usage.

This branch is up-to-date with current master. The build continues to fail with the same error.

packages/pipeline/src/mocks/IntersectionObserver.ts
scottsidwell: Adding some context on @matt-d-rat behalf here^:

Jest is using jsdom to configure a browser-like environment in Node, however it [doesn't actually support layout](https://github.com/jsdom/jsdom/blob/master/README.md#unimplemented-parts-of-the-web-platform) or methods like `getBoundingClientRects()` which the polyfill you linked uses heavily.

As a result, even with the polyfill, code that tests behaviour that relies on the IntersectionObserver just won't work.

The alternatives that remain would be to either:
* run the tests using something like jest-puppeteer
* mock out `getBoundingClientRects()` and any other layout/size methods used by the IntersectionObserver
* mock out the IntersectionObserver, which Matt did here

matt-d-rat: Thanks @scott I was having trouble wording 🤣

hswolff: Ah, didn't know that the polyfill relied on `getBoundingClientRects`. That makes complete sense.

I'm fine with either mocking out `getBoundingClientRects` or using the mock that @matt-d-rat made, up to ya'll which you prefer.

If we go with the current mock can we move it somewhere more general so we can re-use it across other packages if needed? We have `scripts/jestSetup.js` which has one mock created, so ideally we can build on top of that (make a folder for this file, inline the file, up to you).

packages/pipeline/src/utils/isElementOverflowed.ts
hswolff: Yeah let's keep the name, but let's inline this comment description as it's gold jerry, gold.

hswolff:
Ugh.... yeah....I wonder if we added a new `build:ci` command that turns off parallel that would help? Mind trying that?

brookescarlett:
Also, don't forget to run `yarn changeset` documenting your changes before merging :)

packages/pipeline/src/mocks/IntersectionObserver.ts
matt-d-rat: I'll stick with the mock I have implemented already, and attempt to move it to a re-usable location.

packages/pipeline/src/utils/isElementOverflowed.ts
matt-d-rat: Resolved by b383cf5

packages/pipeline/src/mocks/IntersectionObserver.ts
matt-d-rat: I attempted to move the file to a re-usable location, but when I attempt to build the project I get the following error:

```bash
$ tsc --build tsconfig.json
packages/pipeline/src/Stage.spec.tsx:2:36 - error TS6059: File '/Users/matt/Documents/projects/mongodb/leafygreen-ui/__mocks__/IntersectionObserver.ts' is not under 'rootDir' '/Users/matt/Documents/projects/mongodb/leafygreen-ui/packages/pipeline'. 'rootDir' is expected to contain all source files.

2 import { mockIsIntersecting } from '../../../__mocks__/IntersectionObserver';
```

I also tried a variety of jest config options as well (`moduleDirectories`,  `modulePaths`), which complained that the file could not be found when importing. If someone with more jest experience than me could take a look at this then I would appreciate it.

The relevant code that needs moving is located under `packages/pipeline/mocks`

matt-d-rat:
> Ugh.... yeah....I wonder if we added a new `build:ci` command that turns off parallel that would help? Mind trying that?

I updated the CircleCI `.config.yaml` in 79c7d43 to turn off the `--parallel` flag.

.gitignore
```diff
@@ -18,3 +18,6 @@ reports
 # Yarn Error Log
 yarn-error.log
 tsconfig.tsbuildinfo
+
+# VSCode config
+.vscode/
```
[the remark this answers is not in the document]
matt-d-rat: I don't have an opinion on this.
Removed addition to `.gitignore` in 0cf5485.

packages/pipeline/src/Pipeline.tsx
matt-d-rat: > In this case we have quite a lot of code that could be eliminated without the prop, and AFAIK we don't have any other components that has a variant prop when there's only a single variant.
> 
> While I know this makes adding variants easier in the future, is there any variant that's currently planned to be added right now? If not, I feel that this creates unnecessary overhead when reading the code, and should be removed.

This is starting to feel like bikeshedding. Resolved by b6c1bc0.

packages/pipeline/src/Pipeline.tsx
```diff
@@ -0,0 +1,285 @@
...
+      overflow: hidden;
+      list-style: none;
+      padding: 0;
+      margin: 0;
+      min-width: ${minWidth}px;
+    `,
+  );
+};
+
+export const lastVisibleClassName =
```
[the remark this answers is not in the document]
matt-d-rat: The `:last-of-type` selector allows you to target the last occurrence of an element within its container. This will not work for this use case because what I actually need to find is the last *visible* occurrence of an element within its container (ie: the last element before the parent container clips the others when content overflows).

If I use the `createDataProp` it will clash with another usage of the `createDataProp` on this element (because the `createDataProp` doesn't allow me to set the key of the resulting attribute, only the value). As far as I am aware, `data-` attributes should have unique names, which would make this not an option?

packages/pipeline/src/mocks/IntersectionObserver.ts
hswolff: Changed my mind - let's worry about moving this file when we have a need for it. I keep running out of time to look into it.

matt-d-rat:
Thank you for the approval @hswolff . If I could get an approval from @brookescarlett or @DesignerDave that would be great.

Just to confirm the process, I would:

1. Run `yarn changeset` (which I assume generates some sort of change log?)
2. Merge to `master`

After which I assume someone in the LeafyGreen team will handle the release to npm?

matt-d-rat:
> LGTM! It would be great if we could get a Sketch symbol for this from Dave Hastwell as well, so it's usable outside of Charts 😁

cc @dhastwell

I believe one does exist, though he may need to adjust the height of the symbol to correspond to the changes in this PR (I aligned the height of the Pipeline to match the heights of the LeafyGreen buttons at all sizes). I can work with him on that.

dhastwell:
@DesignerDave @matt-d-rat After I make those adjustments, I will happily add this component to our LeafyGreen Sketch library.

PR #198, opened 2019-11-19
PD-454: Tooltip Trigger Accepts Nested Components

brookescarlett:
## ✍️ Proposed changes

- Patch fix to Tooltip, to allow trigger to accept nested components
- Added test to prove that fix works

🎟 _Jira ticket:_ [PD-454](https://jira.mongodb.org/browse/PD-454)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/tooltip/src/Tooltip.tsx
+++ b/packages/tooltip/src/Tooltip.tsx
@@ -274,7 +274,7 @@ function Tooltip({
       });
     }
 
-    const triggerChildren = trigger.props.children;
+    const { children: triggerChildren } = trigger.props;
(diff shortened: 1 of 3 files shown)


packages/tooltip/src/Tooltip.tsx
```diff
@@ -274,11 +274,17 @@ function Tooltip({
       });
     }
 
+    const triggerChildren = trigger.props.children;
```
DesignerDave: Opinion on 
```js
const triggerChildren = trigger.props.children;
```
vs
```js
const { children: triggerChildren } = trigger.props
```
?

PR #201, opened 2019-11-21
PD-457: Allow target/rel props to be passed to MenuItem

brookescarlett:
## ✍️ Proposed changes

- Make MenuItem props extend HTML element based on whether it's being rendered as a button or an anchor tag
- Allow consumers to customize `target` and `rel` props
- Added test to prove that fix works

🎟 _Jira ticket:_ [PD-457](https://jira.mongodb.org/browse/PD-457)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/menu/src/Menu.spec.js
+++ b/packages/menu/src/Menu.spec.js
@@ -71,7 +71,12 @@ describe('packages/Menu', () => {
         >
           Item 1
         </MenuItem>
-        <MenuItem href="http://mongodb.design" data-testid="second-item">
+        <MenuItem
+          href="http://mongodb.design"
+          data-testid="second-item"
+          target="_self"
+          rel=""
+        >
           Item 2
         </MenuItem>
       </div>,
@@ -91,6 +96,11 @@ describe('packages/Menu', () => {
 
     test('renders inside of an `a` instead of a `button` tag, when `href` prop is supplied', () => {
       expect(secondItem.tagName.toLowerCase()).toBe('a');
+    });
+
+    test('renders with correct target and rel values when set', () => {
+      expect(secondItem.target).toBe('_self');
+      expect(secondItem.rel).toBe('');
     });
   });
 });
--- a/packages/menu/src/Menu.spec.js
+++ b/packages/menu/src/Menu.spec.js
@@ -75,7 +75,7 @@ describe('packages/Menu', () => {
           href="http://mongodb.design"
           data-testid="second-item"
           target="_self"
-          rel=""
+          rel="help"
         >
           Item 2
         </MenuItem>
@@ -100,7 +100,7 @@ describe('packages/Menu', () => {
 
     test('renders with correct target and rel values when set', () => {
       expect(secondItem.target).toBe('_self');
-      expect(secondItem.rel).toBe('');
+      expect(secondItem.rel).toBe('help');
     });
   });
 });
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -1,8 +1,8 @@
 import React, { RefObject } from 'react';
 import PropTypes from 'prop-types';
+import { HTMLElementProps, createDataProp } from '@leafygreen-ui/lib';
 import { css, cx } from '@leafygreen-ui/emotion';
 import { uiColors } from '@leafygreen-ui/palette';
-import { createDataProp } from '@leafygreen-ui/lib';
 
 const menuItemContainer = createDataProp('menu-item-container');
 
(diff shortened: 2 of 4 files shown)


packages/menu/src/MenuItem.tsx
```diff
@@ -255,6 +257,12 @@ const MenuItem = React.forwardRef(
         </Root>
       </li>
     );
+
+    if (usesLinkElement({ ...rest })) {
```
DesignerDave: This is the same as `usesLinkElement(rest)` I believe.

packages/menu/src/MenuItem.tsx
```diff
@@ -137,43 +137,49 @@ const disbaledTextStyle = css`
...
+    SharedMenuItemProps {
+  href?: null;
+}
+
+type MenuItemProps = LinkMenuItemProps | ButtonMenuItemProps;
+
+function usesLinkElement(
+  props: LinkMenuItemProps | ButtonMenuItemProps,
+): props is LinkMenuItemProps {
+  return props.href != null;
```
DesignerDave: I know we've talked about whether we should write `value != null` or `value !== undefined && value !== null`...

@hswolff @brookescarlett Do we want to take a stance on this and prefer `value != null` in the future? In the past, I think Harry and I spoke about this and preferred the more verbose option in the past for readability, but I'd personally be fine with this instead.

packages/menu/src/MenuItem.tsx
DesignerDave: I'm also not convinced this is worth an entire function to check `href != null`...

DesignerDave: If we got rid of that function, we could destructure href from props, and write the declaration for `anchorProps` above like so:
```js
const anchorProps = href && {
  target: `_blank`,
  rel: 'noopener noreferrer',
  href,
}
```

packages/menu/src/MenuItem.tsx
```diff
@@ -205,19 +210,17 @@ const MenuItem = React.forwardRef(
     }: MenuItemProps,
     forwardedref,
   ) => {
-    const Root = href ? 'a' : 'button';
-
-    const anchorProps = href && {
+    const anchorProps = rest.href && {
       target: '_blank',
```
DesignerDave: While we're in here, I'm not sure it's right to indiscriminately set `target='_blank'`. We may want to make the default either open in the same tab, or by default parse the URL to see if this should be set.

I'm fine with leaving it for now if we just wanna make a ticket, but I wanted to note this while it's fresh.

packages/menu/src/MenuItem.tsx
brookescarlett: Leaving as-is based on our conversation

packages/menu/src/MenuItem.tsx
brookescarlett: Personally support the less verbose option, but will move forward based on whatever you think.

packages/menu/src/MenuItem.tsx
brookescarlett: Outside the scope of this PR but we should have this discussion and decide what we want the default behavior here to be.

packages/menu/src/MenuItem.tsx
seanmhanson: FWIW at least on MMS we don't have a standard. If anything, I run into `value != null` more often, and we have gone as far as configuring ESLint to allow that as an exception to `eqeqeq`.

packages/menu/src/MenuItem.tsx
seanmhanson: Unsolicited opinion as a user: I would expect the default target to be set to `_self` with `rel` omitted, since for the majority of navigation I write, I stay within our application. 

From my usage, the most major exception would be usage in `mongo-menu`. That component is shared, though, so as a user I'd just expect `mongo-menu` to deal with that on my behalf.

packages/menu/src/Menu.spec.js
```diff
@@ -71,7 +71,12 @@ describe('packages/Menu', () => {
         >
           Item 1
         </MenuItem>
-        <MenuItem href="http://mongodb.design" data-testid="second-item">
+        <MenuItem
+          href="http://mongodb.design"
+          data-testid="second-item"
+          target="_self"
+          rel=""
```
hswolff: Should probably check that an actual value is carried through. Empty string may be a false positive.

packages/menu/src/MenuItem.tsx
hswolff: I'm completely fine with `!= null` shortcut.

brookescarlett: 💯

packages/menu/src/MenuItem.tsx
brookescarlett: Think you're right about that @seanmhanson: https://jira.mongodb.org/browse/PD-459

packages/menu/src/Menu.spec.js
brookescarlett: Updated!

PR #204, opened 2019-11-21
PD-461: Update anchor styles for Menu and MongoMenu components

seanmhanson:
## ✍️ Proposed changes

This updates the styles specifically for focus/active states for anchors in Menu and MongoMenu with rules that are already inherited but need to be specifically included in order to avoid third party libraries from overwriting these with values set on `a` tags specifically. This was found during [library upgrades for the quarter](https://jira.mongodb.org/browse/CLOUDP-49387), and we figured we needed to address this soon anyway for the work we're doing on the Account app.

🎟 _Jira ticket:_ [Update MongoMenu and MenuItem styles for Cloud usage](https://jira.mongodb.org/browse/PD-461)

## 🛠 Types of changes

## ✅ Checklist

### Menu (before)
Screen Shot 2019-11-21 at 2 08 43 PM
### Menu (current)
Screen Shot 2019-11-21 at 2 10 05 PM

### MongoMenu (before)
Screen Shot 2019-11-21 at 2 08 38 PM
### MongoMenu (current)
Screen Shot 2019-11-21 at 2 10 11 PM

### DIFF
--- a/packages/mongo-menu/src/MongoMenu.tsx
+++ b/packages/mongo-menu/src/MongoMenu.tsx
@@ -134,12 +134,6 @@ const accountButtonStyle = css`
   justify-content: center;
   align-items: center;
   margin-bottom: 14px;
-
-  &:focus,
-  &:hover {
-    text-decoration: none;
-    color: ${uiColors.gray.dark2};
-  }
 `;
 
 const descriptionStyle = css`
--- a/packages/mongo-menu/src/MongoMenu.tsx
+++ b/packages/mongo-menu/src/MongoMenu.tsx
@@ -26,6 +26,7 @@ const buttonReset = css`
   }
 
   &:active {
+    outline: none;
     color: ${uiColors.gray.dark2};
 
     &:before {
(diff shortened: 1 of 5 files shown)


packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -133,6 +134,12 @@ const accountButtonStyle = css`
   justify-content: center;
   align-items: center;
   margin-bottom: 14px;
+
+  &:focus,
+  &:hover {
+    text-decoration: none;
```
brookescarlett: I think it might make more sense to apply this rule to the button component rather than here, because I imagine other LG buttons are having the same problem

brookescarlett:
hi @seanmhanson 👋, thanks so much for the PR. Left one piece of feedback, but otherwise LGTM!

packages/mongo-menu/src/MongoMenu.tsx
seanmhanson: Good catch. I checked and it looks like it is in fact impacting all buttons we use that are anchors (however, that's virtually none, hence this not showing up before).

brookescarlett:
Make sure you run Changest for the new Button updates before merging, but otherwise LGTM :D Thank you!

PR #206, opened 2019-11-25
PD-426: Upgrade Typescript to 3.7

brookescarlett:
## ✍️ Proposed changes

- Uprgrade Typescript and associated deps to 3.7
- Added babel plugins to support optional chaining and nullish coalescing 

🎟 _Jira ticket:_ [PD-426](https://jira.mongodb.org/browse/PD-426)


packages/icon-button/src/IconButton.tsx
```diff
@@ -57,6 +57,12 @@ interface ButtonIconButtonProps
 
 type IconButtonProps = LinkIconButtonProps | ButtonIconButtonProps;
 
+function usesLinkElement(
```
hswolff: What's with these changes?

brookescarlett: Typescript started to complain about onClick handling. Have encountered this before with a component dynamically switching between 'a' and 'button' tags so just followed pattern that we've implemented in other places.

brookescarlett: Should run a changeset for this component, great catch.

packages/icon-button/src/IconButton.tsx
```diff
@@ -129,10 +135,12 @@ const iconButtonVariants: { readonly [K in Variant]: string } = {
 const disabledStyle: { readonly [K in Variant]: string } = {
   [Variant.Light]: css`
     color: ${uiColors.gray.light2};
+    pointer-events: none;
```
hswolff: What about these changes?

brookescarlett: Handles disabling pointer events when we're no longer conditionally implementing the onClick

PR #211, opened 2019-12-02
Clean Storybook

brookescarlett:
- Remove Variants from sidebar of storybook, in favor of using `select knobs` to navigate between variant-options
- Not running changesets here, as this will not effect consuming applications
Screen Shot 2019-12-02 at 2 12 17 PM


DesignerDave:
Can we also rename instances of `Default`? Right now we either say `Default` or the component's name.

brookescarlett:
I actually think in most cases Default makes more sense than the component name, because where we are using multiple sub-tabs, it doesn't make sense to have one that says the component's name and one that says some variant of the component (IMO)

brookescarlett:
Happy to standardize either way, but let me know if you think that makes sense @DesignerDave

hswolff:
Using the new storybook standard format for storybook files may be a better time to do a rename as it has more flexibility I believe.

PR #212, opened 2019-12-03
PD-443: Trap Focus in Modal Component

brookescarlett:
## ✍️ Proposed changes

- Bringing in`react-trap-focus` library to help trap focus in our Modal component
- Bringing in this library, rather than handling manually, because there are many necessary steps to take in order to ensure that focus is properly trapped. Which includes creating a list of focusable children, updating tabIndexes, and keeping track of what element is focused such that focus can be properly updated on `Tab` or `Shift+Tab`
- Add tests to show that focus is properly trapped 

🎟 _Jira ticket:_ [PD-443](https://jira.mongodb.org/browse/PD-443)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/.changeset/hot-geckos-guess.md
+++ b/.changeset/hot-geckos-guess.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/modal': major
+---
+
+Use `react-trap-focus` to handle trapping focus in component
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -14,9 +14,11 @@
   "dependencies": {
     "@leafygreen-ui/hooks": "^1.1.0",
     "@leafygreen-ui/icon": "^2.1.0",
+    "@leafygreen-ui/icon-button": "2.0.0",
     "@leafygreen-ui/lib": "^3.2.0",
     "@leafygreen-ui/palette": "^1.1.1",
     "@leafygreen-ui/portal": "^1.1.8",
-    "react-transition-group": "^4.2.1"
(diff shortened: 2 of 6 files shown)


.changeset/hot-geckos-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/modal': major
```
hswolff: Why major? Isn't this additional functionality? What functionality does this break?

packages/modal/package.json
```diff
@@ -14,9 +14,11 @@
   "dependencies": {
     "@leafygreen-ui/hooks": "^1.1.0",
     "@leafygreen-ui/icon": "^2.1.0",
+    "@leafygreen-ui/icon-button": "2.0.0",
     "@leafygreen-ui/lib": "^3.2.0",
     "@leafygreen-ui/palette": "^1.1.1",
     "@leafygreen-ui/portal": "^1.1.8",
-    "react-transition-group": "^4.2.1"
+    "react-transition-group": "^4.2.1", 
+    "react-focus-lock": "^2.2.1"
```
hswolff: I don't think it's a good idea to use this library:

- It doesn't seem like it's maintained anymore, last update was over a year ago.
- It hasn't been updated to the latest React APIs, meaning it'll have deprecation warnings in the console.

I'd be more comfortable to port it into leafygreen (either as a child inside modal or as its own package) and update it to reflect current best React practices. This should also include tests for this.

PR #215, opened 2019-12-07
PD-473: Disable MongoMenu account button when account url is empty

seanmhanson:
## ✍️ Proposed changes

The account app needs to use MongoMenu, but display the link to the account app as disabled (because the user is already there). Rather than add more props for a one-time flag, I opted to add support for displaying as a disabled button when the accountURL is purposefully set to be empty/null. Related stories and tests have been updated.

In the process, I found a typescript bug in MenuItem that impacted MongoMenu (rest params were not part of the types accepted by MenuItem). I've fixed that accordingly.

🎟 _Jira ticket:_ [PD-473](https://jira.mongodb.org/browse/PD-473)

## 🛠 Types of changes

## ✅ Checklist

### Screenshot (when account is disabled)
Screen Shot 2019-12-07 at 2 39 26 PM

### DIFF
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -174,11 +174,11 @@ interface ButtonMenuItemProps
   href?: null;
 }
 
-type MenuItemProps = (LinkMenuItemProps | ButtonMenuItemProps) & {
-  [key: string]: any;
-};
+type MenuItemProps = LinkMenuItemProps | ButtonMenuItemProps;
 
-function usesLinkElement(props: MenuItemProps): props is LinkMenuItemProps {
+function usesLinkElement(
+  props: LinkMenuItemProps | ButtonMenuItemProps,
+): props is LinkMenuItemProps {
   return props.href != null;
 }
 
@@ -204,7 +204,7 @@ const MenuItem = React.forwardRef(
       <li role="none">
         <Root
           {...anchorProps}
-          {...(rest as HTMLElementProps<any>)}
+          {...rest}
           {...menuItemContainer.prop}
           className={cx(
             containerStyle,
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -174,11 +174,11 @@ interface ButtonMenuItemProps
   href?: null;
 }
 
-type MenuItemProps = LinkMenuItemProps | ButtonMenuItemProps;
+type MenuItemProps = (LinkMenuItemProps | ButtonMenuItemProps) & {
+  [key: string]: any;
+};
 
-function usesLinkElement(
-  props: LinkMenuItemProps | ButtonMenuItemProps,
-): props is LinkMenuItemProps {
+function usesLinkElement(props: MenuItemProps): props is LinkMenuItemProps {
   return props.href != null;
 }
 
@@ -204,7 +204,7 @@ const MenuItem = React.forwardRef(
       <li role="none">
         <Root
           {...anchorProps}
-          {...rest}
+          {...(rest as HTMLElementProps<any>)}
           {...menuItemContainer.prop}
           className={cx(
             containerStyle,
--- a/packages/mongo-menu/src/MongoMenu.spec.tsx
+++ b/packages/mongo-menu/src/MongoMenu.spec.tsx
@@ -1,18 +1,19 @@
 import React from 'react';
 import 'jest-dom/extend-expect';
 import { render, fireEvent, cleanup } from '@testing-library/react';
-import MongoMenu, { Product } from '.';
+import MongoMenu from '.';
 
 afterAll(cleanup);
 
 describe('packages/MongoMenu', () => {
+  const user = { name: 'Leafy', email: 'leafy@mongodb.com' };
   const onLogout = jest.fn();
   const onProductChange = jest.fn();
 
   const renderedComponent = render(
     <MongoMenu
-      user={{ name: 'Leafy', email: 'leafy@mongodb.com' }}
-      activeProduct={Product.Atlas}
+      user={user}
+      activeProduct={'atlas'}
       onLogout={onLogout}
       onProductChange={onProductChange}
     />,
@@ -22,32 +23,28 @@ describe('packages/MongoMenu', () => {
   const trigger = getByText('Leafy');
   fireEvent.click(trigger);
 
+  const assertHrefByText = (expectedText: string, expectedHref: string) => {
+    const { href } = getByText(expectedText).parentElement as HTMLAnchorElement;
+    expect(href).toBe(expectedHref);
+  };
+
   test('Account button links to the default account app', () => {
-    const accountItem = getByText('MongoDB Account').parentElement;
-    expect((accountItem as HTMLAnchorElement).href).toBe(
+    assertHrefByText(
+      'MongoDB Account',
       'https://cloud.mongodb.com/v2#/account',
     );
   });
 
   test('Atlas menu item links to cloud.mongodb.com', () => {
-    const atlasItem = getByText('Atlas').parentElement;
-    expect((atlasItem as HTMLAnchorElement).href).toBe(
-      'https://cloud.mongodb.com/',
-    );
+    assertHrefByText('Atlas', 'https://cloud.mongodb.com/');
   });
 
   test('University menu item links to university.mongodb.com', () => {
-    const universityItem = getByText('University').parentElement;
-    expect((universityItem as HTMLAnchorElement).href).toBe(
-      'https://university.mongodb.com/',
-    );
+    assertHrefByText('University', 'https://university.mongodb.com/');
   });
 
   test('Support menu item links to support.mongodb.com', () => {
-    const supportItem = getByText('Cloud Support').parentElement;
-    expect((supportItem as HTMLAnchorElement).href).toBe(
-      'https://support.mongodb.com/',
-    );
+    assertHrefByText('Cloud Support', 'https://support.mongodb.com/');
   });
 
   test('onProductChange fires when a product is clicked', () => {
@@ -67,21 +64,11 @@ describe('packages/MongoMenu', () => {
   });
 
   test('renders the account link as a disabled button when set to the empty string', () => {
-    renderedComponent.rerender(
-      <MongoMenu
-        user={{ name: 'Leafy', email: 'leafy@mongodb.com' }}
-        activeProduct={Product.Atlas}
-        onLogout={onLogout}
-        onProductChange={onProductChange}
-        accountURL={''}
-      />,
-    );
+    renderedComponent.rerender(<MongoMenu user={user} accountURL={''} />);
 
-    const accountItem = getByText('MongoDB Account').parentElement;
-    const accountButton = accountItem as HTMLButtonElement;
-    expect((accountButton as HTMLButtonElement).tagName.toLowerCase()).toBe(
-      'button',
-    );
+    const accountButton = getByText('MongoDB Account')
+      .parentElement as HTMLButtonElement;
+    expect(accountButton.tagName.toLowerCase()).toBe('button');
     expect(accountButton.disabled).toBe(true);
     expect(accountButton.getAttribute('aria-disabled')).toBe('true');
   });
--- a/packages/mongo-menu/src/MongoMenu.spec.tsx
+++ b/packages/mongo-menu/src/MongoMenu.spec.tsx
@@ -9,7 +9,7 @@ describe('packages/MongoMenu', () => {
   const onLogout = jest.fn();
   const onProductChange = jest.fn();
 
-  const { getByText } = render(
+  const renderedComponent = render(
     <MongoMenu
       user={{ name: 'Leafy', email: 'leafy@mongodb.com' }}
       activeProduct={Product.Atlas}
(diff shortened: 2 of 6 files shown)


packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -168,14 +168,16 @@ const menuItems = [
...
-export const Product = {
+const Product = {
   Atlas: 'atlas',
   University: 'university',
   Support: 'support',
 } as const;
 
 type Product = typeof Product[keyof typeof Product];
 
+export { Product };
```
seanmhanson: Likewise, this fixes a small typescript error in the stories for MongoMenu: 
(1) the array of products was typed erroneously as `Array<Size>` instead of `Array<Product>` in the stories
(2) we exported Product as data rather than as a type (I checked this against several other packages to ensure exporting the type was the intent)

packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -255,9 +261,10 @@ function MongoMenu({
         <FocusableMenuItem>
           <Button
             size="small"
-            href={accountURL}
+            href={accountURL || undefined}
```
seanmhanson: (by explicitly setting this as undefined, `href` never gets provided to `Button` and we avoid a little console noise)

packages/mongo-menu/src/MongoMenu.spec.tsx
```diff
@@ -57,4 +65,24 @@ describe('packages/MongoMenu', () => {
 
     expect(onLogout).toHaveBeenCalledTimes(1);
   });
+
+  test('renders the account link as a disabled button when set to the empty string', () => {
+    renderedComponent.rerender(
```
seanmhanson: I've never used Jest or this testing library before, so **please** feel free to correct any style things I may have done incorrectly here.

packages/mongo-menu/src/MongoMenu.story.tsx
```diff
@@ -3,16 +3,19 @@ import { storiesOf } from '@storybook/react';
...
 
-storiesOf('MongoMenu', module).add('Default', () => (
-  <MongoMenu
-    user={{
-      name: text('name', 'Alex Smith'),
-      email: text('email', 'alex.smith@mongodb.com'),
-    }}
-    activeProduct={select(
-      'activeProduct',
-      Object.values(Product) as Array<Size>,
```
seanmhanson: (`Size` here was undefined, but is `Product` now)

packages/mongo-menu/src/MongoMenu.spec.tsx
```diff
@@ -18,9 +18,17 @@ describe('packages/MongoMenu', () => {
     />,
   );
 
+  const { getByText } = renderedComponent;
```
DesignerDave: We're trying to move away from `getByText` to rendering the component passing a `data-testid` and using `getByTestId`. `getByText` tends to be overly brittle in most cases.

packages/mongo-menu/src/MongoMenu.spec.tsx
```diff
@@ -18,9 +18,17 @@ describe('packages/MongoMenu', () => {
...
   const trigger = getByText('Leafy');
   fireEvent.click(trigger);
 
+  test('Account button links to the default account app', () => {
+    const accountItem = getByText('MongoDB Account').parentElement;
+    expect((accountItem as HTMLAnchorElement).href).toBe(
```
DesignerDave: I might be missing something, but is there a reason we're explicitly typing accountItem as `HTMLAnchorElement`? If so, might it be better to do it on the line above like so?

```ts
const accountItem: HTMLAnchorElement = getByText('MongoDB Account').parentElement;
```

packages/mongo-menu/src/MongoMenu.spec.tsx
```diff
@@ -57,4 +65,24 @@ describe('packages/MongoMenu', () => {
...
+        user={{ name: 'Leafy', email: 'leafy@mongodb.com' }}
+        activeProduct={Product.Atlas}
+        onLogout={onLogout}
+        onProductChange={onProductChange}
+        accountURL={''}
+      />,
+    );
+
+    const accountItem = getByText('MongoDB Account').parentElement;
+    const accountButton = accountItem as HTMLButtonElement;
```
DesignerDave: You're typing accountItem twice here. We should be able to rely on:
```ts
    const accountItem: HTMLButtonElement = getByText('MongoDB Account').parentElement;
```

Then we can remove the definition of `accountButton` altogether, and the inline typing within the assertion below.

packages/mongo-menu/src/MongoMenu.spec.tsx
```diff
@@ -57,4 +65,24 @@ describe('packages/MongoMenu', () => {
 
     expect(onLogout).toHaveBeenCalledTimes(1);
   });
+
+  test('renders the account link as a disabled button when set to the empty string', () => {
+    renderedComponent.rerender(
+      <MongoMenu
+        user={{ name: 'Leafy', email: 'leafy@mongodb.com' }}
+        activeProduct={Product.Atlas}
```
DesignerDave: I don't expect people to directly use the enum in most cases, so a better way to test this might be to explicitly write the string out here (just `'atlas'` I believe). That way, this test will break if we make a breaking API change.

DesignerDave: Additionally (and correct me if I'm wrong), but you're setting props here that aren't required or being tested at this point. You *should* be fine omitting `onLogout`, and `onProjectChange` at the very least.

packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -255,9 +261,10 @@ function MongoMenu({
...
             className={accountButtonStyle}
-            as="a"
+            as={accountURL ? 'a' : 'button'}
```
DesignerDave: The Button component *should* handle disabled states when rendered as a link. I think this change is unnecessary.

packages/mongo-menu/src/MongoMenu.spec.tsx
seanmhanson: This block was more or less copy-pasted and just moved, but I can definitely make these changes.

seanmhanson: Both of these were already provided in the tests prior to my changes, but I'm happy to clean them up now.

packages/mongo-menu/src/MongoMenu.spec.tsx
seanmhanson: That isn't helpful here, as we're reaching pretty far into the component and examining specific links. The usage of `getByText` was actually already in this test and I just pulled it out here instead so I'd be able to use `renderedComponent.rerender()` more easily.

packages/mongo-menu/src/MongoMenu.spec.tsx
seanmhanson: `getByText('MongoDB Account')` will always return a value (or throw an error), but the call to `parentElement` can be null, and otherwise returns an `HTMLElement`

You cannot set `accountItem` to `HTMLAnchorElement` for two reasons:
(a) it could be null, which is fine, we can get around that by just adding `| null` or something similar
(b) HTMLElement is missing all of the unique properties that are expected on `HTMLAnchorElement`, which we cannot take care of easily.

As a result, we have to explicitly cast. Presumably I can cast earlier though:
```
const accountItem = getByText('this test').parentElement as HTMLAnchorElement | null;
expect(accountItem.href).toBe('wow this is not mocha/chai');
```
However to be honest, I did this to keep with the pattern already established in this test.

packages/mongo-menu/src/MongoMenu.spec.tsx
seanmhanson: Yeah this is no longer relevant anyway, since this will be an `HTMLAnchorElement` as per your feedback on the component itself. However that does mean that unfortunately we can't test for `accountItem.disabled` since `disabled` is not a property on `HTMLAnchorElement`, without doing some more digging into typing. Aria props, however, will still be testable.

packages/mongo-menu/src/MongoMenu.spec.tsx
seanmhanson: All of these were how the test was written before I jumped in, though I can definitely leaf off `onLogout` and `onProjectChange` when I rerender here. I have no problem just replacing the usage of `Product` here.

packages/mongo-menu/src/MongoMenu.tsx
seanmhanson: We can change this back to an `a` tag if you insist, but this isn't updated because of my concern over the component, but rather clarity and accessibility. Justin brought this up way back when web chapter was just being born (or maybe before?), but I was taught by him and other sources:
- **only** use an `a` tag if you are going to set an `href` and not modify any behavior
- if you have no `href` tag, if you interrupt the call with a JS handler, if you directly manipulate `window.location`, etc, then you should **always** use a `button` tag instead, so that no user (especially on a screen reader) experiences an unexpected behavior.

This is a picky opinion I follow pretty religiously, but if that goes against the accessibility / code style / etc standards for this repo, I have no problem rolling things back to always be an `a`.

(aside: as a bonus if it is a button—we can very easily test the `disabled` prop in the unit tests)

seanmhanson:
@DesignerDave 
> As a more meta question, when might the accountURL not be set to any value?

Currently I can anticipate the following apps using this component:
- Account
- Cloud/Atlas
- Support
- University (to be added to the account app)
- some sort of JIRA portal (this was spitballed for the account app but seems to have been removed)
- Charts (this will be integrated during IA though, so maybe not)
- Stitch (this will be integrated during IA though, so maybe not)

Theoretically these could be modified too, but are unlikely to since they're tied more to mongodb itself rather than cloud;
- Compass
- Free Monitoring

Of all of these, it is more than reasonable to expect that the `accountURL` prop won't be provided, and we'll rely on the default value (which ultimately should be `account.mongodb.com`. It is possible that these may be implemented using the centralURL (I did that for Cloud) so that we can test/qa the relevant Account app.

Account is the only situation I can imagine where you would explicitly set the URL to be `''` or `null` instead of just letting it go `undefined` and defaulting to the URL in the component. That's why I thought this would be the least disruptive change.

seanmhanson:
I've updated the test with the feedback I didn't have questions about. I also made `activeProduct` an optional prop/type, as the account app should not show any product as active. RFAL.

packages/mongo-menu/src/MongoMenu.spec.tsx
DesignerDave: Ahh, you're right 👍

packages/mongo-menu/src/MongoMenu.tsx
DesignerDave: @brookescarlett Might be able to make a better call of what we should do here than I in this case.

brookescarlett: @seanmhanson's logic makes sense to me

DesignerDave: Sounds good, seems fine then.

packages/mongo-menu/src/MongoMenu.tsx
```diff
@@ -212,18 +216,20 @@ interface MongoMenuProps {
     activeProduct="atlas"
     onLogout={() => console.log('On logout')}
     onProductChange={() => console.log('Switching products')}
+    accountUrl="https://cloud.mongodb.com/account/profile"
```
brookescarlett: Thank you for updating!

packages/menu/src/MenuItem.tsx
```diff
@@ -172,11 +172,11 @@ interface ButtonMenuItemProps
   href?: null;
 }
 
-type MenuItemProps = LinkMenuItemProps | ButtonMenuItemProps;
+type MenuItemProps = (LinkMenuItemProps | ButtonMenuItemProps) & {
+  [key: string]: any;
```
brookescarlett: What prop are we passing here that's not being caught by Link/ButtonMenuItemProps? Open to handling as above if necessary, but curious about what's causing the error

packages/mongo-menu/src/MongoMenu.spec.tsx
```diff
@@ -1,45 +1,50 @@
...
 
+  const { getByText } = renderedComponent;
   const trigger = getByText('Leafy');
   fireEvent.click(trigger);
 
-  test('Atlas menu item links to cloud.mongodb.com', () => {
-    const atlasItem = getByText('Atlas').parentElement;
-    expect((atlasItem as HTMLAnchorElement).href).toBe(
-      'https://cloud.mongodb.com/',
+  const assertHrefByText = (expectedText: string, expectedHref: string) => {
```
brookescarlett: 👍 We might consider moving this to lib so that we can reuse, but can do this when we want to leverage the pattern again later.

brookescarlett:
@seanmhanson Thanks for raising this PR! Like the `href` helper method in the spec file, and think that down the line we'll probably want to move that into `lib`. No blocking feedback on my part, curious as to why we have to open up MenuItemProps to allow any prop to pass through, but if that's a necessary solution then I'm on board.

packages/menu/src/MenuItem.tsx
seanmhanson: Good catch, this was from when I was debugging something and completely forgot about this. Sorry!

brookescarlett: 👍

packages/mongo-menu/src/MongoMenu.story.tsx
```diff
@@ -3,16 +3,19 @@ import { storiesOf } from '@storybook/react';
...
+
+const activeProduct = select(
+  'activeProduct',
+  Object.values(Product) as Array<Product>,
+  Product.Atlas,
+);
+
+storiesOf('MongoMenu', module)
+  .add('Default', () => <MongoMenu user={user} activeProduct={activeProduct} />)
+  .add('Disabled MongoDB Account link', () => (
```
DesignerDave: We're trying to refrain from adding this sort of thing to the sidebar. If you remove this addition to the story, I think the PR is good to go 👍

seanmhanson:
@DesignerDave reverted the story file. Let me know if there's anything else to take care of!

PR #229, opened 2020-01-14
PD-483: MongoNav

brookescarlett:
## ✍️ Proposed changes

👋 there is definitely lots of outstanding work still to be done here, but wanted to open this PR to start getting feedback on the already published components (Menu and IconButton) so that we can release a beta version of `<MongoNav />`

In addition to feedback on the Menu and IconButton components, would be happy to receive early feedback on any of the parts of MongoMenu, I just want to call out that the component is far from finished. Lots of cleaning, testing, styling, etc. still to do.

🎟 _Jira ticket:_ [PD-483](https://jira.mongodb.org/browse/PD-483)

## 🛠 Types of changes

## ✅ Checklist


packages/menu/src/Menu.spec.js
```diff
@@ -103,4 +108,49 @@ describe('packages/Menu', () => {
       expect(secondItem.rel).toBe('help');
     });
   });
+
+  describe('packages/sub-menu', () => {
```
hswolff: Think it's a little confusing to name it this way, as sub-menu isn't its own package. I think just naming it `SubMenu` is fine. We don't really have hard conventions here, do we?

brookescarlett: I think we have been doing it for subcomponents, but you're right, it doesn't make sense. I'll fix here and can take a look around the repo and make sure it's cleaned up there too.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -0,0 +1,55 @@
...
+  constructProjectURL?: (orgID: string, projID: string) => string;
+  overrides?: OverridesInterface;
+  showProjNav?: boolean;
+}
+
+export default function MongoNav({
+  activeProduct,
+  activeNav,
+  data,
+  overrides,
```
hswolff: What if we un-nest this by one level? Such that we have a `urls` and a `hosts` props directly on the component? Not sure we get much benefit nesting them within another object.

hswolff: Expanding on that...the biggest source of complexity for all these components is understanding what the default URL is and when to use a URL that is explicitly being set.

I think what may make that easier to reason about is if, within this component, `MongoNav`, we do all the logic around urls and default urls here.

Such that what we produce from that is the final shape of the URLs. So that we can just pass a `urls` prop to `OrgNav` and `ProjectNav` and they just use those values directly.

So all that logic is consolidated in one location, making it easy to reason about how some value ended up somewhere else.

brookescarlett: Done

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import OrgNav from './org-nav/index';
+import ProjNav from './proj-nav/index';
```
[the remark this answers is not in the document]

packages/mongo-nav/src/types.ts
```diff
@@ -0,0 +1,140 @@
...
+} as const;
+
+type Variant = typeof Variant[keyof typeof Variant];
+
+export { Variant };
+
+const PlanType = {
+  Cloud: 'cloud',
+  Atlas: 'atlas',
+  OM: 'OM',
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/tsconfig.json
```diff
@@ -0,0 +1,40 @@
+{
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,151 @@ function Menu({
...
+        }
+      };
+
+      const onFocus = ({ target }: { target: HTMLElement }) => {
+        setFocused(target);
+      };
+
+      if (isComponentType(child, 'SubMenu')) {
+        if (titleArr.includes(child.props.title)) {
+          throw 'SubMenu titles must be unique';
```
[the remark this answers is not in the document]

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+  const iconButtonRef: React.MutableRefObject<HTMLElement | null> = useRef(
+    null,
+  );
+
+  const onRootClick = (
+    e: React.MouseEvent<HTMLAnchorElement, MouseEvent> &
+      React.MouseEvent<HTMLButtonElement, MouseEvent>,
+  ) => {
+    if (iconButtonRef?.current?.contains(e.target as HTMLElement)) {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+  top: 0;
+`;
+
+type Glyph = keyof typeof glyphs;
+
+interface SharedSubMenuProps {
+  open?: boolean;
+  setOpen?: React.Dispatch<React.SetStateAction<boolean>>;
+  className?: string;
+  description: string | React.ReactElement;
```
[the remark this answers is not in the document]
brookescarlett: yes you're right 👍

packages/mongo-nav/src/types.ts
```diff
@@ -0,0 +1,140 @@
...
+  orgName: string;
+  planType: PlanType;
+}
+
+export interface CurrentOrganizationInterface extends OrganizationInterface {
+  paymentStatus?: OrgPaymentLabel;
+}
+
+export interface DataInterface {
+  account: AccountInterface;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
 
   const popoverRef: React.RefObject<HTMLUListElement> = useRef(null);
 
+  const setFocus = (el: HTMLElement) => {
+    setFocused(el);
+    el?.focus();
+  };
+
   useMemo(() => {
     if (open === false) {
```
DesignerDave: Is there a reason we don't just do this?
```js
useMemo(() => {
  hasSetInitialFocus.current = open;
  hasSetInitialOpen.current = open;
}, [open])
```

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+const activeMainTextStyle = css`
+  font-weight: bolder;
+  color: ${uiColors.green.dark3};
+`;
+
+const iconButtonStyle = css`
+  position: absolute;
+  z-index: 1;
+  right: 8px;
+  top: ${52 / 2 - 22 / 2}px;
```
DesignerDave: Can you comment these magic numbers, or make them variables?

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+const openIconButtonStyle = css`
+  background-color: ${uiColors.white};
+  transition: background-color 150ms ease-in-out;
+`;
+
+const mainIconStyle = css`
+  color: ${uiColors.gray.base};
+  // 15px of padding from MenuItem + 32px width of SVG + 13px = 60px
+  // which is the padding-left for children of SubMenus.
```
DesignerDave: I don't see any use of 60px in the code below.

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+  list-style: none;
+  padding: 0;
+  height: 0;
+  overflow: hidden;
+  transition: height 150ms ease-in-out;
+`;
+
+const menuItemBorder = css`
+  position: absolute;
+  width: 100%;
```
DesignerDave: With absolute-positioned elements, we should use:
```css
left: 0;
right: 0;
```

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+    onExited = () => {},
+    open = false,
+    active = false,
+    disabled = false,
+    ...rest
+  } = props;
+
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+
+  const iconButtonRef: React.MutableRefObject<HTMLElement | null> = useRef(
```
DesignerDave: I don't think I've seen `MutableRefObject` used. Why that vs `refObject`?

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+              return React.cloneElement(child, {
+                children: (
+                  <>
+                    <div className={menuItemBorder}></div>
+                    {child.props.children}
+                  </>
+                ),
+                className: cx(
+                  css`
+                    padding-left: ${glyph ? '60px' : '28px'};
```
DesignerDave: I don't understand from looking at this why those values are used

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+              `]: state === 'entered',
+            })}
+            role="menu"
+            aria-label={title}
+          >
+            {React.Children.map(children as React.ReactElement, child => {
+              return React.cloneElement(child, {
+                children: (
+                  <>
+                    <div className={menuItemBorder}></div>
```
DesignerDave: Self-terminate here

packages/menu/src/styles.tsx
```diff
@@ -0,0 +1,96 @@
...
+  margin: unset;
+  width: 100%;
+  font-family: Akzidenz, ‘Helvetica Neue’, Helvetica, Arial, sans-serif;
+  box-sizing: border-box;
+  background: ${uiColors.white};
+  text-align: left;
+  &:focus {
+    outline: none;
+  }
+  &:before {
```
DesignerDave: Let's add line breaks before all pseudo-classes and states that are preceded by style declarations.

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
-            setFocused(refs[0]);
-            hasSetInitialFocus.current = true;
-          }
-        },
-        onFocus: (e: Event) => setFocused(e.target as HTMLElement),
-      });
-    }
+  function updateChildren(children: any): Array<React.ReactElement> {
+    return React.Children.map(children, child => {
+      if (child.props?.disabled) {
```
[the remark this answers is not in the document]

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
+        return child;
+      }
 
-    return child;
-  });
+      let currentChildRef: HTMLElement | null = null;
+
+      const setRef = (ref: HTMLElement) => {
```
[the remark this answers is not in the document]

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
+        if (ref == null) {
+          return;
+        }
+
+        refs.push(ref);
+        currentChildRef = ref;
+
+        if (open && hasSetInitialFocus.current === false) {
+          setFocus(refs[0]);
+          setUsingKeyboard(true);
```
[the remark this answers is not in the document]

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
+          open: isCurrentSubMenu,
+          setOpen: (state: boolean) => {
+            if (currentChildRef) {
+              setFocused(currentChildRef);
+            }
+
+            setCurrentSubMenu(state ? child : null);
+          },
+          onKeyDown: (e: KeyboardEvent) => {
+            if (e.keyCode === keyMap.ArrowLeft) {
```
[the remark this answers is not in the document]

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
+
+  // When a SubMenu becomes open, it's set to currentSubMenu, and we focus on the first child.
+  useEffect(() => {
+    if (!currentSubMenu) {
+      const focusedRefIndex = refs.indexOf(focused);
+      const subMenu = refs[focusedRefIndex];
+
+      subMenu?.focus();
+      setUsingKeyboard(true);
+      return;
```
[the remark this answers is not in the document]
brookescarlett: should just be wrapped in an `if, else` you're right.

packages/menu/src/Menu.tsx
brookescarlett: yes because we manually set it to open above, we're just resetting back to false when the menu is closed.

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
brookescarlett: reorg'd math to make more sense, was explaining why we have margin-right at 13px

packages/menu/src/SubMenu.tsx
brookescarlett: left: 0 makes the border start all the way at the left; we want the border to be more indented and not fill the entire line

brookescarlett: Screen Shot 2020-01-21 at 3 42 07 PM

brookescarlett: ^ hard to articulate

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
hswolff: Yeah I actually played around with this a lot locally. What is here isn't great but I can't think of anything that's drastically better.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -0,0 +1,155 @@
...
+  activeNav,
+  data,
+  hosts,
+  onOrganizationChange,
+  onProjectChange,
+  urls,
+  showProjNav = true,
+  admin = false,
+  constructOrganizationURL = orgId =>
+    `https://cloud.mongodb.com/v2#/org/${orgId}/projects`,
```
hswolff: It'd be nice if these default functions could default to the hosts value for constructing these URLs. Would allow consumers to only provide the hosts value.

brookescarlett: Hm so just accept `(host, orgId)` and then we construct?

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -0,0 +1,155 @@
...
+        userPreferences:
+          urls?.mongoMenu?.cloud?.userPreferences ??
+          `${cloudHost}/v2#/preferences/personalization`,
+        organizations:
+          urls?.mongoMenu?.cloud?.organizations ??
+          `${cloudHost}/v2#/preferences/organizations`,
+        invitations:
+          urls?.mongoMenu?.cloud?.invitations ??
+          `${cloudHost}/v2#/preferences/invitations`,
+        tfa: urls?.mongoMenu?.cloud?.tfa ?? `${cloudHost}/v2#/preferences/2fa`,
```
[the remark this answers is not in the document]

packages/mongo-nav/src/mongo-menu/MongoMenu.tsx
```diff
@@ -0,0 +1,309 @@
...
+            </Button>
+          </FocusableMenuItem>
+        </div>
+        <MenuSeparator />
+
+        <SubMenu
+          {...subMenuContainer.prop}
+          {...sharedProps}
+          active={isCloud}
+          href="https://cloud.mongodb.com"
```
hswolff: Should use hosts value for non-prod environments too. Ditto others.

packages/mongo-nav/src/mongo-menu/MongoMenu.tsx
```diff
@@ -0,0 +1,309 @@
...
+          description={<Description isActive={isCloud} product="cloud" />}
+          title="Atlas"
+          glyph="Cloud"
+          className={cx(subMenuContainerStyle, {
+            [subMenuActiveContainerStyle]: isCloud,
+          })}
+        >
+          <MenuItem href={urls?.mongoMenu?.cloud?.userPreferences}>
```
hswolff: These don't need the optionals anymore, right?

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -0,0 +1,155 @@
...
+      university: {
+        videoPreferences:
+          urls?.mongoMenu?.university?.videoPreferences ?? `${universityHost}`,
+      },
+      support: {
+        userPreferences:
+          urls?.mongoMenu?.support?.userPreferences ?? `${supportHost}/profile`,
+      },
+      account: {
+        accountURL:
```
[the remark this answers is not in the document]

packages/mongo-nav/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import MongoNav from './MongoNav';
+import MongoMenu from './mongo-menu/index';
+import MongoSelect from './mongo-select/index';
+export { MongoMenu, MongoSelect };
```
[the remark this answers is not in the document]
brookescarlett: Not sure about MongoSelect, but can see a world in which we'd want tot expose just MongoMenu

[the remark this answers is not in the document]
brookescarlett: (University?)

packages/mongo-nav/src/types.ts
```diff
@@ -0,0 +1,150 @@
...
+export { Product };
+
+const NavItem = {
+  Billing: 'billing',
+  Support: 'support',
+  AccessManager: 'accessManager',
+  AllClusters: 'allClusters',
+  Admin: 'admin',
+  OrgSettings: 'orgSettings',
+};
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/src/types.ts
```diff
@@ -0,0 +1,150 @@
...
+  EmbargoConfirmed: 'embargo confirmed',
+  Ok: 'ok',
+  Warning: 'warning',
+  Suspended: 'suspended',
+  Closing: 'closing',
+  AdminSuspended: 'admin suspended',
+  Dead: 'dead',
+  Locked: 'locked',
+  Closed: 'closed',
+};
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -0,0 +1,280 @@
...
+  data,
+  constructProjectURL,
+  urls,
+  alerts,
+  activeProduct,
+  onProjectChange,
+}: ProjectNavInterface) {
+  const { projectNav } = urls;
+
+  function calcStyle() {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -0,0 +1,280 @@
...
+          <MenuItem href={projectNav.settings}>Project Settings</MenuItem>
+          <MenuItem href={projectNav.accessManager}>
+            Project Access Manager
+          </MenuItem>
+          <MenuItem href={projectNav.support}>Project Support</MenuItem>
+          <MenuItem href={projectNav.integrations}>Integrations</MenuItem>
+        </Menu>
+        <ol className={olStyle}>
+          <li role="none" className={productStyle}>
+            <a href={'https://cloud.mongodb.com'} className={productTextStyle}>
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/menu/src/Menu.tsx
```diff
@@ -39,15 +40,15 @@ interface MenuProps
...
+  setOpen?: React.Dispatch<React.SetStateAction<boolean>>;
 
   /**
    * Callback to determine whether or not Menu should close when user tries to close it.
    *
    */
   shouldClose?: () => boolean;
 
-  children: React.ReactNode;
+  children: Array<React.ReactElement>;
```
DesignerDave: Hmm... Isn't there a type for children specifically that includes text nodes, arrays, and react elements?

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
   ...rest
 }: MenuProps) {
   const refs: Array<HTMLElement> = [];
+  const titleArr: Array<string> = [];
+
   const hasSetInitialFocus = useRef(false);
+  const hasSetInitialOpen = useRef(false);
+  // @ts-ignore: using this setState to trigger a subMenu close after animation
+  const [closed, setClosed] = useState(false); // eslint-disable-line
```
DesignerDave: So rather than adding ts/eslint ignores, I think there's two better ways to write this:
```js
const [ , setClosed] = useState(false);
```
OR
```js
const {1: setClosed} = useState(false);
```

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
+          hasSetInitialFocus.current = true;
+        }
+      };
+
+      const onFocus = ({ target }: { target: HTMLElement }) => {
+        setFocused(target);
+      };
+
+      if (isComponentType<'SubMenu'>(child, 'SubMenu')) {
+        if (titleArr.includes(child.props.title)) {
```
DesignerDave: Boy we use `child.props.title` quite a bit. Might be worth destructuring props to make the code more succinct.

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
-  const setOpen = isControlled ? controlledSetOpen : uncontrolledSetOpen;
+  const setOpen =
+    (typeof controlledOpen === 'boolean' && controlledSetOpen) ||
+    uncontrolledSetOpen;
+  const open = controlledOpen ?? uncontrolledOpen;
+
+  // When a SubMenu becomes open, it's set to currentSubMenu, and we focus on the first child.
+  useEffect(() => {
+    if (!currentSubMenu) {
+      const focusedRefIndex = refs.indexOf(focused);
```
DesignerDave: Rather than using optional chaining to check for the existence of `subMenu` before focusing, it might be better to instead check that focusedRefIndex is not `-1`.

DesignerDave: Ditto in the `else` block.

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,152 @@ function Menu({
...
+    currentSubMenu,
+    open,
+    refs,
+  ]);
 
   const popoverRef: React.RefObject<HTMLUListElement> = useRef(null);
 
+  const setFocus = (el: HTMLElement) => {
+    setFocused(el);
+    el?.focus();
```
DesignerDave: Rather than checking here if `el` exists, we should probably make the function return on the first line if el is undefined. Otherwise we may have a bug in the future where we pass `undefined` to `setFocused()`.

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+  font-weight: bolder;
+  color: ${uiColors.green.dark3};
+`;
+
+const iconButtonStyle = css`
+  position: absolute;
+  z-index: 1;
+  right: 8px;
+  // 56 is the height of the SubMenuItem container and 28 is the height of the IconButton
+  top: ${56 / 2 - 28 / 2}px;
```
DesignerDave: Probably best to move these values to variables with meaningful names.

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+  margin: auto;
+  background-color: ${uiColors.gray.light3};
+  transition: background-color 150ms ease-in-out;
+  ${subMenuContainer.selector}:hover + & {
+    background-color: ${uiColors.gray.light2};
+    transition: background-color 150ms ease-in-out;
```
DesignerDave: This transition is redundant with the one on line :76

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+const iconButtonFocusedStyle = css`
+  ${subMenuContainer.selector}:focus + & {
+    background-color: ${uiColors.blue.light3};
+    transition: background-color 150ms ease-in-out;
+  }
+`;
+
+const openIconButtonStyle = css`
+  background-color: ${uiColors.white};
+  transition: background-color 150ms ease-in-out;
```
DesignerDave: We're setting this transition a lot, but we likely only need to do it once.

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+`;
+
+const mainIconStyle = css`
+  color: ${uiColors.gray.base};
+  // 60px (padding-left for SubMenus) - 32px (width of SVG) - 15px (padding from MenuItem) = 13
```
DesignerDave: Hmm... Wouldn't this annotation be unnecessary if these values were simply variables?

packages/menu/src/SubMenu.tsx
DesignerDave: ~Hmm... So I got it working, though I had to make a few other CSS tweaks to its ancestors.~

Nevermind, that created other problems. Carry on! 😅

packages/mongo-nav/src/mongo-menu/MongoMenu.tsx
brookescarlett: 👍

packages/mongo-nav/src/mongo-menu/MongoMenu.tsx
brookescarlett: 👍

packages/menu/src/Menu.tsx

packages/menu/src/Menu.tsx

packages/menu/src/Menu.tsx

packages/menu/src/Menu.tsx

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
brookescarlett: you're correct

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.tsx

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,158 @@ function Menu({
...
+
+        if (open && hasSetInitialFocus.current === false) {
+          setFocus(refs[0]);
+          // This allows the focus state to be shown via the LeafyGreen Provider
+          setUsingKeyboard(true);
+          hasSetInitialFocus.current = true;
+        }
+      };
+
+      const title = child?.props?.title != null && child.props.title;
```
DesignerDave: Hmm... Is this better than:

```js
const title = child?.props?.title ?? false
```

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -0,0 +1,155 @@
...
+  onProjectChange,
+  urls,
+  showProjNav = true,
+  admin = false,
+  constructOrganizationURL = orgId =>
+    `https://cloud.mongodb.com/v2#/org/${orgId}/projects`,
+  constructProjectURL = (orgId, projectId) =>
+    `https://cloud.mongodb.com/v2#/${projectId}`,
+}: MongoNavInterface) {
+  const accountHost = hosts?.account ?? `https://account.mongodb.com`;
```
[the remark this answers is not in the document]
DesignerDave: Is there an alternative you'd propose? Looking at the built `index.web.js` file, it looks less egregious to me:

```js
const accountHost = hosts?.account ?? `https://account.mongodb.com`;
// to
Ee = null !== (t = null == fe ? void 0 : fe.account) && void 0 !== t ? t : "https://account.mongodb.com",
```

Ultimately whether or not we're okay with this is based on what would be shipped to the user. I'm not convinced that the bundle would be smaller if we wrote this differently.

That said, the `index.web.js` file altogether looks like it's about 60KB (20K gzipped), which seems a tad large. I do think it's partially because of the SVG paths that are being bundled (which should decrease in this package once the final versions are included in the Icon component).

I wonder if there are any quick wins in terms of bundle size we could make here...

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const cloudHost = hosts?.cloud ?? `https://cloud.mongodb.com`;
+  // @ts-ignore
```
DesignerDave: Can we either solve for why this is here, or comment why the TS ignore is necessary (same for the eslint-disable-line)

packages/mongo-nav/src/MongoNav.tsx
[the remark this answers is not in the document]
DesignerDave: I suppose the fixture data we're using won't be in v1 either.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -0,0 +1,155 @@
...
+        `${cloudHost}/v2/${data.currentProject.projectId}#activity`,
+    },
+  };
+
+  return (
+    <>
+      <OrgNav
+        account={data.account}
+        activeProduct={activeProduct}
+        current={data.currentOrganization}
```
DesignerDave: Ya might consider destructuring `data` to clean up stuff like this a tad.

packages/mongo-nav/src/index.ts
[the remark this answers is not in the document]
DesignerDave: MongoSelect doesn't have a use-case outside of MongoNav, but there are plans by (I believe) both support and University to use MongoMenu in the near future. We should definitely keep exporting it.

There may be an argument for moving MongoMenu back to its own package in that case though, especially given tree-shaking our components isn't really a thing yet. Thoughts?

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+        mountOnEnter
+        unmountOnExit
+        onExited={onExited}
+      >
+        {(state: string) => (
+          <ul
+            className={cx(ulStyle, {
+              [css`
+                height: ${subMenuItemHeight *
+                  React.Children.toArray(children).length}px;
```
DesignerDave: Might be worth moving the length to a variable before the JSX block to make this bit a little more readable.

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+          </div>
+
+          <div className={cx({ [disabledTextStyle]: disabled })}>
+            {description}
+          </div>
+        </div>
+      </Root>
+      <IconButton
+        ref={iconButtonRef}
+        ariaLabel={open ? 'Caret Up' : 'Caret Down'}
```
DesignerDave: Wouldn't it be better for accessibility to set the label to the button's action?
```
ariaLabel={open ? 'Close Sub-menu' : 'Open Sub-menu'}
```

packages/menu/src/SubMenu.tsx
```diff
@@ -0,0 +1,349 @@
...
+        className={cx(iconButtonStyle, {
+          [openIconButtonStyle]: open,
+          [iconButtonFocusedStyle]: showFocus,
+        })}
+        onClick={(e: React.MouseEvent) => {
+          e.nativeEvent.stopImmediatePropagation();
```
DesignerDave: Can we comment why we need this? I can't even remember myself, but I know it was necessary.

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: removed 😄

packages/mongo-nav/src/MongoNav.tsx

packages/menu/src/Menu.tsx

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
brookescarlett: Good point

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/menu/src/Menu.spec.js
```diff
@@ -102,5 +111,54 @@ describe('packages/Menu', () => {
...
+          <MenuItem data-testid="sub-menu-item-c">SubMenu Item 3</MenuItem>
+        </SubMenu>
+      </Menu>,
+    );
+
+    const subMenuItemA = getByTestId('sub-menu-item-a');
+    const subMenuB = getByTestId('sub-menu-b');
+    const subMenuBArrow = subMenuB.parentNode.querySelectorAll('button')[1];
+
+    test('renders a SubMenu open by default, when the SubMenu is Active', () => {
```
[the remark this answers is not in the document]
brookescarlett: Yes we do, the next test: `when a SubMenu is clicked, it opens and closes the previously opened SubMenu`

packages/mongo-nav/src/mongo-menu/MongoMenu.tsx
```diff
@@ -0,0 +1,317 @@
...
+            </Button>
+          </FocusableMenuItem>
+        </div>
+        <MenuSeparator />
+
+        <SubMenu
+          {...subMenuContainer.prop}
+          {...sharedProps}
+          active={isCloud}
+          href={hosts?.cloud ?? 'https://cloud.mongodb.com'}
```
hswolff: Ditto elsewhere, shouldn't this already be sanitized?

packages/menu/src/styles.tsx
brookescarlett: 👍

packages/mongo-nav/src/types.ts
```diff
@@ -0,0 +1,150 @@
...
+const NavItem = {
+  Billing: 'billing',
+  Support: 'support',
+  AccessManager: 'accessManager',
+  AllClusters: 'allClusters',
+  Admin: 'admin',
+  OrgSettings: 'orgSettings',
+} as const;
+
+type NavItem = typeof NavItem[keyof typeof NavItem] | '';
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/src/data.ts
```diff
@@ -0,0 +1,51 @@
+import { DataInterface } from './types';
+
+export default {
+  account: {
+    email: 'harry.wolff@mongodb.com',
```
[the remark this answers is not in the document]
brookescarlett: Gave it a stab, if you want the data to be more specific, include something in particular lmk. Otherwise just named orgs/projects Demo Org/Demo Project

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -0,0 +1,252 @@
...
+          data={data}
+          current={current}
+          constructOrganizationURL={constructOrganizationURL}
+          variant="organization"
+          urls={urls}
+          onChange={onOrganizationChange}
+          isActive={activeNav === 'orgSettings'}
+        />
+        <ul className={ulContainer}>
+          {variant && current.paymentStatus && (
```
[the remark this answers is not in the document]
brookescarlett: https://jira.mongodb.org/browse/PD-493

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -0,0 +1,292 @@
...
+          onChange={onProjectChange}
+        />
+        <Menu
+          trigger={
+            <IconButton ariaLabel="More" className={menuIconButtonStyle}>
+              <Icon glyph="Ellipsis" className={menuIconStyle} />
+            </IconButton>
+          }
+        >
+          <MenuItem href={projectNav.settings}>Project Settings</MenuItem>
```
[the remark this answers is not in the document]
brookescarlett: Do you mean conditionally deciding whether the path should live inside an `href` tag or a `to` tag? I added support for rendering MenuItems as a custom element in this PR, so passing Link to the as prop should just work

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -0,0 +1,292 @@
...
+          <MenuItem href={projectNav.accessManager}>
+            Project Access Manager
+          </MenuItem>
+          <MenuItem href={projectNav.support}>Project Support</MenuItem>
+          <MenuItem href={projectNav.integrations}>Integrations</MenuItem>
+        </Menu>
+        <ol className={olStyle}>
+          <li role="none" className={productStyle}>
+            <a
+              href={hosts?.cloud ?? 'https://cloud.mongodb.com'}
```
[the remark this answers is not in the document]
brookescarlett: currently we aren't sanitizing the hosts, just the URLs, I can do that though!

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -0,0 +1,222 @@
...
+      </MenuItem>
+    );
+  }
+
+  const renderProjectOption = (
+    props: ProjectMongoSelectProps,
+    datum: ProjectInterface,
+  ) => {
+    const id = datum.projectId;
+    const content = (
```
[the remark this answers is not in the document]

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -0,0 +1,222 @@
...
+      </FocusableMenuItem>
+
+      <li role="none" className={listContainerStyle}>
+        <ul
+          onKeyDown={e => e.preventDefault()}
+          className={ulContainerStyle}
+          role="menu"
+        >
+          {isProject(props)
+            ? props.data.map(datum => renderProjectOption(props, datum))
```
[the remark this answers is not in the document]
brookescarlett: Unless the data is properly typed, Typescript complains about trying to map over the data: ```This expression is not callable.
  Each member of the union type '(<U>(callbackfn: (value: ProjectInterface, index: number, array: ProjectInterface[]) => U, thisArg?: any) => U[]) | (<U>(callbackfn: (value: OrganizationInterface, index: number, array: OrganizationInterface[]) => U, thisArg?: any) => U[])' has signatures, but none of those signatures are compatible with each other.ts(2349)``` So checking the type of data/datum before the map helps to avoid that error

packages/mongo-nav/src/project-nav/ProjectNav.tsx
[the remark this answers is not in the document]
brookescarlett: Done

packages/mongo-nav/src/mongo-menu/MongoMenu.tsx
brookescarlett: 👍

.changeset/plenty-yaks-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/menu': minor
+---
+
+MenuItems accept an as prop to determine what component to render as. Introduce SubMenu components with stronger accessibility logic
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/menu/src/Menu.tsx
```diff
@@ -88,46 +89,158 @@ function Menu({
...
+        if (child.props.active && !hasSetInitialOpen.current) {
+          setCurrentSubMenu(child);
+          hasSetInitialOpen.current = true;
+        }
+
+        const isCurrentSubMenu = currentSubMenu?.props.title === title;
+
+        return React.cloneElement(child, {
+          ref: setRef,
+          open: isCurrentSubMenu,
```
DesignerDave: Not a blocker, but we might consider reordering these for readability, putting all the inline function declarations last.

PR #231, opened 2020-01-22
PD-486: SideNav Component

seanmhanson:
## ✍️ Proposed Changes and Additions

As part of IA Navigation, one major goal outside of the new MongoNav is to unify the side navigation appearance and behavior for most of our Cloud products, including:
- Atlas & Cloud Manager (Atlas, Cloud Manager, Organizations, Projects, User Preferences)
- Ops Manager 4.4 (Ops Manager, Organizations, Projects, User Account)
- MongoDB Account
- Federations / IDPs
- Stitch/Realm
- Charts

This component establishes the style and pattern we will use across all applications. It allows individual items to be customized to use specific components, allowing integration with React Router and other components.

One benefit of adding this component early is that we can update all of Cloud's navigation immediately (save for the removal of Stitch and Charts links), since it telegraphs no upcoming changes and is functionally independent of the rest of IA. Get pumped.

🎟 _Jira ticket:_ [PD-486](https://jira.mongodb.org/browse/PD-486)

## 🛠 Types of Changes

## ✅ Checklist

I have yet to complete the documentation, and some automatically generated documentation is not yet present but should be easy to add.

## 📷 Demo

Screen Shot 2020-01-24 at 12 33 11 PM

### DIFF
--- a/.changeset/clever-dolls-know.md
+++ b/.changeset/clever-dolls-know.md
@@ -1,0 +1,7 @@
+---
+'@leafygreen-ui/side-nav': major
+'@leafygreen-ui/lib': minor
+'@leafygreen-ui/menu': patch
+---
+
+Introduces the SideNav component. Adds enumerated aria-current values to lib for general use. Fixes some tiny internal typos in the Menu component.
--- a/.changeset/clever-dolls-know.md
+++ b/.changeset/clever-dolls-know.md
@@ -1,7 +1,6 @@
 ---
 '@leafygreen-ui/side-nav': major
 '@leafygreen-ui/lib': minor
-'@leafygreen-ui/menu': patch
 ---
 
-Introduces the SideNav component. Adds enumerated aria-current values to lib for general use. Fixes some tiny internal typos in the Menu component.
+Introduces the SideNav component. Adds enumerated aria-current values to lib for general use.
--- a/.changeset/tiny-lies-type.md
+++ b/.changeset/tiny-lies-type.md
@@ -1,6 +1,0 @@
----
-'@leafygreen-ui/menu': minor
-'@leafygreen-ui/side-nav': minor
----
-
-Mongo Menu now adds a class to style title text. Side Nav now has MVP functionality.
--- a/.changeset/tiny-lies-type.md
+++ b/.changeset/tiny-lies-type.md
@@ -1,0 +1,6 @@
+---
+'@leafygreen-ui/menu': minor
+'@leafygreen-ui/side-nav': minor
+---
+
+Mongo Menu now adds a class to style title text. Side Nav now has MVP functionality.
--- a/.changeset/two-mirrors-glow.md
+++ b/.changeset/two-mirrors-glow.md
@@ -1,6 +1,0 @@
----
-'@leafygreen-ui/side-nav': major
-'@leafygreen-ui/menu': minor
----
-
-MongoMenu now adds a class to style title text. SideNav has been introduced as a new component.
--- a/.changeset/two-mirrors-glow.md
+++ b/.changeset/two-mirrors-glow.md
@@ -3,4 +3,4 @@
 '@leafygreen-ui/menu': minor
 ---
 
-Mongo Menu now adds a class to style title text. Side Nav has been introduced as a new component.
+MongoMenu now adds a class to style title text. SideNav has been introduced as a new component.
--- a/.changeset/two-mirrors-glow.md
+++ b/.changeset/two-mirrors-glow.md
@@ -1,0 +1,6 @@
+---
+'@leafygreen-ui/side-nav': major
+'@leafygreen-ui/menu': minor
+---
+
+Mongo Menu now adds a class to style title text. Side Nav has been introduced as a new component.
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,13 @@ storybook/public/
 # JUnit report files
 reports
 
+# OS generated files
+.DS_Store
+.DS_Store?
+.Trashes
+ehthumbs.db
+Thumbs.db
+
 # Yarn Error Log
 yarn-error.log
 tsconfig.tsbuildinfo
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -63,10 +63,10 @@ export const keyMap = {
 /**
  * An enum of accepted values for the "aria-current" attribute, used for
  * indicating current/active state across several contexts.
- * 
+ *
  * The values "false", the empty string, and an ommission of this attribute
  * are all treated identically by user agents and screen readers.
- * 
+ *
  * W3C Recommendation: https://www.w3.org/TR/wai-aria-1.1/#aria-current
  */
 export const AriaCurrentValue = {
@@ -76,7 +76,7 @@ export const AriaCurrentValue = {
   Date: 'date', // current value in a calendar or date picker
   Time: 'time', // current value in a timetable or time picker
   True: 'true', // (fallback) current value in any set of options/elements
-  Unset: 'false' // equivalent to omitting the aria-current attribute
+  Unset: 'false', // equivalent to omitting the aria-current attribute
 } as const;
 
 export type AriaCurrentValue = typeof AriaCurrentValue[keyof typeof AriaCurrentValue];
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -59,3 +59,24 @@ export const keyMap = {
   Space: 32,
   Tab: 9,
 };
+
+/**
+ * An enum of accepted values for the "aria-current" attribute, used for
+ * indicating current/active state across several contexts.
+ * 
+ * The values "false", the empty string, and an ommission of this attribute
+ * are all treated identically by user agents and screen readers.
+ * 
+ * W3C Recommendation: https://www.w3.org/TR/wai-aria-1.1/#aria-current
+ */
+export const AriaCurrentValue = {
+  Page: 'page', // current value in a set of pagination links
+  Step: 'step', // current value in a step indicator
+  Location: 'location', // current value in a chart or other visual flow
+  Date: 'date', // current value in a calendar or date picker
+  Time: 'time', // current value in a timetable or time picker
+  True: 'true', // (fallback) current value in any set of options/elements
+  Unset: 'false' // equivalent to omitting the aria-current attribute
+} as const;
+
+export type AriaCurrentValue = typeof AriaCurrentValue[keyof typeof AriaCurrentValue];
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import { HTMLElementProps, createDataProp } from '@leafygreen-ui/lib';
 import { css, cx } from '@leafygreen-ui/emotion';
 import { uiColors } from '@leafygreen-ui/palette';
-import { AriaCurrentValue } from '@leafygreen-ui/lib';
 import { useUsingKeyboardContext } from '@leafygreen-ui/leafygreen-provider';
 
 const menuItemContainer = createDataProp('menu-item-container');
@@ -141,29 +140,21 @@ const disabledTextStyle = css`
 
 interface SharedMenuItemProps {
   /**
+   * Class name that will be applied to root MenuItem element.
+   */
+  className?: string;
+  /**
    * Determines whether or not the MenuItem is active.
    */
   active?: boolean;
-  /**
-   * Determines whether or not the MenuItem is disabled.
-   */
-  disabled?: boolean;
   /**
    * Description text displayed below title in MenuItem.
    */
   description?: string;
   /**
-   * The value for aria-current if the SideNavItem is active
+   * Determines whether or not the MenuItem is disabled.
    */
-  ariaCurrentValue?: AriaCurrentValue;
-  /**
-   * Class name that will be applied to root MenuItem element.
-   */
-  className?: string;
-  /**
-   * Class name that will be applied to the title text/children wrapper.
-   */
-  titleTextClassName?: string;
+  disabled?: boolean;
   /**
    * Content that will appear inside of the underlying MenuItem's content wrapper.
    */
@@ -181,7 +172,7 @@ interface ButtonMenuItemProps
   href?: null;
 }
 
-export type MenuItemProps = LinkMenuItemProps | ButtonMenuItemProps;
+type MenuItemProps = LinkMenuItemProps | ButtonMenuItemProps;
 
 function usesLinkElement(
   props: LinkMenuItemProps | ButtonMenuItemProps,
@@ -193,16 +184,14 @@ function usesLinkElement(
  * # MenuItem
  *
  * ```
-  <MenuItem href="#homeward" titleTextClassName="header-bold" active>
-    My Beautiful Menu Item
+  <MenuItem href="#homeward" active>
+    Content
   </MenuItem>
  * ```
  @param props.active Determines whether or not the MenuItem is active.
  @param props.disabled Determines whether or not the MenuItem is disabled.
  @param props.description Description text optionally displayed below the MenuItem's title.
- @param props.ariaCurrentValue The value for aria-current if the MenuItem is active
  @param props.className Class name that will be applied to root MenuItem element.
- @param props.titleTextClassName Class name that will be applied to the title text/children wrapper.
  @param props.href When provided, the underlying MenuItem's root component will be rendered as an anchor with this href value.
  @param props.children Content that will appear inside of the underlying MenuItem's content wrapper.
  *
@@ -210,14 +199,12 @@ function usesLinkElement(
 const MenuItem = React.forwardRef(
   (props: MenuItemProps, forwardRef: React.Ref<any>) => {
     const {
-      active = false,
       disabled = false,
-      description,
-      ariaCurrentValue = 'page',
+      active = false,
       className,
-      titleTextClassName,
-      href,
       children,
+      description,
+      href,
       ...rest
     } = props;
     const { usingKeyboard: showFocus } = useUsingKeyboardContext();
@@ -245,20 +232,15 @@ const MenuItem = React.forwardRef(
             className,
           )}
           role="menuitem"
-          aria-current={active ? ariaCurrentValue : ''}
           aria-disabled={disabled}
           ref={forwardRef}
           tabIndex={disabled ? -1 : undefined}
         >
           <div
-            className={cx(
-              titleTextStyle,
-              {
-                [activeTitleTextStyle]: active,
-                [disabledTextStyle]: disabled,
-              },
-              titleTextClassName,
-            )}
+            className={cx(titleTextStyle, {
+              [activeTitleTextStyle]: active,
+              [disabledTextStyle]: disabled,
+            })}
           >
             {children}
           </div>
@@ -288,13 +270,11 @@ MenuItem.displayName = 'MenuItem';
 
 // @ts-ignore: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37660
 MenuItem.propTypes = {
-  active: PropTypes.bool,
-  disabled: PropTypes.bool,
-  description: PropTypes.string,
-  ariaCurrentValue: PropTypes.oneOf(Object.values(AriaCurrentValue)),
-  className: PropTypes.string,
-  titleTextClassName: PropTypes.string,
   href: PropTypes.string,
+  className: PropTypes.string,
+  description: PropTypes.string,
+  disabled: PropTypes.bool,
+  active: PropTypes.bool,
   children: PropTypes.node,
   ref: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
 };
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -143,18 +143,22 @@ interface SharedMenuItemProps {
    * Class name that will be applied to root MenuItem element.
    */
   className?: string;
+
   /**
    * Determines whether or not the MenuItem is active.
    */
   active?: boolean;
+
   /**
    * Description text displayed below title in MenuItem.
    */
   description?: string;
+
   /**
    * Determines whether or not the MenuItem is disabled.
    */
   disabled?: boolean;
+
   /**
    * Content that will appear inside of the underlying MenuItem's content wrapper.
    */
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -225,11 +225,13 @@ const MenuItem = React.forwardRef(
           tabIndex={disabled ? -1 : undefined}
         >
           <div
-            className={cx(titleTextStyle, {
-              [activeTitleTextStyle]: active,
-              [disabledTextStyle]: disabled,
+            className={cx(
+              titleTextStyle,
+              {
+                [activeTitleTextStyle]: active,
+                [disabledTextStyle]: disabled,
               },
-              titleTextClassName
+              titleTextClassName,
             )}
           >
             {children}
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -88,7 +88,7 @@ const titleTextStyle = css`
   }
 `;
 
-const activetitleTextStyle = css`
+const activeTitleTextStyle = css`
   font-weight: bold;
   color: ${uiColors.green.dark3};
 `;
@@ -134,7 +134,7 @@ const disabledStyle = css`
   }
 `;
 
-const disbaledTextStyle = css`
+const disabledTextStyle = css`
   color: ${uiColors.gray.light1};
 `;
 
@@ -143,6 +143,11 @@ interface SharedMenuItemProps {
    * Class name that will be applied to root MenuItem element.
    */
   className?: string;
+
+  /**
+   * Class name that will be applied to the title text/children wrapper.
+   */
+  titleTextClassName?: string;
 
   /**
    * Determines whether or not the MenuItem is active.
@@ -170,7 +175,7 @@ interface ButtonMenuItemProps
   href?: null;
 }
 
-type MenuItemProps = LinkMenuItemProps | ButtonMenuItemProps;
+export type MenuItemProps = LinkMenuItemProps | ButtonMenuItemProps;
 
 function usesLinkElement(
   props: LinkMenuItemProps | ButtonMenuItemProps,
@@ -184,6 +189,7 @@ const MenuItem = React.forwardRef(
       disabled = false,
       active = false,
       className,
+      titleTextClassName,
       children,
       description,
       href,
@@ -220,9 +226,11 @@ const MenuItem = React.forwardRef(
         >
           <div
             className={cx(titleTextStyle, {
-              [activetitleTextStyle]: active,
-              [disbaledTextStyle]: disabled,
-            })}
+              [activeTitleTextStyle]: active,
+              [disabledTextStyle]: disabled,
+              },
+              titleTextClassName
+            )}
           >
             {children}
           </div>
@@ -230,7 +238,7 @@ const MenuItem = React.forwardRef(
             <div
               className={cx(descriptionTextStyle, {
                 [activeDescriptionTextStyle]: active,
-                [disbaledTextStyle]: disabled,
+                [disabledTextStyle]: disabled,
               })}
             >
               {description}
@@ -255,6 +263,7 @@ MenuItem.propTypes = {
   href: PropTypes.string,
   onClick: PropTypes.func,
   className: PropTypes.string,
+  titleTextClassName: PropTypes.string,
   description: PropTypes.string,
   disabled: PropTypes.bool,
   active: PropTypes.bool,
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -245,7 +245,7 @@ const MenuItem = React.forwardRef(
             className,
           )}
           role="menuitem"
-          aria-current={active ? ariaCurrentValue : '' }
+          aria-current={active ? ariaCurrentValue : ''}
           aria-disabled={disabled}
           ref={forwardRef}
           tabIndex={disabled ? -1 : undefined}
--- a/packages/menu/src/MenuItem.tsx
+++ b/packages/menu/src/MenuItem.tsx
@@ -88,7 +88,7 @@ const titleTextStyle = css`
   }
 `;
 
-const activeTitleTextStyle = css`
+const activetitleTextStyle = css`
   font-weight: bold;
   color: ${uiColors.green.dark3};
 `;
@@ -134,7 +134,7 @@ const disabledStyle = css`
   }
 `;
 
-const disabledTextStyle = css`
+const disbaledTextStyle = css`
   color: ${uiColors.gray.light1};
 `;
 
@@ -153,16 +153,10 @@ interface SharedMenuItemProps {
    * Description text displayed below title in MenuItem.
    */
   description?: string;
-
   /**
    * Determines whether or not the MenuItem is disabled.
    */
   disabled?: boolean;
-
-  /**
-   * Content that will appear inside of the underlying MenuItem's content wrapper.
-   */
-  children?: React.ReactNode;
   ref?: React.Ref<any>;
 }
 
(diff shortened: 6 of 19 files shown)


.gitignore
```diff
@@ -15,6 +15,13 @@ storybook/public/
 # JUnit report files
 reports
 
+# OS generated files
+.DS_Store
+.DS_Store?
+.Trashes
+ehthumbs.db
+Thumbs.db
+
```
seanmhanson: This was a drive-by, because I was seeing macOS things try to sneak into my commits, and the thumbs.db tend to sneak in similarly when I am on my desktop back home.

packages/menu/src/MenuItem.tsx
```diff
@@ -184,6 +189,7 @@ const MenuItem = React.forwardRef(
       disabled = false,
       active = false,
       className,
+      titleTextClassName,
```
seanmhanson: Before this is complete, I need to examine this, because for some reason in some cases this is showing up as an attribute `titletextclassname` on the link/button, rather than as part of the class composition by Emotion.

packages/side-nav/README.md
```diff
@@ -0,0 +1,13 @@
+# Side Nav
```
seanmhanson: (to be added when next possible)

packages/side-nav/src/SideNav.spec.tsx
```diff
@@ -0,0 +1,224 @@
...
+    sideNavGroup: 'side-nav-group',
+    sideNavItem: 'side-nav-item',
+    sideNavLink: 'side-nav-link',
+  };
+
+  let renderedEls: RenderedElements = {};
+
+  afterEach(() => {
+    document.body.innerHTML = '';
+    renderedEls = {};
```
seanmhanson: [nitpick] I can remove the `afterAll` and add a call to `cleanup` here since this `afterEach` is at the root.

packages/side-nav/src/SideNav.spec.tsx
```diff
@@ -0,0 +1,224 @@
...
+      beforeEach(() => {
+        const { sideNavLink } = testIds;
+        const children = (
+          <a href="#clusters" data-testid={sideNavLink}>
+            Clusters
+          </a>
+        );
+
+        renderSideNavItem({ children });
+      });
```
seanmhanson: [TODO] this lacks a test to detect the correctly rendered child (this was a brief oversight, it should be simple)

packages/side-nav/src/SideNav.story.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { SideNav, SideNavItem, SideNavGroup } from './index';
+
+function MockAtlasSideNav() {
+  return (
+    <SideNav>
+      <SideNavGroup headerText="Data Storage">
+        <SideNavItem href="#clusters" active>
+          Clusters (active)
+        </SideNavItem>
+        {/* <SideNavItem>Triggers (not a link)</SideNavItem>
```
seanmhanson: These were commented out when testing something, and I accidentally committed the comments. These will be re-enabled.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+import { createDataProp } from '@leafygreen-ui/lib';
+import { ulStyleOverrides, LEFT_RIGHT_OFFSET } from './styles';
+
+export const sideNavGroupDataProp = createDataProp('side-nav-group-section');
```
seanmhanson: [question] I don't think this is actually being used, but I saw it in other components. Can this safely be removed?

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+      )}
+      <ul className={ulStyleOverrides}>{children}</ul>
+    </li>
+  );
+}
+
+SideNavGroup.displayName = 'SideNavGroup';
+
+SideNavGroup.propTypes = {
+  children: PropTypes.node,
```
seanmhanson: [correction needed] `children: PropTypes.node.isRequired`

packages/side-nav/src/SideNav.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <ul className={ulStyleOverrides}>{children}</ul>
+    </nav>
+  );
+}
+
+SideNav.displayName = 'Side Nav';
+
+SideNav.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.node,
```
seanmhanson: [correction needed] `children: PropTypes.node.isRequired`

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+SideNavItem.displayName = 'SideNavItem';
+
+SideNavItem.propTypes = {
+  active: PropTypes.bool,
+  disabled: PropTypes.bool,
+  className: PropTypes.string,
+  description: PropTypes.string,
+  children: PropTypes.node,
```
seanmhanson: [correction needed] `children: PropTypes.node.isRequired`

.changeset/tiny-lies-type.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/menu': minor
+'@leafygreen-ui/side-nav': minor
```
brookescarlett: new components should be a major bump so we can get to 1.0.0 when we release; unless we are planning to release this as an alpha/beta like we are doing with MongoNav

packages/side-nav/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "types": "./dist/index.d.ts",
+    "scripts": {
+      "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js"
+    },
+    "license": "Apache-2.0",
+    "publishConfig": {
+      "access": "public"
+    },
+    "devDependencies": {
+      "@leafygreen-ui/lib": "^3.2.0"
```
brookescarlett: I have to update this in the boilerplate, but please use lib at 4.0.0

packages/side-nav/src/SideNav.spec.tsx
```diff
@@ -0,0 +1,224 @@
+import React from 'react';
+import '@testing-library/jest-dom/extend-expect';
+import { render, cleanup } from '@testing-library/react';
+import { MenuItemProps } from '@leafygreen-ui/menu';
+import { SideNav, SideNavGroup, SideNavItem } from './index';
+
+type renderedElement = HTMLElement | null | undefined;
```
brookescarlett: Because you are implementing this type as optional, undefined is already inferred for you so you shouldn't need tot explicitly define here

packages/side-nav/src/SideNav.story.tsx
brookescarlett: Can add knobs here, so that users can play with active/non-active, href/nohref and the text of the SideNavItem

packages/side-nav/src/SideNav.tsx
```diff
@@ -0,0 +1,34 @@
...
+      aria-label="side-nav"
+      className={cx(sideNavStyles, className)}
+      {...rest}
+    >
+      <ul className={ulStyleOverrides}>{children}</ul>
+    </nav>
+  );
+}
+
+SideNav.displayName = 'Side Nav';
```
brookescarlett: [nit]: value of displayName should be one word: SideNav

packages/side-nav/src/SideNavGroup.tsx
brookescarlett: yes this can be removed if you aren't using the sideNavGroup selector, this prop is used to give us more specificity with css styles since we can't reference exact classNames

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+ */
+function SideNavGroup({
+  children,
+  className,
+  headerText,
+  headerClassName,
+  ...rest
+}: SideNavGroupProps) {
+  return (
+    <li className={className} {...rest} {...sideNavGroupDataProp.prop}>
```
brookescarlett: might need to add aria-level props to the list items (on this level and the one below), see Accessibility tab on storybook for more details

brookescarlett: actually seems to be complaining about lack of menu/menu bar labels, which is interesting because wai-aria doesn't implement those roles on it's own side nav... something to consider

packages/side-nav/src/styles.ts
```diff
@@ -0,0 +1,11 @@
+import { css } from '@leafygreen-ui/emotion';
+
+export const LEFT_RIGHT_OFFSET = 15;
```
brookescarlett: 👍

packages/side-nav/package.json
brookescarlett: also add the rest of the devDeps (menu, palette, emotion, etc.)

.gitignore
DesignerDave: Hmm... So far we've avoided adding editor / OS-specific files like these, and instead have each person use their own global gitignore to avoid files specific to their environment from being added. That said, I don't feel strongly about this.

@hswolff Do you feel strongly about it?

packages/menu/src/MenuItem.tsx
DesignerDave: Oh weird. I don't see anything particularly wrong here with that 🤷‍♂

packages/menu/src/MenuItem.tsx
```diff
@@ -219,18 +225,22 @@ const MenuItem = React.forwardRef(
...
+            )}
           >
             {children}
           </div>
           {description && (
             <div
               className={cx(descriptionTextStyle, {
                 [activeDescriptionTextStyle]: active,
-                [disbaledTextStyle]: disabled,
+                [disabledTextStyle]: disabled,
```
DesignerDave: I'd coordinate with @brookescarlett around changes to this file in particular. I know she also has a pull request in where this is fixed.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,79 @@
...
+  </SideNavItem>
+</SideNavGroup>
+ * ```
+ * @param props.children Content that will appear inside of SideNavGroup component.
+ * @param props.className ClassName that will be applied to root SideNavGroup element.
+ * @param props.headerText Header text displayed above the SideNavGroup contents.
+ * @param props.headerClassName ClassName that will be applied to SideNavGroup's header element.
+ *
+ */
+function SideNavGroup({
```
DesignerDave: Just want to get @brookescarlett to weigh in here – from an accessibility standpoint, this seems correct overall to me, but I could be wrong.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,79 @@
...
+  children,
+  className,
+  headerText,
+  headerClassName,
+  ...rest
+}: SideNavGroupProps) {
+  return (
+    <li className={className} {...rest}>
+      {headerText && (
+        <div className={cx(sideNavLabelStyle, headerClassName)} role="heading">
```
DesignerDave: We should probably use a more semantic tag than `div` here. Perhaps something like `label` or `h4`?

DesignerDave: Additionally, I think we get `role="header"` for free that way.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,79 @@
...
+ */
+function SideNavGroup({
+  children,
+  className,
+  headerText,
+  headerClassName,
+  ...rest
+}: SideNavGroupProps) {
+  return (
+    <li className={className} {...rest}>
```
DesignerDave: Perhaps I'm missing something, but we can probably pass className via `rest` here.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,79 @@
...
+  font-weight: bold;
+  text-transform: uppercase;
+  padding: 5px ${LEFT_RIGHT_OFFSET}px 5px ${LEFT_RIGHT_OFFSET}px;
+`;
+
+interface SideNavGroupProps {
+  /**
+   * Content that will appear inside of SideNavGroup component.
+   */
+  children: React.ReactNode;
```
DesignerDave: Is there a reason this is a required prop?

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+import { ulStyleOverrides, LEFT_RIGHT_OFFSET } from './styles';
+
+const sideNavLabelStyle = css`
```
DesignerDave: If we already have styles split from the component, should this class live in the styles file instead?

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,79 @@
...
+  display: block;
```
DesignerDave: We should get this for free so long as we're using a div or header. This is fine to keep if we switch to a label tag I believe.

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+const containerStyleOverrides = css`
+  padding: 10px ${LEFT_RIGHT_OFFSET}px 10px ${LEFT_RIGHT_OFFSET}px;
+  min-height: 0;
+  border-radius: 5px;
+
+  &:before {
+    content: normal;
+  }
+
+  &:focus:active {
```
DesignerDave: It would be nice to implement this focus state conditionally with the [useUsingKeyboardContext](https://github.com/mongodb/leafygreen-ui/tree/master/packages/leafygreen-provider#useusingkeyboardcontext) hook from the leafygreen-provider package.

This way applications with the provider can benefit from better focus state handling (and it gracefully degrades when the provider isn't a parent).

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+`;
+
+const titleTextStyleOverrides = css`
+  font-size: 15px;
+  text-transform: capitalize;
+`;
+
+const disabledStyleOverrides = css`
+  background-color: ${uiColors.white};
```
DesignerDave: We should probably set to transparent (assuming the intent is for it to not have a visible background). Additionally, we probably should change the color to a lighter gray to make disabled states more apparent.

packages/side-nav/src/SideNav.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { ulStyleOverrides } from './styles';
+
+const sideNavStyles = css`
```
DesignerDave: You might consider inlining this single-liner to improve readability (not that it's terrible now).

packages/side-nav/src/SideNav.tsx
```diff
@@ -0,0 +1,34 @@
...
+  width: 200px;
+`;
+
+interface SideNavProps {
+  className?: string;
+  children: React.ReactNode;
```
DesignerDave: The props are currently pretty simple, but probably worth adding TSDoc annotations here for a consistent DX.

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+    content: normal;
+  }
+
+  &:focus:active {
+    background-color: ${uiColors.gray.light2};
+  }
+`;
+
+const titleTextStyleOverrides = css`
+  font-size: 15px;
```
DesignerDave: This is a larger font size than in the prototype

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+import { MenuItem, MenuItemProps } from '@leafygreen-ui/menu';
```
DesignerDave: So there's one main reason I think we might consider not using MenuItem as the base here: We lose a lot of control over the transition between states of the MenuItem.

I think with the design of these, we have a great micro-interaction opportunity in animating the scale of the background between different states, but (barring some hacky implementation), those animations will be difficult to add without full control of the rendered DOM.

Additionally, there's the added benefit of not needing to modify MenuItem for this component which is nice for two reasons:
1. We don't need to expose a className prop for MenuItem, which we want to avoid style overrides on in general.
2. There won't be conflicts with Brooke's work.

Open to thoughts on this, of course.

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { LEFT_RIGHT_OFFSET } from './styles';
+
+// style overrides for menu item
+const containerStyleOverrides = css`
+  padding: 10px ${LEFT_RIGHT_OFFSET}px 10px ${LEFT_RIGHT_OFFSET}px;
+  min-height: 0;
+  border-radius: 5px;
+
+  &:before {
+    content: normal;
```
DesignerDave: Is this to unset the left bar in the MenuItem? I think `initial` is preferable, if so.

packages/side-nav/package.json
DesignerDave: @brookescarlett Is there a reason it's in the boilerplate? I don't think every component uses this.

brookescarlett: ^ yes, but can make a separate PR quickly to update

packages/side-nav/src/SideNavGroup.tsx
brookescarlett: :+1 @DesignerDave

.gitignore
hswolff: Nope, looks grrrreat!

packages/side-nav/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
hswolff: We need to add a reference to this project in the root `tscnofig.json` file so that when we run `yarn ts` from the root it picks up and properly type checks this package.

.changeset/two-mirrors-glow.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/side-nav': major
+'@leafygreen-ui/menu': minor
+---
+
+Mongo Menu now adds a class to style title text. Side Nav has been introduced as a new component.
```
hswolff:
```suggestion
MongoMenu now adds a class to style title text. SideNav has been introduced as a new component.
```

To mirror the component names themselves?

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+  disabled = false,
+  className,
+  titleTextClassName,
+  children,
+  description,
+  href,
+  ...rest
+}: MenuItemProps) {
+  if (active) {
+    rest['aria-current'] = 'page';
```
hswolff: Shouldn't be modifying the props value.

seanmhanson: I landed on a better solution that is more usable as well.

packages/side-nav/src/SideNavGroup.tsx
seanmhanson: I wanted to use a h tag, but didn't know what level you all use for what. H4 it is!

packages/side-nav/src/SideNavGroup.tsx
seanmhanson: Oh gosh, I completely forgot about that and just pulled everything out by habit

packages/side-nav/src/SideNavGroup.tsx
seanmhanson: Nope, I think it can be optional, if you want to use a self closing tab. However, what would that mean in terms of the UI? A group with no members behind a label?

packages/side-nav/src/SideNavGroup.tsx
seanmhanson: I only split out common styles (anything reused). Whenever possible otherwise, I try to keep styles alongside components.

packages/side-nav/src/SideNav.tsx
seanmhanson: Yup, part of my current work is consistency amongst all forms of docs

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: I'm sure we'll have stricter guidelines in the future, but for the time being this should do!

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: In theory – I could go either way here. I suppose there isn't a lot of importance to typing it either way, just that it's perfectly valid to not pass children.

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: That's fair enough!

packages/side-nav/src/SideNavGroup.tsx
seanmhanson: H4 it is, so we'll get this and bold for free, but pay for it in margin top and bottom.

packages/side-nav/src/SideNavItem.tsx
seanmhanson: The lighter color is already taken care of by the default styles for `MenuItem`, but I've updated the background.

packages/side-nav/src/SideNavItem.tsx
seanmhanson: It is meant to unset the pseudo-elements, but MDN specifically states that this is the intended usage of `normal` (see [refs here](https://developer.mozilla.org/en-US/docs/Web/CSS/content#values))

packages/side-nav/src/SideNavItem.tsx
seanmhanson: Updated to 14px.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: This is more of a preference, but usually I use `initial` rather than the default value to clearly communicate the intent of the rule to set it to its default value. Without looking at documentation, that's unclear here unless you're intimately familiar with the content property.

That said, I don't feel strongly about this. Either value should function the same here.

seanmhanson: This will be addressed as part of removing MenuItem as a dependency for SideNavItem

packages/menu/src/MenuItem.tsx
seanmhanson: This will be addressed as part of removing MenuItem as a dependency for SideNavItem. I'll revert the changes I made beyond pulling out the repeated types.

packages/menu/src/MenuItem.tsx
seanmhanson: This will be addressed as part of removing MenuItem as a dependency for SideNavItem, so no worries

packages/lib/src/index.ts
```diff
@@ -59,3 +59,24 @@ export const keyMap = {
   Space: 32,
   Tab: 9,
 };
+
+/**
+ * An enum of accepted values for the "aria-current" attribute, used for
```
brookescarlett: 👍

packages/side-nav/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+![npm (scoped)](https://img.shields.io/npm/v/@leafygreen-ui/side-nav.svg)
+
+## Example
+
+```js
+import { SideNav, SideNavGroup, SideNavItem } from '@leafygreen-ui/sideNav';
```
brookescarlett: the package name should be side-nav

packages/side-nav/src/SideNav.tsx
```diff
@@ -0,0 +1,47 @@
...
+  </SideNavGroup>
+</SideNav>
+ * ```
+ * @param props.className Class name that will be applied to SideNav root component.
+ * @param props.children Content that will appear inside of SideNav root component.
+ *
+ */
+function SideNav({ children, ...rest }: SideNavProps) {
+  return (
+    <nav style={{ width: '200px' }} aria-label="side-nav" {...rest}>
```
brookescarlett: can we move this to a className with our emotion instance

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,285 @@
...
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+
+  let Root: ElementType<any>;
+
+  if (usesLinkElement(props)) {
+    Root = RootComponentTypes.Link;
+  } else if (usesCustomElement(props)) {
+    Root = props[RootComponentTypes.Custom];
+  } else {
+    Root = RootComponentTypes.Div;
```
brookescarlett: I think we should probably default to a button here. Since this is a navigation component, it's fair to assume that most, if not all, of the items should respond to a click.

packages/side-nav/src/SideNav.tsx
seanmhanson: Haha, @DesignerDave asked the opposite and said to inline it, so I got rid of that. I actually like to have it as a class instead and it's mostly a nitpick, so I'm going to revert it back. Sorry Dave!

brookescarlett: I think maybe he meant `<nav className={css`width: 200px`} />` but perhaps I'm wrong. Either way definitely support using className rather than style 👍

packages/menu/src/MenuItem.tsx
```diff
@@ -143,12 +143,10 @@ interface SharedMenuItemProps {
    * Class name that will be applied to root MenuItem element.
    */
   className?: string;
-
```
hswolff: nit, I prefer the extra white space between the definitions as it gives space for my eyes to rest before reading the next prop name.

Having no line break makes it feel like a run-on sentence.

packages/side-nav/package.json
```diff
@@ -0,0 +1,22 @@
+
+  {
+    "name": "@leafygreen-ui/side-nav",
+    "version": "1.0.0",
```
hswolff: nit, I believe we have to have this be initialized to 0.9.0 so that the changelog script correctly runs the first time we use it. Gonna let @brookescarlett weigh in.

packages/menu/src/MenuItem.tsx
hswolff: That being said, I don't feel strong enough to desire a code change, but just adding my opinion into the mix.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,78 @@
...
+ */
+function SideNavGroup({
+  headerText,
+  headerClassName,
+  children,
+  ...rest
+}: SideNavGroupProps) {
+  return (
+    <li {...rest}>
+      {headerText && (
```
hswolff: In an effort to mildly premature optimize.....I wonder if we should rename this prop to `header` and allow it to to be any valid react node type. If it's a string it does what's here, but if it's not then it just renders itself.

Thinking is instead of us someday adding a `tooltipNode` prop in the future we just expose that additional control. 

Happy to be pushed back on this, @DesignerDave usually does :).

packages/side-nav/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import SideNav from './SideNav';
+import SideNavItem, { SideNavItemProps } from './SideNavItem';
+import SideNavGroup from './SideNavGroup';
+
+export { SideNav, SideNavItem, SideNavGroup, SideNavItemProps };
```
hswolff: I'm wondering if rather than add this to our public export API we just let the test import the props directly from the module.

packages/side-nav/src/SideNav.spec.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+  let renderedEls: RenderedElements = {};
+
+  afterEach(() => {
+    document.body.innerHTML = '';
+    renderedEls = {};
+    cleanup();
+  });
+
+  describe('Side Nav Item', () => {
```
hswolff: Let's have these mirror the component names. Should make grepping easier.

packages/menu/src/MenuItem.tsx
seanmhanson: Definitely agree actually, and I've made those changes. Things changed as a result of copy/pasting when trying to make sure I used the same descriptions across the typescript, the docstring, and the readme.

packages/side-nav/src/index.ts
seanmhanson: Clever. I didn't think of using two import statements like that.

packages/side-nav/src/SideNavGroup.tsx
seanmhanson: I wanted to do this, and opted not to because of premature optimization concerns when I was trying to get this out the door. I agree though and am going to change this. It also simplifies our props, as we don't need to take a custom class anymore when you can just directly pass in the header.

I will have to think a little bit about this and ensuring there is either an `h*` tag or `role="header"` somewhere.

seanmhanson:
@hswolff RFAL if you continue to sneakily do reviews on a Saturday

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -42,23 +41,24 @@ interface SideNavGroupProps {
...
- *
  */
-function SideNavGroup({
-  headerText,
-  headerClassName,
-  children,
-  ...rest
-}: SideNavGroupProps) {
+function SideNavGroup({ header, children, ...rest }: SideNavGroupProps) {
+  const renderHeader = () => {
```
hswolff: micro opt, rather than incurring a new function call cost we could do:

```jsx
const renderedHeader = typeof header === 'string' ? <h4 className={cx(sideNavLabelStyle)}>{header}</h4> : header;
```

Could even inline it.

packages/menu/src/MenuItem.tsx
```diff
@@ -88,7 +88,7 @@ const titleTextStyle = css`
   }
 `;
 
-const activetitleTextStyle = css`
+const activeTitleTextStyle = css`
```
DesignerDave: Brooke fixed both of the typos in this file in her PR. Let's omit these changes so we keep the concerns separated.

.changeset/clever-dolls-know.md
```diff
@@ -0,0 +1,7 @@
+---
+'@leafygreen-ui/side-nav': major
+'@leafygreen-ui/lib': minor
+'@leafygreen-ui/menu': patch
```
DesignerDave: As noted below, let's omit this altogether to keep from any conflicts with Brooke's PR.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,78 @@
...
+ *   If a string is provided, it will be rendered with default styling as a header tag.
+ * @param props.children Class name that will be applied to the component's header.
+ */
+function SideNavGroup({ header, children, ...rest }: SideNavGroupProps) {
+  return (
+    <li {...rest}>
+      {typeof header === 'string' ? (
+        <h4 className={sideNavLabelStyle}>{header}</h4>
+      ) : (
+        header
```
DesignerDave: Hmm... I have mixed feelings about not rendering the H4 when the prop isn't a string. Reason being is that if anyone needed to include an icon or span in the header, they'd have to duplicate the header styles entirely...

packages/menu/src/MenuItem.tsx
seanmhanson: This was a complete oversight (ditto the mention of it in the changeset file). Thanks for noticing this.

packages/side-nav/src/SideNavGroup.tsx
seanmhanson: Definitely, good plan. Theoretically if someone wants a custom header without those styles, they should just pass it in as part of `children`.

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,301 @@
...
+  padding: 10px ${LEFT_RIGHT_OFFSET}px 10px ${LEFT_RIGHT_OFFSET}px;
+  min-height: 0;
+  display: flex;
+  flex-direction: column;
+  border: none;
+  border-radius: 5px;
+  cursor: pointer;
+  text-align: left;
+  font-family: Akzidenz, ‘Helvetica Neue’, Helvetica, Arial, sans-serif;
+  transition: background-color 150ms ease-in-out;
```
DesignerDave: So I was trying to figure out why the scale transition doesn't work between states... We shouldn't care about background color on this element. The background color should be applied to the `:before` pseudo-class instead.

PR #235, opened 2020-01-23
PD-491: Make react-transition-group an external dep of the build

brookescarlett:
- Make react-transition-group an external dep
- Patch bump modal and popover

### DIFF
--- a/.changeset/large-zebras-exercise.md
+++ b/.changeset/large-zebras-exercise.md
@@ -1,0 +1,6 @@
+---
+'@leafygreen-ui/modal': patch
+'@leafygreen-ui/popover': patch
+---
+
+Make react-transition-group an external dependency of the build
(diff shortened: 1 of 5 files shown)


.changeset/large-zebras-exercise.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/modal': patch
+'@leafygreen-ui/popover': patch
+---
+
+Make react-transition-group an external dependency of the build
```
hswolff: This should be moved to a peer dependency then in the package.jsons for both these packages.

brookescarlett: 👍

packages/modal/package.json
```diff
@@ -12,12 +12,14 @@
   "publishConfig": {
     "access": "public"
   },
+  "peerDependencies": {
+    "react-transition-group": "^4.2.1"
```
hswolff: Might have to add this to the root package.json so it's still installed on first installations. If you remove node_modules and re-install does this still work?

PR #236, opened 2020-01-24
PD-492: Box component

brookescarlett:
## ✍️ Proposed changes

Created a Root component to handle logic determining how to render a component (button, anchor or custom component) with all the appropriate typing. We use this logic in many places throughout Leafy, so @seanmhanson  and I thought it might be wise to abstract it away. 

Would be great to have this in sooner than later, so that it can be used in @seanmhanson's SideNav PR as well as in my MongoNav component.

Adding @jetpacmonkey as a reviewer as he wrote the initial code in Button that was largely used to create the Root component.

🎟 _Jira ticket:_ [PD-492](https://jira.mongodb.org/browse/PD-492)

## 🛠 Types of changes

## ✅ Checklist


packages/root/README.md
```diff
@@ -0,0 +1,39 @@
...
+<a href="https://mongodb.design">I shall render as a link</a>
+```
+
+## Properties
+
+| Prop       | Type                                  | Description                                                                                                                     | Default  |
+| ---------- | ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| `as`       | `HTMLElement` or `React.ReactElement` | Takes priority in determining what tag the `<Root />` element will render                                                       | `button` |
+| `href`     | `string`                              | Determines the location that an anchor tag will point to. If prop is set without the `as` prop, `<Root />` will render `a` tags |          |
+| `children` | `React.ReactElement`                  | Content to appear inside of `<Root />` component                                                                                |          |
```
seanmhanson: In this one situation, since we're dealing with both components and HTML elements, I think we should be more explicit:

```suggestion
| `children` | `React.ReactElement` | Content to be rendered in an HTML element, or provided as a prop to the rendered component | `undefined` |

_Any other properties will be spread on the rendered HTML element or component._
```

packages/root/src/Root.story.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { text, select } from '@storybook/addon-knobs';
+import Root from '.';
+
+storiesOf('Root', module)
+  .add('Anchor', () => (
```
seanmhanson: Even though you include Button in the knobs select, I'd add a default state as well. I'd want to make sure all functionality is demonstrated, and then knobs further allows for more introspection.

```suggestion
  .add('Default, () => (
    <Root>
      By default, I am a button.
    </Root>
  ))
  .add('Anchor', () => (
```

packages/root/src/Root.story.tsx
```diff
@@ -0,0 +1,22 @@
...
+    <Root href="https://mongodb.design">
+      {text('Children', 'I am an anchor tag')}
+    </Root>
+  ))
+  .add('Custom', () => {
+    const CustomRoot = select(
+      'div',
+      { div: 'div', span: 'span', button: 'button' },
```
seanmhanson: I would specifically create or import a full component for use with this story (maybe `@leafygreen-ui/badge` since it is simple?) and offer `div`, `span`, and `Badge` instead. This would also avoid any confusion about needing to explicitly pass in `as="button"` since right now that's the default behavior.

packages/root/src/Root.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * # Root
+ *
+ * Root component
+ *
+ * ```
+<Root href="https://mongodb.design">Anchors Away!</Root>
+```
+ * @param props.children Takes priority in determining what tag the `<Root />` element will render.
+ * @param props.href Determines the location that an anchor tag will point to. If prop is set without the `as` prop, `<Root />` will render `a` tags.
+ * @param props.as Callback to change the open state of the Menu.
```
seanmhanson: The `@param props.children` and `@param props.as` are incorrect here. I would also update these to match `README.md` exactly whenever possible.

packages/root/src/Root.tsx
```diff
@@ -0,0 +1,65 @@
...
+ */
+function Root<T>(props: RootProps<T>) {
+  const { children, ...rest } = props;
```
seanmhanson: We should strip off any props that are specific to `Root<T>` only (i.e. `as`). Otherwise they'll be passed in when we use `{...rest}`.

```suggestion
  const { children, as, ...rest } = props;
```
and then just update the reference below from `props.as` to `as`.

packages/root/src/Root.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+type RootProps<T> = ButtonProps<T> | Anchor<T> | CustomElementProps<T>;
+
+function usesCustomElement<T>(
+  props: RootProps<T>,
+): props is CustomElementProps<T> {
+  return (props as any).as != null;
+}
+
+function usesLinkElement<T>(props: RootProps<T>): props is Anchor<T> {
```
seanmhanson: [nitpick] super nitpicky, can we change this to `usesAnchorElement` because there is already a `Link` element (used in `<head>` though, not `<body>`)>?

packages/root/src/Root.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { render, cleanup } from '@testing-library/react';
+import Root from '.';
+
+afterAll(cleanup);
+
+describe('packages/root', () => {
+  describe('by default', () => {
+    const props = { name: 'testName' };
+    const content = 'button content';
+    const { getByTestId } = render(
```
seanmhanson: I've seen this throughout, but I'm uncomfortable with several `render` calls if we're not also taking care of the previously existing elements on the body. In this pattern, each time you call `render` it will be added to the same body, and the order of the tests (and render calls) will impact the results of the test. 

As a result, I prefer to be very strict about using `beforeEach` and `afterEach` similar to Mocha/Chai. That doesn't mean we need to sacrifice what React Testing Library does for us (heavily restricting us to UI actions instead of modifying State/Props/etc). It just means we have to be very deliberate about scope and ensuring every test runs in isolation. 

It can be tempting to do `const { getByTestId, rerender } = render(...)`, but I would push heavily against the use of `rerender` unless you are literally forcing a re-render to account for asynchronous stubbed changes, and jest may do a better job of handling that anyway.

So instead I would look for something like this, in pseudo-code:
```
describe packages/mine
  let <any variables to be referenced repeatedly in tests>;

  beforeEach
    set up any values that will be used throughout the test
    set up your stubs here so they're consistent and in one place

  afterEach
    document.body.innerHTML = "";
    do whatever you need to restore stubs
    cleanup() (clean up each time, rather than only at the very end of the suite like in an afterAll call)

  describe when rendering X component
    describe with Y state
      beforeEach
        render component with props for Y
      test
        assert something about component and state  
      describe and I do an action
        beforeEach
          do the action
        test
          assert that i did the action
    describe with Z state
      beforeEach
        render component with props for Z
      test
        assert something about component and state  
```

Let me know if you have questions about this.

packages/root/src/Root.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    test('it renders children inside of <Root /> tag', () => {
+      expect(button.textContent).toBe(content);
+    });
+
+    test('it renders other props passed, when they are set', () => {
+      expect((button as HTMLButtonElement).name).toBe('testName');
+    });
+  });
+
+  describe('when the href prop is set', () => {
```
seanmhanson: You've established a nice pattern here in the test of rendering and checking the component, the tag, and the expected prop handling. Since you use it three times here *(and since we'll likely have to expand the tests if we ever have other components we explicitly check for), you could abstract this to a list of the data to test and then map over it to run your tests.

For example, here's a good chunk of code that should get you started if you want:
```
const componentsToTest = [
  {
    name: 'anchor',
    tagName: 'a',
    testProps: { href: '...', target: '...' },
    expectedType: 'HTMLAnchorElement',
  },
    ...,
];
```

And then map over the data and write a set of tests like:

```
beforeEach(() => {
  const { getByTestId } = render(<Root {...testProps}>`${name} content`</Root>);
});

test(`it renders as a(n) ${name} element`, () => {
  const el = getByTestId(name);
  expect(el.tagName.toLowerCase()).toBe(tagName);
});

test(`it renders children inside of the ${name}`, () => {
  const el = getByTestId(name);
  expect(el.textContent).toBe(`${name} element`);
});

Object.keys(testProps).forEach(propKey => {
  test(`it correctly sets the ${propKey} element on the ${name} component`, () => {
    const el = getByTestId(name);
    expect((el as expectedType)[propKey]).toBe(testProps[propKey]);
  });
});
```

packages/root/src/Root.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </Root>,
+    );
+
+    const custom = getByTestId('custom');
+
+    test('it renders as a custom element when the as prop is set', () => {
+      expect(custom.tagName.toLowerCase()).toBe('div');
+    });
+
+    test('it renders as a custom component even when the href prop is set', () => {
```
seanmhanson: This is a duplicate test (you rendered the component with `href` set for both of these tests)

packages/root/package.json
```diff
@@ -0,0 +1,19 @@
+
+  {
+    "name": "@leafygreen-ui/root",
+    "version": "0.9.0",
```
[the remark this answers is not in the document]
brookescarlett: We have it at 0.9.0 so that when we run changeset version bump we are able to release at 1.0.0

packages/root/README.md
```diff
@@ -0,0 +1,39 @@
+# Root
```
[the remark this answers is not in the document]
brookescarlett: Makes sense, added 👍

packages/root/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+```html
+<a href="https://mongodb.design">I shall render as a link</a>
+```
+
+## Properties
+
+| Prop       | Type                                  | Description                                                                                                                     | Default  |
+| ---------- | ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| `as`       | `HTMLElement` or `React.ReactElement` | Takes priority in determining what tag the `<Root />` element will render                                                       | `button` |
```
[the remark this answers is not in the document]
brookescarlett: Taking your suggestion for the wording, but `as` is not a required prop

packages/root/README.md
```diff
@@ -0,0 +1,39 @@
...
+| `href`     | `string`                              | Determines the location that an anchor tag will point to. If prop is set without the `as` prop, `<Root />` will render `a` tags |          |
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/root/README.md
brookescarlett: 👍

packages/root/src/Root.spec.tsx
seanmhanson: Note: I used `document.body.innerHTML` to wipe everything, but it would probably be better to take `unmount` from the call to `render` and set a variable scoped for all of your tests and call `unmount && unmount();` in the root level `afterEach` call.

FWIW `unmount` will definitely trigger any hooks etc that occur at the end of the component lifecycle. What I did *should*, but I am a little nervous now that I think about it.

packages/root/src/Root.story.tsx
brookescarlett: 👍

packages/root/src/Root.tsx
brookescarlett: Ah thank you -- oversight when c/ping

packages/root/src/Root.tsx
brookescarlett: You can't destructure `as` without Typescript complaining, because the prop won't exist unless we're using a custom element so we have to check that it exists first.

packages/root/src/Root.tsx
brookescarlett: 👍

packages/root/src/Root.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * Root component
+ *
+ * ```
+<Root href="https://mongodb.design">Anchors Away!</Root>
+```
+ * @param props.children Takes priority in determining what tag the `<Root />` element will render.
+ * @param props.href Determines the location that an anchor tag will point to. If prop is set without the `as` prop, `<Root />` will render `a` tags.
+ * @param props.as Callback to change the open state of the Menu.
+ */
+function Root<T>(props: RootProps<T>) {
```
jetpacmonkey: I feel conflicted about calling this "Root". My first instinct would be for a `Root` component to go at the top level of your app or something. Since this defaults to rendering a `button`, maybe `BaseButton` would be a better name? Not 100% confident on that, but `Root` seems to imply a bunch of things that aren't true to me.

packages/root/src/Root.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+
+type ButtonProps<T> = HTMLElementProps<'button'> &
+  T & {
+    href?: undefined;
+  };
+
+type Anchor<T> = HTMLElementProps<'a'> &
```
jetpacmonkey: Should probably call this `AnchorProps`

packages/root/src/Root.tsx
```diff
@@ -0,0 +1,65 @@
...
+  T & {
+    href: string;
+  };
+
+type CustomElementProps<T> = T & {
+  as: React.ElementType<any>;
+  [key: string]: any;
+};
+
+type RootProps<T> = ButtonProps<T> | Anchor<T> | CustomElementProps<T>;
```
jetpacmonkey: We should probably export this type, right? So then when this gets reused by `Button` it can use it like...

```ts
type ButtonProps = RootProps<SharedButtonProps>
```

packages/root/src/Root.tsx
brookescarlett: i can understand that Root might be confusing. Thoughts @DesignerDave and @seanmhanson ?

packages/root/src/Root.tsx
brookescarlett: 👍

packages/root/src/Root.tsx
brookescarlett: Yes that makes so much sense... good call

packages/root/src/Root.tsx
seanmhanson: I think that even though it defaults to button, we shouldn't name it after button (half the usages of this would have a menuitem role, and half would have a button role, for example). I've referred to it as Root so often that I can't un-think that name right now, but I agree that Root falls into that category with Application, App, etc.

packages/root/package.json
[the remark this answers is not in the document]
seanmhanson: Oh I missed that you hadn't run it yet!

packages/root/README.md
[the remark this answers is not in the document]
seanmhanson: Oh, whoops, I meant an optional prop that doesn't have a default value. Sorry for the confusion. Either way I don't mind.

packages/root/src/Root.tsx
seanmhanson: Taken offline, but we can use Omit to peel off the property right before we actually spread the rest.

packages/root/src/Root.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * ```
+<Root href="https://mongodb.design">Anchors Away!</Root>
+```
+ * @param props.children Content to be rendered in an HTML element, or provided as a prop to the rendered component.
+ * @param props.href When provided, `<Root />` will render an anchor tag with this `href` value.
+ * @param props.as The component or HTML tag to be rendered by the `<Root />` component. **Note**: This will supersede the behavior of any other props.
+ */
+function Root<T>(props: RootProps<T>) {
+  const { children, ...rest } = props;
+  let Root: React.ElementType<any> = 'button';
```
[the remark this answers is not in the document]
seanmhanson: I made the case for `div` as well, but one difficulty for `button` is that the required props for a button (talking specifically about the html spec, not react etc) aren't unique. I think we can get around that by looking for the props that we specifically care about in our button usages, like checking for `onClick` and `name`, and checking for the omission of `href`. I'm not sure exactly what to do about that.

We also paired and came up with a second version of this that is more component agnostic and maintained a mapping of expected component to required/omitted props, but given the power of the `as` prop (or `component`), it seemed like overkill.

packages/root/src/Root.tsx
seanmhanson: I really like both the suggestion of `Box` and of `component` instead of `as`. I also prefer `div` as a default, but Brooke made a good case for `button` since this is more of an internal component.

brookescarlett: Should we call it Component? Reach UI has a Component component and maybe that's more clear? We can default to div if that makes people feel more comfortable we've just followed the Root/button pattern it all of the use-cases thus far

brookescarlett: I feel like when we make a box component it should have more grid/spacing/alignment properties built in @hswolff, but maybe I'm wrong here

packages/root/src/Root.spec.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import Root from '.';
+
+afterAll(cleanup);
+
+describe('packages/root-hi', () => {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/root/src/Root.tsx
hswolff: Naming this Box now doesn't preclude us from adding those props at a future point. The Box component would also have this functionality of the `component={}` prop so this gets us started down that path.

Might require a more rapid discussion on Slack.

brookescarlett: Renamed to Box!

brookescarlett: (with div as default, and component as slot instead of as)

packages/root/src/Root.tsx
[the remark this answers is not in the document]
brookescarlett: Updated 👍

packages/box/src/Box.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+import omit from 'lodash/omit';
+
+type DivProps<T> = HTMLElementProps<'div'> &
+  T & {
+    href?: undefined;
```
seanmhanson: I think you can remove `{ href?: undefined }` because this should only be the `div` props and the shared props. Presumably if the shared props include `href` or `component`, their union should be treated as `AnchorProps` or `CustomElementProps` anyway.

brookescarlett: Removing this line causes Typescript to complain in the `usesAnchorElement` function, because it believes that `href` should not be present in `BoxProps`. I think since we do a loose check of the value of `href` (null/undefined before determining whether or not the component should use AnchorProps) it's okay that we specifically say that `href` should not be defined in div props.

seanmhanson: If so, I think then we should probably add `component?: undefined` as well, to keep things consistent. Does that work? If not, no worries, I don't want this to block a merge.

brookescarlett: Nvm, fixed this!

packages/box/src/Box.tsx
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+type CustomElementProps<T> = T & {
+  component: React.ElementType<any>;
+  [key: string]: any;
+};
+
+export type BoxProps<T> = DivProps<T> | AnchorProps<T> | CustomElementProps<T>;
+
+function usesCustomElement<T>(
```
hswolff:
```suggestion
function isCustomElement<T>(
```

And ditto `isAnchorElement`?

WDYT?

packages/box/src/Box.spec.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const renderBox = (props: testProps = sharedProps) => {
+    const { queryByTestId, unmount } = render(
+      <Box data-testid="box" {...props}>
+        <div data-testid="child">Child Content</div>
+      </Box>,
+    );
+
+    renderedEls.child = queryByTestId('child');
+    renderedEls.box = queryByTestId('box');
+    unmountQueue.push(unmount);
```
hswolff: This seems un-necessarily complex. 

Are we testing anything to do with unmounting?

If not can e just remove this complexity and rely on the builtin cleanup mechanism?

packages/box/src/Box.tsx
brookescarlett: Works for me

packages/box/src/Box.spec.tsx
brookescarlett: @seanmhanson you wrote these tests, so defer to you for a response

packages/box/src/Box.story.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { text, select } from '@storybook/addon-knobs';
+import Box from '.';
+
+storiesOf('Box', module)
+  .add('Default', () => <Box>By default I am a div</Box>)
```
DesignerDave: I think it would be better to use knobs for demonstrating changing props.

packages/box/src/Box.tsx
```diff
@@ -0,0 +1,72 @@
...
+```
+ * @param props.children Content to be rendered in an HTML element, or provided as a prop to the rendered component.
+ * @param props.href When provided, `<Box />` will render an anchor tag with this `href` value.
+ * @param props.component The component or HTML tag to be rendered by the `<Box />` component. **Note**: This will supersede the behavior of any other props.
+ */
+
+const Box = React.forwardRef(
+  <T extends React.ReactNode>(props: BoxProps<T>, ref: React.Ref<any>) => {
+    const { children } = props;
+    const rest = omit(props as any, ['component', 'children']);
```
DesignerDave: Instead of bringing in lodash here, why not destructure the prop argument at the top?

```ts
const Box = React.forwardRef(
  <T extends React.ReactNode>({
    component,
    children,
    ...rest,
  }: BoxProps<T>, ref: React.Ref<any>) => {}
```

packages/box/src/Box.story.tsx
brookescarlett: 👍

brookescarlett: Discussed offline, leaving as is

packages/box/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "types": "./dist/index.d.ts",
+    "scripts": {
+      "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js"
+    },
+    "license": "Apache-2.0",
+    "publishConfig": {
+      "access": "public"
+    },
+    "devDependencies": {
+      "@leafygreen-ui/lib": "^4.0.0",
```
hswolff: These are dependencies.

packages/box/src/Box.spec.tsx
hswolff: I'd like to remove this so we don't have overly complex tests. Do the tests still pass with this removed?

packages/box/package.json
brookescarlett: 👍

packages/box/src/Box.spec.tsx
brookescarlett: Removed it

hswolff: Don't see it. 👀

brookescarlett: Failed to push, I see that now 🤦‍♂

brookescarlett:
To note, I upgraded Storybook in this commit because @DesignerDave and I were playing around with one of the new features -- I believe that's why there's such a discrepancy in yarn lock

packages/box/src/Box.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+      </Box>,
+    );
+
+    renderedEls.child = queryByTestId('child');
+    renderedEls.box = queryByTestId('box');
+  };
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    cleanup();
```
seanmhanson: OH! I didn't realize `cleanup` actually takes care of the unmounting for you. Good improvement, I should fix that in my previous tests! I can't believe I missed that in the docs.

PR #237, opened 2020-01-25
PD-492: Optional Test Changes

seanmhanson:
This isn't necessary to be approved, but I wanted to provide a changeset that updates the tests for this component so they better handle scoping, cleanup tasks, data-id consistency, etc. If you're interested, we can merge these changes into your branch (this PR is pointing at PD-492 for a base, not master). If not, you can just go through it and get from it whatever you like / take whatever is helpful

### DIFF
--- a/packages/root/src/Root.spec.tsx
+++ b/packages/root/src/Root.spec.tsx
@@ -1,77 +1,133 @@
 import React from 'react';
+import '@testing-library/jest-dom/extend-expect';
 import { render, cleanup } from '@testing-library/react';
 import Root from '.';
+import { RootProps } from './Root';
 
-afterAll(cleanup);
+// types
+interface SharedProps {
+  name?: string;
+}
 
-describe('packages/root-hi', () => {
-  const props = { name: 'testName' };
-  const content = 'button content';
+interface RenderedEls {
+  root?: HTMLElement | null;
+  child?: HTMLElement | null;
+}
+
+interface LinkWrapperProps {
+  href: string;
+  target: string;
+  children: React.ReactNode;
+}
+
+type testProps = RootProps<SharedProps>;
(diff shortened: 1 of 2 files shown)


packages/root/src/Root.spec.tsx
```diff
@@ -1,77 +1,133 @@
...
+    renderedEls.child = queryByTestId('child');
+    renderedEls.root = queryByTestId('root');
+    unmountQueue.push(unmount);
+  };
 
   afterEach(() => {
-    document.body.innerHTML = '';
+    while (unmountQueue.length > 0) {
+      const unmount = unmountQueue.pop();
+      unmount && unmount();
```
seanmhanson: Previously I had used `document.body.innerHTML = '';` to clear everything in the `side-nav` tests, but then I realized that it is probably best to allow the component to finish up anything that might be triggered on unmount, so I did this queue thing instead. ‾\_(ツ)_/‾

packages/root/src/Root.spec.tsx
```diff
@@ -1,77 +1,133 @@
...
+type testProps = RootProps<SharedProps>;
+type unmountQueueType = Array<() => boolean>;
+
+describe('packages/root', () => {
+  const unmountQueue: unmountQueueType = [];
+  const renderedEls: RenderedEls = {};
+  const linkWrapperFn = jest.fn();
+
+  function LinkWrapper({ href, target, children, ...rest }: LinkWrapperProps) {
+    linkWrapperFn();
```
seanmhanson: This is a subtle but to me important, to show that in fact we actually did render the component properly (because `linkWrapperFn` will wind up getting called in the process), rather than only checking for the output.

brookescarlett:
Looks great definitely happy to merge this into PD-492. Thank you again for doing this.. @seanmhanson

PR #239, opened 2020-01-28
Improve SideNav Accessibility

DesignerDave:
## ✍️ Proposed changes

This increases the contrast of labels in the SideNav components, as well as improves handling of Aria roles.

## 🛠 Types of changes

## ✅ Checklist

- [ ] ~I have added my new package to the global tsconfig~
- [ ] ~I have added my new package to the Table of Contents on the global README~

### DIFF
--- a/packages/side-nav/README.md
+++ b/packages/side-nav/README.md
@@ -47,7 +47,7 @@ return (
     <li>
       <h4 class="leafygreen-ui-19w7vet">States</h4>
       <ul role="menu" class="leafygreen-ui-bdnco">
-        <li role="menuitem">
+        <li role="none">
           <button
             data-leafygreen-ui="side-nav-item-container"
             class="leafygreen-ui-d18uaj"
@@ -58,7 +58,7 @@ return (
             <div class="leafygreen-ui-1mikaeo">Active State</div>
           </button>
         </li>
-        <li role="menuitem">
+        <li role="none">
           <button
             data-leafygreen-ui="side-nav-item-container"
             class="leafygreen-ui-g84goy"
@@ -78,7 +78,7 @@ return (
         <small>(content header)</small>
       </div>
       <ul role="menu" class="leafygreen-ui-bdnco">
-        <li role="menuitem">
+        <li role="none">
           <button
             data-leafygreen-ui="side-nav-item-container"
             class="leafygreen-ui-1b8asku"
@@ -91,7 +91,7 @@ return (
             </div>
           </button>
         </li>
-        <li role="menuitem">
+        <li role="none">
           <a
             href="/"
             data-leafygreen-ui="side-nav-item-container"
@@ -103,7 +103,7 @@ return (
             <div class="leafygreen-ui-f3d0a6">Anchor root element</div>
           </a>
         </li>
-        <li role="menuitem">
+        <li role="none">
           <label
             for="docs-input"
             data-leafygreen-ui="side-nav-item-container"
--- a/packages/side-nav/README.md
+++ b/packages/side-nav/README.md
@@ -2,6 +2,20 @@
 
 ![npm (scoped)](https://img.shields.io/npm/v/@leafygreen-ui/side-nav.svg)
 
+## Installation
+
+### Yarn
+
+```shell
+yarn add @leafygreen-ui/side-nav
+```
+
+### NPM
+
+```shell
+npm install @leafygreen-ui/side-nav
+```
+
 ## Example
 
 ```js
--- a/packages/side-nav/src/SideNav.spec.tsx
+++ b/packages/side-nav/src/SideNav.spec.tsx
@@ -64,6 +64,7 @@ describe('packages/side-nav', () => {
 
       test('it sets the correct role for the item', () => {
         expect(renderedEls.itemEl).toHaveAttribute('role', 'menuitem');
+        expect(renderedEls.itemEl?.parentNode).toHaveAttribute('role', 'none');
       });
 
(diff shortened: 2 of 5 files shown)


packages/side-nav/README.md
```diff
@@ -47,7 +47,7 @@ return (
     <li>
       <h4 class="leafygreen-ui-19w7vet">States</h4>
       <ul role="menu" class="leafygreen-ui-bdnco">
-        <li role="menuitem">
+        <li role="none">
           <button
```
brookescarlett: Shouldn't the buttons then have role="menuitem" ? 
See HTML source doe here: https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-links.html

brookescarlett:
The implementation in the .tsx file looks correct to me, but the menuitem roles are missing in the markdown

packages/side-nav/README.md
seanmhanson: Oops, @DesignerDave good catch. From W3's spec, an example:
```
<li role="none">
  <a role="menuitem" href="https://www.w3.org/standards/webdesign/accessibility">
    W3C Web Accessibility Initiative
  </a>
</li>
```
So by default if `<ul role="menu">` is your container, `li` elements treated as menu items since they're direct descendants. Since the actual interactive element is the link/button/whatever, we have to reset the role back to `role="none"`.

Does that sound right?

packages/side-nav/src/SideNav.spec.tsx
```diff
@@ -63,10 +63,7 @@ describe('packages/side-nav', () => {
       });
 
       test('it sets the correct role for the item', () => {
-        expect(renderedEls.itemEl?.parentNode).toHaveAttribute(
-          'role',
-          'menuitem',
-        );
+        expect(renderedEls.itemEl).toHaveAttribute('role', 'menuitem');
```
seanmhanson: I would prefer this block:
```
test('it sets the correct role for the item and its container', () => {
   expect(renderedEls.itemEl).toHaveAttribute('role', 'menuitem');
   expect(renderedEls.itemEl?.parentNode).toHaveAttribute('role', 'none');
});
```

Just so the test coverage maintains this in the future, otherwise I could see myself or someone else modifying this without thinking once we've all stopped thinking so much about IA Navigation.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -5,9 +5,10 @@ import { uiColors } from '@leafygreen-ui/palette';
 import { ulStyleOverrides, SIDE_OFFSET } from './styles';
 
 const sideNavLabelStyle = css`
-  color: ${uiColors.green.base};
+  font-size: 11px;
```
seanmhanson: So this is more push-back against the design than this PR, but looking at this in context right now on Cloud, `11px` is borderline legible. I really think `12px` is the smallest we should go for something that will show up on nearly every page of every application.

Example: I have a moderate prescription for glasses (nearsighted + astigmatism), and without my glasses the letters of the side nav group header at 11px start to blur into each other, about one foot from the monitor.

Take it or leave it, just a thought.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -5,9 +5,10 @@ import { uiColors } from '@leafygreen-ui/palette';
...
+  font-weight: bold;
```
seanmhanson: Thank you for the drive-by on behalf of legacy less code

packages/side-nav/README.md
DesignerDave: Ahh, yep! Sorry, I believe this is implemented correctly, just forgot to add the role in the README.

packages/side-nav/src/SideNav.spec.tsx
DesignerDave: Great, will do.

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: 👍

packages/side-nav/README.md
brookescarlett: Shouldn't the button have role="menuitem"?

brookescarlett: I only see the role set on the LI

brookescarlett:
@DesignerDave just noticed that the readme doesn't have the new Installation instructions we added everywhere else, if you don't mind adding while you're in here

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: Rob is working on typography guidelines which will give us the justification to change this. As of now, we should probably stick with the mocks on size, though I did work with Nelle to increase the contrast.

We would need to make some larger modifications to the spacing around the label to maintain the visual hierarchy to make the change to a larger font-size, but I do agree with you. I think we should go forward with this.

That said, if we add `letter-spacing: 0.3px;` to the component, does this address your concerns at all?

seanmhanson: Sure thing. If this is being addressed overall with guidelines, I'm good with whatever we need for now.

packages/side-nav/README.md
DesignerDave: I'm seeing it as an attribute of the button in the PR. It was already there prior, so it's not showing up as a change.

brookescarlett: got it.. the compare view was cutting a bit out so I didn't notice

PR #248, opened 2020-02-04
PD-262: Add Glyph Prop to Button Component

dominoweir:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-262](https://jira.mongodb.org/browse/PD-262)

## 🛠 Types of changes

- [X] New feature (non-breaking change which adds functionality)

## ✅ Checklist

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/packages/button/src/Button.spec.tsx
+++ b/packages/button/src/Button.spec.tsx
@@ -148,6 +148,6 @@ describe('packages/Button', () => {
       throw new Error('No element was rendered');
     }
 
-    expect(buttonComponent.getAttribute('glyph')).not.toBe(null);
+    expect(buttonComponent.innerHTML).toContain('svg');
   });
 });
--- a/packages/button/src/Button.spec.tsx
+++ b/packages/button/src/Button.spec.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { render, fireEvent, cleanup } from '@testing-library/react';
 import { typeIs } from '@leafygreen-ui/lib';
+import Icon from '@leafygreen-ui/icon';
 import Button from './Button';
 
 afterAll(cleanup);
@@ -137,5 +138,15 @@ describe('packages/Button', () => {
     }
 
     expect(buttonComponent.tagName.toLowerCase()).toBe('div');
+  });
+
+  test(`renders a button with an arbitrary element passed in glyph prop`, () => {
+    const glyph = render(
+      <Button glyph={<Icon glyph="Trash" />}>My Trash Button</Button>,
+    ).container.firstChild;
+
+    if (!typeIs.element(glyph)) {
+      throw new Error('No element was rendered');
+    }
   });
 });
--- a/packages/button/src/Button.spec.tsx
+++ b/packages/button/src/Button.spec.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { render, fireEvent, cleanup } from '@testing-library/react';
 import { typeIs } from '@leafygreen-ui/lib';
-import Icon from '@leafygreen-ui/icon';
 import Button from './Button';
 
 afterAll(cleanup);
--- a/packages/button/src/Button.spec.tsx
+++ b/packages/button/src/Button.spec.tsx
@@ -141,12 +141,14 @@ describe('packages/Button', () => {
   });
 
   test(`renders a button with an arbitrary element passed in glyph prop`, () => {
-    const glyph = render(
-      <Button glyph={<Icon glyph="Trash" />}>My Trash Button</Button>,
+    const buttonComponent = render(
+      <Button glyph={<svg />}>My Trash Button</Button>,
     ).container.firstChild;
 
-    if (!typeIs.element(glyph)) {
+    if (!typeIs.element(buttonComponent)) {
       throw new Error('No element was rendered');
     }
+
+    expect(buttonComponent.getAttribute('glyph')).not.toBe(null);
   });
 });
--- a/packages/button/src/Button.story.tsx
+++ b/packages/button/src/Button.story.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { text, select, boolean } from '@storybook/addon-knobs';
 import { css } from '@leafygreen-ui/emotion';
+import Icon from '@leafygreen-ui/icon';
 import Button, { Size, Variant } from '.';
 
 const buttonClass = css`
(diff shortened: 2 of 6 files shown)


packages/button/src/Button.story.tsx
```diff
@@ -45,6 +46,11 @@ storiesOf('Buttons', module)
       { div: 'div', span: 'span', button: 'button' },
       'div',
     );
+    const selectedGlyph = select(
```
brookescarlett: I know that @DesignerDave might disagree with me on this one, but think it could be worth either breaking this out into its own story or bringing this logic into the `default` tab. This behavior isn't related to whether or not it's a custom component, so I think it might be confusing to group this here.

packages/button/src/Button.spec.tsx
```diff
@@ -138,4 +139,14 @@ describe('packages/Button', () => {
 
     expect(buttonComponent.tagName.toLowerCase()).toBe('div');
   });
+
+  test(`renders a button with an arbitrary element passed in glyph prop`, () => {
```
brookescarlett: What exactly are we testing here? I don't see an expect statement... Perhaps we could do something like 

`const { getByTestId } = render( <Button glyph={<Icon glyph="Trash" />} data-testid="trash-button">My Trash Button</Button>)` and checking to make sure that the trash-button contains the Icon?

packages/button/src/Button.tsx
```diff
@@ -333,6 +352,13 @@ const Button = React.forwardRef((props: ButtonProps, forwardRef) => {
     'children',
   ]);
 
+  const modifiedGlyph =
```
brookescarlett: Thinking that maybe when we modify the glyph we should also change the size of the Icon based on the size of the button. What are your thoughts on this @DesignerDave?

packages/button/src/Button.spec.tsx
dominoweir: Dave recommended that I add a test checking if something renders when the glyph prop is set to a value other than null. I don't know that I did it correctly as my jest knowledge is a little rusty. I can definitely use your suggestion instead.

brookescarlett: Feel free to test for that, just be sure to add an assertion statement :)

dominoweir: Ok, went ahead and fixed it. Thanks for catching that!

brookescarlett: 👍

packages/button/src/Button.tsx
DesignerDave: I think let's stick with the mocks on this one. We aren't changing the font size either I believe.

DesignerDave: When we do the flat button upgrade, we can consider making larger changes.

packages/button/src/Button.spec.tsx
```diff
@@ -138,4 +138,16 @@ describe('packages/Button', () => {
...
+    const buttonComponent = render(
+      <Button glyph={<svg />}>My Trash Button</Button>,
+    ).container.firstChild;
+
+    if (!typeIs.element(buttonComponent)) {
+      throw new Error('No element was rendered');
+    }
+
+    expect(buttonComponent.getAttribute('glyph')).not.toBe(null);
```
brookescarlett: Right now this test is testing if the buttonComponent as rendered in the HTML contains a glyph attribute, which is currently occurring but is not the behavior that we want. I'd suggest that first you make sure that we're not spreading the glyph prop along with {...rest} on the Button component and make sure that this test fails. Then, add another test (perhaps checking that the buttonComponent.innerHTML contains <svg>, but feel free to get more creative than that) to ensure that you're testing that we are rendering the desired svg.

dominoweir: Ok, I went ahead and modified the component and the test to account for this. I'm now checking if the component's innerHTML contains "svg" instead.

brookescarlett: :+1

packages/button/src/Button.tsx
brookescarlett: Kk sounds good to me

packages/button/src/Button.story.tsx
brookescarlett: Still think it makes sense to move this out of custom component. Can be in glyph or it's own story up to you

dominoweir: ok, this logic has now been moved to the default tab

brookescarlett: 👍

packages/button/src/Button.tsx
```diff
@@ -348,8 +375,11 @@ const Button = React.forwardRef((props: ButtonProps, forwardRef) => {
           //
           // https://www.w3.org/TR/css-flexbox-1/#painting
           z-index: 1;
+          display: flex;
+          align-items: center;
```
DesignerDave: Given we have several CSS properties here now, we should move this from being an inline style declaration to a variable.

Additionally, would `display: inline-flex` be a better change from the default of `display: inline`? Or is there a particular reason we need this to have block sizing?

packages/button/src/Button.story.tsx
```diff
@@ -34,6 +35,11 @@ storiesOf('Buttons', module)
           null,
         ) as any
       }
+      glyph={
+        <Icon
+          glyph={select('Glyph', Object.keys(glyphs) as Array<any>, 'Edit')}
```
DesignerDave: We probably should handle the glyph knob the same way in both stories (e.g. as a variable or inline).

I also _think_ that TypeScript should actually infer the type for `Object.keys(glyphs)` as being something like `Array<keyof typeof glyphs>`, so we might not need to typecast it.

packages/button/src/Button.tsx
```diff
@@ -331,8 +350,16 @@ const Button = React.forwardRef((props: ButtonProps, forwardRef) => {
     'size',
     'variant',
     'children',
+    'glyph',
   ]);
 
+  const modifiedGlyph =
+    glyph && children
+      ? React.cloneElement(glyph, {
+          className: cx({ [glyphMargins[size]]: glyph !== null }),
```
DesignerDave: It looks like `glyph` can be undefined, so we should use `glyph != null` instead to handle that case.

packages/button/src/Button.story.tsx
dominoweir: You are correct, it does not need to be typecast. The glyph knob no longer appears in both stories- Brooke said to just have it in the default component. I wrote it inline because that's how the rest of the knobs were written for the default component, but I can definitely change that if you would prefer to see it in a variable.

PR #250, opened 2020-02-05
PD-497: API Request in MongoNav

brookescarlett:
## ✍️ Proposed changes

Two things happening in this PR: 
- Broke out MongoSelect into ProjectSelect and OrgSelect (which will help down the line when we need just ProjectSelect exported) 
- Added API requests when mode is set to Production (while keeping fixture data for Dev mode async)

🎟 _Jira ticket:_ [PD-497](https://jira.mongodb.org/browse/PD-497)

## 🛠 Types of changes

## ✅ Checklist

*A big test PR will come soon, I promise*


packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +121,60 @@ export default function MongoNav({
...
+      .catch(error => {
+        onError?.(error);
+        console.warn(error);
+      });
   }
 
-  if (!data) {
-    // Eventually this logic will be more robust, but for an alpha version will return null without data
-    return null;
+  function getFixtureData() {
```
DesignerDave: This was mostly for the sake of testing our handling without making actual fetch requests. I think we should opt to simply consume the fixture data directly instead of making a promise and having random errors.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +121,60 @@ export default function MongoNav({
...
+  const [data, setData] = React.useState<DataInterface | undefined>(undefined);
+
+  const accountHost = hosts?.account ?? `https://account.mongodb.com`;
+  const cloudHost = hosts?.cloud ?? `https://cloud.mongodb.com`;
+  const universityHost = hosts?.university ?? `https://university.mongodb.com`;
+  const supportHost = hosts?.support ?? `https://support.mongodb.com`;
 
-  if (mode === Mode.Dev) {
-    data = devModeData;
+  function getProductionData() {
```
DesignerDave: We only use this function in one place. You might consider returning your fetch call from this function, and consuming the returned promise in your useEffect instead.

```js
const endpointUri = `${cloudHost}/user/shared`
const getProductionData = () => fetch(endpointUri, {
    credentials: 'include',
    method: 'GET',
})
```
```js
 useEffect(() => {
    if (mode === Mode.Dev) {
      getFixtureData().then(data => setData(data as DataInterface));
    } else {
      getProductionData()
        .then(response => response.json())
        .then(data => {
          setData(data);
          onSuccess?.(data);
        })
        .catch(error => {
          onError?.(error);
          console.error(error);
        });
      }
  }, []);
```

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +121,60 @@ export default function MongoNav({
...
+      method: 'GET',
+    })
+      .then(response => response.json())
+      .then(data => {
+        setData(data);
+        onSuccess?.(data);
+      })
+      .catch(error => {
+        onError?.(error);
+        console.warn(error);
```
DesignerDave: `console.error` is more appropriate here.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -144,7 +203,7 @@ export default function MongoNav({
     `${cloudHost}/v2#/${projectId}`;
   const constructProjectURL = constructProjectURLProp ?? defaultProjectURL;
 
-  const constructedUrls: Required<URLSInterface> = {
+  const sanitizedURLs: Required<URLSInterface> = {
```
DesignerDave: "sanitized" to me implies something different than what you mean here.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -12,30 +11,26 @@ import { css } from '@leafygreen-ui/emotion';
...
   ProjectInterface,
   OrganizationInterface,
   URLSInterface,
   CurrentProjectInterface,
   CurrentOrganizationInterface,
+  PlanType,
 } from '../types';
 
-const menuItemHeight = 36;
+// const menuItemHeight = 36;
```
DesignerDave: We can remove this

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -12,30 +11,26 @@ import { css } from '@leafygreen-ui/emotion';
...
 
-const listContainerStyle = css`
-  max-height: ${menuItemHeight * 5}px;
-  overflow-y: auto;
-`;
-
-const ulContainerStyle = css`
-  padding-left: 1px;
-`;
+// const listContainerStyle = css`
```
DesignerDave: Ditto here

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -55,11 +50,33 @@ const projectButtonStyle = css`
...
+  font-weight: bolder;
+  white-space: nowrap;
+`;
+
+const orgOptionContainerStyle = css`
+  display: flex;
+  justify-content: space-between;
+`;
+
+const formatPlanType = (planType: PlanType) => {
```
DesignerDave: We could simplify this by having an enum and consuming it directly:
```ts
const formattedPlanTypes: Record<PlanType, string> = {
  [PlanType.Atlas]: 'Atlas',
  [PlanType.Cloud]: 'Cloud Manager',
  [PlanType.OnPrem]: 'Ops Manager',
} as const;

const formattedAtlasType = formattedPlanTypes[PlanType.Atlas];
```

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -89,140 +104,137 @@ const onKeyDown: React.KeyboardEventHandler = e => {
...
-      />
-    );
+  const errorMessage =
+    'You do not belong to any organizations. Create an organization to start using MongoDB Cloud';
+
+  return (
+    <Menu
+      trigger={
+        <OrganizationTrigger
+          current={current?.orgName ?? 'All Organizations'}
```
DesignerDave: Should we just accept a `placeholder` prop, and if current is unset, render that?

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -89,140 +104,137 @@ const onKeyDown: React.KeyboardEventHandler = e => {
...
+      <FocusableMenuItem>
+        <Input
+          onChange={onChange}
+          onKeyDown={onKeyDown}
+          variant="organization"
+        />
+      </FocusableMenuItem>
+
+      <>
+        {data ? (
```
DesignerDave: This is probably a more succinct way to write this:
```js
<>
  data?.map(renderOrganizationOption) ?? <li>{errorMessage}</li>
</>
```

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -89,140 +104,137 @@ const onKeyDown: React.KeyboardEventHandler = e => {
...
     return (
       <MenuItem
-        key={datum.orgId}
+        key={projectId}
         className={menuItemContainerStyle}
         onClick={onClick}
+        href={constructProjectURL(orgId, projectId)}
       >
-        {content}
+        <span className={nameStyle}>{projectName}</span>
```
DesignerDave: Is there a reason we don't apply the `nameStyle` `className` directly to `MenuItem` rather than rendering a nested span?

packages/mongo-nav/src/mongo-select/Trigger.tsx
```diff
@@ -62,7 +58,7 @@ const border = css`
 
 interface OrganizationTriggerProps {
   children?: React.ReactNode;
-  current: CurrentOrganizationInterface;
+  current: string;
```
DesignerDave: Not sure I understand this change. Why did we make this type more broad?

packages/mongo-nav/src/Loading.tsx
```diff
@@ -0,0 +1,41 @@
...
+  justify-content: space-between;
+  width: 100%;
+  padding-left: 15px;
+  padding-right: 15px;
+  height: 45px;
+  box-shadow: 0 3px 7px 0 rgba(67, 117, 151, 0.08);
+  overflow: hidden;
+`;
+
+export default function Loading() {
```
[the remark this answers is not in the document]
brookescarlett: Figured it'd be worth it to wait for now rather than adding messy conditional logic throughout OrgNav and ProjectNav (at least for now) but once we have the designs depending on how much we're omitting/still rendering we can figure out what makes the most sense (bringing this into its own component or not)

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: I can take out the random errors for sure but @hswolff thought it would be a better idea to have the fixtureData still resolve from a promise

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: we were "sanitizing" the hosts so I thought it was a good parallel, can change back for sure if you want

packages/mongo-nav/src/mongo-select/Trigger.tsx
brookescarlett: we were accepting an object, now just a string

brookescarlett: we don't need all of the other information from the CurrentOrganizationInterface object, simpler to just pass the name of the currentOrg

packages/mongo-nav/src/MongoNav.tsx
DesignerDave: Yeah, I don't think it's particularly helpful to randomly error in people's applications in dev mode.

packages/mongo-nav/src/Loading.tsx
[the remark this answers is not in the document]
DesignerDave: 👍

packages/mongo-nav/src/MongoNav.tsx
DesignerDave: Sanitization usually implies removal of unsafe characters and information, e.g. to prevent script injection attacks.

packages/mongo-nav/src/mongo-select/Trigger.tsx
DesignerDave: Shouldn't this be a string union of the possible values?

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: 😂

brookescarlett: That made me laugh, but removing it!

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/mongo-select/MongoSelect.tsx

packages/mongo-nav/src/mongo-select/MongoSelect.tsx

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
brookescarlett: Do you think different products are going to want to change this value? That would mean exposing two props one for Orgs and one for Projects.. Idk if it's necessary

brookescarlett:
Commented out failing test while I try and work through testing the async data delivery -- in the meantime, responded to your PR feedback @DesignerDave

packages/mongo-nav/src/mongo-select/MongoSelect.tsx

packages/mongo-nav/src/mongo-select/MongoSelect.tsx

packages/mongo-nav/src/mongo-select/Trigger.tsx
brookescarlett: It's whatever a user names their org, so could be anything

packages/mongo-nav/src/MongoNav.story.tsx
```diff
@@ -16,5 +16,6 @@ storiesOf('MongoNav', module).add('Default', () => (
     onOrganizationChange={() => {}}
     onProjectChange={() => {}}
     admin={boolean('admin', true)}
+    hosts={{ cloud: 'http://localhost:8080' }}
```
hswolff: We'll need to remove before merging.

packages/mongo-nav/src/types.ts
```diff
@@ -100,11 +100,11 @@ export interface CurrentOrganizationInterface extends OrganizationInterface {
 }
 
 export interface DataInterface {
-  readonly account: AccountInterface;
-  currentOrganization: CurrentOrganizationInterface;
-  currentProject: CurrentProjectInterface;
-  readonly organizations: Array<OrganizationInterface>;
-  readonly projects: Array<ProjectInterface>;
+  readonly account?: AccountInterface;
```
hswolff: Do we really need to make all these properties optional? What's the impetus for this?

Really the data itself is optional, but we should be able to safely assume certain things of the structure of the data itself.

Namely that account will always exist, currentOrg and currentProject can definitely be undefined, but I believe organizations and projects will always at least be empty arrays.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -60,6 +68,16 @@ interface MongoNavInterface {
    * By default the value is set to `production`
    */
   mode?: Mode;
+
+  /**
+   * Function that is passed an error code, so that consuming application can handle fetch failures
+   */
+  onError?: (code: string) => void;
```
hswolff: This should follow our object-like enum pattern we have, where we accept only the values of all the keys of the object, but export the object itself so downstream users can use the object directly or just the string.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +121,53 @@ export default function MongoNav({
...
+      method: 'GET',
+    });
+
+  function getFixtureData() {
+    return new Promise(resolve => {
+      setTimeout(() => {
+        onSuccess?.(fixtureData);
+
+        resolve(fixtureData);
+      }, 1000);
```
hswolff: I'd put this to maybe 10, just so it's not un-necessarily slow for consumers but still forces people to expect the component to load asyncronously.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +121,53 @@ export default function MongoNav({
...
+        .then(data => {
+          setData(data);
+          onSuccess?.(data);
+        })
+        .catch(error => {
+          onError?.(error);
+          console.error(error);
+        });
+    }
+  }, []);
```
hswolff: Might want to have this re-run if `mode` or `endpointUri` changes.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +121,53 @@ export default function MongoNav({
...
+
+  if (!data || !data.account) {
```
```suggestion
  if (!data?.account) {
```

Or my preference:

```jsx
  if (data?.account == null) {
```

packages/mongo-nav/src/mongo-select/Trigger.tsx
hswolff: We should rename this prop because as is it's confusing what it means, ditto for project.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
hswolff: Only benefit to function is easier behavior for default case, but yeah either works for me. The return for OnPrem looks like a bug there too.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -89,140 +103,131 @@ const onKeyDown: React.KeyboardEventHandler = e => {
...
-    datum: OrganizationInterface,
-  ) => {
-    const content = (
-      <OrganizationOption
-        orgName={datum.orgName}
-        planType={datum.planType}
-        href={props.constructOrganizationURL(datum.orgId)}
-      />
-    );
+  const errorMessage =
```
hswolff: nit, this variable isn't used anywhere else. Better to inline IMO rather than having the reader read this variable here and then have to read where it's being used to understand the full behavior of the component.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -89,140 +103,131 @@ const onKeyDown: React.KeyboardEventHandler = e => {
...
+    >
+      <FocusableMenuItem>
+        <Input
+          onChange={onChange}
+          onKeyDown={onKeyDown}
+          variant="organization"
+        />
+      </FocusableMenuItem>
+
+      <>{data?.map(renderOrganizationOption) ?? <li>{errorMessage}</li>}</>
```
hswolff: Why does this need to be in a fragment? Does TS complain otherwise?

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -89,140 +103,131 @@ const onKeyDown: React.KeyboardEventHandler = e => {
...
-        </ul>
-      </li>
+      <>{data?.map(datum => renderProjectOption(datum))}</>
 
       <MenuSeparator />
 
-      {footer}
+      <li onKeyDown={onKeyDown} role="none" className={projectButtonStyle}>
+        <FocusableMenuItem>
+          <Button href={urls.mongoSelect.viewAllProjects} as="a">
```
hswolff: You're using href, you don't need to specify `as="a"`. Perhaps that default behavior is confusing?

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -129,18 +129,19 @@ export default function OrgNav({
...
           constructOrganizationURL={constructOrganizationURL}
-          variant="organization"
           urls={urls}
           onChange={onOrganizationChange}
           isActive={activeNav === 'orgSettings'}
         />
+
         <ul className={ulContainer}>
-          {variant && current.paymentStatus && (
+          {variant && current?.paymentStatus && (
```
hswolff: nit, we should rename variant to `paymentVariant` to disambiguate the multiple variant references we have in this file.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -258,7 +257,9 @@ export default function ProjectNav({
               href={projectNav.alerts as string}
               className={alertIconButtonStyle}
             >
-              {alerts && <div className={alertBadgeStyle}>{alerts}</div>}
+              {typeof alerts === 'number' && (
```
hswolff: Can have the default value of alerts be 0 and then check if alerts > 0 as we don't want to show an icon for 0 alerts.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx

packages/mongo-nav/src/mongo-select/Trigger.tsx
brookescarlett: Should we call the prop `placeholder` as @DesignerDave suggested?

packages/mongo-nav/src/MongoNav.story.tsx
brookescarlett: 👍

packages/mongo-nav/src/types.ts
brookescarlett: I think was just trying to defend against 'what ifs...' and handling cases where certain pieces of data are missing (even though they probably shouldn't be). I'll change account to be non-optional, same with orgs and projects

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/project-nav/ProjectNav.tsx
brookescarlett: 👍

packages/mongo-nav/src/org-nav/OrgNav.tsx

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
brookescarlett: Was supplying as "a" because typescript was typing urls.mongoSelect.newProject as possibly undefined, for now typecasting as string but need to figure out why Typescript is confused about that because should be <Required> by the time it gets to MongoSelect

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
brookescarlett: Yes TS saying undefined is not a type of ReactElement

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
brookescarlett: It's going to end up being more complicated logic down the line, because IIRC the error message will be different depending on what product you're in (CM, OM or Atlas) but can inline for now and can always pull out later if it becomes necessary

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/Loading.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { LogoMark } from '@leafygreen-ui/logo';
+import { css } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+
+const orgNavContainer = css`
+  height: 60px;
+  width: 100%;
```
DesignerDave: We shouldn't need this with `display: flex;`

packages/mongo-nav/src/Loading.tsx
```diff
@@ -0,0 +1,41 @@
...
+  line-height: 15px;
+  color: ${uiColors.gray.dark3};
+  border-bottom: 1px solid ${uiColors.gray.light2};
+`;
+
+const projNavContainer = css`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  width: 100%;
```
DesignerDave: Ditto here.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +126,60 @@ export default function MongoNav({
...
 
-  if (mode === Mode.Dev) {
-    data = devModeData;
+  const endpointURI = `${cloudHost}/user/shared`;
+
+  function getProductionData() {
+    return fetch(endpointURI, {
+      credentials: 'include',
+      method: 'GET',
+      mode: 'cors',
```
DesignerDave: I _think_ this is the default value for `mode`, so we should be able to get rid of this.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -89,140 +92,138 @@ const onKeyDown: React.KeyboardEventHandler = e => {
...
 
-      {footer}
+      <li onKeyDown={onKeyDown} role="none" className={projectButtonStyle}>
+        <FocusableMenuItem>
+          <Button href={urls.mongoSelect.viewAllProjects as string}>
+            View All Projects
+          </Button>
+        </FocusableMenuItem>
+        <FocusableMenuItem>
+          <Button href={urls.mongoSelect.newProject as string}>
```
DesignerDave: This typecasting is throwing red flags for me. We shouldn't need to do this, and if it's complaining, we're probably doing something that's not type-safe or typed correctly.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -160,10 +160,10 @@ export default function ProjectNav({
   data,
   constructProjectURL,
   urls,
-  alerts,
   activeProduct,
   onProjectChange,
   hosts,
+  alerts = 0,
```
DesignerDave: Why not allow this to be undefined?

brookescarlett: harry suggested this, which allows us to conditionally render alerts/the badge based on whether or not alerts is > 0

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
brookescarlett: <Required> only addresses the top-level, I was trying to be DRY but it seems like I need to make a new URLSInterface type that has every level <Required>

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +126,60 @@ export default function MongoNav({
...
+    });
+  }
+
+  function getFixtureData() {
+    return new Promise(resolve => {
```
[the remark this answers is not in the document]
brookescarlett: Leaving in promise, but removing the timeout

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +126,60 @@ export default function MongoNav({
...
+        onSuccess?.(fixtureData);
+
+        resolve(fixtureData);
+      }, 10);
+    });
+  }
+
+  async function handleResponse(response: Response) {
+    if (!response.ok) {
+      onError?.(ErrorCodeMap[response.status as 401]); //typecasting for now until we have more types to handle
```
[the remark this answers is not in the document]
brookescarlett: can do this: `const status = response.status as 401; //typecasting for now until we have more types to handle` but can't tell it that a response.status will always be 401 because TS knows it can be any number. Right now, we're only handling 401 which is why we have to cast it

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,16 +126,60 @@ export default function MongoNav({
...
+      onSuccess?.(data);
+    }
   }
 
-  if (!data) {
-    // Eventually this logic will be more robust, but for an alpha version will return null without data
-    return null;
+  useEffect(() => {
+    if (mode === Mode.Dev) {
+      getFixtureData().then(data => setData(data as DataInterface));
```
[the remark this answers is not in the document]
brookescarlett: TS doesn't allow you to assert the value of a promise like that. It says type DataInterface doesn't overlap with `unknown`

packages/mongo-nav/src/Loading.tsx
brookescarlett: 👍

packages/mongo-nav/src/Loading.tsx
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -111,27 +136,55 @@ export default function MongoNav({
...
+  onSuccess = () => {},
 }: MongoNavInterface) {
-  let data;
+  const [data, setData] = React.useState<DataInterface | undefined>(undefined);
 
-  if (mode === Mode.Dev) {
-    data = devModeData;
+  const hosts = defaultsDeep(hostsProp, defaultHosts);
+  const endpointURI = `${hosts.cloud}/user/shared`;
+
```
hswolff: Up to you but you could move all this fetching related behavior to a custom hook.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -111,27 +136,55 @@ export default function MongoNav({
...
-  } = data;
+  async function handleResponse(response: Response) {
+    if (!response.ok) {
+      const status = response.status as 401; //typecasting for now until we have more types to handle
+      onError?.(ErrorCodeMap[status]);
+      console.error(ErrorCodeMap[status]);
+    } else {
+      const data = await response.json();
+      setData(data);
+      onSuccess?.(data);
```
hswolff: Let's be consistent with the order in which we call the callback and the internal functions. Is it prop callback first, then internal functions, or vice versa? The order is swapped in getFixtureData.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -111,27 +136,55 @@ export default function MongoNav({
...
+  }
 
-  const hosts = defaultsDeep(hostsProp, defaultHosts);
+  useEffect(() => {
+    if (mode === Mode.Dev) {
+      getFixtureData().then(data => setData(data as DataInterface));
+    } else {
+      getProductionData()
+        .then(handleResponse)
+        .catch(console.error);
```
hswolff: I can believe a future where we gracefully fail and just show a top nav with the mongo nav, rather than nothing.

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: Right now displaying the `<Loading />` component, but waiting on this https://jira.mongodb.org/browse/EXPO-254 before implementing any further logic

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: Talked offline, leaving as-is for now

PR #252, opened 2020-02-06
PD-497-Testing: Add Test Coverage for MongoNav component

seanmhanson:
This adds unit testing using Jest for covering the MongoNav component. It currently handles the testing of the fetch request, but does not handle the handling of hosts and overrides. I'll dig into that a bit later.

This also adds Cloud's `Deferred` package we usually use when we need to test asynchronous code and want to be resolve or reject the resulting promise lazily. The functionality is built into Sinon, but I couldn't find an equivalent with Jest, but it's very lightweight.

If helpful, I'll also outline the test structure for other components, and am happy to do further work as you would like (I enjoy this stuff a lot).

### DIFF
--- a/packages/lib/src/Deferred.ts
+++ b/packages/lib/src/Deferred.ts
@@ -1,0 +1,43 @@
+export default class Deferred<T> {
+  private _promise: Promise<T>;
+
+  private _resolve: (val: T | PromiseLike<T>) => void;
+
+  private _reject: (reason?: any) => void;
+
+  constructor() {
+    this._resolve = () => {};
+    this._reject = () => {};
+
+    this._promise = new Promise((resolve, reject) => {
+      this._resolve = resolve;
+      this._reject = reject;
+    });
+  }
+
+  resolve(value: T) {
+    this._resolve(value);
+    return this;
+  }
+
+  reject(value: any) {
+    this._reject(value);
+    return this;
+  }
+
+  promise() {
+    return this._promise;
+  }
+
+  then: Promise<T>['then'] = (...args) => {
+    return this._promise.then(...args);
+  };
+
+  catch: Promise<T>['catch'] = (...args) => {
+    return this._promise.catch(...args);
+  };
+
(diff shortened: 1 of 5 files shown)


packages/lib/src/Deferred.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  promise() {
+    return this._promise;
+  }
+
+  then: Promise<T>['then'] = (...args) => {
+    return this._promise.then(...args);
+  };
+
+  catch: Promise<T>['catch'] = (...args) => {
```
DesignerDave: Hmm... This approach seems a bit heavy-handed, but I can't think of a much better way to go about it.

That said, since we aren't using `then` or `catch` anywhere, is there a reason you wanna keep them around?

Additionally, we might consider moving this out of `lib` somewhere. Right now, there isn't a good way to tree-shake dependencies for our packages, and consumers wouldn't be using this, we would currently be including this in every app using `leafygreen-ui` since pretty much all of our components include this package. I'm working on the tree-shaking issue sooner-than-later, but for the time being, it's still an issue.

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -1,81 +1,208 @@
...
+    describe('when the data is not returned successfully', () => {
+      beforeEach(function() {
+        fetchDeferred.reject(new Error('fetch calls rarely result in a rejection'));
+      });
+
+      it('logs the error to the console', () => {
+        expect(consoleErrorMock).toHaveBeenCalledTimes(1);
+      });
+
+      it('still displays the loading component', () => {
```
DesignerDave: It may be premature to include this. Our loading state may not end up being an entirely different component – rather smaller differences within the existing nav. The current implementation is a placeholder until the designs are finalized.

seanmhanson:
This wasn't meant to be reviewed and merged, so I'm closing this out but not deleting it, so Brooke can keep it as reference as needed.

PR #256, opened 2020-02-10
Cleaner construction of defaults in MongoNav

brookescarlett:
## ✍️ Proposed changes

- Using lodash `defaultsDeep` to construct default URLs in MongoNav

## 🛠 Types of changes

### DIFF
--- a/packages/mongo-nav/package.json
+++ b/packages/mongo-nav/package.json
@@ -22,6 +22,7 @@
     "@leafygreen-ui/badge": "^3.0.0",
     "@leafygreen-ui/button": "^4.0.0",
     "@leafygreen-ui/hooks": "^2.0.0",
-    "@leafygreen-ui/icon-button": "^4.0.0"
+    "@leafygreen-ui/icon-button": "^4.0.0",
+    "lodash": "^4.17.15"
   }
 }
--- a/packages/mongo-nav/src/MongoNav.tsx
+++ b/packages/mongo-nav/src/MongoNav.tsx
@@ -5,6 +5,15 @@ import { Product, URLSInterface, HostsInterface, NavItem, Mode } from './types';
 import devModeData from './data';
 import defaultsDeep from 'lodash/defaultsDeep';
 
+const defaultHosts: Required<HostsInterface> = {
+  account: 'https://account.mongodb.com',
+  cloud: 'https://cloud.mongodb.com',
+  university: 'https://university.mongodb.com',
+  support: 'https://support.mongodb.com',
+  charts: 'https://charts.mongodb.com',
+  stitch: 'https://stitch.mongodb.com',
+};
+
 interface MongoNavInterface {
   /**
    * Describes what product is currently active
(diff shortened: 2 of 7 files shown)


packages/mongo-nav/package.json
```diff
@@ -22,6 +22,7 @@
     "@leafygreen-ui/badge": "^3.0.0",
     "@leafygreen-ui/button": "^4.0.0",
     "@leafygreen-ui/hooks": "^2.0.0",
-    "@leafygreen-ui/icon-button": "^4.0.0"
+    "@leafygreen-ui/icon-button": "^4.0.0",
+    "lodash": "^4.17.15"
```
DesignerDave: Can we make sure that lodash is a dependency wherever we use it, and then add it to our webpack externals? I think most applications already install it, so there's not much of a need to bundle it.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -121,110 +122,70 @@ export default function MongoNav({
...
 
-  const defaultOrgURL = (orgId: string) =>
-    `${cloudHost}/v2#/org/${orgId}/projects`;
-  const constructOrganizationURL =
-    constructOrganizationURLProp ?? defaultOrgURL;
-
-  const defaultProjectURL = (projectId: string) =>
-    `${cloudHost}/v2#/${projectId}`;
-  const constructProjectURL = constructProjectURLProp ?? defaultProjectURL;
+  const sanitizedHosts = defaultsDeep(hosts, defaultHosts);
```
DesignerDave: Still could probably use a better name than `sanitizedX`...

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -121,110 +122,70 @@ export default function MongoNav({
...
-  const supportHost = hosts?.support ?? `https://support.mongodb.com`;
-
-  const sanitizedHosts: Required<HostsInterface> = {
-    account: accountHost,
-    cloud: cloudHost,
-    university: universityHost,
-    support: supportHost,
-    charts: hosts?.charts ?? `https://charts.mongodb.com`,
-    stitch: hosts?.stitch ?? `https://stitch.mongodb.com`,
+  const defaultHosts: Required<HostsInterface> = {
```
hswolff: Can move out of the component so we don't redefine it every time.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -121,110 +122,70 @@ export default function MongoNav({
...
 
-  const defaultOrgURL = (orgId: string) =>
-    `${cloudHost}/v2#/org/${orgId}/projects`;
-  const constructOrganizationURL =
-    constructOrganizationURLProp ?? defaultOrgURL;
-
-  const defaultProjectURL = (projectId: string) =>
-    `${cloudHost}/v2#/${projectId}`;
-  const constructProjectURL = constructProjectURLProp ?? defaultProjectURL;
+  const constructedHosts = defaultsDeep(hosts, defaultHosts);
```
hswolff: optional: Perhaps rename the internal value of the prop so it's `({ hosts = hostsProps })` and then just name this variable `hosts`.

Ditto for URLs.

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx

PR #264, opened 2020-02-14
PD-541: Disabled Org Nav state, tweaks

seanmhanson:
## ✍️ Proposed changes
This adds a disabled state for the org nav if the preferences navigation item is active (user preferences and Ops Manager account settings do not have a current org or project). It also makes a few tweaks, like centering some off-center links in the org nav, exporting nav heights so we can programmatically adjust our app layout's styles, and increase test coverage for those components I worked with. I pulled out some common types into a test helper file for lib, which will someday have more than just types and be worthy of that title.

🎟 _Jira ticket:_ [PD-541](https://jira.mongodb.org/browse/[PD-541])

## 🛠 Types of changes

## ✅ Checklist

### Screenshot
Screen Shot 2020-02-13 at 7 31 12 PM


packages/mongo-nav/src/mongo-select/Trigger.tsx
```diff
@@ -67,25 +72,44 @@ export function OrganizationTrigger({
...
+      <button
+        {...rest}
+        className={buttonContainer}
+        data-testid="org-trigger"
+        disabled={disabled}
       >
         <Icon
-          glyph={'Settings'}
-          fill={isActive ? uiColors.green.base : uiColors.gray.base}
+          className={cx({ [disabledStyle]: disabled })}
```
DesignerDave: Could we move this to a variable outside of the return statement?

As a neat trick, I believe this works:
```js
const disabledStyleClassName = cx({ [disabledStyle]: disabled })
return (
        <span className={cx(selectedStyle, disabledStyleClassName)}>
          {disabled ? 'All Organizations' : placeholder}
        </span>
)
```

packages/mongo-nav/src/mongo-select/Trigger.tsx
```diff
@@ -67,25 +72,44 @@ export function OrganizationTrigger({
...
+          size="small"
+          glyph="Building"
         />
-      </a>
+        <span className={cx(selectedStyle, { [disabledStyle]: disabled })}>
+          {disabled ? 'All Organizations' : placeholder}
```
DesignerDave: I usually would be against this for a more generalized component, but since this is so specific, we can hard-code the placeholder value here rather than accepting it as a prop.

packages/mongo-nav/src/mongo-select/Trigger.tsx
```diff
@@ -67,25 +72,44 @@ export function OrganizationTrigger({
...
+        </span>
+        <Icon
+          className={cx({ [disabledStyle]: disabled })}
+          size="small"
+          glyph="CaretDown"
+        />
+      </button>
+      {!disabled && (
```
DesignerDave: I don't think we're conditionally rendering this. We should apply a disabled state appearance to it instead.

packages/mongo-nav/src/mongo-select/Trigger.tsx
```diff
@@ -67,25 +72,44 @@ export function OrganizationTrigger({
...
+        <a
+          href={urls.mongoSelect.orgSettings}
+          className={cx(anchorStyle, border)}
+          aria-label="settings"
+          data-testid="org-trigger-settings"
+        >
+          <Icon
+            glyph={'Settings'}
```
DesignerDave: We can just have this as a string rather than evaluating it as a JS statement:
```js
glyph='Settings'
```

packages/mongo-nav/src/mongo-select/Trigger.tsx
```diff
@@ -106,6 +130,7 @@ export function ProjectTrigger({
   return (
     <button
       {...rest}
+      data-testid="project-trigger"
```
DesignerDave: "project-select-trigger"? Only reason I'm being nit-picky about these is because we will need to make a major version bump to update these IDs in the future.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -17,8 +17,10 @@ import {
 import { OrgSelect } from '../mongo-select/index';
 import UserMenu from '../user-menu/index';
 
+export const orgNavHeight = 60;
```
DesignerDave: Out of curiosity, why is this needed?

packages/mongo-nav/src/mongo-select/Trigger.tsx
DesignerDave: I realize I was looking at this PR with the wrong idea in mind. This is for user preference pages not associated with an org, correct? If so, ignore this comment.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -109,6 +123,7 @@ export default function OrgNav({
   hosts,
 }: OrgNav) {
   const { orgNav } = urls;
+  const disabled = activeNav === 'preferences';
```
DesignerDave: I'm wondering if the name for this `activeNav` value should be more general than 'preferences'. It's possible we may need this for somewhere other than preference pages.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -241,6 +265,7 @@ export default function OrgNav({
             className={cx(rightSideLinkStyle, linkText, {
               [activeLink]: activeNav === 'admin',
             })}
+            data-testid="admin-link"
```
DesignerDave: What are the chances this test ID might collide with another on the page?

packages/mongo-nav/src/data.ts
```diff
@@ -49,3 +49,60 @@ export default {
...
+  account: 'https://account.mongodb.com',
+  cloud: 'https://cloud.mongodb.com',
+  charts: 'https://charts.mongodb.com',
+  realm: 'https://stitch.mongodb.com',
+  support: 'https://support.mongodb.com',
+  university: 'https://university.mongodb.com',
+};
+
+export const constructOrganizationURL = (orgId: string) =>
+  `https://cloud-dev.mongodb.com/v2#/org/${orgId}/projects`;
```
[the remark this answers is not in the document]
seanmhanson: We can do that. I preserved the existing behavior. Mind if we handle this outside of this PR?

packages/mongo-nav/src/mongo-select/Trigger.tsx
seanmhanson: Oops, this is something I'm used to linters handling. No prob

packages/mongo-nav/src/org-nav/OrgNav.tsx
seanmhanson: We export the heights of each Nav so that we can account for the offset of the correct Nav(s) when in tandem with the banners and other components that push content down in Cloud.

It also means the component can change height without the risk of breaking the integration

packages/mongo-nav/src/org-nav/OrgNav.tsx
seanmhanson: I took this from Brooke, but if you provide a better name, I'll change it. I feel uncomfortable trying to rename this without a suggestion that you'd prefer.

packages/mongo-nav/src/org-nav/OrgNav.tsx
seanmhanson: On Cloud? None, because we don't use Jest, plus our Admin app is partially Java Static Pages.

On this component? Likely none as well, unless you see another potential usage in the design that I missed.

In general, if the testid needs to be read outside of the component, it's the user's responsibility for selecting which one (ex: if there are many usages of button).

packages/mongo-nav/src/data.ts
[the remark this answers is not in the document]
brookescarlett: Sorry I think that was a mistake on my part, agreed that we should default to production URLs but can handle outside the PR if you prefer

packages/mongo-nav/src/mongo-select/MongoSelect.spec.tsx
```diff
@@ -1,224 +1,257 @@
...
-        .firstChild as HTMLAnchorElement;
-      expect(anchor?.href).toBe('https://cloud-dev.mongodb.com/v2#/5d729a93');
+          it('calls the on click handler', () => {
+            expect(onClick).toHaveBeenCalledTimes(1);
+          });
+        });
+      });
     });
   });
 });
```
brookescarlett: 👀

brookescarlett: Grateful 🙏

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: @DesignerDave do you prefer `userPreferences`? For now, all of the activeNav values have been matched to possible `activeNav` states

brookescarlett: My thought is more around perhaps not needing to check for `disabled` and `current` separately and finding a way to combine that logic, but that might be for another PR once I understand all of the moving parts that go into determining whether or not to show a payment badge... 🤔

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
```diff
@@ -76,8 +26,8 @@ describe('packages/UserMenu', () => {
       activeProduct={'cloud'}
       onLogout={onLogout}
       onProductChange={onProductChange}
-      urls={urls}
-      hosts={hosts}
+      urls={urlFixtures}
```
brookescarlett: So much cleaner thank you for moving this

brookescarlett:
First of all, thank you so much for cleaning this up AND for adding all of the tests. Really really appreciate it.

My only hesitation, which may not even need to be addressed now, is combining logic for the conditional rendering on OrgNav. Do we need to be checking for current and disabled, or can the disabled state be used to represent multiple things (i.e. if there's not currentActiveOrg, if the page is on userPreferences, etc.)

packages/mongo-nav/src/org-nav/OrgNav.tsx
seanmhanson: @hswolff `atlasPreferences` seems quite misleading to me. The page that sets this active nav is the preferences (for Atlas/CM) or account (for OM) page. Those pages do not deal with anything Atlas/CM/deployment/cluster/etc related.

Were this only for atlas/cloud manager, I'd suggest `userPreferences`. Were this only for Ops Manager I'd suggest `userAccount`. I think a good compromise would be `userSettings` instead.

hswolff: Yeah I guess the full name would be `atlasUserPreferences`.  `userSettings` wfm.

packages/mongo-nav/src/data.ts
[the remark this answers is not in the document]
seanmhanson: Oh, that's just a drive by. Done.

packages/mongo-nav/src/org-nav/OrgNav.tsx
seanmhanson: I still wouldn't want `atlasUserPreferences`, because then if I'm an Ops Manager user updating my username, the page is (Ops Manager) Account Settings. `atlasUserPreferences` both isn't platform-agnostic and doesn't describe the page, due to our unfortunately bifurcation of preferences vs. account settings.

Is this just to separate things from the potential `realmUserSettings` and `chartsUserSettings`, or some global `userSettings`? If so, we can definitely change it to `cloudUserSettings`?

I also think this might be worth addressing in a pre-release review rather than during alpha, if there is still disagreement.

.changeset/tender-baboons-learn.md
```diff
@@ -0,0 +1,10 @@
+---
+'@leafygreen-ui/mongo-nav': minor
+'@leafygreen-ui/side-nav': minor
+'@leafygreen-ui/lib': patch
```
DesignerDave: Nit: This PR adds features to `lib`, so it should probably be a minor change.

PR #266, opened 2020-02-14
PD-542: Side Nav scrolling and margin/height/padding tweaks

seanmhanson:
## ✍️ Proposed changes

Per feedback from design and users, it was brought up that `side-nav` doesn't allow for vertical scrolling, and some specs regarding height and padding/margins were incorrect still (though off by very little). This updates those values according to the design screenshot attached to the Jira ticket.

🎟 _Jira ticket:_ [https://jira.mongodb.org/browse/PD-542](https://jira.mongodb.org/browse/[https://jira.mongodb.org/browse/PD-542])

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/side-nav/src/SideNav.tsx
+++ b/packages/side-nav/src/SideNav.tsx
@@ -5,6 +5,8 @@ import { ulStyleOverrides } from './styles';
 
 const navStyles = css`
   width: 200px;
+  overflow-y: scroll;
+  padding-bottom: 24px;
 `;
 
 interface SideNavProps {
--- a/packages/side-nav/src/SideNav.tsx
+++ b/packages/side-nav/src/SideNav.tsx
@@ -5,8 +5,6 @@ import { ulStyleOverrides } from './styles';
 
 const navStyles = css`
   width: 200px;
-  overflow-y: scroll;
-  padding-bottom: 24px;
 `;
 
 interface SideNavProps {
--- a/packages/side-nav/src/SideNavItem.tsx
+++ b/packages/side-nav/src/SideNavItem.tsx
@@ -21,7 +21,7 @@ const defaultStyle = css`
   position: relative;
   width: 100%;
   margin: unset;
-  padding: 8px ${SIDE_OFFSET}px 8px ${SIDE_OFFSET}px;
+  padding: 6px ${SIDE_OFFSET}px 6px ${SIDE_OFFSET}px;
   min-height: 0;
   display: flex;
   flex-direction: column;
--- a/.changeset/clean-plums-swim.md
+++ b/.changeset/clean-plums-swim.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/side-nav': patch
+---
+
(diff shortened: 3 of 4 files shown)


packages/side-nav/src/SideNavItem.tsx
```diff
@@ -21,7 +21,7 @@ const defaultStyle = css`
   position: relative;
   width: 100%;
   margin: unset;
-  padding: 8px ${SIDE_OFFSET}px 8px ${SIDE_OFFSET}px;
+  padding: 6px ${SIDE_OFFSET}px 6px ${SIDE_OFFSET}px;
```
DesignerDave: Hmm... can you point me to the mocks these values are from? I worked with Nelle on this, and to align with the typography / spacing scale we’re slowly moving to.

seanmhanson: designWithSpecs

This is the screenshot from Invision that passed through to me via Harry (and I think Nelle?)

packages/side-nav/src/SideNav.tsx
```diff
@@ -5,6 +5,8 @@ import { ulStyleOverrides } from './styles';
 
 const navStyles = css`
   width: 200px;
+  overflow-y: scroll;
```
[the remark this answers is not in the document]
seanmhanson: > Use `auto` instead

Good call, I missed this because macOS hides scrollbars when not in use.

>  I’m not convinced this should ever scroll independent of the page

I was careful to preserve existing behavior, ex: [Testing Cluster on QA](https://cloud-qa.mongodb.com/v2/5e18d3e0a7cc6835f8f7d364#clusters)

> Who suggested this change?

Cory Mintz (VP of Atlas Engineering) brought this up this morning (Feb 14). After he did, several other teams throughout Cloud reached out about the same, mostly people who use Atlas (which has a gigantic side nav).

**However**, I totally agree with you, and think it definitely reasonable to say that `<SideNav>` should be agnostic in terms of scrolling and bottom padding related to scrolling, should anyone mention it outside of Cloud. Users can either just use a container, or style the `<SideNav>` themselves if they use Emotion (I'm merging a PR right now to do that).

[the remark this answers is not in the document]
DesignerDave: An update – Harry, Nelle, and I are talking about the right solution for this. The main factors are whether this is simply a product of the transitional phase, or if this will be an issue in products after the transition. If it's the former, I'm likely to suggest adding these changes as overrides within MMS itself.

I appreciate moving the scrolling behavior to MMS, as that specific issue seems specific to there.

As an FYI, you should be able to pass a non-emotion className as a string to our components, and have that style applied as well. `cx()` should handle that case by doing old-school string concatenation between the string className and the emotion className.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: As noted in my previous comment, we're discussing how best to handle the length issue, and I'll get back to you about it.

seanmhanson:
Closing this, we can override on Cloud if we want, but should stick to the component of possible.

PR #268, opened 2020-02-18
PD-436: Create text input component

dominoweir:
## ✍️ Proposed changes

Create a new text input component.

🎟 _Jira ticket:_ [PD-436](https://jira.mongodb.org/browse/PD-436)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

There are several places in this code where I made an educated guess at what to do, figuring someone would correct me down the line. If you see something weird and think to yourself, "why did she do it that way?" then the answer is probably that I took a guess and need some constructive advice about the proper way to go about things.


packages/text-input/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "main": "./dist/index.node.js",
+    "types": "./dist/index.d.ts",
+    "scripts": {
+      "build": "../../node_modules/.bin/webpack --config ../../webpack.config.js"
+    },
+    "license": "Apache-2.0",
+    "publishConfig": {
+      "access": "public"
+    },
+    "devDependencies": {
```
brookescarlett: Check to make sure that you're using the correct/latest version of each component. Some of these components aren't at version 4.0.0 yet which is causing your build to fail with our CI tool

packages/text-input/src/TextInput.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+    />,
+  );
+
+  const renderedChildren = renderedTextInput.container.firstChild;
+
+  if (!typeIs.element(renderedChildren)) {
+    throw new Error('TextInput component failed to render');
+  }
+
+  const renderedInputElement =
```
brookescarlett: We're trying to move away form relying on the order of the DOM to find our components when testing. What might make that easier for you is to try grabbing getByText and getByTestId methods from the container when you render.

`const {getByText, getByTestId} = render(<TextInput />)`

This should allow you to look for your component more directly, without having to do as much DOM traversal.

packages/text-input/src/TextInput.story.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { boolean, text, select } from '@storybook/addon-knobs';
+import TextInput from '.';
+
+storiesOf('TextInput', module).add('Default', () => (
```
brookescarlett: Great use of knobs here!

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+  none: 'none',
+  valid: 'valid',
+  error: 'error',
+} as const;
+
+export type State = typeof State[keyof typeof State];
+
+interface TextInputProps {
+  /**
+   * Text shown in bold above the input element.
```
brookescarlett: nit: when referring to the component in documentation prefer to use capital letters, e.g. Input or <Input />

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+   */
+  label?: string;
+  /**
+   * Text that gives more detail about the requirements for the input.
+   */
+  description?: string;
+  /**
+   * Whether or not the field is optional. This is false by default.
```
brookescarlett: nit, next line
* Default: `false`

brookescarlett: (ditto elsewhere)

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+  className?: string;
+}
+
+/**
+ * # TextInput
+ *
+ * TextInput component
+ *
+ * ```
+<TextInput label={'Input Label'} onChange={() => execute when value of input field changes}/>
```
brookescarlett: Don't need brackets around string for label

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+      color: #cf4a22;
+      font-size: 14px;
+      height: 20px;
+      padding-top: 4px;
+      font-weight: normal;
+    `;
+
+    return (
+      <div className={className}>
+        <label className={textInputStyle + ' ' + labelStyle}>
```
brookescarlett: these classes can be composed by using `cx` from our instance of emotion.

`className={cx(textInputStyle, labelStyle)}`

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+    const errorMessageStyle = css`
+      color: #cf4a22;
+      font-size: 14px;
+      height: 20px;
+      padding-top: 4px;
+      font-weight: normal;
+    `;
+
+    return (
+      <div className={className}>
```
brookescarlett: Is there a reason that we're using a div as the container element? I think we should leave the container element as the label and apply the supplied className prop to the label element.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+        <label className={textInputStyle + ' ' + labelStyle}>
+          {label}
+          <label className={descriptionStyle}>{description}</label>
+          <div className={inputContainerStyle}>
+            <input
+              {...inputSelectorProp.prop}
+              className={cx(
+                inputStyle,
+                css`
```
brookescarlett: There's enough css in here that it's probably worth pulling out into it's own style and composing through cx

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+      }
+    `;
+
+    const interactionRingHoverStyle = css`
+      ${inputSelectorProp.selector}:hover ~ & {
+        transform: scale(1);
+      }
+    `;
+
+    const textInputStyle = css`
```
brookescarlett: Is there a reason that we're doing all of this styling inside the component? Should probably be able to do out of scope of the component, ditto elsewhere.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+              <Icon
+                glyph="Checkmark"
+                className={cx(
+                  validIconStyle,
+                  css`
+                    z-index: ${hasFocus ? 2 : 1};
+                  `,
+                )}
+              />
+            )}
```
brookescarlett: nit: let's add line breaks in between each state check to make more readable

packages/text-input/src/TextInput.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const renderedTextInput = render(
+    <TextInput
+      className={className}
+      label={label}
+      description={description}
+      onChange={onChange}
+      state={State.none}
+      optional={true}
+    />,
+  );
```
DesignerDave: I have a trick that'll make this a bit neater!

Instead of:
```js
  const className = 'test-text-input-class';
  const label = 'Test Input Label';
  const description = 'This is the description';
  const error = 'This is the error message';
  const validEmail = 'test.email@mongodb.com';
  const invalidEmail = 'invalid.email';
  const onChange = jest.fn();
  onChange.mockReturnValue('none');

  const renderedTextInput = render(
    <TextInput
      className={className}
      label={label}
      description={description}
      onChange={onChange}
      state={State.none}
      optional={true}
    />,
  );
```

try this:
```js
  const error = 'This is the error message';
  const validEmail = 'test.email@mongodb.com';
  const invalidEmail = 'invalid.email';
  const props = {
    className: 'test-text-input-class',
    label: 'Test Input Label',
    description: 'This is the description',
    onChange: jest.fn(),
    state: State.none,
    optional: true,
  };
 
  props.onChange.mockReturnValue('none');

  const renderedTextInput = render(<TextInput {...props} />);
```

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import Icon from '@leafygreen-ui/icon';
+import { uiColors } from '@leafygreen-ui/palette';
+import { useUsingKeyboardContext } from '@leafygreen-ui/leafygreen-provider';
+import { createDataProp } from '@leafygreen-ui/lib';
+
+export const State = {
+  none: 'none',
```
DesignerDave: Quick note: typically we PascalCase enum values (this being a pseudo-enum).

```js
export const State = {
  None: 'none',
  ...
};
```

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+      if (state === State.valid || state === State.error) {
+        return '30px';
+      } else {
+        return optional ? '60px' : '12px';
+      }
+    }
+
+    function onValueChange(e: React.ChangeEvent<HTMLInputElement>) {
+      if (onChange != undefined) {
+        state = onChange(e.target.value);
```
DesignerDave: I believe we typically pass the event, rather than the value, directly to the onChange method that the user passes to a component to mirror the default API. Let me know if you've seen otherwise though!

packages/text-input/src/TextInput.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
dominoweir: in this particular case, I'm referring to the HTML <input /> element, although I see why that would be unclear. I can still change it, if you want.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+        return state === State.valid ? '#13AA52' : '#CF4A22';
+      }
+    }
+
+    function getInputPaddingFromState() {
+      if (state === State.valid || state === State.error) {
+        return '30px';
+      } else {
+        return optional ? '60px' : '12px';
+      }
```
DesignerDave: So one coding pattern we try to follow religiously is the concept of "happy path coding."

If you're unfamiliar, the idea is that you don't need an `else` if your other conditions return. This would make the above function look like this:

```js
    function getInputPaddingFromState() {
      if (state === State.valid || state === State.error) {
        return '30px';
      }
      
      return optional ? '60px' : '12px';
    }
```

This makes it clear that the condition is an exception, and clearly shows the "happy path," or frequently the default one.

packages/text-input/src/TextInput.tsx
dominoweir: I just didn't know exactly where all this belonged, so I left it inside the component for the time being. I'll move all of the styling that isn't dependent on the component in some way to be out of scope.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+    const setValue = isControlled ? setControlledValue : setUncontrolledValue;
+    const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+    const [hasFocus, setHasFocus] = useState(false);
+    const inputSelectorProp = createDataProp('input-selector');
+
+    function getInputColorFromState() {
+      if (state === State.none) {
+        return '#CCC';
+      } else {
+        return state === State.valid ? '#13AA52' : '#CF4A22';
```
DesignerDave: We should look at pulling in values from our palette package wherever possible. 

```js
import {uiPalette} from '@leafygreen-ui/palette';
```

I believe `#13AA52` becomes `uiColors.green.base` in the case above, for example.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+    function getInputPaddingFromState() {
+      if (state === State.valid || state === State.error) {
+        return '30px';
+      } else {
+        return optional ? '60px' : '12px';
+      }
+    }
+
+    function onValueChange(e: React.ChangeEvent<HTMLInputElement>) {
+      if (onChange != undefined) {
```
DesignerDave: We should probably be doing either `onChange != null` or `onChange !== undefined` here. We try to avoid `!=` wherever possible except in the `!= null` case.

packages/text-input/src/TextInput.tsx
dominoweir: Good catch, I use cx in other places but totally missed this one

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+              value={value}
+              required={!optional}
+              disabled={disabled}
+              placeholder={placeholder}
+              onChange={e => onValueChange(e)}
+              ref={forwardRef}
+              onFocus={() => setHasFocus(true)}
+              onBlur={() => setHasFocus(false)}
+            />
+            {state === State.valid && (
```
DesignerDave: Nit: You shouldn't need the parentheses here.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,328 @@
...
+                )}
+              />
+            )}
+            {state === State.error && (
+              <Icon
+                glyph="Warning"
+                className={cx(
+                  errorIconStyle,
+                  css`
+                    z-index: ${hasFocus ? 2 : 1};
```
DesignerDave: Question: What's the reason this is needed?

packages/text-input/src/TextInput.tsx
dominoweir: Nope, no particular reason. I've reworked this so that the container element is the label.

packages/text-input/src/TextInput.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
dominoweir: Fixed, I put this in a variable called conditionalStyling

packages/text-input/src/TextInput.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
dominoweir: Good to know! I have gone through and replaced all the colors where possible.

packages/text-input/src/TextInput.tsx
dominoweir: This is a new concept for me, thanks for sharing! I went through and got rid of the extraneous `else` statements found throughout my code.

packages/text-input/src/TextInput.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
dominoweir: I got rid of these parentheses, but running `yarn format` added them back in.

packages/text-input/src/TextInput.spec.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.spec.tsx
dominoweir: Much prettier. Thanks!

packages/text-input/src/TextInput.tsx
dominoweir: Without this line, the interaction ring always renders on top of the input element. I don't know that I could explain why this particular CSS fixes the problem, it was just something I found in Brooke's wrapper code.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+   */
+  placeholder?: string;
+  /**
+   * The message shown below the input field if the value is invalid.
+   */
+  errorMessage?: string;
+  /**
+   * The current state of the TextInput. This can be none, valid, or error.
+   */
+  state?: State;
```
hswolff: Now that I see this in code it seems it may be a confusing word when put against React state. But I don't really have a better alternative.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+  /**
+   * The current value of the input field. If a value is passed to this prop, component will be controlled by consumer.
+   */
+  value?: string;
+  /**
+   * Callback used to set the value of the input field.
+   */
+  setValue?: Function;
```
hswolff: This shouldn't exist. If it's controlled then you just set the `value={value}` prop. There's no setter, we just all onChange when the value is changing.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+      if (onChange !== undefined) {
+        state = onChange(e);
+      }
+
+      if (setValue != undefined) {
+        setValue(e.target.value);
+      }
+    }
+
+    const interactionRingFocusStyle = css`
```
hswolff: Ideally we define things that don't rely on props or state outside the component so they're only defined once. It's a micro-optimization, but also a best practice.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+    const conditionalStyling = css`
+      z-index: ${hasFocus ? 2 : 1};
+      background: ${disabled ? uiColors.gray.light2 : '#FFFFFF'};
+      padding-right: ${getInputPaddingFromState()};
+      border: 1px solid ${getInputColorFromState()};
+    `;
+
+    return (
+      <label className={cx(textInputStyle, labelStyle, className)}>
```
hswolff: This should be a fieldset instead, right?

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+        {label}
```
hswolff: This should be inside the label element instead.

Did you know about a `label`'s `for` attribute? It's what lets a user click on a label and have it act on the input element. [Docs here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)

We'll need to create a unique id for each TextInput but we should definitely have a `for` attribute to lean into the default browser behavior.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+        <div className={inputContainerStyle}>
+          <input
+            {...inputSelectorProp.prop}
+            className={cx(inputStyle, conditionalStyling)}
+            type="text"
+            value={value}
+            required={!optional}
+            disabled={disabled}
+            placeholder={placeholder}
+            onChange={e => onValueChange(e)}
```
hswolff: FYI this is also valid. Rather than making another inline function you can just pass the function reference directly. 

```suggestion
            onChange={onValueChange}
```

packages/text-input/src/TextInput.tsx
DesignerDave: Happy to share! It definitely increases readability for conditional returns.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+    const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+    const [hasFocus, setHasFocus] = useState(false);
+    const inputSelectorProp = createDataProp('input-selector');
+
+    function getInputColorFromState() {
+      if (state === State.None) {
+        return '#CCC';
+      }
+
+      return state === State.Valid ? uiColors.green.base : uiColors.red.base;
```
DesignerDave: I might suggest taking the concept of happy-path further in this function in a couple ways.

- First is to break out the ternary statement you're returning into separate return statements.
- Second is to make the final return statement the default state, and check for the valid and error states in the conditionals.

Example of what this would look like:
```js
function getInputColorFromState() {
  if (state === State.Error) {
    return uiColors.red.base;
  }

  if (state === State.Valid) {
    return uiColors.green.base;
  }

  return '#CCC';
}
```

packages/text-input/src/TextInput.tsx
DesignerDave: I think @brookescarlett looked into this, and actually found the W3C recommendation was for this pattern. She can weigh in a bit more on that though.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+      color: ${uiColors.gray.dark1};
+      font-size: 14px;
+      height: ${description === '' ? '0' : '20'}px;
+      font-weight: normal;
+      padding-bottom: 8px;
+    `;
+
+    const conditionalStyling = css`
+      z-index: ${hasFocus ? 2 : 1};
+      background: ${disabled ? uiColors.gray.light2 : '#FFFFFF'};
```
DesignerDave: Nit: We should be specific about the property we're changing, rather than using the shorthand to change one value. In this case, `background-color` instead of `background`.

DesignerDave: Additionally, is there a reason these rules aren't a part of `inputStyle`?

dominoweir: changed to background-color. I don't have a good reason why those rules are defined separately, although they all depend on the component in some way. I will merge those two sets of rules and just have inputStyle defined inside the component.

packages/text-input/src/TextInput.tsx
dominoweir: I've seen the `for` attribute mentioned before, but I didn't touch it because I wasn't sure how to handle the unique ID piece. How would I use that attribute here? Also- isn't {label} already inside the element? I'm not sure what you mean.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,331 @@
...
+  optional?: boolean;
+  /**
+   * Whether or not the field is currently disabled.
+   * Default: false
+   */
+  disabled?: boolean;
+  /**
+   * Callback to be executed when the value of the input field changes.
+   */
+  onChange?: Function;
```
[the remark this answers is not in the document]
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
dominoweir: Can you explain further? I just kind of modeled my code after other hybrid controlled/uncontrolled components so maybe I'm missing something.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,337 @@
...
+
+      if (optional) {
+        return '60px';
+      }
+
+      return '12px';
+    }
+
+    function onValueChange(e: React.ChangeEvent<HTMLInputElement>) {
+      if (onChange !== undefined) {
```
brookescarlett: can we just check if(onChange)

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,337 @@
...
+        onChange(e);
+      }
+
+      if (!isControlled) {
+        setUncontrolledValue(e.target.value);
```
brookescarlett: Can just call this setValue now that we're not accepting this as a prop

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,337 @@
...
+      &:focus {
+        outline: none;
+      }
+    `;
+
+    return (
+      <label className={cx(textInputStyle, labelStyle, className)}>
+        {label}
+        <label className={descriptionStyle}>{description}</label>
+        <div className={inputContainerStyle}>
```
brookescarlett: do we need this div or can these styles go somewhere else?

packages/text-input/src/TextInput.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
dominoweir: Yup, fixed

packages/text-input/src/TextInput.tsx
dominoweir: yes, we need the div so that the icons can be vertically centered

packages/text-input/src/TextInput.tsx
dominoweir: Never mind, I figured it out! setValue prop has been removed.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,337 @@
...
+            </div>
+          )}
+
+          <div
+            className={cx(interactionRing, interactionRingHoverStyle, {
+              [interactionRingFocusStyle]: showFocus,
+            })}
+          />
+        </div>
+        {state === State.Error && (
```
hswolff: Should also check if `errorMessage` exists before we render this block?

dominoweir: yes, I think so- unless there is a case where you would want the error state/icon to show but not an error message? but that sounds weird so I'm going to guess it wouldn't happen.

packages/text-input/src/TextInput.tsx
```diff
@@ -171,9 +178,13 @@ const TextInput = React.forwardRef(
     const { usingKeyboard: showFocus } = useUsingKeyboardContext();
     const [hasFocus, setHasFocus] = useState(false);
     const inputSelectorProp = createDataProp('input-selector');
+    const [generatedId, setId] = useState(
+      id || `text-input-${Math.floor(Math.random() * 10000000)}`,
```
hswolff: The bug with what you have now is if someone were to put in `''` or `0` as the prop value, that's falsy, and we'd default to using our generated id.

```suggestion
      id ?? `text-input-${Math.floor(Math.random() * 10000000)}`,
```

hswolff: I think we can simplify this more:

```jsx
const [generatedId] = useState(`text-input-${Math.floor(Math.random() * 10000000)}`);
const id = propsId ?? generatedId;
```

packages/text-input/src/TextInput.tsx
```diff
@@ -278,7 +294,7 @@ const TextInput = React.forwardRef(
             />
           )}
 
-          {state === State.Error && (
+          {state === State.Error && errorMessage && (
```
hswolff: So this I think should not depend on the existence of `errorMessage`, only for showing the message.

Such that someone could show an error state but not show a message if they desire.

packages/text-input/src/TextInput.tsx
dominoweir: Would that simplification account for someone potentially changing the id, or do I still need useEffect to sync?

hswolff: Whenever the id prop changes the component re-runs. The generated id stays the same, but what value we set on the elements changes depending on the presence of the `id` prop. So they are effectively synced 'for free'.

packages/text-input/src/TextInput.spec.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
+import { render, cleanup, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import TextInput, { State } from './TextInput';
+import { typeIs } from '@leafygreen-ui/lib';
+
+afterAll(cleanup);
+
+describe('packages/text-input', () => {
```
hswolff: Additional things to assert for:

- [ ] When you don't define an `optional` prop the optional text isn't present
- [ ] When optional is true and state is valid or error we don't see optional
- [ ] If state is Error and errorMessage is not set, we still see an error state
- [ ] If state is not Error and errorMessage is set, we don't see an error state

dominoweir: thank for this- I felt like I was missing some important tests but I couldn't articulate what they were. I have added checks for all these things now!

packages/text-input/src/TextInput.tsx
DesignerDave: `validity` is the only thing I can think of, but I don't think it's better.

I also think since we've moved away from class components, state is a bit less confusing since we just use the hook now.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,354 @@
...
+    const interactionRingHoverStyle = css`
+      ${inputSelectorProp.selector}:hover ~ & {
+        transform: scale(1);
+      }
+    `;
+
+    const descriptionStyle = css`
+      color: ${uiColors.gray.dark1};
+      font-size: 14px;
+      height: ${description === '' ? '0' : '20'}px;
```
DesignerDave: Rather than doing this, let's not render the paragraph tag for the description at all when there isn't one.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,354 @@
...
+      }
+
+      if (state === State.Valid) {
+        return uiColors.green.base;
+      }
+
+      return '#CCC';
+    }
+
+    function getInputPaddingFromState() {
```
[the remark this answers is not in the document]
dominoweir: Great idea! I've made this change now.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,354 @@
...
+    function getInputColorFromState() {
+      if (state === State.Error) {
+        return uiColors.red.base;
+      }
+
+      if (state === State.Valid) {
+        return uiColors.green.base;
+      }
+
+      return '#CCC';
```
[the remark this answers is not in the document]
dominoweir: I pulled all the colors directly from the mockups, but I cannot for the life of me figure out where I found the `#CCC`  as I don't see it used anywhere. The borders in the mockups all seem to use `uiColors.gray.light1`, so I swapped that in.

packages/text-input/src/TextInput.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,354 @@
...
+
+    const inputStyle = css`
+      width: 400px;
+      height: 36px;
+      border-radius: 4px;
+      padding-left: 12px;
+      font-size: 14px;
+      font-weight: normal;
+      z-index: ${hasFocus ? 2 : 1};
+      background-color: ${disabled ? uiColors.gray.light2 : '#FFFFFF'};
```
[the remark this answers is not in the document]
dominoweir: good to know! fixed.

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,354 @@
...
+            className={inputStyle}
+            type="text"
+            value={value}
+            required={!optional}
+            disabled={disabled}
+            placeholder={placeholder}
+            onChange={onValueChange}
+            ref={forwardRef}
+            onFocus={() => setHasFocus(true)}
+            onBlur={() => setHasFocus(false)}
```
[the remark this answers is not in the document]
dominoweir: would love some help with this

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,354 @@
...
+      label,
+      description,
+      optional = false,
+      disabled = false,
+      onChange,
+      placeholder,
+      errorMessage,
+      state,
+      value: controlledValue,
+      className,
```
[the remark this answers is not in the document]
dominoweir: Ok neat- I knew y'all allow `...rest` but I couldn't figure out where it would be spread in this case. Added now!

.changeset/small-vans-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/text-input': major
+---
+
+Create a text input component
```
brookescarlett: Nit, for first release we usually write: "Initial release of TextInput"

packages/text-input/README.md
```diff
@@ -0,0 +1,58 @@
...
+import TextInput from '@leafygreen-ui/text-input';
+
+const [value, setValue] = useState('');
+
+return (
+  <TextInput
+    label='Email'
+    description='Enter your email below'
+    placeholder='your.email@example.com'
+    onChange={event => setValue(event.target.value)}
```
brookescarlett: I believe we fixed this to be called with just the event sent back, if that's the case update docs accordingly :)

packages/text-input/README.md
```diff
@@ -0,0 +1,54 @@
...
+  className='email-text-input'
+  onClick={(event) => {/* Something to handle the click event */}}
+/>
+
+```
+
+**Output HTML**
+
+```
+not really sure what should go here...
```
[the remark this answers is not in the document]
brookescarlett: Be sure to add this in before merging!

packages/text-input/README.md
```diff
@@ -0,0 +1,58 @@
...
+## Properties
+
+| Prop           | Type                     | Description                                                                  | Default |
+| -------------- | ------------------------ | ---------------------------------------------------------------------------- | ------- |
+| `id`           | `string`                 | id associated with the TextInput component.                                  |         |
+| `label`        | `string`                 | Text shown in bold above the input element.                                  |         |
+| `description`  | `string`                 | Text that gives more detail about the requirements for the input.            |         |
+| `optional`     | `boolean`                | Marks the input as optional                                                  | `false` |
+| `disabled`     | `boolean`                | Disabled the input                                                           | `false` |
+| `onChange`     | `function`               | The event handler function for the 'onchange' event.                         |         |
```
brookescarlett: Can provide information here about what is passed back to the function (i.e. receives the event object as its argument)

packages/text-input/src/TextInput.tsx
brookescarlett: got it, makes sense

packages/text-input/README.md
brookescarlett: Also usually last line of readme provides information about where we spread `{...rest}` props

brookescarlett:
Just a few more nits from me, but it's looking really really good

packages/text-input/README.md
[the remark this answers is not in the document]
dominoweir: Done

packages/text-input/README.md
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,345 @@
...
+  z-index: 0;
+`;
+
+const inputStyle = css`
+  width: 400px;
+  height: 36px;
+  border-radius: 4px;
+  padding-left: 12px;
+  font-size: 14px;
+  font-weight: normal;
```
DesignerDave: It looks like browsers explicitly set the font-family for text inputs, so we should do the same to override those styles. That's causing the text in the input to appear larger than the description, despite being the same font-family.

Could you set the following CSS rule here?
```css
font-family: Akzidenz, ‘Helvetica Neue’, Helvetica, Arial, sans-serif;
```

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,345 @@
...
+  background-color: ${uiColors.white};
+  border: 1px solid;
+  z-index: 1;
+
+  &:placeholder {
```
DesignerDave: This rule isn't doing anything right now. This is a bit of an oddity, but the placeholder pseudo-class needs to be preceded with two colons for it to work.

This should do the trick:
```less
&::placeholder {
  color: ${uiColors.gray.base};
}
```

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,345 @@
...
+    color: ${uiColors.gray.base};
+  }
+
+  &:focus {
```
DesignerDave: Right now, the gray input border is visible when the element is focused. The design specs have it match the color of the interaction ring on focus.

dominoweir: I need some help figuring out the CSS magic for how to make this happen

packages/text-input/src/TextInput.tsx
dominoweir: Fixed

packages/text-input/src/TextInput.tsx
dominoweir: Fixed (the text does look nicer now)

packages/text-input/src/TextInput.tsx
DesignerDave: It should be as simple as changing `border-color` to match the interaction ring. That, and making sure there's a border-color transition.

dominoweir: Figured it out- I was doing that, just in the wrong place 🤦‍♀

packages/text-input/src/TextInput.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+const inputContainerStyle = css`
+  position: relative;
+  display: flex;
+  align-items: center;
+  z-index: 0;
+`;
+
+const inputStyle = css`
+  width: 400px;
```
DesignerDave: Setting width on this element causes some issues when the label is longer:

Screen Shot 2020-02-25 at 11 35 01 AM

What may be the best solution is setting this to `width: 100%`, and set the fixed width in storybook to be applied on the outer wrapper. This helps the consumer use this to their design specifications more easily.

dominoweir: yup, that solution worked nicely. lingering question though- when the label/description are longer than one line, what do I need to do from a CSS perspective to make sure everything still renders nicely with no overlapping elements?

PR #269, opened 2020-02-18
PD-545: Add new types such that usermenu can be used without mongonav

brookescarlett:
## ✍️ Proposed changes
- University needs a Menu to use ASAP, and we've deprecated MongoMenu, so we need to make UserMenu resilient when used in environments outside of the MongoNav component
- Only blocker here is that the Hosts/URLs are currently required for the entire MongoNav component, so suggested changes.. 

- Hosts/URLs are no longer required props
- Duplicated logic for creating defaults for Hosts and URLs from MongoNav, as if this component is to be used outside of the MongoNav component, it will need this logic implemented
- To Note: I know there is a PR floating around out there made by @seanmhanson which moves some of this URL logic outside of the <MongoNav /> component, so I didn't want to duplicate this work right now and was thinking we could merge without the more generalized defaults and then update once their PR is in

🎟 _Jira ticket:_ [PD-545](https://jira.mongodb.org/browse/[PD-545])

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/mongo-nav/src/user-menu/UserMenu.tsx
+++ b/packages/mongo-nav/src/user-menu/UserMenu.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import defaultsDeep from 'lodash/defaultsDeep';
 import Badge from '@leafygreen-ui/badge';
 import Button from '@leafygreen-ui/button';
 import Icon from '@leafygreen-ui/icon';
@@ -20,9 +21,16 @@ import {
   URLSInterface,
   Product,
   HostsInterface,
-  UserMenuURLSInterface,
-  UserMenuHostsInterface,
 } from '../types';
+
+const defaultHosts: Required<HostsInterface> = {
+  account: 'https://account.mongodb.com',
+  cloud: 'https://cloud.mongodb.com',
+  university: 'https://university.mongodb.com',
+  support: 'https://support.mongodb.com',
+  charts: 'https://charts.mongodb.com',
+  realm: 'https://stitch.mongodb.com',
+};
 
 const subMenuContainer = createDataProp('sub-menu-container');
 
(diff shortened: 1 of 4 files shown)


packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -22,6 +23,15 @@ import {
...
 } from '../types';
 
+const defaultHosts: Required<HostsInterface> = {
+  account: 'https://account.mongodb.com',
+  cloud: 'https://cloud.mongodb.com',
+  university: 'https://university.mongodb.com',
+  support: 'https://support.mongodb.com',
+  charts: 'https://charts.mongodb.com',
+  realm: 'https://stitch.mongodb.com',
+};
```
DesignerDave: Should this have an `as const`?

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -22,6 +23,15 @@ import {
   HostsInterface,
 } from '../types';
 
+const defaultHosts: Required<HostsInterface> = {
```
DesignerDave: This doesn't need to be marked `Required` I don't believe.

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -154,19 +164,74 @@ interface UserMenuProps {
...
   onLogout = () => {},
   onProductChange = () => {},
-  urls,
-  hosts,
+  urls: urlsProp,
+  hosts: hostsProp,
 }: UserMenuProps) {
+  const hosts = defaultsDeep(hostsProp, defaultHosts);
+
+  const defaultURLs = {
```
DesignerDave: Might wanna drop a comment that this is temporary.

packages/mongo-nav/src/user-menu/UserMenu.tsx

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: this is just the defaults and we are supplying them, so we should be able to leave it as required

packages/mongo-nav/src/user-menu/UserMenu.tsx

packages/mongo-nav/src/user-menu/UserMenu.tsx
hswolff: Chatted with Brooke offline- this is temporary and another PR unifies these data structures that are used in multiple places. So this works for me for now, so long as we clean it up later.

DesignerDave: SGTM, resolving.

PR #273, opened 2020-02-20
Add more logic to determine whether or not Payment Status is displayed in MongoNav

brookescarlett:
## ✍️ Proposed changes

- Add logic to determine whether or not Payment Status is displayed in MongoNav

🎟 _Jira ticket:_ [PD-493](https://jira.mongodb.org/browse/PD-493)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -141,15 +149,15 @@ export default function OrgNav({
         />
 
         <ul className={ulContainer}>
-          {paymentVariant && current?.paymentStatus && (
+          {paymentVariant && showPaymentStatus && (
```
DesignerDave: Is there a reason why we don't add the `paymentVariant` check to `showPaymentStatus`?

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -141,15 +149,15 @@ export default function OrgNav({
...
             <li>
               <Badge
                 className={css`
                   margin-right: 25px;
                 `}
                 variant={paymentVariant}
               >
-                {current.paymentStatus.toUpperCase()}
+                {current?.paymentStatus?.toUpperCase()}
```
DesignerDave: Two things:

First, to make the transition to rollup a bit smoother, can we use the old syntax for now? Not ideal or permanent, but we won't be able to use optional chaining with it for a bit.

Second, what's the case where paymentStatus is truthy, but also isn't a string? The `?.toUpperCase` strikes me as unnecessary.

.changeset/tasty-spiders-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/mongo-nav': minor
```
DesignerDave: Hmm... this could probably be a patch, but I don't feel all that strongly about it.

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: Good call 👍

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: it's odd because we are checking for the presence of both current and current.paymentStatus with `showPaymentStatus` so not sure why I even need to check down here

brookescarlett: when I inline the check, that goes away... perhaps I should just do that

.changeset/tasty-spiders-march.md
brookescarlett: 👍

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -141,18 +141,24 @@ export default function OrgNav({
...
-                  margin-right: 25px;
-                `}
-                variant={paymentVariant}
-              >
-                {current.paymentStatus.toUpperCase()}
-              </Badge>
-            </li>
-          )}
+          {current &&
+            current.paymentStatus &&
```
hswolff: Could you simplify this to `current?.paymentStatus ...etc`?

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -141,18 +141,24 @@ export default function OrgNav({
...
+            paymentVariant &&
+            (admin ||
+              ['supsended', 'locked', 'admin suspended'].includes(
```
hswolff: typo, suspended

hswolff: Also it's not matching the values from mms.

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: Trying to move away from chaining for @DesignerDave's rollup PR

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: good catch on the typo... maybe i messed up but the enum i have for payment status is: 

const OrgPaymentLabel = {
  Embargoed: 'embargoed',
  EmbargoConfirmed: 'embargo confirmed',
  Ok: 'ok',
  Warning: 'warning',
  Suspended: 'suspended',
  Closing: 'closing',
  AdminSuspended: 'admin suspended',
  Dead: 'dead',
  Locked: 'locked',
  Closed: 'closed',
} as const;

hswolff: Yeah I'm not sure what's being passed in from the API tbh. I don't think we even are.

Regardless, here we should instead have it be `[OrgPaymentLabel.Suspended, ...etc]`. That's the value of using enums - reduces string errors!

brookescarlett: If I use the value of the enums, typescript says that "EMBARGOED" is not assignable to "SUSPENDED", "ADMIN_SUSPENDED" etc.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -136,6 +149,12 @@ export default function OrgNav({
     }
   }
 
+  const paymentValues: Array<OrgPaymentLabel> = [
+    'SUSPENDED',
```
hswolff: Values should still be enum values, `OrgPaymentLabel.Suspended` etc.

brookescarlett: Fixed!

PR #279, opened 2020-02-24
PD-425: Better documentation for Tooltip when not rendering children inside a trigger

dominoweir:
## ✍️ Proposed changes

Update documentation to clarify how to use a function as a trigger in the Tooltip component. 

🎟 _Jira ticket:_ [PD-425](https://jira.mongodb.org/browse/PD-425)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

I modified one of the tooltips in Storybook to use a function as `trigger` while I was working on understanding the issue. I can revert that change if y'all want, but I thought it might be nice to have that as an example. 

I don't think it would be possible to check for the presence of a `children` argument if someone passes a function as the trigger, but let me know if you have any ideas.


brookescarlett:
Looks great so far but still need to address this piece of it:
"What would be even better from a user experience point of view is if we could somehow detect that children prop is not being called in the render function of the tooltip – and log an error to the dev tools console. This would help greatly for users who like me skim read the documentation and made incorrect assumptions."

Would recommend checking to see whether or not trigger is a react element. If it's not, then log some sort of warning to console advising the consumer to explicitly return children in their function

dominoweir:
That makes sense. I added a warning that is logged when `React.isValidElement(trigger)` returns false, but I'm not sure exactly where that check belongs. For now I have placed it in the `createTriggerProps` function.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -168,6 +168,12 @@ function Tooltip({
     triggerEvent: TriggerEvent,
     triggerProps?: any,
   ) => {
+    if (!React.isValidElement(trigger)) {
```
brookescarlett: I feel like this doesn't get at the problem with meaningful specificity. That being said, it'd be really tough to. Spoke with @DesignerDave and we're cool with just updating the docs and being more explicit in storybook for now. We can go ahead and remove this warning

dominoweir: got it, I removed the warning 👍

PR #280, opened 2020-02-24
PD-500: MongoNav Ops Manager State

brookescarlett:
## ✍️ Proposed changes

- Add testing and documentation for two new props added to MongoNav: isOnPrem and onPremVersion
- Add styles for MongoNav Ops Manager State

🎟 _Jira ticket:_ [PD-500](https://jira.mongodb.org/browse/PD-500)

## 🛠 Types of changes

## ✅ Checklist

Screen Shot 2020-02-24 at 12 38 56 PM


packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -198,10 +219,12 @@ export default function OrgNav({
           onChange={onOrganizationChange}
           isActive={activeNav === 'orgSettings'}
           disabled={disabled}
+          isOnPrem={isOnPrem}
         />
         {!disabled && (
           <ul className={ulContainer}>
             {!isTablet &&
```
DesignerDave: This has gotten to a critical mass of conditionals where I think we may be better off moving the logic outside of the return statement. Something like this:

```ts
let badgeItem: React.ReactElement | null = null

if (
  !isTablet &&
  !isOnPrem &&
  current?.paymentStatus &&
  paymentVariant &&
  (admin || paymentValues.includes(current.paymentStatus))
) {
  badgeItem = (
    <li>
      <Badge
        className={css`margin-right: 25px;`}
        variant={paymentVariant}
        data-testid="org-payment-status"
      >
        {current.paymentStatus.split('_').join()}
      </Badge>
    </li>
  )
}

return (
  ...
  {badgeItem}
  ...
)
```

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -237,11 +260,16 @@ export default function OrgNav({
                   </OrgNavLink>
 
                   <Menu
-                    open={open}
-                    setOpen={setOpen}
+                    open={accessManagerOpen}
```
DesignerDave: I'm not sure I understand this change. Isn't this menu for both AM and Atlas?

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -100,6 +100,21 @@ interface MongoNavInterface {
...
+
+  /**
+   * Callback executed when user logs out
+   */
+  onLogout?: React.MouseEventHandler;
+
+  /**
+   * Version of Ops Manager that a user is using
+   */
+  onPremVersion?: string;
```
seanmhanson: (ditto reminder for an `multifactorAuthEnabled` prop)

packages/mongo-nav/src/org-nav/OrgNav.spec.tsx
```diff
@@ -53,6 +53,7 @@ describe('packages/mongo-nav/src/org-nav', () => {
     expectedElements.billing = queryByTestId('org-billing');
     expectedElements.allClusters = queryByTestId('all-clusters-link');
     expectedElements.admin = queryByTestId('admin-link');
+    expectedElements.version = queryByTestId('onPrem-version');
```
seanmhanson: [nitpick] this should be snake-cased as `on-prem-version`

packages/mongo-nav/src/org-nav/OrgNav.tsx
seanmhanson: Harry and I often disagree about when we should combine conditions, etc., and he usually comes down on the side of having more logic in the jsx is more readable, whereas I come down on the side of pulling out logic making things simpler for when you just need to grok code, etc.

That being said, I think nobody would disagree if we simplified any conditional rendering when there are three or more expressions.

seanmhanson: [aside] also, I wanted to share how we deal with responsive conditional rendering, etc. in cloud. We use `react-responsive`, which includes a hook `useMediaQuery` which handles detecting resizing, reading the window, etc., and takes in a min and max width as options.

We then apply that to create responsive wrapper components. For simplicity, I'm going to assign some arbitrary values and pull out types:

```
// breakpoint globals
const SMALL = 767;
const LARGE = 1025;
const MEDIUM_MIN = SMALL + 1;
const MEDIUM_MAX = LARGE - 1;

// internal helper
const getMediaQueryComponent = mediaQueryProps => {
  return ({ children }) => {
    const isMatchingSize = useMediaQuery(mediaQueryProps);
    return isMatchingSize ? children : none;
  };
};

// conditionally render content only for small viewports
const smallViewportProps = { maxWidth: SMALL };
export const SmallViewport = getMediaQueryComponent(smallViewportProps);

// conditionally render content for medium viewports ONLY
const mediumViewportProps = { minWidth: MEDIUM_MIN, maxWidth: MEDIUM_MAX };
export const MediumViewport = getMediaQueryComponent(mediumViewportProps);

// conditionally render content for medium viewports OR LARGER
const mediumPlusViewportProps = { minWidth: MEDIUM_MIN };
export const MediumPlusViewport = getMediaQueryComponent(mediumPlusViewportProps);
```

Which means we can. do things like
```
<SmallViewport>
  all my rendering for small screens can go here
</SmallViewport>
```
or, as needed, we can just use the hook mentioned at the top. 

**TL;DR I am totally not saying any of this should be changed**. I just wanted to expose the patterns we fell into for responsive design, in case they seem interesting or were different from what you found useful. We all don't do nearly as much responsive design as your average startup, etc, so I think it's good to share these things.

packages/mongo-nav/src/org-nav/OrgNav.tsx
seanmhanson: I think that this is just cleaning up variable names,  to avoid confusion about which indicate the on prem user dropdown is open vs. the access manager is open, etc.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -304,12 +340,40 @@ export default function OrgNav({
...
+                Two-factor Authentication
+              </MenuItem>
+              <MenuItem href={urls.onPrem.personalization}>
+                Personalization
+              </MenuItem>
+              <MenuItem href={urls.onPrem.invitations}>Invitations</MenuItem>
+              <MenuItem href={urls.onPrem.organizations}>
+                Organizations
+              </MenuItem>
+              <MenuItem href={urls.onPrem.featureRequest}>
```
seanmhanson: Right now, I think you have the default for this set to an empty string. Is there a corresponding page for this? If not, maybe we can conditionally hide this for now?

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -304,12 +340,40 @@ export default function OrgNav({
...
+                Feature Request
+              </MenuItem>
+              <MenuItem onClick={onLogout}>Sign Out</MenuItem>
```
seanmhanson: This is going to be a major one where we'll want a `data-testid` to be assigned, because we click this many times in the E2E tests. The same should already exist (or be added) to the UserMenu.

I would actually go as far as to say it would be nice to have testids for each of the links, plus the trigger, for both menus. That's a fair number of similar testids, so to quote PEP20 (The Zen of Python):
> "Namespaces are one honking great idea—let's do more of those!"

(aka I hadn't thought this through, but it'd be nice to use prefixes like `om-user-menu` and `user-menu` to keep things tidy)

...also, I am so glad I didn't get PEP20 tattooed on my arm when I was going through my Python phase 😬

packages/mongo-nav/src/org-nav/OrgNav.spec.tsx
```diff
@@ -86,6 +87,18 @@ describe('packages/mongo-nav/src/org-nav', () => {
     );
   };
 
+  const testForVersion = (isVisible = true) => {
```
seanmhanson: To be honest, I think this is fine to just test:
- once, given default props, that this is invisible
- once, given onPrem, that this is visible

I sometimes go overboard in these things, so my apologies if this is different from `testForPaymentStatus` and such, totally my fault.

packages/mongo-nav/src/org-nav/OrgNav.spec.tsx
```diff
@@ -146,12 +165,30 @@ describe('packages/mongo-nav/src/org-nav', () => {
...
+
     testForPaymentStatus(false);
+    testForVersion(false);
 
     Object.keys(linkNamesToUrls).forEach(linkName =>
       testForNavLink(linkName, ['allClusters', 'admin'].includes(linkName)),
     );
   });
+
+  describe('when rendered onPrem', () => {
```
seanmhanson: We now additionally need to test:
- the expected user menu is rendered by default
- the on prem user menu is rendered when on prem is true

And I think that means adding testing for the contents of the on prem menu if there are any conditional things rendered / if urls are overwritten, etc.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -90,6 +89,21 @@ const accessManagerMenuItem = css`
   margin-bottom: 0px;
 `;
 
+const onPremMenuWrapper = css`
+  display: inline-block;
+  position: relative;
+  z-index: 0;
```
seanmhanson: I think I mentioned this generally, but we want. to be really careful about the selection of our z-index values. This will be a lower z-index than the fixed header itself by default. I am not 100% sure how that will turn out as a result.

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: there are two menus on the page that we're controlling instead of just one

brookescarlett: so had to be more specific w naming

packages/mongo-nav/src/org-nav/OrgNav.tsx

packages/mongo-nav/src/org-nav/OrgNav.spec.tsx
seanmhanson: Also having just written that last comment for the review, it might be nice to prefix the testids with the component name, so they're namespaced. I think @DesignerDave was trying to surreptitiously get me to do that in a previous PR.

brookescarlett: 👍

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: I asked in our channel if people knew where this should point to. I do not. If you and @hswolff don't know yet, then I can definitely hide for now.

packages/mongo-nav/src/org-nav/OrgNav.tsx

packages/mongo-nav/src/org-nav/OrgNav.spec.tsx
brookescarlett: 👍

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: Going to do a big zindex sweep per your previous ticket, dw!

packages/mongo-nav/src/org-nav/OrgNav.spec.tsx
brookescarlett: 👍

packages/mongo-nav/src/org-nav/OrgNav.tsx
DesignerDave: @seanmhanson `z-index: 0` is only different than `z-index: auto` (the browser default) in that it forms a new stacking context. Largely, this does almost nothing aside from allowing us to safely use z-index on the element's descendants. We do this frequently for things like our interaction ring pattern where we rely on z-index to render it properly, but don't want elements to appear above things like modals.

packages/mongo-nav/README.md
```diff
@@ -32,6 +32,8 @@
...
 | `onSuccess`                | `(response) => {}`                                                             | Callback that receives the response of the fetched data, having been converted from JSON into an object                                                   | `() => {}`                                              |
+| `isOnPrem`                 | `boolean`                                                                      | Whether or not a user is using Ops Manager                                                                                                                | `false`                                                 |
+| `onPremVersion`            | `string`                                                                       | Version of Ops Manager that a user is using                                                                                                               |                                                         |
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/org-nav/OrgNav.tsx
DesignerDave: Codepen to demonstrate: https://codepen.io/mdbdave/pen/PoqWRGN

If you un-comment the `z-index` rule on `.context-parent`, it should demonstrate the effect `z-index: 0;` has on its children.

seanmhanson: I didn't know this! In that case 0 here sounds fine.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -20,17 +20,16 @@ import {
...
+import UserMenu, { UserMenuTrigger } from '../user-menu/index';
 
 export const orgNavHeight = 60;
 
 const navContainer = css`
   height: ${orgNavHeight}px;
   width: 100%;
   display: flex;
-  justify-content: space-between;
+  // justify-content: space-between;
```
hswolff: remove?

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -130,6 +131,10 @@ interface OrgNav {
   admin: boolean;
   hosts: Required<HostsInterface>;
   currentProjectName?: string;
+  isOnPrem?: boolean;
```
hswolff: I wonder if we should encapsulate all onPrem config inside an object.

```jsx
onPrem={{
  enabled: true,
  version: string,
   mfaEnabled: boolean
}}
```

WDYT @brookescarlett @seanmhanson ?

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -169,81 +179,154 @@ export default function OrgNav({
...
+        variant={paymentVariant}
+        data-testid="org-nav-payment-status"
+      >
+        {current.paymentStatus.split('_').join()}
+      </Badge>
+    );
+  }
+
+  const renderedUserMenu = isOnPrem ? (
+    <div className={onPremMenuWrapper}>
```
hswolff: It'd be nice to have these as their own components but then we'd have to pass down all props again.

The context trick we talked about would be nice, but again, not a big deal at this point.

packages/mongo-nav/src/styles.tsx
```diff
@@ -0,0 +1,16 @@
+import { css } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+
+export const removePointerEvents = css`
+  pointer-events: none;
```
hswolff: nice utility class!

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -232,6 +260,14 @@ export default function MongoNav({
       alerts: `${hosts.cloud}/v2/${data?.currentProject?.projectId}#alerts`,
       activityFeed: `${hosts.cloud}/v2/${data?.currentProject?.projectId}#activity`,
     },
+    onPrem: {
+      profile: `${hosts.cloud}/v2#/account/profile`,
+      mfa: `${hosts.cloud}/v2#/preferences/2fa`,
+      personalization: `${hosts.cloud}/v2#/account/personalization`,
+      invitations: `${hosts.cloud}/v2#/preferences/invitations`,
+      organizations: `${hosts.cloud}/v2#/preferences/organizations`,
+      featureRequest: ``,
```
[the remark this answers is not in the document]

packages/mongo-nav/src/data.ts
```diff
@@ -90,6 +90,14 @@ export const urlDefaults: Required<URLSInterface> = {
     alerts: `https://cloud.mongodb.com/v2/currentProjectId098#alerts`,
     activityFeed: `https://cloud.mongodb.com/v2/currentProjectId098#activity`,
   },
+  onPrem: {
+    profile: `https://cloud.mongodb.com/v2#/account/profile`,
+    mfa: `https://cloud.mongodb.com/v2#/preferences/2fa`,
+    personalization: `https://cloud.mongodb.com/v2#/account/personalization`,
+    invitations: `https://cloud.mongodb.com/v2#/preferences/invitations`,
+    organizations: `https://cloud.mongodb.com/v2#/preferences/organizations`,
+    featureRequest: ``,
```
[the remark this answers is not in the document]
brookescarlett: These urls are for testing purposes... they have the fixture data orgIds and projectIds interpolated.. but I can merge the host objects where appropriate.

packages/menu/src/Menu.tsx
```diff
@@ -54,7 +53,7 @@ interface MenuProps
    */
   shouldClose?: () => boolean;
 
-  children: Array<React.ReactElement>;
+  // children: React.ReactNode;
```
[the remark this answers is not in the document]
brookescarlett: children is now coming from PopoverProps instead of being re-typed here.

packages/menu/src/Menu.tsx
```diff
@@ -225,13 +224,16 @@ function Menu({
...
-    currentSubMenu,
-    open,
-    refs,
-  ]);
+  const updatedChildren = React.useMemo(() => {
+    if (
+      children == null ||
+      ['boolean', 'number', 'string'].includes(typeof children)
+    ) {
+      return;
```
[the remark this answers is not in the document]
brookescarlett: We're just saying you don't have to update the children if the typeof children isn't a ReactElement. I was doing some weird hacky thing with TS earlier to get it to work very specifically in context, but it was causing problems when trying to use Menu as it should work in other places. This removes those TS errors. For example, I was trying to conditionally render a MenuItem but was getting an error from TS because previously we said the children of Menu had to be an array of React Elements and undefined/null is not included in that

packages/mongo-nav/src/org-nav/OrgNav.tsx

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: I agree with you. Dave and I disagreed about this... @DesignerDave are you okay with me pulling this into a helper component now?

packages/mongo-nav/src/styles.tsx
brookescarlett: 👍

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: I don't hate that... eliminates some props and there are _so_ many on this component

brookescarlett: Thoughts @seanmhanson?

brookescarlett: Updated! Let me know what you think!

packages/mongo-nav/src/org-nav/OrgNav.tsx
DesignerDave: As we chatted about on Slack, I don't feel strongly about this specifically so go for it. I'm just cognizant that the way we've broken this component up makes it somewhat difficult to follow right now.

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -235,7 +227,7 @@ function UserMenu({
 
   const [open, setOpen] = useState(false);
 
-  const name = `${firstName} ${lastName}`;
+  const name = account ? `${account?.firstName} ${account?.lastName}` : '';
```
DesignerDave: Hmm... The way this is written, it looks like the output could be something like "Dave undefined" if one is defined and the other isn't.

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -298,22 +298,36 @@ function UserMenu({
...
-            {openInvitations ? (
+
+          <MenuItem
+            href={userMenu?.cloud?.invitations}
+            data-testid="user-menuitem-cloud-invitations"
+          >
+            {account?.openInvitations ? (
               <span className={subMenuItemStyle}>
-                Invitations <Badge variant="blue">{openInvitations}</Badge>
+                Invitations{' '}
```
DesignerDave: Not sure I understand this. If the idea is to add whitespace between "Invitations" and the badge, we should use margin instead of a space.

Additionally, is there a reason we can't just conditionally render the badge? We're repeating rendering the "Invitations" text right now.

packages/mongo-nav/src/org-nav/OrgNav.tsx

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: we add a container to control spacing when we have invitations and don't need to worry about styles/spacing when we're not rendering the badge.

DesignerDave: Is there a difference beyond having an additional element in the DOM? Otherwise, something like this may be best:

```js
<span className={subMenuItemStyle}>
  Invitations

  {account?.openInvitations &&
    <Badge variant="blue">{account.openInvitations}</Badge>
  }
</span>
```

brookescarlett: we don't need to add justify-content space-between when it's just "invitations"

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -162,20 +141,19 @@ export default function MongoNav({
...
   constructOrganizationURL: constructOrganizationURLProp,
   constructProjectURL: constructProjectURLProp,
   onError = () => {},
   onSuccess = () => {},
   onLogout = () => {},
-  onPremVersion,
-  onPremMFA = false,
+  onPrem = { mfa: false, enabled: false, version: '' },
 }: MongoNavInterface) {
+  const { mfa, enabled, version } = onPrem;
```
hswolff: I may argue to keep these props on the `onPrem` object to keep the context of their values nearby. I.e. on line 270 it's a little confusing what the enabled variable relates to.

Also should probably change the child props as well to keep them consistent.

packages/mongo-nav/src/helpers/OnPremUserMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+  open,
+  setOpen,
+  urls,
+  onLogout,
+  mfa,
+}: OnPremUserMenuProps) {
+  return (
+    <div className={onPremMenuWrapper}>
+      <UserMenuTrigger
+        name={name ?? ''}
```
hswolff: Can just be default value in props?

packages/mongo-nav/src/user-menu/UserMenu.tsx
DesignerDave: It also shouldn't matter if it's there, and the implementation is cleaner.

packages/mongo-nav/src/helpers/OnPremUserMenu.tsx
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/user-menu/UserMenu.tsx

packages/mongo-nav/src/user-menu/UserMenu.tsx

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -255,62 +284,69 @@ export default function OrgNav({
...
-
-        <UserMenu
-          account={account}
-          activeProduct={activeProduct}
-          urls={urls}
-          hosts={hosts}
-        />
       </div>
+
+      {renderedUserMenu}
```
seanmhanson: [test coverage] I'd suggest testing the following cases, particularly because we're very careful about avoiding UI bugs in Ops Manager (different release schedules, supporting multiple versions, etc., makes it tedious to get a bugfix out, and then the release may still not be installed by a group)

### When On Prem is enabled
- I see an on prem user menu
- I do not see the default user menu
- I see the on prem version
- I do not see a billing link

#### When admin is true as well
- I still do not see an admin link

### Default state
- I do not see the on prem user menu
- I do not see the on prem version

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -13,18 +12,9 @@ import {
   DataInterface,
   ErrorCode,
 } from './types';
-import { dataFixtures } from './data';
+import { dataFixtures, hostDefaults } from './data';
```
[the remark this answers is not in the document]

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -100,6 +90,16 @@ interface MongoNavInterface {
...
+
+  /**
+   * Callback executed when user logs out
+   */
+  onLogout?: React.MouseEventHandler;
+
+  /**
+   * onPrem config object with three keys: enabled, version and mfa
+   */
+  onPrem?: { mfa?: boolean; version?: string; enabled?: boolean };
```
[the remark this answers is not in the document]
brookescarlett: Not at the moment, but probably smart to pull out anyway

packages/mongo-nav/README.md
```diff
@@ -32,6 +32,7 @@
...
 | `onError`                  | `(ErrorCode) => {}`,                                                           | Function that is passed an error code as a string, so that consuming application can handle fetch failures                                                | `() => {}`                                              |
 | `onSuccess`                | `(response) => {}`                                                             | Callback that receives the response of the fetched data, having been converted from JSON into an object                                                   | `() => {}`                                              |
+| `onPrem`                   | `{enabled: boolean, mfa: boolean, version: string}`                            | onPrem config object with three keys: enabled, version and mfa                                                                                            | `{enabled: false, mfa: false, version: ''}`             |
```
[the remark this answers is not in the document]
brookescarlett: I broke out onPrem.enabled, etc. each to its own line. Let me know if you think that's cleaner!

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -245,37 +252,33 @@ export default function MongoNav({
...
         hosts={hosts}
-        currentProjectName={currentProject?.projectName}
+        currentProjectName={data?.currentProject?.projectName}
+        onLogout={onLogout}
+        onPremEnabled={onPrem.enabled}
+        onPremVersion={onPrem.version}
+        onPremMFA={onPrem.mfa}
       />
-      {showProjNav && currentProject && (
+      {showProjNav && !onPrem.enabled && (
```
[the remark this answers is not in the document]
brookescarlett: You should never see the ProjetNav when onPrem is enabled. Need to add a ProjectNav.spec file but I think that might be outside of the scope of this PR

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: - When onPrem is enabled: 
I already tested for lack of billing link and that the proper UserMenu is displayed
- Default
I already test for proper User Menu displayed

Will add tests for: 
- onPrem: version is displayed, admin not displayed even when true
- Default: do not see version

brookescarlett: Okay added!

packages/mongo-nav/src/helpers/OnPremUserMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onLogout?: React.MouseEventHandler;
+  mfa: boolean;
+}
+
+export default function OnPremUserMenu({
+  name,
+  open,
+  setOpen,
+  urls,
+  onLogout,
```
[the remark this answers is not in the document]
brookescarlett: I did stick it up higher in the tree so just making non-optional here.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -145,10 +147,12 @@ export default function MongoNav({
   constructProjectURL: constructProjectURLProp,
   onError = () => {},
   onSuccess = () => {},
+  onLogout = () => {},
+  onPrem = { mfa: false, enabled: false, version: '' },
```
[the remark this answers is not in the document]
brookescarlett: Leaving for now so that all the defaults are in one place, but can adjust later if you think necessary

packages/mongo-nav/src/styles.tsx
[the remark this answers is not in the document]
brookescarlett: Great catch, fixed this!

packages/mongo-nav/src/MongoNav.tsx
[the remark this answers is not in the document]
seanmhanson: No need.

packages/mongo-nav/src/MongoNav.tsx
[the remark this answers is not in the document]
seanmhanson: Oh, I mean add to `MongoNav.spec.tsx` a test that the Mongo Nav, if provided `onPrem.enabled = true`, will check this. No need. for a ProjectNav spec yet.

PR #290, opened 2020-02-28
PD-482: Code component styling updates

dominoweir:
## ✍️ Proposed changes

Styling updates to the Code component, including italicizing comments, adding a shadow when content overflows, and creating a "copy to clipboard" button. This also includes a smaller ticket that updates the abbreviation used in the Syntax component for C#. 

🎟 _Jira ticket:_ [PD-482](https://jira.mongodb.org/browse/PD-482)
🎟 _Additional ticket:_ [PD-551](https://jira.mongodb.org/browse/PD-551)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

I am having issues with testing the new copy button and would love some input on how to go about it (see comments on Code.spec.tsx for more).


packages/code/src/Code.spec.tsx
```diff
@@ -1,32 +1,74 @@
...
+
     test('root element renders as a <pre> tag', () => {
       expect(codeRoot.tagName).toBe('PRE');
     });
 
     test(`renders "${className}" in the root element's classList`, () => {
       expect(codeRoot.classList.contains(className)).toBe(true);
     });
+
+    test(`copy button actually copies text to clipboard`, () => {
```
dominoweir: This is my test that's causing issues. The comments I left in the test explain the current state of things. Currently, the problem I am having is that I cannot access the contents of the clipboard to paste in to an arbitrary input element.

packages/code/src/Code.tsx
```diff
@@ -140,8 +224,34 @@ function Code({
   chromeTitle = '',
   ...rest
 }: CodeProps) {
+  const [copied, setCopied] = useState(false);
+  const [canScrollLeft, setCanScrollLeft] = useState(false);
+  const [canScrollRight, setCanScrollRight] = useState(false);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const scrollableMultiLine = useRef<HTMLPreElement>(null);
+  const scrollableSingleLine = useRef<HTMLDivElement>(null);
+  useEffect(() => {
```
dominoweir: The reason I have a useEffect here is to force a rerender after the initial render. We cannot determine if the Code content overflows its container until after the initial render, so it is impossible to know until then if a shadow should be shown on the right side of the container. Brooke mentioned that this was kind of a hack-y solution, but for now it gets the job done. I would love to know if y'all have better ideas for how to accomplish showing the shadow initially.

packages/code/src/Code.tsx
```diff
@@ -169,16 +280,55 @@ function Code({
...
+
   if (!multiline) {
     return (
       <div className={wrapperStyle}>
         {renderedWindowChrome}
 
         <div
-          {...(rest as DetailedElementProps<HTMLDivElement>)}
-          className={wrapperClassName}
+          className={css`
```
dominoweir: Not sure if this needs to be in a separate variable since it's so short (and because I couldn't think of a good variable name), let me know what y'all think

typings/copy.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'react-copy-to-clipboard';
```
dominoweir: I wasn't sure where this declaration file belonged, so for now it's living in the same directory as the highlight module declarations.

packages/code/src/Code.tsx
```diff
@@ -140,8 +224,34 @@ function Code({
...
+  const [copied, setCopied] = useState(false);
+  const [canScrollLeft, setCanScrollLeft] = useState(false);
+  const [canScrollRight, setCanScrollRight] = useState(false);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const scrollableMultiLine = useRef<HTMLPreElement>(null);
+  const scrollableSingleLine = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    if (multiline) {
+      if (!isLoaded && scrollableMultiLine.current !== null) {
+        setIsLoaded(true);
```
hswolff: When you see yourself setting multiple states at once - that due to one condition two state variables need to change - it's generally better to reach for `useReducer`. [Shameless self promotion, but also more rich explanation of this pattern as a blog post I wrote.](https://hswolff.com/blog/why-i-love-usereducer/)

packages/code/src/Code.tsx
hswolff: The reason makes sense. You need to read the DOM before you can make decisions. There's a different hook just for this need, [useLayoutEffect](https://reactjs.org/docs/hooks-reference.html#uselayouteffect).

packages/code/src/Code.tsx
brookescarlett: My understanding of useLayoutEffect may be wrong, but could we remove this all together once useLayoutEffect replaces useEffect, because we'll have information about the layout changing rather than having to force the rerender?

packages/code/src/Code.tsx
```diff
@@ -169,16 +280,55 @@ function Code({
...
 
+  function handleScroll(e: React.UIEvent) {
+    const scrollWidth = e.currentTarget.scrollWidth;
+    const elementWidth = e.currentTarget.clientWidth;
+    const isScrollable = scrollWidth > elementWidth;
+
+    if (isScrollable) {
+      const scrollPosition = e.currentTarget.scrollLeft;
+      const maxPosition = scrollWidth - elementWidth;
+      setCanScrollLeft(scrollPosition > 0);
```
brookescarlett: Would it be easier/more readable to combine and do something like [scroll, setScroll] = useState('none') and have it be an enum of values: `{'left', 'right', 'both', 'none'}`. You don't have to run with this, just a suggestion that may help clean things up a bit

packages/code/src/Code.tsx
brookescarlett: Should be fine to leave as-is when it's just one rule 👍

dominoweir:
I have refactored this code to take advantage of `useLayoutEffect` and `useReducer`. let me know what y'all think!

packages/code/src/Code.tsx
dominoweir: This ended up becoming much cleaner once I started using an enum 👍

packages/code/src/Code.tsx
```diff
@@ -80,6 +92,121 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+  return '';
+}
+
+const initialState = {
+  scroll: ScrollState.None,
+  copied: false,
+  isLoaded: false,
+};
+
+function scrollReducer(state: any, action: any) {
```
dominoweir: I know these shouldn't be typed as `any` but it's not clear to me what the correct type would be... thoughts?

packages/code/src/Code.tsx
```diff
@@ -80,6 +92,131 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+    color: ${uiColors.gray.base};
+  `;
+}
+
+const ScrollState = {
+  None: 'none',
+  Left: 'left',
+  Right: 'right',
+  Both: 'both',
+};
```
brookescarlett: this can probably be `as const` right?

brookescarlett: And then we can turn this into a type as well so that we can use it in the getScrollShadowStyle function

packages/code/src/Code.tsx
```diff
@@ -80,6 +92,131 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+
+function getScrollShadowStyle(scrollState: string, variant: Variant): string {
```
brookescarlett: See above

packages/code/src/Code.tsx
```diff
@@ -80,6 +92,131 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+    `;
+  }
+
+  if (scrollState === ScrollState.Left) {
+    return css`
+      box-shadow: inset 6px 0 6px -6px ${shadowColor};
+    `;
+  }
+
+  if (scrollState === ScrollState.Right) {
```
brookescarlett: So much cleaner this looks great

packages/code/src/Code.tsx
```diff
@@ -80,6 +92,131 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+
+  return '';
+}
+
+interface ActionType {
+  type: string;
+}
+
+interface StateType {
+  scroll: string;
```
brookescarlett: this should also be able to be more specifically typed, see above

packages/code/src/Code.tsx
```diff
@@ -187,17 +398,45 @@ function Code({
...
+          display: flex;
+        `}
       >
-        {showLineNumbers && (
-          <LineNumbers variant={variant} lineCount={lineCount} />
-        )}
+        <pre
+          {...(rest as DetailedElementProps<HTMLPreElement>)}
+          className={wrapperClassName}
+          onScroll={e => {
```
brookescarlett: I believe the `e` is implied as the first argument so you should just be able to `onScroll={handleScroll}`

dominoweir: yup, that's what I had initially. I was playing around with it when I was running in to "Maximum depth exceeded" errors implementing useReducer and never changed it back to `onScroll={handleScroll}`. Fixed now!

packages/code/src/Code.tsx
```diff
@@ -80,6 +92,136 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+  isLoaded: boolean;
+}
+
+const initialState = {
+  scroll: ScrollState.None,
+  copied: false,
+  isLoaded: false,
+};
+
+function scrollReducer(state: StateType, action: ActionType) {
```
hswolff: So...seems like your refactor cleaned up the state changes we have to support, such that we're only support one state change per dispatch.

That being the case...it may make sense to go back to using useState as our state transitions are simple again and not inter-related.

I don't want to just keep pointing you in a new direction, but hopefully you understand the reasoning behind the shifts in my thinking?

packages/code/src/Code.tsx
```diff
@@ -140,8 +282,34 @@ function Code({
   chromeTitle = '',
   ...rest
 }: CodeProps) {
+  const scrollableMultiLine = useRef<HTMLPreElement>(null);
+  const scrollableSingleLine = useRef<HTMLDivElement>(null);
+  const [state, dispatch] = useReducer(scrollReducer, initialState);
+  useLayoutEffect(() => {
```
hswolff: nit, line break to break up variable declarations and the hook

packages/code/src/Code.tsx
```diff
@@ -7,8 +7,12 @@ import Syntax, {
   Language,
   variantColors,
 } from '@leafygreen-ui/syntax';
+import Icon from '@leafygreen-ui/icon';
+import IconButton from '@leafygreen-ui/icon-button';
 import LineNumbers from './LineNumbers';
 import WindowChrome from './WindowChrome';
+import { CopyToClipboard } from 'react-copy-to-clipboard';
+import { uiColors } from '@leafygreen-ui/palette/src';
```
DesignerDave: We should import directly from the package, rather than its source directory:

```js
```

packages/code/src/Code.tsx
```diff
@@ -70,6 +74,14 @@ const codeWrapperStyleWithWindowChrome = css`
   border-left: 0;
 `;
 
+const copyStyle = css`
+  min-width: 38px;
```
DesignerDave: How come we don't just use `width` here?

packages/code/src/Code.tsx
hswolff: We only want this running on first run of the component?

`useLayoutEffect` has the same API as `useEffect` just differs in when the callback is called.

That being said if you go to [the docs page for useEffect](https://reactjs.org/docs/hooks-reference.html#useeffect) and search for the text "If you want to run an effect and clean it up only once" to see how we can have this hook only run once.

packages/code/src/Code.tsx
```diff
@@ -169,16 +338,63 @@ function Code({
...
+            display: flex;
+          `}
         >
-          {renderedSyntaxComponent}
+          <div
+            {...(rest as DetailedElementProps<HTMLDivElement>)}
+            className={wrapperClassName}
+            onScroll={e => {
+              handleScroll(e);
+            }}
```
hswolff: Rather than making a new closure just pass the function directly to the handler. 

```suggestion
            onScroll={handleScroll}
```

packages/code/src/Code.tsx
```diff
@@ -80,6 +92,136 @@ function getWrapperVariantStyle(variant: Variant): string {
   `;
 }
 
+function getSidebarVariantStyle(variant: Variant): string {
+  const colors = variantColors[variant];
+
+  switch (variant) {
+    case 'light':
```
DesignerDave: We should use the enum here:
```js
switch (variant) {
  case Variant.Light:
    return css``;
}

File: packages/code/src/Code.tsx [hunk 19]
```diff
@@ -140,8 +282,34 @@ function Code({
   chromeTitle = '',
   ...rest
 }: CodeProps) {
+  const scrollableMultiLine = useRef<HTMLPreElement>(null);
+  const scrollableSingleLine = useRef<HTMLDivElement>(null);
+  const [state, dispatch] = useReducer(scrollReducer, initialState);
```
DesignerDave: Hmm... I'm thinking useReducer might be a bit heavy-handed here. Thoughts @hswolff?

packages/code/src/Code.tsx
```diff
@@ -140,8 +282,34 @@ function Code({
...
+  useLayoutEffect(() => {
+    if (!state.isLoaded) {
+      dispatch({ type: 'load' });
```
DesignerDave: I'm not sure I understand the case for a loading action.

packages/code/src/Code.tsx
```diff
@@ -140,8 +282,34 @@ function Code({
...
+      if (multiline) {
+        if (
+          scrollableMultiLine.current !== null &&
```
DesignerDave: I might recommend destructuring these to make the code overall more readable:

```js
const { current: multiLineEl } = scrollableMultiLine

if (
  multiLineEl != null &&
  multiLineEl.scrollWidth > multiLineEl.clientWidth
) { ... }
```

Also, did you mean to do `scrollableMultiLine.current != null`?

packages/code/src/Code.tsx
```diff
@@ -169,16 +338,63 @@ function Code({
     </Syntax>
   );
 
+  function handleScroll(e: React.UIEvent) {
```
DesignerDave: I think you can use something like `React.ScrollEvent` here to be more specific.

packages/code/src/Code.tsx
dominoweir: Made this change earlier today in one place but not others. Thanks for catching it!

packages/code/src/Code.tsx
```diff
@@ -169,16 +338,63 @@ function Code({
...
+    const scrollWidth = e.currentTarget.scrollWidth;
+    const elementWidth = e.currentTarget.clientWidth;
```
DesignerDave: Tip – the above two lines can be:

```js
const { scrollWidth, clientWidth: elementWidth } = e.currentTarget;
```

packages/code/src/Code.tsx
```diff
@@ -169,16 +338,63 @@ function Code({
...
+            <div className={cx(copyStyle, getSidebarVariantStyle(variant))}>
+              <CopyToClipboard
+                text={content}
+                onCopy={() => {
+                  dispatch({ type: 'copy' });
+                }}
+              >
+                <IconButton
+                  variant={variant}
+                  ariaLabel={'Copy'}
```
DesignerDave: Since this is a string, you can omit the curly braces (`{}`).

packages/code/src/Code.tsx
```diff
@@ -169,16 +338,63 @@ function Code({
...
-          className={wrapperClassName}
+          className={css`
+            display: flex;
+          `}
         >
-          {renderedSyntaxComponent}
+          <div
+            {...(rest as DetailedElementProps<HTMLDivElement>)}
+            className={wrapperClassName}
+            onScroll={e => {
```
DesignerDave: This can be shortened to `onScroll={handleScroll}` since you've already made a function that accepts the event.

packages/code/src/Code.tsx
```diff
@@ -169,16 +338,63 @@ function Code({
...
         <div
-          {...(rest as DetailedElementProps<HTMLDivElement>)}
-          className={wrapperClassName}
+          className={css`
+            display: flex;
+          `}
         >
-          {renderedSyntaxComponent}
+          <div
+            {...(rest as DetailedElementProps<HTMLDivElement>)}
```
DesignerDave: Usually we prefer to spread `rest` on the root element. Any particular reason you did it here?

typings/copy.d.ts
DesignerDave: This is great 👍

packages/code/src/Code.tsx
dominoweir: I totally understand. Refactoring back to useState was actually a pretty quick process and I think it came out a little more clear.

packages/code/src/Code.tsx
```diff
@@ -80,6 +92,136 @@ function getWrapperVariantStyle(variant: Variant): string {
   `;
 }
 
+function getSidebarVariantStyle(variant: Variant): string {
+  const colors = variantColors[variant];
```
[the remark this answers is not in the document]
dominoweir: I'm not sure I understand... colors is an array, wouldn't that mean that if `variantColors[variant]` doesn't exist, then we would default to the string `'light'` rather than (I think) `variantColors['light']`?

packages/code/src/Code.tsx
dominoweir: Swapped back to useEffect now 👍

packages/code/src/Code.tsx
dominoweir: Tried that but got this error: `Namespace 'React' has no exported member 'ScrollEvent'.`

packages/code/src/Code.tsx
dominoweir: I found that when I just used `width`, as more text was added and the element widened, the sidebar shrank from 38px to like 29px. If I set `min-width` instead, that shrinkage didn't happen.

packages/code/src/Code.tsx
DesignerDave: Whoops, it's actually `React.WheelEvent` I believe.

packages/code/src/Code.tsx
DesignerDave: Ahh, its parent is `display: flex`. You should be able to use `width` in combination with `flex-shrink: 0`.

packages/code/src/Code.tsx
dominoweir: This is where it was spread initially, so I can't really speak to the reasoning.

packages/code/src/Code.spec.tsx
```diff
@@ -1,32 +1,79 @@
 import React from 'react';
-import { render, cleanup } from '@testing-library/react';
+import { render, cleanup, fireEvent } from '@testing-library/react';
 import { typeIs } from '@leafygreen-ui/lib';
 import Code from './Code';
 
+// I had to add this in to prevent another error being thrown on the click event (in addition to the TypeError)
+window.prompt = jest.fn();
```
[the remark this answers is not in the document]
dominoweir: I defined a mock function for window.prompt because the error I saw was telling me that it wasn't defined and I wanted to see what would happen. Adding in that definition got rid of the error, but I wasn't sure why, so I left it in and decided to come back and ask for y'alls thoughts once I had the PR submitted. Here's the exact error I get when I comment out that line: 
```
      Error: Not implemented: window.prompt
          at module.exports (/Users/dominoweir/leafygreen-ui/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
          at /Users/dominoweir/leafygreen-ui/node_modules/jsdom/lib/jsdom/browser/Window.js:594:7
          at copy (/Users/dominoweir/leafygreen-ui/node_modules/react-copy-to-clipboard/node_modules/copy-to-clipboard/index.js:93:14)
          at /Users/dominoweir/leafygreen-ui/node_modules/react-copy-to-clipboard/lib/Component.js:69:52
          at HTMLUnknownElement.callCallback (/Users/dominoweir/leafygreen-ui/node_modules/react-dom/cjs/react-dom.development.js:336:14)
          at invokeEventListeners (/Users/dominoweir/leafygreen-ui/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:193:27)
          at HTMLUnknownElementImpl._dispatch (/Users/dominoweir/leafygreen-ui/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:119:9)
          at HTMLUnknownElementImpl.dispatchEvent (/Users/dominoweir/leafygreen-ui/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:82:17)
          at HTMLUnknownElementImpl.dispatchEvent (/Users/dominoweir/leafygreen-ui/node_modules/jsdom/lib/jsdom/living/nodes/HTMLElement-impl.js:30:27)
          at HTMLUnknownElement.dispatchEvent (/Users/dominoweir/leafygreen-ui/node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:157:21) undefined
```

packages/code/src/Code.tsx
dominoweir: It looks like the types are still different: 
`Type '(e: WheelEvent<Element>) => void' is not assignable to type '(event: UIEvent<HTMLDivElement>) => void'.`
I checked the typing for onScroll and it looks like it actually does use `UIEvent`: 
`(JSX attribute) DOMAttributes<HTMLPreElement>.onScroll?: ((event: React.UIEvent<HTMLPreElement>) => void) | undefined`

packages/code/src/Code.spec.tsx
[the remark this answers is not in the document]
dominoweir: See my comment to Dave- without this line I see an error telling me that `window.prompt` is not implemented. I'm not really sure what's up with that.

packages/code/src/Code.tsx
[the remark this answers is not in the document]
hswolff: Ah, you're right! So fallback to `Variant.Light`.

My basic idea was to make sure we always have a fallback if we don't have an empty state.

packages/code/src/Code.tsx
```diff
@@ -169,16 +290,60 @@ function Code({
...
-          className={wrapperClassName}
+          className={css`
+            display: flex;
+          `}
         >
-          {renderedSyntaxComponent}
+          <div
+            {...(rest as DetailedElementProps<HTMLDivElement>)}
+            className={wrapperClassName}
+            onScroll={handleScroll}
```
hswolff: Do we want to [debounce this? ](https://davidwalsh.name/javascript-debounce-function) Generally good practice for onScroll events.

packages/code/src/Code.spec.tsx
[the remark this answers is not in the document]
hswolff: Seems like it's not provided by [jsdom](https://github.com/jsdom/jsdom). I think this is fine. 

I might suggest we put it into beforeEach instead so it's reset before every test run, and then on afterEach we clear it out so it's put back to the state we found it in.

packages/code/src/Code.spec.tsx
hswolff: I may need to pair on this to understand it.

To not hold things up I may say let's just .skip for now and revisit later.

packages/code/src/Code.tsx
```diff
@@ -80,6 +93,86 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+}
+
+const ScrollState = {
+  None: 'none',
+  Left: 'left',
+  Right: 'right',
+  Both: 'both',
+} as const;
+
+type ScrollState = typeof ScrollState[keyof typeof ScrollState];
```
brookescarlett: 👍

packages/code/src/Code.tsx
DesignerDave: Oh interesting... I'm alright with this then.

packages/code/src/Code.tsx
```diff
@@ -169,16 +291,65 @@ function Code({
...
+            ref={scrollableSingleLine}
+          >
+            {renderedSyntaxComponent}
+          </div>
+          {!showWindowChrome && (
+            <div className={cx(copyStyle, getSidebarVariantStyle(variant))}>
+              <CopyToClipboard
+                text={content}
+                onCopy={() => {
+                  setCopied(true);
```
DesignerDave: We should unset this state after about a second so that the checkmark doesn't persist forever.

packages/code/src/Code.tsx
```diff
@@ -80,6 +94,86 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+      `;
+    case Variant.Dark:
+      return css`
+        border-color: ${colors[1]};
+        background-color: ${colors[1]};
+      `;
+  }
+}
+
+function getCopyButtonStyle(variant: Variant, copied: boolean): string {
```
DesignerDave: Noticed a problem – The background colors here should all be applied to the `:before` pseudo-element for the button, not the button itself.

Right now, the hover and focus states will cover the button's background color, making the interaction a bit weird.

packages/code/src/Code.tsx
```diff
@@ -80,6 +94,86 @@ function getWrapperVariantStyle(variant: Variant): string {
   `;
 }
 
+function getSidebarVariantStyle(variant: Variant): string {
+  const colors = variantColors[variant] ?? variantColors[Variant.Light];
```
DesignerDave: Tip – you can destructure arrays similar to objects. Instead of this, you could do:

```js
const [, variantColor] = variantColors[variant] ?? variantColors[Variant.Light];

switch (variant) {
    case Variant.Light:
      return css`
        border-color: ${variantColor};
        background-color: white;
      `;

    case Variant.Dark:
      return css`
        border-color: ${variantColor};
        background-color: ${variantColor};
      `;
}
```

packages/code/src/Code.tsx
```diff
@@ -80,6 +94,86 @@ function getWrapperVariantStyle(variant: Variant): string {
...
+
+  if (variant === Variant.Dark) {
+    return css`
+      align-self: center;
+      color: ${uiColors.gray.base};
+      background-color: ${uiColors.gray.dark3};
+    `;
+  }
+
+  return css`
```
DesignerDave: Nice happy-pathing! There's a bit of style repetition here that we can clean up using `cx`:

```js
  const baseStyle = css`
    align-self: center;
    color: ${uiColors.gray.base};
  `;

  if (copied) {
    return cx(baseStyle, css`
      color: ${uiColors.white};
      background-color: ${uiColors.green.base};
    `);
  }

  if (variant === Variant.Dark) {
    return cx(baseStyle, css`
      background-color: ${uiColors.gray.dark3};
    `);
  }

  return baseStyle;
```

packages/code/src/Code.tsx
```diff
@@ -140,8 +234,35 @@ function Code({
...
+        singlelineEl.scrollWidth > singlelineEl.clientWidth
+      ) {
+        setScrollState(ScrollState.Right);
+      }
+    }
+  }, []);
+
   const wrapperStyle = css`
-    display: inline-block;
+    display: block;
```
DesignerDave: Divs are `display: block` by default, so this probably isn't needed.

packages/code/src/Code.tsx
DesignerDave: Hmm... might need to explore this a bit more after this change then.

dominoweir: I think the idea is that `rest` is spread on the `<pre>` or `<div>` element that actually holds  `renderedSyntaxComponent` so that you can make changes to how the code is displayed... but that's just my guess.

packages/code/src/Code.tsx
dominoweir: I tried this but got back the error: `Type 'Base16Palette' must have a '[Symbol.iterator]()' method that returns an iterator.`

packages/code/src/Code.tsx
```diff
@@ -169,16 +307,67 @@ function Code({
...
+        setScrollState(ScrollState.Left);
+      } else if (scrollPosition < maxPosition) {
+        setScrollState(ScrollState.Right);
+      }
+    }
+  }
+
+  const debounceScroll = debounce(handleScroll, 50, { leading: true });
+
+  new ClipboardJS('.copy-btn');
```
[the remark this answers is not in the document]
hswolff: Also it looks like [ClipboardJS supports instantiating with a node element](https://github.com/zenorocha/clipboard.js/blob/master/demo/constructor-node.html#L19-L20), which I think may be a little more robust for our use case.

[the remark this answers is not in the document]
dominoweir: yup, we can stick this in useLayoutEffect so that it only runs after that initial render, and that instantiation with a node element works great.

packages/code/src/Code.tsx
```diff
@@ -140,8 +257,48 @@ function Code({
...
+
+  useEffect(() => {
+    let timeoutId: any;
+
+    if (copied) {
+      timeoutId = setTimeout(() => setCopied(false), 1500);
+    }
+
+    return () => clearTimeout(timeoutId);
+  });
```
hswolff:
```suggestion
  }, [copied]);
```

We only want this to run when the value of copied changes.

dominoweir:
Yeah, I'm seeing the same problem as you @DesignerDave 
I last tested on Friday and things seemed to be fine, but it's totally possible that I had the correct copied text in my clipboard already and therefore didn't see/comprehend the problem. Want me to look in to it?

DesignerDave:
If you have the time to check into it, that would be awesome! That said, you just started a new rotation, so let us know if you need us to look into it (I totally understand if that's the case).

dominoweir:
@DesignerDave I think I have resolved the problem, can you let me know if you see the same thing on your end?

PR #291, opened 2020-03-02
PD-538: Add activeOrgId and activeProjectId props

brookescarlett:
## ✍️ Proposed changes

- Add `activeOrgId` and `activeProjectId` props 
- Conditionally POST/GET based on the presence of those props
- Add post functionality to /user/shared

🎟 _Jira ticket:_ [PD-538](https://jira.mongodb.org/browse/PD_538
## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -163,6 +180,16 @@ export default function MongoNav({
     });
   }
 
+  function fetchProductionData(body: PostBodyInterface) {
```
hswolff: Fairly sure you could figure out a way to merge this and the get function together. So all fetching is in one location. Don't think it'd add that much complexity either.

hswolff: Related to that we should test that the proper behavior occurs too. 

Ideally with new functionality we also add tests to add coverage there.

packages/mongo-nav/src/MongoNav.story.tsx
```diff
@@ -25,6 +25,7 @@ storiesOf('MongoNav', module).add('Default', () => (
         onOrganizationChange={() => {}}
         onProjectChange={() => {}}
         admin={boolean('admin', true)}
+        hosts={{ cloud: 'http://localhost:8080' }}
```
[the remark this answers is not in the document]
brookescarlett: Removed already 👍

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -101,6 +102,18 @@ interface MongoNavInterface {
    * onPrem config object with three keys: enabled, version and mfa
    */
   onPrem?: OnPremInterface;
+
+  /**
+   * ID for active organization, will cause a POST request to the database to update
```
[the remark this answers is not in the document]

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -185,12 +212,17 @@ export default function MongoNav({
   useEffect(() => {
     if (mode === Mode.Dev) {
       getDataFixtures().then(data => setData(data as DataInterface));
+    } else if (activeProjectId || activeOrgId) {
+      const body = activeProjectId ? { activeProjectId } : { activeOrgId };
+      fetchProductionData(body)
```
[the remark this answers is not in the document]

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: Something like:

```fetchProductionData(body) {
  if(body) { POST }
  else { GET } 
}```

Comment (hswolff) [reply packages/mongo-nav/src/MongoNav.tsx]: Ideally we only use `fetch` function once and change the args we send to it.

File: packages/mongo-nav/src/MongoNav.tsx [hunk 5]
```diff
@@ -186,11 +216,18 @@ export default function MongoNav({
     if (mode === Mode.Dev) {
       getDataFixtures().then(data => setData(data as DataInterface));
     } else {
-      getProductionData()
+      const body =
+        activeProjectId || activeOrgId
+          ? activeProjectId
+            ? { activeProjectId }
```
hswolff: You can send both values to the server - `{ activeProjectId, activeOrgId }`. The server will ignore orgId if projectId is given.

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -1,68 +1,143 @@
...
+
+  let onOrganizationChange: jest.Mock;
+  let onProjectChange: jest.Mock;
+  let fetchMock: jest.Mock;
+
+  beforeEach(() => {
+    onOrganizationChange = jest.fn();
+    onProjectChange = jest.fn();
+    fetchMock = jest.fn();
+    window.fetch = fetchMock;
```
hswolff: Gotta make sure you restore the original fetch function.

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.spec.tsx
brookescarlett: Updated!

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -287,3 +324,31 @@ export default function MongoNav({
     </section>
   );
 }
+
+MongoNav.displayName = 'MongoNav';
```
seanmhanson: NOT ALL HEROES WEAR CAPES 🙇

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -172,6 +203,7 @@ export default function MongoNav({
 
   async function handleResponse(response: Response) {
     if (!response.ok) {
+      console.log(response);
```
hswolff: rm

packages/mongo-nav/src/MongoNav.spec.tsx
hswolff: If you go to another test does window.fetch exist?

What I had in mind was:

```jsx
let originalFetch = window.fetch;
window.fetch = fetchMock;

// afterEach
window.fetch = originalFetch
```

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -1,68 +1,310 @@
...
+    const responseObject = {
+      ok: true,
+      json: () => Promise.resolve(dataFixtures),
+    };
+
+    beforeEach(async () => {
+      fetchMock.mockResolvedValue(responseObject);
+      renderComponent();
+
+      await act(() => responseObject.json());
```
hswolff: nice

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -1,68 +1,310 @@
...
-    test('it renders admin as false', () => {
-      const orgNav = getByTestId('organization-nav');
-      expect(orgNav.innerHTML.includes('Admin')).toBe(false);
+    test('fetch is called', () => {
+      expect(fetchMock).toHaveBeenCalledTimes(1);
+    });
+
+    test('the organization nav is rendered', () => {
+      const organizationNav = expectedElements['orgNav'];
+      expect(organizationNav).toBeInTheDocument();
```
hswolff: Why so much indirection? Why not just:

```suggestion
      expect(expectedElements.orgNav).toBeInTheDocument();
```

Ditto for other methods. 

I'd also just combine this and the next test to assert that the dom contains the elements you expect. No value added having them be separate.

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -1,68 +1,310 @@
...
-        'https://cloud.mongodb.com/access-manager-test',
+    test('the project nav is rendered', () => {
+      const projectNav = expectedElements['projectNav'];
+      expect(projectNav).toBeInTheDocument();
+    });
+
+    test('current orgId is set based on data returned from fetch', () => {
+      const billing = expectedElements['billing'];
+      expect((billing as HTMLAnchorElement).href).toBe(
+        `https://cloud.mongodb.com/v2#/org/${dataFixtures?.currentOrganization?.orgId}/billing/overview`,
```
hswolff: We know this value should exist right? In that case we can do:

```suggestion
        `https://cloud.mongodb.com/v2#/org/${dataFixtures!.currentOrganization!.orgId}/billing/overview`,
```

The `!` is a TS feature that tells TS that I know this value exists.

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: 👍 sorry about that

packages/mongo-nav/src/MongoNav.spec.tsx
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.spec.tsx
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.spec.tsx
brookescarlett: 👍

tomhollander:
What's the interplay between `activeOrgId` and `activeProjectId`? Each project is in a single org, so if I provide the `activeProjectId` do I also need to provide the `activeOrgId`? What if I provide both but the project isn't in that org? What if I provide an `activeOrgId` but no `activeProjectId`? 

Note, in Charts we have easy access to the Project ID but I think it would be harder to get the Org ID - so if possible it would be best to only set the Project ID and have the Org ID inferred.

brookescarlett:
Hi @tomhollander, I hope this answers your question: 
- If you provide an activeProjectID you do not need to supply an activeOrgId, the orgID will be inferred. 
- If you provide both, the server will ignore the activeOrgId, and only use the activeProjectId. 
- If you provide an activeOrgId without an activeProjectId the org will appear active but without a specific project.

PR #295, opened 2020-03-03
Update Popover with center alignments and Justify.Fit

jetpacmonkey:
## ✍️ Proposed changes

As part of the work for adding a `<Select>` component, Popover needed some additional functionality. This PR adds support for `align={Align.CenterHorizontal}`, `align={Align.CenterVertical}`, and `justify={Justify.Fit}`. I also removed the `Justification` enum, as it's become redundant with `Align`, and refactored a bunch of internals in `positionUtils` to derive values directly from `Align` and `Justify`, without using `Justification` as a go-between.

Some of the changes led to test failures in the `positionUtils` unit tests because of a change from `30` to `'30px'` (for example), so instead of changing them all over, I wrote a custom matcher that accepts either of the two formats. This feels like a better test (to me, at least) since it should generally avoid false positives on changes that don't actually impact behavior.

🎟 _Jira ticket:_ [PD-271](https://jira.mongodb.org/browse/PD-271)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

This is, unfortunately, a breaking change, since `justification` was being passed to the function-as-a-child pattern. I've also included a fix for Tooltip, which was relying on that behavior. I would assume that we probably don't have any external consumers relying on Popover, so that _should_ be sufficient.


jetpacmonkey:
Oh, I guess I should probably add testcases for the new enum values. Regardless, this is ready for review while I do that.

jetpacmonkey:
Alright, that should be full test coverage for the new values now

jetpacmonkey:
Note: The `Justify.Fit` value impacts the width/height of the child, which can cause some wonkiness when changing from `Fit` to one of the other `Justify` values. For instance...

1. `align={Align.Top} justify={Justify.Fit}`: left: 0, width: parent width
2. `justify={Justify.Middle}`: Since child width and parent width are equal, centering means left: 0
3. Browser re-renders, and now that we're not explicitly setting the child width it changes back to being based on its content, while being left-aligned on the parent.

The popover will then correct its position the next time the browser window is resized or if the mutation observer is triggered.

Changes to the `justify` prop seems like a pretty rare edge case to me, but thought it should be mentioned.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -67,7 +66,7 @@ const notchVariants = {
 
 interface PopoverFunctionParameters {
   alignment: Align;
-  justification: Justification | Justify;
+  justification: Justify;
```
hswolff: Can we change this key to justify too since it's going to be a breaking change anyways?

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -1,8 +1,9 @@
 import { css } from '@leafygreen-ui/emotion';
+import { Align, Justify } from '@leafygreen-ui/popover';
 
 export function trianglePosition(
-  alignment: string,
-  justification: string,
+  alignment: Align,
+  justification: Justify,
```
hswolff: Ditto rename here.

scripts/expectToBePx.ts
```diff
@@ -0,0 +1,53 @@
+import diff from 'jest-diff';
+
+expect.extend({
+  toBePx(received, expected) {
```
hswolff: I mean, I guess so? I'd just as much update the tests to have them match so we don't have to maintain this.

packages/tooltip/src/Tooltip.tsx
jetpacmonkey: should we also change `alignment` to `align` then?

hswolff:
I'm in favor of that. I think our internal names are good enough to be the
same that's used externally, right?

scripts/expectToBePx.ts
jetpacmonkey: My first pass updated all the tests with a regex replace, but then it turned out that not all of the internal functions seemed to be consistent with whether it used a number or a string with `px`. I suppose a helper function in the test file might be easier to maintain, though.

hswolff: I feel semi strongly to not take on this debt, but I'm curious about @DesignerDave's thoughts.

packages/tooltip/src/Tooltip.tsx
hswolff: I'm in favor of that. I think our internal names are good enough to be the
same that's used externally, right?

scripts/expectToBePx.ts
jetpacmonkey:
```ts
function checkPixelValue(actual: string | number, expected: number) {
  if (typeof actual === 'string') {
    expect(actual).toBe(`${expected}px`);
  } else {
    expect(actual).toBe(expected);
  }
}
```

it's a pretty easy regex find/replace to just use this instead, so that's probably worth the small tradeoff in having very slightly worse diffs and error messages.

jetpacmonkey: Updated with the helper function

packages/tooltip/src/Tooltip.tsx
jetpacmonkey: Actually, nice catch! It looks like I'd already changed it in `Popover` to be `justify` and TS somehow didn't catch it...

jetpacmonkey:
@hswolff addressed your comments

jetpacmonkey:
@DesignerDave I just pushed up a commit that fixes the issue when changing to/from `justify={Justify.Fit}` that I described on Tuesday

packages/popover/src/Popover.story.tsx
```diff
@@ -8,13 +8,19 @@ import Popover, { Align, Justify } from '.';
 
 const containerStyle = css`
   position: absolute;
+  min-height: 50px;
+  appearance: none;
+  box-shadow: 0 0 4px #000;
+  background-color: ${colors.gray[5]};
```
brookescarlett: Nit, but while you're in here...
Can we pick a gray from palette instead of theme? I believe this is the only time that the theme package is used in Popover and if we can swap that over to palette that'd be great as we are trying to deprecate theme. @DesignerDave not sure if you have a preference on which gray?

jetpacmonkey: Sure, I just used it because it was already being imported, but it'd be easy enough to switch that over. It should probably just be a devDependency, too, if it's only being used in storybook. Looks like right now it's listed under `dependencies`.

brookescarlett: 👍  Thank you so much

packages/popover/src/positionUtils.ts
```diff
@@ -123,70 +118,51 @@ export function getElementPosition(
...
+  [Align.CenterHorizontal]: { x: 'center' },
+  [Align.CenterVertical]: { y: 'center' },
+};
 
-    case Justification.CenterVertical:
-      y = 'center';
-      break;
-  }
+// Constructs the transform origin for any given pair of alignment / justification
+function getTransformOrigin({ align, justify }: TransformOriginArgs): string {
```
brookescarlett: love this

packages/popover/src/positionUtils.ts
```diff
@@ -208,97 +189,186 @@ interface AbsolutePositionObject {
...
+  })}px`;
 
-    case Align.Right:
-      positionObject.left = `calc(100% + ${spacing}px)`;
-      break;
+  if (justify !== Justify.Fit) {
+    return { left, top };
   }
 
-  switch (justification) {
```
brookescarlett: Thank you for removing these crazy switch statements

DesignerDave:
Thanks for the PR Jeremy! Sorry I didn't finish up with the review last week.

A few things I've found:
- It looks like the API just got more complex by switching from `center` to `center-horizontal` and `center-vertical`. It's something we can, and used to infer from justify, and I'm not sure we should be taking steps backwards in usability. That said, outside of our own internal usage, I'm not sure how many people use the component and I could be convinced this is fine... I also see how it's hard to define the center behavior for fit without this, but this still feels like we're going backwards a bit.
- It looks like the README wasn't updated for this change.
- I've run into some issues when switching between different alignments and justifications. Mostly, switching from anything with `alignment='center-vertical' justify='fit'` to `alignment='center-vertical' justify={/* anything else */}`seems to result in this:
Screen Shot 2020-03-09 at 2 03 15 PM

jetpacmonkey:
That last point _should_ have been fixed by 8b7b60d, when did you pull the branch down? As far as the two centers, `center` never existed for `Align`, and I'm fairly certain you can't infer it based on `justify`. The only way you'd be able to have a simple `center` value is if you had some sort of equivalent to `flex-direction`.

I'll make sure to update the README, I definitely did miss that.

DesignerDave:
Ahh, my apologies, may have been on an outdated version. I'll give that another shot.

We inferred center vertical / horizontal based on the alignment being left/right vs top/bottom previously.

jetpacmonkey:
Right, and that's still effectively the case if you do align=left and justify=middle or whatever, although since I removed the `Justification` enum the end user might have to do a bit of that logic for themselves in the function-as-a-child version. But if you want your primary _alignment_ to be in the center, you need some way to distinguish which axis it is that you're centering on, and there's no way to infer that based off of just `start`, `end`, `middle`, or `fit`.

DesignerDave:
Alright, I think I was misunderstanding.

With the latest I'm still finding issues when switching between alignments:

`align: center-vertical` with `justify: middle`:
Screen Shot 2020-03-09 at 3 55 14 PM

`align: center-vertical` with `justify: fit`:
Screen Shot 2020-03-09 at 3 57 07 PM

`align: fit` with `justify: fit`:
Screen Shot 2020-03-09 at 4 03 28 PM

All of these were triggered from switching to, and then from a center alignment. I'm having trouble reproducing the padding issue in particular, but I can pretty consistently reproduce the alignment issues.

jetpacmonkey:
I switched the storybook popover to have `overflow: hidden` to more directly see it changing the widths/heights, so that's probably the padding issue. The alignment being off, though, is odd, and I'm not seeing the same behavior on my machine. Also what is `align: fit`?

DesignerDave:
Whoops, copy + paste error. `align: left` I believe that was.

Try this:
1. Refresh the page
2. Open the popover
3. Change align to `center-vertical`
4. Change justify to `fit`
5. Change align to `left`

That should get you to at least the last screenshot I sent.

Doing the following will also result in a strange alignment:
1. Refresh the page
2. Open the popover
3. Change align to `center-vertical`
4. Change justify to `fit`
5. Change align to `center-horizontal`

jetpacmonkey:
Oh, there it is! Just reproduced it.

jetpacmonkey:
Alright, so I had it re-render when changing to or from `fit`, but it seems like I probably also should have it re-render when justify is `fit` and you're switching between one of the "column" direction alignments and one of the "row" direction alignments.

jetpacmonkey:
Thanks for the thorough debugging @DesignerDave! Just pushed up an update that should cover those additional cases. I'm not sure if there's a good way to write an automated test for those types of cases without a browser environment, but I also doubt that there's a realistic usecase for switching between a vertically-oriented alignment and a horizontal one, so we should be fine I think.

DesignerDave:
@jetpacmonkey Those fixes seem to be working well! Thanks for the update.

Little more QA on components currently using popover:

**Tooltip –** `align: center-horizontal | center-vertical; justify: middle;`
Screen Shot 2020-03-11 at 10 43 42 AM
The arrow is now positioned strangely likely due to the alignment not being handled (the top of the "I" is being cut off). I wouldn't be opposed to not supporting `center-horizontal` and `center-vertical` at all in tooltip, but if we do, we should at least hide the arrow. Ditto for disabling `justify: fit` for tooltips for the time being.

Everything else is looking good now it seems! Awesome stuff.

jetpacmonkey:
I'm back from the worst-timed vacation in history! I don't think I realized that Tooltip was extending `PopoverProps` directly, I agree that the new alignments probably don't need to be supported at all. `justify: fit` is probably harmless to allow on its own, but I'll skip it if there seem to be issues.

jetpacmonkey:
@DesignerDave I think that should be everything. I left `justify: fit` in there for Tooltip even though it's not a super likely usecase, because it seemed to work without any other modifications.

jetpacmonkey:
Oh I should also add this to the table of contents in the root README, shouldn't I

DesignerDave:
Hey @jetpacmonkey, sorry for the delayed response.

I was testing out tooltip, and it looks like align left/right with justify fit doesn't work:
Screen Shot 2020-03-26 at 2 41 47 PM

Fit should work for those alignments, right? Otherwise this seems ready to go!

jetpacmonkey:
Yeah, that's actually working the same as it does in `Popover`. If your popover/tooltip has a defined height (or width for top/bottom alignments) that's larger than the trigger, or enough content and padding to make the popover grow to be bigger without `overflow: hidden`, that overrides setting `top`/`bottom`. That's why I made the change to the `Popover` storybook to have a taller button and `overflow: hidden` on the popover.

PR #297, opened 2020-03-04
PD-558: Add onElementClick prop

brookescarlett:
## ✍️ Proposed changes

- Adds onElementClick prop which receives a type and an event
- type can be from OnElementClick type
- Replace onLogout with onElementClick(OnElementClick.Logout) 

🎟 _Jira ticket:_ [PD-558](https://jira.mongodb.org/browse/PD-558)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/types.ts
```diff
@@ -1,3 +1,14 @@
+const OnElementClick = {
```
hswolff: Just to spice things up...I have half a mind that this should legitimately be an enum, such that we completely control the cases and force end users to use the enum to switch behaviors off of.

I might be getting ahead of myself tho, so this is still fine as is.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -108,6 +109,13 @@ interface MongoNavInterface {
    * Applies a className to the root element
    */
   className?: string;
+
+  /**
+   * Click EventHandler that receives a `type` as its first argument and the associated `MouseEvent` as its second
+   * `type` can be ['logout', 'cloud', 'realm', 'charts']
```
hswolff: I wouldn't include this here, just ask the user to refer to the object, so there's no risk in things getting out of sync.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -277,7 +285,7 @@ export default function MongoNav({
         admin={admin}
         hosts={hosts}
         currentProjectName={data?.currentProject?.projectName}
-        onLogout={onLogout}
+        onElementClick={onElementClick}
```
hswolff: So rather than tunnel this down to every level one thought I had was to make a local Context that just exposes this prop so any descendent component can access it. And then you can expose the function via a hook.

```jsx
// ProjectTrigger

// or even better. This works if there's no conditionals before it.
<div onClick={useOnElementClick()} />
```

Up to you.

hswolff: Also a more fun API for children is to allow this usage:

```jsx
<div name="logout" onClick={useOnElementClick(OnElementClick.Logout)} />
```

Such that `useOnElementClick` returns a function that takes in the event and then calls the `props.onElementClick` with the expectant API.

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/project-nav/ProjectNav.spec.tsx
```diff
@@ -146,6 +146,10 @@ describe('packages/mongo-nav/src/project-nav', () => {
       testForVisibleProducts(product, product !== 'cloudManager'),
     );
 
+    // Object.values(Products).forEach(product =>
```
hswolff: rm

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -29,6 +30,20 @@ const navContainerStyle = css`
   position: relative;
 `;
 
+type OnElementClickType = (
+  type?: OnElementClick,
+  event?: React.MouseEvent,
+) => void;
+
+export const OnElementClickContext = createContext<OnElementClickType>(
```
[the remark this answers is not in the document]
hswolff: Also let's move this to its own file so we don't have many circular dependencies in other files in this package.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -29,6 +30,20 @@ const navContainerStyle = css`
...
+  () => {},
+);
+
+export function useOnElementClick() {
+  const onElementClick = useContext(OnElementClickContext);
+  return onElementClick;
```
hswolff: Can just do 

```suggestion
  return useContext(OnElementClickContext);
```

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -189,20 +191,29 @@ interface UserMenuProps {
...
   account,
   activeProduct,
-  onLogout = () => {},
+  onLogout: onLogoutProp,
   onProductChange = () => {},
   urls: urlsProp,
   hosts: hostsProp,
 }: UserMenuProps) {
   const hosts = defaultsDeep(hostsProp, hostDefaults);
+  const onElementClick = useOnElementClick();
```
hswolff: Does this work?

Also ideally the prop gets called with the same API as everything else. 

```suggestion
  const onElementClick = useOnElementClick() ?? onLogoutProp;
```

packages/mongo-nav/src/project-nav/ProjectNav.spec.tsx
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: the prop doesn't need the type of onElementClick though

hswolff: Oh, didn't realize it was still named onLogout. Rightio, SGTM.

packages/mongo-nav/src/MongoNav.tsx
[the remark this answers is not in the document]

packages/mongo-nav/src/user-menu/UserMenu.tsx

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -123,6 +120,12 @@ interface MongoNavInterface {
    */
   className?: string;
 
+  /**
+   * Click EventHandler that receives a `type` as its first argument and the associated `MouseEvent` as its second
+   * This prop provides a hook into product link and logout link clicks and allows consuming applications to handle routing internally
+   */
+  onElementClick?: (type?: OnElementClick, event?: React.MouseEvent) => void;
```
hswolff: Are these arguments optional? Aren't they always going to be there?

PR #301, opened 2020-03-06
Harry mongo nav tweaks

hswolff:
Check each commit for description of that atomic change. 

- Unify OnElementClick and NavItem enums to one enum

- Revise exports from index.ts and also export NavItem type

- Rename NavItem to NavElement

- When clicking a project menu item close the menu

- move all useState calls in Menu into one location

- Add remaining activeNav values

## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/menu/src/Menu.tsx
+++ b/packages/menu/src/Menu.tsx
@@ -97,11 +97,20 @@ function Menu({
 
   const hasSetInitialFocus = useRef(false);
   const hasSetInitialOpen = useRef(false);
+  const [focused, setFocused] = useState<HTMLElement>(refs[0] || null);
+
   const [, setClosed] = useState(false);
   const [currentSubMenu, setCurrentSubMenu] = useState<React.ReactElement<
     HTMLLIElement
   > | null>(null);
+  const [uncontrolledOpen, uncontrolledSetOpen] = useState(false);
+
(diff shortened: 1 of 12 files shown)


packages/menu/src/Menu.tsx
```diff
@@ -97,12 +97,21 @@ function Menu({
 
   const hasSetInitialFocus = useRef(false);
   const hasSetInitialOpen = useRef(false);
+  const [focused, setFocused] = useState<HTMLElement>(refs[0] || null);
```
hswolff: Drive by, just moving some code around that increased my readability. functionality is the same

brookescarlett: Should we run a changeset against this, or just wait until we update Menu further to publish?

packages/mongo-nav/src/types.ts
```diff
@@ -1,13 +1,23 @@
-const OnElementClick = {
+const NavElement = {
```
brookescarlett: Much happier with this combined, and also makes the OnElementClick logic more readable 👍

packages/menu/src/Menu.tsx
hswolff: This has no functional change at all, so can just wiat.

PR #302, opened 2020-03-06
PD-561: Logic to determine when to display Realm/Stitch

brookescarlett:
## ✍️ Proposed changes

- Hard coding into the component to switch from Stitch to Realm

🎟 _Jira ticket:_ [PD-561](https://jira.mongodb.org/browse/PD-561)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -198,6 +198,17 @@ const tooltipStyles = css`
   text-align: center;
 `;
 
+function displayProductName() {
```
hswolff: I'd just do this one and save it in a variable. No need to keep running this function on every render.

brookescarlett: secondTabName it is

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -229,6 +240,8 @@ export default function ProjectNav({
   const isCloudManager = current?.planType === PlanType.Cloud;
   const onElementClick = useOnElementClick();
 
+  const secondTabName = displayProductName();
```
hswolff: You can make this variable outside of the component!

brookescarlett: Sorry duh that was the whole point

packages/mongo-nav/src/project-nav/ProjectNav.spec.tsx
```diff
@@ -169,4 +171,51 @@ describe('packages/mongo-nav/src/project-nav', () => {
 
     testForAlerts(alerts, true);
   });
+
+  describe('when the date is before MongoDB World', () => {
+    test('Stitch is displayed in the ProjectNav', () => {
+      const productName = displayProductName();
```
hswolff: You should mock this, cuz after May 4th this test will fail.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -198,6 +198,19 @@ const tooltipStyles = css`
   text-align: center;
 `;
 
+export function displayProductName() {
```
hswolff: FYI another strategy for testing the dates is to do this:

```suggestion
export function displayProductName(today = new Date(Date.now()), ) {
```

packages/mongo-nav/src/project-nav/ProjectNav.spec.tsx
brookescarlett: Smart 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
brookescarlett: Even better...

PR #304, opened 2020-03-06
PD-564: SideNavItem active vs hover highlight sizes are inconsistent

brookescarlett:
## ✍️ Proposed changes

- Fix active/hover highlight size inconsistency

🎟 _Jira ticket:_ [PD-564](https://jira.mongodb.org/browse/PD-564)

## 🛠 Types of changes

## ✅ Checklist


packages/side-nav/src/SideNav.story.tsx
```diff
@@ -17,7 +17,7 @@ function MockSideNav() {
     <SideNav>
       <SideNavGroup header={textHeader}>
         <SideNavItem active>Active State</SideNavItem>
-        <SideNavItem disabled>Disabled State</SideNavItem>
+        <SideNavItem>Disabled State</SideNavItem>
```
matt-d-rat: I believe you are missing the `disabled` prop in this demo.

brookescarlett: you're right, thank you! added back in

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -83,9 +85,11 @@ const activeStyle = css`
   background-color: ${uiColors.green.light3};
 
   &:hover {
+    background-color: ${uiColors.green.light3};
```
DesignerDave: So instead of setting background color on the button itself, that color should always be on the pseudo-element `:before`. Right now, this is making what was previously a smooth transition into one that looks stepped.

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -83,9 +85,11 @@ const activeStyle = css`
...
     color: ${uiColors.green.dark3};
 
     &:before {
+      transform: scale(0.95, 0.85);
```
DesignerDave: Shouldn't we make this `scale(1)`?

PR #307, opened 2020-03-09
PD-569: Hide Scrollbars in MongoSelect

brookescarlett:
## ✍️ Proposed changes

- Room for scrollbars displaying unnecessarily in OrgSelect and ProjectSelect components
- Made overflow logic more specific to overflow-y, and conditionally applying the CSS rule based on the number of MenuItems that will appear in the Select components 

- To test locally, update OS preferences to always display scroll bars and add varying levels of orgs/projects to the data.ts file

- With 5 MenuItems: Screen Shot 2020-03-09 at 9 46 02 AM
- With 5+: Screen Shot 2020-03-09 at 9 46 10 AM

🎟 _Jira ticket:_ [PD-569](https://jira.mongodb.org/browse/PD-569)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/mongo-nav/src/mongo-select/MongoSelect.tsx
+++ b/packages/mongo-nav/src/mongo-select/MongoSelect.tsx
@@ -39,10 +39,7 @@ const ulStyle = css`
   max-height: ${5 * menuItemHeight}px;
   list-style: none;
   padding: unset;
-`;
-
-const overflowScroll = css`
-  overflow-y: scroll;
+  overflow-y: auto;
 `;
 
 const viewAllStyle = css`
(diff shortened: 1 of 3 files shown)


packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -37,11 +37,14 @@ const menuItemContainerStyle = css`
 
 const ulStyle = css`
   max-height: ${5 * menuItemHeight}px;
-  overflow: scroll;
```
hswolff: Does just doing `overflow-y: auto` here solve this without adding the conditional css classes?

brookescarlett: Didn't think of that, let me try

brookescarlett: Lol yes, much simpler. Thank you

PR #311, opened 2020-03-10
PD-534: Support Segment Tracking calls in new navs

brookescarlett:
## ✍️ Proposed changes

- Definitely breaking some rules of `react-testing-library` which I am gathering based on the number of errors in the console when I run the test suite, but they all passed and everythign else works as expected so I figured I'd let y'all review while I clean up the tests

- Added testids throughout
- Added several elements to the NavElement enum and prefixed names for clarity
- Now every nav element that @nellem23 wanted to test should be available through the onElementClick function

🎟 _Jira ticket:_ [PD-534](https://jira.mongodb.org/browse/PD-534)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -224,7 +235,10 @@ function ProjectSelect({
       placeholder={current?.projectName ?? ''}
       open={open}
       loading={loading}
-      onClick={() => setOpen(curr => !curr)}
+      onClick={e => {
```
DesignerDave: This onClick doesn't have a typed event parameter like the others.

As an aside, you might consider making a wrapper for these repeated onClick handlers. Something like this:

```ts
type ClickHandler = (e: React.MouseEvent) => void

function getOnClickHandler(elementId: NavElement, callback?: ClickHandler): ClickHandler {
  return function(e) {
    onElementClick(elementId)

    if (callback) {
      callback(e)
    }
  }
}
```

Then you can use it in most cases like:
```js
onClick={getOnClickHandler(NavElement.ProjectNavViewAllProjects)}
```
And in cases like this, you can do:
```js
onClick={getOnClickHandler(NavElement.ProjectNavProjectSelectTrigger, () => setOpen(curr => !curr))}
```

packages/mongo-nav/src/on-element-click-provider/OnElementClick.spec.tsx
```diff
@@ -0,0 +1,273 @@
...
+    expectedElements[NavElement.OrgNavDropdown] = queryByTestId(
+      'org-nav-dropdown',
+    );
+    expectedElements[NavElement.OrgNavDropdownOrgAccessManager] = queryByTestId(
+      'org-nav-dropdown-org-access-manager',
+    );
+    expectedElements[
+      NavElement.OrgNavDropdownProjectAccessManager
+    ] = queryByTestId('org-nav-dropdown-project-access-manager');
+    expectedElements[NavElement.OrgNavSupport] = queryByTestId(
```
DesignerDave: This is a lot of repeat logic. Couldn't you do something like this?
```js
const expectedElements = Object.keys(NavElement).reduce((acc, val) => {
  acc[val] = queryByTestId(NavElement[val])
  return acc
}, {} as ExpectedElements)
```

packages/mongo-nav/src/on-element-click-provider/OnElementClick.spec.tsx
```diff
@@ -0,0 +1,273 @@
+import React from 'react';
+import { render, cleanup, act, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import { nullableElement, Queries } from '@leafygreen-ui/lib/src/testHelpers';
+import MongoNav from '../MongoNav';
+import { NavElement } from '../types';
```
DesignerDave: Could we add a test to ensure all values are unique in the NavElement object?

packages/mongo-nav/src/on-element-click-provider/OnElementClick.spec.tsx
brookescarlett: Had to rewrite test suite, still not super clean but lmk what you think now... significantly decreased the warnings in the console..

packages/mongo-nav/src/on-element-click-provider/OnElementClick.spec.tsx
brookescarlett: 👍

packages/mongo-nav/src/on-element-click-provider/OnElementClick.spec.tsx
DesignerDave: Seems significantly cleaner than before 👍

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
brookescarlett: Spoke off-line, leaving as-is

brookescarlett:
@hswolff @DesignerDave do we want to release this as MongoNav v1 once it's in?

.changeset/hip-gorillas-joke.md
```diff
@@ -0,0 +1,5 @@
+---
```
hswolff: Need additional changeset for segment tracking calls.

packages/mongo-nav/src/MongoNav.story.tsx
```diff
@@ -24,17 +24,18 @@ storiesOf('MongoNav', module).add('Default', () => (
         activeNav={select(
           'activeNav',
           Object.values(NavElement),
```
hswolff: Hrm....this is now a lie. `NavElement` contains way more than what you can actually set as an activeNav value.

We should probably make a separate object that just contains the values that we'll actually use here, like `ActiveNav` as the object and its values can be taken from `NavElement`.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -243,75 +246,95 @@ export default function OrgNav({
...
+            Access Manager
+          </OrgNavLink>
+
+          <IconButton
+            ariaLabel="Dropdown"
+            active={accessManagerOpen}
+            disabled={!current}
+            data-testid="org-nav-dropdown"
+            onClick={(e: React.MouseEvent) => {
+              onElementClick(NavElement.OrgNavDropdown, e);
```
hswolff: So here's an idea of a refactor for the value we get back from `useOnElementClick`.

What if we wrap the value given as a prop into a function that looks like this:

```jsx
const wrappedOnElementClick = (navElement: NavElement, additionalWork: (e) => void) {
   return (e) => {
        onElementClick(NavElement.OrgNavDropdown, e);
        return additionalWork(e);
  };
};

// So you could use it one of these two ways:
<MyComponent onClick={onElementClick(NavElement.OrgNavDropdown)} />
// or
<MyComponent onClick={onElementClick(NavElement.OrgNavDropdown, () => setStateThing())} />
```

.changeset/hip-gorillas-joke.md
brookescarlett: I know but wasn't sure if we wanted this to be major as our first official release, or if we were going to continue with minor updates. What do you think?

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: 👍 woo!

.changeset/hip-gorillas-joke.md
hswolff: I think we should hold off. I'm going to have a few more things to change from integrating this into cloud.

I think once we go live on cloud-dev we can bump this to 1.0.0.

brookescarlett: 👍

packages/mongo-nav/src/MongoNav.story.tsx
brookescarlett: 👍 Didn't construct the most beautifully but it does work

packages/mongo-nav/src/on-element-click-provider/OnElementClickProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export function useOnElementClick() {
+  return useContext(OnElementClickContext);
+}
+
+export default function OnElementClickProvider({
+  children,
+  onElementClick,
+}: OnElementClickProviderProps) {
+  const wrappedOnElementClick = (
```
hswolff: Need this memoized. Your usage of `useMemo` below requires the second argument: `useMemo(functionToMemo, dependencyArrayLikeUseEffect)`.

So for example `useMemo(wrappedOnElementClick, [onElementClick])`

packages/mongo-nav/src/types.ts
```diff
@@ -1,18 +1,50 @@
+const ActiveNavElement = {
+  OrgNavOrgSettings: 'orgNavOrgSettings',
```
hswolff: We should at least do 
```suggestion
  OrgNavOrgSettings: NavElement.OrgNavOrgSettings,
```
etc

packages/mongo-nav/src/types.ts
```diff
@@ -1,18 +1,50 @@
+const ActiveNavElement = {
```
hswolff: We should update all usages of `activeNav === XX` to use this object instead, as that's the object that dictates what in the nav can be set as active.

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -275,6 +218,9 @@ function MongoNav({
       orgSettings: `${hosts.cloud}/v2#/org/${data?.currentOrganization?.orgId}/settings/general`,
     },
     orgNav: {
+      leaf: data?.currentOrganization
+        ? `${hosts.cloud}/v2#/org/${data?.currentOrganization?.orgId}/`
```
```suggestion
        ? `${hosts.cloud}/v2#/org/${data.currentOrganization?.orgId}/`
```

Can drop that one as you checked it in the ternary. It'd be nice to not also have to check if `currentOrganization` exists, and do that in the first part of the ternary too.

packages/mongo-nav/src/user-menu/UserMenuTrigger.tsx
```diff
@@ -99,10 +101,13 @@ export default function UserMenuTrigger({
   children,
   ...rest
 }: UserMenuTriggerProps) {
+  const onElementClick = useOnElementClick();
   const buttonRef = useRef<HTMLButtonElement>(null);
   const { width: viewportWidth } = useViewportSize();
   const isTablet = viewportWidth < breakpoints.medium;
 
+  // console.log('open is', open, onElementClick);
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/src/on-element-click-provider/OnElementClickProvider.tsx
brookescarlett: 👍

packages/mongo-nav/src/types.ts
brookescarlett: 👍

packages/mongo-nav/src/types.ts
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: Isn't this checking that currentOrg exists in the first part of the ternary?

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -219,7 +220,7 @@ function MongoNav({
     },
     orgNav: {
       leaf: data?.currentOrganization
-        ? `${hosts.cloud}/v2#/org/${data?.currentOrganization?.orgId}/`
+        ? `${hosts.cloud}/v2#/org/${data.currentOrganization?.orgId}/`
```
```suggestion
        ? `${hosts.cloud}/v2#/org/${data.currentOrganization.orgId}/`
```

Should also be valid then?

packages/mongo-nav/src/on-element-click-provider/OnElementClickProvider.tsx
```diff
@@ -37,29 +37,12 @@ export default function OnElementClickProvider({
...
-  //     return (e: React.MouseEvent) => {
-  //       onElementClick(navElement, e);
-  //       if (additionalWork) {
-  //         return additionalWork(e);
-  //       }
-  //     };
-  //   },
-  //   [onElementClick],
-  // );
+  const memoizedWrapper = React.useMemo(() => wrappedOnElementClick, [
```
hswolff:
```suggestion
  const memoizedWrapper = React.useMemo(wrappedOnElementClick, [
```

`useMemo` is like `debounce`, it passes on whatever arguments are given to the function returned from memo.

packages/mongo-nav/src/MongoNav.tsx

packages/mongo-nav/src/on-element-click-provider/OnElementClickProvider.tsx
brookescarlett: 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -289,55 +295,53 @@ export default function ProjectNav({
...
           >
             <Icon glyph="Ellipsis" className={menuIconStyle} />
           </IconButton>
         }
       >
         <MenuItem
           href={projectNav.settings}
-          onClick={onMenuClick}
           data-testid="project-nav-settings"
+          onClick={onElementClick(ProjectNavProjectSettings, () =>
```
hswolff: FYI could have also done: 
```suggestion
          onClick={onElementClick(ProjectNavProjectSettings, onMenuClick)}
```

packages/mongo-nav/src/on-element-click-provider/OnElementClickProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+      onElementClick(navElement, e);
+      return additionalWork?.(e);
+    };
+  };
+
+  const callbackWrapper = useCallback(
+    (navElement, additionalWork) =>
+      wrappedOnElementClick(navElement, additionalWork),
+    [onElementClick],
+  );
```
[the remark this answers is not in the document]
brookescarlett: 👍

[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
brookescarlett: I can revert if you prefer, lmk

hswolff: Nah, up to you, just letting you know alternative ways to solve the problem.

PR #314, opened 2020-03-17
PD-575: Filter projects based on current org

brookescarlett:
## ✍️ Proposed changes

- Only display projects in ProjectSelect that have the same currentOrg as the currentProject

🎟 _Jira ticket:_ [PD-575](https://jira.mongodb.org/browse/PD-575)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/mongo-nav/src/MongoNav.tsx
+++ b/packages/mongo-nav/src/MongoNav.tsx
@@ -312,7 +312,7 @@ function MongoNav({
   const constructProjectURL = constructProjectURLProp ?? defaultProjectURL;
 
   const filteredProjects = data?.projects.filter(project => {
-    return project.orgId === data?.currentProject?.orgId;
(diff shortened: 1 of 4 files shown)


packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -311,6 +311,10 @@ function MongoNav({
     `${hosts.cloud}/v2/${projectId}#`;
   const constructProjectURL = constructProjectURLProp ?? defaultProjectURL;
 
+  const filteredProjects = data?.projects.filter(project => {
+    return project.orgId === data?.currentProject?.orgId;
```
```suggestion
    return project.orgId === data.currentProject?.orgId;
```
Should be fine as we already validated that data exists in the first line.

brookescarlett: you're right 🤦‍♀

PR #322, opened 2020-03-19
PD-581: Updated UserMenu state for non-cloud users

seanmhanson:
## ✍️ Proposed changes

This updates the UserMenu in Mongo-Nav to conditionally render Atlas as a submenu only if in an active Cloud/Realm/Charts product, since all of these require an AppUser and will always have the necessary metadata. When in other products, Atlas will render as a MenuItem linking to Cloud. This eliminates the need for checking if an MMS AppUser exists when used by University, Support, and in the Account app, and prevents any case where the links would be broken as a result.

🎟 _Jira ticket:_ [PD-581](https://jira.mongodb.org/browse/PD-581])

## 🛠 Types of changes

## ✅ Checklist

### State when Cloud/Realm/Charts are not active, on hovering on atlas
(note: the arrow animation is preserved)

Screen Shot 2020-03-19 at 12 20 48 PM


packages/menu/src/MenuItem.tsx
```diff
@@ -262,7 +262,7 @@ MenuItem.propTypes = {
   href: PropTypes.string,
   onClick: PropTypes.func,
   className: PropTypes.string,
-  description: PropTypes.string,
+  description: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
DesignerDave: I'm 80% sure that PropTypes.node includes string, so you might be able to omit it.

seanmhanson: Good catch, I am pretty sure about that too.

PR #323, opened 2020-03-19
PD-583: Add client-side filtering to ProjectSelect and OrgSelect

brookescarlett:
## ✍️ Proposed changes

- Add client-side filtering to ProjectSelect and OrgSelect

🎟 _Jira ticket:_ [PD-583](https://jira.mongodb.org/browse/PD-583)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -115,28 +115,61 @@ interface OrganizationMongoSelectProps extends BaseMongoSelectProps {
...
+
+  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setValue(e.target.value);
+
+    if (onChangeProp) {
+      onChangeProp(e);
+      return;
+    }
+
+    const search = new RegExp(`${e.target.value}`, 'i');
```
hswolff: If you're using the same value (`e.target.value`) twice in a scope it's generally good practice to cache that value in a local variable.

This is a micro performance optimization, as every property access takes some (negligible) cost to do, but most importantly it guarantees the consistency of that value used throughout the function.

Such that you sanitize the value into a local variable at the beginning of the function and then delegate all usage to that one source of (consistent) truth.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -115,28 +115,61 @@ interface OrganizationMongoSelectProps extends BaseMongoSelectProps {
   isOnPrem?: boolean;
 }
 
-const onKeyDown: React.KeyboardEventHandler = e => {
```
hswolff: You can  definitely update this function so you can reuse it across both selectors. Seems important to guarantee the similarity in behavior.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -115,28 +115,61 @@ interface OrganizationMongoSelectProps extends BaseMongoSelectProps {
...
 
+  useEffect(() => {
+    setRenderedData(data);
+  }, [data]);
+
+  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setValue(e.target.value);
+
+    if (onChangeProp) {
+      onChangeProp(e);
```
hswolff: So basically we only filter internally if the user doesn't set `onChangeProp`, right?

I think two things are possible to improve here:

- Abstract the filter logic into a standalone function that can be shared
- Not have to create an intermediate state value of `renderedData`. I think you can do this all inline in the component body.

Happy to pair on this.

packages/mongo-nav/src/mongo-select/MongoSelect.spec.tsx
```diff
@@ -175,6 +172,40 @@ describe('packages/mongo-select', () => {
         expect(orgSettingsIcon).toBeNull();
       });
     });
+
+    describe('when a user filters organizations', () => {
```
hswolff: Need a test for when a user sets their `onChange` handler to make sure we don't filter.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx

packages/mongo-nav/src/mongo-select/MongoSelect.tsx

packages/mongo-nav/src/mongo-select/MongoSelect.spec.tsx
brookescarlett: 👍

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -115,28 +132,48 @@ interface OrganizationMongoSelectProps extends BaseMongoSelectProps {
...
   loading = false,
 }: OrganizationMongoSelectProps) {
+  const [value, setValue] = useState('');
   const [open, setOpen] = useState(false);
   const onElementClick = useOnElementClick();
 
+  let renderedData = data;
+
+  const filterData = () => {
+    const search = new RegExp(`${value}`, 'i');
```
brookescarlett: Known bug if you typeL "\" it breaks, should I not turn the value into a regex?

packages/mongo-nav/src/mongo-select/MongoSelect.tsx

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -106,22 +106,21 @@ const paymentStatusMap: Record<Colors, ReadonlyArray<OrgPaymentLabel>> = {
   ],
 } as const;
 
-interface OrgNav {
+type OrgNavProps = Pick<
```
hswolff: Niiiicee!!

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
hswolff: Can you not just do `new RegExp(String(value), 'i')` ?

Do we have tests for these cases?

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -92,11 +92,28 @@ const formattedPlanTypes: Record<PlanType, string> = {
   [PlanType.OnPrem]: 'Ops Manager',
 } as const;
 
+const onKeyDown = (e: React.KeyboardEvent, callback: Function) => {
```
hswolff: The callback will always be the useState setter function, so we can just pass that in directly.

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
brookescarlett: Yes the simplification works but still fails on \

packages/mongo-nav/src/org-nav/OrgNav.tsx

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
brookescarlett: What do you mean?

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -92,11 +92,28 @@ const formattedPlanTypes: Record<PlanType, string> = {
   [PlanType.OnPrem]: 'Ops Manager',
 } as const;
 
+const onKeyDown = (e: React.KeyboardEvent, setValue: Function) => {
```
hswolff: Heh, misunderstood.

This should literally be the setValue function from useState.

Cuz then in the body you can just do `setValue(currentValue => `${currentValue} `)`. That usage is consistent everywhere.

brookescarlett: 🤦‍♀ got it

PR #325, opened 2020-03-20
PD-566: Refactor Icon, and allow titles to be unset

DesignerDave:
Allow null to unset Glyph titles

## ✍️ Proposed changes

This pull request does a few things that both improve on our ability to maintain the Icon component, as well as adds the ability to unset Icon titles.
- Eliminates the SVGR string literal template in favor of wrapping each glyph in a component, decoupling much of how Icon works from SVGR.
- Restructures the `SVGR` namespace, and adds a `LGGlyph` namespace to represent the wrapped glyph components.
- Adds handling for null values being passed into the `title` prop, unsetting title entirely if that's the case. This is what the ticket below pertained to, and was exceedingly difficult (if not impossible) to do with SVGR templates, thanks to the lack of documentation for the feature.

Out of an abundance of caution given the scope of changes to the component, this is being marked as a breaking change, though the intent is that this is really a minor change.

🎟 _Jira ticket:_ [Titles can't be unset on Icon](https://jira.mongodb.org/browse/PD-566)

## 🛠 Types of changes

## ✅ Checklist


scripts/jestSetup.js
```diff
@@ -1,3 +1,6 @@
+require('core-js/stable');
+require('regenerator-runtime/runtime');
```
hswolff: What's this for? I would think jest would handle this for us?

packages/icon/src/Icon.spec.tsx
```diff
@@ -5,6 +5,7 @@ import { toJson } from 'xml2json';
 import { render, cleanup } from '@testing-library/react';
```
hswolff: No test for when title is null?

scripts/jestSetup.js
DesignerDave: Whoops, you're right. I only need `core-js/stable` for `Object.entries`. I don't quite remember why I added this 😬

hswolff: May be good to add just the one polyfill directly so we don't add more than we need (and cause tests to run slower).

Can check out mms' polyfill.js file to see how we do it.

packages/icon/src/Icon.spec.tsx
DesignerDave: So since we can't actually test the SVGR output, we aren't able to test that the correct title is rendered.

I think there's a decent way we can still do this – I can move all the title logic to a self-contained function we can independently test. I'll re-request when that's done.

DesignerDave: Alright, let me know what you think of my solution for testing this.

packages/icon/src/createGlyphComponent.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+
+export function getGlyphTitle(name: string, title?: string | null) {
+  if (title === null) {
```
hswolff: This may be too late but this seems like weird implicit behavior. null and undefined are almost always used interchangeably. What if we had this behave if title was `false` instead?

PR #328, opened 2020-03-22
PD-588: Active States, URLs, and export OnPremUserMenu

seanmhanson:
## ✍️ Proposed changes

This PR includes four sets of small changes impacting the On Prem state and integration:
- the default URLs are updated to point to the on prem account app urls (`/account` rather than `/preferences`)
- active state handling is added to match the org and project dropdowns (marking both the menu item and the trigger as active)
- export the component, for usage outside of the mongo-nav (ex: for the minimal header used in the set of all full-screen pages in mms)
- export the ProjectSelect component, for usage in the on prem side navigation

🎟 _Jira ticket:_ [PD-588](https://jira.mongodb.org/browse/PD-588)

## 🛠 Types of changes

## ✅ Checklist

- [x] I have run `yarn changeset` and documented my change

### Example Active State
Screen Shot 2020-03-21 at 9 25 25 PM


packages/mongo-nav/src/user-menu/UserMenuTrigger.tsx
```diff
@@ -123,7 +125,7 @@ export default function UserMenuTrigger({
         {...buttonDataProp.prop}
         ref={buttonRef}
         className={cx(baseButtonStyles, {
-          [openBaseButtonStyle]: open,
+          [openBaseButtonStyle]: open || active,
```
DesignerDave: Hey Sean!

Everything else in this PR looks stellar, but I don't think we should make this specific change. It removes the change of state in the button when you open it on the associated pages, and I don't think the indicator is particularly useful on the button itself.

Would you mind taking this out? The PR looks ready to go otherwise!

PR #335, opened 2020-03-24
Create usePoller hook

hswolff:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


packages/hooks/src/usePoller.ts
```diff
@@ -0,0 +1,74 @@
...
+      savedCallback.current?.();
+    }
+
+    if (immediate) {
+      poll();
+    }
+
+    const id = setInterval(poll, interval);
+
+    return () => clearInterval(id);
```
jetpacmonkey: Do we want to have any special handling for async poll functions? The mms poller will avoid firing additional requests if previous ones are still pending.

DesignerDave: +1 – I like the idea of doing long-polling if the callback returns a promise.

packages/hooks/src/usePoller.ts
```diff
@@ -0,0 +1,74 @@
+import { useState, useRef, useEffect } from 'react';
+import useEventListener from './useEventListener';
+
+function useVisibilityChange() {
+  const isVisibilityStateVisible = () => document.visibilityState === 'visible';
```
DesignerDave: I dig this!

If we're doing this sort of fancy thing, what are your thoughts on slowing down polling on slower internet connections using the `NetworkInformation` API? (as an option for the hook, not a default of course)

hswolff: How much should we slow polling? I assume most consumers would set their own interval, so if we exposed a hook that returned the `NetworkInformation` data then they could set the interval depending on the network speed.

DesignerDave: Hmm... Yeah, I suppose exposing a hook for that might be a better path. I was thinking something like a `networkSpeedFactor` option that slows the polling by the passed multiple that could be less effort, though that would limit how granular a consumer could be.

jetpacmonkey: There's no guarantee that a poller is _necessarily_ making network requests, right? Probably best to allow the consumer to make that choice.

packages/hooks/src/usePoller.ts
hswolff: Adding it in.

packages/hooks/src/usePoller.ts
DesignerDave: Yeah, I 100% agree that if we were to even bake it into usePoller, it shouldn't be enabled by default.

hswolff:
One behavior I can't figure out what to about is...what we should do if the onPoll handler changes.

Right now we're just updating the useRef, such that the next interval will call with the new onPoll handler.

However is it fair that a person would assume that if the onPoll handler changes it should essentially 'restart' the poller and re-trigger it at the point the onPoll handler changes?

If we do that then we need to assume onPoll is memoized, which is something I don't think we can assume is true.

I'm inclined to keep the behavior as is for now. If someone needs to retrigger the onPoll themselves they can extract it to its own function and call it themselves. 

But...food for thought.

package.json
```diff
@@ -99,7 +99,7 @@
     "file-loader": "^4.2.0",
     "gh-pages": "^2.1.1",
     "husky": "^3.0.9",
-    "jest": "23.6.0",
+    "jest": "^25.2.0",
```
hswolff: I thought upgrading jest would help with the timing tests but it didn't. But also I upgraded jest and the failure consoles are soooo much better. So I kept it.

scripts/jestSetup.js
```diff
@@ -1,3 +1,8 @@
+// As of Jest 24 (https://jestjs.io/blog/2019/01/25/jest-24-refreshing-polished-typescript-friendly#breaking-changes)
+// this is no longer included by default for async tests
+// So we're including it so we can use async/await
+require('regenerator-runtime/runtime');
```
DesignerDave: Should we also be polyfilling this in our build? 🤔

packages/hooks/README.md
```diff
@@ -99,3 +99,28 @@ Hook to subscribe to changes in viewport size
 ```js
 const viewportSize = useViewportSize();
 ```
+
+## usePoller
+
+Hook to create a Poller that polls at a given interval.
+
+If your `onPoll` handler returns a `Promise` it will wait for the Promise to resolve or reject before scheduling the next interval.
```
DesignerDave: Is it worth noting the behavior when the window isn't active?

.changeset/silent-bananas-speak.md
```diff
@@ -0,0 +1,8 @@
+---
+'@leafygreen-ui/hooks': minor
+'@leafygreen-ui/text-input': patch
```
DesignerDave: We should probably have two changesets here given the changes are unrelated.

I believe this currently will generate a changelog for text input that includes the creation of `usePoller`.

scripts/jestSetup.js
hswolff: We're not using async/await in our code anymore. Probably should have some eslint rule to forbid it so we don't have to worry about it.

If we do transpile it down though I do think it should be an assumption of the application to provide their own, and we just make note of that.

packages/hooks/README.md
hswolff: Yeah wasn't sure where to put that. Here works.

.changeset/silent-bananas-speak.md
hswolff: Oooh that's how it works? Interesting.

scripts/jestSetup.js
DesignerDave: Ahh wait, this is just for writing async functions in tests. Disregard 😅

PR #336, opened 2020-03-25
Set Atlas tab link to the full URL

hswolff:
This prevents full page refreshes in an Atlas project.


brookescarlett:
Approving, but add changeset before you merge please 😈

PR #338, opened 2020-03-26
Add polling for alerts to MongoNav

hswolff:
## ✍️ Proposed changes

This also fixes mongo nav tests to so the warning about changes not being wrapped in `act` are fixed. The issue was we had a Promise in a `useEffect` that changed state and as such we needed to wrap it in an `act` call.

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -62,17 +62,21 @@ describe('packages/mongo-nav', () => {
...
-        <MongoNav
-          activeProduct="cloud"
-          onOrganizationChange={onOrganizationChange}
-          onProjectChange={onProjectChange}
-          {...props}
-        />,
-      ),
-    );
+  const renderComponent = async (props = {}) => {
+    await act(async () => {
```
brookescarlett: !! You fixed the testing console errors 🙏

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -167,7 +167,6 @@ function OrgSelect({
 
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const val = e.target.value.replace(/\\/g, '\\');
-    console.log(val);
```
brookescarlett: Thank you! I caught this in my open PR as well 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -223,7 +224,7 @@ const secondTabName = displayProductName();
 
 type ProjectNavProps = Pick<
   MongoNavInterface,
-  'activeProduct' | 'activeNav' | 'onProjectChange'
+  'activeProduct' | 'activeNav' | 'onProjectChange' | 'mode'
```
brookescarlett: Woohooo this makes me happy

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -245,9 +246,11 @@ export default function ProjectNav({
   activeNav,
   onProjectChange,
   hosts,
-  alerts = 0,
+  mode = Mode.Production,
```
DesignerDave: Hmm... Can't we require this to be passed into ProjectNav, and set the default only in MongoNav?

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -256,6 +259,32 @@ export default function ProjectNav({
   const isCloudManager = current?.planType === PlanType.Cloud;
   const isLoading = !!current;
 
+  const currentProjectId = current?.projectId;
+
+  function fetchAlertsCount() {
+    return fetch(
+      `${hosts.cloud}/user/shared/alerts/project/${currentProjectId}`,
```
DesignerDave: Given that `currentProjectId` is possibly undefined (based on the optional chaining operator in its declaration), we might want to check that it's defined before fetching.

packages/mongo-nav/src/project-nav/ProjectNav.spec.tsx
```diff
@@ -66,29 +67,48 @@ describe('packages/mongo-nav/src/project-nav', () => {
...
-          urls={urlDefaults}
-          activeProduct="cloud"
-          hosts={hostDefaults}
-          onProjectChange={onProjectChange}
-          {...props}
-        />,
-      ),
-    );
+  const renderComponent = async (props = {}) => {
+    await act(async () => {
```
DesignerDave: Not sure, but is there a reason we need the callback to act to be an async function? I'm not sure what we get for doing that.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
hswolff: Sure, same difference. Not sure how many assumptions we should make about props. This one seems fine though.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
hswolff: I tricked you! Look at the enabled value for usePoller! Won't run unless it's set.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
DesignerDave: Fair enough – I think for ProjectNav in particular, we can make a lot of assumptions given it's not built to be used outside of MongoNav.

packages/mongo-nav/src/project-nav/ProjectNav.spec.tsx
hswolff: Without the inner async we get this error:
image

Without the outer await we get this:

image


It's because we're waiting for a Promise to resolve inside the component.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
DesignerDave: Ahah! Missed that. Looks good 😁

packages/mongo-nav/src/project-nav/ProjectNav.spec.tsx
DesignerDave: Wow that's unintuitive. Okay 😬lol

PR #342, opened 2020-03-27
PD-594: Tweaks to active navs for User Menus, docs

seanmhanson:
## ✍️ Proposed changes

This is a catch-all for some things I ran into when starting in on the Ops Manager integration:
- [x] previously we disabled the org nav when passing in a placeholder active nav element `UserSettings`. This removes that and replaces it with the correct logic, and adds two additional values, `UserMenuCloudOther` and `UserMenuOnPremOther`, to indicate the org nav is disabled when on a page that doesn't contain an active link in `MongoNav`
- [x] the User Feedback related active nav element I overzealously added last PR has been removed since they it is an external link
- [x] the submenu items for Atlas now are active when on the appropriate page as a Cloud user
- [x] the documentation has been updated to reflect the full list of possible active nav element values

🎟 _Jira ticket:_ [PD-594](https://jira.mongodb.org/browse/PD-594)

## 🛠 Types of changes

## ✅ Checklist

## 🖼 Screenshot
Screen Shot 2020-03-26 at 5 00 13 PM


packages/mongo-nav/README.md
```diff
@@ -45,6 +45,45 @@
 
 _Any other properties will be spread on the root element_
 
+### ActiveNav Values
```
brookescarlett: Thank you 👍

packages/mongo-nav/src/helpers/OnPremUserMenu.tsx
```diff
@@ -12,14 +12,60 @@ const onPremMenuWrapper = css`
 `;
 
 interface OnPremUserMenuProps {
+  /**
```
brookescarlett: Also thank you for this

PR #344, opened 2020-03-27
More tweaks

hswolff:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/mongo-nav/src/data.ts
+++ b/packages/mongo-nav/src/data.ts
@@ -98,6 +98,7 @@ export const urlDefaults: Required<URLSInterface> = {
     alerts: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#alerts`,
     activityFeed: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#activity`,
     invite: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#access/add`,
+    realm: `https://stitch.mongodb.com/groups/${dataFixtures.currentProject?.projectId}`,
     charts: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#charts`,
   },
   onPrem: {
--- a/packages/mongo-nav/src/data.ts
+++ b/packages/mongo-nav/src/data.ts
@@ -98,7 +98,7 @@ export const urlDefaults: Required<URLSInterface> = {
     alerts: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#alerts`,
     activityFeed: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#activity`,
     invite: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#access/add`,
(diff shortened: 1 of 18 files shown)


brookescarlett:
I might be going crazy but I think the animation on the close of MongoSelect components closed is now off. It's almost like it's closing then opening and then closing again really quickly. Let me know if you see this too @DesignerDave. The logic otherwise LGTM

packages/mongo-nav/src/data.ts
```diff
@@ -98,6 +98,7 @@ export const urlDefaults: Required<URLSInterface> = {
     alerts: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#alerts`,
     activityFeed: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#activity`,
     invite: `https://cloud.mongodb.com/v2/${dataFixtures.currentProject?.projectId}#access/add`,
+    realm: `https://stitch.mongodb.com/groups/${dataFixtures.currentProject?.projectId}`,
```
mburdette: i would just update this to `/apps` at the end to avoid redirect on our end.

packages/mongo-nav/src/mongo-select/MongoSelect.spec.tsx
```diff
@@ -340,6 +372,31 @@ describe('packages/mongo-select', () => {
...
+          );
+        });
+
+        // Need to wait for the UI to update as clicking on the orgTrigger so rapidly causes
+        // React to coalesce the events
+        await waitFor(() => {
+          setExpectedElements();
+          expect(expectedElements!.projectInput).not.toBeInTheDocument();
+        });
+        // queries.debug();
```
DesignerDave: This doesn't look intentional 🤔

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
```diff
@@ -174,11 +180,15 @@ function OrgSelect({
     setValue(val);
 
     if (onChangeProp) {
-      onChangeProp(value, e);
-      return;
+      return onChangeProp(value, e);
```
DesignerDave: Hmm... This looks like we may not want to return anything since it's a change handler. Any reason we shouldn't just remove the return statement entirely, especially since we only return conditionally here AFAICT?

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -225,7 +225,7 @@ const tooltipStyles = css`
 `;
 
 export function displayProductName(today = new Date(Date.now())) {
-  const mdbLiveDate = new Date('May 4, 2020 0:00:00');
+  const mdbLiveDate = new Date('June 8, 2020 0:00:00');
```
DesignerDave: 😷

packages/mongo-nav/src/mongo-select/MongoSelect.tsx
hswolff: You can return false from change handlers to effectively do e.preventDefault and e.stopPropagation.

DesignerDave: 🤦‍♂ You're right – I completely forgot that's a thing you can do

PR #346, opened 2020-03-30
PD-597: Updates to OrgSelect, ProjectSelect, and OnPremUserMenu

seanmhanson:
## ✍️ Proposed changes

This makes several changes involving the `mongo-select` components used in `mongo-nav`. First, regarding organization:
- [x] `MongoSelect` is broken out into `ProjectSelect` and `OrgSelect`, ibid for tests
- [x] `Trigger` is broken out into `ProjectTrigger` and `OrgTrigger`
- [x] common styles and types have been pulled into `styles` and `types` local to the directory
- [x] a helper function is pulled out of the components

The following changes in behavior/design have been made:
- [x] the select components no longer require the full `URLSInterface` and instead only require the values present in` urls.mongoSelect`. This allows external use of `ProjectSelect` without declaring the full urls defaults for `mongo-nav`. I think this is actually a good practice overall.
- [x] the active project should be indicated with an active state in the select
- [x] the active organization should be indicated with an active state in the select
- [x] either `className` should be set on the base select props and then applied, or rest params need to be exposed, in order to allow external usage of the selects to override positioning

Additionally, one change was requested for Ops Manager's user menu, as just approved by the Ops Manager team:
- [x] add Public API Access as an option to the Ops Manager user dropdown

🎟 _Jira ticket:_ [PD-597](https://jira.mongodb.org/browse/[PD-597])

## 🛠 Types of changes

## ✅ Checklist

### Updated structure for mongo-select
Screen Shot 2020-03-30 at 1 38 43 PM

### Updated active state for Project
Screen Shot 2020-03-30 at 1 38 23 PM

### Updated active state for Organization
Screen Shot 2020-03-30 at 1 38 30 PM

### Updated On Prem User Menu options
Screen Shot 2020-03-30 at 3 36 51 PM


packages/mongo-nav/README.md
```diff
@@ -79,8 +79,9 @@ _Any other properties will be spread on the root element_
 `UserMenuOnPremInvitations`  
 `UserMenuOnPremOrganizations`  
 `UserMenuOnPremOther`
-`UserMenuOnPremPersonalization`  
+`UserMenuOnPremPersonalization`
```
seanmhanson: That whitespace needs to be readded or the markdown won't insert a line break

packages/mongo-nav/src/data.ts
```diff
@@ -26,7 +26,7 @@ export const dataFixtures = {
     orgId: 'fakeOrgId1',
     planType: 'atlas',
     projectId: 'fakeProjectId1',
-    projectName: 'Test Project',
+    projectName: 'Demo Project 1',
```
seanmhanson: [background] This data fixture had to be updated because it didn't match any project in the `currentProjects` data fixtures.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -0,0 +1,225 @@
...
+          {!isOnPrem && (
+            <span className={productStyle}>{formattedPlanTypes[planType]}</span>
+          )}
+        </div>
+      </MenuItem>
+    );
+  };
+
+  return (
+    <OrgTrigger
```
brookescarlett: This is a nit, and I'm happy either way but now that we consolidating should we just inline the Trigger functions directly into the component, rather than breaking them out as their own component? Either way the readability has gone up 1000% so very grateful for this PR.

packages/mongo-nav/src/mongo-select/OrgTrigger.tsx
```diff
@@ -194,10 +121,15 @@ export function OrganizationTrigger({
...
-            [textLoadingStyle]: loading,
-          })}
+          className={cx(
+            baseButtonStyles,
+            orgButtonStyles,
+            {
+              [activeButtonColor]: open,
+              [textLoadingStyle]: loading,
+            },
+            className,
```
brookescarlett: (Re above, this would be mitigated if we collapse into one file, but still not at all necessary)

packages/mongo-nav/src/mongo-select/OrgTrigger.tsx
```diff
@@ -237,7 +169,7 @@ export function OrganizationTrigger({
       {!disabled && (
         <a
           {...anchorDataProp.prop}
-          href={urls.mongoSelect.orgSettings}
+          href={urls.orgSettings}
```
brookescarlett: 🙏

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
seanmhanson: Great idea.

seanmhanson:
Heads up: right now 175 tests fail because I have not yet updated the usage or testing of the org/project selects after pulling in the trigger components and consolidating code. I'll take care of that either later tonight or tomorrow morning

packages/mongo-nav/src/mongo-select/README.md
```diff
@@ -0,0 +1,60 @@
+# Project Select
```
brookescarlett: Should we add information about OrgSelect too, if we're putting this here?

seanmhanson: I omitted this because we are not exporting OrgSelect, and only using it internally.

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -0,0 +1,250 @@
...
+        active={isActive}
+        className={cx(menuItemContainerStyle, nameStyle)}
+        href={constructProjectURL(datum)}
+      >
+        {projectName} {isActive && '(current)'}
+      </MenuItem>
+    );
+  };
+
+  const onClickTrigger = onElementClick(
```
hswolff: Can we inline this please?

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -0,0 +1,363 @@
...
+  const [open, setOpen] = useState(false);
+  const onElementClick = useOnElementClick();
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+  const { width: viewportWidth } = useViewportSize();
+  const isTablet = viewportWidth < breakpoints.medium;
+
+  const toggleOpen = () => {
+    setOpen(!open);
+    if (!open) {
+      setValue('');
```
hswolff: Why the change?

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -0,0 +1,363 @@
...
+}: OrganizationMongoSelectProps) {
+  const [value, setValue] = useState('');
+  const [open, setOpen] = useState(false);
+  const onElementClick = useOnElementClick();
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+  const { width: viewportWidth } = useViewportSize();
+  const isTablet = viewportWidth < breakpoints.medium;
+
+  const toggleOpen = () => {
+    setOpen(!open);
```
hswolff: Toggles typically should use the updater as function to ensure you're always getting the up to date value.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
seanmhanson: This previously used `useEffect` and split handling the open and value states over the trigger and the select. Now that they're integrated, this would update its own state instead, and was causing an infinite loop of renders.

PR #352, opened 2020-04-01
PD-567: IconButton accessibility, TS fixes

DesignerDave:
## ✍️ Proposed changes

- Adds a helper type `Either<>` that constructs a type that requires one of a set of optional keys in an interface.
- Changes `ariaLabel` to the HTML `aria-label`, and requires that either `aria-label` or `aria-labelledby` is present on the IconButton component.
- If `size` isn't explicitly set on an Icon within IconButton, IconButton will now set the icon's size according to its own size prop.
- Some general cleanup in Icon and IconButton.
- Resolves an issue where TypeScript would add a triple-slash reference directive to built `.d.ts` files in Icon and Portal pointing to types in a Storybook package. This was a problem in TypeScript apps that don't already use Storybook.

🎟 _Jira ticket:_ [Apply accessibility attributes correctly to Icons and Icon Buttons
](https://jira.mongodb.org/browse/PD-567)

## 🛠 Types of changes

## ✅ Checklist


package.json
```diff
@@ -35,7 +35,7 @@
     "prettier:check": "prettier --check \"**/*.{js,ts,tsx,json,md}\"",
     "test": "jest --env jsdom",
     "ts": "tsc --build tsconfig.json",
-    "prepublishOnly": "yarn lint && yarn test && yarn build",
+    "prepublishOnly": "yarn lint && yarn test && yarn build && yarn ts",
```
DesignerDave, line 38: Previously, only new JS files would be built as a part of the pre-publish script, leaving potentially (though unlikely) out of date TS files in the `dist` directory of packages being published.

packages/icon-button/src/IconButton.tsx
```diff
@@ -205,27 +143,102 @@ const activeStyle: { readonly [K in Variant]: string } = {
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
 `;
 
+interface IconProps extends React.SVGProps<SVGSVGElement> {
+  glyph: string;
+  size?: Size | number;
+  title?: string | null | boolean;
+}
```
DesignerDave, line 163: Since applications can't yet tree-shake, we're duplicating this interface from the types in the namespaces within the Icon package rather than importing the Icon package.

packages/icon/src/types.ts
```diff
@@ -0,0 +1,15 @@
+export namespace SVGR {
```
DesignerDave, line 1: Rather than declaring these namespaces in a definition file like we were previously, I moved them here so we can import the namespace directly, and not include a triple-slash reference directive in `index.ts`.

DesignerDave:
> **Size Change:** +571 B (0%)
> 
> **Total Size:** 530 kB
> 
> Filename	Size	Change	
> `packages/code/dist/index.node.js`	7.74 kB	+1 B	
> `packages/code/dist/index.web.js`	7.79 kB	+1 B	
> `packages/icon-button/dist/index.node.js`	2.58 kB	+306 B (11%)	⚠️
> `packages/icon-button/dist/index.web.js`	2.59 kB	+305 B (11%)	⚠️
> `packages/icon/dist/index.node.js`	8.29 kB	-10 B (0%)	
> `packages/icon/dist/index.web.js`	8.29 kB	-9 B (0%)	
> `packages/lib/dist/index.node.js`	1.25 kB	+1 B	
> `packages/lib/dist/index.web.js`	1.25 kB	+1 B	
> `packages/menu/dist/index.node.js`	6.77 kB	-4 B (0%)	
> `packages/menu/dist/index.web.js`	6.78 kB	-4 B (0%)	
> `packages/mongo-nav/dist/index.node.js`	31.8 kB	-8 B (0%)	
> `packages/mongo-nav/dist/index.web.js`	31.9 kB	-9 B (0%)	
> ℹ️ **View Unchanged**
> [compressed-size-action](https://github.com/preactjs/compressed-size-action)

Lol at the warning for +305 Bytes

packages/lib/src/index.ts
```diff
@@ -34,7 +13,43 @@ export type HTMLElementProps<
...
+ *   requiredOtherTimes: boolean,
+ * }
+ *
+ * type EquivalentToExampleEither = FirstIsRequired | SecondIsRequired
+ * ```
+ */
+export type Either<T, Keys extends keyof T = keyof T> = Omit<T, Keys> &
+  {
+    [K in Keys]-?: Required<Pick<T, K>> &
+      Partial<Record<Exclude<Keys, K>, undefined>>;
```
jetpacmonkey, line 51: Shouldn't this be `Partial<Pick<T, Exclude<Keys, K>>>`? At least, I'm pretty sure that would match your example. As it is, if I'm reading it right, this would be equivalent to

```ts
interface FirstIsRequired extends SharedInExampleInterface {
  sometimesRequired: boolean;
  requiredOtherTimes?: undefined;
}
```

with `undefined` instead of `boolean`. Of course, if the goal is to disallow people from providing _both_ keys, then this is right and the example should be updated.

packages/icon-button/README.md
```diff
@@ -57,8 +56,11 @@ import IconButton from '@leafygreen-ui/icon-button';
 | `href`      | `string`                     | If a href is supplied, the component renders inside of an `a` tag instead of inside of a `button` tag.            |           |
 | `className` | `string`                     | Adds a className to the class attribute on the container element.                                                 |           |
 | `children`  | `node`                       | Content rendered inside of the `<IconButton />` component                                                         |           |
-| `ariaLabel` | **Required** `string`        | Value passed to `aria-label` attribute to ensure component's accessibility                                        |           |
 | `size`      | `default`, `large`, `xlarge` | Determines the size of the IconButton                                                                             | `default` |
 | `active`    | `boolean`                    | Determines whether the `<IconButton />` will appear `active`                                                      | `false`   |
 
+### Special Case: Aria Labels
```
brookescarlett, line 62: We have required props across multiple components, why break this out into its own section here?

DesignerDave: I broke it out because in the table, it's particularly difficult to indicate that having one of the two is required, but they're both optional. It's a little strange compared to our usual prop typing.

packages/lib/src/index.ts
DesignerDave: Hmm... I think you're right here – good catch! I tested, and confirmed this change works. As is, it was requiring that only one of the two was passed when really having both is okay.

DesignerDave: Just pushed the fix 😁

packages/lib/src/index.ts
```diff
@@ -34,7 +13,42 @@ export type HTMLElementProps<
...
+ *   sometimesRequired?: boolean,
+ *   requiredOtherTimes: boolean,
+ * }
+ *
+ * type EquivalentToExampleEither = FirstIsRequired | SecondIsRequired
+ * ```
+ */
+export type Either<T, Keys extends keyof T = keyof T> = Omit<T, Keys> &
+  {
+    [K in Keys]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<Keys, K>>>;
```
jetpacmonkey, line 50: So just so I understand... the `-?:` here is removing the optionality from `keyof T` that's being "looped" through here, not from any of the keys that are actually output, right? Essentially just avoiding having an additional `| undefined` in the generated type.

DesignerDave: That's my understanding. In all fairness, this is a modified version of an answer in this SO thread: https://stackoverflow.com/questions/40510611/typescript-interface-require-one-of-two-properties-to-exist

I didn't know what `-?` did in TS until I found this hah

packages/icon-button/README.md
brookescarlett: Kk that makes sense

package.json
hswolff: Should we swap the order of this then? Cuz yarn lint already runs yarn ts.

packages/icon-button/src/IconButton.spec.tsx
```diff
@@ -41,14 +36,44 @@ describe('packages/icon-button', () => {
...
   });
 
+  const renderedIcon = getByTestId('icon-test-id');
+
+  test('renders icon as button content', () => {
+    expect(iconButton.contains(renderedIcon)).toBe(true);
+  });
+
+  test("the rendered icon doesn't include a title tag", () => {
+    expect(renderedIcon.getElementsByTagName('title')[0]).toBeUndefined();
```
hswolff, line 46: `.length === 0` ?

packages/icon-button/src/IconButton.tsx
hswolff: Great comment! May want to inline this comment to preserve it in the code base.

packages/icon-button/src/IconButton.tsx
```diff
@@ -205,27 +143,102 @@ const activeStyle: { readonly [K in Variant]: string } = {
...
+
+  /**
+   * Content to appear inside of `IconButton`.
+   */
+  children?: React.ReactNode;
+
+  /**
+   * Determines whether or not `IconButton` is disabled.
+   */
+  disabled?: boolean;
```
hswolff, line 184: I gotta say the more I see JSDoc annotations used with interfaces and how it renders in VSCode the happier I am.

That being said it'd be great if we started using the [@default](https://jsdoc.app/tags-default.html) tag so we can have the annotations show the default value. Check out usePoller's `PollerOptions` to see an example.

packages/icon-button/src/IconButton.tsx
```diff
@@ -234,31 +247,60 @@ const getIconStyle = (size: Size) => css`
...
+
+    return child;
+  });
+
+  const renderIconButton = (Root: React.ElementType = 'button') => (
     <Root
       {...rest}
-      href={href ? href : undefined}
+      ref={ref}
+      href={href ?? undefined}
```
hswolff, line 299: Is this just so it's not null?

packages/icon/src/types.ts
hswolff: Probably better in general to do this!

packages/lib/src/index.ts
```diff
@@ -1,28 +1,7 @@
...
-  const prefix = 'data-leafygreen-ui' as string;
-
-  return {
-    prop: {
-      [prefix]: name,
-    },
-    selector: `[${prefix}="${name}"]`,
-  };
-}
+export { typeIs, testHelpers };
```
hswolff, line 4: Can do new export syntax now!

packages/icon-button/src/IconButton.tsx
```diff
@@ -234,31 +247,60 @@ const getIconStyle = (size: Size) => css`
...
-    ariaLabel,
+    'aria-label': ariaLabel,
+    'aria-labelledby': ariaLabelledBy,
     ...rest
   } = props;
 
-  const renderIconButton = (Root: React.ElementType<any> = 'button') => (
+  // We do our own proptype validation here to ensure either 'aria-label' or 'aria-labelledby' are passed to the component.
+  if (!ariaLabel && !ariaLabelledBy) {
+    console.error(
```
hswolff, line 270: A ticket for us to do in the future is to make our own leafygreen.error function that does what React does and doesn't show these warnings in production.

packages/icon-button/src/IconButton.tsx
DesignerDave: Good point lol

I updated this without really thinking why we don't just pass href in directly.

packages/icon-button/src/IconButton.tsx
DesignerDave: True. Right now, we're super reliant on consuming apps stripping out console messages in their build.

https://jira.mongodb.org/browse/PD-600

packages/lib/src/index.ts
DesignerDave: Is there a way to neatly do this: `typeIs` and  `testHelpers` are imported like so:

```js
import * as typeIs from './typeIs';
```

AFAIK we can't do:
```js
export * as typeIs from './typeIs';
```

packages/icon-button/src/IconButton.spec.tsx
DesignerDave: You're right, that's probably better.

package.json
DesignerDave: Whoops, seemed to have forgotten about that. Should we actually have lint run tsc without emitting declaration files? It seems like a very side effect-y thing to be doing for a linting script.

packages/icon-button/src/IconButton.tsx
DesignerDave: Which comment in particular are you talking about? There's no comment where you left this.

packages/icon-button/src/IconButton.tsx
DesignerDave: AFAICT TSDoc doesn't actually support default annotations. I remember exploring this when we first started doing them, and it appears to remain true.

https://github.com/Microsoft/tsdoc/issues/151

packages/icon/src/types.ts
DesignerDave: That's the hope 😁

packages/icon-button/src/IconButton.tsx
hswolff: Oh, what you wrote on GitHub:

> Since applications can't yet tree-shake, we're duplicating this interface from the types in the namespaces within the Icon package rather than importing the Icon package.

packages/icon-button/src/IconButton.tsx
hswolff: Hmm interesting. We'd still show it in VSCode as it shows any JSDoc comments there.

packages/icon-button/src/IconButton.tsx
```diff
@@ -276,16 +276,17 @@ const IconButton = React.forwardRef((props: IconButtonProps, ref) => {
     if (isComponentType(child, 'Icon')) {
       const { size: childSize, title }: IconProps = child.props;
 
-      if (typeof title === 'string' && title.length > 0) {
-        return child;
-      }
+      const newChildProps: Partial<IconProps> = {
+        size: childSize || size,
```
hswolff, line 280:
```suggestion
        size: childSize ?? size,
```

Instead?

packages/lib/src/index.ts
hswolff: I believe you can 😱  https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export

package.json
hswolff: If we can do that easily then that WFM but it may be tricky to actually pull off.

Might be easier to just remove it from lint entirely and have it be its own individual task that we call individually?

packages/icon-button/src/IconButton.tsx
DesignerDave: 🤦‍♂ Right, will do hah

packages/icon-button/src/IconButton.tsx
DesignerDave: Right – we've done that for a few components before, so I suppose I can add em here.

packages/icon-button/src/IconButton.tsx
DesignerDave: Hmm... I'm kinda okay here if `''` or `0` cause it to default to the IconButton's size, but I'm open to changing that if you feel strongly.

packages/lib/src/index.ts
DesignerDave: Oh hmm... it was yelling at me previously for doing this but it seems to be working now! Will do.

DesignerDave: Actually, the current version of TS is still yelling. 
```sh
MongoDB/leafygreen-ui [PD-567-icon-button-accessibility●] » yarn ts
yarn run v1.16.0
$ tsc --build tsconfig.json

packages/lib/src/index.ts:1:10 - error TS1005: 'from' expected.
1 export * as typeIs from './typeIs';

packages/lib/src/index.ts:1:13 - error TS1005: ';' expected.
1 export * as typeIs from './typeIs';

packages/lib/src/index.ts:1:20 - error TS1005: ';' expected.
1 export * as typeIs from './typeIs';

packages/lib/src/index.ts:1:25 - error TS1005: ';' expected.
1 export * as typeIs from './typeIs';
```
Leaving for now and we can try again with the next version of TS if that's alright.

hswolff: WFM

packages/icon-button/src/IconButton.tsx
hswolff: Ah...wfm.

hswolff:
Did you forget to push? I'm not seeing changes you said you made?

package.json
DesignerDave: I tried just passing `--noEmit`, but that didn't work out the box. I wanna keep this on our radar as an improvement we could make, because I feel pretty strongly relying on a lint script to build part of our packages is a bad idea. I'll make a ticket.

DesignerDave: https://jira.mongodb.org/browse/PD-603

DesignerDave:
Sorry Harry, I've been going through the comments as I address them locally. I'll re-request when I push.

DesignerDave:
AKA now

PR #353, opened 2020-04-01
PD-598: Adds cloud manager logo

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-598](https://jira.mongodb.org/browse/[PD-598])

## 🛠 Types of changes

Screen Shot 2020-04-01 at 4 55 58 PM
Screen Shot 2020-04-01 at 4 56 07 PM

### DIFF
--- a/packages/mongo-nav/src/helpers/Icons.tsx
+++ b/packages/mongo-nav/src/helpers/Icons.tsx
@@ -6,6 +6,171 @@ type ProductIconProps = HTMLElementProps<'svg'> & {
   active: boolean;
 };
 
+export function CloudManagerSmall() {
+  return (
+    <svg width="10px" height="10px" viewBox="0 0 10 10">
+      <path
+        className={css`
+          fill: #00804b;
+        `}
+        d="M3.19,9.66h.4A1.32,1.32,0,0,0,3.59,7H.43A5,5,0,0,0,3.19,9.66Z"
+      />
+      <path
+        className={css`
+          fill: #03aa4f;
+        `}
+        d="M1.58,5a2,2,0,0,1,.5-1.32H.18a4.93,4.93,0,0,0,0,2.64H2.07A2,2,0,0,1,1.58,5Z"
+      />
+      <path
+        className={css`
+          fill: #00804b;
+        `}
+        d="M5.61,8.34A2,2,0,0,1,4.75,10H5A5,5,0,0,0,9.57,7H5.12A2,2,0,0,1,5.61,8.34Z"
+      />
+      <path
+        className={css`
+          fill: #007dab;
+        `}
+        d="M5,0A5,5,0,0,0,.43,3H9.57A5,5,0,0,0,5,0Z"
+      />
+      <path
+        className={css`
+          fill: #03aa4f;
+        `}
+        d="M2.28,5A1.32,1.32,0,0,0,3.6,6.32H9.82a4.93,4.93,0,0,0,0-2.64H3.6A1.32,1.32,0,0,0,2.28,5Z"
+      />
+    </svg>
+  );
+}
+
+export function AtlasSmall() {
+  return (
+    <svg width="10px" height="10px" viewBox="0 0 10 10">
+      <path
+        className={css`
+          fill: #03aa4f;
+        `}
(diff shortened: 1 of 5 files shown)


packages/mongo-nav/src/helpers/Icons.tsx
```diff
@@ -6,6 +6,171 @@ type ProductIconProps = HTMLElementProps<'svg'> & {
   active: boolean;
 };
 
+export function CloudManagerSmall() {
+  return (
+    <svg width="10px" height="10px" viewBox="0 0 10 10">
+      <path
+        className={css`
+          fill: #00804b;
```
hswolff, line 14: Can these just be direct attributes and remove emotion from the equation?

brookescarlett: 👍

packages/mongo-nav/src/helpers/Icons.tsx
```diff
@@ -6,6 +6,146 @@ type ProductIconProps = HTMLElementProps<'svg'> & {
   active: boolean;
 };
 
+export function CloudManagerSmall() {
```
DesignerDave, line 9: Is there a reason we need this in addition to the regular CM logo component? We should just be able to set the flat colors for the fill instead of the gradients as a part of a variant.

DesignerDave: Ditto for Atlas. It might be nice to determine whether to display the flat colors vs gradients based on a size prop actually!

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -228,7 +238,16 @@ function OrgSelect({
           </span>
 
           {!isOnPrem && (
-            <span className={productStyle}>{formattedPlanTypes[planType]}</span>
+            <span className={displayFlex}>
```
DesignerDave, line 241: Let's use a div since we're manually setting the display property anyways.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -404,7 +405,13 @@ export default function ProjectNav({
               isCloudManager ? 'cloud-manager' : 'atlas'
             }`}
           >
-            {!isMobile && (
+            {!isMobile && isCloudManager ? (
+              <CloudManagerIcon
+                active={activeProduct === Product.Cloud && isLoading}
+                {...productIconProp.prop}
```
DesignerDave, line 411: Nit: Can we put this splat first?

packages/mongo-nav/src/helpers/Icons.tsx
brookescarlett: 👍

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
brookescarlett: 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
brookescarlett: 👍

packages/mongo-nav/src/helpers/Icons.tsx
```diff
@@ -1,20 +1,149 @@
...
+          x2="-3.78"
+          y2="9"
+          xlinkHref="#cloud-manager-linear-gradient"
+        />
+      </defs>
+      <path
+        className={css`
+          fill: ${variant === 'small'
+            ? '#00804b'
+            : active
```
DesignerDave, line 75: Hmm... I'd think we'd also want to render the logo as gray when it's small and not active, right?

packages/mongo-nav/src/helpers/Icons.tsx
```diff
@@ -1,20 +1,149 @@
...
 };
 
-export function AtlasIcon({ active, ...rest }: ProductIconProps) {
+export function CloudManagerIcon({
+  variant = 'default',
+  active = false,
+  className,
+  ...rest
+}: ProductIconProps) {
+  const size = variant === 'small' ? 10 : 18;
```
DesignerDave, line 16: For future flexibility, we may want to render the small version based on the size, not the other way around.

packages/mongo-nav/src/helpers/Icons.tsx
brookescarlett: Nope, per the spec there's always a color: https://mongodb.invisionapp.com/d/main/#/console/19026513/396615702/preview

DesignerDave: In this case, that's true. We may want to be able to show a monochromatic version of the smaller logo in future cases though.

PR #358, opened 2020-04-06
PD-606: Fix onOrgChange and onProjectChange functions

brookescarlett:
## ✍️ Proposed changes

- Fixes onChange and onProjectChange functions, such that they now accept an object with value, setData and event as keys, allowing consuming applications to conditionally render organizations/projects based on a user's search

🎟 _Jira ticket:_ [PD-606](https://jira.mongodb.org/browse/PD-606)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -40,17 +40,13 @@ describe('packages/mongo-nav', () => {
     expectedElements.charts = queryByTestId('project-nav-charts');
   };
 
-  let onOrganizationChange: jest.Mock;
```
hswolff, line 43: We should have coverage that this API usage works as expected. Where a consumer can call `setValue` and it works.

brookescarlett: 👍

PR #359, opened 2020-04-06
PD-607: Conditionally render logo in orgselect

brookescarlett:
## ✍️ Proposed changes

Only render logo in OrgSelect when not all orgs have the same plantype

🎟 _Jira ticket:_ [PD-607](https://jira.mongodb.org/browse/PD-607)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -189,6 +189,15 @@ function OrgSelect({
   const { width: viewportWidth } = useViewportSize();
   const isTablet = viewportWidth < breakpoints.medium;
 
+  const checkPlanType = data?.[0]?.planType;
+  const showPlanType = data.reduce((shouldDisplay, datum) => {
```
hswolff, line 193: We probably want to use [every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) instead. This way if we find a case where we shouldn't display the logo the function will return right away, rather than iterating on every item.

brookescarlett: 👍

PR #369, opened 2020-04-14
New build process

hswolff:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


hswolff:
Investigating how material-ui builds and publishes their icons.

* [Their docs page](https://material-ui.com/components/icons/) that documents how you can consume it. This is ideally how we want to do it too.
* [The source code for their icons package](https://github.com/mui-org/material-ui/tree/master/packages/material-ui-icons)
  * Seems like they [use babel and compile their `src/` folder to `build` folder](https://github.com/mui-org/material-ui/blob/master/packages/material-ui-icons/package.json#L27)
  * Interestingly it looks like they then use the icon package.json as a base but [then transform it and copy it into the `build/` folder as well](https://github.com/mui-org/material-ui/blob/master/scripts/copy-files.js#L64)
  * Can see the modified [package.json file of icon here](https://unpkg.com/browse/@material-ui/icons@4.9.1/package.json)
  * Also they [only publish their build folder via `npm publish build`](https://github.com/mui-org/material-ui/blob/master/packages/material-ui-icons/package.json#L33) (I didn't know you could define a folder to publish!)

So basically they make the final deployable version entirely in a created `build/` folder, from all assets to package.json file as well. Fascinating!

hswolff:
For handling inlining images to `Checkbox` we'd probably have to write a custom script to transform the images to their base64 values (which webpack is doing for us right now).

The webpack file looks like

```
/* harmony default export */ __webpack_exports__["default"] = ("data:image/png;base64,iVBO
```

package.tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "module": "esnext",
-    "target": "esnext",
-    "jsx": "preserve",
+    "target": "ES5",
```
DesignerDave, line 4: Hmm... I'm curious about this change. Will this compile to ESM?

package.tsconfig.json
```diff
@@ -15,8 +15,11 @@
...
     },
-    "emitDeclarationOnly": true,
+    "sourceMap": true,
+    "declarationMap": true,
+    "emitDeclarationOnly": false,
+    "importHelpers": true,
     "composite": true
   },
-  "exclude": ["*.story.*", "*.spec.*", "dist"]
+  "exclude": ["**/*/*.spec.*", "**/*/*.story.*"]
```
DesignerDave, line 24: Do we not want to exclude compiling dist directories?

package.tsconfig.json
hswolff: Yep, that's they they have module and target fields. Module defines what module system to compile down to (esnext === ESM and also commonjs).

Target is more like babel, to what language spec to transpile down to. So having Object.assign() instead of `{ ...foo }`.

So, in short, yes.

package.tsconfig.json
hswolff: It seemed to not make a difference afaict. I can iterate on it more, but didn't seem to make a difference.

package.tsconfig.json
DesignerDave: 🤦 Right, forgot about the separate module field. Thanks!

DesignerDave:
Closed in favor of: https://github.com/mongodb/leafygreen-ui/pull/412

PR #372, opened 2020-04-15
PD-622: Update Org/Project Selects with serverside filtering for administrators

seanmhanson:
## ✍️ Proposed changes

The logic for the project and org selects have been updated to support more complex serverside filtering by global administrators necessary for support (and to handle such a large dataset at the database/service layer instead of the UI). 

In doing this, along with adding the fetch and logic I made three changes for performance:
- the filtered data is now handled as state, to prevent refetching and refiltering each render
- filtering now occurs when the provided data changes (i.e. after the initial mongo-nav call resolves), and when the filter value changes (i.e. when the user interacts with the select)
- additionally, I ensure that we never fetch the admin search results on the initial render, which would be blank anyway (an empty search term returns an empty response)

On an aside, per @DesignerDave's blessing, eslint is updated to allow using the rest operator to omit fields from a provided object (ex: `({me, ...coolKidsInGymClass}) => haveFun(coolKidsInGymClass)`)

🎟 _Jira ticket:_ [PD-622](https://jira.mongodb.org/browse/PD-622)

## 🛠 Types of changes

## ✅ Checklist

### Example: Searching for an org by objectid
Screen Shot 2020-04-14 at 8 30 00 PM

### Example: Searching for a project by scan
Screen Shot 2020-04-14 at 8 30 23 PM


packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -208,17 +238,43 @@ function OrgSelect({
...
+
+    // manage serverside filtering for a global user
+    if (admin) {
+      // don't fetch data initially, only when filtering
+      if (!didMount.current) {
+        didMount.current = true;
+        return;
+      }
+
+      filterDataAsAdmin();
```
hswolff, line 259: We need to debounce this as we currently do in ContextSwitcher. Otherwise every character input is going to create an API request.

Looks like we debounce for 300ms in mms.

hswolff: Also the behavior we have in ContextSwitcher when `value === ''` we reset back to the original data for the user, so the passed in `data` value.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -208,17 +238,43 @@ function OrgSelect({
...
+      return;
+    }
+
+    // otherwise, filter the provided data clientside
+    filterData();
+  }, [value]);
+
+  useEffect(() => {
+    setFilteredData(data);
+  }, [data]);
```
hswolff, line 269: So...I _think_ we can get rid of this if we assume that if `value === ''` we show the initial data. Anything that's not an empty string we show the filtered data.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -200,6 +206,30 @@ function OrgSelect({
...
+    const endpointURI = `${hosts.cloud}/user/shared/organizations/search${queryString}`;
+    return fetch(endpointURI, {
+      credentials: 'include',
+      mode: 'cors',
+      method: 'GET',
+    });
+  }
+
+  const filterDataAsAdmin = async () => {
+    const response = await fetchOrgsAsGlobalUser(value);
```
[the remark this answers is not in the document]
seanmhanson: I looked into this and found a set of eslint rules by Salesforce that include one for async/await. It seems like it has more detail than we would want, but I just wanted to throw it out there in case we want to add a custom rule to prevent async/await:

https://github.com/salesforce/eslint-plugin-lwc/blob/master/lib/rules/no-async-await.js

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
seanmhanson: I still needed this, but it allowed me to keep the logic about `value === ''` simple as a result.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -200,6 +207,24 @@ function OrgSelect({
     }
   };
 
+  const fetchOrgsAsGlobalUser = (searchTerm: string) => {
```
DesignerDave, line 210: Is there a reason we use the terminology "GlobalUser" here, but "Admin" below?

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -112,6 +120,29 @@ function ProjectSelect({
     }
   };
 
+  const fetchProjectsAsGlobalUser = (searchTerm: string) => {
```
DesignerDave, line 123: Shouldn't `searchTerm` be optional here? You're treating it as though it is in the ternary below.

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -112,6 +120,29 @@ function ProjectSelect({
...
+      mode: 'cors',
+      method: 'GET',
+    });
+  };
+
+  const filterDataAsAdmin = debounce(() => {
+    fetchProjectsAsGlobalUser(value)
+      .then(response => response.json())
+      .then(data =>
+        data.map(({ cid, gn }: ProjectFilterResponse) => ({
```
DesignerDave, line 137: Just a thought for potential improvement here – you could reassign these variables to the desired name in the function parameters to give them more meaning here:

```ts
.then(data => data.map(({ cid: projectId, gn: projectName }: ProjectFilterResponse) => ({ projectId, projectName }))
```

One more thought is that it may be more readable to move this map function (and the similar one in I believe the org select) to its own variable:

```ts
const filterDataAsAdmin = debounce(() => {
    const processData = ({
        cid: projectId,
        gn: projectName,
    }: ProjectFilterResponse) => ({ projectId, projectName });

    fetchProjectsAsGlobalUser(value)
      .then(response => response.json())
      .then(processData)
      .then(setFilteredData)
      .catch(console.error);
});

```

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -213,7 +275,7 @@ function ProjectSelect({
           </FocusableMenuItem>
 
           <ul className={ulStyle}>
-            {renderedData?.map(datum => renderProjectOption(datum))}
+            {filteredData?.map(datum => renderProjectOption(datum))}
```
DesignerDave, line 278: Nit: `filteredData?.map(renderProjectOption)`

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -208,17 +233,49 @@ function OrgSelect({
...
+    }
+
+    // clear filtering if the search term is cleared
+    if (value === '') {
+      setFilteredData(data);
+      return;
+    }
+
+    // serverside filtering (global read only)
+    if (admin) {
```
brookescarlett, line 253: This is breaking in Dev mode. Might want to add `if (admin && mode !== 'dev')` or something like that

packages/mongo-nav/src/MongoNav.story.tsx
```diff
@@ -27,6 +27,9 @@ storiesOf('MongoNav', module).add('Default', () => (
           'orgNavAccessManager',
         )}
         admin={boolean('admin', true)}
+        hosts={{
+          cloud: 'http://localhost:8080',
```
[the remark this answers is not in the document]
seanmhanson: I reverted this file to origin but didn't apparently include it in my commit. Coming at you in a second.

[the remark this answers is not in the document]
seanmhanson: Okay, updated. Good catch.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
seanmhanson: Ah, this wasn't something that I assumed we wanted to handle, but is an easy fix!

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
seanmhanson: good catch

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
seanmhanson: Habit, since we don't use the phrase "admin" in cloud code typically and instead refer to global users and global roles. I'll make it consistent for leafygreen's usages and keep everything to admin.

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
seanmhanson: It will always be passed in, but may be an empty string, since it is pulled from `value` and `value` is set using `setState` and should never be nullable. However, assigning a default value to `''` is definitely a good safeguard to add.

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
seanmhanson: Sure thing! I think that will be cleaner as well since there is so much chaining going on.

packages/mongo-nav/src/types.ts
```diff
@@ -176,8 +176,8 @@ export interface AccountInterface {
 export interface ProjectInterface {
   projectId: string;
   projectName: string;
-  planType: PlanType;
-  orgId: string;
+  planType?: PlanType;
+  orgId?: string;
```
seanmhanson, line 180: [commentary] these are marked as optional because the serverside endpoint used for admin filtering does not return a planType or orgId as part of the `GroupInfoView` class that is prepared and returned.

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -120,17 +153,48 @@ function ProjectSelect({
...
+    }
+
+    // clear filtering if the search term is cleared
+    if (value === '') {
+      setFilteredData(data);
+      return;
+    }
+
+    // serverside filtering (admin users only)
+    if (admin && mode !== 'dev') {
```
brookescarlett, line 173: Thank you :D

packages/mongo-nav/src/mongo-select/types.ts
```diff
@@ -1,15 +1,18 @@
 import { MouseEventHandler, ReactNode } from 'react';
-import { URLSInterface, OnChangeInterface } from '../types';
+import { MongoNavInterface, URLSInterface, OnChangeInterface } from '../types';
 
 export type MongoSelectUrls = NonNullable<URLSInterface['mongoSelect']>;
 
-export interface BaseMongoSelectProps {
+export type BaseMongoSelectProps = Pick<
```
brookescarlett, line 6: 👍

.changeset/neat-olives-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/mongo-nav': minor
```
brookescarlett, line 2: @hswolff should this be a major bump, or are we adding additional functionality before we release v1

seanmhanson: In reviewing this, I'm not sure documentation is up to date for the exported components. Maybe the upgrade to v1.0.0 could include a pass of all docs?

brookescarlett: I can take a docs pass right now!

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -192,14 +193,22 @@ function OrgSelect({
...
-  const showPlanType = !data?.every(datum => datum.planType === checkPlanType);
+  const isFiltered = value !== '';
+  const isAdminSearch = isFiltered && admin && mode === Mode.Production;
+
+  const renderedData = isFiltered ? filteredData : data;
+
+  const checkPlanType = renderedData?.[0]?.planType;
+  const showPlanType = !renderedData?.every(
+    datum => datum.planType === checkPlanType,
+  );
```
hswolff, line 211: I'm hoping this reduces the code paths, with `renderedData` ultimately being what's rendered.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -218,15 +227,20 @@ function OrgSelect({
...
-  const filterDataAsAdmin = debounce(() => {
-    const mapResponseToOrgs = (response: Array<OrgFilterResponse>) =>
-      response.map(({ groups, ...org }) => org);
-    fetchOrgsAsAdmin(value)
-      .then(response => response.json())
-      .then(mapResponseToOrgs)
-      .then(setFilteredData)
-      .catch(console.error);
-  }, 300);
+  const filterDataAsAdmin = useCallback(
```
hswolff, line 230: Without useCallback the debounce was being redefined on every component render (aka every time `value` changed). Such that if you typed in `test` you'd get 4 definitions of `filterDataAsAdmin` with each ultimately being invoked after their debounce time elapsed.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -380,7 +383,13 @@ function OrgSelect({
             )}
 
             <ul className={ulStyle}>
-              {filteredData?.map(renderOrganizationOption) ?? (
+              {isAdminSearch && isFetching && (
+                <li className={emptyStateStyle}>Searching...</li>
+              )}
+              {isAdminSearch && !isFetching && renderedData.length === 0 && (
+                <li className={emptyStateStyle}>No matches found</li>
+              )}
```
hswolff, line 391: This will be revised with Nelle's styles but for now I wanted to get some visual indicator for admins that something was happening.

PR #374, opened 2020-04-15
PD-585: Popover scroll fixes

DesignerDave:
## ✍️ Proposed changes

Popovers were using viewport-relative measurements for positioning in places that needed to be accounting for scrolling as well. This should fix popover (and by association, tooltip) when the page is scrolled).

🎟 _Jira ticket:_ [Tooltip is improperly positioned when page scrolls](https://jira.mongodb.org/browse/PD-585)

## 🛠 Types of changes

## ✅ Checklist


packages/popover/src/Popover.tsx
```diff
@@ -98,25 +102,35 @@ function Popover({
...
-    viewportSize,
-    lastTimeRefElMutated,
-    active,
-    align,
-    justify,
-    forceUpdateCounter,
-  ]);
+  // We don't memoize these values as they're reliant on scroll positioning
+  const referenceElViewportPos = getElementViewportPosition(referenceElement);
+  const contentElViewportPos = getElementViewportPosition(contentNode);
```
DesignerDave, line 107: Thinking about this more, we might be able to similarly memoize these values. I'm open to thoughts on this.

packages/popover/src/positionUtils.ts
```diff
@@ -125,13 +164,14 @@ interface TransformOriginArgs {
 type XOrigin = 'left' | 'right' | 'center';
 type YOrigin = 'top' | 'bottom' | 'center';
 
-const yJustifyOrigins: { [J in Justify]: YOrigin } = {
+const yJustifyOrigins: Record<Justify, YOrigin> = {
```
jetpacmonkey, line 167: These aren't equivalent! Using `Record` means it won't fail if we add a value to the `Justify` enum in the future.

packages/code/README.md
```diff
@@ -76,13 +76,13 @@ const SomeComponent = () => <Code>{codeSnippet}</Code>;
...
+| `multiline`        | `boolean`                                                                                                                                                                                                                         | This prop determines whether or not the code snippet will visually retain line breaks when rendered.                                                                                                                                            | `true`    |
+| `lang`             | `'javascript'`, `'typescript'`, `'cs'`, `'cpp'`, `'go'`, `'java'`, `'perl'`, `'php'`, `'python'`, `'ruby'`, `'scala'`, `'swift'`, `'kotlin'`,`'objectivec'`, `'bash'`, `'shell'`, `'sql'`, `'yaml'`, `'json'`, `'auto'`, `'none'` | The language to render the code block as. When set to `'none'`, no syntax highlighting will be applied. When set to `'auto'`, the component will attempt to do its best to guess what language the code snippet is.                             | `'auto'`  |
```
brookescarlett, line 83: 👍

packages/popover/src/Popover.tsx
brookescarlett: Would that require creating a useScroll hook?

packages/popover/src/positionUtils.ts
DesignerDave: Interesting – I just tested this, and adding a new property to the Justify enum does cause an error because the property is missing.

DesignerDave: Looking into it, the definition for the Record helper is:
```ts
/**
 * Construct a type with a set of properties K of type T
 */
type Record<K extends keyof any, T> = {
    [P in K]: T;
};
```

packages/popover/src/positionUtils.spec.ts
```diff
@@ -2581,12 +2852,28 @@ describe('positionUtils', () => {
     });
   });
 
-  describe('getElementPosition', () => {
+  describe('getElementDocumentPosition', () => {
+    test('given an element, it returns an object with information about its position', () => {
+      const div = document.createElement('div');
```
brookescarlett, line 2857: Should we set some initial information about the document to test this better? Like we do for the above tests?

packages/popover/src/Popover.tsx
DesignerDave: We don't need to update the component on scroll, we just need to take into account the current scroll values at the time when we're calculating the position.

packages/popover/src/positionUtils.spec.ts
DesignerDave: Potentially, though this is a direct port of what was already here. I was gonna start making some test improvements, but I started going down a rabbit-hole. 

For the time being, my goal here was simply to maintain the existing test coverage rather than expand it.

packages/popover/src/positionUtils.ts
jetpacmonkey: Oh, I stand corrected, then!

packages/popover/src/positionUtils.ts
```diff
@@ -1,20 +1,29 @@
 import { Align, Justify, ElementPosition } from './types';
 
-type ReferencePosition = ElementPosition;
-type ContentPosition = ElementPosition;
+interface ElementViewportPositions {
+  referenceElViewportPos: ElementPosition;
+  contentElViewportPos: ElementPosition;
```
hswolff, line 5: optional nit, the namespace already defines what this properties are.

```suggestion
  referenceEl: ElementPosition;
  contentEl: ElementPosition;
```

Ditto for `ElementDocumentPositions`

DesignerDave: I named them as such because the two interfaces are combined at a point, so having `referenceEl, contentEl` in both wouldn't work.

hswolff: Got it, I dig it.

packages/popover/src/positionUtils.ts
```diff
@@ -103,8 +114,29 @@ const defaultElementPosition = {
...
 
+export function getElementDocumentPosition(
+  element: HTMLElement | null,
+): ElementPosition {
+  if (!element) {
+    return defaultElementPosition;
+  }
+
+  const { top, bottom, left, right } = element.getBoundingClientRect();
+  const { offsetHeight: height, offsetWidth: width } = element;
```
jetpacmonkey, line 125: Doesn't `getBoundingClientRect` also return `height` and `width`?

jetpacmonkey: Oh but this was probably copy/paste and not new. So whatever :)

DesignerDave: Yeah, this component could use a bit of cleaning up, but I was trying to change as little as possible here to fix the bug.

PR #385, opened 2020-04-20
PD-625: Using keyboard back/forward shortcut should no longer switch tabs

brookescarlett:
## ✍️ Proposed changes

Bug discovered: Using keyboard back/forward shortcut switches tabs
This PR addresses that bug by adding a new hook: useKeyPress to determine if a modifier key is being pressed while a user is using the arrowkeys.

🎟 _Jira ticket:_ [PD-625](https://jira.mongodb.org/browse/PD-625)

## 🛠 Types of changes

## ✅ Checklist


packages/hooks/README.md
```diff
@@ -126,3 +136,21 @@ usePoller(onPoll, {
...
 | `optional.enabled`   | `boolean`  | Is polling enabled.                                                  | `true`  |
+
+## useKeyPress
+
+Hook that accepts a keyCode and returns a boolean based on whether or not that key is being pressed.
+
+### Example
+
+```js
+import { useKeyPress } from '@leafygreen-ui/hooks';
```
hswolff, line 147: Can we show an example of using this with lib's `keyMap` variable? Noting it's optional to use it?

packages/tabs/src/Tabs.tsx
```diff
@@ -132,36 +133,43 @@ function Tabs({
...
+          (current - 1 + enabledIndexes.length) % enabledIndexes.length
+        ],
+      );
     }
-  }
+  };
+
+  const enabled =
+    navigator.platform.indexOf('Mac') > -1
+      ? !useKeyPress(keyMap.Command)
```
hswolff, line 167: Conditional hooks can break their behavior. Instead we should change what argument we pass into `useKeyPress` and let the hook update accordingly.

Also we should be able to pull this logic outside of the hook so we don't re-run it every time.

```
const isMac = navigator.platform.indexOf('Mac') > -1;
const keyMapValue = isMac? command : control;
```

packages/tabs/src/Tabs.tsx
```diff
@@ -132,36 +133,43 @@ function Tabs({
...
+      : !useKeyPress(keyMap.Control);
+
+  useEventListener('keydown', handleArrowKeyPress, {
```
hswolff, line 170: Feels weird to have this additional hook being used. Why not just add a handler to `useKeyPress`?

Also this feels borderline like `useKeyPress` is only ever going to be needed by Tabs, I'm not sure I see the reusability of it yet.

May make more sense to define the hook in the Tabs file and we can see if we see ourselves needing it again in the future before we export it and make it a fully supported hook.

packages/hooks/README.md
brookescarlett: N/A -- moved useKeyPress to Tabs for now and am not exporting for external consumption

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: Not sure I understand what you mean here. (As an aside: I moved useKeyPress to tabs so its local and not external.) Happy to make the hook more robust if that's what you're getting at? But thought it made sense to reuse logic we already had.

packages/tabs/src/Tabs.tsx
```diff
@@ -132,36 +134,42 @@ function Tabs({
...
+          (current - 1 + enabledIndexes.length) % enabledIndexes.length
+        ],
+      );
     }
-  }
+  };
+
+  const isMac = navigator.platform.indexOf('Mac') > -1;
+  const keyMapValue = isMac ? keyMap.Command : keyMap.Control;
+  const enabled = !useKeyPress(keyMapValue);
```
DesignerDave, line 168: This change shouldn't require new keys in the keyMap. Keyboard events have the boolean property `metaKey` (if the windows key is pressed on windows, or the command key is pressed on mac). For handling ctrl+arrow (I believe that's the right shortcut on windows for the same behavior), there's the `ctrlKey` boolean property on the keyboard event.

packages/tabs/src/Tabs.tsx
brookescarlett: Nvm. Removed the hook entirely!

packages/tabs/src/Tabs.tsx
brookescarlett: Done!

packages/tabs/src/Tabs.tsx
```diff
@@ -132,36 +133,38 @@ function Tabs({
     setSelected(index);
   }
 
-  function handleKeyDown(e: React.KeyboardEvent) {
-    e.stopPropagation();
+  const enabledIndexLogic: () => [Array<number>, number] = () => {
```
DesignerDave, line 136: Most code is logic. Maybe `getEnabledIndexes`?

brookescarlett: 😂 kk

packages/tabs/src/Tabs.tsx
```diff
@@ -132,36 +133,38 @@ function Tabs({
     setSelected(index);
   }
 
-  function handleKeyDown(e: React.KeyboardEvent) {
-    e.stopPropagation();
+  const getEnabledIndexes: () => [Array<number>, number] = () => {
+    const enabledIndexes: Array<number> = childrenArray.reduce(
```
hswolff, line 137: This looks like it should be a `.filter`?

brookescarlett: It's a filter/map because we need to return the index rather than the child itself. Does this look cleaner?

``` const testIndexes = childrenArray
      .filter(child => !child.props.disabled)
      .map(child => childrenArray.indexOf(child));```

Comment (DesignerDave) [reply packages/tabs/src/Tabs.tsx:137]: Much cleaner. You could even take the map a step further – 

```js
.map(childrenArray.indexOf);
```

I can't think of a case where this indexOf would return -1, so this should be fine unless you can.

brookescarlett: Get this error: ```Array.prototype.indexOf called on null or undefined

DesignerDave: And you don't hit that issue when passing in the anonymous function wrapper? Seems strange...

hswolff: It's being given the wrong `this` value in that usage. With `.map(childrenArray.indexOf);` you're just passing in a reference to `indexOf` so when it's called the this value is undefined.

PR #391, opened 2020-04-21
PD-632: When onPremEnabled and in a project we're incorrectly showing project access as disabled

brookescarlett:
## ✍️ Proposed changes

- When onPremEnabled and in a project we're incorrectly showing project access as disabled

🎟 _Jira ticket:_ [PD-632](https://jira.mongodb.org/browse/PD-632)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -227,6 +226,14 @@ function OrgNav({
     );
   }
 
+  let displayProjectAccess: boolean;
+
+  if (onPremEnabled) {
+    displayProjectAccess = currentProjectName !== 'None' ? true : false;
+  } else {
+    displayProjectAccess = showProjectNav;
+  }
```
hswolff, line 235: Can define the happy path as the default path.

```suggestion
  let displayProjectAccess = showProjectNav;

  if (onPremEnabled) {
    displayProjectAccess = currentProjectName !== 'None' ? true : false;
  }
```

hswolff: Why is it being compared against the string 'None'? Someone could name their project that. We should check if the currentProjectId is set no?

brookescarlett: Checking against none currently bc that's the value we're setting as default. You're right, smarter to check against id. Will update now.

brookescarlett: Updated!

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -237,6 +237,7 @@ function MongoNav({
           hosts={hosts}
           mode={mode}
           currentProjectName={data?.currentProject?.projectName}
+          currentProjectId={data?.currentProject?.projectId}
```
hswolff, line 240: We have `currentProjectId` on line 99 to use.

packages/mongo-nav/src/org-nav/OrgNav.spec.tsx
```diff
@@ -340,6 +340,78 @@ describe('packages/mongo-nav/src/org-nav', () => {
     });
   });
 
+  describe('the Access Manager dropdown displays correctly', () => {
+    test('when onPrem and currentProject exist, the Project Access Manager link is not disabled and project name is displayed', () => {
```
hswolff, line 344: Maybe helps to make a describe block with onPrem is true and false to bucket those two conditionals? These sentences are looong.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -227,6 +228,14 @@ function OrgNav({
     );
   }
 
+  let displayProjectAccess: boolean;
+
+  if (onPremEnabled) {
+    displayProjectAccess = !!currentProjectId;
+  } else {
+    displayProjectAccess = showProjectNav;
+  }
```
hswolff, line 237: Is the same now?

```suggestion
  let displayProjectAccess = onPremEnabled ? !!currentProjectId : showProjectNav;
```

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: 👍

packages/mongo-nav/src/org-nav/OrgNav.spec.tsx
brookescarlett: 👍

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: 👍

PR #394, opened 2020-04-22
Update url to new project invite url

hswolff:
## ✍️ Proposed changes

## 🛠 Types of changes

I changed this URL on mms in a ticket yesterday. I have a redirect in place on mms so there's no urgency on merging this, but figured we might as well get the right version out.

FYI this can't be published until after mms goes live as this URL doesn't exist yet.

## ✅ Checklist


packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -287,6 +288,12 @@ function OrgSelect({
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const val = e.target.value.replace(/\\/g, '\\');
     setValue(val);
+
+    if (open === true && wasOpen !== open) {
+      // Opt out of type-checking as we just want to trigger the onElementClick handler
+      onElementClick(NavElement.OrgNavOrgSelectSearch)(e as any);
```
hswolff, line 294: Not a huge fan of this but the alternative would be to dropdown the type of `onElementClick` when this is the outlier.

DesignerDave: Could we at least type it as `React.Event` or something slightly more specific like that?

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -198,6 +199,12 @@ function ProjectSelect({
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const value = e.target.value;
     setValue(value);
+
+    if (open === true && wasOpen !== open) {
+      // Opt out of type-checking as we just want to trigger the onElementClick handler
+      onElementClick(NavElement.ProjectNavProjectSelectSearch)(e as any);
```
DesignerDave, line 205: I feel like I'm missing something, but what's the reason for treating this onChange event as an onClick from the POV of the consumer?

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
hswolff: I tried a few but none worked.

The issue is we type it as `MouseEvent` which is the most exact.

`interface MouseEvent<T = Element, E = NativeMouseEvent> extends UIEvent<T, E> {`

And using UIEvent:
image

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
hswolff: Simply? Ease.

I could expose another handler for notifying when this event occurs but that's going to involve a lot more conversation.

DesignerDave: My primary worry here is that we may want to deprecate this event firing in the future since it's inconsistent with the name and other uses of the prop. If that's acceptable to ya, I'm fine with it.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
DesignerDave: Ahh yeah, you're right. I suppose the primary issue here is the inconsistency of using the API like this (addressed in my other comment).

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
hswolff: You mean in general? And replace it with what? 

I mean I'm fine for this to fit here awkwardly or I can add a new prop..but that seems so much more painful.

Or I can just remove this code for now and deal with it in a follow-up ticket.

DesignerDave: To clarify, I don't feel too strongly about this, just that I could see this being a point of 
 confusion in the future because the name of the prop doesn't reflect the usage.

We can "awkwardly" add this here for now, though in the longer term I expect us to want to do one of the following:
1. We create an `onElementChange` prop that a consumer can hook into.
2. We rename and restructure the prop to something like `onNavEvent(navElement, eventType)`.

hswolff: I'm fine to change that in the future. When we upgrade mongo nav we'll just change the handler we use. 

This is just for analytics so I don't care terribly.

PR #403, opened 2020-04-29
Box component updates

brookescarlett:
## ✍️ Proposed changes

- Makes Box component more resilient
- Button now uses Box (Card and Typography also updated, as they were using the old version of Box)

🎟 _Jira ticket:_ [PD-610](https://jira.mongodb.org/browse/PD-610)

## 🛠 Types of changes

## ✅ Checklist


packages/box/src/Box.tsx
```diff
@@ -1,68 +1,69 @@
...
-  children: PropTypes.node,
+  as: PropTypes.elementType,
   href: PropTypes.string,
-  // @ts-ignore
-  component: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
 };
 
 export default Box;
+
+export type OverrideComponentProps<
```
DesignerDave, line 58: Locally, I was able to eliminate this type entirely by modifying `BoxProps` like so:
```ts
export type BoxProps<
  C extends React.ElementType,
  H extends string | undefined,
  P extends Record<string, any> = {}
> = P & BoxComponentProps<C, H> &
  Omit<
    React.ComponentPropsWithRef<C extends React.ComponentType ? any : C>,
    keyof BoxComponentProps
  >;
```

And then replacing all instances of `OverrideComponentProps` with it.

packages/box/src/Box.tsx
```diff
@@ -1,68 +1,69 @@
...
-  [key: string]: any;
-};
-
-export type BoxProps<T> = DivProps<T> | AnchorProps<T> | CustomElementProps<T>;
-
-function isCustomElement<T>(
-  props: BoxProps<T>,
-): props is CustomElementProps<T> {
-  return (props as any).component != null;
+interface BoxComponentProps<
```
DesignerDave, line 8: This interface can be removed by changing `BoxProps` to:
```ts
export type BoxProps<
  C extends React.ElementType,
  H extends string | undefined,
  P extends Record<string, any> = {}
> = P & {
  href?: H;
  as?: InferComponentType<C, H>;
} & Omit<
  React.ComponentPropsWithRef<C extends React.ComponentType ? any : C>,
  'href' | 'as'
>;
```

Generally, I'm just seeing a lot of repeating type parameters and such, so this can help trim that down a bit.

packages/box/src/Box.tsx
brookescarlett: 👍  Done

packages/box/src/Box.tsx
brookescarlett: 👍

packages/box/src/Box.spec.tsx
```diff
@@ -38,89 +24,111 @@ describe('packages/box', () => {
   }
 
   const sharedProps = { name: 'testName' };
-  const anchorProps = {
+  const anchorProps: {
```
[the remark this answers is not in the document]
brookescarlett: We can. It looks a little weird because the generics are static here but it works.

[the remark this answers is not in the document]
DesignerDave: May still be worth it just in case we update the interface.

packages/box/src/Box.spec.tsx
```diff
@@ -38,89 +24,109 @@ describe('packages/box', () => {
...
     test('it renders the child content', () => {
-      expect(renderedEls.child).toBeInTheDocument();
-    });
-
-    test('it preserves the shared props', () => {
-      expect(renderedEls.box).toHaveAttribute('name', sharedProps.name);
+      expect(child).toBeInTheDocument();
     });
   });
 });
```
lazytype, line 132: could we have a TODO to add "type tests" that test that type errors occur if there are missing or invalid props? we should be able to write these tests after we upgrade TS.

packages/box/src/Box.tsx
```diff
@@ -1,68 +1,58 @@
...
-
-    return <Box {...rest} ref={ref} />;
+type InferComponentType<C, H> = H extends string
+  ? 'a' | React.ComponentType
+  : C;
+
+export type BoxProps<
+  C extends React.ElementType,
+  H extends string | undefined,
+  P extends Record<string, any> = {}
```
hswolff, line 11: Can we expand the single letter characters to something more readable?

packages/box/src/Box.tsx
```diff
@@ -1,68 +1,58 @@
...
-  // @ts-ignore
-  component: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
 };
 
 export default Box;
+
+export type OverrideComponentCast<P> = <
+  C extends React.ElementType = 'div',
+  H extends string | undefined = undefined
+>(
```
hswolff, line 56: Ditto here, can we expand the generic variable names?

packages/box/src/Box.tsx
brookescarlett: Yes!!

packages/box/src/Box.tsx
brookescarlett: Yes!

packages/box/src/Box.spec.tsx
```diff
@@ -38,89 +24,123 @@ describe('packages/box', () => {
...
 
     test('it renders the child content', () => {
-      expect(renderedEls.child).toBeInTheDocument();
+      expect(child).toBeInTheDocument();
     });
+  });
 
-    test('it preserves the shared props', () => {
-      expect(renderedEls.box).toHaveAttribute('name', sharedProps.name);
+  describe('types work as expected', () => {
```
brookescarlett, line 133: @lazytype thoughts?

lazytype: ❤️
(technically we don't need to call `render`)

brookescarlett: yes good point, omitting now. Thank you!

PR #408, opened 2020-05-05
PD-649: Add "enabled" prop to Tooltip

DesignerDave:
## ✍️ Proposed changes

This PR adds an `enabled` prop to tooltip that defaults to `true`.

I also added a test for the tooltip becoming visible on hover while I was in there.

🎟 _Jira ticket:_ [PD-649](https://jira.mongodb.org/browse/PD-649)

## 🛠 Types of changes

## ✅ Checklist


packages/tooltip/src/Tooltip.story.tsx
```diff
@@ -45,11 +46,13 @@ storiesOf('Tooltip', module)
       trigger={({ children, ...rest }: any) => {
         return <button {...rest}>trigger {children}</button>;
       }}
+      data-testid="hello there"
```
brookescarlett, line 49: Do we need this?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -81,8 +83,11 @@ interface PopoverFunctionParameters {
   referenceElPos: ElementPosition;
 }
 
+type ModifiedPopoverProps = Omit<PopoverProps, 'active' | 'spacing' | 'refEl'>;
+
 interface TooltipProps
-  extends Omit<PopoverProps, 'active' | 'spacing' | 'refEl'> {
+  extends Omit<HTMLElementProps<'div'>, keyof ModifiedPopoverProps>,
```
brookescarlett, line 89: nice

packages/tooltip/src/Tooltip.tsx
```diff
@@ -211,7 +227,7 @@ function Tooltip({
       onClick: (e: MouseEvent) => {
         // ensure that we don't close the tooltip when content inside tooltip is clicked
         if (e.target !== tooltipRef.current) {
-          setOpen((curr: boolean) => !curr);
+          setOpen((curr: boolean) => enabled && !curr);
```
brookescarlett, line 230: Should we check for enabled on the line above, rather than within the setOpen function?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -237,9 +253,13 @@ function Tooltip({
   };
 
   useEventListener('click', handleBackdropClick, {
-    enabled: open && triggerEvent === 'click',
+    enabled: enabled && open && triggerEvent === 'click',
   });
 
+  // @ts-ignore
```
brookescarlett, line 259: can we explicitly pull this out of props rather than bringing in lodash/omit

packages/tooltip/src/Tooltip.story.tsx
DesignerDave: Nope lol good catch!

packages/tooltip/src/Tooltip.tsx
DesignerDave: I put it here because in this case, it will close it if it's disabled. If enabled is `false` and the tooltip is open for some reason, we want the tooltip to close.

packages/tooltip/src/Tooltip.tsx
DesignerDave: I wasn't able to get TypeScript to understand this is a valid prop when trying to destructure it. Maybe if we're ts-ignoring here though, we could just ts-ignore it where we would destructure it...

We have set a precedent of using omit elsewhere though, and lodash is likely to be a dependency of most apps already, so I don't think the cost is all that high here.

Still, if you feel strongly about this, I can change it!

brookescarlett: I'm trying to remove many usages of lodash/omit with my Box PR, which is why I called this out here. Reasoning makes sense, so I'm ok leaving as-is although my preference would probably just be to push the @ts-ignore to where we're destructuring props

DesignerDave: I think I actually was able to remove the handling of the testid altogether and handle it through rest. Take a look!

brookescarlett: 👍

packages/tooltip/src/Tooltip.spec.tsx
```diff
@@ -53,6 +65,67 @@ describe('packages/tooltip', () => {
...
+      test(`when triggerEvent is "${triggerEvent}"`, async () => {
+        const { queryByTestId, button } = renderTooltip({
+          triggerEvent,
+          enabled: false,
+        });
+
+        fireEventMap[triggerEvent].on(button);
+
+        // Wait for 200ms to allow the component to animate out and remove the element from the DOM
+        await waitForTimeout(200);
```
hswolff, line 111: Can we use `waitFor` here instead, and wait until the element is removed from the DOM? This seems more brittle than declaring what we're waiting for.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -159,6 +170,7 @@ function Tooltip({
   trigger,
   variant = Variant.Light,
   triggerEvent = TriggerEvent.Hover,
+  enabled = true,
```
hswolff, line 173: What happens if a tooltip is shown and then enabled is set to false? It should hide the tooltip at that point right?

We need a test for this too?

DesignerDave: The way it works now is that it will close and remain closed on the next hover or click event (based on the `triggerEvent`).

packages/tooltip/src/Tooltip.spec.tsx
DesignerDave: This test is specifically testing that nothing changes when the trigger is interacted with, so there's no DOM change to wait for in a passing case.

packages/tooltip/src/Tooltip.spec.tsx
```diff
@@ -53,6 +65,67 @@ describe('packages/tooltip', () => {
...
+
+        expect(queryByTestId(tooltipTestId)).toBeFalsy();
```
hswolff, line 113: ?

```suggestion
        expect(queryByTestId(tooltipTestId)).not.toBeInTheDocument();
```

Ditto below.

packages/tooltip/src/Tooltip.tsx
hswolff: I feel like if a user sets enabled to false they expect the tooltip to not exist at that point in time. I.e. the JSX that is being rendered in the tooltip content may now be in a state that could be broken?

DesignerDave: I suppose that point is fair. I will say that the JSX is always being based on the same state as it was prior, and shouldn't be all that brittle. I'll make the change though, since I think the case you're describing is fair.

packages/tooltip/src/Tooltip.spec.tsx
DesignerDave: In this case they are functionally identical – queryByTestId returns null when the element is not in the document. Still, I'll update for consistency's sake.

hswolff: Yeah I know. The benefit of this named expectation is two fold:

- Is clearer to the reader what you're asserting, so easier to maintain as you know what you're testing for
- If the expectation is not met then the error message given by react-testing-library is clearer.

packages/tooltip/src/Tooltip.tsx
DesignerDave: Alright – changed the behavior here, and added a test for that behavior!

PR #409, opened 2020-05-05
PD-662: Allows consumers to override type prop in Input component

brookescarlett:
## ✍️ Proposed changes

Allows consumers to override type prop in Input component

🎟 _Jira ticket:_ [PD-662](https://jira.mongodb.org/browse/PD-662)

## 🛠 Types of changes

Screen Shot 2020-05-05 at 5 09 31 PM

### DIFF
--- a/packages/text-input/src/TextInput.tsx
+++ b/packages/text-input/src/TextInput.tsx
@@ -286,10 +286,10 @@ const TextInput = React.forwardRef(
         {description && <p className={descriptionStyle}>{description}</p>}
         <div className={inputContainerStyle}>
           <input
+            type="text"
             {...inputSelectorProp.prop}
             {...rest}
             className={cx(inputStyle, getStatefulInputStyles(state, optional))}
-            type="text"
             value={value}
             required={!optional}
             disabled={disabled}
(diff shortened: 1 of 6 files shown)


packages/text-input/src/TextInput.tsx
```diff
@@ -286,10 +286,10 @@ const TextInput = React.forwardRef(
         {description && <p className={descriptionStyle}>{description}</p>}
         <div className={inputContainerStyle}>
           <input
+            type="text"
```
DesignerDave, line 289: Could we make this the default value in the destructuring assignment, and then pass type explicitly?

brookescarlett, line 289: To note: the type will error if it's not compliant with expected type values on an HTML input element, as the props are extending those of an HTMLInputElement

DesignerDave: I'd also love for us to add explicit typing to allow for just text input types. I think they're:
`email`, `password`, `search`, `text`, `url`, `tel`. I could be missing one...

brookescarlett: 🤦 makes sense -- updated!

brookescarlett:
Just updated!

PR #412, opened 2020-05-08
PD-431 : Build glyphs individually

DesignerDave:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-431](https://jira.mongodb.org/browse/PD-431)

## 🛠 Types of changes

## ✅ Checklist


packages/icon/scripts/build.ts
```diff
@@ -0,0 +1,177 @@
+const svgr = require('@svgr/core').default;
+// @ts-ignore
+const fs = require('fs');
+// @ts-ignore
+const template = require('../src/template');
```
hswolff, line 5: nit, put relative imports last

packages/icon/scripts/build.ts
```diff
@@ -0,0 +1,177 @@
...
+// @ts-ignore
+const path = require('path');
+// @ts-ignore
+const meow = require('meow');
+
+const cli = meow(
+  `
+	Usage
+		$ node ./build.js <filename(s)>
```
hswolff, line 14:
```suggestion
		$ ts-node ./build.js <filename(s)>
```

packages/icon/src/template.ts
```diff
@@ -0,0 +1,143 @@
+/**
+usage:
```
hswolff, line 2: Guess we can remove this comment block now.

packages/icon/src/template.ts
```diff
@@ -0,0 +1,143 @@
...
+        title={getGlyphTitle(glyphName, title)}
+        height={size}
+        width={size}
+        {...props}
+      />
+    );
+  }
+}
+**/
+interface FirstArg extends Record<string, any> {
```
hswolff, line 40: Taking from here: https://react-svgr.com/docs/custom-templates/#custom-component-template

```suggestion
interface BabelAPI extends Record<string, any> {
```

packages/icon/src/template.ts
```diff
@@ -0,0 +1,143 @@
...
+  template: {
+    smart: (opts: Record<string, any>) => Record<string, any>;
+  } & Record<string, any>;
+}
+
+interface SecondArg extends Record<string, any> {
```
hswolff, line 46: Taking from here: https://react-svgr.com/docs/custom-templates/#custom-component-template

```suggestion
interface SVGROptions extends Record<string, any> {
```

packages/icon/src/template.ts
```diff
@@ -0,0 +1,143 @@
...
+  state: {
+    componentName: string;
+  } & Record<string, any>;
+  typescript: boolean;
+}
+
+interface ThirdArg extends Record<string, any> {
```
hswolff, line 53: Taking from here: https://react-svgr.com/docs/custom-templates/#custom-component-template

```suggestion
interface ASTParts extends Record<string, any> {
```

packages/icon/src/template.ts
```diff
@@ -0,0 +1,143 @@
...
+  jsx.openingElement.attributes = jsxAttributes.expression.openingElement.attributes.concat(
+    jsx.openingElement.attributes[2],
+  );
+
+  return typeScriptTpl.ast`
+    ${imports}
+    import PropTypes from 'prop-types';
+    import { css, cx } from '@leafygreen-ui/emotion';
+
+    function getGlyphTitle(name, title) {
```
hswolff, line 96: A size perf idea I had was to eventually move this function into its own file and import it, instead of re-creating the function every time.

Can do it later though.

packages/icon/src/template.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      return title;
+    }
+
+    function generateGlyphTitle() {
+      return '${componentName}' + '-' + Math.floor(Math.random() * 1000000);
+    }
+
+    const ${componentName} = ({ className, size = 16, title, customTitleId,  fill, ...props }) => {
+      const { current: titleId } = React.useMemo(() => customTitleId || generateGlyphTitle());
```
hswolff, line 124: I love this! Gotta [tweak it according to the docs](https://reactjs.org/docs/hooks-reference.html#usememo)

```suggestion
      const { current: titleId } = React.useMemo(() => customTitleId || generateGlyphTitle(), [customTitleId]);
```

packages/icon/scripts/build.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    fs.writeFileSync(path.resolve(outputDir, `${name}.d.ts`), fileContent);
+  });
+}
+
+function buildSvgFiles(input: Array<string>, flags: Flags) {
+  let svgFiles: Array<FileObject>;
+
+  if (input && input.length) {
```
hswolff, line 92: Does ts-node let us do:

```suggestion
  if (input?.length) {
```

packages/icon/scripts/build.ts
```diff
@@ -0,0 +1,177 @@
...
+    svgFiles = input.filter(filterSvgFiles).map(filePath => {
+      const splitPath = filePath.split('/');
+      const fileName = splitPath[splitPath.length - 1].replace('.svg', '');
```
hswolff, line 95: You want [path.basename](https://nodejs.org/api/path.html#path_path_basename_path_ext) if for the rare chance a Windows contributor wants to use this repo, then this will allow that to still work.

(I've learned this the hard way a long time ago)

packages/icon/scripts/build.ts
```diff
@@ -0,0 +1,177 @@
...
+      .map((fileName: string) => ({
+        name: fileName.replace('.svg', ''),
+        path: path.resolve(glyphsDir, fileName),
+      }));
+  }
+
+  svgFiles.forEach(file => {
+    const fileContent = fs.readFileSync(file.path, { encoding: 'utf8' });
+
+    svgr(
```
hswolff, line 117: May want `svgr.sync` cuz you're mixing a sync loop with async code.

packages/icon/scripts/build.ts
```diff
@@ -0,0 +1,177 @@
...
+        if (!fs.existsSync(outputDir)) {
+          fs.mkdirSync(outputDir, { recursive: true });
+        }
+
+        fs.writeFileSync(
+          path.resolve(outputDir, `${file.name}.js`),
+          moduleCode,
+        );
+      })
+      .catch(console.error);
```
hswolff, line 172: We probably want to `process.exit(1)` with an error if we encounter an error, rather than just logging?

packages/icon/src/template.ts
DesignerDave: Thank you so much. I hated this lol

packages/icon/src/template.ts
DesignerDave: Good catch

packages/icon/scripts/build.ts
DesignerDave: Looks like it does!

packages/icon/scripts/build.ts
DesignerDave: I'm not sure I see a problem here 🤔 Since the file it's reading and writing to are different for every iteration of the loop, we should be fine here and not have to worry about any race conditions right?

packages/icon/scripts/build.ts
DesignerDave: Oh boy, sometimes I forget people code on windows! haha

Will fix this.

packages/icon/scripts/build.ts
DesignerDave: 👍 Think I glossed over this because we're using `require()`.

packages/icon/scripts/build.ts
DesignerDave: Good catch!

packages/icon/scripts/build.ts
hswolff: In practice we're probably fine, but theoretically we're not. So might as well be safer than sorry.

packages/icon/scripts/build.ts
DesignerDave: Done!

packages/icon/scripts/build.ts
DesignerDave: Alright, fair enough. Done!

packages/icon/src/template.ts
hswolff: I think this is actually pretty easy to do now, just need to make a new file that exports this function and update this template to import it. TypeScript will transpile that file for us automatically now so it'll be a nice and easy relative import.

hswolff:
This looks good to me as far as I'm concerned!

I don't love the addition of `pre-build` cuz it implies to me that one day we're going to need a `pre-pre-build`, but....when that day comes we can tackle that dependency tree issue then.

Let's move `getGlyphTitle` to its own file, should be a nice quick win.

I'm interested to see the externals change you're talking about too.

Also I think to fix the build being broken we need to update `.github/workflows/nodejs.yml` to also run `yarn pre-build`.

I love this PR.

DesignerDave:
> This looks good to me as far as I'm concerned!
> 
> I don't love the addition of `pre-build` cuz it implies to me that one day we're going to need a `pre-pre-build`, but....when that day comes we can tackle that dependency tree issue then.
> 
> Let's move `getGlyphTitle` to its own file, should be a nice quick win.
> 
> I'm interested to see the externals change you're talking about too.
> 
> Also I think to fix the build being broken we need to update `.github/workflows/nodejs.yml` to also run `yarn pre-build`.
> 
> I love this PR.

RFAL!

I don't love pre-build either, but it's a bit of a side-effect of a build that uses Typescript for some stuff, and Webpack for others.

As we spoke about, due to the dependency graph, moving the function to another file is actually a heavier lift than that, so we're omitting that change here.

Externals change is in – looks pretty similar to how we list all our packages as externals dynamically.

Giving that GitHub CI change a try now, we'll see if that works!

.github/workflows/nodejs.yml
```diff
@@ -52,6 +51,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
       - run: yarn install
+      - run: yarn build
```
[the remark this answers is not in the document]
DesignerDave: You're right for the time being, though it's strange to run it independently IMO, and we'll need to do this anyways to eventually test against our built files instead of src (if that's still something we want to do).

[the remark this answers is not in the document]
hswolff: Meh, it's not like `pre-build` is a private command, it's broken out for convenience. And it's not like our CI build scripts follow any conventions, they just do what they need to test what we want.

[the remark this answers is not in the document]
DesignerDave: Fair enough, will change that

PR #414, opened 2020-05-20
PD-525: Table Component

brookescarlett:
👋 Was hoping to get a first pass from you both. Doesn't need to be a really in-depth PR review, but wanted to make sure that we were all on the same page re overall approach. If a first pass looks good to you, then would be happy to get more granular feedback while I start writing the test suite/documentation. Thank you very much!


packages/table/src/Table.story.tsx
```diff
@@ -0,0 +1,110 @@
...
+  color: string;
+  location: string;
+}
+
+storiesOf('Table', module)
+  .add('Default', () => (
+    <Table
+      selectable={boolean('isSelectable', true)}
+      data={defaultData}
+      columns={[
```
hswolff, line 20: Do you think its possible to also accept a Fragment here so we could do:

```jsx
columns={
   <>
       <TableHeader />
        Location
   </>
}
```

Is that possible?

packages/table/src/Table.story.tsx
```diff
@@ -0,0 +1,110 @@
...
+          dataType={DataType.String}
+          sortable
+          label="Favorite Color"
+          key="color"
+          accessor="color"
+        />,
+        'Location',
+      ]}
+    >
+      {({ datum }: { datum: DemoDataInterface; rowIndex: number }) => (
```
hswolff, line 44: I wonder if we could infer the interface for `datum` by inferring the shape of one object from the `data` prop.

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+// * Add nodeRef to Row
+
+const tableStyles = css`
+  border-collapse: collapse;
+  box-sizing: border-box;
+`;
+
+interface TableProps extends React.ComponentPropsWithoutRef<'table'> {
+  data?: Array<any>;
+  columns?:
```
hswolff, line 19: I think columns and data is required, no? Otherwise what do we render?

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+  };
+
+  const [state, dispatch] = useReducer(reducer, initialState);
+
+  React.useEffect(() => {
+    const rowCheckedState: {
+      [n in number]: undefined | boolean;
+    } = {};
+
+    rows.forEach((datum, index) => {
```
hswolff, line 49: This is confusing as `rows` is defined below. Can we make it so the variable exists before the effect is defined?

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+    if (!children) {
+      return null;
+    }
+
+    if (state.selectable) {
+      return rows.map((row, index) => {
+        const selectCell = <CheckboxCell index={index} />;
+
+        return React.cloneElement(row, {
+          children: [selectCell, [...Array.from(row.props.children)]],
```
hswolff, line 180: We probably want to be using [React.Children](https://reactjs.org/docs/react-api.html#reactchildren) utils when manipulating children.  Ditto elsewhere.

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+        });
+      });
+    }
+
+    return rows;
+  };
+
+  return (
+    <TableProvider state={state} dispatch={dispatch}>
+      <table cellSpacing="0" cellPadding="0" className={tableStyles} {...rest}>
```
hswolff, line 190: We should support `className` to append className styles here.

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+      type: Types.SortTableData,
+      payload: {
+        columnId,
+        key,
+        data,
+      },
+    });
+  };
+
+  const renderHeader = (array: Array<any>): React.ReactNode => {
```
hswolff, line 113: We could potentially make this its own component so we could test this behavior in isolation. Might help us find all cases easily.

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+    const cols = array.map((child, index) => {
+      if (isComponentType(child, 'HeaderRow')) {
+        usingHeaderRow = true;
+
+        const { children } = child?.props;
+
+        const renderedChildren = Array.isArray(children)
```
hswolff, line 120: Maybe https://reactjs.org/docs/react-api.html#reactchildrentoarray helps?

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+          children: renderHeader(renderedChildren),
+        });
+      }
+
+      if (isComponentType(child, 'TableHeader')) {
+        const { label, accessor: accessorProp } = child.props;
+
+        let glyph = 'Unsorted';
+
+        const accessor = accessorProp || label?.toLowerCase();
```
hswolff, line 134: I thought we'd do `accessorProp || key`? You're always going to at least need a key so it's doubtful we'd ever fall back to the label lowercased.

Also `accessProp` needs to support being a function so the table supports sorting nested objects. `accessor={data => data.whatever.foo}`.

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+        if (state.sort?.key?.toLowerCase() === accessor) {
+          glyph =
+            state.sort?.direction === 'asc'
```
hswolff, line 138: Let's not duplicate our checks on `state.sort?`. Let's grab the object once and then use it after we've ensured it exists.

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+          index,
+          glyph,
+        });
+      }
+
+      if (typeof child === 'string') {
+        return (
+          <TableHeader
+            // safe until columns need to be reordered
+            key={index}
```
hswolff, line 156: This is where I'd say to do `child.toLowerCase()` as a nice default.

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+      } else {
+        dispatch({
+          type: Types.ToggleHeaderIndeterminate,
+          payload: true,
+        });
+      }
+    }
+  }, [state.data, state.rowCheckedState]);
+
+  let usingHeaderRow = React.useMemo(() => false, [children]);
```
hswolff, line 95: Can we move this to the first line of renderHeader? Why is it out here? What does this memo get us?

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+  let rows: Array<React.ReactElement>;
+
+  if (typeof children === 'function') {
```
hswolff, line 98: We're going to require children be a function, no? So maybe we just assume it is and use it as such.

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,196 @@
...
+  data = [],
+  selectable: selectableProp = false,
+  children,
+  ...rest
+}: TableProps) {
+  const initialState: State = {
+    sort: {
+      direction: 'asc',
+    },
+    data,
```
hswolff, line 36: When the data prop changes how are we applying those new values to the table? I don't see any code that supports that.

packages/table/src/TableHeader.tsx
```diff
@@ -0,0 +1,120 @@
...
+  align-items: center;
+  color: ${uiColors.gray.dark2};
+  padding-right: 4px;
+`;
+
+interface TableHeaderInterface {
+  label: React.ReactElement | string;
+  onClick?: (colId: number, key: string) => void;
+  index?: number;
+  glyph?: string;
```
hswolff, line 40: These are known string literals right? Let's do that!

packages/table/src/index.ts
```diff
@@ -0,0 +1,6 @@
+export { default } from './Table';
```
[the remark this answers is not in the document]
brookescarlett: Yeah that makes sense.

packages/table/src/storybookdata.ts
```diff
@@ -0,0 +1,19 @@
+export const defaultData = [
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/table-context.tsx
```diff
@@ -0,0 +1,247 @@
...
+  ZipCode: 'zipCode',
+  String: 'string',
+  Date: 'date',
+} as const;
+
+type DataType = typeof DataType[keyof typeof DataType];
+
+export interface State {
+  sort?: Sort;
+  data?: Array<any>;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/table-context.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+    if (direction !== 'desc') {
+      return alphanumericCollator.compare(aVal, bVal);
+    }
+
+    return alphanumericCollator.compare(bVal, aVal);
+  });
+};
+
+function updateRowCheckedState(
```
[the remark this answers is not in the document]
brookescarlett: Just for clarity it was a lot of code so wanted to break it out. But I can inline.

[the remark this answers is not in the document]
brookescarlett: Good call.

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/TableHeader.tsx
brookescarlett: good point

packages/table/src/Table.story.tsx
brookescarlett: Yes possible, done!

packages/table/src/Table.tsx
```diff
@@ -14,19 +14,25 @@ const tableStyles = css`
...
+  index: number;
+}
+export interface TableProps extends React.ComponentPropsWithoutRef<'table'> {
+  data: Array<any>;
+  columns:
     | Array<ReactElement<HeaderRowProps>>
     | Array<ReactElement<TableHeaderProps> | string>;
   selectable?: boolean;
+  children: (ChildArgObject: ChildrenArgInterface) => React.ReactElement;
 }
```
hswolff: I tried playing around with this and couldn't figure it out but it'd be awesome if we could do something like this:

```
interface ChildrenArgInterface<Shape> {
  datum: Shape;
  index: number;
}


export interface TableProps<Datum extends{} infer U ? U : any> extends React.ComponentPropsWithoutRef<'table'> {
  data: Array<U>;
  children: (ChildArgObject: ChildrenArgInterface<U>) => React.ReactElement;
}
```

packages/table/src/Table.tsx
```diff
@@ -14,19 +14,25 @@ const tableStyles = css`
   box-sizing: border-box;
 `;
 
-interface TableProps extends React.ComponentPropsWithoutRef<'table'> {
-  data?: Array<any>;
-  columns?:
+interface ChildrenArgInterface {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: Updated this, let me know what you think.

packages/table/src/Table.tsx
brookescarlett: But what if the child is a react element? child.toString().toLowercase() ?

brookescarlett: Sorry, this works here, but not when cloning the element right above.

packages/table/src/Table.tsx
brookescarlett: When using `key` here React throws: 

```Warning: TableHeader: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://fb.me/react-special-props)```

File: packages/table/src/Cell.tsx [hunk 22]
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { commonCellStyles } from './styles';
+
+interface CellProps extends React.ComponentPropsWithRef<'td'> {}
+
+const tdStyles = css`
+  line-height: 16px;
+  letter-spacing: 0px;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Cell.tsx
```diff
@@ -0,0 +1,35 @@
...
+  position: relative;
+`;
+
+const innerDivStyles = css`
+  display: flex;
```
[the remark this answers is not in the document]
brookescarlett: it helps us align when there's an expandable Icon.. open to feedback, but this was the best solution that I could figure out.

packages/table/src/Cell.tsx
```diff
@@ -0,0 +1,35 @@
...
+  flex-direction: row;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/CheckboxCell.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const {
+    state: { headerCheckState },
+    dispatch,
+  } = useTableContext();
+  const [checked, setChecked] = React.useState(headerCheckState);
+
+  const handleChange = () => {
+    setChecked(curr => !curr);
+
+    if (typeof index === 'number') {
```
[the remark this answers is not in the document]
brookescarlett: I can't just check if index exists because if it's 0 it'll be false.

packages/table/src/HeaderRow.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import Checkbox from '@leafygreen-ui/checkbox';
+import { uiColors } from '@leafygreen-ui/palette';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { useTableContext, Types } from './table-context';
+
+const thStyles = css`
+  width: 40px;
+  border-width: 0px 1px 3px 1px;
```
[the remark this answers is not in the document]
brookescarlett: Kk! Will look into an eslint rule right now.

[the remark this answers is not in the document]
brookescarlett: thoughts: https://stylelint.io/user-guide/rules/length-zero-no-unit

packages/table/src/HeaderRow.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const handleChange = () => {
+    dispatch({
+      type: Types.ToggleHeaderCheckedState,
+      // investigate not passing a payload
+      payload: undefined,
+    });
+  };
+
+  let checkColSpan = true;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/HeaderRow.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  React.Children.forEach(children, child => {
+    if ((child as React.ReactElement)?.props?.colSpan) {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,310 @@
...
+          {...rest}
+        >
+          {renderedChildren}
+        </tr>
+
+        <Transition in={isExpanded} timeout={150} nodeRef={ref}>
+          {(state: string) => {
+            const props = {
+              ['data-testid']: isExpanded
+                ? 'nested-row-is-expanded'
```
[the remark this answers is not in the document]
brookescarlett: Yeah that's probably a better call, I'll look into that.

packages/table/src/HeaderRow.tsx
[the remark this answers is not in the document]
brookescarlett: .. talked offline, will take to a separate PR

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,310 @@
...
+      }
+    });
+
+    const shouldAltRowColor = data && data.length >= 10 && !hasNestedRows;
+
+    const alignmentStyles: Array<string> = [];
+
+    for (const key in columnInfo) {
+      alignmentStyles.push(styleColumn(key, columnInfo?.[key]?.dataType));
+    }
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

brookescarlett: 👍 fixed

packages/table/src/Table.tsx
brookescarlett: I think this works!!!
`type DataShape<Shape> = Shape extends infer U ? U : Shape;

interface TableRowInterface<Shape> {
  datum: DataShape<Shape>;
  index?: number;
}

export interface TableProps<Shape>
  extends React.ComponentPropsWithoutRef<'table'> {
  data: Array<DataShape<Shape>>;
  columns:
    | Array<ReactElement<HeaderRowProps>>
    | Array<ReactElement<TableHeaderProps> | string>
    | React.ReactFragment;
  selectable?: boolean;
  children: (TableRowArgs: TableRowInterface<Shape>) => JSX.Element;
}`

packages/table/src/Table.story.tsx
brookescarlett: Think I got it!

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,98 @@
...
+  datum: DataShape<Shape>;
+  index?: number;
+}
+
+export interface TableProps<Shape>
+  extends React.ComponentPropsWithoutRef<'table'> {
+  data: Array<DataShape<Shape>>;
+  columns:
+    | Array<ReactElement<HeaderRowProps>>
+    | Array<ReactElement<TableHeaderProps> | string>
```
hswolff, line 27:
```suggestion
    | Array<ReactElement<HeaderRowProps | TableHeaderProps> | string>
```

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,98 @@
...
+    data,
+    selectable: selectableProp,
+    headerCheckState: false,
+    headerIndeterminate: false,
+    hasNestedRows: false,
+  };
+
+  const [state, dispatch] = useReducer(reducer, initialState);
+
+  React.useEffect(() => {
```
hswolff, line 54: Let's be consistent with how we use hooks. We have a plain `useReducer` above and then this is off the React object. I prefer importing the individual functions, but prefer consistency above all else.

packages/table/src/table-context.tsx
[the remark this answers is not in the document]
hswolff: Can we also ensure that data will always at least be an array? Can and should we always make that assumption?

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,98 @@
...
+    | Array<React.ReactElement>
+    | undefined = state.data?.map((datum, index) => children({ datum, index }));
+
+  const renderBody = () => {
+    if (!children) {
+      return null;
+    }
+
+    return rows;
+  };
```
hswolff, line 83: This is a simple ternary we can inline.

packages/table/src/table-context.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+// type State
+interface Sort {
+  columnId?: number;
+  direction: 'asc' | 'desc';
+  key?: string;
+}
+
+const DataType = {
+  NominalNumber: 'nominalNumber',
```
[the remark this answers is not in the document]
brookescarlett: because they all sort of represent numbers, and this is what it was called in the docs that Rob sent me. Happy to make this just "number" though

packages/table/src/Table.tsx
brookescarlett: Sorry yes good point

packages/table/src/table-context.tsx
```diff
@@ -0,0 +1,233 @@
+import React, { createContext, useContext, useMemo } from 'react';
```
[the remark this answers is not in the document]
brookescarlett: Happy to update.

packages/table/src/table-context.tsx
```diff
@@ -0,0 +1,233 @@
...
+    </TableContext.Provider>
+  );
+}
+
+// useTableContext
+function useTableContext() {
+  return useContext(TableContext);
+}
+
+export { TableProvider, useTableContext, reducer, Types, DataType };
```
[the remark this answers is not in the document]
brookescarlett: yes happy to do that

packages/table/src/table-context.tsx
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/table-context.tsx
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/table-context.tsx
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Table.story.tsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { boolean } from '@storybook/addon-knobs';
+import { Table, Row, Cell, TableHeader, HeaderRow, DataType } from '.';
+import { defaultData, multiRowData } from './fixtures';
+
+storiesOf('Table', module)
+  .add('Default', () => (
+    <Table
+      selectable={boolean('isSelectable', true)}
```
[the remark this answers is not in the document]
brookescarlett: Good catch, idk when introduced this. Will investigate.

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.story.tsx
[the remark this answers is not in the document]
brookescarlett: Fixed

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,239 @@
...
+        payload: A[Key];
+      };
+};
+
+type Action = ActionMap<ActionPayload>[keyof ActionMap<ActionPayload>];
+
+// type Dispatch
+type Dispatch = (action: Action) => void;
+
+// type State
```
hswolff, line 58: This comment is misplaced? Or you're breaking up sections of the file?

I'd suggest being bolder if that's the case:

```
//
// State
//

```

brookescarlett: Was breaking up sections, but I can eliminate if you don't think its necessary

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,239 @@
...
+      };
+
+    case Types.SetHasNestedRows:
+      return {
+        ...state,
+        hasNestedRows: action.payload,
+      };
+
+    case Types.SetRowCheckedState:
+      console.log('ayo', action);
```
hswolff, line 127: ayoooo

brookescarlett: lol idk why i always leave these in 🤦

packages/table/src/TableContext.tsx
brookescarlett: 👍

packages/table/src/TableContext.tsx
brookescarlett: 👍

packages/table/src/TableHead.tsx
```diff
@@ -0,0 +1,93 @@
...
+          children: createCols(React.Children.toArray(children)),
+        });
+      }
+
+      if (isComponentType(child, 'TableHeader')) {
+        const { label, accessor: accessorProp } = child.props;
+
+        let glyph = 'Unsorted';
+
+        const accessor = accessorProp || label?.toLowerCase();
```
hswolff, line 45: This doesn't handle calling accessor if it's a function.

What we did in mms is normalized the data such that the accessor is always a function. What we also did is when the accessor was a function is we passed in each datum object when sorting so you could do custom value look ups.

For example if you have a complex array like:

```jsx
const data = [
    {
        name: 'harry',
        job: {
            team: 'cloud',
            title: 'punmaster'
        },
    }
];

// Accessor could be:
<TableHeader
  accessor={(data) => data.job.title}
>
```

For convienience we also let you do dot lookup as a string, so with the above example you could have done `accessor="job.title"` but I don't think anyone really knows that exist.

Luckily we have an [existing implementation of this normalize behavior in mms.](https://github.com/10gen/mms/blob/624053ad04198e130771ebd0d723e1dd6c7011f3/client/js/common/components/Table/TablePlain.js#L16-L33)

packages/table/src/TableHead.tsx
```diff
@@ -0,0 +1,93 @@
...
+            onClick={sortRows}
+          />
+        );
+      }
+
+      return child;
+    });
+  }
+
+  const columnArray: Array<React.ReactElement> =
```
hswolff, line 76: This feels like we could memo-ize this but I'd want to meaures that before we do.

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,228 @@
...
+export interface State {
+  sort?: Sort;
+  data: Array<any>;
+  columnInfo?: {
+    [k in number]: { dataType?: DataType };
+  };
+  selectable?: boolean;
+  headerCheckState?: boolean;
+  headerIndeterminate?: boolean;
+  rowCheckedState?: Record<number, boolean | undefined>;
```
hswolff, line 84: Do you think we could have this and the two header state's be simply booleans and not also undefined? It'd be nice if we didn't have to worry about the undefined case and simply rely on `true || false`.

packages/table/src/CheckboxCell.tsx
```diff
@@ -0,0 +1,74 @@
...
+  children,
+  className,
+  disabled,
+  index,
+}: CheckboxCellProps) {
+  const {
+    state: { headerCheckState },
+    dispatch,
+  } = useTableContext();
+  const [checked, setChecked] = React.useState(headerCheckState);
```
hswolff, line 28: Is this just to make things look faster?

We shouldn't do this if we don't have to and just fix the core underlying performance issues.

hswolff: We know what to toggle it to when we actually dispatch the event, right before we set the new value we pull the old one and invert it.

packages/table/src/fixtures.ts
```diff
@@ -0,0 +1,19 @@
+export const defaultData = [
+  { name: 'Alice', age: 19, color: 'white', location: 'bedford' },
```
hswolff, line 2: Let's add some nested objects here for a richer example set.

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,312 @@
...
+import IconButton from '@leafygreen-ui/icon-button';
+import Icon from '@leafygreen-ui/icon';
+import { isComponentType } from '@leafygreen-ui/lib';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+import CheckboxCell from './CheckboxCell';
+import { useTableContext, Types, DataType } from './TableContext';
+
+const rowStyle = css`
+  cursor: pointer;
```
hswolff, line 12: This should only be pointer if we're in selectable mode right?

It's weird each row is a pointer when you can't click on anything.

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,312 @@
...
+      ...rest
+    }: RowProps,
+    ref: React.Ref<any>,
+  ) => {
+    const {
+      state: { data, columnInfo, hasNestedRows, hasRowSpan, selectable },
+      dispatch,
+    } = useTableContext();
+
+    const indexRef = React.useRef(Math.floor(Math.random() * 1000000 + 1));
```
hswolff, line 134: So React will only use the value given from here the first time, but JavaScript will run this code everytime. This can be slow.

Let's instead move the logic into a function (defined outside the component) and I believe you can pass in a function to useRef to get the initial value: `useRef(generateNumber)`.

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,312 @@
...
+          }
+
+          return React.cloneElement(child, {
+            children: <span className={truncation}>{children}</span>,
+          });
+        }
+      }
+    });
+
+    const shouldAltRowColor = data && data.length >= 10 && !hasNestedRows;
```
hswolff, line 250: May be my screen, but I don't see this in storybook.

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,312 @@
...
+      >
+        <Icon
+          aria-label="chevron"
+          glyph={isExpanded ? 'ChevronDown' : 'ChevronRight'}
+          color={uiColors.gray.dark2}
+        />
+      </IconButton>
+    );
+
+    const nestedRows = React.Children.map(children, child => {
```
hswolff, line 194: Can we merge this into the below map so we only iterate over children once?

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,312 @@
...
+            <CheckboxCell
+              index={indexRef.current}
+              className={cx({ [disabledCell]: disabled })}
+              disabled={disabled}
+            />
+          )}
+          {renderedChildren}
+        </tr>
+
+        <Transition
```
hswolff, line 283: Can we conditionally render this if nestedRows exist?

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,312 @@
...
+                  {chevronButton}
+                  <span className={truncation}>{children}</span>
+                </>
+              ),
+              className: cx(displayFlex, className),
+              key: children,
+            });
+          }
+
+          if (!children) {
```
hswolff, line 239: Do we care if there's no children? Can we just render it anyways?

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+// type Action
+const Types = {
+  SelectableTable: 'SELECTABLE_TABLE',
+  ToggleHeaderCheckedState: 'TOGGLE_HEADER_CHECKED',
+  ToggleHeaderIndeterminate: 'TOGGLE_HEADER_INDETERMINATE',
+  SetColumnInfo: 'SET_COLUMN_INFO',
+  SortTableData: 'SORT_TABLE_DATA',
+  ToggleIndividualChecked: 'TOGGLE_INDIVIDUAL_CHECKED',
+  SetRowCheckedState: 'SET_ROW_CHECKED_STATE',
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,94 @@
...
+          type: Types.ToggleHeaderIndeterminate,
+          payload: true,
+        });
+      }
+    }
+  }, [state.rowCheckedState]);
+
+  const rows:
+    | Array<React.ReactElement>
+    | undefined = state.data.map((datum, index) => children({ datum, index }));
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    <TableProvider state={state} dispatch={dispatch}>
+      <table
+        cellSpacing="0"
+        cellPadding="0"
+        className={cx(tableStyles, className)}
+        {...rest}
+      >
+        <TableHead data={data} columns={columns} />
+        <tbody>{!children ? null : rows}</tbody>
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/TableHead.tsx
```diff
@@ -0,0 +1,93 @@
...
+      }
+
+      if (isComponentType(child, 'TableHeader')) {
+        const { label, accessor: accessorProp } = child.props;
+
+        let glyph = 'Unsorted';
+
+        const accessor = accessorProp || label?.toLowerCase();
+
+        if (sort && sort.key?.toLowerCase() === accessor) {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: Did you look at no nested rows?

brookescarlett: Screen Shot 2020-07-09 at 11 52 13 AM

packages/table/src/TableContext.tsx
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: Good call

packages/table/src/Row.tsx
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,94 @@
...
+  useEffect(() => {
+    dispatch({
+      type: Types.SelectableTable,
+      payload: selectableProp,
+    });
+  }, [selectableProp]);
+
+  useEffect(() => {
+    if (state.rowCheckedState) {
+      const boolArray = Object.values(state.rowCheckedState);
```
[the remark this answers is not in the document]
brookescarlett: 👍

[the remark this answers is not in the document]
brookescarlett: Didn't figure out how to remove the loop altogether yet, but I did put this in the reducer.

packages/table/src/fixtures.ts
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: We do care if there's no children when rowSpan/colSpan are introduced

packages/table/src/Row.tsx
brookescarlett: Struggling to combine these because the second iteration relies on the updated children from the first map.

brookescarlett: Combined !

packages/table/src/Table.tsx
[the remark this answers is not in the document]
brookescarlett: Simplified this dramatically -- thank you @DesignerDave

packages/table/src/TableHead.tsx
brookescarlett: Got it! Just need to add tests for this, but I believe we have this working as intended.

packages/table/src/CheckboxCell.tsx
brookescarlett: Removed!

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,356 @@
...
+function getIndentLevelStyle(indentLevel: number) {
+  return css`
+    & > td:nth-child(2) {
+      padding-left: ${8 + indentLevel * 16}px;
+    }
+  `;
+}
+
+function generateIndexRef() {
+  return Math.floor(Math.random() * 1000000 + 1);
```
lazytype, line 106: not that I'd ever hope for a table with tens of thousands of rows for the probability of collision to be feasible, but we could a larger set of possible doing something like `Math.random().toString(36).substring(2)`

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,356 @@
...
+      onChange: () =>
+        dispatch({
+          type: Types.ToggleIndividualChecked,
+          payload: {
+            index: indexRef.current,
+            checked: !rowState[indexRef.current].checked,
+          },
+        }),
+      disabled,
+      checked: rowState[indexRef.current]?.checked || false,
```
lazytype, line 300: could do the `!!` trick here

packages/table/src/Table.spec.tsx
```diff
@@ -0,0 +1,299 @@
...
+        </Row>
+      )}
+    </Table>,
+  );
+  const table = screen.getByTestId('table');
+  return { table };
+}
+
+describe('packages/table', () => {
+  afterEach(cleanup);
```
lazytype, line 53: thought this happens automatically?

packages/table/src/Table.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+// Add tests for nested data structures to prove that sortBy prop works as expected
+
+const tableStyles = css`
+  border-collapse: collapse;
+  box-sizing: border-box;
+  border-bottom: 1px solid ${uiColors.gray.light2};
+`;
+
+type DataShape<Shape> = Shape extends infer U ? U : Shape;
```
lazytype, line 17: I think this will just always resolve to `Shape`?

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,314 @@
...
+type DataType = typeof DataType[keyof typeof DataType];
+
+export { DataType };
+
+export interface State {
+  sort?: Sort;
+  data: Array<any>;
+  columnInfo?: {
+    [k in number]: { dataType?: DataType };
+  };
```
lazytype, line 87: I like using the shorthand `Record<number, { dataType?: DataType }>`

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,314 @@
...
+}
+
+export function TableProvider({
+  children,
+  state,
+  dispatch,
+}: TableProviderInterface) {
+  const contextValue = useMemo(() => {
+    return { state, dispatch };
+  }, [state, dispatch]);
```
lazytype, line 236: neat!

packages/table/src/Cell.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const Cell = React.forwardRef(
+  ({ children, className, ...rest }: CellProps, ref: React.Ref<any>) => {
+    return (
+      <td
+        ref={ref}
+        className={cx(commonCellStyles, tdStyles, className)}
+        {...rest}
+      >
+        <div className={cx(innerDivStyles)}>{children}</div>
```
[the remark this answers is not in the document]
brookescarlett: Good catch

packages/table/src/CheckboxCell.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface CheckboxCellProps {
+  checked: boolean;
+  onChange?: React.ChangeEventHandler;
+  disabled?: boolean;
+  className?: string;
+  children?: React.ReactNode;
+}
+
+function CheckboxCell({
```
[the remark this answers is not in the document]
brookescarlett: Love this idea.

packages/table/src/Centered.tsx
```diff
@@ -0,0 +1,40 @@
...
+  align-items: center;
+`;
+
+interface CenteredProps {
+  children: React.ReactNode;
+  horizontal?: boolean;
+  vertical?: boolean;
+}
+
+function Centered({
```
brookescarlett, line 23: @DesignerDave I think you're opposed to this as an export but would be interested in making a utils package for us to consume this internally :)

brookescarlett: cc @lazytype

packages/table/src/HeaderRow.tsx
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  let checkColSpan = false;
+
+  React.Children.forEach(children, child => {
+    if (!child) {
+      return null;
+    }
+
+    if ((child as React.ReactElement).props?.colSpan) {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,356 @@
...
+      React.Children.forEach(children, child => {
+        if (
+          isComponentType(child, 'Row') &&
+          !hasDispatchedHasNestedRows &&
+          !hasNestedRows
+        ) {
+          dispatch({
+            type: Types.SetHasNestedRows,
+            payload: true,
+          });
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: 👍

packages/table/src/Table.spec.tsx
brookescarlett: it does indeed -- good call!

packages/table/src/TableContext.tsx
brookescarlett: 👍

packages/table/src/TableContext.tsx
brookescarlett: 👍

packages/table/src/Centered.tsx
```diff
@@ -0,0 +1,40 @@
...
+  children,
+  horizontal = false,
+  vertical = false,
```
lazytype, line 26: Hm, would it be counter-intuitive for neither prop being passed in `<Centered>` to default to centering both? just as an example of how this logic might look
```jsx
<Centered horizontal> or <Centered vertical={false}> // center horizontal
<Centered vertical> or <Centered horizontal={false}> // center vertical
<Centered> or <Centered horizontal vertical> // center both
```

brookescarlett: yeah actually i think you're right. if it's centered then the default behavior should be to center it

packages/table/src/Table.tsx
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
```diff
@@ -201,7 +201,7 @@ interface SubMenuProps {
 
   onClick?: React.MouseEventHandler;
 
-  onExited?: ExitHandler;
+  onExited?: ExitHandler<HTMLUListElement>;
```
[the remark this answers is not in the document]
brookescarlett: removed -- not necc

packages/table/src/CheckboxCell.tsx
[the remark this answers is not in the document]
brookescarlett: @DesignerDave and I debated this for a while and he's not the biggest fan. Still omitting <CheckboxCell /> but inlining the div styles

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+function generateIndexRef() {
+  return Math.floor(Math.random() * 1000000 + 1);
+}
+
+interface RowProps extends React.ComponentPropsWithoutRef<'tr'> {
+  expanded?: boolean;
+  disabled?: boolean;
+  indentLevel?: number;
+  isParentExpanded?: boolean;
```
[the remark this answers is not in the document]
brookescarlett: Great catch

packages/menu/src/SubMenu.tsx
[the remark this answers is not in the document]
brookescarlett: spoke too soon, it is necessary here, but onExited is only internal we don't expose that prop so should be fine

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,373 @@
...
+  `,
+};
+
+const styleMap = {
+  left: [DataType.String, DataType.Weight, DataType.ZipCode, DataType.Date],
+  right: [DataType.Number, DataType.Quantity],
+} as const;
+
+function styleColumn(index: string, dataType: DataType) {
+  let justify = '';
```
lazytype, line 83: would this produce invalid css? e.g. `justify-content: ;`

packages/table/src/Row.tsx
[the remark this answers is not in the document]
lazytype: should it be inherited by the children's children too, i.e. should this be `!isAncestorCollapsed`?

packages/table/src/TableHeader.tsx
```diff
@@ -0,0 +1,119 @@
...
+      return onClick?.(index, sortBy!);
+    }
+  };
+
+  React.useEffect(() => {
+    if (typeof index === 'number') {
+      dispatch({
+        type: Types.SetColumnInfo,
+        payload: {
+          index: selectable ? index + 2 : index + 1,
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/TableHeader.tsx
```diff
@@ -0,0 +1,119 @@
...
+  ...rest
+}: TableHeaderProps) {
+  const {
+    state: { selectable, sort },
+    dispatch,
+  } = useTableContext();
+
+  const glyph =
+    typeof sort?.columnId === 'number' && sort.columnId === index
+      ? sort?.direction
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Table.spec.tsx
```diff
@@ -0,0 +1,334 @@
...
+}
+
+describe('packages/table', () => {
+  // Test sortBy prop is a string, and a function
+  // Test table header as a string and a jsx element explicitly
+
+  test('by default, it renders unsorted table data, based on "data" prop', () => {
+    renderTable();
+    const firstRow = screen.getAllByRole('row')[1];
+    expect(firstRow.children[0].innerHTML).toContain('Alice');
```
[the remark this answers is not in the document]
brookescarlett: Was purposefully grabbing the first row and checking if it contained Alice rather than just searching for whether or not Alice was in the table

packages/table/src/Row.tsx
brookescarlett: Good call, providing a fallback.

packages/table/src/Row.tsx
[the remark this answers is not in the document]
brookescarlett: It's working without that additional prop, but what do you think isAncestorCollapsed would help with?

packages/table/src/Table.spec.tsx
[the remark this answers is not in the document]
lazytype: I meant specifically using it with the first row, `getByText(firstRow, "Alice")`

packages/table/src/Row.tsx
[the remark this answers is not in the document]
lazytype: Playing around with it locally, it seems like if I have a parent row that has a sub-row that itself has a sub-row, then collapsing the parent row doesn't collapse the sub-row's sub-row.

packages/table/src/Table.spec.tsx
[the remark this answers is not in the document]
brookescarlett: Ah yes that works

packages/table/src/Row.tsx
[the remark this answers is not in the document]
brookescarlett: https://jmp.sh/YT6hEZn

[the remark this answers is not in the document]
lazytype: Ok looks like I was actually testing a sub-sub-sub-row

https://jmp.sh/PIgMD0G
This is the change I made to the default Story
```diff
--- a/packages/table/src/Table.story.tsx
+++ b/packages/table/src/Table.story.tsx
@@ -53,6 +53,12 @@ storiesOf('Table', module)
                   <Cell>expanded age: {datum.age.second ?? datum.age}</Cell>
                   <Cell>expanded color: {datum.color}</Cell>
                   <Cell>{datum.location}</Cell>
+                  <Row>
+                    <Cell>expanded name: {datum.name}</Cell>
+                    <Cell>expanded age: {datum.age.second ?? datum.age}</Cell>
+                    <Cell>expanded color: {datum.color}</Cell>
+                    <Cell>{datum.location}</Cell>
+                  </Row>
                 </Row>
               )}
             </Row>
```

packages/table/src/HeaderRow.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const thStyles = css`
+  width: 40px;
+  border-width: 0 1px 3px 1px;
+  border-color: ${uiColors.gray.light2};
+  border-style: solid;
+`;
+
+export interface HeaderRowProps extends React.ComponentPropsWithoutRef<'tr'> {
+  indeterminate?: boolean;
```
lazytype, line 15: This doesn't seem like it should be a public prop. Can this be pulled from the table context instead?

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,317 @@
...
+    };
+  }
+
+  return {
+    checked: false,
+    indeterminate: true,
+  };
+};
+
+const setEveryrowState = (
```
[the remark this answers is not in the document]
brookescarlett: 👍  good catch

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,317 @@
...
+  sensitivity: 'base',
+});
+
+export const sortFunction = ({
+  data,
+  accessorValue,
+  direction,
+}: {
+  data: Array<any>;
+  accessorValue: (data: any) => string;
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+interface Sort {
+  columnId?: number;
+  direction?: 'asc' | 'desc';
+  accessorValue?: () => string;
+}
+
+const DataType = {
+  Number: 'Number',
+  Quantity: 'quantity',
```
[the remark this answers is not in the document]
brookescarlett: That's fine, that can be removed. The differentiations came from the design specs, but Number and Quantity are styled the same way so I don't see a reason why they can't be combined.

packages/table/src/HeaderRow.tsx
brookescarlett: Yes good call I actually don't think either of these are required.

PR #427, opened 2020-05-27
Adds yarn prettier to Version action

brookescarlett:
Adds yarn prettier to Version action

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,9 @@ jobs:
       - name: Build packages
         run: yarn build
 
+      - name: Run Prettier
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -30,3 +30,6 @@ jobs:
         uses: changesets/action@master
         env:
           GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+
+      - name: Run Prettier
+        run: yarn format
```
hswolff, line 35: I'd imagine we'd have to also commit these changes after, cuz right now it'd just make a dirty working tree.

PR #428, opened 2020-05-28
[Callout] Add a new Callout package

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Create Call-Outs React Component](https://jira.mongodb.org/browse/PD-653])

## 🛠 Types of changes

## ✅ Checklist


packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,203 @@
...
+  title?: string;
+  children: string;
+}
+
+interface CalloutProps
+  extends Omit<CustomCalloutProps, 'colorSet' | 'headerLabel' | 'headerIcon'> {
+  variant: Variant;
+}
+
+function CustomCallout({
```
hswolff, line 120: Curious, why is this a separate component? Why is this not embedded into the component itself?

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,203 @@
...
+    background: PropTypes.string.isRequired,
+    text: PropTypes.string.isRequired,
+  }).isRequired,
+  headerIcon: PropTypes.oneOf(Object.keys(glyphs)).isRequired,
+  headerLabel: PropTypes.string.isRequired,
+  title: PropTypes.string,
+  children: PropTypes.string.isRequired,
+};
+
+function Callout({ variant, title, children: contents }: CalloutProps) {
```
hswolff, line 184: We should probably support `className` so consumers can set a className for layout purposes.

packages/callout/src/Callout.tsx
lazytype: My thinking is that similar to breaking up a complex function into smaller functions, I wanted to break up the component into ones with more focused concerns. I would want to write more exhaustive tests for `CustomCallout` and more shallow tests for each variant of `Callout`.

hswolff: Hrm, I understand the sentiment but in this case I'm not sure I see how this allows for more exhaustive testing. I'm happy to hold that judgement until I see the tests, but as it looks now it just looks like we're moving the tests from Callout to CustomCallout. I'm not sure Callout is complex enough to require this decoupling of logic and API.

packages/callout/src/Callout.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+    });
+  });
+
+  describe('CustomCallout component', () => {
+    afterEach(async () => {
+      await cleanup();
+    });
+
+    test(`renders class name in classList: "${className}"`, async () => {
+      let { container } = renderCustomCallout({});
```
brookescarlett, line 62: Recently learned that it's actually preferred to grab this from screen rather than destructuring render. You can import `screen` where you import render from `react-testing-library`: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

packages/callout/src/Callout.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+          </CustomCallout>,
+        );
+
+        expect(calloutMarkup).toEqual(customCalloutMarkup);
+      }
+    });
+  });
+
+  describe('CustomCallout component', () => {
+    afterEach(async () => {
```
brookescarlett, line 57: No longer necessary to manually write this out, react-testing-library will do this for you

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+          fill={colorSet.foreground}
+          className={cx(headerIconStyle)}
+        />
+      )}
+      <div>{headerLabel}</div>
+    </div>
+  );
+
+  return (
+    <div
```
brookescarlett, line 151: I think this could benefit from having the role="note" applied: https://www.digitala11y.com/note-role/

packages/callout/src/Callout.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+    headerLabel,
+  };
+  return render(<CustomCallout {...defaultProps} {...props} />);
+}
+
+describe('packages/callout', () => {
+  describe('Callout component', () => {
+    test('renders each variant correctly', () => {
+      for (const variant of Object.values(Variant)) {
+        const calloutMarkup = renderToStaticMarkup(
```
brookescarlett, line 34: Have never used the `renderToStaticMarkup` helper before, will need to check it out but this test suite looks great :)

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+  }).isRequired,
+  headerIcon: PropTypes.oneOf(Object.keys(glyphs)),
+  headerLabel: PropTypes.string.isRequired,
+  title: PropTypes.string,
+  children: PropTypes.string.isRequired,
+  className: PropTypes.string,
+};
+
+function Callout({
+  variant,
```
brookescarlett, line 189: We probably should be setting a default variant rather than requiring this prop. Thoughts @DesignerDave?

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+  Tip: 'tip',
+  Important: 'important',
+  Warning: 'warning',
+} as const;
+
+export type Variant = typeof Variant[keyof typeof Variant];
+
+const baseStyle = css`
+  border-radius: 6px;
+  box-shadow: 0px 2px 5px 0 ${uiColors.gray.light2};
```
DesignerDave, line 18: I didn't catch this when reviewing the callout Sketch library, but we should be setting the color as a dark-gray or black RGBA value to work better on more backgrounds than white.

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+  display: flex;
+  overflow: hidden;
+  width: 700px;
```
DesignerDave, line 21: We should be relying on block element styling to set the width here to be 100% of its parent. The [guidelines](https://wiki.corp.mongodb.com/display/DESIGN/Call-Outs) don't call out any specific fixed width.

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+    content: '';
+    position: relative;
+    width: 6px;
+    left: 0px;
+    border-radius: 6px 0px 0px 6px;
+  }
+`;
+
+const headerStyle = css`
+  display: flex;
```
DesignerDave, line 33: From what I can tell, this shouldn't need to be display flex since the header icon is position absolute, and there's only one other child.

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+  font-size: 12px;
+  font-weight: bold;
+  letter-spacing: 0.8px;
+  line-height: 16px;
+  padding: 12px 24px 12px 52px;
+  text-transform: uppercase;
+`;
+
+const headerIconStyle = css`
+  margin-left: -32px;
```
DesignerDave, line 44: A note: We usually avoid negative margins wherever possible. In this case, I believe this CSS property can be changed simply by using `left` instead.

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+  position: absolute;
+`;
+
+const titleStyle = css`
+  font-weight: bold;
+  margin: 8px 0;
```
DesignerDave, line 50: Nit: Usually, we prefer to do:
```css
margin-top: 8px;
margin-bottom: 8px;
```

When you do `margin: 8px 0;` it's explicitly setting the left and right margin, which doesn't appear to be the intent here.

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+          background-color: ${colorSet.background};
+          color: ${colorSet.foreground};
+        `,
+      )}
+    >
+      {headerIcon && (
+        <Icon
+          glyph={headerIcon}
+          fill={colorSet.foreground}
+          className={cx(headerIconStyle)}
```
DesignerDave, line 143: `cx()` is used to compose multiple classNames. Since only one is being used here, you can simply do `className={headerIconStyle}`!

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+        />
+      )}
+      <div>{headerLabel}</div>
```
DesignerDave, line 146: Question: What's the purpose of this div here vs just rendering `headerLabel`?

packages/callout/src/Callout.tsx
DesignerDave: Ditto for other instances 😄

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+        <div className={cx(bodyStyle)}>
+          {title && <div className={cx(titleStyle)}>{title}</div>}
+          <div className={cx(bodyTextStyle)}>{contents}</div>
+        </div>
+      </div>
+    </div>
+  );
+}
+
+CustomCallout.propTypes = {
```
DesignerDave, line 175: We are usually fine relying on TS interfaces for anything internal. Generally, we include component PropTypes for the benefit of users not working with TypeScript.

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { uiColors } from '@leafygreen-ui/palette';
+import Icon, { glyphs } from '@leafygreen-ui/icon';
```
DesignerDave, line 5: We're beginning to move to importing icons directly from dist internally to improve the tree-shaking situation in users' apps since right now, the main files in our package.json point at a webpack-bundle, rather than an ES module 😞 

The pattern looks like:
```js
import Bulb from '@leafygreen-ui/icon/dist/Bulb;
```

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+};
+
+export const headerIcons: Record<Variant, keyof typeof glyphs> = {
+  [Variant.Note]: 'Edit',
+  [Variant.Tip]: 'Bulb',
+  [Variant.Important]: 'InfoWithCircle',
+  [Variant.Warning]: 'Warning',
+};
+
+export const colorSets: Record<Variant, ColorSet> = {
```
DesignerDave, line 78: 👍 I dig this approach!

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,215 @@
...
+  font-size: 14px;
+  line-height: 20px;
+`;
+
+export const headerLabels = {
+  [Variant.Note]: 'Note',
+  [Variant.Tip]: 'Tip',
+  [Variant.Important]: 'Important',
+  [Variant.Warning]: 'Warning',
+};
```
DesignerDave, line 69: I'd consider using `as const` for any of these objects with statically-known members.

packages/callout/src/Callout.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+        );
+
+        expect(calloutMarkup).toEqual(customCalloutMarkup);
+      }
+    });
+  });
+
+  describe('CustomCallout component', () => {
+    afterEach(async () => {
+      await cleanup();
```
DesignerDave, line 58: If cleanup is already a promise, is there a reason we can't do this?
```js
afterEach(cleanup)
```

packages/callout/src/Callout.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    test(`renders class name in classList: "${className}"`, async () => {
+      let { container } = renderCustomCallout({});
+      expect(container.firstChild).not.toHaveClass(className);
+      await cleanup();
+
+      ({ container } = renderCustomCallout({ className }));
```
DesignerDave, line 66: Rather than mutating the value of container, there's a supported API that we use elsewhere for updating the props and re-rendering a component through the testing library:

```js
const { container } = render(<MyComponent someProp={true} />);

render(
  <MyComponent someProp={false} />,
  { container },
);
```

packages/callout/src/Callout.spec.tsx
brookescarlett: We don't need cleanup anymore! react-testing-library calls this for us automatically!

packages/callout/src/Callout.spec.tsx
brookescarlett: I actually think you should be able to grab container from screen.
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

DesignerDave: Actually, even better would be just to use the `rerender` method: https://testing-library.com/docs/react-testing-library/api#rerender

packages/callout/src/Callout.tsx
hswolff: So the idea is to make this unit testable?

I'm not sure this component is complex enough to warrant this separation. IMO it makes it harder to read the code in full and piece together what the resulting rendered code looks like.

The tests you have for the CustomCallout seem simple enough to apply directly to just one component. That also lets us test the entire public API of the component, which we don't really have right now. We moved some of those tests to this inner component.

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,229 @@
...
+  Tip: 'tip',
+  Important: 'important',
+  Warning: 'warning',
+} as const;
+
+export type Variant = typeof Variant[keyof typeof Variant];
+
+const baseStyle = css`
+  border-radius: 6px;
+  box-shadow: 0px 2px 5px 0 rgba(33, 49, 60, 0.1);
```
lazytype, line 21: might be nice to have a help like this so we can re-use the palette colors
https://github.com/Chris-Baker/Emotion-RGBA

DesignerDave: Already have one! We use [Polished](https://polished.js.org/), and it's included via lib. It has a helper for this.

https://github.com/mongodb/leafygreen-ui/blob/master/packages/lib/package.json#L24

packages/callout/src/Callout.tsx
```diff
@@ -0,0 +1,196 @@
...
+function Callout({
+  variant,
+  title,
+  children: contents,
+  className,
+}: CalloutProps) {
+  const colorSet = colorSets[variant];
+  const Icon = headerIcons[variant];
+
+  const header = (
```
hswolff, line 150: I'm always a fan to inline things when possible, it makes reading and understanding the overall behavior of code easier than having to piece together different parts of code into one to get the complete picture.

I'm not going to make this desire blocking, mostly wanting to voice my dissent 😁

packages/callout/src/Callout.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+const children = 'this is the callout content.';
+const className = 'test-classname';
+
+const defaultProps = {
+  children,
+  variant: Variant.Note,
+};
+
+describe('packages/callout', () => {
+  describe('Callout component', () => {
```
hswolff, line 16: can remove this nesting now, no?

PR #429, opened 2020-06-01
Fix glyph `displayName`s

lazytype:
## ✍️ Proposed changes

Changes the generated icon code from:
```Edit.displayName = Edit;```
to
```Edit.displayName = 'Edit';```

## 🛠 Types of changes

This could technically be a breaking change, but I doubt anyone would be depending on the previous behavior.

## ✅ Checklist

### DIFF
--- a/packages/icon/src/Icon.spec.tsx
+++ b/packages/icon/src/Icon.spec.tsx
@@ -150,7 +150,7 @@ describe('packages/Icon/createIconComponent', () => {
   });
 });
 
-describe('Generated icons', () => {
+describe('Generated glyphs', () => {
   test('Edit icon has displayName: "Edit"', () => {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/icon/src/Icon.spec.tsx [hunk 1]
```diff
@@ -148,3 +149,9 @@ describe('packages/Icon/createIconComponent', () => {
     expect(glyph.textContent).toBe(text);
   });
 });
+
+describe('Generated icons', () => {
```
brookescarlett, line 153: Would change this to generated glyphs as I believe @DesignerDave made a distinction between the icon component itself and glyphs being the individual exports.

brookescarlett:
Don't forget to add a changeset but otherwise LGTM 👍

PR #433, opened 2020-06-02
PD-689: Removes inline-block from Body and InlineCode components

brookescarlett:
## ✍️ Proposed changes

Removes inline-block from Body and InlineCode components
🎟 _Jira ticket:_ [PD-689](https://jira.mongodb.org/browse/PD-689)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/typography/src/Typography.tsx
+++ b/packages/typography/src/Typography.tsx
@@ -14,14 +14,12 @@ const sharedStyles = css`
 `;
 
 const typeScale1 = css`
-  display: inline-block;
   font-size: 14px;
   line-height: 20px;
   letter-spacing: 0px;
 `;
 
 const typeScale2 = css`
-  display: inline-block;
   font-size: 16px;
   line-height: 24px;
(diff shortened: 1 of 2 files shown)


DesignerDave:
Hey @brookescarlett!

InlineCode needs to be `display: inline-block`, but that same CSS property breaks Body text. This change seems like it will fix one case, and break another.

brookescarlett:
Fixed @DesignerDave

packages/typography/src/Typography.tsx
```diff
@@ -14,14 +14,12 @@ const sharedStyles = css`
 `;
```
DesignerDave, line 14: (can't comment on line 10, thanks GitHub)

I realized that we're setting display block on just about everything. Headers are already display block by default, and the only tag that block being set on that isn't already is the `<small>` tag.

brookescarlett: Kk fixed!

PR #434, opened 2020-06-03
PD-616: Adds closeOnBackdropClick prop

brookescarlett:
## ✍️ Proposed changes

- Addresses: Cannot prevent closing modal when clicking backdrop by adding `closeOnBackdropClick` prop

🎟 _Jira ticket:_ [PD-616](https://jira.mongodb.org/browse/PD-616)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/modal/src/Modal.spec.tsx
+++ b/packages/modal/src/Modal.spec.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import Modal from './Modal';
 
@@ -57,6 +57,12 @@ describe('packages/modal', () => {
 
       window.dispatchEvent(new MouseEvent('click', { bubbles: true }));
       expect(container.innerHTML).toBe('');
+    });
+
+    test('when "closeOnBackdropClick" is false', () => {
+      renderModal({ closeOnBackdropClick: false, open: true });
+      window.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+      expect(screen.getByText(modalContent)).toBeInTheDocument();
     });
   });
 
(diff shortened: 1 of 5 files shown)


packages/modal/src/Modal.spec.tsx
```diff
@@ -58,6 +58,12 @@ describe('packages/modal', () => {
       window.dispatchEvent(new MouseEvent('click', { bubbles: true }));
       expect(container.innerHTML).toBe('');
     });
+
+    test('when "closeOnBackdropClick" is false', () => {
+      renderModal({ closeOnBackdropClick: false, open: true });
+      window.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+      expect(screen.getByText(modalContent)).toBeInTheDocument();
```
DesignerDave, line 65: I still don't like the pattern of selecting by text and other attributes, and would prefer a test-id. Technically, this only gets a child of the modal, and doesn't test that the whole modal is still in the document.

brookescarlett: Happy to switch this, just trying to follow the documented preferred patterns but if you feel strongly, let me know.

packages/modal/src/Modal.spec.tsx
```diff
@@ -58,6 +58,12 @@ describe('packages/modal', () => {
       window.dispatchEvent(new MouseEvent('click', { bubbles: true }));
       expect(container.innerHTML).toBe('');
     });
+
+    test('when "closeOnBackdropClick" is false', () => {
```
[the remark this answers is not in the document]
brookescarlett: Added!

packages/modal/src/Modal.spec.tsx
DesignerDave: I think I feel strongly enough about this – especially given that it doesn't technically test this correctly (as noted above).

brookescarlett: Fixed @DesignerDave!

PR #435, opened 2020-06-04
PD-690: Change host URL in MongoNav

brookescarlett:
## ✍️ Proposed changes

- Change host URL from 'stitch.mongodb.com' to 'realm.mongodb.com'
🎟 _Jira ticket:_ [PD-690](https://jira.mongodb.org/browse/PD-690)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/mongo-nav/src/MongoNav.spec.tsx
+++ b/packages/mongo-nav/src/MongoNav.spec.tsx
@@ -8,7 +8,7 @@ import {
 } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import { nullableElement, Queries } from 'packages/lib/src/testHelpers';
-import { dataFixtures } from './data';
+import { dataFixtures, getProductHost } from './data';
 import MongoNav from './MongoNav';
 
 // types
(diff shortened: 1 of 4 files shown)


packages/mongo-nav/src/data.ts
```diff
@@ -114,11 +114,21 @@ export const urlFixtures: Required<URLSInterface> = {
   },
 };
 
+export function getProductHost(today = new Date(Date.now())) {
```
DesignerDave, line 117: We're only using this in one place for the next week or two. I don't think we need to make `today` configurable.

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -417,4 +417,27 @@ describe('packages/mongo-nav', () => {
...
+    test('getProductHost function returns stitch', () => {
+      expect(getProductHost()).toBe('stitch');
+    });
+  });
+
+  describe('when the date is the day of MongoDB World', () => {
+    const mdbworld = new Date('June 8, 2020 0:00:00');
+
+    test('getProductHost function returns realm', () => {
+      const productName = getProductHost(mdbworld);
```
DesignerDave, line 431: Testing this like so doesn't seem to actually test that our implementation is correct. We don't pass any parameters to the function, so we should be testing that the function returns the correct name when nothing is passed.

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -417,4 +417,27 @@ describe('packages/mongo-nav', () => {
       });
     });
   });
+
+  describe('when the date is before MongoDB World', () => {
+    test('getProductHost function returns stitch', () => {
+      expect(getProductHost()).toBe('stitch');
```
DesignerDave, line 423: This test will break on the 8th.

packages/mongo-nav/src/data.ts
brookescarlett: The reason to do this is to make it testable. Followed the same convention used in ProjectNav

packages/mongo-nav/src/MongoNav.spec.tsx
brookescarlett: We do that on line 422.

packages/mongo-nav/src/MongoNav.spec.tsx
brookescarlett: You're right, I fixed this.

packages/mongo-nav/src/data.ts
DesignerDave: I don't think this is an effective test for how we're using the function.

You might consider overwriting the global Date constructor in the test to set it to the desired dates, and using the function with no parameter like it is in the code.

DesignerDave: https://stackoverflow.com/questions/8922145/overriding-the-javascript-date-constructor

PR #440, opened 2020-06-05
PD-677: Convert instances of Icon imports to import individual icons from dist

brookescarlett:
## ✍️ Proposed changes

Convert instances of Icon imports to import individual icons from dist

🎟 _Jira ticket:_ [PD-677](https://jira.mongodb.org/browse/PD-677)

## 🛠 Types of changes

## ✅ Checklist


packages/icon-button/src/IconButton.tsx
```diff
@@ -240,6 +240,21 @@ function usesLinkElement(
   return props.href != null;
 }
 
+function isComponentGlyph<T extends React.ReactElement>(
+  child: React.ReactNode,
+): child is T {
```
DesignerDave, line 245: This is indicating to typescript that the child is simply a ReactElement, not a Glyph. Not sure if there is a more strict way to type this. Perhaps something roughly like:
```ts
function isComponentGlyph<T extends React.ReactElement>(child: React.ReactNode): child is React.ComponentType<GlyphProps> {...}
```

packages/icon-button/src/IconButton.tsx
```diff
@@ -240,6 +240,21 @@ function usesLinkElement(
...
+  if (
```
DesignerDave, line 246: We can just do:

```ts
return child != null &&
    typeof child === 'object' &&
    'type' in child &&
    (child.type as any).displayName?.slice(0, 5) === 'Glyph'
```

DesignerDave: Also it's definitely worth adding some comments around why this is needed. We know, but I could see this being confusing and seeming unnecessary to someone less familiar.

packages/icon/src/Icon.spec.tsx
```diff
@@ -79,7 +79,7 @@ describe('packages/Icon/createGlyphComponent createGlyphComponent()', () => {
   });
 
   test('The function returned by createGlyphComponent has the displayName: "MyGlyph"', () => {
-    expect(GlyphComponent.displayName).toBe('MyGlyph');
+    expect(GlyphComponent.displayName).toBe('GlyphMyGlyph');
```
DesignerDave, line 82: We may need to add some documentation around custom glyph display names being used with `IconButton` if our tests needed to be updated here.

packages/icon/src/createGlyphComponent.tsx
```diff
@@ -3,6 +3,27 @@ import PropTypes from 'prop-types';
...
+
+export const sizeMap: Record<Size, number> = {
+  small: 14,
+  default: 16,
+  large: 20,
+  xlarge: 24,
+} as const;
+
+// We omit size here because we map string values for size to numbers in this component.
+export interface GlyphProps extends Omit<LGGlyph.ComponentProps, 'size'> {
```
DesignerDave, line 23: This might not be necessary now – I think this was from before Glyphs directly supported these string default sizes.

packages/icon/src/createGlyphComponent.tsx
```diff
@@ -47,20 +68,21 @@ export default function createGlyphComponent(
           className,
         )}
         title={getGlyphTitle(glyphName, title)}
-        height={size}
-        width={size}
+        height={typeof size === 'number' ? size : sizeMap[size]}
```
DesignerDave, line 71: Can we move this above and set this to a variable to avoid repeating this logic?

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -2,7 +2,11 @@ import React, { useState, useEffect, useCallback } from 'react';
 import debounce from 'lodash/debounce';
 
 // leafygreen-ui
-import Icon from '@leafygreen-ui/icon';
+import BuildingIcon from '@leafygreen-ui/icon/dist/Building';
+import CaretUpIcon from '@leafygreen-ui/icon/dist/CaretUp';
+import CaretDownIcon from '@leafygreen-ui/icon/dist/CaretDown';
+import SettingsIcon from '@leafygreen-ui/icon/dist/Settings';
+import { Size } from '@leafygreen-ui/icon';
```
DesignerDave, line 9: This might not work how we hope... I have a sinking feeling that this might include the full bundle still. There is a silver lining – I think you might be able to import this from `dist` as well now.

DesignerDave: Since we have typescript hinting, we could just use the string, `'small'` like we were before as well.

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -2,7 +2,10 @@ import React, { useState, useEffect, useCallback } from 'react';
 import debounce from 'lodash/debounce';
 
 // leafygreen-ui
-import Icon from '@leafygreen-ui/icon';
+import FolderIcon from '@leafygreen-ui/icon/dist/Folder';
+import CaretUpIcon from '@leafygreen-ui/icon/dist/CaretUp';
+import CaretDownIcon from '@leafygreen-ui/icon/dist/CaretDown';
+import { Size } from '@leafygreen-ui/icon';
```
DesignerDave, line 8: Ditto here

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -266,11 +268,18 @@ function ProjectSelect({
...
+          <CaretUpIcon
+            size={Size.Small}
+            className={cx(caretBaseStyle, { [iconLoadingStyle]: loading })}
+          />
+        ) : (
+          <CaretDownIcon
+            size={Size.Small}
+            className={cx(caretBaseStyle, { [iconLoadingStyle]: loading })}
+          />
+        )}
```
DesignerDave, line 282: These ternaries inline feel a bit worse than doing something like:

```js
const CurrentIcon = open ? CaretUpIcon : CaretDownIcon;

...
<CurrentIcon size='small' className={cx(caretBaseStyle, { [iconLoadingStyle]: loading })} /?
```

I don't feel that strongly about it if you do though.

packages/icon-button/src/IconButton.tsx
brookescarlett: 👍

packages/icon-button/src/IconButton.tsx
brookescarlett: 👍

packages/icon/src/createGlyphComponent.tsx
brookescarlett: 👍

packages/icon-button/src/IconButton.tsx
brookescarlett: If we bring in glyph props then we have to bring the entire icon library in and i'm not sure we want to do that, right?

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
brookescarlett: 👍

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
brookescarlett: 👍

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
brookescarlett: 👍

packages/icon/src/Icon.spec.tsx
brookescarlett: What do you think should be documented?

packages/icon/src/createGlyphComponent.tsx
brookescarlett: I added this -- I think it is necessary so that we can overwrite the expected value of the size key.

packages/icon/src/Icon.spec.tsx
DesignerDave: I suppose, since we're relying on "Glyph" being the prefix to any glyph display name in places, we should document something around that.

packages/icon/src/createGlyphComponent.tsx
DesignerDave: It doesn't need to be overwritten? `LGGlyph.ComponentProps` has the same type for size as this is overwriting. Additionally, the LGGlyph namespace should remain the single source of truth for these types.

packages/icon/src/createGlyphComponent.tsx
```diff
@@ -47,20 +70,21 @@ export default function createGlyphComponent(
...
+  GlyphComponent.displayName = `Glyph${glyphName}`;
 
   GlyphComponent.propTypes = {
     fill: PropTypes.string,
-    size: PropTypes.number,
+    size: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
     className: PropTypes.string,
   };
 
+  // @ts-ignore
```
DesignerDave, line 88: Why?

packages/icon/src/createIconComponent.tsx
```diff
@@ -32,20 +25,15 @@ export default function createIconComponent<
...
+    return <SVGComponent {...rest} size={size} />;
   };
 
   Icon.displayName = 'Icon';
 
   Icon.propTypes = {
     glyph: PropTypes.oneOf(Object.keys(glyphs)).isRequired,
     size: PropTypes.oneOfType([
-      PropTypes.oneOf(Object.keys(sizeMap)),
+      PropTypes.oneOf(Object.values(Size)),
```
DesignerDave, line 36: This should also accept a number or a string, right?

packages/icon/src/createGlyphComponent.tsx
```diff
@@ -47,20 +70,21 @@ export default function createGlyphComponent(
...
     );
   }
 
-  GlyphComponent.displayName = glyphName;
+  GlyphComponent.displayName = `Glyph${glyphName}`;
 
   GlyphComponent.propTypes = {
     fill: PropTypes.string,
-    size: PropTypes.number,
+    size: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
DesignerDave, line 84: We can be more specific here:
```js
    size: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf(Object.values(Size))]),
```

packages/icon/src/template.ts
```diff
@@ -7,6 +7,7 @@ interface BabelAPI extends Record<string, any> {
 interface SVGROptions extends Record<string, any> {
   state: {
     componentName: string;
+    size: 'small' | 'default' | 'large' | 'xlarge' | number;
```
DesignerDave, line 10: I don't think this belongs in `SVGROptions`.

packages/menu/src/SubMenu.tsx
```diff
@@ -251,6 +252,10 @@ const SubMenu = React.forwardRef((props: SubMenuProps, ref) => {
       ),
     });
 
+  const caretIconStyles = cx(open ? openIconStyle : closedIconStyle, {
```
DesignerDave, line 255: For cleanliness, it's probably best to make this ternary its own variable before using it here.

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -377,11 +379,17 @@ function OrgSelect({
...
           </span>
-          <Icon
-            size="small"
-            glyph={open ? 'CaretUp' : 'CaretDown'}
-            className={cx(caretBaseStyle, { [iconLoadingStyle]: loading })}
-          />
+          {open ? (
+            <CaretUpIcon
+              size="small"
+              className={cx(caretBaseStyle, { [iconLoadingStyle]: loading })}
```
DesignerDave, line 385: Can this className value be moved to a variable? We're repeating it twice.

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -3,7 +3,8 @@ import PropTypes from 'prop-types';
 import defaultsDeep from 'lodash/defaultsDeep';
 import Badge from '@leafygreen-ui/badge';
 import Button from '@leafygreen-ui/button';
-import Icon from '@leafygreen-ui/icon';
+import ArrowRightIcon from '@leafygreen-ui/icon/dist/ArrowRight';
+import { Size } from '@leafygreen-ui/icon';
```
DesignerDave, line 7: I think we might have to avoid these types of imports still since this is coming from a commonjs module. Otherwise, this may mitigate the benefits we'll get from doing direct imports.

Nothing wrong with using a string below anyways – typechecking will catch any typos.

packages/icon/src/Icon.spec.tsx
brookescarlett: Sounds good, I added as a note to the readme.

packages/icon/src/createGlyphComponent.tsx
brookescarlett: You're right -- sorry. This can be omitted.

packages/icon/src/createGlyphComponent.tsx
brookescarlett: Fixed!

packages/icon/src/createIconComponent.tsx
brookescarlett: number or specific string value, not any string value

packages/icon/src/createGlyphComponent.tsx
brookescarlett: Yep!

packages/icon/src/template.ts
brookescarlett: 👍

packages/menu/src/SubMenu.tsx
brookescarlett: 👍

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
brookescarlett: 👍

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: 👍

packages/icon/src/createIconComponent.tsx
```diff
@@ -32,20 +25,15 @@ export default function createIconComponent<
   const Icon = ({ glyph, size = Size.Default, ...rest }: IconProps) => {
     const SVGComponent = glyphs[glyph];
 
-    return (
-      <SVGComponent
-        {...rest}
-        size={typeof size === 'number' ? size : sizeMap[size]}
-      />
-    );
+    return <SVGComponent {...rest} size={size} />;
```
hswolff, line 28: Should we just remove size from the arguments and just pass through all props as the default is now handled in the glyph itself?

packages/icon/src/template.ts
```diff
@@ -93,11 +100,11 @@ module.exports = function template(
       return ${jsx};
     }
 
-    ${componentName}.displayName = ${componentName};
+    ${componentName}.displayName = 'Glyph' + '${componentName}';
```
hswolff, line 103: Let's remove the runtime cost.

```suggestion
    ${componentName}.displayName = 'Glyph${componentName}';
```

packages/icon-button/src/IconButton.tsx
```diff
@@ -240,6 +241,17 @@ function usesLinkElement(
...
 }
 
+function isComponentGlyph(
+  child: React.ReactNode,
+): child is React.ReactElement<LGGlyph.ComponentProps> {
+  return (
+    child != null &&
+    typeof child === 'object' &&
+    'type' in child &&
+    (child.type as any).displayName?.slice(0, 5) === 'Glyph'
```
hswolff, line 251: This is pretty gross...I wonder if we just inject a prop on the function so it's more discrete. `Glyph.isGlyph = true` and that lets this check be less fuzzy.

packages/icon/src/createGlyphComponent.tsx
```diff
@@ -3,6 +3,22 @@ import PropTypes from 'prop-types';
...
+export const Size = {
+  Small: 'small',
+  Default: 'default',
+  Large: 'large',
+  XLarge: 'xlarge',
+} as const;
+
+export type Size = typeof Size[keyof typeof Size];
+
+export const sizeMap: Record<Size, number> = {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/icon/src/template.ts
brookescarlett: 👍

packages/icon/src/createIconComponent.tsx
brookescarlett: 👍

packages/icon-button/src/IconButton.tsx
brookescarlett: 👍

packages/icon/src/template.ts
brookescarlett: Changing this causes a Babel error: `Unknown substitution "$5" given`

packages/icon/src/createGlyphComponent.tsx
```diff
@@ -74,6 +74,8 @@ export default function createGlyphComponent(
 
   GlyphComponent.displayName = `Glyph${glyphName}`;
 
+  GlyphComponent.isGlyph = 'true';
```
hswolff, line 77: Can this just be a boolean?

packages/icon/src/template.ts
hswolff: I wonder if you have to escape the characters. Or perhaps use a template string?

packages/icon/src/createGlyphComponent.tsx
brookescarlett: 👍

packages/icon/src/template.ts
brookescarlett: This is actually unnecessary now, because we are checking for isGlyph rather than needing to prefix the displayName as such. So reverted this back to '${componentName}'

PR #442, opened 2020-06-08
Upgrade TypeScript version to 3.9.5

lazytype:
## ✍️ Proposed changes

As title states

## 🛠 Types of changes

## ✅ Checklist


packages/pipeline/src/Stage.tsx
```diff
@@ -34,7 +34,7 @@ export interface StageProps {
   /**
    * The DOM node to use as the root node for the intersectionObserver. Defaults to window when null or undefined.
    **/
-  intersectionNode?: Element;
+  intersectionNode: HTMLElement | null;
```
DesignerDave, line 37: Based on the TSDoc annotation, this should be `intersectionNode?: HTMLElement | null;`, right?

DesignerDave: Also, this probably needs a patch changeset, even though it's small.

lazytype: sgtm!

PR #447, opened 2020-06-11
Add `portalClassName` prop to Popover

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Tooltip: Support providing a className to the portal root node](https://jira.mongodb.org/browse/PD-668)

Depends on #446. 

## 🛠 Types of changes

## ✅ Checklist


packages/popover/src/types.ts
```diff
@@ -115,4 +107,25 @@ export interface PopoverProps {
...
-}
+} & (
+  | {
+      /**
+       * Specifies that the popover content will appear portaled to the end of the DOM,
+       * rather than in the DOM tree.
+       *
+       * default: `true`
+       */
+      usePortal?: true;
```
DesignerDave, line 118: Do we need to add the TSDoc annotations to the other instance of this prop below?

packages/popover/src/Popover.tsx
```diff
@@ -219,7 +222,7 @@ function Popover({
               display: none;
             `}
           />
-          <Root>
+          <Root {...rootProps}>
```
DesignerDave, line 225: How come we aren't inlining the className here?
```js
<Root className={usePortal ? portalClassName : ''}>
```

The intent here seems a bit more clear than the `rootProps` constant above.

packages/popover/src/types.ts
lazytype: I actually don't know where the TSDoc annotations are used. To me it feels a bit redundant with the existing documentation of the the component.

packages/popover/src/Popover.tsx
lazytype: Since `Root` can be a `React.Fragment`, always passing in a `className` prop would give a runtime warning:
```Warning: Invalid prop `className` supplied to `React.Fragment`. React.Fragment can only have `key` and `children` props.```

PR #449, opened 2020-06-11
Add `portalClassName` prop to Tooltip

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Tooltip: Support providing a className to the portal root node](https://jira.mongodb.org/browse/PD-668)

Depends on #447. 

## 🛠 Types of changes

## ✅ Checklist


packages/tooltip/src/Tooltip.tsx
```diff
@@ -252,14 +277,20 @@ function Tooltip({
...
   const tooltip = (
     <Popover
+      key="tooltip"
       active={enabled && open}
       align={align}
       justify={justify}
-      usePortal={usePortal}
       adjustOnMutation={true}
       spacing={12}
+      {...portalProps}
```
hswolff, line 293: micro perf nit: seems un-necessary to create another object here, can we just directly pass the props directly?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -84,55 +84,78 @@ interface PopoverFunctionParameters {
...
+        /**
+         * If using a portal, specifies a class name to apply to the root element of the portal.
+         *
+         * default: undefined
+         */
+        portalClassName?: string;
+      }
+    | {
+        usePortal: false;
+        portalClassName?: never;
```
hswolff, line 156: Not sure how many times we'll duplicate this, but if we see us doing it more we should extract to something reusable.

packages/tooltip/src/Tooltip.tsx
lazytype: from a typing angle we can't because once we've destructured the props out, they lose the "OneOf"iness. we create another object in order to get to get the "OneOf"iness back.

PR #450, opened 2020-06-11
Add Yarn script command to type-check only

lazytype:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


dev.tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "pretty": true,
+    "strictNullChecks": true,
+    "noUnusedLocals": true,
+    "strict": true,
+    "allowSyntheticDefaultImports": true,
+    "moduleResolution": "node",
+    "baseUrl": ".",
+    "outFile": "/dev/null",
+    "tsBuildInfoFile": "./tsconfig.tsbuildinfo",
+    "incremental": true
```
lazytype, line 16: we can't just use "noEmit" because it would also prevent creating the `.tsbuildinfo` file and slow down type-checking

packages/tabs/src/Tabs.tsx
```diff
@@ -99,7 +99,9 @@ function Tabs({
   as = 'button',
   ...rest
 }: TabsProps) {
-  const childrenArray = React.Children.toArray(children);
+  const childrenArray = React.Children.toArray(children) as Array<
+    React.ReactElement
+  >;
```
lazytype, line 104: I don't know why this wasn't caught before 🤔

dev.tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "include": ["packages/**/*", "typings/**/*"],
+  "compilerOptions": {
```
hswolff, line 3: Can we not extend our `package.tsconfig.json`? Ideally we'd be able to do that and keep our compiler options defined in there consistent with here.

packages/tabs/src/Tabs.tsx
DesignerDave: Does toArray return `Array<React.ReactNode>`?

As an aside, it looks like this component doesn't have any specific code preventing a non-Tab ReactElement from being passed in, but there's some assumptions made that it will always be a tab (or have the same props as a Tab) when rendering. Might be worth looking into a bit more at a later time.

package.json
```diff
@@ -115,7 +116,7 @@
     "source-map-loader": "^0.2.4",
     "storybook-addon-performance": "^0.9.0",
     "ts-node": "^8.8.2",
-    "typescript": "^3.9.5",
+    "typescript": "^4.0.1-rc",
```
lazytype, line 119: using 4 so that `"noEmit": true` can be used together with `"incremental": true` (implied by `"composite": true`)

PR #453, opened 2020-06-17
PD-610: Box Component Updates

brookescarlett:
New PR! Thank you @lazytype for all of your help in nailing the typing of our Box component.  Was hoping you could take a look at all of the downstream changes I made as a result of our updates to Box itself.

## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-610](https://jira.mongodb.org/browse/PD-610)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -307,6 +307,7 @@ function OrgNav({
                 usePortal={false}
               >
                 <MenuItem
+                  as={orgNav.accessManager ? 'a' : 'button'}
```
lazytype, line 310: I'm surprised this typechecks since from TS's point of view we might be giving an "href" to a "button". is there a chance that `MenuItem`'s typing is not catching error cases?

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
```diff
@@ -54,26 +54,20 @@ describe('packages/mongo-nav/user-menu', () => {
...
-    const university = document.querySelectorAll(
-      '[data-leafygreen-ui="sub-menu-container"]',
-    )[1];
-
-    const universityArrowButton = university?.parentNode?.querySelector(
-      'button',
-    );
+    const universityTrigger = getAllByRole('button', {
+      name: 'Open Sub-menu',
+    })[1];
```
lazytype, line 65: I think this test is failing. I could see it being more stable if the button was selected based on its text content.

brookescarlett: Great call, fixed! Thank you!

packages/card/src/Card.tsx
```diff
@@ -18,16 +17,20 @@ const containerStyle = css`
...
+interface CardProps {
+  className?: string;
 }
 
+const Card: ExtendableBox<CardProps> = ({ className, ...rest }: CardProps) => {
+  return <Box className={cx(containerStyle, className)} {...rest} />;
+};
+
+// @ts-ignore
 Card.displayName = 'Card';
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/mongo-nav/src/helpers/OnPremUserMenu.tsx
```diff
@@ -87,7 +87,7 @@ export default function OnPremUserMenu({
 
       <Menu open={open} setOpen={setOpen} usePortal={false}>
         <MenuItem
-          href={urls.onPrem.profile}
+          href={urls.onPrem.profile as string}
```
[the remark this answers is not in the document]
brookescarlett: Yes ! can definitely be used here. I need to look into whether or not we're sure about having access to values for those hrefs by the time were in this component (hrefs coming from MongoNav). If we are, I need to change the typing of the urls prop in this component and that should fix things.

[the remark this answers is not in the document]
brookescarlett: Confirming, updated the type for URLs to solve, because we are sure at this point that they exist.

packages/mongo-nav/src/org-nav/OrgNav.tsx
brookescarlett: Definitely possible... 🤔 But 
```
<MenuItem target="xx" />;
<MenuItem target="xx" as="button" />;
```

Both of these fail as expected

packages/card/src/Card.tsx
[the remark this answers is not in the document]
hswolff: I'd prefer not to incur the (small) runtime hit. Can we we just expect-error instead? I don't want TS to dictate our code that it affects performance.

packages/box/README.md
```diff
@@ -41,10 +41,10 @@ import Box from '@leafygreen-ui/box';
...
-| `href`      | `string`                              | When provided, `<Box />` will render an anchor tag with this `href` value.                                                          |         |
-| `children`  | `React.ReactElement`                  | Content to be rendered in an HTML element, or provided as a prop to the rendered component                                          |         |
+| Prop       | Type                                  | Description                                                                                                                         | Default |
+| ---------- | ------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `as`       | `HTMLElement` or `React.ReactElement` | The component or HTML tag to be rendered by the `<Box />` component. **Note**: This will supersede the behavior of any other props. | `div`   |
```
hswolff, line 46: This I guess will be breaking?

packages/button/src/Button.tsx
```diff
@@ -289,128 +288,94 @@ const disabledStyle = css`
...
+      className: cx(
+        baseStyle,
+        buttonSizes[size],
+        buttonVariants[variant],
+        { [disabledStyle]: disabled },
+        { [focusStyle]: showFocus },
+        className,
+      ),
+      // only add a disabled prop if not an anchor
+      ...(rest.href === undefined && { disabled }),
```
hswolff, line 327: Thoughts on this vs `rest.href == null` to cover both cases?

packages/button/src/Button.tsx
```diff
@@ -289,128 +288,94 @@ const disabledStyle = css`
...
+    );
 
-  if (usesLinkElement(props)) {
-    return renderButton('a');
-  }
+    if (rest.href !== undefined) {
+      return (
+        // we give button a default "as" value based on the `href` prop, if a custom
+        // "as" prop is supplied, it will overwrite this value through {...rest}
+        <Box as="a" {...commonProps} {...rest}>
```
hswolff, line 362: If there's a href won't Box handle setting it to an anchor element?

packages/menu/package.json
```diff
@@ -21,6 +21,7 @@
     "@leafygreen-ui/lib": "^4.4.1",
     "@leafygreen-ui/popover": "^5.0.2",
     "@leafygreen-ui/palette": "^2.0.1",
+    "@leafygreen-ui/box": "1.1.1",
```
hswolff, line 24: Let's always default to ranges so we can enable better hoisting.

```suggestion
    "@leafygreen-ui/box": "^1.1.1",
```

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -341,6 +342,7 @@ function UserMenu({
             })}
           >
             <MenuItem
+              as={userMenu.cloud?.userPreferences ? 'a' : 'button'}
```
hswolff, line 345: It'd be nice if we could do the existence check of `userMenu.cloud?` once so we don't bloat the file size.

```
cloudUrls = userMenu.cloud ?? {}
```

packages/button/src/Button.tsx
brookescarlett: the check is `as ? as : href ? : a : 'div'`

We need to supply an `as` when there is no href, but I suppose you're right and we can omit the explicit as setting on line 362.

packages/menu/package.json
brookescarlett: 👍

packages/button/src/Button.tsx
brookescarlett: WFM 👍

packages/box/README.md
brookescarlett: Yes indeed

brookescarlett: Was holding off on Changesets until we decide impact on downstream components and how to version those

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: Going to hold off on updating this until I spend a little more time playing with URLSInterface, but if I leave it as-is then will implement.

brookescarlett: 👍

packages/button/src/Button.tsx
brookescarlett: NVM! Ts evaluates the common props and sees attributes that don't exist on "div" tags, so it's expecting to receive a value for "as"

brookescarlett:
@hswolff did lots of updates as far as types go in MongoNav so would love another check from you with specific attention to that 🙏

packages/button/src/Button.tsx
DesignerDave: This seems like a core part of the typing of Box that I'd expect to work. It shouldn't care about what a div can receive because href is set.

This might be an issue because based on the conditional – `rest.href !== undefined`. Based on that condition,`href` may still be falsy (or what TS cares about, not a string). If we check `typeof rest.href === 'string'` instead of whether `href` is undefined, that could resolve the issue.

packages/card/README.md
```diff
@@ -23,7 +23,7 @@ import Card from '@leafygreen-ui/card';
 
 <Card
   className='card-styles'
-  component='div'
+  as='div'
```
DesignerDave, line 26: This is the default for Box, right? Can't this be omitted, or changed to a more applicable example like "article"?

packages/card/src/Card.tsx
```diff
@@ -18,16 +17,20 @@ const containerStyle = css`
   }
 `;
 
-function Card<T extends React.ReactNode>(props: BoxProps<T>) {
-  const rest = omit(props as any, ['className']);
-
-  return <Box className={cx(containerStyle, props.className)} {...rest} />;
+interface CardProps {
+  className?: string;
```
DesignerDave, line 21: Maybe I'm missing something, but isn't className also defined as an optional string from the implicit props in ExtendableBox?

packages/menu/src/MenuItem.tsx
```diff
@@ -122,54 +118,34 @@ interface SharedMenuItemProps {
...
+   */
+  className?: string;
 
-function usesCustomElement(props: MenuItemProps): props is CustomMenuItemProps {
-  return (props as any).as != null;
-}
+  /**
+   * Content to appear inside of `<MenuItem />` component
+   */
+  children?: React.ReactNode;
```
DesignerDave, line 129: Ditto here – isn't this included in the types for ExtendableBox?

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -431,7 +431,7 @@ function OrgSelect({
               <MenuSeparator />
               <MenuItem
                 onKeyDown={onKeyDown}
-                href={urls.viewAllOrganizations}
+                href={urls.viewAllOrganizations!}
```
DesignerDave, line 434: Weren't the typing changes in mongo nav explicitly to address the typing issue here?

packages/mongo-nav/src/org-nav/OrgNav.tsx
DesignerDave: Maybe TS is smart enough to understand we're checking that the same value used for href is truthy here?

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
```diff
@@ -54,26 +54,20 @@ describe('packages/mongo-nav/user-menu', () => {
...
-    const university = document.querySelectorAll(
-      '[data-leafygreen-ui="sub-menu-container"]',
-    )[1];
-
-    const universityArrowButton = university?.parentNode?.querySelector(
-      'button',
-    );
+    const universityTrigger = getAllByRole('button', {
+      name: 'Open Sub-menu',
+    })[0];
```
DesignerDave, line 65: This seems brittle – if the university button is no longer the first in the menu, I wouldn't expect a test to fail. I'd rather we use test ids here to avoid this failing for reasons that we aren't trying to test for.

If we wanted to test the order of the menu, we should do that explicitly elsewhere.

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
```diff
@@ -135,15 +129,15 @@ describe('packages/mongo-nav/user-menu', () => {
...
     });
     const trigger = getByTestId('user-menu-trigger');
     fireEvent.click(trigger);
-    const accountButton = getByText('Manage your MongoDB Account')
-      .parentElement as HTMLButtonElement;
+    const accountButton = getByTestId('user-menu-account-button');
+
     expect(accountButton.tagName.toLowerCase()).toBe('button');
-    expect(accountButton.disabled).toBe(true);
+    expect((accountButton as HTMLButtonElement).disabled).toBe(true);
```
DesignerDave, line 140: Why are we typecasting here? It's probably desirable to do something like this if we don't know that accountButton will be defined or contain the disabled attribute:

```js
expect(accountButton?.disabled).toBe(true)
```

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -260,8 +260,10 @@ function UserMenu({
     },
   };
 
-  const urls: URLSInterface = defaultsDeep(urlsProp, defaultURLs);
+  const urls: Required<URLSDefaults> = defaultsDeep(urlsProp, defaultURLs);
   const userMenu = urls.userMenu ?? {};
+  const cloudUrls: Partial<UserMenuURLS['userMenu']['cloud']> =
+    userMenu.cloud ?? {};
```
DesignerDave, line 266: Wont userMenu.cloud always be defined now?

packages/card/README.md
brookescarlett: Yes, good point!

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
brookescarlett: 👍

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
brookescarlett: We don't have a testId for this button

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
brookescarlett: We're typecasting bc accountButton is inferred as HTMLElement not HTMLButtonElement and HTMLElement doesn't have disabled as an attribute

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: We can't assume that from the outset, no

DesignerDave: Looking at the implementation on line `:263`, it looks otherwise. defaultsDeep ensures that every value in the urls object is defined, doesn't it? I might be missing something here...

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
DesignerDave: Could we type the `accountButton` const instead of typecasting here?

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
DesignerDave: Could we add one?

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: UserMenu should be safe to use outside of the Mongonav context

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
brookescarlett: Yes we can!

packages/mongo-nav/src/user-menu/UserMenu.tsx
DesignerDave: Aren't we applying the defaults either way? Line 263 appears to apply default values regardless of context.

packages/card/src/Card.tsx
brookescarlett: Chatted offline!

packages/menu/src/MenuItem.tsx
brookescarlett: Chatted offline!

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
brookescarlett: Actually, not as straightforward as I would've thought, because I dont think that we have glyph components set up to spread props. I can include that change in this PR though, if you would like.

DesignerDave: Looks like it should be spreading props actually!

packages/button/src/Button.tsx
brookescarlett: TS isn't understanding the check on line 358, I'm not sure why. If you inline href={rest.href} it works without defining an as..

brookescarlett: Screen Shot 2020-06-23 at 5 07 30 PM

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
brookescarlett: yes that works for sure

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: yep you're right

brookescarlett: fixed

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -206,7 +206,7 @@ function MongoNav({
   const currentOrgId = data?.currentOrganization?.orgId;
   const currentProjectId = data?.currentProject?.projectId;
 
-  const defaultURLS: Required<Omit<URLSInterface, 'userMenu'>> = {
+  const defaultURLS: Required<Omit<URLSDefaults, 'userMenu'>> = {
```
hswolff, line 209: Do we still need `Required` here? Seems like that's now supported via `URLSDefaults`. Ditto below too.

packages/mongo-nav/src/data.ts
```diff
@@ -58,7 +58,7 @@ export const dataFixtures = {
   ],
 } as DataInterface;
 
-export const urlFixtures: Required<URLSInterface> = {
+export const urlFixtures: Required<URLSDefaults> = {
```
hswolff, line 61: ditto required usage

packages/mongo-nav/src/helpers/OnPremUserMenu.tsx
```diff
@@ -31,7 +31,7 @@ interface OnPremUserMenuProps {
    * Object that supplies URL overrides to OnPremUserMenu component.
    * Shape: { onPrem: { profile, mfa, personalization, invitations, organizations, publicApiAccess, featureRequest }}
    */
-  urls: Pick<Required<URLSInterface>, 'onPrem'>;
+  urls: Pick<Required<URLSDefaults>, 'onPrem'>;
```
hswolff, line 34: ditto here

packages/mongo-nav/src/mongo-select/types.ts
```diff
@@ -1,7 +1,7 @@
 import { MouseEventHandler, ReactNode } from 'react';
-import { MongoNavInterface, URLSInterface, OnChangeInterface } from '../types';
+import { MongoNavInterface, URLSDefaults, OnChangeInterface } from '../types';
 
-export type MongoSelectUrls = NonNullable<URLSInterface['mongoSelect']>;
+export type MongoSelectUrls = NonNullable<URLSDefaults['mongoSelect']>;
```
hswolff, line 4: do we need non nullable?

packages/mongo-nav/src/types.ts
```diff
@@ -220,60 +220,78 @@ export interface OnPremInterface {
...
+
+export interface URLSProp {
+  userMenu?: UserMenuURLS;
+  mongoSelect?: MongoSelectURLS;
+  orgNav?: OrgNavURLS;
+  projectNav?: ProjectNavURLS;
+  onPrem?: OnPremURLS;
+}
+
+export interface URLSDefaults {
```
hswolff, line 289: Defaults implies there's values associated with it. Also we use this everywhere internally...I wonder if we should just call it `URLS`.

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -260,8 +260,9 @@ function UserMenu({
     },
   };
 
-  const urls: URLSInterface = defaultsDeep(urlsProp, defaultURLs);
+  const urls: Required<URLSDefaults> = defaultsDeep(urlsProp, defaultURLs);
   const userMenu = urls.userMenu ?? {};
+  const cloudUrls: Partial<UserMenuURLS['userMenu']['cloud']> = userMenu.cloud;
```
hswolff, line 265: Lets us remove the existence checks below.

```suggestion
  const cloudUrls: Partial<UserMenuURLS['userMenu']['cloud']> = userMenu.cloud ?? {};
```

packages/menu/src/Menu.spec.tsx
```diff
@@ -90,6 +90,11 @@ describe('packages/menu/menu-item', () => {
     expect(menuItem.classList.contains(className)).toBe(true);
   });
 
+  test('renders MenuItem inside button tag by default', () => {
```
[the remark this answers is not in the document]
brookescarlett: This was always the default behavior for MenuItem but I can include something like this in the changelog still

packages/button/src/Button.tsx
brookescarlett: 👍

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
brookescarlett: Discussed offline: props weren't spreading in the template so I updated, but that's actually not the problem in this case. In order to add a test-id we'd have to update SubMenu itself, which would also require relying on an order of the submenus. I actually don't think this is as brittle as you think, because we wouldn't change the order of the items in the UserMenu without updating the tests but happy to discuss any alternative thoughts that you have.

packages/menu/src/Menu.spec.tsx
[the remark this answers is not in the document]
hswolff: ah, nvm then!

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: 👍

packages/mongo-nav/src/data.ts
brookescarlett: 👍

packages/mongo-nav/src/helpers/OnPremUserMenu.tsx
brookescarlett: 👍

packages/mongo-nav/src/types.ts
brookescarlett: kk

packages/mongo-nav/src/mongo-select/types.ts
brookescarlett: yes i believe we still do

packages/mongo-nav/src/types.ts
brookescarlett: updated

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: Took me an embarrassingly long time to find a typing that would support it but arrived at a very simple conclusion 😥

packages/menu/src/Menu.spec.tsx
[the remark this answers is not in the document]
brookescarlett: Will add the changesets now

.changeset/many-rats-tease.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/card': major
+'@leafygreen-ui/typography': major
+---
+
+Renames "component" prop to "as"
```
hswolff, line 6:
```suggestion
Renames `component` prop to `as`
```

As it's code related.

.changeset/metal-shrimps-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/box': major
+---
+
+Box component now accepts "as" prop instead of "component" and exports generic ExtendableBox as helper type for components wrapping Box
```
hswolff, line 5:
```suggestion
Box component now accepts `as` prop instead of `component` and exports generic `ExtendableBox` as helper type for components wrapping Box.

Example usage of `ExtendableBox`: ...
```

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -246,7 +246,7 @@ function MongoNav({
     },
   };
 
-  const urls: Required<URLSInterface> = defaultsDeep(urlsProp, defaultURLS);
+  const urls: Required<URLS> = defaultsDeep(urlsProp, defaultURLS);
```
hswolff, line 249: Is required still required here?

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -251,7 +252,7 @@ type ProjectNavProps = Pick<
   onPremMFA?: boolean;
   alerts?: number;
   constructProjectURL: NonNullable<MongoNavInterface['constructProjectURL']>;
-  urls: Required<NonNullable<MongoNavInterface['urls']>>;
+  urls: Required<URLS>;
```
hswolff, line 255: Required still needed?

.changeset/many-rats-tease.md
brookescarlett: 👍

.changeset/metal-shrimps-yawn.md
brookescarlett: 👍

packages/mongo-nav/src/MongoNav.tsx
brookescarlett: 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
brookescarlett: 👍

PR #462, opened 2020-06-26
PD-646: Conditionally render label tag on TextInput

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-646](https://jira.mongodb.org/browse/PD-646)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/text-input/src/TextInput.tsx
+++ b/packages/text-input/src/TextInput.tsx
@@ -295,9 +295,11 @@ const TextInput = React.forwardRef(
 
     return (
       <div className={cx(textInputStyle, className)}>
-        <label htmlFor={id} className={labelStyle}>
-          {label}
-        </label>
+        {label && (
+          <label htmlFor={id} className={labelStyle}>
+            {label}
+          </label>
+        )}
         {description && <p className={descriptionStyle}>{description}</p>}
         <div className={inputContainerStyle}>
           <input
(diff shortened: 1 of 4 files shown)


packages/text-input/src/TextInput.tsx
```diff
@@ -295,9 +295,11 @@ const TextInput = React.forwardRef(
 
     return (
       <div className={cx(textInputStyle, className)}>
-        <label htmlFor={id} className={labelStyle}>
-          {label}
-        </label>
+        {label && (
```
DesignerDave, line 298: Hmm.... I wonder if there's a way to enforce there being some label for the input. Maybe require `aria-labelledby` if label isn't present?

brookescarlett:
Good call, @DesignerDave - followed pattern from IconButton

packages/text-input/src/TextInput.tsx
```diff
@@ -258,16 +261,17 @@ function getStatefulInputStyles(state: State, optional: boolean) {
...
     {
-      id: propsId,
       label,
       description,
-      optional = false,
-      disabled = false,
       onChange,
       placeholder,
       errorMessage,
+      optional = false,
```
brookescarlett, line 269: Reorg'd the props, bc I like to have to aliased props grouped together as well as the props that we provide a default value for grouped together

packages/text-input/src/TextInput.tsx
brookescarlett: Great call -- updated this.

PR #465, opened 2020-06-29
Add Stepper component

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Stepper React Component](https://jira.mongodb.org/browse/PD-694)

## 🛠 Types of changes

## ✅ Checklist


packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+    >
+      {Object.entries(layerToLayerSteps).map(([layer, layerSteps]) => (
+        <div
+          key={layer}
+          className={cx(layerStyle, {
+            [hiddenLayerStyle]: visibleLayer !== layer,
+          })}
+        >
+          {layerSteps.map((props, index) => (
+            <InternalStep key={index} {...props}></InternalStep>
```
brookescarlett, line 491: Can this be self-closing

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+
+const upcomingStepLabelStyle = css`
+  background-color: white;
+  border-color: ${uiColors.gray.light1};
+  color: ${uiColors.gray.base};
+`;
+
+const stepLabelTextStyle = css('padding: 0 8px;');
+
+function InternalStep({
```
brookescarlett, line 235: I like this pattern, not cloning Step itself but creating an Internal-only component that handles all of the added functionality. Will be applying this elsewhere 😏

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+        { [currentStepStyle]: isCurrent },
+        className,
+      )}
+      {...containerProps}
+    >
+      {stepLabel && (
+        <div
+          className={cx(
+            stepLabelStyle,
+            state === 'current'
```
brookescarlett, line 259: Might be cleaner to use `isCurrent` here since you've already created the variable

brookescarlett: I also think we should be consistent, either conditionally applying styles with the object notation as you did above, or through a nested ternary as you're doing here.

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+  ...containerProps
+}: InternalStepProps) {
+  const isCurrent = state === 'current';
+  const isUpcoming = state === 'upcoming';
+  const isCompleted = state === 'completed';
+  return (
+    <div
+      className={cx(
+        stepStyle,
+        { [upcomingStepStyle]: isUpcoming },
```
brookescarlett, line 249: I think this can be collapsed into one object

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+          [slideInAnimationStyle]: visibleLayer === layer,
+          [slideOutAnimationStyle]: visibleLayer !== layer,
+        });
+      }
+    });
+  });
+
+  return (
+    <div
+      aria-label="breadcrumbs"
```
brookescarlett, line 479: Have been trying to find documentation to support this label, have you seen any? Open to it if it makes sense, but not sure that it does here

packages/stepper/src/index.ts
```diff
@@ -0,0 +1 @@
+export { default, Step } from './Stepper';
```
hswolff, line 1: My stance on compound components is that they should all be named exports.

My opinion (so take with a grain of salt) is that a default export should be expected to work by itself. 

That's not the case here, these components need to work together, hence they should all be named exports.

Happy to discuss more!

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+          <div />
+        </span>
+      </div>
+    </div>
+  );
+}
+
+InternalStep.displayName = 'Step';
+
+type StepContents = string | number | Array<StepContents>;
```
hswolff, line 288: Feels like we should break this into two types, `StepContent` and `StepContents` for the array instance. Otherwise as you have the `Step` component typed now doing `<Step>{[1,2,3]}</Step>` is allowed.

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+  &:before,
+  &:after {
+    display: inline-block;
+    height: 100%;
+    background: white;
+    border: 1px solid #dee0e3;
+  }
+
+  & > * {
+    width: calc(100% - ${46 / 3}px);
```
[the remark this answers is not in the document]
hswolff: Also how did we come to these magic numbers?

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+
+  & > * {
+    border-bottom: none;
+  }
+
+  &:before,
+  &:after {
+    border-bottom: none;
+    transform-origin: 50% 0%;
+    transform: skewX(${Math.atan(2 / 3)}rad);
```
hswolff, line 87: 🤯  atan??

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+  justify-content: center;
+  height: 100%;
+  width: 100%;
+  margin-left: -${46 / 3 + 1}px;
+
+  color: ${uiColors.gray.dark3};
+
+  font-family: Akzidenz, 'Helvetica Neue', Helvetica, Arial, sans-serif;
+  font-size: 14px;
+  font-weight: bold;
```
hswolff, line 119: Can't wait till we standardize these values on some common context.

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+  color: white;
+`;
+
+const upcomingStepLabelStyle = css`
+  background-color: white;
+  border-color: ${uiColors.gray.light1};
+  color: ${uiColors.gray.base};
+`;
+
+const stepLabelTextStyle = css('padding: 0 8px;');
```
hswolff, line 233: So many styles. Feels like we're bordering on needing to put them into their own file for clarity sake. Up to you.

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+function stepContentsAriaLabel(contents: StepContents): string {
+  return contents instanceof Array
+    ? contents.map(content => stepContentsAriaLabel(content)).join('')
+    : contents.toString();
+}
+
+type StepElement = React.ReactElement<{ children: StepContents }>;
+export type StepElements = StepElement | Array<StepElement>;
+
+interface StepperProps {
```
hswolff, line 299: super nit, can we put this right above the Stepper component so they're closer together which makes it easier for me to read?

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+export default function Stepper({
+  children,
+  currentStep,
+  maxDisplayedSteps,
+}: StepperProps) {
+  const [visibleLayer, setVisibleLayer] = useState<
+    'previous' | 'next' | 'current'
+  >('current');
+
+  const allSteps: Array<StepItem> = (children instanceof Array
```
hswolff, line 393: Why not `Array.isArray`?

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+    : [children]
+  ).map((child, index) => {
+    let state: StepCompletionState;
+
+    if (index < currentStep) {
+      state = 'completed';
+    } else if (index === currentStep) {
+      state = 'current';
+    } else {
+      state = 'upcoming';
```
hswolff, line 404: THE FUTURE

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,521 @@
...
+type StepElement = React.ReactElement<{ children: StepContents }>;
+export type StepElements = StepElement | Array<StepElement>;
+
+interface StepperProps {
+  children: StepElements;
+  currentStep: number;
+  maxDisplayedSteps: number;
+}
+
+function computeRangeToDisplay({
```
hswolff, line 305: Easy to unit test if we desire.

packages/stepper/src/Stepper.tsx
lazytype: Yep, that's by design. For example, `<Step>Step {index}<Step>` is equivalent to `<Step>{['Step ', 1]}</Step>`. This lends to slightly more convenience than needing to do ``<Step>{`Step ${index}`}<Step>``

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,319 @@
...
+interface StepperProps {
+  children: StepElements;
+  currentStep: number;
+  maxDisplayedSteps: number;
+}
+
+export function Stepper({
+  children,
+  currentStep,
+  maxDisplayedSteps,
```
hswolff, line 103: Can we make this prop optional and default to step.length?

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,319 @@
...
+}: StepperProps) {
+  const [visibleLayer, setVisibleLayer] = useState<
+    'previous' | 'next' | 'current'
+  >('current');
+
+  const allSteps: Array<StepItem> = (Array.isArray(children)
```
hswolff, line 109: I think I know the answer but just want to double check - wrapping this function inside a `useMemo(fn, [children])` won't give us great value as children will, by default, be a new instance?

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+interface StepperProps {
+  children: StepElements;
+  currentStep: number;
+  maxDisplayedSteps: number;
+}
+
+export function Stepper({
+  children,
+  currentStep,
```
hswolff, line 102: 0 indexed or 1 indexed?

Which do we want?

packages/stepper/src/Stepper.tsx
```diff
@@ -0,0 +1,319 @@
...
+        </span>
+        <span className={stepBoxBottomStyle}>
+          <div />
+        </span>
+      </div>
+    </div>
+  );
+}
+
+InternalStep.displayName = 'Step';
```
hswolff, line 83: sneaky

packages/stepper/src/Stepper.spec.tsx
```diff
@@ -0,0 +1,272 @@
...
+  expectedSteps?: Array<string>;
+}) {
+  const visiblePreviousStepsElement = onlyElement(
+    screen.queryAllByLabelText('Previous steps').filter(isVisible),
+  );
+
+  if (expectedSteps.includes('Previous steps')) {
+    expect(visiblePreviousStepsElement).not.toBeNull();
+  } else {
+    expect(visiblePreviousStepsElement).toBeNull();
```
hswolff, line 59: Generally I prefer `toBeInDocument` as the error messages when it's incorrect is more rich. Can we use that instead?

packages/stepper/src/Stepper.tsx
lazytype: I think it makes more sense to be 0-indexed since `currentStep` will probably be managed programmatically by the consumer

packages/stepper/src/Stepper.tsx
lazytype: the function also uses `currentStep`, so the only prop change that shouldn't result in a re-evaluation is `maxDisplaySteps`, but I would generally expect it to not change.

PR #471, opened 2020-07-09
Clean up mongo-nav tests and remove testHelpers

lazytype:
testHelpers.ts only exposes types which aren't necessary. Only the mongo-nav tests seems to be using them, so I removed usage of them + I refactored the tests to not depend on mutating non-local state 🤢. In addition I fixed up some typing issues here and there.

## ✍️ Proposed changes

🎟 _Jira ticket:_ [Enable code sharing between tests](https://jira.mongodb.org/browse/PD-723)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -92,63 +56,64 @@ describe('packages/mongo-nav', () => {
...
     test('the organization and project navs are rendered', () => {
-      expect(expectedElements.orgNav).toBeInTheDocument();
-      expect(expectedElements.projectNav).toBeInTheDocument();
+      expect(queryByTestId('organization-nav')).not.toBeNull();
+      expect(queryByTestId('project-nav')).not.toBeNull();
     });
 
     test('current orgId is set based on data returned from fetch', () => {
-      expect((expectedElements.billing as HTMLAnchorElement).href).toBe(
+      expect((getByTestId('org-nav-billing') as HTMLAnchorElement).href).toBe(
```
hswolff, line 64: I'm curious if we make a page object that collects all of these string id's for a clearer representation of the parts of the nav.

```jsx
const MongoNavElements = {
   OrgNav: {
       Billing: 'org-nav-billing'
   }
}
```

and etc?

I'm a little on the fence as it seems like adding indirection, but we are repeating ourselves a lot. Also having a canonical list of all id references seems pleasant as well.

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -92,63 +56,64 @@ describe('packages/mongo-nav', () => {
     });
 
     test('the organization and project navs are rendered', () => {
-      expect(expectedElements.orgNav).toBeInTheDocument();
-      expect(expectedElements.projectNav).toBeInTheDocument();
+      expect(queryByTestId('organization-nav')).not.toBeNull();
```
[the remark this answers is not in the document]
lazytype: Since now we're querying the element, it must be in the document if it's not `null`. Before the element could've been stale and no longer in the document. Same rationale as https://github.com/mongodb/leafygreen-ui/pull/465/files#r455182402.

packages/mongo-nav/src/MongoNav.spec.tsx
lazytype: would we want something like this to also be used in the components themselves? e.g. `data-testid={MongoNavElements.OrgNav.Billing}`. any thoughts on where to locate these, e.g. separate file, same file as component

packages/mongo-nav/src/MongoNav.spec.tsx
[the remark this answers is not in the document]
hswolff: What's the difference in log if we made this intentionally fail?

```jsx
expect(queryByTestId('organization-nav')).toBeNull();
// vs
expect(queryByTestId('organization-nav')).toBeInTheDocument();
```

Also what does the query log error show when it errors?

packages/mongo-nav/src/MongoNav.spec.tsx
hswolff: I don't feel as strongly about collecting all source `data-testid` values as they're only used once.

The obvious benefit to collect them is it becomes easily reusable for tests.

For MongoNav it may be useful to consolidate them all as it's so massive.

packages/mongo-nav/src/MongoNav.spec.tsx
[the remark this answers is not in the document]
lazytype: the former gives us
```
expect(received).not.toBeNull()

Received: null

      58 |     test.only('the organization and project navs are rendered', () => {
    > 59 |       expect(queryByTestId('doesnt-exist')).not.toBeNull();
```
the latter gives us
```
expect(received).toBeInTheDocument()

    received value must be an HTMLElement or an SVGElement.
    Received has value: null

      58 |     test.only('the organization and project navs are rendered', () => {
    > 59 |       expect(queryByTestId('doesnt-exist')).toBeInTheDocument();
```

The latter adds some extra context about the expected type, but since you can see the line it failed on, it should be pretty obvious either way what we were expecting.

packages/mongo-nav/src/MongoNav.spec.tsx
lazytype: I think I want to hold off on his for now, just to reduce the scope of the change

hswolff: WFM

packages/mongo-nav/src/MongoNav.spec.tsx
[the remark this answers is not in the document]
hswolff: Yeah, the latter definitely seems more accurate to me. We want to make sure that it's in the document and visible (which we're not strictly testing for visibility, but in document is closer than just not null).

packages/stepper/src/Stepper.spec.tsx
```diff
@@ -54,7 +58,7 @@ function assertVisibleSteps({
   );
 
   if (expectedSteps.includes('Previous steps')) {
-    expect(visiblePreviousStepsElement).not.toBeNull();
+    expect(visiblePreviousStepsElement).toBeVisible();
```
brookescarlett, line 61: Just out of curiosity, why the change?

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -92,63 +61,64 @@ describe('packages/mongo-nav', () => {
     });
 
     test('the organization and project navs are rendered', () => {
-      expect(expectedElements.orgNav).toBeInTheDocument();
-      expect(expectedElements.projectNav).toBeInTheDocument();
+      expect(queryByTestId('organization-nav')).toBeVisible();
```
brookescarlett, line 64: again, just out of curiosity, why are we changing the test from `toBeInTheDocument()` to `toBeVisible()`

packages/stepper/src/Stepper.spec.tsx
lazytype: Making this consistent everywhere based on the discussion in this thread
https://github.com/mongodb/leafygreen-ui/pull/471/files/840f8c5dd7dc762382d69676dbdad60b74a1000e#diff-832b8f9837f5879d08b9fddbcc69af38

brookescarlett: Can't see that thread gh is erroring :/

packages/mongo-nav/src/MongoNav.spec.tsx
lazytype: Also for consistency and I prefer `toBeVisible` since it's stricter than `toBeInTheDocument`

packages/stepper/src/Stepper.spec.tsx
lazytype: oops, updated the link

PR #472, opened 2020-07-10
PD-263: Add Link Component

brookescarlett:
## ✍️ Proposed changes

Adds link component to typography package

🎟 _Jira ticket:_ [PD-263](https://jira.mongodb.org/browse/PD-263)

## 🛠 Types of changes

## ✅ Checklist


.changeset/rude-penguins-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
```
lazytype, line 1: looks like two changesets were created?

packages/typography/src/Typography.story.tsx
```diff
@@ -24,6 +33,19 @@ storiesOf('Typography', module).add('Default', () => (
       <InlineCode className={displayBlock}>InlineCode</InlineCode>
       <Disclaimer className={displayBlock}>Disclaimer</Disclaimer>
       <Overline>Overline</Overline>
+      <Link
+        href={select(
+          'href',
+          [
+            'http://localohst:9001',
```
lazytype, line 40: typo(?): localohst -> localhost

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +137,138 @@ const Overline: ExtendableBox<{
...
+`;
+
+const arrowRightIcon = css`
+  height: 10px;
+  width: 10px;
+  opacity: 0;
+  transform: translate3d(-3px, 0, 0px);
+  transition: all 100ms ease-in;
+  ${anchorDataProp.selector}:hover &,
+  ${anchorDataProp.selector}:hover & {
```
lazytype, line 211: dupe?

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +137,138 @@ const Overline: ExtendableBox<{
...
+      background-color: #9dd0e7;
+    }
+  }
+`;
+
+const arrowRightIcon = css`
+  height: 10px;
+  width: 10px;
+  opacity: 0;
+  transform: translate3d(-3px, 0, 0px);
```
lazytype, line 208: would "0" work in lieu of "0px"?

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +137,138 @@ const Overline: ExtendableBox<{
...
+  target: targetProp,
+  showArrow = false,
+  ...rest
+}: LinkProps) => {
+  if (!href) {
+    console.error(
+      'Link components are wrapped anchor tags. Please provide a destination URL through the `href` prop or consider using another component.',
+    );
+    return null;
+  }
```
lazytype, line 234: could we enforce this in LinkProps?

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +137,138 @@ const Overline: ExtendableBox<{
...
+      target = '_self';
+    } else {
+      target = '_blank';
+    }
+  }
+
+  const icon =
+    target === '_blank' ? (
+      <OpenInNewTab />
+    ) : showArrow ? (
```
lazytype, line 256: it might be confusing that `showArrow` being set to `true` doesn't guarantee that the arrow is shown. is there another name we could use to more clearly indicate that, e.g. `canShowArrow`

.changeset/rude-penguins-wonder.md
brookescarlett: Good catch, thank you

packages/typography/src/Typography.story.tsx
brookescarlett: 👍

packages/typography/src/Typography.tsx
brookescarlett: 👍

packages/typography/src/Typography.tsx
brookescarlett: yep!

packages/typography/src/Typography.tsx
brookescarlett: 👍

brookescarlett: Should I remove this check here then or leave both?

packages/typography/src/Typography.tsx
brookescarlett: yes good call

brookescarlett: the design specs say that the external icon should override this one

packages/typography/src/Typography.tsx
lazytype: I think it can be removed since `propTypes` should give a warning

brookescarlett: Makes sense!

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +138,166 @@ const Overline: ExtendableBox<{
...
+
+const linkStyles = css`
+  text-decoration: none;
+  color: ${uiColors.blue.base};
+  cursor: pointer;
+
+  span {
+    position: relative;
+
+    &:after {
```
DesignerDave, line 172: This approach worked great for the MongoNav, but unfortunately, this will break sometimes when being used in traditional body copy if the link text wraps.

If you'd like to see the rabbit-hole this problem sent medium down, here's a good read for ya: https://medium.design/crafting-link-underlines-on-medium-7c03a9274f9

I think we'll need to get clever here...

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +138,153 @@ const Overline: ExtendableBox<{
...
+        size={10}
+        className={cx({
+          [arrowRightIconHover]: canShowArrow === CanShowArrow.Hover,
+          [arrowRightIconPermanent]: canShowArrow === CanShowArrow.Permanent,
+        })}
+      />
+    ) : null;
+
+  return (
+    <a
```
hswolff, line 270: Ideally we support the `as` prop so we can let a user use react-router's Link component for proper SPA routing: `<Link as={ReactRouterLink} href="foo" />`

brookescarlett: Ok!

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +137,145 @@ const Overline: ExtendableBox<{
   return <Box className={cx(sharedStyles, overline, className)} {...rest} />;
 };
 
-export { H1, H2, Subtitle, Body, InlineCode, Disclaimer, Overline };
+const OpenInNewTab = (props: JSX.IntrinsicElements['svg']) => {
```
hswolff, line 140: Curious if it'd be better to make a standalone `A.tsx` file or `AnchorElement`?

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +137,145 @@ const Overline: ExtendableBox<{
...
+  ...rest
+}: LinkProps) => {
+  if (!href) {
+    return null;
+  }
+
+  const size = useBaseFontSize();
+  const fontSize = size === 16 ? typeScale2 : typeScale1;
+
+  const hrefHostname = new URL(href).hostname;
```
hswolff, line 240: Maybe `useMemo` this value?

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +137,145 @@ const Overline: ExtendableBox<{
...
+  const currentHostname = window.location.hostname;
```
hswolff, line 241: Not server side friendly.  Not sure we have a standard way of being server side friendly yet.

packages/typography/src/Typography.tsx
```diff
@@ -136,4 +137,145 @@ const Overline: ExtendableBox<{
...
+  }
+`;
+
+const openInNewTabStyles = css`
+  margin-bottom: -1px;
+  margin-left: -1px;
+  margin-right: -2px;
+`;
+
+const CanShowArrow = {
```
hswolff, line 214: Perhaps: `ArrowAppearance = 'none' | 'hover' | 'persist'`?

packages/typography/src/Typography.tsx
brookescarlett: Kk!

packages/typography/src/Typography.tsx
brookescarlett: 👍

packages/typography/src/Typography.tsx
brookescarlett: Can we keep this logic but either expose a prop or default to an empty string for graceful degradation in the case where we don't have `currentHostname`

packages/typography/src/Typography.tsx
brookescarlett: I like this better, thank you.

packages/typography/src/Typography.tsx
brookescarlett: @DesignerDave thoughts?

DesignerDave: We have a build variable we can use in a conditional to only use the window when the target is a browser. Check out our emotion package to see how this is done there.

packages/typography/README.md
```diff
@@ -76,3 +95,11 @@ All props extend the HTMLElementProps of their root tag, however `<Body />` and
...
+
+## Link
+
+| Prop              | Type                           | Description                                                                                                                                                                              | Default |
+| ----------------- | ------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `arrowAppearance` | `'hover' | 'persist' | 'none'` | Displays a right arrow adjacent to the anchor tag. When set to `persist` the arrow will always be present. When set to `hover`, the arrow will only appear when hovering over the arrow. | `none`  |
+
+_Note:_ If the current host name is different from the host of the destination URL, we will add provide the "\_blank" value for the `target` prop. When the target is set to open in a new tab, we render an icon as a visual affordance.
```
hswolff, line 105:
```suggestion
_Note:_ If the current host name is different from the host of the destination URL, we will provide the "\_blank" value for the `target` prop. When the target is set to open in a new tab, we render an icon as a visual affordance.
```

packages/typography/src/index.ts
```diff
@@ -7,3 +7,6 @@ export {
   Disclaimer,
   Overline,
 } from './Typography';
+export { Link } from './Link';
+
+export declare const __TARGET__: 'web' | 'node';
```
hswolff, line 12: Why in here? It's only being used inside `Link`?

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,164 @@
...
+        <path
+          d="M9.29411765,3 L10.7727647,4.47864706 L6.42788235,8.82352941 L7.17647059,9.57211765 L11.5218824,5.22776471 L13,6.70588235 L13,3 L9.29411765,3 Z M4,4.58823529 L4,12 L11.4122941,12 L11.4122941,7.76470588 L10.3529412,7.76470588 L10.3529412,10.9411765 L5.05882353,10.9411765 L5.05882353,5.64705882 L8.23529412,5.64705882 L8.23529412,4.58823529 L4,4.58823529 Z"
+          fill="#007CAD"
+        />
+      </g>
+    </svg>
+  );
+};
+
+// declare const __TARGET__: 'web' | 'node';
```
hswolff, line 85: rm?

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+function isServer() {
+  if (__TARGET__ === 'node') {
+    return true;
+  }
+
+  return false;
+}
+
```
hswolff, line 94:
```jsx
function isServer() {
  return __TARGET__ === 'node';
}
```

packages/typography/src/index.ts
brookescarlett: Was going to use in spec, but then I didn't. Relocating!

packages/typography/src/Link.tsx
brookescarlett: 👍

packages/typography/src/Link.tsx
brookescarlett: Re-added

packages/typography/src/index.ts
brookescarlett: NVM. I had this in index.ts because I'm getting an error that `__TARGET__` is not defined when I declare it in a `.tsx` file, it only works for me when declared in a `.ts` file

packages/typography/src/Link.tsx
brookescarlett: Removed!

packages/typography/README.md
brookescarlett: 👍

packages/typography/src/Typography.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+        const openInNewTab = screen.getByTitle('Open in New Tab');
+        const rightArrow = screen.queryByTitle('Glyphs / Arrow / Right');
+        expect(openInNewTab).toBeInTheDocument();
+        expect(rightArrow).not.toBeInTheDocument();
+      });
+
+      test('and the "arrowAppearance" prop is not set', () => {
+        renderLink({ href: 'http://mongodb.design' });
+        const openInNewTab = screen.getByTitle('Open in New Tab');
+        expect(openInNewTab).toBeInTheDocument();
```
hswolff, line 23: What's rightArrow behavior?

packages/typography/src/Typography.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    describe('when the current host is the same as the destination URL host', () => {
+      describe('by default', () => {
+        test('it renders the correct tag to the DOM', () => {
+          renderLink({ href: 'http://localhost:9001' });
+          const anchor = screen.getByText('Link').parentNode;
+          expect((anchor as HTMLElement).tagName.toLowerCase()).toBe('a');
+        });
+      });
+      test('and the "arrowAppearance" prop is set to "hover"', () => {
```
hswolff, line 35: nit, line break

brookescarlett: Fixed!

packages/typography/src/Typography.spec.tsx
brookescarlett: Added!

packages/typography/src/Typography.story.tsx
```diff
@@ -24,6 +33,23 @@ storiesOf('Typography', module).add('Default', () => (
...
+          'href',
+          [
+            'http://localhost:9001',
+            'https://mongodb.github.io/leafygreen-ui/?path=/story/*',
+          ],
+          'http://localhost:9001',
+        )}
+        arrowAppearance={select(
+          'arrowAppearance',
+          ['hover', 'persistant' | 'none'],
```
DesignerDave, line 47: This probably should be `['hover', 'persistent', 'none']`. Right now the options in storybook are `'hover'` and `0` because of this.

packages/typography/src/Typography.story.tsx
```diff
@@ -24,6 +33,23 @@ storiesOf('Typography', module).add('Default', () => (
...
       <Disclaimer className={displayBlock}>Disclaimer</Disclaimer>
       <Overline>Overline</Overline>
+      <Link
+        href={select(
+          'href',
+          [
+            'http://localhost:9001',
+            'https://mongodb.github.io/leafygreen-ui/?path=/story/*',
+          ],
+          'http://localhost:9001',
```
DesignerDave, line 43: Hmm... This is fun 😄 

Screen Shot 2020-07-20 at 2 08 30 PM

It's not infinite FYI, but some iFrame fun regardless!

packages/typography/src/Typography.story.tsx
```diff
@@ -24,6 +33,23 @@ storiesOf('Typography', module).add('Default', () => (
       <InlineCode className={displayBlock}>InlineCode</InlineCode>
       <Disclaimer className={displayBlock}>Disclaimer</Disclaimer>
       <Overline>Overline</Overline>
+      <Link
+        href={select(
+          'href',
```
DesignerDave, line 38: We should indicate which typography element each of these knobs apply to. Right now, it's hard to tell which element you're changing props for.

Even better, you could make them each their own page under Typography to remove any confusion.

packages/typography/src/Typography.story.tsx
```diff
@@ -24,6 +33,23 @@ storiesOf('Typography', module).add('Default', () => (
       <InlineCode className={displayBlock}>InlineCode</InlineCode>
       <Disclaimer className={displayBlock}>Disclaimer</Disclaimer>
       <Overline>Overline</Overline>
+      <Link
```
DesignerDave, line 36: Can we add a knob to show or hide the external link to demonstrate that behavior?

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,158 @@
...
+  } else {
+    if (hrefHostname === currentHostname) {
+      target = '_self';
+    } else {
+      target = '_blank';
+    }
+  }
+
+  const icon =
+    target === '_blank' ? (
```
DesignerDave, line 132: Not a fan of the nested ternary here. Could we please break this up for readability?

packages/typography/src/index.ts
DesignerDave: I wasn't having this issue earlier today 🤔

packages/typography/src/Typography.story.tsx
brookescarlett: Typo, my bad! Thank you!

packages/typography/src/Typography.story.tsx
brookescarlett: Haha! had to do that in order to demonstrate what a local link looks like..

packages/typography/src/Typography.story.tsx
brookescarlett: Thought it was nice to see everything all together... would it be alright if I leave it this way but make the knobs more clear?

packages/typography/src/Typography.story.tsx
brookescarlett: That's why the href is selectable

packages/typography/src/Typography.story.tsx
DesignerDave: Sure!

DesignerDave: Sure!

packages/typography/src/Typography.story.tsx
DesignerDave: Oh whoops. How would you disable the external link icon in rare cases where that's necessary? I believe it's something we identified as needing a prop for before.

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,158 @@
...
+`;
+
+const openInNewTabStyles = css`
+  margin-bottom: -1px;
+  margin-left: -1px;
+  margin-right: -2px;
+`;
+
+const OpenInNewTab = (props: JSX.IntrinsicElements['svg']) => {
+  return (
```
DesignerDave, line 65: Why not simply add this to the Icon library? (cc @audroue we should probably do that in the Sketch library as well)

packages/typography/src/Typography.story.tsx
brookescarlett: I didn't have a prop for this, as I thought this was part of the design guidelines. I'm happy to add a prop, maybe `externalIconAppearance`?

packages/typography/src/Link.tsx
brookescarlett: Fixed!

packages/typography/src/index.ts
brookescarlett: Screen Shot 2020-07-20 at 2 27 36 PM

brookescarlett: I'm seeing this issue coming from Menu too...

packages/typography/src/Link.tsx
brookescarlett: What should we call the icon?

brookescarlett: I can add to the library

packages/typography/src/Link.tsx
brookescarlett: Yes, done!

packages/typography/src/Typography.story.tsx
DesignerDave: Rob and I spoke about this one, and acknowledged there will be cases where it wouldn't be able to be shown. It's not noted in the guidelines, but the cases do exist.

brookescarlett: Added support for this, let me know what you think.

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,142 @@
...
+const Link: ExtendableBox<LinkProps, 'a'> = ({
+  href,
+  children,
+  className,
+  target: targetProp,
+  arrowAppearance = ArrowAppearance.None,
+  hideExternalIcon = false,
+  ...rest
+}: LinkProps) => {
+  if (!href) {
```
DesignerDave, line 92: Elements rendered as `<Link>` components from libraries use `to` rather than `href`. We should allow those to render properly.

brookescarlett: good call -- i can omit this. What do we want this component to fallback on if no href is provided and can't be assumed to be an `a`. (If they pass a value to `as`, it will overwrite this fallback)

DesignerDave: I think in the case where `as` is undefined and `href` is also, we should render the link as a span.

brookescarlett: sounds good!

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+declare const __TARGET__: 'web' | 'node';
+
+function isServer() {
+  return __TARGET__ === 'node';
+}
+
+const ArrowAppearance = {
+  Hover: 'hover',
+  Persist: 'persist',
```
DesignerDave, line 72: Hey! I noticed while playing with the link that the persistent arrows aren't spaced the same from the text as the hover ones. We might want to apply the same translation we do on hover to the persistent version as well.

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,139 @@
...
+  }
+`;
+
+const underline = css`
+  background-repeat: repeat-x;
+  background-position: 0 1em;
+  background-size: 3px 3px;
+
+  ${anchorDataProp.selector}:hover & {
+    background-image: linear-gradient(
```
DesignerDave, line 29: I'm fine if we wait on this, but it would be great to transition the gradient in smoothly. That's not supported directly by CSS unfortunately so it'll probably take some JS to make the transition work in this case.

If you don't want to do this for v1, would you mind making a ticket?

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,139 @@
...
+  }
+`;
+
+const arrowRightIconPersist = css`
+  transform: translate3d(3px, 0, 0);
+`;
+
+const arrowRightIconHover = css`
+  height: 10px;
+  width: 10px;
```
DesignerDave, line 47: There's a size prop for this, right?

packages/typography/src/Link.tsx
```diff
@@ -0,0 +1,139 @@
...
+  return (
+    <Box
+      as={href ? 'a' : 'span'}
+      href={href && href}
+      target={target}
+      className={cx(linkStyles, fontSize, className)}
+      {...anchorDataProp.prop}
+      {...rest}
+    >
+      <span className={underline}>{children}</span>
```
DesignerDave, line 133: I'm not sure I understand the reason to have a span tag here. Could we not apply the underline to the Box? Anchor tags behave identically to span tags from a display perspective.

packages/typography/src/Typography.story.tsx
DesignerDave: Looks good!

packages/typography/src/Link.tsx
brookescarlett: There's a span tag because the underline doesn't appear underneath the icon

brookescarlett: but the icon has the behavior of a link

DesignerDave: Why doesn't the underline appear beneath the link? Is that from the design spec?

brookescarlett: You mean why doesn't the link appear underneath the icon? Yes from the spec

packages/typography/src/Link.tsx
brookescarlett: 👍

packages/typography/src/Link.tsx
brookescarlett: https://jira.mongodb.org/browse/PD-752

packages/typography/src/Link.tsx
brookescarlett: Yes you're right.

packages/typography/src/Link.tsx
brookescarlett: That's a storybook problem. I had the value `persistant` when the value should be `persist`
Will update storybook to pull from the enum directly

PR #475, opened 2020-07-10
Fix Portal creating multiple container elements in dev

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Tooltips append multiple portal roots to the DOM when hovering on the trigger](https://jira.mongodb.org/browse/PD-728)

## 🛠 Types of changes

## ✅ Checklist


packages/portal/src/Portal.tsx
```diff
@@ -19,8 +19,28 @@ function createPortalContainer(className?: string): HTMLElement {
 }
 
 function Portal(props: PortalProps) {
-  const [container] = useState(
-    props.container ?? createPortalContainer(props.className),
+  const [container, setContainer] = useState(props.container);
```
hswolff, line 22: I'm wondering if you could do:

```jsx
const [container] = useState(
    props.container ?? () => createPortalContainer(props.className));
```

As it should only run once and useState will call the anonymous function.

lazytype: See see the comment down below. The TL;DR is that it will run more than once in dev
https://github.com/facebook/react/issues/15074#issuecomment-471197572

PR #479, opened 2020-07-14
PD-732: Add onCopy prop to Code component

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-732](https://jira.mongodb.org/browse/PD-732)

## 🛠 Types of changes

## ✅ Checklist


packages/code/src/Code.tsx
```diff
@@ -236,6 +236,12 @@ interface CodeProps extends SyntaxProps {
    * default: `true`
    */
   copyable?: boolean;
+
+  /**
+   * Callback fired when Code is copied
+   *
+   */
+  onCopy: React.MouseEventHandler;
```
DesignerDave, line 244: Couple things:
- I'm inclined to think that, at least this time around, we shouldn't pass anything to this callback (e.g. this is only a function, not a mouse event handler). `onCopy` is a real thing in HTML, and we may want to expand this handler to behave the same in the future.
- This needs to be an optional prop 😄

brookescarlett: Good points, just updated.

PR #489, opened 2020-07-20
PD-741: SSR fixes, TypeScript <3.8 compatibility fix

DesignerDave:
## ✍️ Proposed changes

Usages of Menu and Popover were breaking SSR contexts due to the usage of `Element`. within their PropTypes. This PR conditionally uses Element based on the build target, and otherwise uses `PropTypes.any`.

Additionally, this PR restores the compatibility of some components with TypeScript <3.8 by removing usages of the `import type` syntax introduced in 3.8.

🎟 _Jira ticket:_ [PD-741](https://jira.mongodb.org/browse/PD-741)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/menu/src/Menu.tsx
+++ b/packages/menu/src/Menu.tsx
@@ -399,7 +399,10 @@ Menu.propTypes = {
   className: PropTypes.string,
   align: PropTypes.oneOf(Object.values(Align)),
   justify: PropTypes.oneOf(Object.values(Justify)),
-  refEl: PropTypes.shape({ current: __TARGET__ === 'web' ? PropTypes.instanceOf(Element) : PropTypes.any }),
+  refEl: PropTypes.shape({
+    current:
+      __TARGET__ === 'web' ? PropTypes.instanceOf(Element) : PropTypes.any,
+  }),
   usePortal: PropTypes.bool,
   trigger: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
   open: PropTypes.bool,
--- a/packages/menu/src/Menu.tsx
+++ b/packages/menu/src/Menu.tsx
@@ -388,12 +388,18 @@ function Menu({
 
 Menu.displayName = 'Menu';
 
+// __TARGET__ is a global variable that indicates the webpack build target.
+//
+// We're typing this here because doing it globally was proving problematic.
+// We should solve for this if we need to use __TARGET__ elsewhere.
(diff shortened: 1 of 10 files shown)


packages/menu/src/Menu.tsx
```diff
@@ -388,12 +388,18 @@ function Menu({
 
 Menu.displayName = 'Menu';
 
+// __TARGET__ is a global variable that indicates the webpack build target.
+//
+// We're typing this here because doing it globally was proving problematic.
+// We should solve for this if we need to use __TARGET__ elsewhere.
```
brookescarlett, line 394: Should we type this in lib and have places import. I think this will be the third usage in our repo

brookescarlett: Didn't even see that it was used everywhere in this PR -- maybe now is a good time to add to lib?

DesignerDave: I actually think that may break the build. This is one of those things that `import type` might actually be good for. It's worth testing, but trying to import a constant that's initialized without a value might overwrite the build variable.

scripts/jestSetup.js
```diff
@@ -15,3 +15,5 @@ global.IntersectionObserver = class {
   unobserve() {}
   disconnect() {}
 };
+
+global.__TARGET__ = 'web'
```
brookescarlett, line 19: 👏

packages/menu/src/Menu.tsx
DesignerDave: Yep, so if you try to add it to `lib`, it outputs:

```js
current:"web"===a.__TARGET__?u.default.instanceOf(Element):u.default.any
```

The problem with this is that the bundle doesn't have an instance of the `__TARGET__` variable, and will default to using the `node` environment. If we keep it as it's done in the PR, any code in a conditional that relies on the value of `__TARGET__` is removed altogether when appropriate.

Something interesting I realized is that now that we're outputting ES Modules as well, the ES Modules will output the `__TARGET__` usage directly. I might need to look into this, though I'd suggest this not be a blocker for the PR since we:
1. Don't truly support ES Modules as an API yet.
2. The emotion package already outputs this, and it hasn't yet been problematic.

I will make a ticket for the issue though.

brookescarlett: 👍

PR #494, opened 2020-07-22
PD-751: Make existence checks include null

brookescarlett:
## ✍️ Proposed changes

- Make existence checks include null
= This means that if a user passes in null it'll be treated as not existing.

🎟 _Jira ticket:_ [PD-751](https://jira.mongodb.org/browse/PD-751)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/box/src/Box.tsx
+++ b/packages/box/src/Box.tsx
@@ -45,7 +45,7 @@ function InlineBox(props: BoxProps, ref: React.Ref<any>) {
     return <Component {...rest} ref={ref} />;
   }
 
-  if (props.href != null) {
+  if (typeof props.href === 'string') {
     return <a {...props} ref={ref} />; //eslint-disable-line jsx-a11y/anchor-has-content
   }
 
--- a/packages/box/src/Box.tsx
+++ b/packages/box/src/Box.tsx
@@ -45,7 +45,7 @@ function InlineBox(props: BoxProps, ref: React.Ref<any>) {
     return <Component {...rest} ref={ref} />;
   }
 
-  if (typeof props.href === 'string') {
+  if (props.href != null) {
     return <a {...props} ref={ref} />; //eslint-disable-line jsx-a11y/anchor-has-content
   }
 
--- a/packages/box/src/Box.tsx
+++ b/packages/box/src/Box.tsx
@@ -39,13 +39,13 @@ function InlineBox<TProps>(
 ): JSX.Element;
 
 function InlineBox(props: BoxProps, ref: React.Ref<any>) {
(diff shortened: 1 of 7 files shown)


packages/box/src/Box.tsx
```diff
@@ -39,13 +39,13 @@ function InlineBox<TProps>(
...
 function InlineBox(props: BoxProps, ref: React.Ref<any>) {
-  if (props.as !== undefined) {
+  if (props.as != null) {
     const { as: Component, ...rest } = props;
     // @ts-expect-error
     return <Component {...rest} ref={ref} />;
   }
 
-  if (props.href !== undefined) {
+  if (props.href != null) {
```
lazytype, line 48: can we go even so far as to make this`typeof props.href !== 'string'`?

brookescarlett: wfm

brookescarlett: actually, getting this from ts

```Types of property 'onCopy' are incompatible.
        Type '((event: ClipboardEvent<HTMLAnchorElement>) => void) | undefined' is not assignable to type '((event: ClipboardEvent<HTMLDivElement>) => void) | undefined'.
          Type '(event: ClipboardEvent<HTMLAnchorElement>) => void' is not assignable to type '(event: ClipboardEvent<HTMLDivElement>) => void'.
            Types of parameters 'event' and 'event' are incompatible.
              Type 'ClipboardEvent<HTMLDivElement>' is not assignable to type 'ClipboardEvent<HTMLAnchorElement>'.
                Type 'HTMLDivElement' is missing the following properties from type 'HTMLAnchorElement': charset, coords, download, hreflang, and 21 ```

Comment (brookescarlett) [reply packages/box/src/Box.tsx]: I can enforce `string` in other components (Menu, SubMenu and Button) but not in Box 🤔

Comment (lazytype) [reply packages/box/src/Box.tsx:48]: my guess is that
```js
if (typeof props.href != 'string') {
  return ...
}
// props is not being narrowed from `BoxAnchorDefault | BoxDivDefault` to `BoxDivDefault` anymore
```
but no biggie

packages/button/src/Button.spec.tsx
```diff
@@ -61,6 +61,11 @@ describe('packages/button', () => {
     expect(button.tagName.toLowerCase()).toBe('a');
   });
 
+  test(`renders component inside of button tag when "href" prop is set to null`, () => {
+    const { button } = renderButton({ href: null });
+    expect(button.tagName.toLowerCase()).toBe('button');
```
hswolff, line 66: nice!!!

PR #498, opened 2020-07-25
Allow testing server side rendering

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LeafyGreenProvider is not compatible with SSR](https://jira.mongodb.org/browse/PD-738)
🎟 _Jira ticket:_ [Approximate SSR in tests](https://jira.mongodb.org/browse/PD-755)

## 🛠 Types of changes

## ✅ Checklist


packages/checkbox/src/Checkbox.tsx
```diff
@@ -206,7 +205,8 @@ export default class Checkbox extends PureComponent<
   };
 
   render() {
-    const checkboxId = this.props.id || this.checkboxId;
+    const checkboxId =
+      this.props.id || `checkbox-${Checkbox.idAllocator.generate()}`;
```
DesignerDave, line 209: This ID generation runs on every render right now – each time the checkbox becomes checked, the ID changes which could pose accessibility problems.

Additionally, the ID seems to increment by two every time `generate()` is invoked, which seems unintentional (it's probably strict mode).

.changeset/serious-ligers-film.md
```diff
@@ -0,0 +1,18 @@
+---
+'@leafygreen-ui/checkbox': minor
+'@leafygreen-ui/emotion': minor
+'@leafygreen-ui/hooks': minor
+'@leafygreen-ui/lib': minor
```
DesignerDave, line 5: I believe we removed the default of `document` from `useEventListener` which would be breaking.

packages/emotion/src/emotion.ts
```diff
@@ -21,7 +15,7 @@ function createEmotionInstance() {
     key: 'leafygreen-ui',
   };
 
-  if (__TARGET__ === 'web') {
+  if (typeof window !== 'undefined') {
```
DesignerDave, line 18: If we're abandoning the `__TARGET__` build variable, we can update our webpack configuration and package.json files to only output and point to one build file.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -153,11 +153,26 @@ function OrgNav({
...
   const [onPremMenuOpen, setOnPremMenuOpen] = useState(false);
-  const { width: viewportWidth } = useViewportSize();
   const onElementClick = useOnElementClick();
-  const { orgNav } = urls;
-  const isTablet = viewportWidth < breakpoints.medium;
-  const isMobile = viewportWidth < breakpoints.small;
+  const viewportSize = useViewportSize();
+  const [showProductTour, setShowProductTour] = useState(false);
+
+  const isTablet = viewportSize
```
DesignerDave, line 160: Could we do this rather than the ternary?

```js
const isTablet = viewportSize && viewportSize.width < breakpoints.medium;
```

packages/box/src/Box.spec.tsx
```diff
@@ -140,7 +140,7 @@ describe('packages/box', () => {
 
     test('it renders the box component as the custom component', () => {
       expect(box).toBeInTheDocument();
-      expect(linkWrapperFn).toHaveBeenCalledTimes(1);
+      expect(linkWrapperFn).toHaveBeenCalled();
```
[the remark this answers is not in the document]
lazytype: It gets called twice with SSR because the server render and hydrate both call it.

packages/lib/src/index.ts
```diff
@@ -139,3 +139,46 @@ export type RecursivePartial<T> = {
     ? RecursivePartial<U> | null
     : T[P];
 };
+
+export class IdAllocator {
```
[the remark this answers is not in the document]
lazytype: Not global, each instance of `IdAllocator` starts at 0.

packages/lib/src/index.ts
```diff
@@ -139,3 +139,46 @@ export type RecursivePartial<T> = {
...
+  private static registry: Array<IdAllocator> = [];
+
+  private value = 0;
```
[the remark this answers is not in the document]
lazytype: Since the current expectation is that the caller with pre/append a unique string, I'm thinking we can take in the prefix as an argument to the constructor. how does that sound?

.changeset/serious-ligers-film.md
lazytype: It still defers defaulting to `document` in the `useEffect` (i.e. `?? document`) of `useEventListener`, so the behavior won't change there

packages/lib/src/index.ts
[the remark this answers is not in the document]
DesignerDave: Right, sorry I misread.

packages/lib/src/index.ts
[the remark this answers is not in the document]
DesignerDave: I dig that. Especially since the IDs aren't incrementing globally, that makes the prefix absolutely necessary.

packages/emotion/src/emotion.ts
lazytype: I think there are still a non-zero number of usages, but they don't impact SSR vs client rendering

.changeset/serious-ligers-film.md
DesignerDave: 🤦 Missed that, makes sense!

packages/mongo-nav/src/org-nav/OrgNav.tsx
lazytype: sure, though I think we need `!!` in front of `viewportSize` so that `isTablet` becomes `false` instead of `null`

packages/emotion/src/emotion.ts
DesignerDave: They don't? The `browser` and `main` fields in the package.json for our components point at different builds that rely on these variables. I'm also not seeing any instances of `__TARGET__` in source code on your branch.

If we simplify to output one build, we could speed up our builds a bit which would be nice.

packages/lib/src/index.ts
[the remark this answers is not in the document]
brookescarlett: Love this, we've needed something like this for a while.

packages/mongo-nav/src/org-nav/OrgNav.tsx
```diff
@@ -409,19 +425,16 @@ function OrgNav({
...
-            <OrgNavLink
-              // @ts-ignore 'Cannot find name Appcues'
-              onClick={() => Appcues.show('-M4PVbE05VI91MJihJGv')} // eslint-disable-line no-undef
-              className={cx(rightLinkMargin, productTourColor)}
-              data-testid="org-nav-see-product-tour"
-            >
-              See Product Tour
-            </OrgNavLink>
-          )}
+        {showProductTour && (
```
brookescarlett, line 428: While we're in here, are we past `showProductTour` needing to be visible in the nav @DesignerDave?

packages/mongo-nav/src/org-nav/OrgNav.tsx
DesignerDave: Is there a case where useViewportSize returns null?

packages/mongo-nav/src/org-nav/OrgNav.tsx
DesignerDave: I think you're right! https://jira.mongodb.org/browse/PD-621 <- we've been clear since July 15th.

brookescarlett: @lazytype if you want to remove the product tour link and surrounding feel free, otherwise happy to include in my PR.

packages/emotion/src/emotion.ts
lazytype: yeah you're right. I was thinking of the storybook fix, which we actually don't need anymore

packages/mongo-nav/src/org-nav/OrgNav.tsx
lazytype: yeah, on initial render, since the initial read was moved into an effect

packages/mongo-nav/src/org-nav/OrgNav.tsx
lazytype: @brookescarlett I'll let you include it. since I don't have as much context there's more chance I might accidentally screw something up

packages/hooks/src/useEventListener.ts
```diff
@@ -44,10 +44,18 @@ export default function useEventListener<Type extends keyof DocumentEventMap>(
...
-    element.addEventListener(type, callback as EventListener, options);
+    (element ?? document).addEventListener(
+      type,
+      callback as EventListener,
+      options,
+    );
 
     return () => {
-      element.removeEventListener(type, callback as EventListener, options);
+      (element ?? document).removeEventListener(
```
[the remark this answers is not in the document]
lazytype: This is to lazily set the default `element` as `document`. We were using default de-structured values before `{element = document}`, but that causes `document` to be evaluated at the top-level of the function, which isn't valid to do on the server.

packages/hooks/src/useViewportSize.ts
```diff
@@ -1,17 +1,26 @@
...
     width: window.innerWidth,
     height: window.innerHeight,
   };
 }
 
-export default function useViewportSize() {
-  const [viewportSize, setViewportUpdateVal] = useState(getViewportSize);
+export default function useViewportSize(): ViewportSize | null {
+  const [viewportSize, setViewportUpdateVal] = useState<ViewportSize | null>(
+    null,
```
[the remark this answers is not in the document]
lazytype: Having explicit `null` provides more explicit information to the caller. The caller can then decide what a reasonable default is. For example, if it were 0, 0, then all the mongo-nav components would be server-rendered for mobile by default. So instead I explicitly check for `null` and make the default be non-mobile/tablet rendering.

packages/hooks/src/usePoller.ts
```diff
@@ -4,17 +4,15 @@ import useEventListener from './useEventListener';
 function useVisibilityChange() {
   const isVisibilityStateVisible = () => document.visibilityState === 'visible';
 
-  const [isVisible, setIsVisible] = useState(isVisibilityStateVisible);
```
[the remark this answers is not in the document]
lazytype: `isVisibilityStateVisible` is computed using `document` which doesn't exist on the server. it needs to be deferred by moving it to an effect.

lazytype:
Splitting this up into smaller pieces.

PR #499, opened 2020-07-27
Hooks: update `useEventListener` API, fix and type tests

lazytype:
## 🛠 Types of changes

## ✅ Checklist


packages/hooks/src/useEventListener.ts
```diff
@@ -1,9 +1,9 @@
 import { useEffect, useRef } from 'react';
 
 export interface UseEventOptions {
-  options?: AddEventListenerOptions;
+  options?: Omit<AddEventListenerOptions, 'once'>;
   dependencies?: Array<any>;
-  enabled?: boolean;
+  enabled?: boolean | 'once';
```
DesignerDave, line 6: Should we handle `true` and `false` as strings as well?

lazytype: I would want to disallow it in the types and documentation but handle them in the implementation in case non-TS codebases make a mistake. thoughts?

packages/hooks/src/useEventListener.ts
```diff
@@ -44,7 +44,10 @@ export default function useEventListener<Type extends keyof DocumentEventMap>(
       memoizedEventCallback.current(e);
 
     // NOTE(JeT): I'm pretty sure there should be a way to avoid this type assertion, but I couldn't figure it out.
-    element.addEventListener(type, callback as EventListener, options);
+    element.addEventListener(type, callback as EventListener, {
```
hswolff, line 47: To correctly removeEventListener do we need to pass in the same options instance?

DesignerDave: I'm not sure that the instance of the object is required, just that the values are the same.

From the MDN docs, it looks like they don't even do a strict comparison – if I'm reading it right, removeEventListener only cares about the `capture` option's value: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

lazytype: good call

packages/hooks/src/useEventListener.ts
DesignerDave: I dig that, though I'm a tad worried that it'll be confusing 😬 I'll leave it to you. If we do string "true"/"false", I like your approach.

packages/hooks/src/useEventListener.ts
```diff
@@ -40,14 +40,34 @@ export default function useEventListener<Type extends keyof DocumentEventMap>(
       return;
     }
 
+    // @ts-expect-error Handle this in case non-TypeScript users pass in the wrong value
+    if (enabled === 'false') {
```
hswolff, line 44: Just for this case, not for `'true'`? `typeof enabled === 'string' && enabled !== 'once'` then we show a warning?

This will also throw some React warnings cuz it makes the `useEffect` conditional. I think that's fine, cuz it's better we do this than silently coerce it to a boolean.

packages/hooks/src/useEventListener.ts
```diff
@@ -40,14 +40,34 @@ export default function useEventListener<Type extends keyof DocumentEventMap>(
...
 
+    const eventListenerOptions = {
+      ...options,
+      once: enabled === 'once',
+    };
     // NOTE(JeT): I'm pretty sure there should be a way to avoid this type assertion, but I couldn't figure it out.
-    element.addEventListener(type, callback as EventListener, options);
+    element.addEventListener(
+      type,
+      callback as EventListener,
```
hswolff, line 61: You can't figure this out @lazytype ???!?!!? :P

packages/hooks/src/useEventListener.ts
```diff
@@ -40,14 +40,34 @@ export default function useEventListener<Type extends keyof DocumentEventMap>(
       return;
```
hswolff, line 40: in the line above, I'm wondering if we should be more explicit and say `enabled === false` to early return. Seems like we get that behavior due to fortune that `'once'` is truthy. 

Don't feel strongly about this tho.

packages/hooks/src/useEventListener.ts
lazytype: I did look into it actually :P. the typing that TS provides for `addEventListener` isn't intelligent enough. not motivated enough at the moment to get it changed upstream

PR #500, opened 2020-07-27
PD-746: Fix tooltip notch positioning

DesignerDave:
## ✍️ Proposed changes

This PR does a couple things to improve the Tooltip component:
- The positioning logic for the notch is now much simpler, solving several visual discrepancies between different alignments, as well as addressing the core issue of the ticket. Note that align left/right with `justify: fit` is not fixed with this change, as those combinations are obscure at most (I don't know that there's anything using that currently), and the fix is more involved than what this PR is addressing.
- Updates the Button component so that the span element which wraps the Button's children is able to be used properly for popover/tooltip positioning. This change _shouldn't_ impact consuming applications at all, but I marked it as a breaking change because consumers should be careful that this didn't break their usages of the component.
- Adds `typeIs.array` type guard method to lib to clean up some logic within Tooltip.

🎟 _Jira ticket:_ [Tooltip arrow is centered on trigger element regardless of that trigger's width.](https://jira.mongodb.org/browse/PD-746)

## 🛠 Types of changes

## ✅ Checklist


packages/lib/src/typeIs.ts
```diff
@@ -9,3 +9,7 @@ export function button(el?: Node | null): el is HTMLButtonElement {
 export function input(el?: Node | null): el is HTMLInputElement {
   return element(el) && el.tagName.toLowerCase() === 'input';
 }
+
+export function array(item?: any): item is Array<any> {
```
lazytype, line 13: `Array<unknown>` would be preferable. Here's an [example](https://www.typescriptlang.org/play/#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVwIZRR4CeAgisQBSYzAC2A-AFz4UCULN9cmAznKUIkAPHgoA+OAG8AUHDhRgMJFFxc6PFLxhi0wCAgFDiAbhkBfGTNCRYiVBmy4CRYgFUUAaxQQA7impaRhYxYg4eIJ5+QVdhVG8-FElZeUVlVQjuTC0ddH1DGJIzSysZZHQsHDgAIwJKBAgIFm0obIBzNmk5eR5DShcScioGiDZOlJ7JtBxtOBAWFCQ6auAoOABeREaAbQAGAF0zSfM4YAAbXmAuycmAelu4AAEYXgBaG2AMd8JoOAAVAAWfDgvABECQZwAJjUrpg6JBeLxMNUzsBujc4NMcnNmjBWig2hsthAjj0SpNMH0Bu4vD5-PVGmNrhi4Pdnm8Pl9VlBfoDgaDwVCYTx4RBEcjUeibljZvM4ItlqsiSM9ocpSdzpdmRiZfA5S12srGqT5JZzEA) demonstrating why.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -10,11 +10,27 @@ import Popover, {
...
+import { HTMLElementProps, typeIs } from '@leafygreen-ui/lib';
 import { transparentize } from 'polished';
 import debounce from 'lodash/debounce';
 import { trianglePosition } from './tooltipUtils';
 
+/**
+ * Converts any type to an array of that type if it isn't already an array,
+ * or an empty array for nullish values.
+ * */
+function toArray<T>(item: T): Array<any> | Array<T> {
```
lazytype, line 22: This doesn't exactly behave as described. `toArray(42)` will return `Array<any> | Array<number>`. You probably want an overload. [Example](https://www.typescriptlang.org/play/#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVwIZRR4CeAFJjMALYD8AXPisQJQMXVyYDOcAgoSQA8qANYoIAdxQA+OAG8AUHDhRgMJFFzsqnFFxh50wCAj4DiAbgUBfBXeTosOODAj8iZbQxRIANr7gAHzhUABNgBEwUYFDWOABtAF0rBwxsXFd3IQAVaXJKKgYs4kFcuOLS6RTUNOdM80r86gZs8obc6sd0lzd2vK84VvklTlMmnQBeCbgff2Zh5WVVdU0E5JHbEcwxgg9x5nnFRZU1DS0Cq2VNpdPV+O11zYU0HH18OGn6vYAWACZmKwvPTwABGHx6xVI8T+iQBz1e8DQ4K+JFIs18cKBb1CyN6ezCESiMQBQA)

packages/lib/src/typeIs.ts
DesignerDave: Can you help me understand this example better?

```ts
function bar(foo: string) {
    if (arrayAny(foo)) {
        // In the example, this condition would never be hit given the type of the `foo` argument.
        const x: number = foo[0];
    } else {
        // Why isn't this possible? The `foo` argument is typed as a string, so shouldn't this assignment work?
        // @ts-expect-error This should be impossible
        const x: string = foo;
    }

    if (arrayUnknown(foo)) {
        //@ts-expect-error This should be impossible
        const x: number = foo[0];
    } else {
        const x: string = foo;
    }
}
```

packages/tooltip/src/Tooltip.tsx
DesignerDave: You're right – I'll give an overload here a shot.

packages/lib/src/typeIs.ts
lazytype: if we know `foo` is `string`, then `const x: number = foo[0]` should definitely be an error, regardless of the outcome of `array(foo)`, but the `arrayAny` case doesn't produce an error.
since `arrayAny(foo)` causes the inferred type of `foo` to become `Array<any> & string` which implies that:
`foo[0]: (Array<any> & string)[0]`
`foo[0]: Array<any>[0] & string[0]`
`foo[0]: any & string` (string[0] is the character at position 0)
`foo[0]: any`

DesignerDave: I think I understand – I was thrown off by the condition that would never evaluate to true.

I'll update 😄

packages/button/src/Button.tsx
```diff
@@ -176,26 +176,41 @@ const buttonVariants: { readonly [K in Variant]: string } = {
...
     font-size: 14px;
   `,
 
   [Size.Large]: css`
     height: 45px;
     font-size: 16px;
+  `,
+} as const;
+
+const spanSizes: Record<Size, string> = {
```
brookescarlett, line 200: Should these values conform with our spacing tokens?

packages/tooltip/src/Tooltip.tsx
brookescarlett: Could this be solved with `Array.from()` ?

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -63,15 +61,15 @@ export function trianglePosition(
         notchStyleObj.left = `${notchOverlap}px`;
 
         // right align is .8px off, factoring that in here to correct offset
-        containerStyleObj.left = `calc(100% - 1.8px)`;
+        containerStyleObj.right = `-${containerSize}px`;
```
brookescarlett, line 64: Looks like we're no longer factoring in that .8px so we can rm the comment above

packages/button/src/Button.tsx
DesignerDave: Probably in the future – This PR shouldn't have any style changes associated with it though.

packages/tooltip/src/tooltipUtils.tsx
DesignerDave: Done!

PR #503, opened 2020-07-29
Fix non-admin search for project and organization

lazytype:
## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/mongo-nav/src/mongo-select/OrgSelect.tsx
+++ b/packages/mongo-nav/src/mongo-select/OrgSelect.tsx
@@ -257,8 +257,9 @@ function OrgSelect({
   );
 
   const filterData = () => {
+    const normalizedString = value.toLowerCase();
     const filtered = data?.filter(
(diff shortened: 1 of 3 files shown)


packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -257,13 +257,9 @@ function OrgSelect({
...
   const filterData = () => {
-    const invalid = /[()[\]{}\\]/g;
-    const sanitizedValue = value.replace(invalid, '');
-    const search = new RegExp(String(sanitizedValue), 'i');
-
-    const filtered = data?.filter(datum => {
-      return search.test(datum.orgName);
-    });
+    const filtered = data?.filter(
+      datum => datum.orgName.toLowerCase().indexOf(value.toLowerCase()) !== -1,
```
hswolff, line 261: micro nit, pull up the toLowerCase so we don't do it on every iteration.

hswolff: Also this is hella clever.

PR #507, opened 2020-07-30
Attempt to make tests run when changeset versions branch updated

lazytype:
## 🛠 Types of changes

## ✅ Checklist


brookescarlett:
This looks good to me, but did I misread the issue that you linked to me earlier? Is the problem that it's a force push that the bot is making and actions have trouble running after force pushes?

lazytype:
> This looks good to me, but did I misread the issue that you linked to me earlier? Is the problem that it's a force push that the bot is making and actions have trouble running after force pushes?

That's what looks to be the case to me. On the Version Packages pull request it looks like the Github Actions bot is making a force push to the branch, which doesn't trigger the "synchronize" event.

brookescarlett:
So this is our way of circumventing that?

lazytype:
> 
> 
> So this is our way of circumventing that?

I would say this is our way of compensating for what seems to be buggy behavior on Github's side. A possibly nicer solution might be to make the versions packages workflow action directly trigger the test workflow action to run. It wouldn't be as straight-forward, seems possible with this API https://docs.github.com/en/actions/reference/events-that-trigger-workflows#manual-events.

PR #508, opened 2020-07-31
PD-655: Add Design Tokens to LeafyGreen UI

brookescarlett:
## ✍️ Proposed changes

Adds tokens package

🎟 _Jira ticket:_ [PD-655](https://jira.mongodb.org/browse/PD-655)

## 🛠 Types of changes

## ✅ Checklist


brookescarlett:
Screen Shot 2020-07-31 at 11 19 38 AM

packages/tokens/src/Tokens.story.tsx
```diff
@@ -0,0 +1,51 @@
...
+  padding: ${spacing[6]};
+`;
+const spacing7 = css`
+  background-color: #a5fd8b;
+  padding: ${spacing[7]};
+`;
+
+storiesOf('Tokens', module).add('Spacing', () => (
+  <>
+    <div className={cx(spacing1, gutter)}>{spacing[1]}</div>
```
hswolff, line 43: Let's show the variable and its value when rendering here:

```jsx
<Code>spacing.1: {spacing[1]}</Code>
```

packages/tokens/src/Spacing.tsx
```diff
@@ -0,0 +1,9 @@
+export const spacing = {
+  1: '4px',
+  2: '8px',
+  3: '16px',
+  4: '24px',
+  5: '32px',
+  6: '64px',
+  7: '88px',
```
DesignerDave, line 8: @brookescarlett @hswolff Any thoughts around providing aliases for these? I think having a human-readable name for each of these (`small`, `large`, etc) could help map design specs using spacers to the appropriate value. Providing those both would also allow for rapid iteration through these values.

packages/tokens/src/Tokens.story.tsx
brookescarlett: sgtm

packages/tokens/src/Spacing.tsx
DesignerDave: Oh, I suppose the other thing is: We might consider providing these as numbers instead of strings for added flexibility. Not convinced that's right though.

brookescarlett:
Screen Shot 2020-07-31 at 3 16 45 PM
@hswolff

packages/tokens/src/Spacing.tsx
brookescarlett: Personally, I feel more passionately about only exposing one way to access this rather than what the keys themselves are called.

brookescarlett: I also think that because these are being used for layout it makes sense to include the pixels but thats just an opinion and open to dropping them

packages/tokens/package.json
```diff
@@ -0,0 +1,19 @@
+
+  {
+    "name": "@leafygreen-ui/tokens",
+    "version": "0.9.0",
```
hswolff, line 4: I'm wondering if we release this as a beta version.

I feel pretty confident that we're going to want to change this API as we learn more. And the best way for us to learn more is for us to start consuming these values in other leafygreen components.

I think we should see how our usages change the APIs in here. I'd feel more comfortable giving ourselves a pre-release version so we can have space to iterate (in public) as we refine our usages.

packages/tokens/src/Spacing.tsx
```diff
@@ -0,0 +1,9 @@
+export const spacing = {
```
hswolff, line 1: Nit, lower case file name. Also should this be a default export?

packages/tokens/src/Tokens.story.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const textWrapper = css`
+  margin-bottom: ${spacing[2]};
+  font-size: 10px;
+`;
+
+const spacingBlockVariants = {
+  1: css`
+    background-color: #fb4949;
+    width: ${spacing[1]};
```
hswolff, line 18: I wish we could do `spacing.1`

If anything having `spacing-1` as a value is cool, but dunno how we'd do that.

packages/tokens/package.json
brookescarlett: I'm fine with that. I felt most strongly about getting this out the door as a utility for .design but definitely agree that we'll know more about the desired API as we use it ourselves. We don't have to announce this package externally until we feel more confident

packages/tokens/src/Spacing.tsx
brookescarlett: yes makes sense to me

packages/tokens/src/Tokens.story.tsx
brookescarlett: me too re `.1` we can do `tokens.spacing-1`

jetpacmonkey: :eyes:

Not sure if this is actually better, but I'm awake for some reason so here's a thing 😁 

```tsx
const colors = [
  '#fb4949',
  '#497ffb',
  '#62e3fd',
  '#52c825',
  '#fdd063',
  '#fd7fec',
  '#a5fd8b',
];

const spacingBlockVariants = Object.keys(spacing).reduce((acc, key, idx) => {
  acc[key] = css`
    background-color: ${colors[idx]};
    width: spacing[key];
    height: spacing[key];
  `;
  return acc;
}, {});
```

brookescarlett: 👋 @jetpacmonkey i dig this

packages/tokens/package.json
hswolff: Let's change this version in that case, as long as we get agreement from everyone! @DesignerDave

brookescarlett: SGTM. So are we leaving the API as is for now?

packages/tokens/src/Tokens.story.tsx
```diff
@@ -0,0 +1,56 @@
...
+  '#497ffb',
+  '#62e3fd',
+  '#52c825',
+  '#fdd063',
+  '#fd7fec',
+  '#a5fd8b',
+];
+
+const spacingBlockVariants = Object.keys(spacing).reduce((acc, key, idx) => {
+  console.log(key);
```
lazytype, line 26: Looks like this snuck in

PR #511, opened 2020-07-31
PD-758: Add imperative API to force reload of data in MongoNav

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-758](https://jira.mongodb.org/browse/PD-758)

## 🛠 Types of changes

## ✅ Checklist


packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -157,32 +167,44 @@ const navContainerStyle = css`
...
+    ...rest
+  }: MongoNavInterface,
+  ref: React.Ref<any>,
+) {
+  const [reload, setReload] = useState(false);
+
+  useImperativeHandle(ref, () => ({
+    reloadData: () => {
+      console.log('inside the exposed property');
+      setReload(true);
```
lazytype, line 204: if it's never set back to `false`, how can it be reloaded again?

packages/mongo-nav/src/MongoNav.tsx
```diff
@@ -157,32 +167,44 @@ const navContainerStyle = css`
...
+    ...rest
+  }: MongoNavInterface,
+  ref: React.Ref<any>,
+) {
+  const [reload, setReload] = useState(false);
+
+  useImperativeHandle(ref, () => ({
+    reloadData: () => {
+      console.log('inside the exposed property');
+      setReload(curr => !curr);
```
brookescarlett, line 204: Good call 👀 @lazytype

lazytype: could it be renamed to `reloadToggle` or something like that to indicate that whether it's true/false doesn't matter? alternatively do something like `forceUpdateCounter`, e.g. `forceReloadCounter`

brookescarlett: what do think should be renamed `reloadData` or `setReload`

brookescarlett: Also am having trouble with my attempts at testing this... not sure if you have any ideas

lazytype: > what do think should be renamed reloadData or setReload

sorry, I meant `setReload` -> `setReloadToggle`/`setForceReloadCounter`.

> Also am having trouble with my attempts at testing this... not sure if you have any ideas

what sort of issue are you encountering? I'd expect the high-level structure of the test to look something like:
```
// mock fetch data
// render mongo nav
// wait for screen to display the data
// mock fetch with different data
// call reloadData
// wait for screen to display the new data
```

```

Comment (brookescarlett) [reply packages/mongo-nav/src/MongoNav.tsx]: Will rename right now and try this... i was trying something like: 

``` 
 describe('able to access "reload" on a ref passed to MongoNav', () => {
    function MongoNavWrapper() {
      const mongoNavRef = React.useRef();

      setTimeout(() => {
// Below throwing error, not logging
         console.log(Object.keys(mongoNavRef.current));
        mongoNavRef.current.reloadData();
      }, 5000);

     

      return (
        <>
          <MongoNav mode="dev" ref={mongoNavRef} />
          <div>{mongoNavRef.current}</div>
        </>
      );
    }

    beforeEach(async () => {
      await render(<MongoNavWrapper />);
    });

    test('test', () => {
      // console.log(getByTestId('user-menu-trigger'));
    });
  });```

lazytype: if you need to get access to the `reloadData` function in the test, maybe something like this will work:
```
let ref;
function MongoNavWrapper() {
  ref = React.useRef();
  ...
}

...

ref.current.reloadData(); // maybe this needs `act`?
```

brookescarlett: yes got it!

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -123,6 +124,52 @@ describe('packages/mongo-nav', () => {
...
+    };
+
+    beforeEach(async () => {
+      fetchMock.mockResolvedValue(responseObject);
+
+      await renderWrapperComponent();
+
+      act(() => {
+        responseObject.json();
+      });
```
lazytype, line 154: can we reset the mock here?

brookescarlett: yep!

packages/mongo-nav/src/MongoNav.story.tsx
```diff
@@ -17,49 +17,58 @@ const setStorybookWidth = css`
...
-      />
-    </div>
-  </LeafygreenProvider>
-));
+storiesOf('MongoNav', module).add('Default', () => {
+  const mongoNavRef = React.useRef(null);
+
+  setTimeout(() => {
+    mongoNavRef.current.reloadData();
+  }, 5000);
```
lazytype, line 25: This could make it more difficult to use the story for development since it could cause re-renders while you're trying to inspect stuff. Could you add a knob that triggers the data to reload instead? https://github.com/storybookjs/storybook/tree/master/addons/knobs#button

brookescarlett: Fixed that was a a great suggestion, didn't know about buttons!
Just went back through your comments and thought I hit all of them, but let me know if I'm missing anything

packages/mongo-nav/src/MongoNav.spec.tsx
lazytype: if we reset it we can remove the comment below
>// twice, once for main data and once for alerts polling

brookescarlett: Updated! Let me know if this is what you had in mind

packages/mongo-nav/README.md
```diff
@@ -69,6 +69,30 @@ npm install @leafygreen-ui/mongo-nav @leafygreen-ui/leafygreen-provider@1.1.0
...
+## Imperative API Handle
+
+We have customized the ref instance value that is exposed to components wrapping `MongoNav`, such that the instance's current property contains a `reloadData` key. Invoking `reloadData` triggers a refetch of data from MongoNav.
+
+## Example
+
+```js
+const mongoNavRef = React.useRef(null);
+
+if (shouldReloadData) {
```
hswolff, line 81: Pseudo code examples confuse people, as people tend to copy and paste blindly.

Let's show a concrete example:

```jsx
const onClick = () {
   mongooNavRef.current.reloadData();
};
```

packages/mongo-nav/src/MongoNav.spec.tsx
```diff
@@ -165,13 +170,13 @@ describe('packages/mongo-nav', () => {
 
     test('when the reloadData function is called, MongoNav refetches data', async () => {
       // twice, once for main data and once for alerts polling
-      expect(fetchMock).toHaveBeenCalledTimes(2);
+      expect(fetchMock).toHaveBeenCalledTimes(1);
```
lazytype: comment no longer applies. shouldn't the number of calls be zero since it was reset?

PR #515, opened 2020-07-31
Remove __TARGET__ variable

lazytype:
## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/typography/src/Link.tsx
+++ b/packages/typography/src/Link.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import ArrowRightIcon from '@leafygreen-ui/icon/dist/ArrowRight';
 import OpenNewTabIcon from '@leafygreen-ui/icon/dist/OpenNewTab';
 import Box, { ExtendableBox } from '@leafygreen-ui/box';
@@ -59,12 +59,6 @@ const openInNewTabStyles = css`
   margin-right: -2px;
 `;
 
-declare const __TARGET__: 'web' | 'node';
-
-function isServer() {
-  return __TARGET__ === 'node';
-}
-
 const ArrowAppearance = {
   Hover: 'hover',
   Persist: 'persist',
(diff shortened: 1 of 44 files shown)


packages/typography/src/Link.tsx
```diff
@@ -89,11 +83,15 @@ const Link: ExtendableBox<LinkProps, 'a'> = ({
...
 }: LinkProps) => {
+  const [currentHostname, setCurrentHostname] = useState('');
+  const hrefHostname = useMemo(() => href && new URL(href).hostname, [href]);
+
   const size = useBaseFontSize();
   const fontSize = size === 16 ? typeScale2 : typeScale1;
 
-  const hrefHostname = useMemo(() => href && new URL(href).hostname, [href]);
-  const currentHostname = isServer() ? '' : window.location.hostname;
+  useEffect(() => {
```
hswolff, line 92: Is there a reason we can't do the inline check `typeof window !== 'undefined'` instead of moving it to an effect?

I'm worried on a page with many Link components, an initial render of the page will cause an immediate re-render.

lazytype: This would be a problem for server side rendering. From https://reactjs.org/docs/react-dom.htm
> React expects that the rendered content is identical between the server and the client. It can patch up differences in text content, but you should treat mismatches as bugs and fix them. In development mode, React warns about mismatches during hydration. There are no guarantees that attribute differences will be patched up in case of mismatches.

Since we're using `window.location` to determine more than just text contents, the only way to guarantee the correct rendering is to have an effect as the hydration might not fix the initial render.

To avoid a possible re-render, there needs to be a way for the server to provide the initial location to the component. Since the component is server-agnostic, it probably needs to receive this through a prop. One approach is for an app to be wrapped in a `<LocationProvider>` and the consumer could provide the prop to the link component. The provider could then have server-specific logic to determine the location, e.g. `typeof window !== 'undefined' ? window.location.hostname : server.hostname`

hswolff: Do we have tests for this mismatch in another PR? I feel like that'd be something easy to regress on.

hswolff: I guess we'll see if this becomes a performance issue at a later point.

lazytype: This should show up as a console warning in SSR tests. Haven't made a pull request for it yet since it depends on all the other ones being merged first.

PR #517, opened 2020-07-31
Add IdAllocator class to make ID generation deterministic

lazytype:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/checkbox/src/Checkbox.tsx
+++ b/packages/checkbox/src/Checkbox.tsx
@@ -171,7 +171,7 @@ export default class Checkbox extends PureComponent<CheckboxProps> {
     }
   }
 
-  private static idAllocator = new IdAllocator('checkbox');
+  private static idAllocator = IdAllocator.create('checkbox');
   private _defaultCheckboxId?: string;
 
   private get defaultCheckboxId(): string {
--- a/packages/checkbox/src/Checkbox.tsx
+++ b/packages/checkbox/src/Checkbox.tsx
@@ -1,6 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import { createDataProp } from '@leafygreen-ui/lib';
+import { createDataProp, IdAllocator } from '@leafygreen-ui/lib';
 import { css, cx } from '@leafygreen-ui/emotion';
 import { colors } from '@leafygreen-ui/theme';
 import {
@@ -120,7 +120,7 @@ const disabledTextStyle = css`
   color: ${colors.gray[5]};
 `;
 
-interface CheckboxProps {
+interface CheckboxProps extends React.InputHTMLAttributes<HTMLInputElement> {
   variant: Variant;
   checked?: boolean;
   label: React.ReactNode;
@@ -131,9 +131,7 @@ interface CheckboxProps {
   bold: boolean;
 }
 
-export default class Checkbox extends PureComponent<
-  CheckboxProps & React.InputHTMLAttributes<HTMLInputElement>
-> {
+export default class Checkbox extends PureComponent<CheckboxProps> {
   static displayName = 'Checkbox';
 
   static propTypes = {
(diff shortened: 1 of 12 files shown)


packages/lib/src/IdAllocator.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  constructor(private prefix: string) {
+    IdAllocator.registry.push(this);
+  }
+
+  generate(): string {
+    return `${this.prefix}-${this.value++}`;
+  }
+
+  /* The following are only for testing!! */
```
hswolff, line 14: Wonder if we prefix the name to be clear: `testing_snapshot`? Maybe over kill...

hswolff: Also where are the tests that are using these?

packages/tooltip/src/Tooltip.tsx
```diff
@@ -232,8 +234,8 @@ function Tooltip({
   const tooltipRef = useRef<HTMLDivElement>(null);
 
   const tooltipId = useMemo(
-    () => id || `tooltip-${Math.floor(Math.random() * Math.floor(10))}`,
-    [id],
+    () => id ?? tooltipRef.current?.id ?? idAllocator.generate(),
+    [id ?? tooltipRef.current?.id],
```
hswolff, line 238: 1. Are you able to change the values of the dependency array? TIL :woah:
2. Won't these be the same? If `id` is not set then wouldn't `tooltipRef.current?.id` be unset as well?

packages/checkbox/src/Checkbox.tsx
```diff
@@ -173,7 +171,17 @@ export default class Checkbox extends PureComponent<
     }
   }
 
-  checkboxId = `checkbox-${Math.floor(Math.random() * 10000000)}`;
+  private static idAllocator = new IdAllocator('checkbox');
```
hswolff, line 174: I'm curious...would it be simpler to:

```jsx
// in module space
const idAllocator = new IdAllocator('checkbox');

class Checkbox {
  checkboxId = idAllocator.generate();
}
```

packages/lib/src/IdAllocator.ts
lazytype: I was relying on the SSR tests warnings to surface these, but I'll add some unit tests

packages/checkbox/src/Checkbox.tsx
lazytype: it would definitely be simpler. I just wanted to be stingy by only allocating an ID when we need to. also that way if you're looking through the html of the page, the ids you see are more likely to be consecutive and not have "holes" where we use a provided ID instead of generated.

hswolff: How would my way generate additional IDs? Wouldn't it be once per instance as well?

packages/tooltip/src/Tooltip.tsx
lazytype: >Are you able to change the values of the dependency array? TIL :woah:

could you clarify what you mean by "change the values"? 

> Won't these be the same? If id is not set then wouldn't tooltipRef.current?.id be unset as well?

On initial render `tooltipRef.current?.id` won't be set, but on subsequent renders it will be `id` if it's set, or a generated ID

hswolff:
```
[id ?? tooltipRef.current?.id]
```

If the `id` prop changes then the dependency array for the useMemo will change - that's allowed?

packages/checkbox/src/Checkbox.tsx
lazytype: It will always generate 1 ID, whereas lazily generating them will generate 0 or 1 IDs.

hswolff: I don't really like this overhead to get the lazy benefit. Most of this goes away if we switch to function components tho right? So I'll just bite my tongue until then.

packages/lib/src/IdAllocator.ts
hswolff: Mind adding some JSDocs to explain what the below methods do?

packages/tooltip/src/Tooltip.tsx
lazytype: > If the `id` prop changes then the dependency array for the useMemo will change - that's allowed?

it's not really any different from assigning to an intermediate variable, i.e.
```tsx
const givenId = id ?? tooltipRef.current?.id;
useMemo(
  ...,
  [givenId],
);
```

React doesn't actually know or care what goes into a dependency array; it's effectively just an array whose shallow comparison is used to determine whether an operation should be recomputed.

packages/checkbox/src/Checkbox.tsx
lazytype: Right, Brooke already has a PR out that converts the Radio components to FCs

packages/tooltip/src/Tooltip.tsx
hswolff: Yeah I don't see it mentioned anywhere on the docs, but I know Hooks has weird rules.

PR #520, opened 2020-08-03
Enable testing server-side rendering

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Approximate SSR in tests](https://jira.mongodb.org/browse/PD-755)

## 🛠 Types of changes

## ✅ Checklist


greggb:
🙏

scripts/jestSSRSetup.js
```diff
@@ -0,0 +1,101 @@
+const { Script } = require('vm');
+const React = require('react');
+const ReactDOM = require('react-dom');
+const ReactDOMServer = require('react-dom/server');
+const { IdAllocator } = require('@leafygreen-ui/lib');
+const { Context } = require('@leafygreen-ui/testing-lib');
+
+class ArtificialServerContext {
```
hswolff, line 8: I'm a little worried this is overly complex for our needs. Is there a more community standard way of emulating SSR? I'd be hard pressed to maintain this in a satisfactory way, so I'm curious if there's perhaps a more simple or plain way to do this.

lazytype: There's no such standard I know of or found for testing components.
- If we were testing whole web pages then we could use something like Cypress. I think Cypress is working on a solution for testing individual components.
- If we didn't care about testing hydration then we could run Jest tests in a Node environment instead of a DOM environment, but there are some bugs that can only be surfaced through testing hydration.
Both of these would require writing new tests for all of the components and they would have a large overlap with the existing tests.

hswolff: I think the usage of Context, which is new, is primarily what's throwing me off. 

Were we to not use Context most of these global mutations would be done at the module level and affect the global running environment? Context is used here to provide a sandbox for emotion and React??

scripts/jestSSRSetup.js
```diff
@@ -0,0 +1,101 @@
...
+
+const originalRender = ReactDOM.render;
+
+const stylesToCleanup = new Set();
+let registeredStyleCleanup = false;
+
+ReactDOM.render = (element, container, callback) => {
+  element = React.createElement(CacheProvider, { value: cache }, element);
+
+  if (container._reactRootContainer) {
```
hswolff, line 70: Delving into internal APIs of React here? Might be unpleasant to maintain if React team moves this around too much.

lazytype: Yeah, I don't like it, but I don't know of another way. My solution was to find what [`ReactDOM.unmountComponentAtNode`](https://github.com/facebook/react/blob/4c7036e807fa18a3e21a5182983c7c0f05c5936e/packages/react-dom/src/client/ReactDOMLegacy.js#L358) uses to detect react roots. At the very least, we can allows try to follow what it does (unless they deprecate `unmountComponentAtNode` one day)

scripts/jestSSRSetup.js
lazytype: >Context is used here to provide a sandbox for emotion and React??

Right, that's a way to think of it and hopefully not just a sandbox but also a blackbox. You might not care what it's doing specifically, but can trust it will clean up after itself like a good citizen.

You could call the implementation of `[Context.enter]` yourself, but using `within` gives you the peace of mind that `[Context.exit]` will be called, _even if an exception is thrown_ inside of it.

PR #522, opened 2020-08-04
0.1.0 Table Release

brookescarlett:
## ✍️ Proposed changes

New Table component PR for initial release of Table component, without the `selectable` API. 

🎟 _Jira ticket:_ [PD-525](https://jira.mongodb.org/browse/PD-525)

## 🛠 Types of changes

## ✅ Checklist


packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+      [...alignmentStyles],
+      {
+        [altColor]: shouldAltRowColor,
+        [disabledStyle]: disabled,
+      },
+      className,
+    );
+
+    const ariaExpanded =
+      nestedRows.length > 0
```
hswolff, line 274: Should hoist this to its own variable: `const hasNestedRows = nestedRows?.length > 0`

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+        {nestedRows && nestedRows.length > 0 && (
+          <Transition
+            in={isExpanded && !isAnyAncestorCollapsedProp}
+            timeout={150}
+            nodeRef={nodeRef}
+          >
+            {(state: string) => {
+              return (
+                <>
+                  {nestedRows?.map(element =>
```
hswolff, line 301: Should be redundant due to the conditional it's rendered within?

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+          key: `${indexRef.current}-${renderedChildren[0].props.children}`,
+        });
+      }
+
+      return renderedChildren;
+    }, [children, disabled, className, isExpanded, setIsExpanded]);
+
+    // Iterating over children twice because generated memoized values have different dependants
+    const nestedRows = React.useMemo(() => {
+      let hasSeenFirstCell = false;
```
hswolff, line 227: What does this flag do? From my reading it seems un-necessary but I imagine I'm missing something.

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+      if (shouldDispatchHasRowSpan) {
+        tableDispatch({
+          type: TableTypes.SetHasRowSpan,
+          payload: true,
+        });
+      }
+    }, [children, hasNestedRows, hasRowSpan, tableDispatch]);
+
+    // Iterating over children twice because generated memoized values have different dependants
```
hswolff, line 170: This comment is confusing, I see it twice and I think it's more confusing than helpful.

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+            React.cloneElement(child, {
+              children: (
+                <span className={truncation}>{child.props.children}</span>
+              ),
+              key: `${indexRef.current}-${index}`,
+              disabled: child.props.disabled || disabled,
+            }),
+          );
+        } else {
+          if (isComponentType(child, 'Row')) {
```
hswolff, line 201: Can merge these two lines.

hswolff: Does htis even need to be an else if? Can it be its own if?

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+        <IconButton
+          onClick={() => setIsExpanded(curr => !curr)}
+          aria-label="chevron"
+          className={iconButtonMargin}
+        >
+          <Icon aria-label="chevron" color={uiColors.gray.dark2} />
+        </IconButton>
+      );
+      const renderedChildren: Array<React.ReactElement> = [];
+      let hasSeenRow = false;
```
hswolff, line 183:
```suggestion
      let hasNestedRow = false;
```

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+      let hasSeenFirstCell = false;
+      const nestedRows: Array<React.ReactElement> = [];
+
+      React.Children.forEach(children, (child, index) => {
+        if (isComponentType(child, 'Cell') && !hasSeenFirstCell) {
+          hasSeenFirstCell = true;
+        }
+
+        if (child != null && isComponentType(child, 'Row')) {
+          nestedRows.push(
```
hswolff, line 236: We're rendering nestedRows optimistically. This means that we render them even if they're not appearing on the screen. This is a big performance hit, as most commonly people don't have all rows expanded.

Instead we need to change this to being lazy. Only render and compute nested rows when they're expanded and visible in the page.

This is absolutely a big area of performance cost, and switching to lazy rendering will be a very short way of fixing that.

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+    // alignment, would rather wait for proper information before rendering
+    if (!columnInfo) {
+      return null;
+    }
+
+    const shouldAltRowColor = data && data.length >= 10 && !hasNestedRows;
+
+    const alignmentStyles = Object.entries(
+      columnInfo,
+    ).map(([key, { dataType }]) => styleColumn(key, dataType!));
```
hswolff, line 260: We should fix the argument syntax of `styleColumn` as `dataType` can absolutely be undefined, we're not requiring them set that on columns are we?

packages/table/src/TableHeader.tsx
```diff
@@ -0,0 +1,156 @@
...
+        payload: {
+          columnId: index,
+          accessorValue: normalizedAccessor,
+          data,
+        },
+      });
+    }
+  };
+
+  const sortButton = (
```
hswolff, line 126: nit, I'd inline this. It's not being reused.

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,181 @@
...
+  [TableTypes.RegisterColumn]: {
+    dataType?: DataType;
+    index: number;
+  };
+  [TableTypes.SetHasRowSpan]: boolean;
+  [TableTypes.SetHasNestedRows]: boolean;
+  [TableTypes.SortTableData]: {
+    columnId: number;
+    accessorValue: (data: any) => string;
+    data: Array<T>;
```
hswolff, line 24: Why do we need to include data in here? We can pull that straight from state, no need to pass it through an action.

packages/table/src/TableHead.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useTableContext } from './TableContext';
+
+type TableHeaderProps<Shape> = Pick<TableProps<Shape>, 'columns'>;
+
+function TableHead<Shape>({ columns = [] }: TableHeaderProps<Shape>) {
+  const usingHeaderRow = React.useRef(false);
+
+  // Breaks tables with a colSpan when we don't subscribe to state here, trying to figure out why
+  // @ts-expect-error
+  const { state } = useTableContext(); // eslint-disable-line @typescript-eslint/no-unused-vars
```
hswolff, line 15: I don't feel comfortable merging this in if we don't know why. This isn't going to go away any time soon, if anything table will get more complex before it gets simpler. We should solve this when it's simpler.

packages/table/src/TableHead.tsx
```diff
@@ -0,0 +1,63 @@
...
+import HeaderRow from './HeaderRow';
+import TableHeader from './TableHeader';
+import { TableProps } from './Table';
+import { isComponentType } from '@leafygreen-ui/lib';
+import { useTableContext } from './TableContext';
+
+type TableHeaderProps<Shape> = Pick<TableProps<Shape>, 'columns'>;
+
+function TableHead<Shape>({ columns = [] }: TableHeaderProps<Shape>) {
+  const usingHeaderRow = React.useRef(false);
```
hswolff, line 11: What if we just require the use of `HeaderRow`? How much does allowing this be implicit get us? They still need to use a Fragment, so might as well just use a proper component?

@DesignerDave ^

hswolff: What if we moved `<HeaderRow />` into the Table body? I may be scope creeping here.

packages/table/package.json
```diff
@@ -0,0 +1,24 @@
+
+  {
+    "name": "@leafygreen-ui/table",
+    "version": "0.0.1",
+    "description": "leafyGreen UI Kit Table",
+      "main": "./dist/index.js",
+      "types": "./dist/index.d.ts",
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+  return css`
+    & > td:nth-child(1) {
+      padding-left: ${8 + indentLevel * 16}px;
+    }
+  `;
+}
+
+const idAllocator = IdAllocator.create('text-input');
+
+interface RowProps extends React.ComponentPropsWithoutRef<'tr'> {
```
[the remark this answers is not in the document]
brookescarlett: yes indeed, good call

packages/table/src/Row.tsx
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: Ah this was from selectable logic. Can remove

packages/table/src/Row.tsx
brookescarlett: Removed it! Had it in because in earlier iterations we were trying to figure out how to combine this map, but the two maps need to update based on different dependents so its more performant to break it up

packages/table/src/Row.tsx
brookescarlett: You're correct

packages/table/src/Row.tsx
brookescarlett: 👍

packages/table/src/TableHeader.tsx
brookescarlett: 👍

packages/table/src/TableHead.tsx
brookescarlett: Because then they can't just do an array of strings which was an option as well

packages/table/src/Row.tsx
brookescarlett: Fixed!

packages/table/src/TableContext.tsx
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: I think originally we were handling this lazily, but we were unable to make the transition work as we wanted. 

The original idea (which did not work with the current transitions) was to not mount the nested rows when `isExpanded` is false. We can take this a step further by not even calculating the nested rows when `isExpanded` is false, just have to decide if it's worth sacrificing the transition. cc: @DesignerDave

packages/table/src/TableHead.tsx
brookescarlett: Found out what was going on, and solved for it :)

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+  expanded?: boolean;
+  disabled?: boolean;
+  indentLevel?: number;
+  isAnyAncestorCollapsed?: boolean;
```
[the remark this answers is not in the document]
brookescarlett: Those are implementation details. We do have some components where we just don't explicitly document the props we don't want consumers to be aware of, but happy to explore thinking about having an Internal Row. If we go this route, what do you think makes sense in terms of splitting up the component? To me it feels like it'd become circular... if we did Rows render NestedRows but NestedRows are Rows.... 🤔 🤯

packages/table/src/Row.tsx
DesignerDave: You'll need to remind me why the transition was dependent on this.

packages/table/src/TableHead.tsx
DesignerDave: I don't feel too strongly here, though I think allowing an array of strings is a nice feature that I believe more implementations than not would be able to use.

packages/table/src/Row.tsx
brookescarlett: There's nothing for the Transition component to animate if the nested rows haven't yet been calculated

brookescarlett: That's what it appears at least

brookescarlett: If we need to prioritize: animation or performance, how do you both feel?

hswolff:
This is a regression? The top Row doesn't show the expandable icon?

image

packages/table/src/Row.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+    const indexRef = useRef(idAllocator.generate());
+    const [isExpanded, setIsExpanded] = useState(expanded);
+    const nodeRef = useRef(null);
+
+    useEffect(() => {
+      let shouldDispatchHasNestedRows = false;
+      let shouldDispatchHasRowSpan = false;
+
+      React.Children.forEach(children, child => {
```
hswolff, line 138: If `hasNestedRows` and `hasRowSpan` are true we don't need to iterate, right?

Also if `data` changes we need to reset those values.

packages/table/package.json
```diff
@@ -0,0 +1,24 @@
+
+{
+  "name": "@leafygreen-ui/table",
+  "version": "0.0.1",
+  "description": "leafyGreen UI Kit Table",
+    "main": "./dist/index.js",
+    "types": "./dist/index.d.ts",
+  "scripts": {
```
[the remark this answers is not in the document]
brookescarlett: Weird, got it.

packages/table/src/TableHeader.tsx
```diff
@@ -0,0 +1,153 @@
...
+      dispatch({
+        type: TableTypes.RegisterColumn,
+        payload: {
+          // Offsetting 0-index
+          index: index + 1,
+          dataType,
+        },
+      });
+    }
+  }, []);
```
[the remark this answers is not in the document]
brookescarlett: I think not for this PR, but definitely something to think about especially as we consider providing support for editable tables.

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,180 @@
+import React, { createContext, useContext, useMemo, useReducer } from 'react';
+
+const TableTypes = {
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,180 @@
...
+    case TableTypes.SetHasRowSpan:
+      return {
+        ...state,
+        hasRowSpan: action.payload,
+      };
+
+    case TableTypes.SetHasNestedRows:
+      return {
+        ...state,
+        hasNestedRows: action.payload,
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,180 @@
...
+        },
+        data: sortFunction({
+          data: state.data,
+          direction: state.sort?.direction === 'desc' ? 'asc' : 'desc',
+          accessorValue: action.payload.accessorValue,
+        }),
+      };
+
+    default:
+      return { ...state };
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/table/src/Row.tsx
brookescarlett: true

packages/table/src/TableContext.tsx
```diff
@@ -0,0 +1,180 @@
...
+  }
+}
+
+export function TableProvider({ children, data }: TableProviderInterface) {
+  const initialState: State = {
+    sort: {
+      direction: undefined,
+    },
+    data,
```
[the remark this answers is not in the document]
brookescarlett: Great catch, fixed and added test

packages/table/src/Table.story.tsx
```diff
@@ -98,18 +101,17 @@ storiesOf('Table', module)
   .add('No nested Rows', () => (
     <Table
       data={defaultData}
-      columns={
-        <>
```
hswolff: Can I be super annoying and request we actually support fragments as well as an array? I mostly didn't want to deal with transforming a string into TableHeader.

brookescarlett: yes will revert

brookescarlett: done!

brookescarlett:
I can't believe this is finally happening.. W O W

PR #526, opened 2020-08-05
Introduce new ConfirmationModal component

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Implement React Confirmation Modal component](https://jira.mongodb.org/browse/PD-764)

## 🛠 Types of changes

## ✅ Checklist


brookescarlett:
What are your thoughts on shipping ConfirmationModal from inside of our Modal package, as its own named export, rather than introducing a net new package?

lazytype:
> What are your thoughts on shipping ConfirmationModal from inside of our Modal package, as its own named export, rather than introducing a net new package?

I think the upside of having separate packages is the ability to upgrade them independently. For example, the new modals depend on the new base modal major version which centers modals by default. It could be the case that consumers want to use the new modals, but not have to resolve potential issues with upgrading the base modal package.

packages/modal/src/Modal.tsx
```diff
@@ -48,16 +48,21 @@ const visibleBackdrop = css`
```
DesignerDave, line 48: While updating this file, would you mind switching the top-right "X" button to an `IconButton`?

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+`;
+
+interface ConfirmationModalProps {
+  title: string;
+  children: React.ReactNode;
+  open?: boolean;
+  setOpen?: (open: boolean) => void;
+  className?: string;
+  primaryActionProps: ActionProps;
+  secondaryActionProps: ActionProps;
```
DesignerDave, line 63: I might be missing something, so let me know if I am.

I believe that the "secondary action" is always going to be "Cancel" in confirmation modals, and always close the modal, so I don't think it should be configurable. I might suggest that the cancel button fire an `onClose` callback for additional handling an application might need for clearing a form, or updating state. That would probably need to be handled within the `Modal` component.

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+    position: relative;
+    width: 100%;
+    border-top: 1px solid ${uiColors.gray.light2};
+  }
+`;
+
+const footerContentStyle = css`
+  display: flex;
+  justify-content: right;
+  flex-direction: row-reverse;
```
DesignerDave, line 51: I tend to find this harder to understand when reading the JSX for the component. Is there any reason we need row-reverse here rather than just ordering the DOM appropriately?

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+  column-gap: 5px;
```
DesignerDave, line 52: I believe this is a property that works with CSS grid, but not flexbox. Right now, there's no margin between the two buttons in the footer.

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+  padding: 36px;
+`;
+
+const textEntryInputStyle = css`
+  width: 300px;
+  padding-top: 14px;
+`;
+
+const footerStyle = css`
+  &:before {
```
DesignerDave, line 39: Is there a reason we need to use a separate div for the pseudo-element for this border? From what I see, we should be able to combine these rules with the below `footerContentStyle` and remove unnecessary nesting of divs within the footer content.

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const [primaryActionEnabled, setPrimaryActionEnabled] = useState(
+    !requireTextEntryConfirmation,
+  );
+
+  const textEntryConfirmation = useMemo(() => {
+    setPrimaryActionEnabled(!requireTextEntryConfirmation);
+
+    let textEntryConfirmation = null;
+
+    if (requireTextEntryConfirmation) {
```
DesignerDave, line 93: It might be a bit cleaner of an API if we use whether the app is passing `requiredText` to display this input. Is there a benefit to us requiring the consumer to pass two props to enable this feature?

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+      const requiredText = primaryActionLabel.toLowerCase();
+
+      const onInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setPrimaryActionEnabled(event.target.value === requiredText);
+      };
+      textEntryConfirmation = (
+        <div>
```
DesignerDave, line 100: What is this `div` doing here? We have a `textEntryInputStyle` className that we could set display properties and/or margins on if the desire is to create a line break.

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+  font-size: 16px;
+  letter-spacing: 0;
+  line-height: 24px;
+
+  padding: 36px;
+`;
+
+const textEntryInputStyle = css`
+  width: 300px;
+  padding-top: 14px;
```
DesignerDave, line 35: Nit: it's usually best to use margin for this sort of spacing (in between the current element, and a sibling).

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+    if (requireTextEntryConfirmation) {
+      const requiredText = primaryActionLabel.toLowerCase();
+
+      const onInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setPrimaryActionEnabled(event.target.value === requiredText);
+      };
+      textEntryConfirmation = (
+        <div>
+          <TextInput
+            label={`Type "${requiredText}" to confirm your action`}
```
DesignerDave, line 102: From talking with Sooa, "confirm your action" also needs to be configurable to allow for indicating the action you're taking. E.g. `Type "my_username" to delete your MongoDB Cloud account.`

packages/confirmation-modal/src/ConfirmationModal.tsx
lazytype: my main reasoning was to influence the tabbing order so that the primary action comes first and I think logically makes sense since the buttons are right-aligned.

packages/confirmation-modal/src/ConfirmationModal.tsx
lazytype: I think you're thinking of `grid-column-gap` which is being deprecated in favor of `column-gap`, which supports more contexts such as flex https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap#Examples

packages/confirmation-modal/src/ConfirmationModal.tsx
lazytype: I made it its own div so that the border wouldn't be affected by the margin of `footerContentStyle`. If I merged them then I think I would need to replace the margin-top of the footer content with a margin/padding-bottom for the pseudo-element which is less clean.

packages/confirmation-modal/src/ConfirmationModal.tsx
DesignerDave: Got it – makes sense. I'm down with not manipulating tabIndex 👍

packages/confirmation-modal/src/ConfirmationModal.tsx
DesignerDave: Ahh you're right. Regardless, this implementation seems to be broken from my testing locally. Margin seems like a cleaner solution in this case anyways.

packages/confirmation-modal/src/ConfirmationModal.tsx
DesignerDave: I played a bit around with this, and it looks like you can simply move the `border-top: ${uiColors.gray.light2};` to `footerContentStyle`, and remove this div altogether, pseudo-element and all. It achieves exactly the same effect with one less element in the DOM, and quite a bit less CSS.

DesignerDave: Screen Shot 2020-08-10 at 4 10 34 PM

packages/confirmation-modal/src/ConfirmationModal.tsx
lazytype: Yeah interesting, interesting case of IWOMM. I'll switch to margin.

packages/confirmation-modal/src/ConfirmationModal.tsx
lazytype: I decided to merge `setOpen` and `onClose` since they're kind of redundant + the modal never calls `setOpen(true)`.

lazytype: Would be nice if the base Modal component were to merge `setOpen` and `shouldClose`, but that's a discussion for another time.

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+  dangerous?: boolean;
+  requiredInputText?: string;
+}
+
+const ConfirmationModal = ({
+  children,
+  title,
+  requiredInputText,
+  buttonText,
+  dangerous = false,
```
DesignerDave, line 54: We should probably use the variant pattern that we've established in other components. Perhaps `variant={'danger' | 'default'}` could work fine here.

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+      </div>
+      <Footer>
+        <Button
+          variant={dangerous ? 'danger' : 'primary'}
+          disabled={!confirmEnabled}
+          onClick={() => onClose?.(true)}
+        >
+          {buttonText}
+        </Button>
+        <Button onClick={() => onClose?.(false)}>Cancel</Button>
```
DesignerDave, line 102: Hmm... this pattern seems a bit unintuitive to me. I think I'd expect to have a separate callback for handling confirmation than I do for handling the modal closing.

packages/modal/src/Footer.tsx
```diff
@@ -0,0 +1,41 @@
...
+  left: 0;
+  width: 100%;
+
+  display: flex;
+  justify-content: right;
+  flex-direction: row-reverse;
+  padding: 16px 24px;
+  border-top: 1px solid ${uiColors.gray.light2};
+
+  & > * {
```
DesignerDave, line 17: We should rework this selector to not use `*`. This selector will look at every element on the page to check to see if the parent matches the selector, and it's generally bad practice to do this outside of narrow use-cases. Here, I don't think there's justification for using it.

lazytype: Yeah on second thought it also makes sense to leave the styling of the footer contents to the consumer. I pushed these styles up to the modal.

PR #527, opened 2020-08-05
Add Marketing Modal component

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Implement React Confirmation Modal component](https://jira.mongodb.org/browse/PD-764)

## 🛠 Types of changes

## ✅ Checklist


packages/marketing-modal/README.md
```diff
@@ -0,0 +1,116 @@
...
+      tabindex="0"
+      ><span class="leafygreen-ui-13nqixy">Cancel</span></span
+    >
+  </div>
+</div>
+```
+
+## Properties
+
+| Prop                   | Type                           | Description                                                              | Default |
```
brookescarlett, line 101: Cover style prop is missing I believe

packages/marketing-modal/src/MarketingModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+  primaryActionProps: { label: primaryActionLabel, ...primaryActionProps },
+  secondaryActionProps: {
+    label: secondaryActionLabel,
+    ...secondaryActionProps
+  },
+
+  ...modalProps
+}: MarketingModalProps) => {
+  return (
+    <Modal {...modalProps} contentClassName={baseModalStyle}>
```
brookescarlett, line 103: Clicking the "x" in the top right corner of the modal doesn't force the modal to close, instead the cover image gets highlighted.

brookescarlett: I actually think that's just because the setOpen function isn't being passed to the Modal; however, when I tried to explicitly pass it, I'm getting a TS error.

brookescarlett:
I think both this and ConfirmationModal would be better as named exports from the Modal package but open to thoughts cc: @hswolff

.changeset/slow-cobras-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/confirmation-modal': major
+---
+
+Initial release of confirmation modal
```
brookescarlett, line 5: I think this is an accident, rm

brookescarlett:
Looks like a package lock file local to the MarketingModal also accidentally slid in there

.changeset/short-rules-admire.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/modal': major
+---
+
+- Modals are now vertically centered by default.
```
brookescarlett, line 5: Also believe this is an accident, rm

packages/marketing-modal/src/MarketingModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+const footerContentStyle = css`
+  line-height: 24px;
+  padding-bottom: 50px;
+
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+
+  & > * {
```
DesignerDave, line 68: We're using quite a few `& > *` selectors. We should really avoid these in most cases given that they're particularly expensive from a performance perspective. Consider adding an additional className to these elements rather than using a child selector here.

packages/marketing-modal/src/MarketingModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+    padding: 4px;
+  }
+`;
+
+interface MarketingModalProps {
+  title: string;
+  cover: JSX.IntrinsicElements['img'];
```
DesignerDave, line 75: Should we only allow for image tags? I can imagine wanting to use CSS background images, an inline SVG, or a canvas element here.

packages/marketing-modal/src/MarketingModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+  coverStyle?: 'default' | 'cover';
```
DesignerDave, line 76: If I'm understanding this correctly, this is to switch between full-bleed background images, and centered images correct?

I would consider:
- Using a `variant` prop to be consistent with other components in LeafyGreen. This also lets us make changes unrelated to the image itself in the modal based on the variant in the future if we need.
- If not, I might call this `imageStyle`, and having `fill` and `center` be the variants here.

packages/marketing-modal/src/MarketingModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+    </Modal>
+  );
+};
+
+MarketingModal.displayName = 'MarketingModal';
+
+MarketingModal.propTypes = {
+  title: PropTypes.string.isRequired,
+  cover: PropTypes.element.isRequired,
+  coverStyle: PropTypes.oneOf(['default', 'cover']),
```
DesignerDave, line 130: If we define a `CoverStyle` "enum" (not a true enum) above, we can keep this up-to-date programmatically.

```ts
const CoverStyle = {
  default: 'default',
  cover: 'cover',
};

type CoverStyle = typeof CoverStyle[keyof typeof CoverStyle];

...

MarketingModal.propTypes = {
  coverStyle: PropTypes.oneOf(Object.values(CoverStyle)),
};
```

packages/marketing-modal/src/MarketingModal.tsx
lazytype: I chose just image tags as a starting point since I hadn't come up with a one-size-fits-all styling solution yet. I can work at it more if you have a sense that this will be a blocker for most consumers to begin using the component.

packages/marketing-modal/src/MarketingModal.tsx
lazytype: >If I'm understanding this correctly, this is to switch between full-bleed background images, and centered images correct?

Right. I think I explicitly chose not to use `variant` because I don't have a sense of what other ways this component will be customizable in the future and it may be the case we'll want to allow consumers to choose different combinations of those props. My thinking is that in the future `variant` can set the default configuration of multiple props and then individual props like `imageStyle` can override them, or we could make them mutually exclusive, i.e. one must either provide the `variant` prop or explicit set of more granular props. So for now I think I'll go with `imageStyle`.
> and having `fill` and `center` be the variants here.

should the prop continue to be optional and default to `center`?

DesignerDave (approved):

PR #532, opened 2020-08-10
TypeScript fixes for tests that are currently not typechecked

lazytype:
## ✍️ Proposed changes

Fixes type issues caught by enabling typing for all files. Creating this as its own PR since the change to enable typing everywhere will also be sizeable.

## 🛠 Types of changes

## ✅ Checklist


packages/typography/src/Link.tsx
```diff
@@ -119,12 +119,14 @@ const Link: ExtendableBox<LinkProps, 'a'> = ({
     );
   }
 
+  const elementProps = href
```
brookescarlett, line 122: 👍

packages/box/src/Box.tsx
```diff
@@ -1,33 +1,59 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-type BoxDivDefault = {
-  as?: never;
-  href?: never;
-} & React.ComponentPropsWithRef<'div'>;
+type Override<T, U> = Omit<T, keyof U> & U;
```
brookescarlett, line 4: Should we define these in lib so that they can be used elsewhere?

lazytype: We could go by the rule of three as a rule of thumb. When some code appears for a third time it’s a good sign to refactor

brookescarlett: 👍

packages/lib/src/index.spec.ts
```diff
@@ -5,18 +5,20 @@ describe('packages/lib', () => {
     test('Basic usage', () => {
       type Test = OneOf<{ foo: number }, { bar: string }>;
 
-      const fooOkay: Test = { foo: 42 };
-      // @ts-expect-error
-      const fooBad: Test = { foo: 'hello' };
+      (function* (): Generator<Test, void, never> {
```
hswolff, line 8: What's going on here?

packages/box/src/Box.tsx
```diff
@@ -1,33 +1,59 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-type BoxDivDefault = {
```
hswolff, line 4: Mind summarizing the changes in this file? It seems like a lot.

packages/typography/src/Link.tsx
```diff
@@ -119,12 +119,14 @@ const Link: ExtendableBox<LinkProps, 'a'> = ({
...
+    ? ({ as: 'a', href, target } as const)
```
hswolff, line 123: Is the `as const` what was necessary for typing due to the changes to Box?

packages/lib/src/index.spec.ts
lazytype: typescript was complaining about unused variables and there's no way for me to silence it without also silencing other type errors. this is a small hack to assert the type of a bunch of values without having to give them them names since all the yielded values must confirm to the declared generator type

packages/box/src/Box.tsx
lazytype: The main change was to make `BoxProps` be generic so clients could pass in their own overrides. Most of the rest of the changes are drilling the overrides down into each overload variant. I also added the `Override` helper to prevent issues where clients might be using `as` or `href` for their own purposes.

packages/typography/src/Link.tsx
lazytype: No, this fix is independent of the changes to Box. TypeScript sees the previous code as passing in `{as: string, href: string | undefined, target: string | undefined}`, which does not satisfy any of the overload variants. The fix is typed as `{as: 'a', href: string, targe: string} | {as: 'span'}` and since each type in the union satisfies an overload variant, the whole type does.

packages/tabs/src/Tabs.tsx
```diff
@@ -99,7 +99,9 @@ function Tabs({
   as = 'button',
   ...rest
 }: TabsProps) {
-  const childrenArray = React.Children.toArray(children);
+  const childrenArray = React.Children.toArray(children) as Array<
+    React.ReactElement
```
DesignerDave, line 103: Should this not be `React.ReactNode`?

packages/logo/src/Logo.story.tsx
```diff
@@ -103,5 +103,11 @@ storiesOf('Logo', module)
       );
     };
 
-    return <>{Object.keys(map).map(renderProductLogo)}</>;
+    return (
+      <>
+        {Object.keys(map).map(key =>
+          renderProductLogo(key as keyof typeof map),
```
DesignerDave, line 109: This is an interesting solution to this TS issue! We've definitely worked around this elsewhere in less elegant ways.

packages/tabs/src/Tabs.tsx
lazytype: We actually implicitly assume that the children are <Tab /> elements. We're currently attempting to access their props without doing any validation :(

PR #535, opened 2020-08-12
Fix Types for UserMenu

brookescarlett:
UserMenu should accept the URLS as-is from OrgNav or simply an object of the relevant URLS to UserMenu when used independently of MongoNav.

### DIFF
--- a/packages/mongo-nav/src/user-menu/UserMenu.tsx
+++ b/packages/mongo-nav/src/user-menu/UserMenu.tsx
@@ -520,5 +520,3 @@ UserMenu.propTypes = {
 };
 
 export default UserMenu;
-
-console.log('shmooo');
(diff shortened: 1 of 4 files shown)


packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -185,7 +186,7 @@ interface UserMenuProps {
    * Object that supplies URL overrides to UserMenu component.
    * Shape: { userMenu:{ cloud: { userPreferences, organizations, invitations, mfa }, university: { universityPreferences }, support: { userPreferences }, account: { homepage } }}
    */
-  urls?: URLS;
+  urls?: URLS | { userMenu: UserMenuURLS };
```
hswolff, line 189: Any type test for this?

brookescarlett: Added 👍

PR #537, opened 2020-08-13
Make Card box shadow dark

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Update Card dropshadow to be dark not light](https://jira.mongodb.org/browse/PD-762)

## 🛠 Types of changes

## ✅ Checklist


.changeset/gold-pets-hope.md
```diff
@@ -0,0 +1,7 @@
+---
+'@leafygreen-ui/card': patch
+'@leafygreen-ui/tokens': patch
+'@leafygreen-ui/tooltip': patch
+---
+
+Make Card box shadow darker
```
brookescarlett, line 7: Why does this impact tokens and tooltip?

lazytype: whoops, picked the packages that changeset suggested too quickly

brookescarlett:
LGTM, but not sure from a design perspective so I'll let @DesignerDave be the approver here

packages/card/src/Card.tsx
```diff
@@ -1,28 +1,55 @@
...
+  onClick?: React.MouseEventHandler;
 }
 
 const Card: ExtendableBox<CardProps> = ({ className, ...rest }: CardProps) => {
-  return <Box className={cx(containerStyle, className)} {...rest} />;
+  const clickable = rest.onClick !== undefined;
+  return (
+    <Box
+      className={cx(containerStyle, { [clickableStyle]: clickable }, className)}
+      tabIndex={clickable ? 0 : undefined}
```
DesignerDave, line 49: We shouldn't need to set tabIndex on elements that would typically need an onClick handler like links and buttons. Anything else, we may want to let the consumer add tabIndex themselves in cases where they're setting an onClick handler on a typically non-interactive element.

packages/card/src/Card.tsx
```diff
@@ -1,28 +1,55 @@
...
 `;
 
 interface CardProps {
   className?: string;
+  onClick?: React.MouseEventHandler;
 }
 
 const Card: ExtendableBox<CardProps> = ({ className, ...rest }: CardProps) => {
-  return <Box className={cx(containerStyle, className)} {...rest} />;
+  const clickable = rest.onClick !== undefined;
```
DesignerDave, line 45: To account for links, we should probably also check for the existence of `href` or `to` (for in cases like React router links being used with the `as` prop).

lazytype: I think `href` makes sense, but it's odd for the component's behavior to depend on a specific routing library's interface. This could lead to a non react-router component that happens to have a `to` prop unexpectedly getting the clickable behavior applied. I think I'd rather expose a prop that allows the consumer to explicitly specify if the card is clickable, e.g. `clickable?: boolean;`.

DesignerDave: I dig adding that prop. If that prop isn't passed, it can still fall back to appearing clickable if `onClick` or `href` are passed as well. It also has the added benefit of being a way to disable the behavior in the off-chance you need to have an 'onClick' handler while not having the visual affordance.

PR #538, opened 2020-08-13
Add `animate` prop to Checkbox

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Add a way to speed up/disable checkbox animation](https://jira.mongodb.org/browse/PD-696)

## 🛠 Types of changes

## ✅ Checklist


brookescarlett:
It seems that in our current version of Checkbox we don't animate from checked -> unchecked. This PR adds that animation. I'm not opposed to it (it looks cool) just wondering whether or not that was intentional.

lazytype:
> It seems that in our current version of Checkbox we don't animate from checked -> unchecked. This PR adds that animation. I'm not opposed to it (it looks cool) just wondering whether or not that was intentional.

Ah, I didn't notice since I only started testing it after I changed the code. Maybe I just got used to it, but seeing it without the animation does feel a bit jarring. I kind of like it with the animation, though it could probably be sped up a little. I'm open for discussion.

packages/checkbox/src/Checkbox.tsx
```diff
@@ -313,11 +322,13 @@ export default class Checkbox extends PureComponent<CheckboxProps> {
           {...checkboxWrapper.prop}
           className={cx(wrapperStyle, {
             [wrapperStyleChecked]: checked && !indeterminate && !disabled,
+            [wrapperStyleAnimated]: animate && !indeterminate && !disabled,
           })}
         >
           <div
             className={cx(checkboxStyle, checkboxBackgroundImage, {
               [checkboxStyleChecked]: checked && !indeterminate && !disabled,
+              [checkboxStyleAnimated]: animate && !indeterminate && !disabled,
```
DesignerDave, line 331: I don't think this needs the indeterminate or disabled checks to work properly. Ditto for the other instance.

DesignerDave:
> > It seems that in our current version of Checkbox we don't animate from checked -> unchecked. This PR adds that animation. I'm not opposed to it (it looks cool) just wondering whether or not that was intentional.
> 
> Ah, I didn't notice since I only started testing it after I changed the code. Maybe I just got used to it, but seeing it without the animation does feel a bit jarring. I kind of like it with the animation, though it could probably be sped up a little. I'm open for discussion.

I dig it 😄 Let's keep it

packages/checkbox/src/Checkbox.tsx
lazytype: it's needed for this one because if we go from `checked -> checked + indeterminate -> checked`, we'll animate the check, which looks pretty weird. I kept it for the wrapper just for symmetry.

DesignerDave: Makes sense! 👍

PR #543, opened 2020-08-14
PD-708 and PD-725: Tooltip Component Updates

brookescarlett:
## ✍️ Proposed changes

- `Tooltip` typography previously relied on consumer application's styles. Now, Tooltip children are wrapped in LeafyGreen Body component. 
- Setting `usePortal` to `false` no longer stops Tooltips from appearing on hover when the trigger is a LeafyGreen Button component.

🎟 _Jira ticket:_ [PD-708](https://jira.mongodb.org/browse/PD-708)
🎟 _Jira ticket:_ [PD-725](https://jira.mongodb.org/browse/PD-725)

## 🛠 Types of changes

## ✅ Checklist


packages/button/src/Button.tsx
```diff
@@ -175,24 +175,24 @@ const buttonVariants: { readonly [K in Variant]: string } = {
 
 const buttonSizes: Record<Size, string> = {
   [Size.XSmall]: css`
-    height: 22px;
+    min-height: 22px;
```
brookescarlett, line 178: Made this min-height so that when the text was overflowing, the button grew. I did this to account for removing the `overflow: hidden` rule. If we don't like this approach it's easy to add back the `overflow:hidden` rule; however, apply it to the span contained inside of the button rather than to the button itself.

packages/button/src/Button.tsx
```diff
@@ -360,7 +359,12 @@ const Button: ExtendableBox<BaseButtonProps, 'button'> = React.forwardRef(
     const modifiedGlyph =
       glyph && children
         ? React.cloneElement(glyph, {
-            className: cx({ [glyphMargins[size]]: glyph != null }),
+            className: cx(
+              css`
+                flex-shrink: 0;
```
brookescarlett, line 364: Ensures that the Icon doesn't shrink when the text overpowers it. We should consider building this into our icons perhaps, I feel like we apply this rule a lot.

.changeset/loud-clouds-guess.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/tooltip': patch
+---
+
+- `Tooltip` typography previously relied on consumer application's styles. Now, Tooltip children are wrapped in LeafyGreen Body component. 
+- Setting `usePortal` to `false` no longer stops Tooltips from appearing on hover when the trigger is a LeafyGreen Button component.
```
brookescarlett, line 6: Will adjust the changeset once there's alignment on the approach of updating Button rather than baking this change into Tooltip directly.

packages/tooltip/src/Tooltip.tsx
```diff
@@ -332,7 +345,7 @@ function Tooltip({
                 className={cx(triangleStyle.notchStyle, notchVariants[variant])}
               />
             </div>
-            {children}
+            <Body className={childrenVariants[variant]}>{children}</Body>
```
DesignerDave, line 348: I'm not sure we're gonna want to always wrap children in a body component. We should consider scoping our solution to the ticket by simply applying the font size and color rules to a parent of children that already exists in the component.

packages/button/src/Button.tsx
brookescarlett: 👍

packages/button/src/Button.tsx
brookescarlett: 👍

packages/tooltip/src/Tooltip.tsx
brookescarlett: I'm fine with just applying those two rules, but I thought the point of the Body component was to standardize our typography usage, so this seemed like a great use-case for that.

.changeset/fast-wasps-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/button': patch
```
DesignerDave, line 2: This is also a breaking change, due to the CSS overflow change.

DesignerDave (changes requested):

.changeset/fast-wasps-turn.md
brookescarlett: 👍

.changeset/loud-clouds-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/tooltip': patch
```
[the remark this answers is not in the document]
brookescarlett: 👍

DesignerDave (approved):

PR #545, opened 2020-08-19
Add command to publish Storybook to personal fork

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Create and document workflow to publish storybook to personal fork of leafygreen-ui](https://jira.mongodb.org/browse/PD-771)

## 🛠 Types of changes

- [X] Tooling (updates to workspace config or internal tooling)

## ✅ Checklist

- [X] I have added necessary documentation (if appropriate)


README.md
```diff
@@ -155,6 +155,12 @@ You can deploy a static build of our Storybook site to gh-pages from the master
 1. First be sure you've built a static version of Storybook: `yarn build:storybook`
 2. Then deploy to gh-pages: `yarn release:site`
 
+### To deploy to your own mirror of leafygreen-ui
```
hswolff, line 158: Does this assume you've already created your fork?

scripts/publish-storybook.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
```
hswolff, line 1: This is fantastic! In some ways should be its own npm package. We could create a new repo on https://github.com/mongodb-js and publish on there if we want. May add more overhead than we want right now.

README.md
lazytype: Oh, yeah it does. I need to think about how to approach that case.

scripts/publish-storybook.sh
lazytype: Probably more maintainable to rewrite in JS if it's not gonna be just a one-off and would also allow us to import the gh-pages package directly. One thing I wanted to do was make the storybook build happen it the background while you're prompted for username and such, but couldn't get ctrl+c to kill the child process. I imagine there must be something in npm that could make that easier.

DesignerDave (approved):
I don't know why this wasn't approved before. Was anything blocking this? LGTM

PR #551, opened 2020-08-20
Small Updates

brookescarlett:
- Update idAllocator in Table
- Remove console log in Tokens

### DIFF
--- a/packages/table/src/Row.tsx
+++ b/packages/table/src/Row.tsx
@@ -100,7 +100,7 @@ function getIndentLevelStyle(indentLevel: number) {
   `;
 }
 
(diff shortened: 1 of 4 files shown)


packages/table/src/Row.tsx
```diff
@@ -100,7 +100,7 @@ function getIndentLevelStyle(indentLevel: number) {
   `;
 }
 
-const idAllocator = IdAllocator.create('text-input');
+const idAllocator = IdAllocator.create('table');
```
lazytype, line 103: should it be “table-row” just to be safe from collisions with other table components?

brookescarlett: makes sense, updated!

PR #553, opened 2020-08-21
Add Yarn script command to type-check only (small version)

lazytype:
Alternative implementation to #450. The primary downside is that VSCode still won't show the correct type errors in test files.

## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


lazytype:
> First, it looks like the table package doesn't love the `yarn ts` command, though I think the issues aren't related to the changes here. Still, we should probably fix those errors here so that the command passes.

Yep, I have a separate PR to address the Table issues here https://github.com/mongodb/leafygreen-ui/pull/556

tsconfig.json
```diff
@@ -1,111 +1,19 @@
...
-    },
-    {
-      "path": "./packages/leafygreen-provider"
-    },
-    {
-      "path": "./packages/typography"
-    }
-  ]
+  "include": ["packages/**/src/**/*", "typings/**/*"],
+  "compilerOptions": {
```
hswolff, line 3: Anyway we can extend from package.tsconfig.json as the settings should be mostly the same? Or vice versa?

DesignerDave:
> > First, it looks like the table package doesn't love the `yarn ts` command, though I think the issues aren't related to the changes here. Still, we should probably fix those errors here so that the command passes.
> 
> Yep, I have a separate PR to address the Table issues here #556

Ahh, I see! Looks good from my perspective then.

common.tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
```
hswolff, line 2: :chefskiss:

PR #557, opened 2020-08-25
PD-779: Configure .design Environment

brookescarlett:
## ✍️ Proposed changes

This PR creates website dir for new mongodb.design. Next and emotion are installed, so that we can be up and running after we configure this skeleton with Realm. Our eslint config is now formatting files contained within this directory, but I created a new .tsconfig local to the website directory, because some of our rules in package.tsconfig.json are in conflict with rules that next.js requires (noEmit: "true").

To note: When testing locally, I discovered a bug with our `run init` script. Before `cd`-ing into the `website` dir and running `yarn dev` first run `yarn build` from our root. Ticket here: [Investigate build script](https://jira.mongodb.org/browse/PD-830)

🎟 _Jira ticket:_ [PD-779](https://jira.mongodb.org/browse/PD-779)

## 🛠 Types of changes

## ✅ Checklist


.eslintignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 dist
 lib
-storybook/public
\ No newline at end of file
+storybook/public
+website/.next
```
brookescarlett, line 5: This doesn't seem to be working, still seeing these files appear when I run `yarn format` 🤔

website/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "compilerOptions": {
+    "rootDir": "pages",
+    "jsx": "preserve",
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true,
+    "noEmit": true,
```
brookescarlett, line 10: This rule is required by next.js and conflicts with our existing package.tsconfig, which is why I'm not extending our rules here.

.eslintignore
hswolff: May need to add it to `.prettierignore` as well.

.eslintrc.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
     },
   },
   rules: {
+    'react/react-in-jsx-scope': 'off',
```
hswolff, line 29: We don't want to apply this to every file right, just the files in website? We can use the overrides config to be more selective in what this rule applies to.

brookescarlett: 👍

.eslintignore
brookescarlett: 🤦

brookescarlett: updated

website/tsconfig.json
hswolff: 😬

PR #565, opened 2020-08-26
Add fonts to tokens package

brookescarlett:
## ✍️ Proposed changes

- Adds `fonts` to tokens package with `default` and `code` values set
- Explicitly sets font for `Syntax` component
- Leverages tokens package to set font in InlineCode 

🎟 _Jira ticket:_ [PD-827](https://jira.mongodb.org/browse/PD-827)
🎟 _Jira ticket:_ [PD-829](https://jira.mongodb.org/browse/PD-829)

I know that there's still debate as to which font will be used, Source Code Pro or Fira Mono, so I thought that abstracting to tokens package would allow this to replace all at once when the time comes. Also would be nice for Akzidenz to have one place where all of the fallback fonts are listed in the proper order. If you like adding fonts to tokens, I'll replace instances of explicitly setting Akzidenz and its respective fallbacks throughout Leafy.

## 🛠 Types of changes

## ✅ Checklist


.changeset/long-zebras-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/syntax': patch
+---
+
+Explicitly sets font-family for syntax component to Source Code Pro
```
hswolff, line 5:
```suggestion
Explicitly sets `font-family` for syntax component to `Source Code Pro`
```

.changeset/thin-lizards-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/typography': patch
+---
+
+Uses Tokens package to define font-family for InlineCode component
```
hswolff, line 5:
```suggestion
Uses Tokens package to define `font-family` for InlineCode component
```

packages/tokens/README.md
```diff
@@ -41,3 +37,18 @@ import { spacing } from '@leafygreen-ui/tokens';
...
 const example = () => <span style={{ margin: spacing[1] }}>Hello World</span>;
 ```
+
+## Fonts
+
+```js
+import { fonts } from '@leafygreen-ui/tokens';
+
+/**
+ * const fonts = {
```
hswolff, line 47: We should...not do this, cuz then it's another place we need to keep maintained as we change things. 

Right now spacing is already out of date.

Better to just link to the source code directly IMO.

packages/tokens/src/fonts.ts
```diff
@@ -0,0 +1,6 @@
+const fonts = {
+  default: `Akzidenz, 'Helvetica Neue', Helvetica, Arial, sans-serif`,
+  code: `'Source Code Pro', monospace`,
```
DesignerDave, line 3: We should have a more robust set of backups for monospaced fonts – `'Source Code Pro', Menlo, monospace` would probably be better here.

packages/tokens/src/fonts.ts
```diff
@@ -0,0 +1,6 @@
...
+};
+
+export default fonts;
```
DesignerDave, line 6: Can we call this `fontFamilies` instead? These are not comprehensive font declarations since they are only meant to be used with the `font-family` CSS rule.

.changeset/long-zebras-brake.md
brookescarlett: 👍

.changeset/thin-lizards-deliver.md
brookescarlett: 👍

packages/tokens/README.md
brookescarlett: 👍

packages/tokens/src/fonts.ts
brookescarlett: 👍

packages/tokens/src/fonts.ts
brookescarlett: 👍

PR #567, opened 2020-08-26
PD-824: Toast v0.1

DesignerDave:
## ✍️ Proposed changes

Adds the Toast component as a Beta (pre-1.0) release. This PR includes all of the features required by the growth team for the "First Time Setup Wizard" project.

🎟 _Jira ticket:_ [PD-824](https://jira.mongodb.org/browse/PD-824)

## 🛠 Types of changes

## ✅ Checklist


packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,335 @@
...
+   * Optional boolean that renders the Toast and makes it visible when true.
+   *
+   * **Default:** `false`
+   */
+  open?: boolean;
+
+  /**
+   * Optional click event handler that, when set, renders a close button that receives the passed handler.
+   */
+  close?: React.MouseEventHandler;
```
brookescarlett, line 234: nit: banners and modals use `onClose` for a similar prop. Maybe worth renaming for consistency?

brookescarlett: Ah playing with it and I see how this works. nvm

DesignerDave: I named it `close` because it:
1. Doesn't fire when the component would close. The Toast is a fully-controlled component right now.
2. IT also enables a UI feature. By naming it `close`, I felt that it's more clear that defining the prop will also show the close button.

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,335 @@
...
+  background-size: 600px;
+  animation: ${backgroundShimmer} 4s infinite linear;
+  transition: width 0.3s ease-in-out;
+`;
+
+interface ProgressBarProps {
+  progress: number;
+}
+
+function ProgressBar({ progress }: ProgressBarProps) {
```
brookescarlett, line 61: do you think this function and the surrounding logic could be it's own package? we can leave it here for now if you prefer, but I can definitely see us wanting to a) standardize the usage of progress bars and b) share this across components as we work on larger loaders

DesignerDave: This probably won't be its own package. The implementation is super specific to this component. That said, I think I'll break it into it's own file in the package...

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,335 @@
...
+  progress: 'progress',
+} as const;
+
+type ToastVariants = typeof ToastVariants[keyof typeof ToastVariants];
+
+const transitionDuration = 150;
+
+const baseToastStyle = css`
+  position: fixed;
+  bottom: ${spacing[6]};
```
brookescarlett, line 96: :)

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,335 @@
...
+};
+
+const dismissButtonStyle = css`
+  position: absolute;
+  top: ${spacing[2]};
+  right: ${spacing[2]};
+  transition: color 0.15s ease-in-out;
+`;
+
+const dismissButtonVariantStyle = {
```
brookescarlett, line 180: nit: Recently added a variant to Callout and it was really easy to do because all of the possible states were in an object for me. It ensured that I didn't miss anything when trying to add a variant and also was super readable. Maybe this is a pattern we think about applying around when working with variants (specifically when those variant sets will probably grow)

``` [Variant.Note]: {
    header: {
      background: uiColors.blue.light3,
      text: uiColors.blue.dark2,
    },
    text: uiColors.blue.dark3,
    bar: uiColors.blue.base,
    icon: uiColors.blue.base,
  },```

File: packages/toast/src/Toast.tsx [hunk 5]
```diff
@@ -0,0 +1,335 @@
...
+
+  return (
+    <Transition
+      in={open}
+      timeout={transitionDuration}
+      mountOnEnter
+      unmountOnExit
+      nodeRef={nodeRef}
+    >
+      {(state: RTGStates) => (
```
brookescarlett, line 268: RTGStates might be worth moving to lib, as we use Transition across the repo

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,335 @@
...
+  transform: translate3d(0, 0, 0) scale(1);
+  opacity: 1;
+`;
+
+const toastTransitionStateStyles: Partial<Record<RTGStates, string>> = {
+  [RTGStates.Entered]: toastVisibleStateStyle,
+};
+
+const baseIconStyle = css`
+  flex-shrink: 0;
```
brookescarlett, line 130: Did we bake this into Icon already? Or is this a part of a PR that hasn't merged yet 🤔

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,335 @@
...
+                    {title}
+                  </Body>
+                )}
+
+                <Body className={bodyVariantStyles[variant]}>{body}</Body>
+              </div>
+            </div>
+
+            {dismissible && (
+              <IconButton
```
brookescarlett, line 316: think you mean to add an onClick here to call `close`

brookescarlett:
Couple of comments but the component looks very good! Don't forget to add a changeset when you add the tests in 😏

packages/toast/src/Toast.tsx
DesignerDave: Unsure – I did this to be safe TBH 😅 I don't think this was included quite yet though.

packages/toast/src/Toast.tsx
DesignerDave: Possibly! I'll leave that for a different PR though to move this one along.

packages/toast/src/Toast.tsx
DesignerDave: Whoops! Should be fixed now.

packages/toast/src/Toast.tsx
DesignerDave: It's possible this is a better approach 🤔

packages/toast/src/Toast.tsx
DesignerDave: As notes on the spacing library, I found two things that I think we should change:
1. Use numbers instead of strings. On multiple occasions, I had to use CSS `calc()` where I wouldn't have otherwise needed to because they are strings.
2. Add `as const` to the spacing enum so that I can see the values I'm entering, rather than simply `string` or `number`.

packages/toast/src/Toast.tsx
DesignerDave: Yep, this definitely feels a lot better! Just updated this.

DesignerDave:
> Couple of comments but the component looks very good! Don't forget to add a changeset when you add the tests in 😏

Thanks Brooke! Already added a changeset 😛

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,290 @@
...
+}
+
+const ToastVariants = {
+  success: 'success',
+  progress: 'progress',
+} as const;
+
+type ToastVariants = typeof ToastVariants[keyof typeof ToastVariants];
+
+const variantStyles: Record<ToastVariants, Partial<Record<StyledElements, string>>> = {
```
brookescarlett, line 64: 👍

packages/toast/src/Toast.tsx
brookescarlett: https://jira.mongodb.org/browse/PD-840

packages/toast/src/Toast.tsx
brookescarlett: 👍

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,301 @@
...
+  body,
+  className,
+  variant,
+  progress = 1,
+  open = false,
+  close,
+  ...rest
+}: ToastProps) {
+  const nodeRef = useRef(null);
+  const dismissible = !!close;
```
hswolff, line 194:
```suggestion
  const dismissible = typeof close === 'function';
```

For 'correctness' sake?

packages/toast/src/ToastProgressBar.tsx
```diff
@@ -0,0 +1,74 @@
...
+  background-size: 600px;
+  animation: ${backgroundShimmer} 4s infinite linear;
+  transition: width 0.3s ease-in-out;
+`;
+
+interface ProgressBarProps {
+  progress: number;
+}
+
+function ToastProgressBar({ progress }: ProgressBarProps) {
```
hswolff, line 53: Feels inevitable we'll use this elsewhere, but no sense in prematurely optimizing yet.

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,301 @@
...
+            {...rest}
+          >
+            <div
+              className={cx(
+                baseElementStyles.contentWrapper,
+                currentVariantStyles.contentWrapper,
+                {
+                  [css`
+                    padding-right: ${spacing[3] * 2}px;
+                  `]: dismissible,
```
hswolff, line 236: Woah, funky.

packages/toast/src/ToastProgressBar.tsx
DesignerDave: I have the feeling we'd need to split out the positioning logic and just include the progress prop handling and animation.

packages/toast/src/Toast.tsx
DesignerDave: I had a feeling you'd say something about coercing this to a boolean like this 😄  Happy to make the change – I do think your suggestion is correct here.

README.md
```diff
@@ -39,6 +39,7 @@ A set of CSS styles and React components built with design in mind.
 - [Toggle](https://github.com/mongodb/leafygreen-ui/tree/master/packages/toggle)
 - [Tooltip](https://github.com/mongodb/leafygreen-ui/tree/master/packages/tooltip)
 - [Typography](https://github.com/mongodb/leafygreen-ui/tree/master/packages/typography)
+- [Toast](https://github.com/mongodb/leafygreen-ui/tree/master/packages/toast)
```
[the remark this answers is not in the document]
DesignerDave: Wish we could automate this... hah

packages/toast/src/Toast.tsx
DesignerDave: Gotta make sure that the text doesn't collide with the button or its state. It's possible that doing ``${spacing[3] * 2}px`` is actually misleading, and rather this should be ``${spacing[3] + ${magicNumber}}px``.

hswolff: Oh I meant more, inlining the CSS function in the key value. Never thought to do that, so caught me by surprise.

DesignerDave: Ahh gotcha 😄

packages/toast/src/Toast.spec.tsx
```diff
@@ -0,0 +1,223 @@
...
+    test('true, Toast is visible', () => {
+      const { queryByRole } = renderToast({
+        body: 'hello world',
+        variant: ToastVariants.success,
+        open: true,
+      });
+
+      const toast = queryByRole('status');
+      expect(toast).toBeInTheDocument();
+      expect(toast).toBeVisible();
```
lazytype, line 40: a bit redundant since being visible implies being in the document

packages/toast/src/Toast.spec.tsx
```diff
@@ -0,0 +1,223 @@
...
+      const titleText = 'this is the title';
+      const { queryByText } = renderToast({
+        open: true,
+        body: 'hello world',
+        title: titleText,
+        variant: ToastVariants.success,
+      });
+
+      const body = queryByText(titleText);
+      expect(body).toBeInTheDocument();
```
lazytype, line 179: toBeVisible?

packages/toast/src/Toast.spec.tsx
```diff
@@ -0,0 +1,223 @@
...
+      const titleElement = <span>{titleText}</span>;
+      const { queryByText } = renderToast({
+        open: true,
+        body: 'hello world',
+        title: titleElement,
+        variant: ToastVariants.success,
+      });
+
+      const titleSpan = queryByText(titleText);
+      expect(titleSpan).toBeInTheDocument();
```
lazytype, line 193: \+

packages/toast/src/Toast.spec.tsx
```diff
@@ -0,0 +1,223 @@
...
+    test('a string, the title element renders', () => {
+      const bodyText = 'this is the title';
+      const { queryByText } = renderToast({
+        open: true,
+        body: bodyText,
+        variant: ToastVariants.success,
+      });
+
+      const body = queryByText(bodyText);
+      expect(body).toBeInTheDocument();
```
lazytype, line 207: \+

packages/toast/src/Toast.spec.tsx
```diff
@@ -0,0 +1,223 @@
...
+      const bodyText = 'this is the body';
+      const bodyElement = <span>{bodyText}</span>;
+      const { queryByText } = renderToast({
+        open: true,
+        body: bodyElement,
+        variant: ToastVariants.success,
+      });
+
+      const bodySpan = queryByText(bodyText);
+      expect(bodySpan).toBeInTheDocument();
```
lazytype, line 220: \+

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,304 @@
...
+    position: absolute;
+    top: ${spacing[2]}px;
+    right: ${spacing[2]}px;
+    transition: color 0.15s ease-in-out;
+  `,
+};
+
+const ToastVariants = {
+  success: 'success',
+  progress: 'progress',
```
lazytype, line 64: usually our variant variables start with a capital letter

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,304 @@
...
+
+  dismissButton: css`
+    position: absolute;
+    top: ${spacing[2]}px;
+    right: ${spacing[2]}px;
+    transition: color 0.15s ease-in-out;
+  `,
+};
+
+const ToastVariants = {
```
lazytype, line 62: `Button` and other components export their variants as `Variant`.

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,304 @@
...
+   * Required style variant to render the Toast as.
+   */
+  variant: ToastVariants;
+
+  /**
+   * Optional number between 0 and 1 that sets the progress bar's progress. Note that the progress bar is only rendered when the Toast variant is set to `'progress'`.
+   *
+   * **Default:** `1`
+   */
+  progress?: number;
```
lazytype, line 170: we can make it a type error to pass a `progress` prop without the progress variant. something like this:
```tsx
import { OneOf } from '@leafygreen-ui/lib';

type ToastProps = {
  ...
} & OneOf<{
  variant: 'progress';
  progress?: number;
}, {
  variant: 'success';
}>;
```

packages/toast/src/Toast.tsx
```diff
@@ -0,0 +1,304 @@
...
+   */
+  close?: React.MouseEventHandler;
+}
+
+function Toast({
+  title,
+  body,
+  className,
+  variant,
+  progress = 1,
```
lazytype, line 190: can this be `1.0` just so it's more visually clear that it's 1.0 out of 1.0, rather than 1 out of 100

packages/toast/src/ToastProgressBar.tsx
```diff
@@ -0,0 +1,80 @@
...
```
lazytype, line 53: can we clamp `progress` to be between 0 and 1?

scripts/package-builder.js
```diff
@@ -112,10 +112,10 @@ npm install @leafygreen-ui/${PACKAGE_LC}
 `;
 
 const rootFile = `
-import React from 'react'
+import React from 'react';
 
 export default function ${PACKAGE_UC}({}) {
-  return(<div>your content here</div>)
+  return(<div>your content here</div>);
```
lazytype, line 118: while you're at it could you add a space after `return` and remove the parens lol

lazytype:
left some non-blocking feedback

packages/toast/src/Toast.spec.tsx
DesignerDave: Fair point!

packages/toast/src/Toast.spec.tsx
DesignerDave: Good point

packages/toast/src/Toast.tsx
DesignerDave: We could, but I consciously decided not to do this. In the case the beta will be used, they don't have an actual progress amount for the initial version, so they will be. hard-coding it as `1`. To keep switching between the progress and success variants easy, and the API simpler, I opted to allow the prop regardless of the variant since the downside seemed negligible.

packages/toast/src/Toast.tsx
DesignerDave: Good point. I'll update this, but I'm not gonna export it for the time being.

packages/toast/src/Toast.tsx
DesignerDave: Sure thing – can't see a reason why not

packages/toast/src/ToastProgressBar.tsx
DesignerDave: Sure! It shouldn't make a difference in the display of the component, but could make the component slightly more resilient in case we accidentally break the CSS.

scripts/package-builder.js
DesignerDave: Hah sure

PR #573, opened 2020-08-31
Use Rollup to generate UMD and ESM files

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Export ES Module version of packages](https://jira.mongodb.org/browse/PD-328)

## 🛠 Types of changes

- [X] Tooling (updates to workspace config or internal tooling)

## ✅ Checklist


storybook/webpack.config.js
```diff
@@ -1,8 +1,143 @@
-module.exports = ({ config, mode }) => {
```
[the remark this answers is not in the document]
hswolff: Nice, ty!

rollup.config.js
```diff
@@ -0,0 +1,173 @@
...
+    return [];
+  }
+
+  return fs
+    .readdirSync(directory)
+    .filter(file => /\.tsx?$/.test(file))
+    .map(file => path.resolve(directory, file));
+}
+
+const allPackages = getAllPackages('../../packages');
```
hswolff, line 58:
```suggestion
const allPackages = getAllPackages(path.join(__dirname, 'packages'));
```

rollup.config.js
```diff
@@ -0,0 +1,173 @@
...
+   *   - lodash/*
+   *   - highlight.js
+   *   - create-emotion
+   *   - create-emotion-server
+   *   - react-transition-group
+   **/
+};
+
+allPackages.forEach(packageName => {
+  globals[packageName] = packageName;
```
hswolff, line 84: Feels like we should namespace this to `global.LG[packageName]`? So we don't pollute the global name space?

rollup.config.js
```diff
@@ -0,0 +1,173 @@
+import fs from 'fs';
```
hswolff, line 1: We'll need to make this a major version change, right?

hswolff (changes requested):

rollup.config.js
lazytype: how so? dependees don't need to change any code and nothing should break for them.

rollup.config.js
lazytype: `globals` is a local variable defined above, not to be confused with `global`.

rollup.config.js
hswolff: The resulting artifact is produced through an entirely different build system. Can you guarantee it'll cause no regressions for applications?

tl;dr: I think it's safer and easier to make this a breaking change for all packages.

rollup.config.js
hswolff: Ah, ty.

rollup.config.js
lazytype: It sounds like you want to use major version to communicate "this should be tested more thoroughly by your app before upgrading", but that's not really what major version means if we follow semver. per semver it just means "the api changed so review the documented changes and if you update your usage accordingly then things should just work". 
i.e. the version number only indicates _intentional_ breakages, not the risk-level associated with unintentional breakages.

I'm ok with wanting version numbers' meaning to extend beyond the semver contract, but I want to be explicit about that, especially since this is not documented. Our readme merely states:
>We follow semver conventions for versioning, so each change will either be major, minor, or patch.

which is not entirely accurate if we're extending the definition of major version to also mean "higher likelihood of undocumented breakages so test more thoroughly".

hswolff: What would you suggest we release this as?

lazytype: If you consider adding ESM module support as a feature, then following semver this would be a minor release

hswolff: My read of semver is that it's not strict with when a major version is released.

[Line 8 says](https://semver.org/#spec-item-8)
> Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. It MAY also include minor and patch level changes. Patch and minor version MUST be reset to 0 when major version is incremented.

hswolff: I also want @DesignerDave to gain consensus on this as well, so ping on that.

hswolff (approved):
This looks AWESOME and LGTM! Waiting on Dave to confirm inline questions!

rollup.config.js
DesignerDave: We've done major versions for similar changes where we think there's a high likelihood of something breaking down the line. By my understanding of these changes, we're no longer building the `index.bundle.js` files with webpack, which will cause very different output builds, so I think this is a change with a high-risk of breaking consuming applications. I'm definitely with Harry on this, and think we should be conservative with our versioning here, and in general.

packages/box/src/index.ts
```diff
@@ -1,2 +1,2 @@
-export { BoxProps, ExtendableBox } from './Box';
+export type { BoxProps, ExtendableBox } from './Box';
```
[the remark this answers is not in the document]
DesignerDave: Something about the build didn't like it, IDR what part.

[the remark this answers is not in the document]
DesignerDave: For the sake of keeping this PR scoped to the build changes, could we omit this change (and the similar other changes) @lazytype and do this later? Or is this essential to the change to rollup?

DesignerDave (approved):
Left a few comments, but overall this looks good. Little bit squeamish about these sort of overarching changes, but this seems to work in my testing against SSR environments, the output looks good, and I'm having a hard time finding places this might break.

If you could, would you mind working with a few folks to link these changes up and make sure their applications and test suites all still run? Maybe @hswolff could link up MMS if he hasn't already.

packages/box/src/index.ts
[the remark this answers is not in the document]
lazytype: I think I had to make these changes because Rollup had trouble resolving type imports.

rollup.config.js
```diff
@@ -0,0 +1,173 @@
...
+    .readdirSync(directory)
+    .filter(file => /\.tsx?$/.test(file))
+    .map(file => path.resolve(directory, file));
+}
+
+const allPackages = getAllPackages(path.resolve(__dirname, 'packages'));
+const directGlyphImports = getDirectGlyphImports();
+
+// Mapping of packages to the `window` property they'd be
+// bound to if used in the browser without a module loader.
```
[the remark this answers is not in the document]
lazytype: >Is there any chance by setting things like window.React ourselves, that we break references in consuming applications?

So this config just specifies where these variables are attempted to be read from and shouldn't affect how they are set.

>Additionally, why include something like highlightjs-graphql when highlight.js itself still needs a module loader?

We effectively gain less warnings from Rollup for each global we explicitly specify.

PR #583, opened 2020-09-02
Generate SVG components and include them as part of the build

lazytype:
### Changes
- the icon pre-build script no longer generates output to `packages/icon/dist/` folder. Now the script will generate each glyph component as a fully-typed React component under the `packages/icon/src/generated/` folder.
- the icon pre-build script has been updated to call async file APIs to increase the chance of getting I/O parallelism.
- the webpack config has been updated so that any files in a package's respective `src/generated` folder will each be built to the package's `dist/` folder. This preserves compatibility with the current direct glyph import style, e.g. `import VerticalEllipsisIcon from '@leafygreen-ui/icon/dist/VerticalEllipsis';`

### Implications
- files that directly import glyph components can now be typechecked without depending on the glyph components being built. That means one can now run `yarn ts` without needing to run `yarn ts:build` first.
- since the glyph components are now built in the same way as any other components, when we move to rollup and rollup becomes responsible for building UMD and ESM modules, we won't need the `yarn ts:build` command at all anymore. TypeScript will only be used for typechecking.

### Follow-up work
- This change also adds a tests that can catch when a generated file has been manually modified or when a generated file needs to be regenerated (i.e. the svg icon file has changed). This is done by generating a checksum hash within each file. The code generation and test logic could at some point be re-used and pulled out into a library for future code generation needs.

## 🛠 Types of changes

## ✅ Checklist


packages/icon/src/Icon.spec.tsx
```diff
@@ -1,43 +1,52 @@
...
+   * './glyphs/' contains an index.ts file, so we filter out
+   * anything that's not an svg.
+   */
+  .filter(path => /.*\.svg$/.test(path))
+  .map(fileName => path.resolve(__dirname, './glyphs/', fileName));
+
+const generatedFilesDirectory = path.resolve(__dirname, './generated');
+const baseNameToGeneratedFilePath: Record<string, string> = {};
 
+fs.readdirSync(generatedFilesDirectory).forEach(filePath => {
```
hswolff, line 30: May be nice to move this out of top of the module, so it's not done on immediate import of the file.

hswolff: Ah...I see cuz of the tests you have. W/e, not a big deal.

packages/icon/src/Icon.spec.tsx
```diff
@@ -154,4 +163,57 @@ describe('Generated glyphs', () => {
...
+        ] = /^\/\*.*@checksum ([^\n]*)\n.*\*\/\n(.*)$/s.exec(
+          generatedFileContents,
+        )!;
+
+        const expectedChecksum = createHash('md5')
+          .update(svgFileContents)
+          .update(checkedContents)
+          .digest('hex');
+
+        expect(checksum).toEqual(expectedChecksum);
```
hswolff, line 215: When this errors can we either inline or log out documentation on how to regenerate the files?

hswolff (approved):
I've made my peace with this and am excited about what this unblocks :)

DesignerDave (approved):
Let's keep close tabs on this to see if we run into roadbumps with it, and we should consider ways to eliminate checking-in these built files in the future.

For the time being, I'm cool with unblocking some exciting work 😄

PR #584, opened 2020-09-02
Auto deploy with Realm

brookescarlett:
- I tested this by manually uploading the contents of hosting/files to Realm and it worked, so hopefully this solves for the automatic portion :)


.eslintignore
```diff
@@ -3,3 +3,5 @@ dist
 lib
 storybook/public
 website/.next
+website/out
+website/hosting
```
lazytype, line 7: do these need to be added to .gitignore?

hswolff: yeah...ideally we won't commit these values....

brookescarlett:
Realm needs the bundled files inside of hosting/files and I believe that those need to be committed in order to deploy directly from GitHub. If we want to gitignore the /hosting directory, then we can manually deploy using Realm's CLI. Would people prefer that?

brookescarlett:
Right now Realm will deploy any time code is pushed to master's website directory

lazytype:
>Right now Realm will deploy any time code is pushed to master's website directory

So Realm is configured specifically to deploy from `mongodb/leafygreen-ui` is that correct? Is there any way to manually deploy code from the repo? Just want to be 200% sure that someone cloning the repo can't deploy any code they want.

brookescarlett:
Yeah Realm is configured specifically to deploy from `mongodb/leafygreen-ui`. I can confirm that because I had to get specific permissions from this repo in order to be able to do that and listed in the options of repo's to select was my local fork of this repo as well.

We can manually deploy code instead of having this automatically deploy, 100%. And, if we do that, then I don't believe that we'd need to commit the files inside of `hosting/files` or `out`

.eslintignore
lazytype: I've been thinking about it more and I'm wondering if we can take an approach a bit more like our Storybook instance. Could these files be pushed to a different branch rather than `master` in the same way that we publish storybook to the `gh-pages` branch?

brookescarlett: yes that's definitely possible as well

brookescarlett: so we'd monitor for changes to the website directory in some other branch? and just not keep website up to date with master?

lazytype: right, it would give us some leeway so that we can make changes on `master` that aren't necessarily ready to be deployed yet.

brookescarlett: yeah i buy that

PR #589, opened 2020-09-03
Tooltip should not navigate when trigger has href

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Tooltips that have usePortal set to false shouldn't propagate click events to their parent](https://jira.mongodb.org/browse/PD-719)

## 🛠 Types of changes

## ✅ Checklist


brookescarlett:
should we use ts to throw an error when an href is passed so we're letting the consumer know that the property isn't valid?

lazytype:
> should we use ts to throw an error when an href is passed so we're letting the consumer know that the property isn't valid?

The property is valid for the trigger itself, e.g. an IconButton, just not the tooltip.

.changeset/fast-kiwis-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/tooltip': minor
```
DesignerDave, line 2: I think this is technically breaking. Otherwise, this looks good to go!

DesignerDave (approved):

PR #593, opened 2020-09-09
Optimize table sorting performance

lazytype:
## ✍️ Proposed changes

We currently re-render every row when changing the sort order because we modify the underlying table data. This change prevents re-renders by:
- memoizing the rendered rows when the data hasn't changed.
- deferring sorting until the render phase and ensuring the underlying data doesn't change.
- ensuring that even when rows are re-sorted, they do not get re-created or re-rendered.

Edit: These are before the context change; the current change is even faster than in the screen captures
Before and after comparison (700 rows, re-rendered elements highlighted)
  

Github doesn't animate the GIFs at the original speed so here are video links:
[before](https://jmp.sh/CKhJ4Fl) [after](https://jmp.sh/0SnWKs2)

## 🛠 Types of changes

## ✅ Checklist


packages/table/src/Table.tsx
```diff
@@ -15,7 +15,6 @@ const tableStyles = css`
 
 interface TableRowInterface<Shape = {}> {
   datum: Shape;
-  index?: number;
```
lazytype, line 18: this prop was undocumented so I don't consider this to be a breaking change

packages/table/src/TableBody.tsx
```diff
@@ -1,17 +1,57 @@
...
+  let compareFn: ((a: Shape, b: Shape) => number) | undefined;
 
-  return (
-    <tbody>{data.map((datum, index) => children({ datum, index }))}</tbody>
-  );
+  if (sort) {
+    const { direction, accessorValue } = sort;
+
+    if (direction && accessorValue) {
+      compareFn = getDataComparisonFunction({ direction, accessorValue });
```
hswolff, line 46: Would we want to wrap this in `useMemo` to make the comparison function stable?

packages/table/src/TableBody.tsx
```diff
@@ -1,17 +1,57 @@
...
+    }
+  }
+
+  const renderFunction = useCallback(datum => children({ datum }), [children]);
```
hswolff, line 50: Does this do anything? Isn't children always going to be a new instance as we instruct users to pass in an inline function?

packages/table/src/TableBody.tsx
```diff
@@ -1,17 +1,57 @@
...
 import { TableProps } from './Table';
 
 type TableBodyProps<Shape> = Pick<TableProps<Shape>, 'children'>;
 
+/**
+ * Hook that avoids re-rendering children when the data used to
+ * render them hasn't changed, even if their ordering has changed.
+ */
+function useRenderedChildren<Props>(
+  propList: Array<Props>,
```
hswolff, line 12: I'm confused by this variable name. Isn't this `data`? Why the rename?

packages/table/src/TableBody.tsx
```diff
@@ -1,17 +1,57 @@
...
+      propList.map((props, index) => [
+        props,
+        <React.Fragment key={index}>{renderFunction(props)}</React.Fragment>,
+      ]),
+    [propList, renderFunction],
+  );
+
+  return useMemo(() => {
+    if (!compareFn) {
+      return pairs.map(([_, renderResult]) => renderResult);
```
hswolff, line 27: Can we think of a way to remove this additional loop?

Hosting two arrays of the datum and the renderResult?

lazytype: we would still need to build some form of mapping from props to renderResult in order to derive the sorted renderResults. I suppose I can rewrite this to use `Map` to make that more obvious.

packages/table/src/TableBody.tsx
lazytype: the variable name is generalized in case this hook is ever used in another context

packages/table/src/TableBody.tsx
lazytype: children doesn't change if the parent doesn't re-render. changing the sort order doesn't cause the parent to re-render.

packages/table/src/TableBody.tsx
hswolff: I struggle to imagine what other context could make use of this hook. For now can we rename to `data` (and `datum` inside the loop respectively) to make reading it in the context of Table clear? If we see a use for it elsewhere we can generalize it then.

packages/table/src/TableBody.tsx
hswolff: In practice this is changing on every render I imagine? This is to catch those instances where we can?

Does this make a material difference if it's removed?

hswolff (changes requested):

packages/table/src/TableBody.tsx
lazytype: no, in general it will not change on every render as I already explained and yes, it makes the difference between re-rendering every row and not.

packages/table/src/TableBody.tsx
lazytype: you could use this hook in any component that has a render prop whether that prop happens to be `children` or not. You can also use it anywhere a `map`, `forEach`, or `for (...)` is used to compute children or list of react elements. 

I don't really understand the line of reasoning that leads to over-specifying the naming. to me the whole point of extracting something out into a function rather than inlining it is to abstract away a high-level operation. the fact that a `prop` is a `datum` is just an implementation detail of `TableBody`. 
this
```ts
   const renderFunction = useCallback(datum => children({ datum }), [children]); 
   const rows = useRenderedChildren(data, renderFunction, compareFn); 
```
could have as easily also been
```ts
   const rows = useRenderedChildren(data.map(datum => ({datum})), children, compareFn); 
```

if it's not clear what the hook definition does in a vacuum, then I'd prefer feedback on how to make it more clear.

hswolff: Nothing's done in a vacuum, this is being written in the context of table so keeping consistent with those variables increases the overall readability of Table as a hole. 

We don't know how this will be applied generically (in a vacuum), however I do doubt a generic form would make use of a `compareFn`. That seems very Table specific itself.

I value readability highly, and the better we can serve that interest then the stronger the codebase is as a whole.

packages/table/src/TableBody.tsx
hswolff: For a contrived example:

```jsx
function MyPage() {
  const childrenFn = () => {};
  return (
    <Button onClick={increment} />
    <Table>{childrenFn}</Table>
  )
}
```

If you click `increment` that'll cause `MyPage` to render and `childrenFn` to be re-created, thus resulting in a new instance and needing to call this `useCallback` hook written here, right? Just making sure we're aligned on understanding.

lazytype: in this example `MyPage` is rerendered and `childrenFn` is recreated because the counter state belongs to `MyPage`. the data in a table belongs to the table's context and changing the sort order is also belongs to the table's context (and doesn't even touch the row data).

hswolff: Aha! So this is specific for the sorting case, which yes, is a great one to cover. Ty for venturing down this explanation path!

packages/table/src/TableBody.tsx
lazytype: >Nothing's done in a vacuum, this is being written in the context of table so keeping consistent with those variables increases the overall readability of Table as a hole.

but the calculus would be different if I had just defined the hook in the hooks package. then the emphasis of the review would be more so on improving its standalone clarity. I chose not to define it separately to reduce the scope of the PR change and reduce the burden of needing to unit test it more thoroughly.

> That seems very Table specific itself.

That's why I made `compareFn` optional, and thus ignorable, part of the API rather than `((a: Props, b: Props) => number) | undefined`. I could technically have split that part out but it would've been pretty unnatural since it either would have to reconstruct the mapping given the list of props and react nodes or have `useRenderedChildren` return the mapping.

hswolff: Seems like we're splitting hairs here? Also seems like the desire is to have this custom hook be reviewed independently is in tension with it needing to still be aware of Table specific needs, I.e. `compareFn`. 

I think what I found jarring when reading this was the introduction of this `propList` variable which is out of context for the rest of the Table component. It causes some mental strain to map what variables we're talking about when I made the jump into this function.

I don't think we're ready to pull this out to the hooks package, so my main ask is that we embrace its locality within Table and keep its variable naming consistent with the rest of Table.

Also - did moving to Map have any impact on performance? (Related - we should get some performance scaffolding in place at some point?).

packages/table/src/TableBody.tsx
```diff
@@ -1,17 +1,58 @@
...
+    }
+
+    return [...data].sort(compareFn).map(datum => resultMap.get(datum));
+  }, [data, resultMap, compareFn]);
+}
+
 function TableBody<Shape>({ children }: TableBodyProps<Shape>) {
   const {
-    state: { data },
+    state: { data, sort },
```
hswolff, line 38: Random thought - would we gain anything moving sort to its own context?

packages/table/src/TableBody.tsx
```diff
@@ -1,17 +1,58 @@
...
+  data: Array<Datum>,
+  renderFunction: React.FunctionComponent<Datum>,
+  compareFn?: (a: Datum, b: Datum) => number,
+): Array<React.ReactNode> {
+  const resultMap = useMemo(() => {
+    const resultMap = new Map<Datum, React.ReactNode>();
+    data.forEach((datum, index) =>
+      resultMap.set(
+        datum,
+        <React.Fragment key={index}>{renderFunction(datum)}</React.Fragment>,
```
hswolff, line 21: We don't sadly have anything more stable than `index` to use as the key, right? Cuz that would gain us some render optimizations?

packages/table/src/TableBody.tsx
```diff
@@ -1,17 +1,58 @@
...
+      ),
+    );
+    return resultMap;
+  }, [data, renderFunction]);
+
+  return useMemo(() => {
+    if (!compareFn) {
+      return Array.from(resultMap.values());
```
hswolff, line 29: niiiice

hswolff (approved):
I dig it!

packages/table/src/TableBody.tsx
lazytype: I checked and yeah, `Row` subscribes to table context and doesn't use `sort`, so that's a lot of re-renders we can prevent.

packages/table/src/TableHeader.tsx
```diff
@@ -120,13 +119,16 @@ function TableHeader<Shape>({
...
-        type: TableActionTypes.SortTableData,
-        payload: {
-          columnId: index,
-          accessorValue: normalizedAccessor,
-        },
-      });
+      setSort(prevSort => ({
+        columnId: index,
+        direction:
+          index === prevSort?.columnId
```
lazytype, line 125: this also fixes the bug I mentioned in chat

packages/table/src/Table.tsx
```diff
@@ -43,8 +43,10 @@ export default function Table<Shape>({
         className={cx(tableStyles, className)}
         {...rest}
       >
-        <TableHead columns={columns} />
```
hswolff, line 46: Should we move `<TableProvider>` in here as well? As it has no impact to `<table`?

hswolff (approved):
Love it! Wonderful work @lazytype!

PR #595, opened 2020-09-09
PD-786: Update Tabs component

brookescarlett:
## ✍️ Proposed changes

Updates Tab component to match design spec

🎟 _Jira ticket:_ [PD_786](https://jira.mongodb.org/browse/PD-786)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

Screen Shot 2020-09-09 at 11 40 07 AM
Screen Shot 2020-09-09 at 11 40 24 AM


packages/tabs/src/Tabs.tsx
```diff
@@ -176,13 +224,15 @@ function Tabs({
 
     let computedX = 0;
 
-    for (let i = 0; i < currentIndex; i++) {
-      computedX += tabListChildren[i].scrollWidth;
+    for (let i = 0; i < current; i++) {
+      computedX += tabListChildren[i].clientWidth;
```
brookescarlett, line 228: scrollWidth doesn't work when we truncate TabTitle's that are longer than 300px

packages/tabs/src/Tabs.tsx
```diff
@@ -4,41 +4,84 @@ import { css, cx } from '@leafygreen-ui/emotion';
 import { uiColors } from '@leafygreen-ui/palette';
 import { isComponentType, keyMap } from '@leafygreen-ui/lib';
 import { useEventListener } from '@leafygreen-ui/hooks';
+import { useUsingKeyboardContext } from '@leafygreen-ui/leafygreen-provider';
 import TabTitle from './TabTitle';
 import omit from 'lodash/omit';
 
-const borderHeight = 3;
+const Variant = {
```
lazytype, line 11: I've seen us define different enums for different components. In some cases we have `'default' | 'light`, but I've also seen `'dark' | 'light'`, `'default' | 'dark'`, and `'dark' | 'light'`. I imagine we'll want to have a standard convention if we'll eventually have some sort of provider that globally selects dark or light mode. just curious on what your thoughts are on this.

brookescarlett: So @DesignerDave's initial distinction between variants was if the colors are lighter than the default its considered `light`. I definitely agree with you in terms of our need to standardize this, and I think it could make sense to consider adding a `mode` prop and slowly deprecate variants when they only refer to `light`/`dark` in their discrepant ways. What do you think?

brookescarlett:
@DesignerDave RFAL, but do you think we should major version bump given all of the changes

DesignerDave:
Yes, for large design changes, we should do a major bump

DesignerDave:
Looks like the styles are broken. There's a background color that appears behind all the tabs, when there shouldn't be.

Screen Shot 2020-09-17 at 3 08 53 PM

One other thing I noticed when playing with the component is that the underline when a tab is active shrinks on hover. We should keep the scaling of the underline at 1 on hover when the tab is active.

DesignerDave (approved):

PR #598, opened 2020-09-10
PD-856: Adds xsmall size to radio

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-856](https://jira.mongodb.org/browse/PD-856)

## 🛠 Types of changes

## ✅ Checklist


DesignerDave (changes requested):
It appears that this PR broke the default size of the input itself.

Before:
Screen Shot 2020-09-17 at 2 38 47 PM

After:
Screen Shot 2020-09-17 at 2 38 47 PM

Could we also decrease the margin between the input and text for the xsmall size? 4px seems to work well:
Screen Shot 2020-09-17 at 2 40 41 PM

Otherwise, looks good!

DesignerDave:
One more comment – I seem to be getting proptypes errors. I don't think `xsmall` is being included.

brookescarlett:
Rookie mistake forgetting you cant check a value against a || b, has to be x === a || x === b!

brookescarlett:
(The default mode is fixed now)

packages/radio-group/src/Radio.tsx
```diff
@@ -295,16 +294,19 @@ export type RadioProps = Omit<React.ComponentPropsWithoutRef<'input'>, 'size'> &
 function Radio({
   children,
   className,
-  onChange,
+  onChange = () => {},
```
brookescarlett, line 297: Adding a default here because on the first render, there's no onChange prop and React complains that we're setting checked in the HTML without an onchange rendering it a `readonly` field, which is not true.

packages/radio-group/src/Radio.tsx
```diff
@@ -295,16 +294,19 @@ export type RadioProps = Omit<React.ComponentPropsWithoutRef<'input'>, 'size'> &
...
   value,
   disabled,
   id,
   name,
   darkMode,
-  checked = false,
+  checked,
```
brookescarlett, line 303: Removing default here, such that we're able to check for the existence of the `checked` prop when deciding if the component is controlled or uncontrolled.

packages/radio-group/src/RadioGroup.spec.tsx
```diff
@@ -69,7 +69,7 @@ describe('packages/radio-group', () => {
       fireEvent.click(secondInput);
 
       expect(firstInput.getAttribute('aria-checked')).toBe('true');
-      expect(secondInput.getAttribute('aria-checked')).toBe('false');
+      expect(secondInput.getAttribute('aria-checked')).toBe(null);
```
brookescarlett, line 72: Now null because we unset the default false

packages/radio-group/src/RadioGroup.tsx
```diff
@@ -75,17 +75,20 @@ function RadioGroup({
   value: controlledValue,
   name: nameProp,
 }: RadioGroupProps) {
-  const isControlled = controlledValue != null;
+  let isControlled = controlledValue != null ? true : false,
```
brookescarlett, line 78: Previously strictly determining if a Radio is controlled based on the value prop. Since we're already iterating through the Radios to find a defaultChecked, makes sense to search for the `checked` prop, which will also let us know that the component is controlled.

DesignerDave (approved):

PR #599, opened 2020-09-10
PD-849 Create Copyable component

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Create Copyable React component](https://jira.mongodb.org/browse/PD-849)

## 🛠 Types of changes

## ✅ Checklist


packages/copyable/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+## Properties
+
+| Prop          | Type                   | Description                                   | Default     |
+| ------------- | ---------------------- | --------------------------------------------- | ----------- |
+| `label`       | `string`               | Label text for the copyable contents.         |             |
+| `description` | `string`               | Further text to describe the contents.        | `undefined` |
+| `children`    | `string`               | The text that will be copied.                 |             |
+| `variant`     | `'default'`, `'dark'`  | Sets the style variant.                       | `'default'` |
```
brookescarlett, line 83: Should we update this to darkMode, since that'll be our convention moving forward

packages/copyable/README.md
```diff
@@ -0,0 +1,85 @@
...
+| `size`        | `'default'`, `'large'` | Display size of the copyable text.            | `'default'` |
+| `copyable`    | `boolean`              | Whether or not a copy button should be shown. | `true`      |
```
brookescarlett, line 85: If this is a copyable component, I'd expect the copy functionality to always be supported. Is there a reason why this should be toggle-able?

brookescarlett: Ah, seeing that this is linked to mobile. But wouldn't the usecase just become a <Code /> component, rather than a Copyable one?

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,283 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import ClipboardJS from 'clipboard';
+import { transparentize } from 'polished';
+
+import Button, { Variant as ButtonVariant } from '@leafygreen-ui/button';
```
brookescarlett, line 6: Ah! I wonder if we should be more proactive about scoping things such as variant/size local to the component, such that we can be more flexible with using these types across components. Probably nothing actionable here in this PR, just thinking

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,283 @@
...
+      }, 1500);
+
+      return () => {
+        clearTimeout(timeoutId);
+        if (!destroyed) {
+          clipboard.destroy();
+        }
+      };
+    }
+  }, [buttonRef, children, copied]);
```
hswolff, line 230: Does buttonRef have any impact here?

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,283 @@
...
+import { uiColors } from '@leafygreen-ui/palette';
+import Tooltip, {
+  Align,
+  Justify,
+  TriggerEvent,
+  Variant as TooltipVariant,
+} from '@leafygreen-ui/tooltip';
+import { InlineCode } from '@leafygreen-ui/typography';
+
+export const Variant = {
```
brookescarlett, line 18: Mentioned above, but let's update this to the darkMode prop

packages/copyable/package.json
```diff
@@ -0,0 +1,24 @@
...
+      "access": "public"
+    },
+    "dependencies": {
+      "@leafygreen-ui/button": "^7.0.0",
+      "@leafygreen-ui/emotion": "^2.0.2",
+      "@leafygreen-ui/icon": "^6.5.0",
+      "@leafygreen-ui/palette": "^2.0.2",
+      "@leafygreen-ui/tooltip": "^3.3.7",
+      "@leafygreen-ui/typography": "^4.2.2",
+      "clipboard": "^2.0.6"
```
hswolff, line 22: Wonder if we should set this to `^2.0.0` to have a bigger chance of having this package deduped from consuming applications. For example in Cloud we have this set to `2.0.6` so we're good, but if we had it set to `2.0.5` it would have resulted in a duped dep.

packages/copyable/src/Copyable.story.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+function renderCopyable(
+  props: Omit<
+    React.ComponentProps<typeof Copyable>,
+    'children' | 'label' | 'description' | 'variant'
+  > = {},
+) {
+  const label = text('Label', 'Label');
+  const description = text('Description', '') || undefined;
+  const children = text('Text', 'npm install @leafygreen-ui/copyable');
```
hswolff, line 14: meta :D

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,283 @@
...
+          clipboard.destroy();
+        }
+      };
+    }
+  }, [buttonRef, children, copied]);
+
+  return (
+    <>
+      {label && (
+        <div className={cx(labelStyle, css(`color: ${colorSet.label};`))}>
```
brookescarlett, line 235: I might be missing something but I think you can inline this css`` without the parens

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,283 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import ClipboardJS from 'clipboard';
+import { transparentize } from 'polished';
+
+import Button, { Variant as ButtonVariant } from '@leafygreen-ui/button';
+import { css, cx } from '@leafygreen-ui/emotion';
+import CopyIcon from '@leafygreen-ui/icon/dist/Copy';
```
hswolff, line 8: Do we think it'd be a good idea to make a lint rule to enforce these imports to reduce the chance of regression?

I think https://eslint.org/docs/rules/no-restricted-imports would make it easy.

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,283 @@
...
+        onClick={() => setCopied(true)}
+      >
+        <CopyIcon size="large" className={iconStyle} />
+        Copy
+      </Button>
+    );
+
+    let tooltipVariant: TooltipVariant;
+
+    switch (variant) {
```
brookescarlett, line 179: This will be very cool when we have a provider from where to pull whether or not the consumer is in darkMode

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,283 @@
...
+      </div>
+    );
+  }
+
+  useEffect(() => {
+    if (!buttonRef) {
+      return;
+    }
+
+    const clipboard = new ClipboardJS(buttonRef, {
```
hswolff, line 210: Every instance of ClipboardJS [creates new listeners](https://github.com/zenorocha/clipboard.js/blob/master/src/clipboard.js#L18). We should destroy and cleanup a previous instance before we create a new one.

packages/copyable/src/Copyable.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+            </Copyable>,
+          );
+
+          expect(queryByText('Copied!')).not.toBeInTheDocument();
+
+          const copyButton = getByText('Copy');
+          expect(copyButton).toBeVisible();
+
+          await Context.within(
+            jest.spyContext(ClipboardJS.prototype, 'onClick'),
```
hswolff, line 44: What does the usage of Context.within provide here? Just an easy way to cleanup the onClick spy?

packages/copyable/src/Copyable.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+              expect(spy).not.toHaveBeenCalled();
+              fireEvent.click(copyButton);
+              await waitFor(() => expect(spy).toHaveBeenCalled());
+            },
+          );
+
+          await waitFor(() => expect(getByText('Copied!')).toBeVisible());
+
+          // Tooltip should remain visible for a while
+          await new Promise(resolve => setTimeout(resolve, 1000));
```
hswolff, line 55: Do we have an easy way to share test utils? It'd be nice to have a sleep util.

packages/copyable/README.md
lazytype: `<Code />` actually also has a `copyable` property. The main difference between `<Copyable />` and `<Code />` I see is that `<Copyable />` isn't necessarily code, e.g. it could be an API token, so semantically I think we still would prefer to use `<Copyable />` for those use-cases on mobile.

packages/copyable/package.json
lazytype: I just chose the version that the `Code` component was using. I'd prefer to keep those in sync if possible, even if it means changing them to both to`^2.0.0`. though I don't quite understand why yarn/node wouldn't just use the `2.0.6` for the `2.0.5` dependency too.

packages/copyable/src/Copyable.spec.tsx
lazytype: also makes the test more self-contained since the spy will be cleaned up even if the test fails

packages/copyable/src/Copyable.spec.tsx
lazytype: we have `@leafygreen-ui/testing-lib`, but theoretically I think we could just have a folder that contains unpackaged files that can be imported directly. maybe worth exploring esp. if we don't want to necessarily maintain testing-lib

packages/copyable/src/Copyable.tsx
lazytype: what restriction(s) would you have in mind?

packages/copyable/src/Copyable.tsx
lazytype: It'll be `undefined` on the initial render

brookescarlett:
This isn't necessary in this PR, but there's been a request to make the hex codes copyable in our palette storybook. If you have time and this isn't scope creep, could make sense to use this component to do that.

DesignerDave (changes requested):
Just played with the component briefly, and had two notes:

1. The description and label for the component don't wrap to the maximum width of the copyable element. Is this intentional? cc @audroue 

Screen Shot 2020-10-07 at 3 13 17 PM

2. Can we make the border radius of the top and bottom right corners of the button match the input's border radius?

packages/copyable/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### NPM
+
+```shell
+npm install @leafygreen-ui/copyable
+```
+
+## Example
+
+```tsx
```
brookescarlett, line 23: can we keep this as `js` for now. Need this to be standardized across readmes for .design

packages/copyable/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Properties
+
+| Prop          | Type                   | Description                                                   | Default     |
+| ------------- | ---------------------- | ------------------------------------------------------------- | ----------- |
+| `label`       | `string`               | Label text for the copyable contents.                         |             |
+| `description` | `string`               | Further text to describe the contents.                        | `undefined` |
+| `children`    | `string`               | The text that will be copied.                                 |             |
+| `darkMode`    | `boolean`              | Determines whether or not the component appears in dark mode. | `false`     |
+| `size`        | `'default'`, `'large'` | Display size of the copyable text.                            | `'default'` |
```
brookescarlett, line 84: \|

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,270 @@
...
+`;
+
+const copyableContainerStyle = css`
+  height: 36px;
+`;
+
+const codeStyle = css`
+  display: inline-flex;
+  align-items: center;
+
```
brookescarlett, line 86: rm

packages/copyable/src/Copyable.tsx
```diff
@@ -0,0 +1,270 @@
...
+  height: 100%;
+  width: 100%;
+  padding-left: 12px;
+  border: 1px solid;
+  border-radius: 4px;
+
```
brookescarlett, line 92: ditto

DesignerDave (approved):
Noticing a couple overflow issues with the rounded corners:

**Static**
Screen Shot 2020-11-09 at 3 13 16 PM

**Hover**
Screen Shot 2020-11-09 at 3 13 22 PM

Otherwise these LGTM. Giving this an approval so you can merge once these are fixed.

PR #606, opened 2020-09-11
PD-861: Adds darkMode to Toggle component

brookescarlett:
## ✍️ Proposed changes

Refactors Toggle from class to functional component
Deprecates variant, adds darkMode

🎟 _Jira ticket:_ [PD-861](https://jira.mongodb.org/browse/PD-861)

## 🛠 Types of changes

## ✅ Checklist


packages/toggle/src/Toggle.tsx
```diff
@@ -389,122 +384,94 @@ const getStatefulStyles = (states: StateForStyle) => ({
...
+  disabled = false,
+  className,
+  onChange: onChangeProp = () => {},
+  checked: checkedProp,
+  name,
+  id,
+  ...rest
+}: ToggleProps) {
+  const [checked, setChecked] = useState(false);
+  const checkboxId = id || `toggle-${idAllocator.generate()}`;
```
lazytype, line 409: I know the code was already like this, but this could actually be changed since `IdAllocator` already adds a prefix.
```suggestion
  const checkboxId = id || idAllocator.generate();
```

packages/toggle/src/Toggle.tsx
```diff
@@ -389,122 +384,94 @@ const getStatefulStyles = (states: StateForStyle) => ({
...
+  const normalizedChecked = checkedProp || checked;
```
lazytype, line 410: I think this should be `??` to maintain the current behavior

packages/toggle/src/Toggle.tsx
lazytype: also, this will generate a new ID on every render. I think there are other function components that demonstrate how to maintain a stable ID with `IdAllocator`

lazytype (changes requested):

packages/toggle/src/Toggle.tsx
brookescarlett: 👍

packages/toggle/src/Toggle.tsx
brookescarlett: 👍

lazytype (approved):

PR #608, opened 2020-09-14
PD-858: Add darkMode prop to Checkbox

brookescarlett:
## ✍️ Proposed changes

- Refactors Checkbox component
- Adds darkMode prop to Checkbox

🎟 _Jira ticket:_ [PD-858](https://jira.mongodb.org/browse/PD-858)

## 🛠 Types of changes

## ✅ Checklist


packages/checkbox/src/Checkbox.tsx
```diff
@@ -1,8 +1,7 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { createDataProp, IdAllocator } from '@leafygreen-ui/lib';
 import { css, cx } from '@leafygreen-ui/emotion';
-import { colors } from '@leafygreen-ui/theme';
```
lazytype, line 5: curious about the motivation for this change

packages/checkbox/package.json
```diff
@@ -12,8 +12,10 @@
     "access": "public"
   },
   "dependencies": {
-    "@leafygreen-ui/lib": "^5.1.1",
-    "@leafygreen-ui/theme": "^2.0.1"
+    "@leafygreen-ui/lib": "^5.1.1"
+  },
+  "devDependencies": {
+    "@leafygreen-ui/palette": "^2.0.2"
```
DesignerDave, line 18: Why is this change being made?

brookescarlett: Isn't deprecating theme all together a goal? This is the last component that was relying on theme, so once we remove it here then we're done with the package throughout our repo. Further, the theme was only used in three places so I just replaced the reference to theme with the same hex value, with the hope that we'd move over to the palette entirely with this component eventually.

packages/checkbox/src/Checkbox.tsx
brookescarlett: see below :)

packages/checkbox/package.json
DesignerDave: Whoops, thought this was moving theme from a dependency to a devDependency. Why is this a devDependency I suppose is a better question 😅

brookescarlett: It's only being used in Storybook, for the three styles I changed, I inlined the hex values (so they're staying the same as they are)

lazytype (approved):

PR #615, opened 2020-09-16
PD-848: Tooltip notch positioning updates

DesignerDave:
## ✍️ Proposed changes

This PR addresses an issue where tooltip notches wouldn't align well with smaller triggers, like icons. It also makes the notch positioning more resilient in general, and a bit more adaptive to the size of the element.

While working with the Tooltip, I discovered an infinite loop in certain cases within Popover. This PR also addresses it, and I'm marking the change as breaking because of a DOM change to the wrapper around Popover children.

Lastly, I fixed a regression where the tooltip appearing in the global nav was now light, when it previously was dark.

🎟 _Jira ticket:_ [PD-848](https://jira.mongodb.org/browse/PD-848)

## 🛠 Types of changes

## ✅ Checklist


packages/tooltip/src/Tooltip.spec.tsx
```diff
@@ -435,7 +435,8 @@ describe('packages/tooltip', () => {
     });
 
     expect(
-      getByTestId(tooltipTestId).parentElement?.parentElement?.className,
+      getByTestId(tooltipTestId).parentElement?.parentElement?.parentElement
+        ?.className,
     ).toBe('test-classname');
```
lazytype, line 440: we can probably make this test more robust doing something like
```suggestion
    const matchedElements = document.querySelectorAll('body > .test-classname');
    expect(matchedElements).toHaveLength(1);
    const portalRoot = matchedElements.item(0)!;

    expect(portalRoot).toContainElement(getByTestId(tooltipTestId));
```

lazytype (commented):
The popover and mongo-nav changes LGTM

packages/tooltip/src/Tooltip.spec.tsx
DesignerDave: Hmm... Usually not a fan of this approach, but in this case I kinda like it. I'll give that a shot.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -542,6 +544,7 @@ export default function ProjectNav({
             align="bottom"
             justify="end"
             usePortal={false}
+            darkMode={true}
```
brookescarlett, line 547: 👍

packages/popover/src/Popover.tsx
```diff
@@ -93,24 +93,29 @@ function Popover({
...
+  // We already recalculate the position when the element becomes active,
+  // So it's safe for us to disable the mutation observers when the element isn't active.
+  const observeMutations = adjustOnMutation && active;
+
   const lastTimeRefElMutated = useMutationObserver(
     referenceElement,
     mutationOptions,
-    () => Date.now(),
-    adjustOnMutation,
+    useCallback(() => Date.now(), []),
```
brookescarlett, line 103: does this work `useCallback(Date.now, [])`

packages/popover/src/Popover.tsx
```diff
@@ -93,24 +93,29 @@ function Popover({
...
+    observeMutations,
   );
 
   const lastTimeContentElMutated = useMutationObserver(
     contentNode,
     mutationOptions,
-    () => Date.now(),
-    adjustOnMutation,
+    useCallback(() => Date.now(), []),
```
brookescarlett, line 110: ditto here

packages/tooltip/src/Tooltip.tsx
```diff
@@ -115,6 +114,28 @@ interface PopoverFunctionParameters {
 
 type ModifiedPopoverProps = Omit<PopoverProps, 'active' | 'refEl'>;
 
+type PortalProps = OneOf<
+  {
+    /**
+     * Specifies that the popover content will appear portaled to the end of the DOM,
+     * rather than in the DOM tree.
+     *
+     * default: `true`
```
brookescarlett, line 123: nit: Can we use @default just for consistency

DesignerDave: From what I see, we don't use this consistently right now, but while "@" holds semantic meaning in JSDoc, it doesn't in TSDoc. In fact, from what I can tell, it may be `@defaultValue`: https://github.com/microsoft/tsdoc/issues/27

I'd recommend that we stick with `default:`, or move entirely to `@defaultValue` rather than trying to use semantically-meaningless `@default`.

packages/popover/src/Popover.tsx
DesignerDave: Makes sense!

packages/tooltip/src/Tooltip.spec.tsx
DesignerDave: Done 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
hswolff: This is to keep the black background of the tooltip? I think I missed this naming, and I'm wondering if it makes sense.

If we have a global theme value for darkmode, when passed to tooltip I'd assume the background would be `white` so it would show for dark mode? If it was dark then how would it be visible?

I may be misunderstanding but figured I'd throw these thoughts out.

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -1,40 +1,83 @@
...
+  // When the trigger is smaller than the minimum offset we require to position the notch over the trigger,
+  // we calculate a transformation to apply to the entire tooltip so that the notch centers on that element.
+  // This is particularly important for things like icons, and icon buttons where without this transformation,
+  // the tooltip's notch could be positioned entirely off of the trigger.
+  let tooltipOffsetTransform = '';
 
   switch (align) {
     case 'top':
     case 'bottom':
+      notchOffsetUpperBound = notchOffsetLowerBound * 3;
```
hswolff, line 51: why * 3?

packages/tooltip/src/tooltipUtils.tsx
```diff
@@ -44,30 +87,51 @@ export function trianglePosition(
...
+          }
+
           break;
       }
 
       break;
 
     case 'left':
     case 'right':
+      notchOffsetUpperBound = notchOffsetLowerBound * 2;
```
hswolff, line 105: why * 2?

packages/popover/src/Popover.tsx
```diff
@@ -93,24 +93,29 @@ function Popover({
 
   const viewportSize = useViewportSize();
 
+  // We already recalculate the position when the element becomes active,
+  // So it's safe for us to disable the mutation observers when the element isn't active.
+  const observeMutations = adjustOnMutation && active;
```
hswolff, line 98: So we only want to run the callback when the Popover is active? I feel like that's at odds with what you wrote?

packages/popover/src/Popover.tsx
```diff
@@ -237,15 +238,18 @@ function Popover({
...
                 { [activeStyle]: state === 'entered' },
                 className,
               )}
             >
-              {renderedChildren}
+              {/*
+                We create this inner node with a ref because placing it on its parent
+                creates an infinite loop in some cases when dynamic styles are applied.
+               */}
+              <div ref={setContentNode}>{renderedChildren}</div>
```
hswolff, line 252: Is this going to impact styles at all? Hard to know I suppose, guess that's why you meant major. Makes sense.

packages/tooltip/src/Tooltip.tsx
hswolff: It'd be nice to at least follow one convention, whether it be JSDoc or TSDoc, rather than us invent our own.

packages/tooltip/src/tooltipUtils.tsx
DesignerDave: This number is sorta "magical". Quite frankly, this was going to be a far more magical number than this. I decided to tie it to an actual value that could create decent proportions instead.

Calculating the exact value this would need to be requires setting a ref on the Tooltip content wrapper, and getting the height / width of it. The problem was that the height / width was changing when the `open` prop was being set, causing the notch to lose its positioning in some cases before the tooltip transitioned out. This was the next best solution IMO, as in my testing it produced nearly the same result, but without the issue I mentioned.

packages/tooltip/src/tooltipUtils.tsx
DesignerDave: Ditto here

packages/popover/src/Popover.tsx
DesignerDave: How so? The idea here is to avoid unnecessary mutation observers on the page. Previously, there was no logic to prevent a mutationObserver from being set when the Popover was inactive.

packages/popover/src/Popover.tsx
DesignerDave: Yep, breaking change. I tested largely with the MongoNav, and that helped me catch a style issue or two. In most cases this shouldn't be breaking, but I was conservative with the changeset.

packages/tooltip/src/Tooltip.tsx
DesignerDave: I agree. The thing is that we've already "invented" the convention, and use it in the majority of cases (according to my search of the repo). If we're gonna change conventions, we should move to one supported by TSDoc, and do it everywhere IMO.

DesignerDave: FWIW, there wasn't a tag in TSDoc for defaults when we started using the convention, so I don't think we were wrong in using this in the past.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
DesignerDave: Yeah, this is something I missed in the darkMode conversion. I don't know that darkMode as a prop makes sense for Tooltips. I actually could see us using a dark background / light text appearance for both dark and light mode to be honest, adjusting only for the contrast. That's more of a design decision though.

This change was just to fix a regression from that PR. Unless you're suggesting not including this change, probably best to talk about the impact of the prop change in the channel.

hswolff: Def fix the regression, was more a note on the conceptual design of a dark mode Tooltip.

packages/tooltip/src/Tooltip.tsx
hswolff: Yeah for sure. Sounds like a good tech debt cleanup ticket.

packages/tooltip/src/tooltipUtils.tsx
hswolff: Wonder if we can inline any of this info into the code for later reference?

packages/popover/src/Popover.tsx
hswolff: Ah I think I got confused by "We already recalculate the position" which made it sound like we don't want to do it again. But we do? Dunno, can ignore me, just reads a little confusing to me.

hswolff (approved):

packages/popover/src/Popover.tsx
DesignerDave: When the Tooltip becomes active, we calculate the desired position initially, so we don't need the mutation observer prior to it becoming active. I can try to make this comment more clear.

packages/tooltip/src/tooltipUtils.tsx
DesignerDave: I can add a comment to make this more clear

PR #616, opened 2020-09-17
Add more parallelization to workflow

lazytype:
## ✍️ Proposed changes

This introduces the following changes:
- Run client-only and SSR tests in separate jobs so they can be parallelized. This saves ~a minute and a half.
- Use cached `node_modules` if `yarn.lock` hasn't changed. Saves ~half a minute to run `yarn install`.
- Use cached build files if no source files have changed. Saves ~a minute.
- `preactjs/compressed-size-action@v1` doesn't provide a way to conditionally build or run different build commands for base vs current, so its optimizability is limited.

Unfortunately Github actions does not have any better way to re-use steps that are common between jobs. They cannot be be sufficiently refactored out to separate actions since actions are not allowed to use `uses` 🤦‍♂️. Refactoring them out to separate jobs would result in significant slowdowns since jobs run on separate machines and can only transfer data by uploading and downloading artifacts between jobs.

🎟 _Jira ticket:_ [Reduce run time for Github action checks](https://jira.mongodb.org/browse/PD-869)

## 🛠 Types of changes

## ✅ Checklist


hswolff:
Haven't reviewed this yet but I'd be curious if we could fork or PR some changes to `preactjs/compressed-size-action@v1` as it is absolutely very limited.

hswolff (approved):
Def bummer we can't DRY it out.

lazytype:
> Haven't reviewed this yet but I'd be curious if we could fork or PR some changes to `preactjs/compressed-size-action@v1` as it is absolutely very limited.

I was thinking of forking it too.

PR #621, opened 2020-09-21
PD-874: Toggle Bug

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-874](https://jira.mongodb.org/browse/PD-874)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/toggle/src/Toggle.story.tsx
+++ b/packages/toggle/src/Toggle.story.tsx
@@ -15,7 +15,6 @@ storiesOf('Toggle', module).add('Default', () => {
         padding: 20px;
       `}
     >
-      <div>Brooke</div>
       <Toggle
         darkMode={darkMode}
         size={select('Size', Object.values(Size) as Array<Size>, Size.Default)}
--- a/packages/toggle/src/Toggle.story.tsx
+++ b/packages/toggle/src/Toggle.story.tsx
@@ -15,6 +15,7 @@ storiesOf('Toggle', module).add('Default', () => {
         padding: 20px;
       `}
     >
+      <div>Brooke</div>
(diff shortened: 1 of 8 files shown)


packages/toggle/src/Toggle.story.tsx
```diff
@@ -15,6 +15,7 @@ storiesOf('Toggle', module).add('Default', () => {
         padding: 20px;
       `}
     >
+      <div>Brooke</div>
```
DesignerDave: This probably wasn't an intentional addition.

packages/toggle/src/Toggle.tsx
```diff
@@ -23,13 +23,13 @@ const toggleInput = createDataProp('toggle-input');
...
-const focusRing = 3;
+export const focusRing = 3;
 
 const inputStyle = css`
   margin: 0;
   position: absolute;
-  left: 100%;
-  top: 100%;
+  left: 0;
+  top: 0;
```
DesignerDave, line 32: Just in case, let's add 
```css
height: 0;
width: 0;
margin: 0;
```

DesignerDave (changes requested):

packages/toggle/src/Toggle.story.tsx
brookescarlett: Yes already rm'd

packages/toggle/src/Toggle.tsx
brookescarlett: we have margin but will add height and width

DesignerDave (approved):

PR #625, opened 2020-09-23
PD-832: Inline Code Component updates

brookescarlett:
## ✍️ Proposed changes

Updates InlineCode component to match design spec

🎟 _Jira ticket:_ [PD-832](https://jira.mongodb.org/browse/PD-832)

## 🛠 Types of changes

## ✅ Checklist


packages/typography/src/Typography.tsx
```diff
@@ -102,22 +105,73 @@ function Body({ children, className, weight = 'regular', ...rest }: BodyProps) {
...
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
   const size = useBaseFontSize();
-  const body = size === 16 ? typeScale2 : typeScale1;
-
-  return (
-    <code {...rest} className={cx(sharedStyles, code, body, className)}>
-      {children}
+  const fontSize = size === 16 ? typeScale2 : typeScale1;
+  const whiteSpace =
+    typeof children === 'string' && children.length <= 30 ? nowrap : normal;
```
lazytype, line 157: I don't think this will have the expected behavior with unicode characters that have multiple code points. I think we want something like this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#Unicode

lazytype (changes requested):

packages/typography/src/Typography.tsx
```diff
@@ -112,22 +115,75 @@ function Body({ children, className, weight = 'regular', ...rest }: BodyProps) {
...
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
   const size = useBaseFontSize();
-  const body = size === 16 ? typeScale2 : typeScale1;
-
-  return (
-    <code {...rest} className={cx(sharedStyles, code, body, className)}>
-      {children}
+  const fontSize = size === 16 ? typeScale2 : typeScale1;
+  const whiteSpace =
+    (typeof children === 'string' && children.match(/./gu)?.length) ?? 0 <= 30
```
lazytype, line 167: I think you need parentheses around `... ?? 0` otherwise it'll get parsed as `... ?? (0 <= 30)`

brookescarlett: 🤦

packages/typography/src/Typography.tsx
```diff
@@ -112,22 +115,75 @@ function Body({ children, className, weight = 'regular', ...rest }: BodyProps) {
...
   font-family: ${fontFamilies.code};
-  display: inline-block;
+
+  ${anchorDataProp.selector}:hover > code > & {
+    border-color: ${uiColors.blue.base};
+  }
+`;
+
+const codeFocus = css`
+  ${anchorDataProp.selector}:focus > code > & {
```
lazytype, line 132: curious why these styles apply to the children rather than the whole element

brookescarlett: Because the border is implemented on the inner span so that it remains the same whether or not the code is wrapped in an anchor tag

lazytype (approved):

PR #628, opened 2020-09-23
PD-876: Adds initialCollapsed prop to SideNavGroup

brookescarlett:
## ✍️ Proposed changes

Adds defaultCollapsable prop to SideNavGroup

🎟 _Jira ticket:_ [PD-876](https://jira.mongodb.org/browse/PD-876)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/side-nav/src/SideNav.spec.tsx
+++ b/packages/side-nav/src/SideNav.spec.tsx
@@ -179,11 +179,11 @@ describe('packages/side-nav', () => {
     const renderGroup = ({
       header,
       collapsible = false,
-      defaultCollapsed = true,
+      initialCollapsed = true,
     }: {
       header?: header;
       collapsible?: boolean;
-      defaultCollapsed?: boolean;
+      initialCollapsed?: boolean;
     } = {}) => {
       const { sideNavGroup, sideNavLink } = testIds;
       render(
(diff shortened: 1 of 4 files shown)


packages/side-nav/src/SideNav.spec.tsx
```diff
@@ -283,6 +285,17 @@ describe('packages/side-nav', () => {
         expect(childContent).toBeInTheDocument();
       });
     });
+
+    describe('when `collapsible` is true and `defaultCollapsed` is false', () => {
```
lazytype, line 289: would we want to make `collapsible={false} defaultCollapsed={true}` be a type error?

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -144,15 +151,17 @@ interface SideNavGroupProps {
...
  */
 function SideNavGroup({
   header,
   children,
   collapsible = false,
+  defaultCollapsed = true,
   ...rest
 }: SideNavGroupProps) {
-  const [open, setOpen] = React.useState(false);
+  const [open, setOpen] = React.useState(!defaultCollapsed);
```
lazytype, line 164: I lean towards naming this `initial`, rather than `default` to more clearly indicate that its value won't be used again after the initial render and will have no effect if changed.

brookescarlett: Yes I like that idea, good call

packages/side-nav/src/SideNav.spec.tsx
brookescarlett: Also good call

lazytype (approved):

PR #633, opened 2020-09-25
PD-629: Code line-highlighting

DesignerDave:
## ✍️ Proposed changes

This PR adds line highlighting, as well as simplifies rendering and makes the syntax and code components render more consistently.

Screen Shot 2020-09-25 at 1 51 25 PM

Screen Shot 2020-09-25 at 1 51 05 PM

**Primary changes:**
- Adds the line highlighting feature through the `highlightLines` prop. This prop accepts an array composed of numbers, and `[number, number]` tuples. Support for those tuples was added to make it simpler for the docs team to integrate the feature, and to allow simpler implementation of highlighted ranges.
- All code examples are now rendered as a table, making unhighlighted code render consistently with highlighted code. Previously, there was inconsistency in how unhighlighed code would render in the DOM since it wouldn't be processed through our highlight.js plugin.
- We now only render one set of highlighted code with React. Previously we were rendering two, but this was simplified by introducing a local context provider that's used to consume whether line numbers should be shown, if dark mode is enabled in the component, and what line numbers to highlight.
- Fixes an issue where users might be unable to manually copy code within the code component. This appears to have been an accidental change when switching to rendering syntax as a table.
- Deprecates the `multiline` prop in the Code component. Single line code examples are now detected automatically via the presence of line breaks in the string, and single line spacing is applied automatically. This change was made necessary when switching to rendering everything as a table, since the multiline prop relied on not rendering the code within a `pre` tag, and not breaking new line characters into separate table rows. The new implementation now adjusts spacing automatically which should be a better DX overall, though this change is breaking.
- Fixes a few design discrepancies with borders in dark mode.
- Addresses an issue where the Code and Syntax components would sometimes break when highlighting syntax that was invalid for the selected language.

🎟 _Jira ticket:_ [PD-629](https://jira.mongodb.org/browse/PD-629)

## 🛠 Types of changes

## ✅ Checklist


.changeset/nasty-seas-marry.md
```diff
@@ -0,0 +1,9 @@
+---
+'@leafygreen-ui/code': major
+'@leafygreen-ui/syntax': major
+---
+
+- Adds the line highlighting feature! This feature is supported through the `highlightLines` prop.
```
brookescarlett, line 6: nit: prefer to keep these separate and document the changes specific to each package in their own changeset

DesignerDave: I can document the multiline change separately, but all the others apply to both components.

packages/code/src/Code.tsx
```diff
@@ -382,55 +352,36 @@ function Code({
...
+      </div>
+    </div>
   );
 }
 
 Code.displayName = 'Code';
 
 Code.propTypes = {
   children: PropTypes.string.isRequired,
-  multiline: PropTypes.bool,
```
brookescarlett, line 433: Replace with highlightLines

packages/code/src/Code.tsx
```diff
@@ -382,55 +352,36 @@ function Code({
...
           className={wrapperClassName}
           onScroll={onScroll}
-          ref={scrollableSingleLine}
+          ref={scrollableElement}
         >
           {renderedSyntaxComponent}
-        </div>
+        </pre>
 
         {copyBar}
```
brookescarlett, line 375: Seeing accessibility issues from the Icon in the copy button

brookescarlett: Definitely possible that they were present before, but if possible we should address while updating the component

packages/code/src/Code.tsx
```diff
@@ -197,49 +195,18 @@ interface CodeProps extends Omit<SyntaxProps, 'onCopy'> {
    *
    */
   onCopy?: Function;
+
+  /**
+   * An array of the line numbers to highlight
+   */
+  highlightLines?: Array<number>;
```
lazytype, line 202: should this also include ranges too?

packages/code/src/Code.tsx
```diff
@@ -178,13 +183,6 @@ interface CodeProps extends Omit<SyntaxProps, 'onCopy'> {
...
 
-  /**
-   * When true, whitespace and line breaks will be preserved.
-   *
-   * default: `true`
-   * */
-  multiline?: boolean;
-
   /**
    * When true, allows the code block to be copied to the user's clipboard
```
brookescarlett, line 187: This description could be better: `If present, displays copy bar next to Code component. The function pased here is invoked when copy button is clicked`. Additionally, noticed that it's not possible to both `showWindowChrome` and `onCopy`. We should probably indicate this through the typing as well

packages/syntax/src/Syntax.tsx
```diff
@@ -1,14 +1,48 @@
...
+  const expandedRange = [];
+
+  for (let i = lower; i <= upper; i++) {
+    expandedRange.push(i);
+  }
+
+  return expandedRange;
+}
+
+export function parseLineHighlightNumbers(
```
lazytype, line 33: when I see "parse" I think of extracting something from a string. can this be called `expandRanges` or something like that?

.changeset/nasty-seas-marry.md
brookescarlett: Wfm

packages/syntax/src/Syntax.tsx
```diff
@@ -1,14 +1,48 @@
...
-import { Language, SyntaxProps } from './types';
+import { Language, SyntaxProps, Mode } from './types';
 import { SupportedLanguages, languageParsers } from './languages';
 import { injectGlobalStyles } from './globalStyles';
-import renderingPlugin from './renderingPlugin';
+import renderingPlugin, { TableContent } from './renderingPlugin';
+import { SyntaxContext } from './SyntaxContext';
+
+export function expandRangeTuple(
+  tuple: [number, number],
```
lazytype, line 16: can this also take in min/max arguments? the max we pass in should just be however many lines of code there are. don't want the window to freeze up when someone passes in `[0, Infinity]`.

packages/syntax/src/Syntax.tsx
```diff
@@ -1,14 +1,48 @@
...
+): number | Array<number> {
+  const [lower, upper] = tuple.sort();
```
lazytype, line 18: mutating the argument seems like an unintended side-effect
```suggestion
  const [lower, upper] = [...tuple].sort();
```

packages/syntax/src/Syntax.tsx
```diff
@@ -1,14 +1,48 @@
...
+
+  if (lower === upper) {
+    return lower;
+  }
```
lazytype, line 22: it would simplify the function to remove this block and just return `Array<number>` even for the single-element case.

packages/syntax/src/renderingPlugin.tsx
```diff
@@ -57,32 +58,94 @@ export function processToken(token: TreeItem, key?: number): React.ReactNode {
...
 
   return (
-    <tr>
-      <td
-        {...numberCellDataProp.prop}
-        className={cx(cellStyle, numberCellStyle)}
-      >
-        {lineNumber}
-      </td>
+    <tr className={cx({ [getHighlightedRowStyle(darkMode)]: highlighted })}>
```
lazytype, line 133: would we benefit from memoizing `getHighlightedRowStyle`?

packages/syntax/src/renderingPlugin.tsx
```diff
@@ -132,61 +195,71 @@ export function treeToLines(
...
-function TableRowLine({ lineNumber, line }: TableRowLineProps) {
-  return (
-    <LineTableRow lineNumber={lineNumber}>
-      {line.map(processToken)}
-    </LineTableRow>
-  );
-}
+export function TableContent({ lines }: TableContentProps) {
+  const { highlightLines, showLineNumbers, darkMode } = useSyntaxContext();
+  const trimmedLines = lines;
```
lazytype, line 207:
```suggestion
  const trimmedLines = [...lines];
```

packages/syntax/src/renderingPlugin.tsx
```diff
@@ -132,61 +195,71 @@ export function treeToLines(
...
-function TableRowFragment({ line }: TableRowFragmentProps) {
   return (
     <>
-      <span>{line.map(processToken)}</span>
-
-      {/* We use a new line character here instead of <br /> so that text will not break outside of a <pre /> tag */}
-      {'\n'}
+      {trimmedLines.map((line, index) => {
+        const currentLineNumber = index + 1;
+        const highlightLine = highlightLines.includes(currentLineNumber);
```
lazytype, line 223: do we polyfill `Array.prototype.includes`? IIRC it's not supported by IE

packages/syntax/src/renderingPlugin.tsx
```diff
@@ -132,61 +195,71 @@ export function treeToLines(
...
+          <div
+            className={css`
+              display: inline-block;
+            `}
+          />
+        );
+
+        return (
+          <LineTableRow
+            key={index}
```
lazytype, line 245: can we use `currentLineNumber` instead? no difference now, but would be more logical in a scenario where we could render a subset of all lines

lazytype (changes requested):

packages/code/src/Code.tsx
DesignerDave: Not a regression, and is an issue with the IconButton component itself. Made a ticket: https://jira.mongodb.org/browse/PD-458

packages/code/src/Code.tsx
DesignerDave: Not trying to overhaul too many existing issues in this component, but I'll update the description.

`showWindowChrome` and `onCopy` are compatible, even if the window chrome and onCopy won't be rendered at the same time. In some edge cases, this will make implementation easier, so I'm not sure I'd want to prevent this.

packages/code/src/Code.tsx
DesignerDave: Good call

packages/syntax/src/Syntax.tsx
DesignerDave: Fair enough, though `expandRanges` doesn't describe it entirely accurately either. It's quite verbose, but I think `normalizeLineHighlightingDefinition` is a bit better.

packages/syntax/src/Syntax.tsx
DesignerDave: Passing `Infinity` seems like an almost impossible edge case, but I'll do something to make sure the tab doesn't freeze in the off chance that happens. Not in love with the min/max args.

packages/syntax/src/Syntax.tsx
DesignerDave: Good call.

packages/syntax/src/renderingPlugin.tsx
DesignerDave: Don't think so – emotion does a lot of caching for these values, and I believe there's a chance that the emotion className might change without a dependency of `useMemo` changing.

packages/syntax/src/renderingPlugin.tsx
DesignerDave: Good call

packages/syntax/src/renderingPlugin.tsx
DesignerDave: I'm not sure, but we use it in more than a few components. If we need to address this, we probably need to do it everywhere.

packages/syntax/src/renderingPlugin.tsx
DesignerDave: Given that `currentLineNumber` is just `index + 1`, it's going to suffer from the same issue as using index in this current implementation. I'm fine with the change though.

packages/syntax/src/renderingPlugin.tsx
lazytype: I'd be happy if we could formally drop support for IE! 🤞

packages/syntax/src/Syntax.tsx
DesignerDave: Fair point!

packages/syntax/src/renderingPlugin.tsx
DesignerDave: Me too...

In our babelrc on master, it looks like we're probably polyfilling this:
```js
const presets = [
    '@babel/preset-typescript',
    '@babel/preset-react',
    [
      '@babel/preset-env',
      {
        targets: {
          browsers: ['last 2 versions', 'safari >= 7', 'ie >= 10'],
        },
        modules: 'commonjs',
      },
    ],
  ];
```

DesignerDave: Correction: From looking at the built bundle itself, I can't find any indication that it's polyfilled.

lazytype: Also, updating the browsers list doesn't seem to result in any change to the bundle. it's pretty unclear what babel uses it for. maybe it only considers whether a browser supports specific syntaxes and not APIs.

packages/code/src/Code.tsx
```diff
@@ -382,61 +352,48 @@ function Code({
...
-  multiline: PropTypes.bool,
   language: PropTypes.oneOf(Object.values(Language)),
   darkMode: PropTypes.bool,
   className: PropTypes.string,
   showLineNumbers: PropTypes.bool,
   showWindowChrome: PropTypes.bool,
   chromeTitle: PropTypes.string,
+  highlightLines: PropTypes.arrayOf(
+    PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.number),
```
lazytype, line 393: I think a way to only accept a pair of numbers is 
```tsx
PropTypes.shape({ 0: PropTypes.number, 1: PropTypes.number })
```

packages/syntax/src/Syntax.story.tsx
```diff
@@ -28,6 +28,15 @@ storiesOf('Syntax', module).add(
         showLineNumbers={boolean('Show line numbers', false)}
         darkMode={boolean('darkMode', false)}
         language={select('language', Language, Language.JavaScript)}
+        highlightLines={select(
+          'highlight lines',
+          {
+            none: undefined,
+            single: [1],
+            multiple: [2, 3, 5],
```
lazytype, line 36: can we have a range example here too

packages/syntax/src/Syntax.tsx
```diff
@@ -1,14 +1,56 @@
...
+  const [lower, upper] = [...tuple]
+    .map(bound => {
+      // Make sure passing infinity doesn't freeze the browser
+      if (bound === Infinity) {
+        // 2000 is arbitrary, but far larger than existing use-cases for our syntax highlighting
+        return 2000;
+      }
+
+      if (bound === -Infinity) {
+        return -2000;
```
lazytype, line 25: if this is only used for source code lines, then it'd make sense to clamp the lower-bound to 0, i.e. `bound = Math.max(bound, 0)`

packages/syntax/src/Syntax.tsx
```diff
@@ -1,14 +1,56 @@
...
 import { injectGlobalStyles } from './globalStyles';
-import renderingPlugin from './renderingPlugin';
+import renderingPlugin, { TableContent } from './renderingPlugin';
+import { SyntaxContext } from './SyntaxContext';
+
+export function expandRangeTuple(tuple: [number, number]): Array<number> {
+  const [lower, upper] = [...tuple]
+    .map(bound => {
+      // Make sure passing infinity doesn't freeze the browser
+      if (bound === Infinity) {
```
lazytype, line 19: I think this is overly-specific and we should clamp any number that's too large, e.g. i.e. `bound = Math.min(bound, 2000)`

packages/syntax/src/renderingPlugin.tsx
```diff
@@ -257,17 +257,10 @@ export function TableContent({ lines }: TableContentProps) {
...
-
-    if (result.illegal) {
-      // If highlight.js identifies invalid syntax, we render the string like we would with language set to "none"
-      lines = result.code.split('\n').map(str => [str]);
-    } else {
-      const { rootNode } = result.emitter;
-      lines = treeToLines(rootNode.children);
-    }
+    const { rootNode } = result.emitter;
+    console.log(result)
```
lazytype, line 261: rm

packages/syntax/src/Syntax.tsx
```diff
@@ -118,7 +118,7 @@ function Syntax({
       return null;
     }
 
-    return hljs.highlight(language, children);
+    return hljs.highlight(language, children, true);
```
lazytype, line 121:
```suggestion
    return hljs.highlight(language, children, /* ignore_illegals */ true);
```

packages/syntax/src/Syntax.story.tsx
DesignerDave: Doing so works, but causes a TS error, presumably because of the typing of the knobs ☹️

packages/syntax/src/Syntax.tsx
DesignerDave: I was avoiding this simply because I'm not a fan of arbitrary limitations. Highlighting 2000 lines should work, even if the use-case isn't really there. Is there a reason you see being so specific would cause issues?

packages/syntax/src/renderingPlugin.tsx
DesignerDave: Oops

packages/syntax/src/Syntax.tsx
DesignerDave: I was avoiding bringing in a new dependency to this package, though since it's `lodash`, it probably wouldn't be a huge deal. With how the component is built, there's no added resilience to a clamp here – it does nothing to include a number that doesn't has a corresponding line.

packages/syntax/src/Syntax.tsx
DesignerDave: As-is, the Infinity case is exceptionally unlikely.

packages/syntax/src/Syntax.tsx
lazytype: no dependency needed; `Math` is a built-in

DesignerDave: 🤦 Duh. Suppose there's no reason not to do this in that case.

packages/code/src/Code.tsx
DesignerDave: Ahh, smart.

packages/syntax/src/Syntax.story.tsx
lazytype: you could pull the values out of the knob, e.g.
```ts
const examples = {
  none: undefined,
  single: [1],
  multiple: [2, 3, 5],
};

highlightLines={
  examples[
    select('highlight lines', Object.keys(examples), 'none')
  ]
}
```

packages/syntax/src/Syntax.tsx
lazytype: it just makes the interface of the function more complex than necessary. it's simpler to say
"this function returns an array of numbers within the range provided, bounded by 0 and 2000" than it is to say "this function returns an array of numbers within the range provided, unless a bound is Infinity or -Infinity in which it will be replaced by 2000 and -2000 respectively"

packages/syntax/src/Syntax.story.tsx
DesignerDave: Smart

packages/syntax/src/Syntax.tsx
DesignerDave: I don't feel that strongly about this. I'll change it.

DesignerDave: Actually I just realized, clamping the bounds to 2000 means that even if the tuple is `[1999, 2001]`, it would output `[1999, 2000]`. If we ever add a lines-start sort of property where they could start line numbering at `2001`, this would effectively disable line highlighting. I feel a bit more strongly about this now given that. We could make sure the upper and lower bound difference is than 2000, but that's less elegant of a solution for the `Infinity` problem.

lazytype (approved):

PR #636, opened 2020-09-29
PD-783: Makes Nav responsive

brookescarlett:
## ✍️ Proposed changes

- Makes Nav responses
- Uses styled API for GridContainer and migrates GridItem off of imports from @emotion/core

🎟 _Jira ticket:_ [PD-783](https://jira.mongodb.org/browse/PD-783)

## 🛠 Types of changes

Screen Shot 2020-09-29 at 5 26 45 PM


website/components/grid/Grid.tsx
```diff
@@ -59,23 +52,22 @@ type GridContainerProps = JSX.IntrinsicElements['div'] & {
...
+  })}
+
+  display: flex;
+  ${['flex-direction']}: ${props =>
+    props.direction ? props.direction : Direction.Row};
+  ${['flex-wrap']}: ${props => (props.wrap ? props.wrap : Wrap.Wrap)};
+  ${['align-items']}: ${props => (props.align ? props.align : Align.Center)};
+  ${['justify-content']}: ${props =>
+    props.justify ? props.justify : Justify.Center};
+`;
```
lazytype, line 68: not a big fan of this api. it looks more like its own language than javascript and is significantly harder to read than the previous code.

website/components/navigation/MobileNavigationGroup.tsx
```diff
@@ -0,0 +1,134 @@
...
+  children,
+  ...rest
+}: MobileNavigationGroupProps) {
+  const [open, setOpen] = useState(false);
+  const nodeRef = useRef(null);
+  const ulRef = useRef<HTMLUListElement>(null);
+
+  return (
+    <li
+      className={css`
```
lazytype, line 83: we probably want to pull out `className` from `rest` and merge it in here

website/components/navigation/MobileNavigationGroup.tsx
```diff
@@ -0,0 +1,134 @@
...
+          <div
+            ref={nodeRef}
+            className={cx(defaultStyle, {
+              [transitionStyles.entering]: state === 'entering',
+              [css`
+                opacity: 1;
+                max-height: ${ulRef?.current?.getBoundingClientRect().height}px;
+              `]: state === 'entered',
+              [transitionStyles.exiting]: state === 'exiting',
+              [transitionStyles.exited]: state === 'exited',
```
lazytype, line 121: this could be simplified to `cx(defaultStyle, transitionStyles[state], ...)`

website/components/navigation/Navigation.tsx
```diff
@@ -19,57 +23,73 @@ const logoStyles = css`
 `;
 
 function Navigation() {
+  const viewport = useViewportSize();
+  const isMobile = viewport && viewport.width < 768;
```
lazytype, line 27: `!!viewport` to keep `isMobile` to a boolean

website/components/navigation/Navigation.tsx
```diff
@@ -19,57 +23,73 @@ const logoStyles = css`
...
+  const Group = isMobile ? MobileNavigationGroup : SideNavGroup;
+  const Item = isMobile ? MobileNavigationItem : SideNavItem;
+
+  const groupProps =
```
lazytype, line 31: prefer using ternary operator here so that `groupProps` can't be `false`

website/components/navigation/Navigation.tsx
```diff
@@ -19,57 +23,73 @@ const logoStyles = css`
...
+        <Item>Toast</Item>
+        <Item>Toggle</Item>
+        <Item>Tokens</Item>
+        <Item>Tooltip</Item>
+        <Item>Typography</Item>
+      </Group>
+    </>
+  );
+
+  if (isMobile) {
```
lazytype, line 85: this definition has a lot of branching based on `isMobile` which makes it harder to reason about. I would explore what it would look like if there was only one branching point.

website/components/navigation/styles.ts
```diff
@@ -0,0 +1,18 @@
...
+  padding-left: ${spacing[4]}px;
+  padding-right: ${spacing[4]}px;
+`;
+
+export const borderColor = uiColors.gray.light2;
+
+export const ulStyleOverrides = css`
+  margin-block-start: 0px;
+  margin-block-end: 0px;
+  padding-inline-start: 0px;
```
lazytype, line 15: saw various places we could replace `0px` with `0` in this PR

lazytype (changes requested):

website/components/grid/Grid.tsx
brookescarlett: Updated because @DesignerDave loves styled components. Let me know what you think in this case!

website/components/navigation/MobileNavigationGroup.tsx
brookescarlett: I can't only because in the style for entered I need access to `ulRef`, although I suppose I can make entered a function and just pass the value 🤔

lazytype: that definition can stay as it is, e.g.
```ts
cx(defaultStyle, transitionStyles[state], {
  [css`
    opacity: 1;
    max-height: ${ulRef?.current?.getBoundingClientRect().height}px;
  `]: state === 'entered'
}
```

website/components/navigation/MobileNavigationGroup.tsx
brookescarlett: 👍

website/components/navigation/Navigation.tsx
brookescarlett: 👍

website/components/navigation/Navigation.tsx
brookescarlett: 👍

website/components/navigation/Navigation.tsx
brookescarlett: The main thing I wanted to do here was share the the `content` between the Mobile and desktop Navigation components.

brookescarlett: Is it easier to make sense of if I move all of the conditional logic to a component, so it only reads as such:

```
function Navigation() {
  const viewport = useViewportSize();
  const isMobile = !!viewport && viewport.width < 768;

  if (isMobile) {
    return (
      <MobileNavigation>
        <Content isMobile />
      </MobileNavigation>
    );
  }

  return (
    <nav className={navWidth}>
      <MDBDesignLogo className={logoStyles} />
      <SideNav>
        <Content />
      </SideNav>
    </nav>
  );
}
```

website/components/navigation/styles.ts
brookescarlett: Fixed

website/components/grid/Grid.tsx
lazytype: I'd have to hear a real rationale behind the preference. the readability of this code does not make a very convincing case.

website/components/navigation/MobileNavigationGroup.tsx
brookescarlett: 👍

website/components/navigation/Navigation.tsx
lazytype: yeah, that could be an improvement

lazytype: oh I see you already made the change. 👍 LGTM

brookescarlett: 👍

brookescarlett:
@DesignerDave: updated so that you can collapse the nav (similar to a modal, collapses when you click the background per convo with @audroue). Also added routing to this PR. Made a separate ticket to better address the header (because it needs to be more responsive: https://jira.mongodb.org/browse/PD-915).

DesignerDave (approved):
I think the animation needs a bit of work, but no need to block this for the time being!

lazytype (approved):

PR #637, opened 2020-09-30
PD-823: Makes Table responsive

brookescarlett:
## ✍️ Proposed changes

Makes Table responsive by adding box shadow to table when the table does not fit in the viewport.

🎟 _Jira ticket:_ [PD-823](https://jira.mongodb.org/browse/PD-823)

## 🛠 Types of changes

## ✅ Checklist


packages/table/src/Table.tsx
```diff
@@ -14,6 +19,42 @@ const tableStyles = css`
...
+  None: 'none',
+  Left: 'left',
+  Right: 'right',
+  Both: 'both',
+} as const;
+
+type ScrollState = typeof ScrollState[keyof typeof ScrollState];
+
+function getScrollShadowStyle(scrollState: ScrollState): string {
+  const shadowColor = 'rgba(93,108,116,0.3)';
```
lazytype, line 32: I think this would be more legible as `transparentize(.7, uiColors.gray.dark1)`

packages/table/src/Table.tsx
```diff
@@ -14,6 +19,42 @@ const tableStyles = css`
...
+
+  if (scrollState === ScrollState.Both) {
```
lazytype, line 34: can we use a `switch` and explicitly handle `ScrollState.None`? if another scroll state were added in the future we'd want this function to have a type error

packages/table/src/Table.tsx
```diff
@@ -35,19 +76,77 @@ export default function Table<Shape>({
   className,
   ...rest
 }: TableProps<Shape>) {
+  const [scrollState, setScrollState] = React.useState<ScrollState>(
+    ScrollState.None,
+  );
+  const divRef = React.useRef<HTMLDivElement>(null);
+  const viewportSize = useViewportSize();
+  console.log(viewportSize);
```
lazytype, line 84: rm

packages/table/src/Table.tsx
```diff
@@ -35,19 +76,77 @@ export default function Table<Shape>({
...
+
+  useIsomorphicLayoutEffect(() => {
+    const ref = divRef.current;
```
lazytype, line 87: can clear up the following blocks by early returning if `ref == null`

packages/table/src/Table.tsx
```diff
@@ -35,19 +76,77 @@ export default function Table<Shape>({
...
+
+    if (ref != null && ref.scrollWidth > ref.clientWidth) {
+      setScrollState(ScrollState.Right);
+    } else if (
+      ref != null &&
+      viewportSize != null &&
+      ref.getBoundingClientRect().width < viewportSize.width
```
lazytype, line 94: would we want `<=` to be more precise?

packages/table/src/Table.tsx
```diff
@@ -35,19 +76,77 @@ export default function Table<Shape>({
...
+    ) {
+      setScrollState(ScrollState.None);
+    }
+  }, [viewportSize]);
+
+  const handleScroll = (e: React.UIEvent) => {
```
lazytype, line 100: I don't think we need the type assertions below if this is changed to`React.UIEvent<HTMLElement>`

lazytype (changes requested):

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: Even with this React.UIEvent<HTMLElement> I get errors when I remove the assertions below

packages/table/src/Table.tsx
brookescarlett: Yes makes sense, but for now going to write

```ScrollState.None: 
default:
```

because the default should be handled the same way

packages/table/src/Table.tsx
brookescarlett: 👍 but .7 looked too light so made it .6

packages/table/src/Table.tsx
lazytype: > because the default should be handled the same way

There is no default case because the only cases possible are those enumerated by `ScrollState`.

If we omit `default:` Typescript won't complain now because all possible cases of `ScrollState` have been handled, but we want it to complain when adding new cases to the enum.

brookescarlett: Removed!

packages/table/src/Table.tsx
lazytype: It seems to be because the `T` in `React.UIEvent<T>` is the element the handler is attached to, but for events in general the actual target could be a child of that element. Since this is a scroll event, the target probably can't ever be a child, but the typings don't reflect that. One way to work around it is to type the event as `React.UIEvent & {target: HTMLElement}`, though ideally the React typings would itself define something like `type ScrollUIEvent<T extends Element> = React.UIEvent<T> & {target: T};`

lazytype (approved):

packages/table/src/Table.tsx
```diff
@@ -35,19 +77,79 @@ export default function Table<Shape>({
...
+        setScrollState(ScrollState.Left);
+      } else if (scrollPosition < maxPosition) {
+        setScrollState(ScrollState.Right);
+      }
+    }
+  };
+
+  const debounceScroll = debounce(handleScroll, 50, { leading: true });
+
+  const onScroll: React.UIEventHandler<HTMLDivElement | HTMLPreElement> = e => {
```
lazytype, line 123: kinda gross, but i think we need something like
```suggestion
  const onScroll: React.EventHandler<React.UIEvent & { target: HTMLDivElement | HTMLPreElement }> = e => {
```

packages/table/src/Table.tsx
brookescarlett:
```Type 'UIEvent<HTMLDivElement, UIEvent>' is not assignable to type '{ target: HTMLDivElement; }'.
        Types of property 'target' are incompatible.```

File: packages/table/src/Table.tsx (hunk 7 above)
Comment (lazytype) [reply packages/table/src/Table.tsx]: mm, it was a valiant effort but looks like React needs to fix the typing of `onScroll` for our types to work. type assertions it is then 😔

Comment (brookescarlett) [reply packages/table/src/Table.tsx]: Reverted back 😞

File: packages/table/src/Table.tsx [hunk 8]
```diff
@@ -14,6 +20,42 @@ const tableStyles = css`
...
+  None: 'none',
+  Left: 'left',
+  Right: 'right',
+  Both: 'both',
+} as const;
+
+type ScrollState = typeof ScrollState[keyof typeof ScrollState];
+
+function getScrollShadowStyle(scrollState: ScrollState): string {
+  const shadowColor = transparentize(0.6, uiColors.gray.dark1);
```
DesignerDave, line 33: This should be transparent black, no?

packages/table/src/Table.tsx
```diff
@@ -35,19 +77,82 @@ export default function Table<Shape>({
...
-    <table
-      cellSpacing="0"
-      cellPadding="0"
-      className={cx(tableStyles, className)}
-      {...rest}
+    <div
+      ref={divRef}
+      onScroll={onScroll}
+      className={cx(
+        css`
```
DesignerDave, line 136: We need to transition the box-shadow here.

DesignerDave (changes requested):

packages/table/src/Table.tsx
brookescarlett: I style QA'd this with Rob, but happy to update if you'd like

DesignerDave: Easy to miss this in a design QA. In general, we've been moving all of our shadows to RGBA values of our black.

DesignerDave: To be fair, I think the Table Sketch library needs this update as well. @audroue Could you wrap that into the migration to Figma?

packages/table/src/Table.tsx
```diff
@@ -1,19 +1,84 @@
...
+  width: 20px;
+  overflow: hidden;
+
+  &:after {
+    opacity: 0;
+    content: '';
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    width: 6px;
```
DesignerDave, line 43: Right now, the shadow doesn't form a hard edge on the outer side of the shadow. My suggestion:

- Bump this width value to at least `12px`.
- Change the x offset of the shadows to `4px` and `-4px`.
- Reduce Alpha value of the box shadow RGBA from `0.4` to `0.3`.

Doing these things will make this look much more polished.

packages/table/src/Table.tsx
```diff
@@ -1,19 +1,84 @@
...
+  max-width: calc(100% - 100px);
+`;
+
 const tableStyles = css`
   border-collapse: collapse;
   box-sizing: border-box;
   border-bottom: 1px solid ${uiColors.gray.light2};
 `;
 
+const shadow = css`
```
DesignerDave, line 30: We need a `pointer-events: none` rule in these styles to ensure that the shadows don't block user interaction with elements underneath the shadow.

DesignerDave (changes requested):

packages/table/src/Table.tsx
brookescarlett: 👍

packages/table/src/Table.tsx
brookescarlett: 👍

DesignerDave (approved):

PR #639, opened 2020-10-02
Starts Header/Tooling for .design

brookescarlett:
- Standardizes our readmes further (which becomes important in the `<CodeDocs />` component)
- Defines routing for our components and the template that component pages will follow
- Adds utils folder


packages/icon/src/Icon.spec.tsx
```diff
@@ -52,6 +52,7 @@ describe('packages/Icon/glyphs/', () => {
         readonly [K in string]: string | SVGNodeObject;
       };
 
+      //@ts-expect-error
```
brookescarlett, line 55: Need to look into this more, but wanted to pass our tests :)

website/pages/component/[id]/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      readme,
+      component: id,
+    },
+  };
+};
+
+export const getStaticPaths: GetStaticPaths = async () => {
+  const packages = fs.readdirSync('../packages');
+
+  const paths = packages.map(packageName => ({
```
Comment (hswolff) [inline website/pages/component/[id]/index.tsx:49]: nit, it's the folder name, not the package name. that's defined inside the package.json

```suggestion
  const paths = packages.map(folderName => ({
```

website/pages/component/[id]/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  let changelogMarkdown, readme;
+
+  if (typeof id === 'string') {
+    changelogMarkdown = fs.readFileSync(
+      path.join('../packages', id, 'changelog.md'),
+    );
+
+    readme = fs
+      .readFileSync(path.join('../packages', id, 'README.md'))
```
Comment (hswolff) [inline website/pages/component/[id]/index.tsx:31]: Let's do this async instead so we don't lock the thread.

website/pages/component/[id]/index.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { GetStaticPaths, GetStaticProps } from 'next';
+import fs from 'fs';
+import path from 'path';
+import Layout from '../../../components/layout';
```
Comment (hswolff) [inline website/pages/component/[id]/index.tsx:5]: Do we intend this component to be re-used elsewhere?

If no, then really we should just inline it in here.

website/utils/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as markdownToHtml } from './markdownToHtml';
```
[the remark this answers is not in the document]
brookescarlett: 👍

website/pages/component/[id]/index.tsx
Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx:49]: 👍

website/pages/component/[id]/index.tsx
Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx:5]: 👍

website/pages/component/[id]/index.tsx
Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx:31]: If we do this async then the prop shows up in the Component component as undefined

Comment (hswolff) [reply website/pages/component/[id]/index.tsx:63]: We have an async await below, that works fine no? May be some bug when switching to the async version.

Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx:63]: That stops working when the original request isn't completed before the value is needed on line 74. I can't await the readFile function, so it's trying to convert the changelog from markdown to HTML before there's a value for the changelogMarkdown.

packages/modal/src/Modal.spec.tsx
```diff
@@ -58,9 +58,10 @@ describe('packages/modal', () => {
       const modal = getByRole('dialog');
       fireEvent.click(modal.parentElement!);
 
-      await expect(waitForElementToBeRemoved(modal)).rejects.toEqual(
+      await expect(waitForElementToBeRemoved(modal)).rejects.toBe(
         Error('Timed out in waitForElementToBeRemoved.'),
-      );
+      ); ///
```
brookescarlett, line 63: The error is slightly different now.

I'm seeing this: 
```
Expected value:  [Error: Timed out in waitForElementToBeRemoved.]
Received object: [Error: Timed out in waitForElementToBeRemoved.·
```

brookescarlett: cc: @lazytype

lazytype: my guess is node or jest dependency changed when the yarn.lock changed

website/pages/component/[id]/index.tsx
Comment (hswolff) [reply website/pages/component/[id]/index.tsx]: They do provide a promise API now: https://nodejs.org/api/fs.html#fs_promise_example

Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx]: used a utils wrapper but I can update to this if you think that's better

Comment (hswolff) [reply website/pages/component/[id]/index.tsx]: Def should be async please!

Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx]: Yes made it async! By using util to turn fs response into a promise
`const getFileContent = util.promisify(fs.readFile);`

website/pages/component/[id]/index.tsx
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import { GetStaticPaths, GetStaticProps } from 'next';
+import fs from 'fs';
+import path from 'path';
+import util from 'util';
+import { css } from 'emotion';
+import { GridContainer, GridItem } from '../../../components/grid/Grid';
```
Comment (hswolff) [inline website/pages/component/[id]/index.tsx:7]: Maybe not for this PR, but may want to look into https://nextjs.org/docs/advanced-features/module-path-aliases to make these imports less painful.

website/pages/component/[id]/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+const topMargin = css`
+  margin-top: 36px;
+`;
+
+export default function Component({
+  component,
+  changelog,
+  readme,
+}: BaseLayoutProps) {
+  console.log(component, changelog, readme);
```
Comment (hswolff) [inline website/pages/component/[id]/index.tsx:29]: 👀

website/pages/component/[id]/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    'utf-8',
+  );
+
+  return {
+    props,
+  };
+};
+
+export const getStaticPaths: GetStaticPaths = async () => {
+  const packages = fs.readdirSync('../packages');
```
Comment (hswolff) [inline website/pages/component/[id]/index.tsx:77]: We should seek to do all things async as we can to make sure we don't block.

website/pages/component/[id]/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const getStaticProps: GetStaticProps = async ({ params }) => {
+  if (!params || typeof params.id !== 'string') {
+    return { props: {} };
+  }
+
+  const { id } = params;
+
+  const props: Partial<BaseLayoutProps> = { component: id };
+
+  const getFileContent = util.promisify(fs.readFile);
```
Comment (hswolff) [inline website/pages/component/[id]/index.tsx:58]: We should promisify outside this method so we don't re-promisify every time the method is called.

hswolff (changes requested):

website/pages/component/[id]/index.tsx
Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx:29]: 👍

website/pages/component/[id]/index.tsx
Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx:7]: I already have baseURL in my tsconfig so this should just work 🤔

Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx:7]: Indeed it does :)

website/pages/component/[id]/index.tsx
Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx:77]: 👍

website/pages/component/[id]/index.tsx
Comment (brookescarlett) [reply website/pages/component/[id]/index.tsx]: 👍

hswolff (approved):

packages/icon/src/Icon.spec.tsx
lazytype: what changed that would even affect this file?

packages/confirmation-modal/src/ConfirmationModal.spec.tsx
```diff
@@ -119,10 +119,10 @@ describe('packages/confirmation-modal', () => {
         requiredInputText: 'Confirm',
       });
 
-      const confirmationButton = getByText('Confirm');
+      const confirmationButton = getByText('Confirm').parentNode;
```
lazytype, line 122: I propose doing `getByText('Confirm').closest('button')` to make this more robust

website/components/grid/Grid.tsx
```diff
@@ -85,7 +85,7 @@ const baseGridItemStyles = css`
 `;
 
 const visibleGridItemStyles = css`
-  flex-grow: 1;
+  // flex-grow: 1;
```
lazytype, line 88: rm

brookescarlett: 👍

packages/confirmation-modal/src/ConfirmationModal.spec.tsx
brookescarlett: 👍

packages/icon/src/Icon.spec.tsx
brookescarlett: not sure, but the error seems to be coming from the toJson typing

PR #643, opened 2020-10-08
Adds breakpoints to tokens package

brookescarlett:
## ✍️ Proposed changes

- Adds breakpoints to tokens package (precursor to building this into the provider and will help standardize bps across website)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/tokens/src/breakpoints.ts
+++ b/packages/tokens/src/breakpoints.ts
@@ -1,2 +1,8 @@
-const breakpoints = [320, 768, 1024, 1440];
-export default breakpoints;
+const Breakpoints = {
+  Mobile: 320,
+  Tablet: 768,
+  Desktop: 1024,
+  XLDesktop: 1440,
+} as const;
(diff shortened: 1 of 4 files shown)


packages/tokens/src/breakpoints.ts
```diff
@@ -0,0 +1,2 @@
+const breakpoints = [320, 768, 1024, 1440];
```
DesignerDave, line 1: Can we change this to an enum so we can describe each breakpoint? Something like `phone`, `tablet`, `desktop`, `largeDesktop`

DesignerDave (changes requested):

packages/tokens/src/breakpoints.ts
brookescarlett: yes that probably makes more sense... 🤦

DesignerDave (approved):

PR #653, opened 2020-10-14
PD-918: Fixes keyboard navigation bug in Tabs component

brookescarlett:
## ✍️ Proposed changes

- Fixes keyboard navigation bug in Tabs component: Previously, if multiple `<Tabs />` were rendered, only the first on the page would be navigable via keyboard. Now, the currently focused `<Tabs />` will be navigable via keyboard, regardless of location on page.

🎟 _Jira ticket:_ [PD-918](https://jira.mongodb.org/browse/PD-918)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/tabs/src/Tabs.spec.tsx
+++ b/packages/tabs/src/Tabs.spec.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { render, fireEvent, screen } from '@testing-library/react';
-import userEvent from '@testing-library/user-event';
 import { Tabs, Tab } from './index';
 
 const setSelected = jest.fn();
--- a/packages/tabs/src/Tabs.spec.tsx
+++ b/packages/tabs/src/Tabs.spec.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, fireEvent } from '@testing-library/react';
+import { render, fireEvent, screen } from '@testing-library/react';
 import { Tabs, Tab } from './index';
 
 const setSelected = jest.fn();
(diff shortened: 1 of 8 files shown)


packages/tabs/src/Tabs.tsx
```diff
@@ -149,7 +150,9 @@ function Tabs({
   };
 
   const handleArrowKeyPress = (e: KeyboardEvent) => {
-    if (!(e.metaKey || e.ctrlKey)) {
+    const isFocusedTabs = tabsRef?.current?.contains(document.activeElement);
```
lazytype, line 153: isn't the problem that we're using `useEventListener` to bind the event handler to the document instead of binding it just to this component? if the handler is bound to the component this condition would always be true.

lazytype (changes requested):

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.spec.tsx
```diff
@@ -150,4 +150,45 @@ describe('packages/tab', () => {
...
+      const tabSet2A = screen.getByText('Tab Set 2-A');
+
+      const tabSet1AContent = screen.getByText('Content 1-A');
+      const tabSet2AContent = screen.getByText('Content 2-A');
+      expect(tabSet1AContent).toBeInTheDocument();
+      expect(tabSet2AContent).toBeInTheDocument();
+
+      tabSet2A.focus();
+
+      fireEvent.keyDown(tabSet2A, {
```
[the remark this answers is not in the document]
brookescarlett: 👍

lazytype (approved):

PR #660, opened 2020-10-21
Create Selectable component

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Select React Component](https://jira.mongodb.org/browse/PD-271)

## 🛠 Types of changes

## ✅ Checklist


packages/select/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+```js
+import Select, { Option, OptionGroup, Size } from '@leafygreen-ui/select';
+
+<Select
+  label="Label"
+  description="Description"
+  placeholder="Placeholder"
+  name="Name"
+  size={Size.Normal}
```
brookescarlett, line 31: I think we should move in the direction of Size.Default. Most of our components follow this convention anyway (with the exception of Button and Pipeline) and it feels more correct

packages/select/README.md
```diff
@@ -0,0 +1,128 @@
...
+| `disabled`     | `boolean`                                          | Disables the component from being edited.                                                                     | `false`    |
+| `value`        | `string`                                           | Sets the `<Option />` that will appear selected and makes the component a controlled component.               | `''`       |
+| `defaultValue` | `string`                                           | Sets the `<Option />` that will appear selected on page load when the component is uncontrolled.              | `''`       |
+| `onChange`     | `function`                                         | A function that gets called when the selected value changes. Receives the value string as the first argument. | `() => {}` |
+| `readOnly`     | `boolean`                                          | Disables the console warning when the component is controlled and no `onChange` prop is provided.             | `false`    |
+
+## Option
```
brookescarlett, line 111: Can this be a single `#` instead? Trying to standardize readmes for parsing for website directory

packages/select/src/MenuButton.tsx
```diff
@@ -0,0 +1,290 @@
...
+              @media only screen and (max-width: ${breakpoints.Desktop}px) {
+                margin-left: ${mobileSizeSet.option.margin}px;
+              }
+            `,
+          )}
+        >
+          {text}
+        </span>
+
+        <CaretDownIcon
```
brookescarlett, line 272: We should transition this such that it rotates based on whether the Select is open or not

packages/select/src/MenuButton.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+  // Only directly hovering over the button should induce hover
+  const preventChildrenFromTriggeringHoverProps = {
+    onMouseEnter: offHover,
+    onMouseLeave: onHover,
+  };
+
+  return (
+    <div
+      // eslint-disable-next-line jsx-a11y/role-has-required-aria-props
```
brookescarlett, line 181: I'm seeing an error in the accessibility tab:
`Required ARIA child role not present: textbox`

packages/select/src/ListMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { cx, css } from '@leafygreen-ui/emotion';
+import { useViewportSize } from '@leafygreen-ui/hooks';
+import { breakpoints } from '@leafygreen-ui/tokens';
+import SelectContext from './SelectContext';
+import { colorSets, mobileSizeSet, sizeSets } from './styleSets';
+
+const menuStyle = css`
+  position: relative;
+  top: 5px;
+
```
brookescarlett, line 11: nit rm space

packages/select/src/ListMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+  width: 100%;
+  border-radius: 3px;
+
```
brookescarlett, line 14: ditto here

packages/select/src/ListMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+    (event: React.KeyboardEvent) => {
+      // No support for modifiers yet
+      /* istanbul ignore if */
+      if (event.ctrlKey || event.shiftKey || event.altKey) {
+        return;
+      }
+
+      let bubble = false;
+
+      switch (event.keyCode) {
```
brookescarlett, line 56: we have a keyMap in lib with these keycodes aliased 😄

packages/select/src/ListMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+      onKeyDown={onKeyDown}
+      className={cx(
+        menuStyle,
+        css`
+          font-size: ${sizeSet.option.text}px;
+          max-height: ${maxHeight}px;
+          background-color: ${colorSet.option.background.base};
+          box-shadow: 0 3px 7px 0 ${colorSet.menu.shadow};
+
+          @media only screen and (max-width: ${breakpoints.Desktop}px) {
```
brookescarlett, line 107: we use `facepaint` to help with media queries. not sure if necessary here, just a thought

packages/select/src/MenuButton.tsx
```diff
@@ -0,0 +1,290 @@
...
+  readOnly?: boolean;
+  refElement: React.MutableRefObject<HTMLDivElement | null>;
+  onFocusFirstOption: () => void;
+  onFocusLastOption: () => void;
+  onDeselect: () => void;
+  onClose: () => void;
+  onOpen: () => void;
+} & Required<
+  Pick<
+    JSX.IntrinsicElements['div'],
```
brookescarlett, line 48: Know we mentioned that this type includes `ref`, but don't see ref included in the props. Are we fine with leaving as-is? Or were we going to establish a new pattern around this? Wondering because I'm going to try and standardize across the repo

packages/select/src/MenuButton.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+  const { usingKeyboard } = useUsingKeyboardContext();
+
+  const [isHovered, setIsHovered] = useState(false);
+
+  const colorSet = colorSets[mode];
+  const sizeSet = sizeSets[size];
+
+  const baseBoxShadow = `inset 0 -1px 0 0 ${colorSet.shadow.base}`;
+  let boxShadowExpanded = `0 0 0 3px ${colorSet.shadow.expanded}`;
```
brookescarlett, line 78: is there a reason not to turn this into a ternary?

packages/select/src/Option.tsx
```diff
@@ -0,0 +1,221 @@
+import React, { useCallback, useContext, useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { usePrevious } from '@leafygreen-ui/hooks';
+import { LGGlyph } from '@leafygreen-ui/icon/src/types';
+import { colorSets } from './styleSets';
+import SelectContext from './SelectContext';
+
+type GlyphElement = React.ReactElement<LGGlyph.ComponentProps> & {
```
brookescarlett, line 9: Should we move this type to the Icon directory, can imagine it'd be useful across the repo

packages/select/src/Select.tsx
```diff
@@ -0,0 +1,470 @@
...
+  readOnly,
+}: Props) {
+  const id = useMemo(() => idProp ?? idAllocator.generate(), [idProp]);
+  const labelId = `${id}-label`;
+  const descriptionId = `${id}-description`;
+  const menuId = `${id}-menu`;
+
+  const [open, setOpen] = useState(false);
+
+  const menuButtonRef = useSmartRef<HTMLDivElement | null>(null);
```
brookescarlett, line 83: for my own understanding, what's the benefit of this over useElementNode?

packages/select/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export { Size } from './styleSets';
+export { default } from './Select';
```
brookescarlett, line 2: we can probably do this as a named export as well, since you can't use select without options. i think the benchmark was "can you use it alone?" in order for a component to be the default export alongside named exports in a package

packages/select/src/utils.tsx
```diff
@@ -0,0 +1,153 @@
...
+import {
+  InternalOption,
+  InternalProps as InternalOptionProps,
+  OptionElement,
+  ReactEmpty,
+} from './Option';
+import { InternalOptionGroup, OptionGroupElement } from './OptionGroup';
+import { isFragment } from 'react-is';
+
+function isReactEmpty(value: React.ReactNode): value is ReactEmpty {
```
brookescarlett, line 12: TIL ReactEmpty

packages/select/src/Option.tsx
```diff
@@ -0,0 +1,221 @@
...
+  type?: { isGlyph?: boolean };
+};
+
+export type ReactEmpty = null | undefined | false | '';
```
brookescarlett, line 13: Ah, just saw you created this type

brookescarlett:
Very very excited to have this in the repo

packages/select/src/Select.tsx
lazytype:
```tsx
function Parent(props) {
  const [ref, setRef] = useElementNode();

  return <Child ref={setRef} />;
}

function Child({ref: setRef}) {
  const ref = useRef();
  return <div ref={/* how to set both `setRef` and `ref`? */} />;
}
```

```tsx
function Parent(props) {
  const ref = useSmartRef();

  return <Child ref={ref} />;
}

function Child({ref}) {
  // can just access `ref.current`

  return <div ref={ref} />;
}
```

packages/select/src/MenuButton.tsx
lazytype: the warnings are still based on aria 1.1. in aria 1.2 a textbox is not required
https://www.w3.org/TR/wai-aria-1.2/#combobox

packages/select/src/MenuButton.tsx
lazytype: we're only picking the aria props from it here so whether or not it has ref doesn't change the outcome :)

packages/select/src/Option.tsx
lazytype: I think it'd be good to do. hoping to keep it as a follow-up to try to keep this PR more self-contained.

packages/select/src/MenuButton.tsx
lazytype: I think Rob and Dave discussed this and decided on keeping it fixed because it would be more complex to also consider cases where the menu opens upwards (not currently implemented).

brookescarlett: Makes sense

packages/select/src/MenuButton.tsx
brookescarlett: Good to know!

packages/select/src/MenuButton.tsx
brookescarlett: 👍

packages/select/src/Option.tsx
brookescarlett: makes sense to me

packages/select/src/Select.tsx
brookescarlett: Ah! Neat!

packages/select/src/ListMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+}: ListMenuProps) {
+  const { mode, size } = useContext(SelectContext);
+
+  const colorSet = colorSets[mode];
+  const sizeSet = sizeSets[size];
+
+  const onKeyDown = useCallback(
+    (event: React.KeyboardEvent) => {
+      // No support for modifiers yet
+      /* istanbul ignore if */
```
brookescarlett, line 48: do we need these comments in production too?

lazytype: it only affects test coverage reporting. it should get stripped out when bundled

packages/select/src/styleSets.ts
```diff
@@ -0,0 +1,322 @@
...
+  };
+  option: {
+    margin: number;
+    text: number;
+  };
+}
+
+export const sizeSets: Record<Size, SizeSet> = {
+  [Size.XSmall]: {
+    height: 22,
```
DesignerDave, line 222: What's the benefit to this versus simply writing out the CSS declarations here?

DesignerDave (commented):
Just took a first look at this PR, and noticed a few things from an interaction perspective.

- The dropdown doesn't transition in any way. Is there a reason we don't use Portal for this component? It has transitions built in.
- The gradient, and border for the trigger don't transition between states. We should transition those like we do our button (if not using the button itself directly).
- The interaction ring's background color doesn't appear to transition between background colors.
- In general, we should transition between all states (selected, hover, focus, open).
- It looks like the line-height of wrapping options is a bit too tight. Is that up to spec?
- I think @audroue has some designs ready for an updated selected state for the component. That would be great to include.

packages/select/README.md
```diff
@@ -0,0 +1,130 @@
...
+    ></span>
+  </button>
+</div>
+```
+
+## Select Properties
+
+| Prop           | Type                                                | Description                                                                                                   | Default     |
+| -------------- | --------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- | ----------- |
+| `children`     | `node`                                              | `<Option />` and `<OptionGroup />` elements.                                                                  |             |
```
brookescarlett, line 99: `name` prop is missing from documentation here

packages/select/README.md
```diff
@@ -0,0 +1,130 @@
...
+| `description`  | `string`                                            | Text that gives more detail about the requirements for the input.                                             |             |
+| `placeholder`  | `string`                                            | The placeholder text shown in the input element when an option is not selected.                               | `'Select'`  |
+| `disabled`     | `boolean`                                           | Disables the component from being edited.                                                                     | `false`     |
+| `value`        | `string`                                            | Sets the `<Option />` that will appear selected and makes the component a controlled component.               | `''`        |
+| `defaultValue` | `string`                                            | Sets the `<Option />` that will appear selected on page load when the component is uncontrolled.              | `''`        |
+| `onChange`     | `function`                                          | A function that gets called when the selected value changes. Receives the value string as the first argument. | `() => {}`  |
```
brookescarlett, line 110: i think we should keep the onChange function signature consistent (so passing up the entire event rather than jus the value, even though the value will be what's used in most use-cases)

packages/select/src/Select.tsx
```diff
@@ -0,0 +1,488 @@
...
+  description?: string;
+  placeholder?: string;
+  name: string;
+} & (
+  | // Uncontrolled
+  ({
+      defaultValue?: string;
+      value?: undefined;
+    } & {
+      onChange?: (value: string) => void;
```
brookescarlett, line 45: mentioned above, but I think we should pass up the entire event for consistency-sake

packages/select/src/Select.tsx
```diff
@@ -0,0 +1,488 @@
...
+  children: React.ReactNode;
+  className?: string;
+  id?: string;
+  darkMode?: boolean;
+  size?: Size;
+  disabled?: boolean;
+  label: string;
+  description?: string;
+  placeholder?: string;
+  name: string;
```
brookescarlett, line 38: why is this a required prop?

packages/select/src/Select.tsx
```diff
@@ -0,0 +1,488 @@
...
+`;
+
+export type Props = {
+  children: React.ReactNode;
+  className?: string;
+  id?: string;
+  darkMode?: boolean;
+  size?: Size;
+  disabled?: boolean;
+  label: string;
```
brookescarlett, line 35: we should allow for passing aria-labelledby instead of only accepting label

packages/select/src/Select.tsx
```diff
@@ -0,0 +1,488 @@
...
+      const index = Math.max(0, enabledOptions.indexOf(focusedOption) - 1);
+      setFocusedOption(enabledOptions[index]);
+    }
+  }, [enabledOptions, focusedOption, onFocusLastOption]);
+
+  const onFocusNextOption = useCallback(() => {
+    if (focusedOption === undefined) {
+      onFocusFirstOption();
+    } else {
+      const index = Math.min(
```
brookescarlett, line 266: We should have this wrap such that the first item gets focused after we've reached the end of the list, rather than not moving forward

packages/select/src/Select.tsx
brookescarlett: Still feel pretty strongly about this

packages/select/src/Select.tsx
brookescarlett: ^

packages/select/README.md
```diff
@@ -0,0 +1,131 @@
...
+| `id`           | `string`                                            | id associated with the Select component.                                                                      |             |
+| `label`        | `string`                                            | Text shown in bold above the input element.                                                                   |             |
+| `description`  | `string`                                            | Text that gives more detail about the requirements for the input.                                             |             |
+| `placeholder`  | `string`                                            | The placeholder text shown in the input element when an option is not selected.                               | `'Select'`  |
+| `disabled`     | `boolean`                                           | Disables the component from being edited.                                                                     | `false`     |
+| `value`        | `string`                                            | Sets the `<Option />` that will appear selected and makes the component a controlled component.               | `''`        |
```
brookescarlett, line 108: Should we consider allowing ReactElements to be passed here as well? A use case I'm thinking about is when we have Icons that would be part of Option value. Let me know what you think cc @DesignerDave

DesignerDave: Agreed, just because weird use-cases do come up, and it's low effort to support the flexibility.

DesignerDave (changes requested):
Hey Michael! This is looking close.

There's a few design-related issues to note that I can see:
- The padding to the left of the text in the select input toggle should be aligned with the padding of the labels in the dropdown. The caret on the right should also reflect this spacing.
- The hover state for select options should transition.
- Clicking the labels for option groups closes the dropdown. I'd confirm with Rob what the behavior should be here, but this feels like it would be annoying if you misclick when selecting an option.
- The dropdown should animate its position as it opens instead of only fading in (though we should keep the opacity transition as well). Adding a `translate3d(0, -8px, 0)` to the inactive state, and `translate3d(0, 0, 0)` to the active one, and transitioning the value should improve things!

We might even want to go a little further and mimic our popover component:

```css
# inactive
transition: opacity 150ms ease-in-out, transform 150ms ease-in-out;
opacity: 0;
transform: translate3d(0, -8px, 0) scale(0.95);
transform-origin: center top;

# active
opacity: 1;
transform: translate3d(0, -8px, 0) scale(0.95);
```

lazytype (commented):
> The caret on the right should also reflect this spacing.

Worth pointing out that the caret might not necessarily align with option labels since there may or may not be a scrollbar

packages/select/README.md
DesignerDave: You're right 🤦 I was thinking this was in reference to the display value for an option for some reason.

DesignerDave:
> > The caret on the right should also reflect this spacing.
> 
> Worth pointing out that the caret might not necessarily align with option labels since there may or may not be a scrollbar

That's fine! It'll look better visually overall, and align better with the specs.

brookescarlett (approved):
Just want to note that I noticed that the combobox aria role is incorrectly applied here (as raised by both the accessibility knob and axe dev tools). I don't want to spend time trying to figure out on our own how to make this more accessible, as we're getting the results of the audit back on Wednesday, and I think their guidance will be very instructive in terms of next steps.

In the meantime, happy to get this in (maybe as a beta, since we know there will be some changes coming soon? cc: @DesignerDave )!  Very excited to have this component in the repo.

lazytype:
> Just want to note that I noticed that the combobox aria role is incorrectly applied here (as raised by both the accessibility knob and axe dev tools).

I've brought this up in a previous meeting but I can document it here for posterity. I've implemented combobox in conformance with the WAI-ARIA 1.2 spec https://www.w3.org/TR/wai-aria-1.2/#combobox. The accessibility guidance provided by storybook and axe is still based on WAI-ARIA 1.1 https://www.w3.org/TR/wai-aria-1.1/#combobox

lazytype:
https://github.com/dequelabs/axe-core/issues/2505 is the tracking issue for Deque to update axe

brookescarlett:
> [dequelabs/axe-core#2505](https://github.com/dequelabs/axe-core/issues/2505) is the tracking issue for Deque to update axe

SGTM

DesignerDave (approved):
Let's get folks pressure-testing this!

PR #666, opened 2020-10-26
Fix React 17 issues

lazytype:
This fixes support in components for React 17. The breakages were primarily due to `useEffect` cleanup functions running asynchronously. This meant that `ref.current` would be stale in most cases. The fixes generally follow the guidance provided here https://reactjs.org/blog/2020/08/10/react-v17-rc.html#potential-issues .

In this PR I also deprecate `useElementNode`. In general, I think it provides little value over using `setState` and has a glaring bug in that it doesn't `null` out the variable when the reference element is unmounted. I also renamed variables that had `ref` in the name that were not refs, but HTMLElements. 

## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


packages/menu/src/SubMenu.tsx
```diff
@@ -172,7 +172,7 @@ interface SubMenuProps {
   /**
    * Content to appear below main text of SubMenu
    */
-  description?: React.ReactElement;
+  description?: string | React.ReactElement;
```
lazytype, line 175: the story file wasn't getting type-checked because it was a .js file. type-checking revealed that this prop wasn't typed correctly.

brookescarlett (approved):
LGTM

PR #673, opened 2020-10-28
PD-345: TextArea component

brookescarlett:
## ✍️ Proposed changes

Worked with @audroue  to create a Beta release of the TextArea component 

🎟 _Jira ticket:_ [PD-345](https://jira.mongodb.org/browse/PD-345)

## 🛠 Types of changes

## ✅ Checklist

Screen Shot 2020-10-28 at 11 44 40 AM
Screen Shot 2020-10-28 at 11 44 36 AM


lazytype:
since this is at least the third time we've included a label + description with a component, it seems like a good time to think about starting to share the implementation between components. Maybe include Label and Description as part of a @leafygreen-ui/form package? @DesignerDave would like your thoughts as well.

brookescarlett:
I don't hate that idea at all. Maybe even include error message?
```
const Form = ({children}) => {
  return (
    <>
      <label>{label}</label>
      <p>{description}</p>
      {children}
      {errorMessage && <div>{errorMessage}</div>
    </>
  )
}
```

brookescarlett:
@DesignerDave what do you think?

packages/text-area/src/TextArea.tsx
```diff
@@ -0,0 +1,278 @@
...
+  border-radius: 7px;
+  position: absolute;
+  top: -3px;
+  bottom: -3px;
+  left: -3px;
+  right: -3px;
+  pointer-events: none;
+  background-color: pink;
+
+  ${textAreaProp.selector}:hover ~ & {
```
DesignerDave, line 85: To clean this up a little, you can write these two styles as:

```
${textAreaProp.selector} {
  &:hover ~ &,
  &:focus ~ & {
    transform: scale(1);
  }
}
```

packages/text-area/src/TextArea.tsx
```diff
@@ -0,0 +1,278 @@
...
+const textAreaStyle = css`
+  width: 100%;
+  min-height: ${spacing[6]}px;
+  resize: none;
+  border-radius: 4px;
+  padding-left: 12px;
+  font-size: 14px;
+  font-weight: normal;
+  font-family: ${fontFamilies.default};
+  z-index: 1;
```
DesignerDave, line 62: The border for the textArea doesn't currently match textInput. We should match the static and focus state border colors exactly.

packages/text-area/src/TextArea.tsx
```diff
@@ -0,0 +1,278 @@
...
+    textAreaColor: uiColors.gray.dark3,
+    textAreaBackgroundColor: uiColors.white,
+    disabledColor: uiColors.gray.base,
+    disabledBackgroundColor: uiColors.gray.light2,
+    interactionRing: uiColors.gray.light2,
+    interactionRingFocus: '#9dd0e7',
+    errorBorder: uiColors.red.base,
+    errorMessage: uiColors.red.base,
+  },
+  [Mode.Dark]: {
```
DesignerDave, line 127: I think I prefer the approach to modes and variants used in [Tabs](https://github.com/mongodb/leafygreen-ui/blob/master/packages/tabs/src/Tabs.tsx#L19) and [Toast](https://github.com/mongodb/leafygreen-ui/blob/master/packages/toast/src/Toast.tsx#L74) to this approach. I find this much harder to read, and I'd like to see us to consolidate our approach to style variants like this over time.

An added benefit to those approaches is not having to inline those stateful styles like you are now (something that I believe you've expressed disliking in the past).

.changeset/violet-rings-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/text-area': minor
```
DesignerDave, line 2: We don't need to do a beta release of this component – it's pretty straightforward.

brookescarlett: I asked Rob if he wanted to beta it because he hasn't had a chance to do an audit of the texarea component, so wasn't sure if the styles would change a little more significantly at some point

brookescarlett:
@DesignerDave what'd you think about creating a form package as @lazytype mentioned above

DesignerDave:
> since this is at least the third time we've included a label + description with a component, it seems like a good time to think about starting to share the implementation between components. Maybe include Label and Description as part of a @leafygreen-ui/form package? @DesignerDave, I'd like your thoughts as well.

Not opposed to this at all, though I think this might be a bit premature. I think the best start would be adding `Label` / `Description` components to `@leafygreen-ui/typography` first. That would be usable in a more broad spectrum of use-cases, and consolidate the styles for these. Once we have that for a bit, I think then we'd consider making a combined component.

DesignerDave:
Also, if we do this lets make it its own PR, rather than including the change here.

.changeset/violet-rings-prove.md
DesignerDave: The API is unlikely to change, as is the design (given how similar the component is to TextArea. I think we should just v1 this.

packages/text-area/src/TextArea.tsx
brookescarlett: 👍

packages/text-area/src/TextArea.tsx
brookescarlett: 🤔 In light mode: 
interactionRing: uiColors.gray.light2,
interactionRingFocus: '#9dd0e7', (same across both components)

In dark mode: 
interactionRing: uiColors.gray.dark1,
interactionRingFocus: uiColors.blue.base,

packages/text-area/src/TextArea.tsx
brookescarlett: An object of objects you mean?

.changeset/violet-rings-prove.md
brookescarlett: Kk that's fine with me, will update now

packages/text-area/src/TextArea.tsx
DesignerDave: An object of `css''` declarations so that you have the full declaration rather than just a color that's consumed elswehere for each state. Tabs is a good example.

packages/text-area/src/TextArea.tsx
DesignerDave: Not the interaction rings, the borders themselves.

packages/text-area/src/TextArea.tsx
brookescarlett: Happy to update

brookescarlett:
```
interface ColorSets {
  labelColor: string;
  descriptionColor: string;
  textArea: {
    textAreaBackgroundColor: string;
    textAreaColor: string;
  };
  disabled: {
    backgroundColor: string;
    color: string;
  };
  interactionRing: {
    backgroundColor: string;
    focusColor: string;
  };
  error: {
    border: string;
    message: string;
  };
}```

Comment (brookescarlett) [reply packages/text-area/src/TextArea.tsx:127]: is that what you had in mind?

File: packages/text-area/src/TextArea.tsx (hunk 1 above)
Comment (brookescarlett) [reply packages/text-area/src/TextArea.tsx]: This implementation actually doesn't work

File: packages/text-area/src/TextArea.tsx (hunk 3 above)
Comment (DesignerDave) [reply packages/text-area/src/TextArea.tsx:146]: More like:
```js
const modes =  {
  [Modes.Dark]: {
    label: css``,
    description: css``,
    textArea: {
      static: css``,
      disabled: css``,
    },
    ...etc
  }
}
```

packages/text-area/src/TextArea.tsx
DesignerDave: What does the CSS output look like? It should work.

DesignerDave (approved):

PR #676, opened 2020-11-06
Adds Label and Description components to Typography package

brookescarlett:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/typography/src/FormElements.tsx
+++ b/packages/typography/src/FormElements.tsx
@@ -42,14 +42,22 @@ const labelStyle = css`
 
 type LabelProps = JSX.IntrinsicElements['label'] & {
   darkMode?: boolean;
-  id: string | number;
+  htmlFor: string;
 };
 
-const Label = ({ id, darkMode = false, children }: LabelProps) => {
+const Label = ({
+  darkMode = false,
+  className,
+  children,
+  ...rest
+}: LabelProps) => {
   const mode = darkMode ? Mode.Dark : Mode.Light;
 
   return (
-    <label htmlFor={id} className={cx(labelStyle, colorSets[mode].labelColor)}>
+    <label
+      className={cx(labelStyle, colorSets[mode].labelColor, className)}
+      {...rest}
+    >
       {children}
     </label>
   );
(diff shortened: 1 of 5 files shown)


packages/typography/src/FormElements.tsx
```diff
@@ -0,0 +1,85 @@
...
+const labelStyle = css`
+  font-size: 14px;
+  font-weight: bold;
+  line-height: 16px;
+  padding-bottom: 4px;
+`;
+
+type LabelProps = JSX.IntrinsicElements['label'] & {
+  darkMode?: boolean;
+  id: string | number;
```
lazytype, line 45: I would call this `inputId` so it doesn't collide with the `id` of the label element itself. should we omit the `htmlFor` property if we're going to rename it? is there a good case for renaming it in the first place?

packages/typography/src/FormElements.tsx
```diff
@@ -0,0 +1,85 @@
...
+};
+
+const Label = ({ id, darkMode = false, children }: LabelProps) => {
+  const mode = darkMode ? Mode.Dark : Mode.Light;
+
+  return (
+    <label htmlFor={id} className={cx(labelStyle, colorSets[mode].labelColor)}>
```
lazytype, line 52: we're not forwarding `className` or the rest of the label props

packages/typography/src/FormElements.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+type DescriptionProps = JSX.IntrinsicElements['p'] & {
+  darkMode?: boolean;
+};
+
+const Description = ({ darkMode = false, children }: DescriptionProps) => {
+  const mode = darkMode ? Mode.Dark : Mode.Light;
+
+  return (
+    <p className={cx(descriptionStyle, colorSets[mode].descriptionColor)}>
```
lazytype, line 77: we're not forwarding props here either

lazytype (changes requested):

packages/typography/src/FormElements.tsx
brookescarlett: Nope, good call... will omit id and just spread htmlFor in rest. Just want to make sure a value for the prop is supplied, so will be sure to type appropriately

packages/typography/src/FormElements.tsx
brookescarlett: great call, thank you

packages/typography/src/FormElements.tsx
brookescarlett: fixed

packages/typography/src/FormElements.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+Label.displayName = 'Label';
+
+const descriptionStyle = css`
+  font-size: 14px;
+  line-height: 16px;
+  font-weight: normal;
+  padding-bottom: 4px;
+  margin-top: 0px;
+  margin-bottom: 0px;
```
lazytype, line 74:
```suggestion
  margin-top: 0;
  margin-bottom: 0;
```

lazytype (approved):

PR #681, opened 2020-11-10
Add custom focus ring to Button

lazytype:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


packages/button/README.md
```diff
@@ -58,6 +58,7 @@ import Button from '@leafygreen-ui/button';
 | Prop        | Type                                                             | Description                                                                                                                                           | Default     |
 | ----------- | ---------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- | ----------- |
 | `variant`   | `'default'` \| `'primary'` \| `'info'` \| `'danger'` \| `'dark'` | Sets the style variant of the button.                                                                                                                 | `'default'` |
+| `darkMode`  | `boolean`                                                        | Determines if the component renders in dark mode                                                                                                      | `false`     |
```
brookescarlett, line 61: Let's also add as a knob to storybook

packages/button/src/Button.tsx
```diff
@@ -329,16 +106,76 @@ const Button: ExtendableBox<
...
+
     const { usingKeyboard: showFocus } = useUsingKeyboardContext();
 
     const commonProps = {
       ref,
       className: cx(
         baseStyle,
-        buttonSizes[size],
-        buttonVariants[variant],
+        css`
```
brookescarlett, line 118: I think @DesignerDave will suggest that we break this entire style up such that we do 

```
cx(baseStyle, colorSet[mode], ...)
```

brookescarlett: I'm open to exploring other patterns, but do think it's probably time for us to define one approach to this and be consistent, since right now we're doing it a bunch of different ways around the repo

brookescarlett: (ditto everywhere else)

brookescarlett:
Happy that we're applying the interaction ring pattern to button! I don't want to be a blocker on this one but would just suggest that we style QA with @audroue and/or @DesignerDave before merging because there were some cases where the border looked a bit off (although that could be a function of the gradient backgrounds, which I believe we're ultimately transitioning away from...)

packages/button/src/styleSets.ts
```diff
@@ -0,0 +1,248 @@
...
+      },
+      image: {
+        base: `linear-gradient(${uiColors.gray.base}, ${uiColors.gray.dark1})`,
+        active: `linear-gradient(${uiColors.gray.dark1}, ${uiColors.gray.base})`,
+        'focus/hover': `linear-gradient(${uiColors.gray.base}, ${uiColors.gray.dark2})`,
+      },
+      shadow: {
+        base: `inset 0 -1px 0 ${uiColors.gray.dark2}`,
+        active: `inset 0 2px 2px ${uiColors.gray.dark2}`,
+        'focus/hover': {
```
DesignerDave, line 154: Using a box-shadow for the interaction ring is inconsistent with other applications in our codebase. This is important since we aren't transitioning the scale of the interaction ring here, just its transparency, so the animation is inconsistent, and less fluid. We should use the approach we have been to apply the interaction ring.

As an aside, the color set approach to variant and mode styling made finding how this style was applied, and where to leave this comment very hard. I'd prefer we take the approach I noted above to our styles, even if it's less typeable.

DesignerDave (changes requested):

packages/button/src/styleSets.ts
lazytype: >As an aside, the color set approach to variant and mode styling made finding how this style was applied, and where to leave this comment very hard. I'd prefer we take the approach I noted above to our styles, even if it's less typeable.

I'm curious as to why that was the case. I would've thought that the definition of `focusRing` would make these things straightforward to find:
https://github.com/mongodb/leafygreen-ui/pull/681/files#diff-fd57aaba57b850d63607350a5dd87f40dff04d2eff66a61ea5b9c363c93d26d6R106
```

packages/button/src/Button.tsx
lazytype: Could you go more into what about this approach has made it work the best? My main concerns with it are that it doesn't really enable DRY-ness and doesn't streamline the process of adding of new variants as much.

lazytype:
Closing in favor of #690

PR #685, opened 2020-11-12
PD-936: Tabs accepts empty children

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-936](https://jira.mongodb.org/browse/PD-936)

## 🛠 Types of changes

## ✅ Checklist


packages/tabs/src/Tabs.tsx
```diff
@@ -184,7 +186,9 @@ function Tabs({
         role="tablist"
         tabIndex={0}
       >
-        {tabs.map((tab, index) => {
+        {tabs?.map((tab, index) => {
+          if (!tab) return null;
```
lazytype, line 190: not completely safe for non-TS users since we don't necessarily want to return `null` for values like `0`

brookescarlett: is this safer: 
```
if (!isComponentType(tab, 'Tab')) {
            return tab
          }
```

lazytype: seems like a reasonable enough behavior to me

lazytype (approved):

PR #689, opened 2020-11-13
PD-940: Copyable Updates

brookescarlett:
## ✍️ Proposed changes

- Makes label optional (since InlineCode is not a form element, this shouldn't break any rules of accessibility)
- Makes Component scrollable on the x-axis
- Removes explicitly setting z-index which was causing problems with stacking contexts in .design
- Updates CodeDocs to use Copyable component

Note: didn't create a changeset yet, because we didn't release Copyable yet (right?) so was hoping we could just publish these with the initial release

🎟 _Jira ticket:_ [PD-940](https://jira.mongodb.org/browse/PD-940)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/copyable/src/Copyable.tsx
+++ b/packages/copyable/src/Copyable.tsx
@@ -173,7 +173,7 @@ export default function Copyable({
     }
 
     // Forward darkMode prop for future versions of Button that support it
-    // const buttonRestProps: {} = { darkMode };
+    const buttonRestProps = 'darkMode' in Button.propTypes! ? { darkMode } : {};
 
     const trigger = (
       <Button
@@ -181,6 +181,7 @@ export default function Copyable({
         variant={buttonVariant}
         className={buttonStyle}
         onClick={() => setCopied(true)}
+        {...buttonRestProps}
       >
         <CopyIcon size="large" className={iconStyle} />
         Copy
--- a/packages/copyable/src/Copyable.tsx
+++ b/packages/copyable/src/Copyable.tsx
@@ -70,12 +70,14 @@ const codeStyle = css`
   display: inline-flex;
   align-items: center;
   height: 100%;
-  width: 100%;
+  // Button is 75px wide. Need to set a smaller width so that we're able to scroll when text overflows.
+  width: calc(100% - 75px);
   padding-left: 12px;
   border: 1px solid;
   border-radius: 4px;
   font-size: 14px;
-  overflow: hidden;
+  overflow-x: auto;
+  overflow-y: hidden;
   white-space: nowrap;
 `;
 
@@ -87,7 +89,6 @@ const buttonWrapperStyle = css`
   display: inline-block;
   height: 100%;
   position: absolute;
-  z-index: 1;
   right: 0;
   top: 0;
 
@@ -129,7 +130,7 @@ const iconStyle = css`
 interface CopyableProps {
   darkMode?: boolean;
   children: string;
-  label: string;
+  label?: string;
   description?: string;
   className?: string;
   copyable?: boolean;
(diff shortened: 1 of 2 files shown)


packages/copyable/src/Copyable.tsx
```diff
@@ -70,12 +70,14 @@ const codeStyle = css`
   display: inline-flex;
   align-items: center;
   height: 100%;
-  width: 100%;
+  // Button is 75px wide. Need to set a smaller width so that we're able to scroll when text overflows.
```
lazytype, line 73: this doesn't seem very reliable since the button is sized by its contents

packages/copyable/src/Copyable.tsx
```diff
@@ -70,12 +70,14 @@ const codeStyle = css`
...
+  width: calc(100% - 75px);
   padding-left: 12px;
   border: 1px solid;
   border-radius: 4px;
   font-size: 14px;
-  overflow: hidden;
+  overflow-x: auto;
+  overflow-y: hidden;
```
lazytype, line 80: theoretically it shouldn't be possible to overflow in the y direction since the child must be a single string

packages/copyable/src/Copyable.tsx
```diff
@@ -70,12 +70,14 @@ const codeStyle = css`
...
   height: 100%;
-  width: 100%;
+  // Button is 75px wide. Need to set a smaller width so that we're able to scroll when text overflows.
+  width: calc(100% - 75px);
   padding-left: 12px;
   border: 1px solid;
   border-radius: 4px;
   font-size: 14px;
-  overflow: hidden;
+  overflow-x: auto;
```
lazytype, line 79: I intentionally made the component non-scrollable because it doesn't look very good with a scrollbar. I think clients should just set a longer width if their content is too long. If it can't fit then maybe they should consider whether the <Code/> component better fits their use-case.

packages/copyable/src/Copyable.tsx
```diff
@@ -175,15 +176,14 @@ export default function Copyable({
     }
 
     // Forward darkMode prop for future versions of Button that support it
-    const buttonRestProps: {} = { darkMode };
+    // const buttonRestProps: {} = { darkMode };
```
lazytype, line 179: why?

lazytype (changes requested):

packages/copyable/src/Copyable.tsx
brookescarlett: I think you're right on this one and so I'm happy to revert the changes back around scrolling. I'll try manually setting the width if my use-case requires in mdb.design

packages/copyable/src/Copyable.tsx
brookescarlett: Right now Button doesn't have a darkMode prop so I'm getting an error saying that darkMode is not a valid HTML attribute

lazytype: hm, can we do something like this:
```tsx
const buttonRestProps = 'darkMode' in Button.propTypes ? { darkMode } : {};
```

brookescarlett: 👍

packages/copyable/src/Copyable.tsx
```diff
@@ -75,7 +75,6 @@ const codeStyle = css`
   border: 1px solid;
   border-radius: 4px;
   font-size: 14px;
-  overflow: hidden;
```
lazytype, line 78: I think we still need this

brookescarlett: 👍

lazytype (approved):

PR #690, opened 2020-11-13
Create Interaction Ring component and replace bespoke implementations in other components

lazytype:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


.changeset/tall-cars-repeat.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/typography': minor
+---
+
+- `Label` component now supports a `disabled` prop
+- Added an `InteractiveRing` component for advanced use cases where one wants to apply standard visual hover and focus states to wrapped components.
```
brookescarlett, line 6: Nit: the second bullet doesn't belong here

packages/button/README.md
```diff
@@ -55,18 +55,20 @@ import Button from '@leafygreen-ui/button';
...
+| `as`         | `React.ElementType`                                              | Determines the root element. For example, `Link` or `a` tags can be supplied to replace `button` from being the DOM element that wraps the component. | `button`    |
+| `href`       | `string`                                                         | If a href is supplied it will change the `as` value, such that the component renders inside of an `a` tag instead of inside of a `button` tag.        |             |
+| `glyph`      | `React.ReactElement`                                             | When supplied a rendered Icon, renders a button with the specified glyph on the left-side of the content, and adjusts Button padding accordingly.     |             |
+| `forceState` | `{ focused?: boolean, active?: boolean }`                        | Force the component to display the specified interaction states.                                                                                      |             |
```
brookescarlett, line 71: why are we adding this prop?

DesignerDave: He ran this one by me. There are cases where we need to force a hover state for, say, an active menu triggered by a button.

packages/interaction-ring/src/InteractionRing.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+interface ColorSet {
+  interactionRingHover: string;
+  interactionRingFocus: string;
+}
+
+const colorSets: Record<Mode, ColorSet> = {
+  [Mode.Light]: {
+    interactionRingHover: uiColors.gray.light2,
+    interactionRingFocus: '#9dd0e7',
```
brookescarlett, line 21: @DesignerDave we should codify this value somewhere

DesignerDave: I'm aware – outside the scope of this though.

An aside, I'm still not a fan of the colorSets approach. Makes it much more difficult to see the underlying CSS implementation, which is more important than being able to type this IMO.

packages/radio-group/src/Radio.tsx
```diff
@@ -67,7 +68,7 @@ const labelStyle = css`
 // Note colors are not in our palette
 const inputColorSet = {
   [Mode.Light]: css`
-    &:checked + ${inputDisplayWrapper.selector} > ${styledDiv.selector} {
+    &:checked + * ${inputDisplayWrapper.selector} > ${styledDiv.selector} {
```
DesignerDave, line 71: Huge red flags go off whenever I see `*` selectors. They present both performance and maintainability problems, and we should avoid them at all costs. We should find a different way to do this, even if it's a bit more work on our end.

packages/interaction-ring/src/InteractionRing.tsx
```diff
@@ -0,0 +1,212 @@
...
+}: InteractionRingProps) {
+  const mode = darkMode ? Mode.Dark : Mode.Light;
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+
+  const [hovered, setHovered] = useState(false);
+  const [focused, setFocused] = useState(false);
+
+  const interactionRingStyle = getInteractionRingStyle({
+    mode,
+    hovered: forceState.hovered ?? hovered,
```
DesignerDave, line 132: Why are we relying on JavaScript to maintain hover and focus states? That's likely to be far more brittle than using CSS, and also causes re-renders where a CSS solution. wont.

packages/radio-box-group/src/RadioBox.tsx
```diff
@@ -33,129 +33,92 @@ const inputStyles = css`
...
-    ${radioBoxWrapper.selector}:hover & {
-      border-color: ${uiColors.gray.base};
-    }
+    width: 100%;
+    // Display behind border
+    z-index: -1;
   `;
 
-  if (disabled) {
+  if (checked) {
```
brookescarlett, line 46: Should we add `active` to the force states in the InteractionRing component? That way we don't have to manually change the zindex in all of the components consuming the interaction ring (since it's largely form elements, most do have an active state)

packages/radio-box-group/src/RadioBox.tsx
```diff
@@ -33,129 +33,92 @@ const inputStyles = css`
...
       baseStyles,
       css`
-        background-color: ${uiColors.green.base};
-        transform: scale(1);
-        z-index: 1;
-
-        ${radioBoxWrapper.selector}:hover & {
-          background-color: ${uiColors.green.base};
-        }
+        // border: 1px solid rgba(0, 0, 0, 0);
```
brookescarlett, line 116: rm

packages/interaction-ring/src/InteractionRing.tsx
```diff
@@ -0,0 +1,212 @@
...
+  font-family: inherit;
+  font-size: 100%;
+`;
+
+interface InteractionRingProps {
+  darkMode?: boolean;
+  className?: string;
+  borderRadius?: string;
+  color?: {
+    focused?: string;
```
brookescarlett, line 102: should this be nested such that we can set colors for dark mode and light mode respectively?

packages/text-area/src/TextArea.tsx
```diff
@@ -1,11 +1,12 @@
 import React, { useState, useMemo } from 'react';
-import { IdAllocator, Either, createDataProp } from '@leafygreen-ui/lib';
 import { css, cx } from '@leafygreen-ui/emotion';
+import InteractionRing from '@leafygreen-ui/interaction-ring';
+import { IdAllocator, Either } from '@leafygreen-ui/lib';
 import { uiColors } from '@leafygreen-ui/palette';
 import { spacing, fontFamilies } from '@leafygreen-ui/tokens';
+import { Description, Label } from '@leafygreen-ui/typography';
```
brookescarlett, line 7: 👍

packages/text-input/tsconfig.json
```diff
@@ -19,11 +19,17 @@
     {
       "path": "../icon"
     },
+    {
+      "path": "../interaction-ring"
+    },
     {
```
brookescarlett, line 25: Can we remove this now, since we're not explicitly relying on the package anymore?

packages/typography/src/FormElements.tsx
```diff
@@ -43,19 +50,26 @@ const labelStyle = css`
...
   darkMode?: boolean;
   htmlFor: string;
+  disabled?: boolean;
 };
 
 const Label = ({
   darkMode = false,
   className,
   children,
+  disabled = false,
```
brookescarlett, line 60: Good catch, ty

brookescarlett:
Stoked to have this in, we've needed this for a while

packages/interaction-ring/src/InteractionRing.tsx
lazytype: we could probably get away with handling hover in css, but focus would be pretty difficult. the problem is that we don't know where the focused element might be in the DOM and would have a hard time writing a selector. `RadioBox` is an example of this. we could force the constraint that the focused element is contained within the interaction ring but that would require reworking some of the components like `RadioBox`.

packages/radio-box-group/src/RadioBox.tsx
lazytype: Are you proposing that we make the interaction ring have a negative z-index for the active state? I think this wouldn't make a difference for most components since there's no overlap between the interaction ring and the surrounded element. `RadioBox` is kind of exceptional in that sense and I think I've heard that we have some interest updating the design of `RadioBox` at some point

packages/radio-box-group/src/RadioBox.tsx
lazytype: 👍

packages/radio-group/src/Radio.tsx
lazytype: I think a quick alternative would be to add some `>` nesting instead but it would be somewhat brittle

packages/text-input/tsconfig.json
lazytype: 👍

packages/radio-group/src/Radio.tsx
lazytype: I changed it to use "* >" which should be significantly more performant since it's gone from:
"given all elements matching `styledDiv.selector` find all ancestors that have a checked input sibling" to
"given all elements matching `styledDiv.selector` find parents that have a checked input sibling"

packages/interaction-ring/src/InteractionRing.tsx
brookescarlett: Reasonable enough, wfm.

packages/radio-box-group/src/RadioBox.tsx
brookescarlett: Kk reasonable enough, was just wondering if this was repeated logic around the consuming components that may as well be shared while we're building this out. But if this is more exceptional than I thought, then I'm fine to leave as-is

packages/radio-group/src/Radio.tsx
DesignerDave: I'd **vastly** prefer that we use selectors not utilizing `*` at all for the other reasons I mentioned as well, not just performance. Using a child instead of descendent selector is an improvement, but not the issue I had  given the selector.

packages/interaction-ring/src/InteractionRing.tsx
DesignerDave: I see – I suppose that's one of the downsides to the generalized approach. If we need to take this approach, then c'est la vie.

packages/interaction-ring/README.md
```diff
@@ -0,0 +1,55 @@
...
+const [inputElement, setInputElement] =
+  (React.useState < HTMLElement) | (null > null);
+
+<InteractionRing focusTargetElement={inputElement}>
+  <div className="Div appearing as button">
+    Click me
+    <input className="Visually hidden input" ref={setInputElement} />
+  </div>
+</InteractionRing>;
+```
```
brookescarlett, line 41: nit: html missing from example

packages/text-area/src/TextArea.tsx
```diff
@@ -247,21 +174,13 @@ export default function TextArea({
   return (
     <div className={containerStyles}>
       {label && (
-        <label
-          htmlFor={id}
-          className={cx(labelStyle, colorSets[mode].labelColor)}
-        >
+        <Label htmlFor={id} disabled={disabled}>
```
brookescarlett, line 177: darkmode prop needs to be passed here

packages/text-area/src/TextArea.tsx
```diff
@@ -247,21 +174,13 @@ export default function TextArea({
...
-        </label>
+        </Label>
       )}
-      {description && (
-        <p className={cx(descriptionStyle, colorSets[mode].descriptionColor)}>
-          {description}
-        </p>
-      )}
-      <div className={textAreaContainer}>
+      {description && <Description>{description}</Description>}
```
brookescarlett, line 181: ditto here

packages/text-input/src/TextInput.tsx
```diff
@@ -382,63 +329,48 @@ const TextInput = React.forwardRef(
...
-            className={cx(
-              labelStyle,
-              css`
-                color: ${disabled
-                  ? colorSets[mode].disabledLabelColor
-                  : colorSets[mode].labelColor};
-              `,
-            )}
-          >
+          <Label htmlFor={id} disabled={disabled}>
```
brookescarlett, line 332: darkMode prop needs to be passed here

packages/text-input/src/TextInput.tsx
```diff
@@ -382,63 +329,48 @@ const TextInput = React.forwardRef(
...
-              css`
-                color: ${colorSets[mode].descriptionColor};
-              `,
-            )}
-          >
-            {description}
-          </p>
+          </Label>
         )}
+        {description && <Description>{description}</Description>}
```
brookescarlett, line 336: ditto here

.changeset/moody-hounds-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/interaction-ring': major
+---
+
+- Added an `InteractiveRing` component for advanced use cases where one wants to apply standard visual hover and focus states to wrapped components.
```
brookescarlett, line 5: nit: naming consistency

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -340,106 +336,115 @@ function OrgSelect({
...
+            usePortal={false}
+            className={menuContainerStyle}
+            justify="start"
+            spacing={0}
+            setOpen={toggleOpen}
+            open={open}
+          >
+            {data && (
+              <FocusableMenuItem>
+                <Input
```
brookescarlett, line 399: Input element not rendering properly anymore 

Screen Shot 2020-11-17 at 11 26 45 AM

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
```diff
@@ -234,102 +231,109 @@ function ProjectSelect({
...
+        >
           <FocusableMenuItem>
-            <Button
-              href={urls.newProject as string}
-              data-testid="project-select-add-new-project"
-              onClick={onElementClick(NavElement.ProjectNavAddProject)}
-            >
-              + New Project
-            </Button>
+            <Input
```
brookescarlett, line 293: ditto here: 
Screen Shot 2020-11-17 at 11 26 53 AM

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
lazytype: seems to be an existing bug too but I have a fix

packages/mongo-nav/src/mongo-select/ProjectSelect.tsx
lazytype: this also seems to show a bug with rendering Button as an anchor link

brookescarlett:
When a MenuItem is focused in the UserMenu the Inputs are still appearing as focused in the Org/Proj Select, not sure if this is by design:
Screen Shot 2020-11-17 at 5 31 33 PM

brookescarlett:
Also this is happening with RadioBoxGroup:

Screen Shot 2020-11-17 at 5 32 52 PM

brookescarlett (approved):
LGTM 👍

PR #699, opened 2020-11-19
PD-954: Fixes focus issues with Tabs package

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-954](https://jira.mongodb.org/browse/PD-954)

## 🛠 Types of changes

## ✅ Checklist


lazytype:
Before this change I could click on a tab and use arrow keys to switch tabs, but I can't after this change

packages/tabs/src/TabTitle.tsx
```diff
@@ -113,6 +113,23 @@ const listTitle = css`
...
+  const handleFocusIn = () => {
+    setActiveEl(document.activeElement);
+  };
+
+  React.useEffect(() => {
+    document.addEventListener('focusin', handleFocusIn);
+    return () => {
+      document.removeEventListener('focusin', handleFocusIn);
+    };
+  }, []);
```
lazytype, line 128: is this the right way to detect this? what if the event doesn't bubble?

packages/tabs/src/TabTitle.tsx
```diff
@@ -113,6 +113,23 @@ const listTitle = css`
   }
 `;
 
+function useDocumentActiveElement() {
+  const [activeEl, setActiveEl] = React.useState<Element | null>(null);
+
+  const handleFocusIn = () => {
+    setActiveEl(document.activeElement);
+  };
```
lazytype, line 121: `useCallback`

packages/tabs/src/TabTitle.tsx
```diff
@@ -132,17 +150,23 @@ const TabTitle: ExtendableBox<BaseTabProps, 'button'> = ({
   ariaControl,
   index,
   darkMode,
+  containerNode,
```
lazytype, line 153: if this is only used to check whether one of the tabs is focused then we can just pass in something like an `isAnyTabFocused` prop instead.

packages/tabs/src/Tabs.tsx
```diff
@@ -204,17 +207,18 @@ function Tabs({
...
               {...filteredRest}
               key={index}
+              containerNode={containerNode}
               ariaControl={`tab-${index}`}
               disabled={disabled}
               selected={selected}
               index={index}
               as={as}
               darkMode={darkMode}
+              onKeyDown={handleArrowKeyPress}
```
lazytype, line 217: it seems more logical to have this adjacent to the `onClick` prop

brookescarlett:
Hm, I'm not seeing that locally

packages/tabs/src/TabTitle.tsx
```diff
@@ -132,17 +150,23 @@ const TabTitle: ExtendableBox<BaseTabProps, 'button'> = ({
...
   const mode = darkMode ? Mode.Dark : Mode.Light;
+  const activeEl = useDocumentActiveElement();
 
   useEffect(() => {
-    if (!disabled && selected && titleRef.current) {
-      titleRef.current.focus();
+    const tabsList = Array.from(containerNode?.current?.children || []);
+
+    if (activeEl !== null && tabsList.indexOf(activeEl) !== -1) {
+      if (!disabled && selected && titleRef.current) {
```
lazytype, line 165: it's gotten to the point where I think a comment to explain to the when and why of what's happening here would be helpful. I'm not really sure what we're trying to do here but can we achieve what we want by checking if `usePrevious(selected) !== selected`?

lazytype (changes requested):
I see a bunch of warnings coming from the Tabs tests

lazytype:
> Hm, I'm not seeing that locally

Seems to only happen in Firefox

packages/tabs/src/TabTitle.tsx
brookescarlett: 👍

packages/tabs/src/TabTitle.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: Was just grouping multiline props at the bottom

packages/tabs/src/TabTitle.tsx
brookescarlett: What was going on is that we were prematurely (manually) setting a TabTitle as focused, so when a consuming application used Tabs and a user landed on a page, it would automatically scroll down to wherever the tabs were located (because the tabs were automatically focused). This PR adds a check to make sure that the TabTitles should receive focus before manually focusing them.

packages/tabs/src/TabTitle.tsx
brookescarlett: > > Hm, I'm not seeing that locally
> 
> Seems to only happen in Firefox

I'm not seeing it broken in Firefox either

brookescarlett:
> > Hm, I'm not seeing that locally
> 
> Seems to only happen in Firefox

I'm not seeing this in Firefox locally either

brookescarlett:
> I see a bunch of warnings coming from the Tabs tests

Rewrote test suite so this should be resolved now

packages/tabs/src/Tabs.tsx
```diff
@@ -60,6 +60,23 @@ const disabledStyle = css`
   cursor: not-allowed;
 `;
 
+function useDocumentActiveElement() {
+  const [activeEl, setActiveEl] = React.useState<Element | null>(null);
+
+  const handleFocusIn = useCallback(() => {
+    setActiveEl(document.activeElement);
+  }, [setActiveEl]);
```
lazytype, line 68: I don't think this needs `setActiveEl` as it should be stable between renders

packages/tabs/src/Tabs.tsx
```diff
@@ -60,6 +60,23 @@ const disabledStyle = css`
   cursor: not-allowed;
 `;
 
+function useDocumentActiveElement() {
+  const [activeEl, setActiveEl] = React.useState<Element | null>(null);
```
lazytype, line 64:
```suggestion
  const [activeEl, setActiveEl] = useState<Element | null>(null);
```

packages/tabs/src/Tabs.tsx
```diff
@@ -121,6 +138,18 @@ function Tabs({
   as = 'button',
   ...rest
 }: TabsProps) {
+  const containerNode = useRef<HTMLDivElement | null>(null);
+  const activeEl = useDocumentActiveElement();
+  const [isAnyTabFocused, setIsAnyTabFocused] = useState(false);
+
+  useEffect(() => {
+    const tabsList = Array.from(containerNode?.current?.children || []);
```
lazytype, line 146:
```suggestion
    const tabsList = Array.from(containerNode.current?.children ?? []);
```

lazytype (approved):

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

PR #700, opened 2020-11-19
PD-852: Fix glyph flexbox issue

DesignerDave:
## ✍️ Proposed changes

Fixes an issue where glyphs would shrink in flex containers when consumed individually.

🎟 _Jira ticket:_ [PD-852](https://jira.mongodb.org/browse/PD-852)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/icon/src/template.ts
+++ b/packages/icon/src/template.ts
@@ -32,7 +32,7 @@ module.exports = function template(
         {
           [fillStyle]: fill != null,
         },
-        noFlexSizing,
+        noFlexShrink,
         className,
       )}
       height={typeof size === 'number' ? size : sizeMap[size]}
@@ -74,9 +74,8 @@ module.exports = function template(
         color: \${fill};
       \`;
 
-      const noFlexSizing = css\`
+      const noFlexShrink = css\`
         flex-shrink: 0;
-        flex-grow: 0;
       \`;
 
       title = getGlyphTitle('${componentName}', title);
--- a/packages/icon/src/template.ts
+++ b/packages/icon/src/template.ts
@@ -32,6 +32,7 @@ module.exports = function template(
         {
           [fillStyle]: fill != null,
         },
+        noFlexSizing,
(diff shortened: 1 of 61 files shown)


packages/icon/src/template.ts
```diff
@@ -73,6 +74,11 @@ module.exports = function template(
         color: \${fill};
       \`;
 
+      const noFlexSizing = css\`
```
brookescarlett, line 77: Why are we setting flex-grow to 0? IIRC, don't we find ourselves adding that property to maintain the icon's size

DesignerDave: You're right, this should be the implicit value already. By being explicit here, I hoped to make it easier to debug issues around this related to setting that value to something other than `0`.

If you feel strongly, I can remove this.

brookescarlett: I thought we had discussions around setting this to 1 explicitly rather than setting to 0 explicitly.

DesignerDave: Fair point, I'll remove.

lazytype:
I think you need to run `yarn pre-build` again

DesignerDave:
> I think you need to run `yarn pre-build` again

Will do. In general, how can we adjust our actions to fix this in the future?

lazytype:
> > I think you need to run `yarn pre-build` again
> 
> Will do. In general, how can we adjust our actions to fix this in the future?

I think the simplest approach is to have the action run the pre-build command and use git to check no files have changed

DesignerDave:
> I think the simplest approach is to have the action run the pre-build command and use git to check no files have changed. This can also be used to check when `yarn.lock` needs to be updated too.

Sounds good! Would you mind making a ticket for that with the details you have? I don't think it should be a priority now, but we should consider doing that for our quality of life 😅

brookescarlett (approved):
lgtm

PR #709, opened 2020-12-01
CLOUDP-76088: Disable Realm and Charts icons in MongoNav

JamesHeppenstall:
## ✍️ Proposed changes

As part of [this epic](https://jira.mongodb.org/browse/CLOUDP-55272), we want to disable the Realm and Charts icons in `MongoNav` when an Atlas project is designated as an China regions-only project. This WIP PR adds the necessary logic to achieve this. See the associated PR for [CLOUDP-76088](https://github.com/10gen/mms/pull/35481) for the changes to the endpoint `/user/shared` from which `MongoNav` gets information about the current project.

🎟 _Jira ticket:_ [CLOUDP-76088](https://jira.mongodb.org/browse/CLOUDP-76088)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

Here's what the icons and tooltip look like when the current project is an China regions-only project:
Screen Shot 2020-12-01 at 1 09 46 PM


lazytype:
From testing this locally I noticed a weird behavior. Clicking one of the tabs briefly displays the cursor as `cursor: pointer` and causes the tooltip disappear and reappear

lazytype:
Pressing the Enter key when one of the tabs is focused still navigates to a new page.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -296,6 +322,11 @@ export default function ProjectNav({
   const isCloudManager = current?.planType === PlanType.Cloud;
   const isLoading = !!current;
   const currentProjectId = current?.projectId;
+  const useCNRegionsOnly =
```
lazytype, line 325: could we do something like the following to avoid the duplicated product logic below
```ts
const useCNRegionsOnly = current?.useCNRegionsOnly ?? false;
const planType = current?.planType;

const isCNRegionsOnly = useCallback(
  (product: Product) => {
    const disabledProducts: Array<Product> = [Product.Realm, Product.Charts];

    return (
      useCNRegionsOnly &&
      planType === PlanType.Atlas &&
      disabledProducts.includes(product)
    );
  },
  [useCNRegionsOnly, planType],
);
```

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -442,43 +477,69 @@ export default function ProjectNav({
...
 
             <li role="none" className={productTabStyle}>
-              <a
-                data-testid="project-nav-charts"
-                href={projectNav.charts}
-                className={getProductClassName('charts')}
-                aria-disabled={!current}
-                tabIndex={current ? 0 : -1}
-                onClick={onElementClick(ProjectNavCharts)}
+              <Tooltip
```
lazytype, line 513: Can we pull out the duplicated tooltip logic into a component? e.g.
```tsx
  const WithCNRegionsOnlyTooltip = useCallback(
    ({ children }: { children: React.ReactElement }) => {
      return (
        <Tooltip
          {...sharedTooltipProps}
          tabIndex={-1}
          align="bottom"
          justify="middle"
          usePortal={false}
          darkMode={true}
          enabled={useCNRegionsOnly}
          trigger={children}
        >
          {useCNRegionsOnlyMessage}
        </Tooltip>
      );
    },
    [sharedTooltipProps, useCNRegionsOnly],
  );

  ...

  <WithCNRegionsOnlyTooltip>
    <a
      data-testid="project-nav-charts"
      href={useCNRegionsOnly ? hosts.cloud : projectNav.charts}
      className={getProductClassName('charts')}
      aria-disabled={!current}
      tabIndex={current ? 0 : -1}
      onClick={onElementClick(ProjectNavCharts)}
    >
      {!isMobile && (
        <ChartsIcon
          {...productIconProp.prop}
          className={getIconStyle(Product.Charts)}
          active={activeProduct === Product.Charts && isLoading}
        />
      )}
      Charts
    </a>
  </WithCNRegionsOnlyTooltip>
```

packages/mongo-nav/src/types.ts
```diff
@@ -203,6 +203,7 @@ export interface CurrentProjectInterface extends ProjectInterface {
   alertsOpen: number;
   chartsActivated: boolean;
   status?: ProjectStatus;
+  useCNRegionsOnly: boolean;
```
lazytype, line 206: would `isCNRegionsOnly` make more sense?

lazytype (changes requested):

packages/mongo-nav/src/project-nav/ProjectNav.tsx
JamesHeppenstall: 👍

packages/mongo-nav/src/project-nav/ProjectNav.tsx
JamesHeppenstall: 👍

packages/mongo-nav/src/types.ts
JamesHeppenstall: I agree that `isCNRegionsOnly` might make more sense but the field name on the project in the Atlas backend is called `useCNRegionsOnly` so I wanted to keep things consistent.

JamesHeppenstall:
@lazytype thanks for the initial review. I addressed all your comments and also decided to only include the anchor tag if `useCNRegionsOnly` is false, which prevents some of the weird behavior you saw. Let me know what you think!

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -442,43 +524,51 @@ export default function ProjectNav({
...
-              </a>
+              {useCNRegionsOnly && (
+                <WithCNRegionsOnlyTooltip>
+                  <div className={getProductClassName(Product.Realm)}>
+                    {getRealmIcon()}
+                    Realm
+                  </div>
+                </WithCNRegionsOnlyTooltip>
+              )}
+              {!useCNRegionsOnly && (
```
lazytype, line 535: hm, is there a way we can get both branches to be defined in `WithCNRegionsOnlyTooltip` ?

packages/mongo-nav/src/project-nav/ProjectNav.tsx
```diff
@@ -247,6 +268,13 @@ type ProjectNavProps = Pick<
   | 'environment'
   | 'alertPollingInterval'
 > & {
+  admin?: boolean;
+  activeProduct?: Product;
+  activeNav?: ActiveNavElement;
+  onProjectChange?: (onChangeInterface: OnChangeInterface) => void;
+  mode?: Mode;
+  alertPollingInterval?: number;
+  environment?: Environment;
```
lazytype, line 277: curious about why this change is needed. does the `Pick` above not already include these for us?

lazytype (commented):
could we also add a couple of test cases to cover the new behaviors:
- clicking on the tabs when `useCNRegionsOnly` is `true` doesn't have any effect, e.g. `expect(getByText('Realm').closest('a')).toBeNull()`
- hovering over the tabs makes the tooltip visible

packages/mongo-nav/src/project-nav/ProjectNav.tsx
JamesHeppenstall: It doesn't seem to render properly when I put both branches in `WithCNRegionsOnlyTooltip`. Also, just wondering why we want to render the tooltip if `useCNRegionsOnly == false`?

packages/mongo-nav/src/project-nav/ProjectNav.tsx
JamesHeppenstall: I thought the same thing but `yarn fix` started throwing errors about those props not being defined.

packages/mongo-nav/src/project-nav/ProjectNav.tsx
lazytype: I was thinking the semantics of it would be more like `WithTooltipIfCNRegionsOnly` and handle the branching so we don't need to repeat it for Realm and Charts

packages/mongo-nav/src/project-nav/ProjectNav.tsx
lazytype: ah, seems like a linter false-positive. you can wrap those lines in
```
/* eslint-disable react/prop-types */
...
/* eslint-enable react/prop-types */
```

packages/mongo-nav/src/project-nav/ProjectNav.tsx
JamesHeppenstall: Ah got it 😄 I'll update the PR to handle this

JamesHeppenstall:
@lazytype this is RFAL 😄

lazytype (approved):
looks great!

PR #719, opened 2020-12-08
Polishing website

brookescarlett:
- Makes interaction ring cleaner on Palette live example 
- Pushes down components in side nav
- Removes guidelines from list of things to iterate from to create side nav as we don't have material for that yet

### DIFF
--- a/website/components/navigation/Navigation.tsx
+++ b/website/components/navigation/Navigation.tsx
@@ -138,7 +138,7 @@ function Content({ isTouchDevice = false }: { isTouchDevice?: boolean }) {
     );
   };
 
-  return <>{[GroupType.Guideline, GroupType.Component].map(renderGroup)}</>;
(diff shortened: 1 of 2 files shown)


website/components/navigation/Navigation.tsx
```diff
@@ -138,7 +138,7 @@ function Content({ isTouchDevice = false }: { isTouchDevice?: boolean }) {
     );
   };
 
-  return <>{[GroupType.Guideline, GroupType.Component].map(renderGroup)}</>;
+  return <>{[GroupType.Component].map(renderGroup)}</>;
```
lazytype, line 141: :)
```suggestion
  return renderGroup(GroupType.Component);
```

brookescarlett: 👍

lazytype (approved):

PR #720, opened 2020-12-08
PD-975: Adds as prop to h1, h2, h3 and subtitle

brookescarlett:
`H1`, `H2`, `H3` and `Subtitle` components now accept an `as` prop, such that we can keep styles consistent via `Component`, but the actual heading level that should be rendered can change based on context. This was done to support making MongoDB an accessible platform, as headings should only decrease by one level (i.e. from `<h1>` to `<h2>`) but the styles don't always need to appear as such.

### DIFF
--- a/packages/typography/src/Typography.tsx
+++ b/packages/typography/src/Typography.tsx
@@ -25,13 +25,9 @@ const h1 = css`
 
 type H1Props = HTMLElementProps<'h1'>;
 
-function H1({ children, className, ...rest }: H1Props) {
-  return (
-    <h1 {...rest} className={cx(sharedStyles, h1, className)}>
-      {children}
-    </h1>
-  );
-}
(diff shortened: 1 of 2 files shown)


packages/typography/src/Typography.tsx
```diff
@@ -25,13 +25,14 @@ const h1 = css`
...
 type H1Props = HTMLElementProps<'h1'>;
 
-function H1({ children, className, ...rest }: H1Props) {
-  return (
-    <h1 {...rest} className={cx(sharedStyles, h1, className)}>
-      {children}
-    </h1>
-  );
-}
+const H1: ExtendableBox<H1Props> = ({
```
lazytype, line 28: we should pass in the default of "h1". same for other components

brookescarlett: Yep, good call

lazytype (approved):

PR #721, opened 2020-12-09
PD-943: Update UserMenu for Account MFA

kevintian:
## ✍️ Proposed changes

We're officially deprecating 2FA in Cloud in favor of the recently released Account MFA. To show this, we are a) advertising the new feature in the UserMenu and b) renaming and hiding the "Legacy 2FA" menu item for users that do not have it set up, since they will no longer be able to enroll in this functionality.

🎟 _Jira ticket:_ [PD-943](https://jira.mongodb.org/browse/PD-943)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments
Invision Link for changes: https://mongodb.invisionapp.com/share/H8XIP6KRVFM#/screens/422003152
Screenshot:

image

### DIFF
--- a/.changeset/metal-cups-switch.md
+++ b/.changeset/metal-cups-switch.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/mongo-nav': minor
+---
+
+Adds banner to advertise Account MFA. Renames Two-Factor Authentication -> Legacy 2FA and removes this menu item for users that do not already have it set up.
--- a/packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
+++ b/packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
@@ -9,16 +9,17 @@ import { JestDOM } from '@leafygreen-ui/testing-lib';
 import { urlFixtures, hostDefaults } from '../data';
 import UserMenu from '.';
 
-const account = {
+const defaultAccount = {
   firstName: 'Leafy',
   lastName: 'Green',
   email: 'leafy@mongodb.com',
+  hasLegacy2fa: true,
   admin: false,
 };
 const onLogout = jest.fn();
 const onProductChange = jest.fn();
 
-function renderUserMenu(props = {}) {
+function renderUserMenu(props = {}, account = defaultAccount) {
   const utils = render(
     <UserMenu
       account={account}
@@ -40,7 +41,7 @@ describe('packages/mongo-nav/user-menu', () => {
     expect(trigger).toBeInTheDocument();
   });
 
-  test('when "activePlatform" is set to "cloud, renders Cloud MenuItems', () => {
+  test('when "activePlatform" is set to "cloud", renders Cloud MenuItems', () => {
     const { getByTestId, getByText } = renderUserMenu({
       activePlatform: 'cloud',
     });
@@ -50,7 +51,7 @@ describe('packages/mongo-nav/user-menu', () => {
     const userPreferences = getByText('User Preferences');
     const invitations = getByText('Invitations');
     const organizations = getByText('Organizations');
-    const mfa = getByText('Two-Factor Authentication');
+    const mfa = getByText('Legacy 2FA');
 
     expect(userPreferences).toBeInTheDocument();
     expect(invitations).toBeInTheDocument();
@@ -78,6 +79,22 @@ describe('packages/mongo-nav/user-menu', () => {
     expect(universityItem).toBeVisible();
   });
 
+  test('when "hasLegacy2fa" is set to false, does not render mfa menu item', () => {
+    const { queryByText, getByTestId } = renderUserMenu({
+      activePlatform: 'cloud',
+      account: {
+        firstName: 'Leafy',
+        lastName: 'Green',
+        email: 'leafy@mongodb.com',
+        hasLegacy2fa: false,
+        admin: false,
+      },
+    });
+
+    const trigger = getByTestId('user-menu-trigger');
+    fireEvent.click(trigger);
+    expect(queryByText('Legacy 2FA')).not.toBeInTheDocument();
+  });
   test('atlas MenuItem links to cloud.mongodb.com', () => {
     const { getByTestId } = renderUserMenu({
       activePlatform: 'cloud',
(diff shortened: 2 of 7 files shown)


packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -442,7 +457,6 @@ function UserMenu({
             {isGovernment ? 'Cloud for Government' : 'Cloud'}
           </MenuItem>
         )}
-
```
kevintian, line 445: Looks like my Prettier auto-removed these extra spaces, but I also didn't see any issue/changes when running `yarn prettier:fix` so let me know if I need to change this back.

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
```diff
@@ -78,6 +79,22 @@ describe('packages/mongo-nav/user-menu', () => {
     expect(universityItem).toBeVisible();
   });
 
+  test('when "hasLegacy2fa" is set to false, does not render mfa menu item', () => {
```
brookescarlett, line 82: Can we add a test to make sure we're correctly showing/hiding the banner as well please

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -335,6 +346,10 @@ function UserMenu({
         )}
         usePortal={false}
       >
+        <li role="none" className={mfaBannerStyle}>
```
brookescarlett, line 349: Should this be conditionally rendered based on whether or not they've already signed up for MFA

packages/mongo-nav/src/user-menu/UserMenu.tsx
```diff
@@ -413,15 +428,15 @@ function UserMenu({
             >
               Organizations
             </MenuItem>
-            {!isGovernment && (
+            {!isGovernment && !!account?.hasLegacy2fa && (
```
brookescarlett, line 431: should this need a `!!` if hasLegacy2fa is returning a boolean?

kevintian: Nope, it doesn't - I noticed that there were places where we did `!!admin` so I wasn't sure if it was a convention that we wanted to force the value to be boolean instead of relying on truthy/falsey. I can remove this.

packages/mongo-nav/src/user-menu/UserMenu.tsx
kevintian: So this was actually changed from the original design because we realized that we wouldn't be able to propagate that boolean value easily to the rest of the services that rely on UserMenu. Instead, we'll temporarily show the banner for a set amount of time (~30 days or so) and then remove it afterwards.

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
kevintian: See comment below about the banner not being a conditional render - I can add a test to ensure the banner is always rendered though

.changeset/metal-cups-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/mongo-nav': minor
```
brookescarlett, line 2: I think we should make this a major version bump as we're changing the UserMenu across products, what do you think cc: @DesignerDave

brookescarlett:
Last thing! Would you mind adding a knob to our storybook to document this change? You can do this by adding a knob inside of the `dataFixtures` prop  in MongoNav.story.tsx

packages/mongo-nav/src/user-menu/UserMenu.spec.tsx
brookescarlett: Thank you!

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: Sounds good, TY for the clarity

packages/mongo-nav/src/user-menu/UserMenu.tsx
brookescarlett: 👍

.changeset/metal-cups-switch.md
DesignerDave: I don't feel strongly either way. This isn't breaking from an API standpoint, so the argument would be simply that the design change is large enough to warrant it, which I don't feel strongly that it is or isn't. Your call 😄

packages/mongo-nav/src/user-menu/UserMenu.tsx
DesignerDave: @kevintian Is there a reason we don't move the logic to the API backing the navigation? It would propagate across other services if we made that change.

packages/mongo-nav/src/user-menu/UserMenu.tsx
DesignerDave: I think that coercion is for cases where it's a boolean or a nullish value in those cases. If `hasLegacy2fa` is always a boolean, that's not an issue.

packages/mongo-nav/src/user-menu/UserMenu.tsx
kevintian: That's a good point - so my one concern is that while Stitch, Charts, and Cloud use the UserMenu component and call the API backing the navigation (and thus would be updated properly), University uses a fork of UserMenu and would have to update their component each time regardless.

I'm fine with moving the logic to the API backing the navigation to make things easier for Stitch/Charts/Cloud if desired!

brookescarlett: Deferring to you here @DesignerDave

DesignerDave: I'd vastly prefer we move this from a prop to the API. University has the option of using the user menu we export as well, but since they don't, how we approach this here is unrelated to their specific case.

.changeset/metal-cups-switch.md
brookescarlett: I feel better about this being a minor version, now that the determination re: whether to render the banner/or not is coming from the server rather than from us imposing it on every product

brookescarlett (approved):

PR #723, opened 2020-12-09
CLOUDP-78527: Add missing data-testid attributes to OrgNav

JamesHeppenstall:
A regression in the `OrgNav` component causes the `data-testid` attribute to be missing from the `All Clusters` link and the `Admin` link. This is resulting in test failures for anything that was looking up these `data-testid` attributes.

🎟 _Jira ticket:_ [CLOUDP-78527](https://jira.mongodb.org/browse/CLOUDP-78527)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/.changeset/hip-impalas-live.md
+++ b/.changeset/hip-impalas-live.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/mongo-nav': patch
+---
+
+Adds missing data-testid attributes to All Clusters and Admin links in OrgNav
--- a/packages/mongo-nav/src/org-nav/OrgNav.tsx
+++ b/packages/mongo-nav/src/org-nav/OrgNav.tsx
@@ -165,12 +165,14 @@ function NavLinks({
     href?: string;
     navId: NavElement;
     text: string;
+    testid: string;
   }> = [];
 
   navLinks.push({
     href: urls.allClusters,
     navId: ActiveNavElement.OrgNavAllClusters,
     text: 'All Clusters',
+    testid: "org-nav-all-clusters-link",
   });
 
   if (admin) {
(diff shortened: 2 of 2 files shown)


lazytype:
Hi there, thanks for surfacing the issue. I would not recommend relying on `data-testid` attributes to test integrations with the mongo-nav component. Since they're not part of the "Public API" of the package they're subject to change even with minor and patch releases. My recommendation would be to query for these elements by role (if available) or text. react-testing-library is great for this use case. The commit that removed the `data-testid` attributes also updated our internal tests to use this convention. Please take a look at [the change](https://github.com/mongodb/leafygreen-ui/commit/e0e6076b1ecdccc04ae237758da8964da579dc81#diff-91f26cf26078eefeb161a82ca69e5c2f4c1b0db7e7997368663d9b27ba99670c) for reference and feel free to ask for any guidance

JamesHeppenstall:
> Hi there, thanks for surfacing the issue. I would not recommend relying on `data-testid` attributes to test integrations with the mongo-nav component. Since they're not part of the "Public API" of the package they're subject to change even with minor and patch releases. My recommendation would be to query for these elements by role (if available) or text. react-testing-library is great for this use case. The commit that removed the `data-testid` attributes also updated our internal tests to use this convention. Please take a look at [the change](https://github.com/mongodb/leafygreen-ui/commit/e0e6076b1ecdccc04ae237758da8964da579dc81#diff-91f26cf26078eefeb161a82ca69e5c2f4c1b0db7e7997368663d9b27ba99670c) for reference and feel free to ask for any guidance

Thanks for bringing this up. I'm going to raise this in the `cloud-web` slack channel to see what people think 😄

DesignerDave:
@lazytype For the time being, can we get this PR in? I think the approach to E2E test selectors, and test selectors outside of React Testing Library needs to be addressed more holistically.

lazytype (approved):

PR #726, opened 2020-12-10
Collapsible Side Nav

lazytype:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Collapsible Side Nav](https://jira.mongodb.org/browse/PD-935)

## 🛠 Types of changes

## ✅ Checklist


packages/collapsible-side-nav/README.md
```diff
@@ -0,0 +1,464 @@
...
+          >
+            <title id="Settings-21">Settings Icon</title>
+            <g
+              id="Settings-Copy"
+              stroke="none"
+              stroke-width="1"
+              fill="none"
+              fill-rule="evenodd"
+            >
+              <path
```
brookescarlett, line 332: can omit the path details here

packages/collapsible-side-nav/README.md
```diff
@@ -0,0 +1,464 @@
...
+          >
+            <title id="Support-11">Support Icon</title>
+            <g
+              id="Support-Copy"
+              stroke="none"
+              stroke-width="1"
+              fill="none"
+              fill-rule="evenodd"
+            >
+              <path
```
brookescarlett, line 407: ditto

packages/collapsible-side-nav/src/SideNav.tsx
```diff
@@ -0,0 +1,204 @@
...
+  // from every sub-component or an equivalent amount of work.
+  const [hovered, setHovered] = useState(false);
+
+  // Debounce hover changes to prevent jitter when rapidly moving the mouse
+  // on and off the nav. Alternatively we could set transition-delay for this
+  // and all sub-components, but that would be less convenient and wouldn't
+  // prevent unnecessary re-renders.
+  const debouncedSetHovered = useMemo(
+    () => ({
+      toTrue: debounce(() => setHovered(true), transitionDurationMilliseconds),
```
brookescarlett, line 94: The transition in still feels a little jittery to me, while the transition out does not

packages/collapsible-side-nav/src/SideNav.tsx
```diff
@@ -0,0 +1,204 @@
...
+    debouncedSetHovered.toTrue.cancel();
+    setHovered(false);
+
+    setCollapsed(collapsed => !collapsed);
+  }, [debouncedSetHovered]);
+
+  useEventListener(
+    'keydown',
+    event => {
+      if (event.keyCode === 219 /* BracketLeft */) {
```
brookescarlett, line 132: Lets add to our keyMap in lib (it seems that keyCode is starting to be deprecated anyway, so when we change how we approach key-detection we'll want to make sure we do this everywhere at once)

packages/collapsible-side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,176 @@
...
+  display: flex;
+  align-items: center;
+  margin-bottom: 4px;
+  padding: 0 16px;
+  text-transform: uppercase;
+  white-space: nowrap;
+  font-size: 11px;
+  font-weight: bold;
+  line-height: 16px;
+  color: ${uiColors.green.dark2};
```
brookescarlett, line 43: Should handle truncating too-long group names here as well

packages/collapsible-side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,176 @@
...
+    const id = useMemo(() => idAllocator.generate(), []);
+
+    const [containedPaths, setContainedPaths] = useState<Set<string>>(
+      () => new Set(),
+    );
+
+    const addPath = useCallback((path: string) => {
+      setContainedPaths(containedPaths => {
+        if (!containedPaths.has(path)) {
+          const updatedPaths = new Set(containedPaths);
```
brookescarlett, line 93: What is the benefit of making a new set each time? If we've checked that the existing set does not have a path, why can't we just add it to the containedPaths set?

packages/collapsible-side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,176 @@
...
+        } else {
+          return containedPaths;
+        }
+      });
+    }, []);
+
+    const removePath = useCallback((path: string) => {
+      setContainedPaths(containedPaths => {
+        if (containedPaths.has(path)) {
+          const updatedPaths = new Set(containedPaths);
```
brookescarlett, line 104: ditto here?

packages/collapsible-side-nav/src/SideNavGroup.tsx
```diff
@@ -0,0 +1,176 @@
...
+        }
+      });
+    }, []);
+
+    const providerData = useMemo(() => ({ addPath, removePath }), [
+      addPath,
+      removePath,
+    ]);
+
+    const showCollapsed = collapsed && !hovered;
```
brookescarlett, line 118: We recreate this variable everywhere, would it make sense to pass this to the provider as well?

packages/collapsible-side-nav/src/SideNavItem.tsx
```diff
@@ -0,0 +1,213 @@
...
+        break;
+      default:
+        enforceExhaustive(glyphVisibility);
+    }
+  } else {
+    showGlyph = false;
+  }
+
+  return (
+    <a
```
brookescarlett, line 169: should we take an `as` prop here to support more types of navigation?

packages/collapsible-side-nav/src/SideNavGroup.tsx
brookescarlett: Another question: My understanding was that the purpose of using Sets was to enforce a uniqueness of paths, such that we are able to determine an active Item based on whatever path is passed to the entire Navigation. If that assumption is correct (and perhaps I'm misunderstanding) I have two questions: 

1. Right now, it seems that we're only enforcing this uniqueness within a Group, so duplicate paths can exist outside of the same group
2. Is there a world in which we'd want to support a group that has items that lead to the same page. If that's the case, does it make more sense for Items to accept an active prop rather than have the Navigation component being responsible for determining what's active?

brookescarlett: Another argument for perhaps supporting the active prop: In .design (for example) we have three routes that map back to the same Item being active.  I need `/component/checkbox/example` and `/component/checkbox/guideline` both to make the Checkbox Item appear active in the SideNav.

lazytype: > why can't we just add it to the containedPaths set

this would not trigger a rerender since the updated `containedPaths` would equal the previous one.

> My understanding was that the purpose of using Sets was to enforce a uniqueness of paths

The purpose of using sets here is to be able to quickly look up whether an item exists. We could use plain objects for this too, but the Set API is better-suited for this.

> we have three routes that map back to the same Item being active

I don't see how an `active` prop would add additional value. the existing logic seems to look like
```jsx
<SideNavItem active={item === activePage} ...
```
which would translate to
```jsx
<SideNav currentPath={activePage} ... >
  <SideNavItem path={item} ...

Comment (brookescarlett) [reply packages/collapsible-side-nav/src/SideNavGroup.tsx:93]: In my case for example I'd like to do: 

```
<SideNav>
   <SideNavItem active={path.includes(componentName)}>{componentName}</SideNavItem>
</SideNav>
```

File: packages/collapsible-side-nav/src/SideNavItem.tsx (hunk 9 above)
Comment (lazytype) [reply packages/collapsible-side-nav/src/SideNavItem.tsx:215]: I would want to see a use-case for it. Accepting `as` and using `Box` usually makes things more difficult on the development side.

File: packages/collapsible-side-nav/src/SideNavGroup.tsx (hunk 6 above)
Comment (brookescarlett) [reply packages/collapsible-side-nav/src/SideNavGroup.tsx:93]: (thank you for the other clarification points)

File: packages/collapsible-side-nav/src/SideNavItem.tsx (hunk 9 above)
Comment (brookescarlett) [reply packages/collapsible-side-nav/src/SideNavItem.tsx:215]: using `Link` for example

File: packages/collapsible-side-nav/src/SideNavGroup.tsx (hunk 6 above)
Comment (lazytype) [reply packages/collapsible-side-nav/src/SideNavGroup.tsx:93]: I think the same can be accomplished by accepting a matching function for `currentPath`, but having something like `path.includes(componentName)` seems more unsound, e.g. `'/components/checkbox'.includes('box')`, compared to normalizing the path, e.g. `normalize('/component/checkbox/example') === '/component/checkbox'`

File: packages/collapsible-side-nav/src/SideNavGroup.tsx (hunk 8 above)
Comment (lazytype) [reply packages/collapsible-side-nav/src/SideNavGroup.tsx:118]: I think the naming is not very clear and would be more clear defined locally since one can look up the definition nearby. I also just prefer not to store redundant data.

File: packages/collapsible-side-nav/src/SideNavGroup.tsx (hunk 6 above)
Comment (brookescarlett) [reply packages/collapsible-side-nav/src/SideNavGroup.tsx:93]: Accepting a matching function would wfm:
```
<SideNav currentPath={(itemPath) => normalizePath(currentPath) === normalizePath(itemPath}

<SideNavItem path="/component/componentName/example"> 
}
```

Comment (lazytype) [reply packages/collapsible-side-nav/src/SideNavGroup.tsx:93]: If you have `normalizePath` then you wouldn't need a matching function
```
<SideNav currentPath={normalizePath(currentPath)} ...
  <SideNavItem path="/component/checkbox" ...

packages/collapsible-side-nav/src/SideNavItem.tsx
lazytype: Fair enough 👍

packages/collapsible-side-nav/src/SideNavGroup.tsx
brookescarlett: Makes sense

brookescarlett:
just spoke to @DesignerDave, and it seems that this sidenav should replace our existing, so we don't need a new package here

packages/icon/src/glyphs/Apps.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
brookescarlett, line 1: Nit: we should add a changeset for Icon additions

packages/side-nav/UPGRADE.md
```diff
@@ -0,0 +1,13 @@
+# v5 to #v6
```
brookescarlett, line 1: nit rm space

packages/side-nav/src/CollapseButton.tsx
```diff
@@ -0,0 +1,152 @@
...
+        [navCollapsedHoveredButtonStyle]: collapsed && hovered,
+      })}
+      aria-label={description}
+      aria-controls={navId}
+      aria-expanded={!shouldRenderCollapsedState}
+      onClick={onTrigger}
+      onKeyDown={onKeyDown}
+      onMouseOver={onMouseOver}
+    >
+      <ChevronLeftIcon
```
brookescarlett, line 117: Seems like there's an accessibility issue with these Chevrons. See the accessibility tab on Storybook for more info

packages/side-nav/src/SideNav.tsx
```diff
@@ -1,53 +1,190 @@
...
+) {
+  const id = useMemo(() => idAllocator.generate(), []);
+
+  // eslint-disable-next-line prefer-const
+  let [collapsed, setCollapsed] = useState(
+    collapsible ? initialCollapsed : false,
+  );
+
+  if (!collapsible) {
+    collapsed = false;
```
brookescarlett, line 95: Why not setCollapsed here

brookescarlett: (Why do we need this outside of an inside of the useEffect hook)

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -1,255 +1,335 @@
...
+    // eslint-disable-next-line prefer-const
+    let [collapsed, setCollapsed] = useState(
+      collapsible ? initialCollapsed : false,
+    );
+
+    if (!collapsible) {
+      collapsed = false;
+    }
+
+    useEffect(() => {
```
brookescarlett, line 163: ditto here

website/pages/component/side-nav/example.tsx
```diff
@@ -56,19 +69,39 @@ function DefaultExample({
...
+        width: ${sideNavWidth}px;
+      `}
+    >
+      <SideNav
+        {...collapsibleProps}
+        currentPath={active ? 'Projects' : undefined}
+      >
+        <SideNavGroup {...collapsibleGroupProps} label={label}>
+          <SideNavItem>Projects</SideNavItem>
+          <SideNavItem disabled={disabled}>Alerts</SideNavItem>
```
brookescarlett, line 95: Not seeing the active and disabled props work here

packages/side-nav/src/SideNav.tsx
lazytype: this is so that the following code we can just write `collapsed` instead of `collapsible && collapsed` everywhere and render the correct thing on this render (and not the next render that would be triggered by the useEffect hook)

packages/side-nav/src/SideNav.spec.tsx
```diff
@@ -1,333 +1,479 @@
...
-type renderedElement = HTMLElement | null;
-
-interface RenderedElements {
-  navEl?: renderedElement;
-  groupEl?: renderedElement;
-  headerContentEl?: renderedElement;
-  defaultHeaderEl?: renderedElement;
-  itemEl?: renderedElement;
-  childEl?: renderedElement;
+import {
```
lazytype, line 2: working on writing tests for collapsing groups

packages/side-nav/src/SideNav.tsx
brookescarlett: I might be missing something but doesn't the ternary above read if it's collapsible then use the initialCollapsed prop and if it's not collapsible set it to false.

brookescarlett: we're branching on the same variable in both places

lazytype: the argument to `useState` only applies to the initial render

brookescarlett (approved):
Have been trying to poke holes in this for a bit and I can't find anything else. I'd wait on @DesignerDave's feedback before merging, but I don't have any blocking feedback. 

(The animation looks great, btw.)

packages/side-nav/README.md
```diff
@@ -18,174 +18,367 @@ yarn add @leafygreen-ui/side-nav
 npm install @leafygreen-ui/side-nav
 ```
 
-## Peer Dependencies
-
-| Package                              | Version  |
-| ------------------------------------ | -------- |
-| `@leafygreen-ui/leafygreen-provider` | `^1.1.0` |
```
DesignerDave, line 25: Is this no longer a peer dependency?

DesignerDave (commented):
Largely just looked at the end result for this – I'm sure you and Brooke have shaved enough yaks around this 😄 

The animation overall is looking absolutely great. I noticed a few things that might make it feel slightly more polished around the edges.

Few things I noticed:
- There's a noticeable area of the expand / collapse button which doesn't show cursor: pointer.
 - It looks like the underline for sections doesn't transition between states like hover.
 - The animation for opening a group is smooth, but the close is less so. It looks like we may be hiding the items within that group via opacity as the transition begins, and it looks a bit jittery.
 -  When transitioning the side navigation from closed to open, the group expand icons (the chevrons) appear with no transition directly on top of the section icons for a brief moment. Transitioning in the expand icons would make the transition feel vastly smoother.
 - More of a note for @audroue, but the text "Expand sidebar ([)" is showing a keyboard shortcut, but it looks an awful lot like a bug, rather than help text.
 - More of a question for @brookescarlett, but when opening the sidebar with the keyboard shortcut, shouldn't the sidebar become focused?

packages/side-nav/README.md
lazytype: I didn't even notice we had this. Is it worth having? A lot of packages will transitively require it too because they use the Button, Interaction-ring, useUsingKeyboardContext, etc. I definitely failed to add this section to the Select and Copyable readmes

DesignerDave: It's _probably_ something we should keep around, but missing it in a component or two isn't going to make a huge impact since it's likely everywhere already has this component installed.

@brookescarlett I don't feel strongly about this. You wanna make this call?

brookescarlett: I think we tried to say that it's a peer dep of every component in the repo. I don't feel strongly one way or another, but do think that we should be consistent.

DesignerDave:
Closing in favor of https://github.com/mongodb/leafygreen-ui/pull/837

PR #730, opened 2020-12-10
Make Popover/Tooltip/Menu compatible with current buttons

lazytype:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/tooltip/src/Tooltip.tsx
+++ b/packages/tooltip/src/Tooltip.tsx
@@ -278,7 +278,7 @@ function Tooltip({
 
       if (triggerProps && triggerProps.onClick) {
         return {
-          onClick: (e: MouseEvent) => {
+          onClick: (e: React.MouseEvent) => {
             // ensure that we don't close the tooltip when content inside tooltip is clicked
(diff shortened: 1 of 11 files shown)


packages/tooltip/src/Tooltip.tsx
```diff
@@ -380,17 +380,14 @@ function Tooltip({
...
-      children: (
-        <>
-          {trigger.props.children}
-          {tooltip}
-        </>
-      ),
-      className: cx(positionRelative, trigger.props.className),
-    });
+    return (
+      <div className={positionRelative} {...createTriggerProps(triggerEvent)}>
```
DesignerDave, line 384: In the past, we purposefully didn't do this, and instead made sure that the button DOM structure supported Tooltips. I'm still of the mind that we shouldn't wrap the trigger element.

lazytype: Looks like you reviewed this PR after I closed it. Please see https://github.com/mongodb/leafygreen-ui/pull/732 instead :)

PR #732, opened 2020-12-14
Make Button work better with Popover and Tooltip

lazytype:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/button/src/Button.tsx
+++ b/packages/button/src/Button.tsx
@@ -276,7 +276,6 @@ const baseStyle = css`
   transition: all 120ms ease;
   user-select: none;
   padding: 0;
-  overflow: hidden;
 
   &:focus {
     outline: none;
--- a/packages/button/src/Button.tsx
+++ b/packages/button/src/Button.tsx
@@ -374,7 +374,8 @@ const Button: ExtendableBox<
           [css`
             border-radius: ${borderRadius};
 
-            &:before, &:after {
+            &:before,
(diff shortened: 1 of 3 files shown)


packages/button/src/Button.tsx
```diff
@@ -276,7 +276,6 @@ const baseStyle = css`
   transition: all 120ms ease;
   user-select: none;
   padding: 0;
-  overflow: hidden;
```
brookescarlett, line 279: I don't remember correctly the rationale, but I do believe there was a reason we have this set explicitly. Maybe @DesignerDave does? Whatever the reasoning, perhaps it's worth it to find an alternative on that side, since these seems such a simple/important fix on ours.

lazytype: I believe for the majority of cases this shouldn't affect clients since we have an internal <span> with overflow:hidden that wraps the button children

DesignerDave: I think I remember – this had to do with the gradient transition we set up with pseudo elements. The `overflow: hidden` I think allowed the rounded corners of the button to clip the corners of those gradients.

lazytype: Just updated the PR with a way to handle the clipping. In short, we just set the border radius on the pseudo elements.

brookescarlett (approved):
Looks like it's working to me!

PR #745, opened 2020-12-18
PD-973: Adds Landing and Coming Soon pages to MDB.design

brookescarlett:
## ✍️ Proposed changes

Adds landing page and coming soon page to mdb.design
Also adds some small design fixes from a QA session with @audroue 

🎟 _Jira ticket:_ [PD-973](https://jira.mongodb.org/browse/PD-973)

## 🛠 Types of changes

## ✅ Checklist


website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+      {renderedStory}
+    </div>
+  );
+}
+
+function News() {
+  return (
+    <div className={newsContainer}>
+      <H2 as="h1">Whats New</H2>
+      <Update
```
lazytype, line 100: it would be more scalable for the stories to not be stored all in one file. if adding a new story was as simple as writing a markdown file, then even non-eng could easily add new ones.

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+      />
+      <Update
+        date="Dec 2, 2020"
+        story="Support for React 17"
+        href="https://github.com/mongodb/leafygreen-ui/commit/c18f16e6632a6688e771334fd1672c9ef7e0f9b4"
+      />
+      <Update
+        date="Nov 29, 2020"
+        story="Installing LeafyGreen in Figma"
+        path=""
```
lazytype, line 113: missing?

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  z-index: -1;
+`;
+
+const layoutProperties = css`
+  ${mq({
+    width: ['calc(100% + 48px)', '100%', '100%', '1077px'],
```
lazytype, line 29: 1077 is oddly specific. what is that derived from?

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+  overflow: hidden;
+
+  ${mq({
+    fontSize: ['24px', '60px', '60px', '60px'],
+  })}
+`;
+
+const lgHeight = css`
+  ${mq({
+    height: ['calc(100vw / 2)', '350px', '350px', '350px'],
```
lazytype, line 120: is this not the same as
```suggestion
    height: ['calc(50vw )', '350px', '350px', '350px'],
```

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+  width: 50%;
+`;
+
+const secondRowContainer = css`
+  display: flex;
+  flex-wrap: wrap;
+`;
+
+interface ComponentPreviewProps {
+  path: string;
```
lazytype, line 140: nit: `route`

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+  background: string;
```
lazytype, line 141: nit: `backgroundURL`

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+        <div className={overlineContainer}>
+          <Overline className={overlineColor}>{content}</Overline>
+        </div>
+      </button>
+    </div>
+  );
+}
+
+interface MarketingPreview {
+  href: string;
```
lazytype, line 171: ditto

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+  children: string;
+  background: string;
```
lazytype, line 173: ditto

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+  padding-left: ${spacing[4]}px;
+  text-align: left;
+  overflow: hidden;
+
+  ${mq({
+    fontSize: ['24px', '60px', '60px', '60px'],
+  })}
+`;
+
+const lgHeight = css`
```
lazytype, line 118: is this "leafygreen" or "large"? it's not worth losing clarity to save a few characters

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+  ${mq({
+    height: ['calc(100vw / 2)', '350px', '350px', '350px'],
+  })}
+`;
+
+const smHeight = css`
```
lazytype, line 124: ditto

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+        className={container}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        <div className={marketingWrapper}>
+          <img
+            src={background}
+            alt=""
+            aria-hidden="true"
+            height="100%"
```
lazytype, line 190: why not css?

website/pages/index.tsx
```diff
@@ -1,5 +1,327 @@
...
+          </div>
+        </GridItem>
+        {!isTouchDevice && (
+          <GridItem sm={6} md={6} lg={6}>
+            {/* @ts-expect-error, awaiting URL for marketing page */}
+            <MarketingPreview background="/images/personas-thumbnail.png">
+              Meet our Personas
+            </MarketingPreview>
+          </GridItem>
+        )}
```
lazytype, line 279: I think it'd be preferable to omit this section or have a meaningful placeholder than have it be non-functional

lazytype (changes requested):

website/components/Grid.tsx
```diff
@@ -147,7 +138,7 @@ function getItemStyle(num: number | undefined) {
   }
 
   if (num) {
-    const width = `${(100 / 12) * num}%`;
+    const width = `${(100 / 12) * num}% `;
```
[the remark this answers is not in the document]
brookescarlett: 👍

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+  font-weight: bolder;
+  display: inline-flex;
+  align-items: center;
+
+  &:hover {
+    & > svg {
+      opacity: 1;
+      transform: translate3d(3px, 0, 0px);
+    }
+  }
```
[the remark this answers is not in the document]
brookescarlett: 👍

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+`;
+
+const iconStyle = css`
+  transform: translate3d(-3px, 0, 0px);
+  transition: all 100ms ease-in;
+
+  ${mq({
+    opacity: [1, 1, 0, 0],
```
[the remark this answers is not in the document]
brookescarlett: I've added this attribute, but I actually think the best way to make this accessible is by making the role presentation and the alt text empty such that the icon is not announced to screenreaders. This requires an overhaul of our Icon component, though, because right now we can't overwrite the role="img" property

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+const overlineMargin = css`
+  margin-top: 28px;
+`;
+
+const overlineColor = css`
+  color: ${uiColors.gray.dark1};
+`;
+interface UpdateProps {
+  date: string;
+  story: string;
```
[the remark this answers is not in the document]
brookescarlett: 👍

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+const overlineMargin = css`
+  margin-top: 28px;
+`;
+
+const overlineColor = css`
+  color: ${uiColors.gray.dark1};
+`;
+interface UpdateProps {
+  date: string;
```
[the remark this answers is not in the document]
brookescarlett: 👍

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+  story: string;
+  path?: string;
+  href?: string;
```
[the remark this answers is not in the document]
brookescarlett: 👍

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+  margin-top: 28px;
+`;
+
+const overlineColor = css`
+  color: ${uiColors.gray.dark1};
+`;
+interface UpdateProps {
+  date: string;
+  story: string;
+  path?: string;
```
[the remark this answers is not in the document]
brookescarlett: 👍

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+        {story}
+        <ArrowRightIcon className={iconStyle} />
+      </Subtitle>
+    );
+  } else {
+    renderedStory = (
+      <Subtitle as="a" className={subtitleStyle} href={href}>
+        {story}
+        <ArrowRightIcon className={iconStyle} />
+      </Subtitle>
```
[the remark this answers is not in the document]
brookescarlett: yep, much cleaner

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+    );
+  }
+
+  return (
+    <div className={overlineMargin}>
```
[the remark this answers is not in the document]
brookescarlett: 👍

website/components/News.tsx
brookescarlett: Yes need a link from @audroue

website/pages/index.tsx
brookescarlett: Derived from @audroue

website/pages/index.tsx
brookescarlett: No it is, sorry.. there was a fancier calc at one point but..

brookescarlett: 👍

website/pages/index.tsx
brookescarlett: 👍

website/pages/index.tsx
brookescarlett: 👍

website/pages/index.tsx
brookescarlett: 👍

website/pages/index.tsx
brookescarlett: 👍

website/pages/index.tsx
brookescarlett: kk

website/pages/index.tsx
brookescarlett: 👍

website/pages/index.tsx
brookescarlett: I had this in so that Rob and I could QA, I can omit if you prefer

website/components/News.tsx
```diff
@@ -0,0 +1,121 @@
...
+    </div>
+  );
+}
+
+function News() {
+  return (
+    <div className={newsContainer}>
+      <H2 as="h1">Whats New</H2>
+      <Update
+        date="Dec 8, 2020"
```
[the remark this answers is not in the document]
brookescarlett: lmk what you think of the new approach

website/pages/index.tsx
brookescarlett: If you feel strongly I can change this

website/components/News.tsx
brookescarlett: I don't have access to file reading methods from inside of components that don't exist  in the pages dir. I can move this there and then read data from a static file if you think that makes more sense. Unless you had something else in mind?

lazytype: This component doesn't need to move. It can receive the data it needs from the page through props.

website/components/News.tsx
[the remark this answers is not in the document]
lazytype: >we can't overwrite the role="img" property

would wrapping the icon in an element with `aria-hidden="true"` achieve the same thing?

website/components/News.tsx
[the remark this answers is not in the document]
lazytype: A question I have is whether this approach will actually use the viewer's locale if it's SSR. Do we need to something to force the date to re-render on the client side? Also, I believe when the date is parsed, it's parsed in UTC+0, not EST, so this could lead to a date being rendered different than even we'd expect.

website/pages/index.tsx
lazytype: It would make updating and debugging styles easier if they're consistently defined via CSS

website/pages/index.tsx
lazytype: It shouldn't be up to my preference. Rather, we should collectively decide whether we want to display non-functional content in production. I think a link to a coming soon page would be better than a non-functional one.

You may want to look into Feature Flags as a general solution to being able to toggle on functionality only for development.

website/components/News.tsx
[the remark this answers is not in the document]
brookescarlett: yep, good call

website/pages/index.tsx
brookescarlett: 👍

website/components/News.tsx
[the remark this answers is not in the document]
brookescarlett: As far as timeZone goes it looks like the default is: 

```timeZone
The time zone to use. The only value implementations must recognize is "UTC"; the default is the runtime's default time zone. Implementations may also recognize the time zone names of the IANA time zone database, such as "Asia/Shanghai", "Asia/Kolkata", "America/New_York".```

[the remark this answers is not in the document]
Comment (brookescarlett) [reply website/components/News.tsx]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

**Comment by brookescarlett** [comment]
> `dateText` is kind of a misnomer now. To avoid issues with time zones I think it'd be better for it to be a string or numeric timestamp

Makes sense -- reverting back.

**Comment by brookescarlett** [comment]
> I think this would throw an error if it's not defined. `typeof navigator !== 'undefined'` should be safe

Updated

File: website/components/News.tsx (hunk 1 above)
Comment (brookescarlett) [reply website/components/News.tsx]: Updated

File: website/updates/Update1.md [hunk 23]
```diff
@@ -0,0 +1,5 @@
+---
```
lazytype, line 1: let's give the file names more meaningful names to make it easier to find specific ones, e.g.`/2020-12-08-copyable-release.md`

website/utils/fetchUpdates.ts
```diff
@@ -0,0 +1,30 @@
...
+const updatesDirectory = join(process.cwd(), 'updates');
+
+function getUpdateSlugs() {
+  return fs.readdirSync(updatesDirectory);
+}
+
+function getIndividualUpdate(updateName) {
+  const slug = updateName.replace(/\.md$/, '');
+  const fullPath = join(updatesDirectory, `${slug}.md`);
+  const fileContents = fs.readFileSync(fullPath, 'utf8');
```
lazytype, line 21: would it be much trouble to use the async api?

website/utils/fetchUpdates.ts
```diff
@@ -0,0 +1,30 @@
...
+  const { data } = matter(fileContents);
+
+  return data;
+}
+
+export function getAllUpdates() {
```
lazytype, line 27: can you annotate this with an explicit return type

website/updates/Update1.md
brookescarlett: updated

website/utils/fetchUpdates.ts
brookescarlett: nope, that's definitely the right call

PR #747, opened 2020-12-21
PD-942: Add isHeader prop to table cell component

DesignerDave:
## ✍️ Proposed changes

This PR adds an `isHeader` prop to the `Cell` component in the `@leafygreen-ui/table` package. This prop removes an outstanding barrier for docs to use our table component.

There were two main approaches which I considered taking for this API:
1. Make the existing TableHeader component work within Rows.
2. Add a prop to the Cell component to render a Cell as a `th` element.

I opted to implement the second option for a few reasons. The first is that the change is a significantly smaller one. Significant work would've needed to be done in TableHeader that would've resulted in far more code, overwriting quite a few styles, and would've duplicated a lot of code. This approach also makes the path to using TH elements in rows for existing Table implementations slightly simpler.

I also broke up the tests for Table by the constituent components. The tests were already structured by component, so while the changes look fairly large, almost nothing has changed within the tests aside from the addition of a test for this prop.

🎟 _Jira ticket:_ [Add support for Table Headers in Rows within Table](https://jira.mongodb.org/browse/PD-942)

## 🛠 Types of changes

## ✅ Checklist


packages/table/src/Cell.tsx
```diff
@@ -5,30 +5,58 @@ import { commonCellStyles } from './styles';
 
 export const tdInnerDiv = createDataProp('td-inner-div');
 
-type CellProps = HTMLElementProps<'td', HTMLTableCellElement>;
+interface HeaderCellProps
+  extends HTMLElementProps<'th', HTMLTableHeaderCellElement> {
+  isHeader?: true;
```
brookescarlett, line 10: Don't think this should be optional. isHeader is necessarily true for us to be using HeaderCellProps

packages/table/src/Cell.tsx
```diff
@@ -5,30 +5,58 @@ import { commonCellStyles } from './styles';
...
+      props.scope = 'row';
+    }
+
     return (
-      <td
-        ref={ref}
-        className={cx(commonCellStyles, tdStyles, className)}
-        {...rest}
-      >
+      <Element {...props} {...rest}>
```
brookescarlett, line 55: nit: Would prefer that we call this Root or Component to be consistent with other occasions across the repo where we do something similar, but this isn't imperative

DesignerDave: Done

packages/table/src/Cell.tsx
DesignerDave: You're right

brookescarlett (approved):
LGTM

PR #752, opened 2020-12-22
PD-917: Major Toggle upgrades

DesignerDave:
## ✍️ Proposed changes

This update is a massive upgrade for the Toggle component, improving accessibility, simplifying the component's structure and styles, and improving documentation.

- Refactors the internal DOM structure of the component to be as accessible as possible, and uses the appropriate role.
- Restructures how the styles are structured in the component to improve maintainability.
- Slightly increases contrast of the "ON" and "OFF" labels in the default size to meet WCAG AA contrast guidelines.
- Hides the "ON" and "OFF" labels for screen readers so that only the current state of the Toggle is read.
- Enforces use of aria-label and aria-labelledby so that Toggles always have screen reader accessible text associated with them.
- Fixes a bug with the rendering of the focus state on Windows machines.
- Uses the LeafyGreen Provider to conditionally show the focus state based on how the user is interacting with the page.
- Introduces an `UPGRADE.md` file to help folks upgrade to the latest more effectively. We can add on to this for each major version if we'd like, but I think it's important here.

🎟 _Jira ticket:_ [PD-916](https://jira.mongodb.org/browse/PD-916)
🎟 _Jira ticket:_ [PD-917](https://jira.mongodb.org/browse/PD-917)

## 🛠 Types of changes

## ✅ Checklist


packages/toggle/UPGRADE.md
```diff
@@ -0,0 +1,7 @@
+# v6 to v7
```
brookescarlett, line 1: AFAICT there's no reason that this shouldn't just be a part of the changelog

packages/toggle/README.md
```diff
@@ -23,49 +23,56 @@ npm install @leafygreen-ui/toggle
...
+| `onClick`   | `(MouseEvent) => void`             | The event handler function for the 'onchange' event. Receives the associated `event` object as the first argument.                                                                                              |             |
+| `onChange`  | `(checked, MouseEvent) => void`    | `onChange` fires when the internally-managed `checked` state of the component is updated. Receives the updated checked state of the toggle as its first argument, and the associated mouse event as the second. |             |
+| `className` | `string`                           | Adds a className to the outermost element.                                                                                                                                                                      | `''`        |
+| ...         | HTML `button` attributes           | Any natively-supported HTML button attributes will be applied to the `button` element                                                                                                                           |             |
+
+## Accessibility
```
brookescarlett, line 74: Would it make sense to have a useAccessibleToggle hook which takes in a Label component and a Toggle component and returns cloned versions of the components that are accessibly associated with one another? Might not be necessary for this version, but also might help us ensure that our Toggle components are being implemented accessibly and reduce work to do so on the consuming applications' end.

DesignerDave: Maybe, but I'd doubt it'd be much easier to get people to use it tbh.

packages/toggle/UPGRADE.md
DesignerDave: The problem with Changelogs is once we have 3-4 other minor and patch versions, this information will get lost easily. This is experimental for sure, but I wanted a dedicated place for upgrade guides between major versions to see if that helps folks make those upgrades. I've been listening to feedback that would point to this being useful.

If it turns out to not be that useful, we don't have to do this again.

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,416 @@ export const Mode = {
...
+   * Determines if the Toggle will render the dark mode styles.
+   *
+   * default: `false`
+   */
   darkMode?: boolean;
-}
 
-type ToggleProps = BaseToggleProps &
-  Omit<HTMLElementProps<'input', never>, keyof BaseToggleProps>;
+  /**
```
brookescarlett, line 317: Why do we recomend this approach?

packages/toggle/README.md
DesignerDave: Honestly, I'd rather educate people a bit on this stuff than us providing abstractions for actions they themselves have to take. It can help the company as a whole understand accessibility a bit better.

packages/toggle/src/Toggle.tsx
DesignerDave: I think you're right. Thinking more through the use-cases of toggle, it's more likely they'll want to use it as a controlled input.

DesignerDave: I'll update this.

packages/toggle/UPGRADE.md
brookescarlett: I don't think it makes sense to be adding yet another place where we document changes, it just becomes another burden on us to maintain. While this specific information is relevant it'll likely remain at the top of the changelog and if needed we can direct people to a certain line number explicitly. We'd also then have to document that we even have such files like an UPGRADE.md so that other devs know to check the resource, and since we don't have this for any other component yet I think that may be confusing.

brookescarlett: Also, as mdb.design will increasingly become the main hub for developer resources, we'd need to think about how this would translate there rather than this information being already available in the changelog.

DesignerDave: I'm not committing to keeping this, just seeing if this is helpful for the folks who have said it's hard to understand what needs to change for major versions. We don't have to document this anywhere except when communicating this specific major version change.

Also worth noting, many apps don't update to our major versions for 3-6 months in some cases. We can't rely on this information remaining accessible for that long in our changelog.

DesignerDave: We **don't** need this to be on mongodb.design. For the time being, think of this as a one-off unless this proves significantly beneficial to folks.

brookescarlett: Ok fine to trial run this, I just worry about maintainability and ensuring that everyone knows about all of the resouces that we have available as it relates to documentation

DesignerDave: I'm looking for ways to improve our communication, and if this is ineffective or confusing, we won't do it again 😄

packages/toggle/src/Toggle.tsx
DesignerDave: Changed it up to no longer offer that recommendation.

packages/toggle/UPGRADE.md
brookescarlett: Sounds good :)

packages/toggle/src/Toggle.tsx
brookescarlett: 👍

packages/toggle/README.md
```diff
@@ -23,49 +23,56 @@ npm install @leafygreen-ui/toggle
...
+| `onClick`   | `(MouseEvent) => void`             | The event handler function for the 'onchange' event. Receives the associated `event` object as the first argument.                                                                                              |             |
+| `onChange`  | `(checked, MouseEvent) => void`    | `onChange` fires when the internally-managed `checked` state of the component is updated. Receives the updated checked state of the toggle as its first argument, and the associated mouse event as the second. |             |
+| `className` | `string`                           | Adds a className to the outermost element.                                                                                                                                                                      | `''`        |
+| ...         | HTML `button` attributes           | Any natively-supported HTML button attributes will be applied to the `button` element                                                                                                                           |             |
```
brookescarlett, line 72: We have a convention for describing rest props now, check out banner's readme

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,416 @@ export const Mode = {
...
+  onChange: onChangeProp,
+  onClick: onClickProp,
   checked: checkedProp,
-  id: idProp,
   ...rest
 }: ToggleProps) {
-  const [checked, setChecked] = useState(false);
-  const toggleId = useMemo(() => idProp ?? idAllocator.generate(), [idProp]);
-  const normalizedChecked = checkedProp || checked;
+  validateAriaLabelling(rest, 'Toggle');
```
lazytype, line 364:
```suggestion
  validateAriaLabelling(rest, Toggle.displayName);
```

packages/toggle/README.md
brookescarlett: Another thought: a hook that accepts the label and toggle props and returns cloned (accessible) props to be spread on the respective elements... while I am all for increasing education/awareness of accessibility best-practices, I also see it as our goal to make these things as baked in as possible, and our components as easy to use accessibly as possible. Again, this doesn't have to happen in this PR but just something I'm thinking about

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,416 @@ export const Mode = {
...
+  /**
+   * `onChange` fires when the internally-managed `checked` state of the component is updated. Receives the updated checked state of the toggle as its first argument, and the associated mouse event as the second.
+   */
+  onChange?: (
+    checked: boolean,
+    mouseEvent: React.MouseEvent<HTMLButtonElement>,
+  ) => void;
+
+  /**
+   * The event handler function for the 'onchange' event. Receives the associated `event` object as the first argument.
```
brookescarlett, line 338: nit: I think you meant 'onclick' event

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,416 @@ export const Mode = {
...
-      setChecked(e.target.checked);
+  const onClick = (e: React.MouseEvent<HTMLButtonElement>) => {
+    if (onClickProp) {
+      onClickProp(e);
     }
+
+    setChecked(curr => {
+      const updatedState = !curr;
+
+      if (onChangeProp) {
```
brookescarlett, line 378: Can do the same here: `onChangeProp?.(updatedState, e)`

packages/toggle/README.md
DesignerDave: I was looking to improve clarity, especially if a developer using our components is unfamiliar with the term "spread". I've met developers at the company who have never used the syntax/don't know what it means, but they use our components.

"native" -> "HTML"
> This is a bit more specific and clear IMO

"Any other props will be spread on the root `div` element" ->" Any natively-supported HTML button attributes will be applied to the `button` element"
> This does a few things. First, the button isn't the root element, so that would be inaccurate. Second, by specifying button attributes it makes the expectations for what you can pass more clear as it mirrors more closely what React documents. Saying  "props" is only relevant if offering an `as` prop in a component IMO.

packages/lib/src/accessibilityHelpers.ts
```diff
@@ -0,0 +1,13 @@
+// Validate that either aria-label or aria-labelledby are present within props.
+export function validateAriaLabelling(
```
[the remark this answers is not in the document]
DesignerDave: Sure

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,416 @@ export const Mode = {
...
-      focusTargetElement={inputElement}
+      focusTargetElement={buttonElement}
+      className={className}
     >
-      <label
-        className={cx(statefulStyles.container, className)}
-        htmlFor={toggleId}
+      <button
+        role="switch"
+        onClick={onClick}
```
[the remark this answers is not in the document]
DesignerDave: Yes – the role handles this behavior. You can test that locally if you want to validate, but I tested this component's keyboard interaction, and tested the changes with a screen reader to land where I did.

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,416 @@ export const Mode = {
...
-    if (onChangeProp) {
-      onChangeProp(e);
-    }
+  const [buttonElement, setButtonElement] = useState<HTMLElement | null>(null);
+  const [checked, setChecked] = useState<boolean>(checkedProp || false);
+  const normalizedChecked = checkedProp ?? checked;
 
-    if (checkedProp == null) {
-      setChecked(e.target.checked);
+  const onClick = (e: React.MouseEvent<HTMLButtonElement>) => {
```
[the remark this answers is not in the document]
DesignerDave: Sounds good.

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,416 @@ export const Mode = {
...
+    if (onClickProp) {
+      onClickProp(e);
```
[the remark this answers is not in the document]
DesignerDave: 👍

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,416 @@ export const Mode = {
...
-  const toggleId = useMemo(() => idProp ?? idAllocator.generate(), [idProp]);
-  const normalizedChecked = checkedProp || checked;
+  validateAriaLabelling(rest, 'Toggle');
 
-  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    if (onChangeProp) {
-      onChangeProp(e);
-    }
+  const [buttonElement, setButtonElement] = useState<HTMLElement | null>(null);
+  const [checked, setChecked] = useState<boolean>(checkedProp || false);
```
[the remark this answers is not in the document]
DesignerDave: Nice catch. Not sure how this was working before.

packages/toggle/src/Toggle.tsx
DesignerDave: Hah yep, nice catch.

packages/toggle/README.md
DesignerDave: I think where I'm not convinced is that I feel that having an external hook isn't baking in the behavior at all. It's effectively as much work to make the component accessible, but stripping any education about accessibility away in the process. If we made a hook, there's still something to learn, but what the developer would learn in that case isn't transferrable to other parts of their work.

 I generally favor standard solutions over proprietary ones where possible, and this is one of those cases.

packages/toggle/src/Toggle.tsx
DesignerDave: 👍

packages/toggle/src/Toggle.tsx
```diff
@@ -24,431 +25,415 @@ export const Mode = {
...
+      onClickProp?.(e);
 
-  const statefulStyles = getStatefulStyles({
-    disabled,
-    size,
-    checked: normalizedChecked,
-    mode: darkMode ? Mode.Dark : Mode.Light,
-  });
+      setChecked(curr => {
+        const updatedState = !(checkedProp ?? curr);
```
DesignerDave, line 375: @brookescarlett @lazytype I wanted to call out this change I made to make sure it makes sense.

When you have a controlled checkbox and are responding to its onChange events, I think you get the value that it should be updated to regardless of whether it's controlled or not. Do y'all agree that this makes more sense to pass to the user's onChange callback than simply `!curr`?

brookescarlett: Taking a closer look, I don't think that it makes sense to break this out into two separate props. They're being triggered by the same event, and there's no difference in how/when they're being called, as the only check is for the respective function's existence. I think we should continue with just the `onChange` prop and passing the event up to the consuming application (as is consistent with most of our form-like elements). If a user wants to use that information to toggle internal state, they are able to do so as-is.

DesignerDave: The reason I included both is mostly to mirror the feature set of using a checkbox. If we don't call onClick in here,  onClick won't work at all, and that's not really okay. If we don't call onChange here, that won't work either since it's non-standard for buttons.

brookescarlett: I've been doing some research and it seems to me that `onChange` really is the conventional event handler when it comes to responding to Toggle/Switch components, so I am inclined to say that we follow-suit, and leave the code you have here as-is, but just not explicitly document the onClick prop. That way, onClick won't break anything, but also isn't what we are recommending in terms of how to use the component.

The only drawback to this approach is that I don't love that the type of event that the onChange would expect to receive isn't the type of event that it actually will receive (expect: React.ChangeEvent, receive: React.MouseEvent). Not sure this needs to be of the utmost concern when making this decision, but it is something I thought I'd raise.

lazytype (approved):

packages/toggle/src/Toggle.tsx
DesignerDave: If it means we're leaving the code here as-is, I don't feel strongly about removing the documentation.

I agree with your feelings about passing a mouse event to a change event handler, but the alternatives are constructing an equivalent of a change event, or not providing one at all. Both seem like worse alternatives than the client receiving the event that is causing an onChange event to fire.

brookescarlett (approved):
Talked offline about a few final comments but nothing blocking. lgtm.

PR #754, opened 2020-12-29
PD-711, PD-878, PD-910, PD-953, PD-972, PD-992, PD-1004: Code component updates

DesignerDave:
## ✍️ Proposed changes

This PR brings a number of bug fixes, accessibility improvements, design tweaks, and code maintenance work to the Code component, and removes the Syntax component due to how tightly coupled with the Code component it is.

🎟 _Jira ticket:_ [Code: Font size too big on touch device](https://jira.mongodb.org/browse/PD-992)
🎟 _Jira ticket:_ [Explore changing the color of comments in the Syntax and Code components to differentiate further from code](https://jira.mongodb.org/browse/PD-711)
🎟 _Jira ticket:_ [In certain cases, the Code component will incorrectly number multi-line statements, only numbering the first line](https://jira.mongodb.org/browse/PD-878)
🎟 _Jira ticket:_ [Don't expand ranges in Syntax component](https://jira.mongodb.org/browse/PD-910)
🎟 _Jira ticket:_ [Deprecate Syntax and co-locate highlight.js code in Code component](https://jira.mongodb.org/browse/PD-953)
🎟 _Jira ticket:_ [Fix highlighting in Syntax component darkMode in Safari](https://jira.mongodb.org/browse/PD-972)
🎟 _Jira ticket:_ [Increase contrast between syntax highlighted tokens for users with poor color vision](https://jira.mongodb.org/browse/PD-1004)

## 🛠 Types of changes

## ✅ Checklist


.changeset/early-lions-turn.md
```diff
@@ -0,0 +1,21 @@
+---
+'@leafygreen-ui/code': major
```
brookescarlett, line 2: Should we add a separate changeset for the Syntax component indicating that it's deprecated?

brookescarlett:
I think we should follow the same pattern for deprecating the Syntax package that we did with MongoMenu, which is to say we leave the package with a package.json, changelog and readme (indicating that the package is deprecated). If we decide we want to do something different here, we should update MongoMenu to follow suit.

packages/code/src/Code.tsx
```diff
@@ -288,14 +274,22 @@ function Code({
...
+    },
   );
 
   const renderedSyntaxComponent = (
     <Syntax
       showLineNumbers={showLineNumbers}
       darkMode={darkMode}
       language={language}
-      highlightLines={highlightLines}
+      highlightLines={highlightLines || []}
```
brookescarlett, line 292: Shouldn't need this, since we highlightLines will default to [] already

packages/code/src/renderingPlugin.tsx
```diff
@@ -136,8 +153,10 @@ export function LineTableRow({
           className={cx(
             cellStyle,
             css`
-              padding-right: 24px;
               user-select: none;
+              text-align: right;
+              padding-left: ${spacing[2]}px;
+              padding-right: 0;
               color: ${highlighted ? highlightedNumberColor : numberColor};
```
brookescarlett, line 160: In axe devtools, I'm seeing that this color does not have sufficient contrast

packages/code/src/Code.tsx
```diff
@@ -49,7 +53,26 @@ const copyStyle = css`
...
+  height: 36px;
+  padding-top: 0;
+  justify-content: center;
+`;
+
+const singleLineWrapperStyle = css`
+  align-items: center;
+  display: flex;
+
+  ${mq({
```
brookescarlett, line 65: Can we combine this and the below in a single media query

packages/code/src/types.ts
```diff
@@ -47,5 +49,44 @@ export interface SyntaxProps extends React.HTMLAttributes<HTMLElement> {
   /**
    * An array of lines to highlight. The array can only contain numbers corresponding to the line numbers to highlight, and / or tuples representing a range (e.g. `[6, 10]`);
    */
-  highlightLines?: Array<number | [number, number]>;
+  highlightLines?: LineHighlightingDefinition;
+}
+
+export interface CodeProps extends Omit<SyntaxProps, 'onCopy'> {
```
brookescarlett, line 55: If we're extending SyntaxProps, why redefine overlapping props? We can probably remove explicit definitions for highlightLines and showLineNumbers.

Also, is there a reason that we're omitting onCopy from Syntax? We're not defining it more explicitly below.

.changeset/early-lions-turn.md
DesignerDave: Decided not to do this. I don't believe anyone's using the package, and if they are, the current version is on NPM.

packages/code/src/Code.tsx
```diff
@@ -288,14 +274,22 @@ function Code({
     },
     className,
     getScrollShadowStyle(scrollState, mode),
+    { [singleLineWrapperStyle]: !isMultiline },
+    {
+      [css`
```
brookescarlett, line 279: we can inline this, but it's enough lines that it wouldn't hurt to pull into it's own variable

packages/code/src/Syntax.tsx
```diff
@@ -12,36 +11,6 @@ import { injectGlobalStyles } from './globalStyles';
...
-      }
-
-      return item;
-    }),
-  );
-}
-
 type FilteredSupportedLanguagesEnum = Omit<
   typeof SupportedLanguages,
   'Cs' | 'JS' | 'TS'
```
brookescarlett, line 16: Can we add a comment here that just explains that these are aliases/dupes

packages/code/src/Code.tsx
DesignerDave: You'd think so, but I was hitting an exception before adding this. Can't remember exactly what it was but if you feel strongly, I can probably reproduce it.

packages/code/src/renderingPlugin.tsx
DesignerDave: Yeah, I mostly left this unchanged here, though IIRC I slightly bumped the contrast in either light or dark mode. The line numbers are decorative, so I'm not sure what the best approach is here to make the devtools pass. It's something we can address down the line, given it's not a regression or goal of this PR.

packages/code/src/renderingPlugin.tsx
```diff
@@ -87,6 +101,9 @@ function getHighlightedRowStyle(darkMode: boolean) {
   return css`
     background-color: ${backgroundColor};
     background-image: ${backgroundImage};
+    // Fixes an issue in Safari where the gradient applied to the table row would be applied
+    // to each cell in the row instead of being continuous across cells.
```
brookescarlett, line 105: I'm still seeing a different highlight color in darkmode across Chrome and Safari.

Chrome:
Screen Shot 2021-01-04 at 10 43 42 AM

Safari:
Screen Shot 2021-01-04 at 10 43 45 AM

.changeset/early-lions-turn.md
brookescarlett: That's fine, then let's do the same for MongoMenu

brookescarlett:
Saw that you removed Syntax from the website directory (thank you) do you mind also removing it from the Navigation component `website/components/navigation/Navigation.tsx`  line 61 so that it doesn't appear on the side bar leading to a 404?

packages/code/src/renderingPlugin.tsx
DesignerDave: That's a difference in the way gradients render across browsers, not a bug in the implementation here.

packages/code/src/types.ts
DesignerDave: This is all pulled from the previous implementation of the component IIRC. You're probably right about removing those two definitions.

If I'm remembering right, `onCopy` being omitted and redefined is due to SyntaxProps including a definition for the native version of this handler. Once again, not something I added or changed aside from moving this code.

packages/code/src/Code.tsx
DesignerDave: I don't feel strongly about this either way. Done!

packages/code/src/types.ts
DesignerDave: Done with that first bit!

.changeset/early-lions-turn.md
DesignerDave: I think our approach to MongoMenu was correct, and was a different case, but it was never meant to stick around as long as it did. I'm alright with removing it at this point.

packages/code/src/Code.tsx
DesignerDave: Didn't know you could do this with facepaint 😮 Done!

packages/code/src/Syntax.tsx
DesignerDave: Sure

.changeset/early-lions-turn.md
DesignerDave: *In a separate PR

DesignerDave:
> Saw that you removed Syntax from the website directory (thank you) do you mind also removing it from the Navigation component `website/components/navigation/Navigation.tsx` line 61 so that it doesn't appear on the side bar leading to a 404?

Sure!

packages/code/src/Code.tsx
brookescarlett: I tried to repro locally and am not seeing an issue 🤔

packages/code/src/renderingPlugin.tsx
brookescarlett: Fair, we should document that this is an outstanding issue somewhere though

packages/code/src/Code.tsx
brookescarlett: 👍

packages/code/src/Code.tsx
DesignerDave: I think it was an edge case when switching between different options in the highlight lines knob. Let me see if I can reproduce it. I remember this just being an edge case I hit when testing the component manually.

packages/code/src/types.ts
brookescarlett: Makes sense, it's easier to detect these things now that they're colocated :)

packages/code/src/renderingPlugin.tsx
brookescarlett: Is there a way to make them look more visually consistent, even if it isn't a bug in implementation?

README.md
```diff
@@ -36,7 +36,6 @@ A set of CSS styles and React components built with design in mind.
 - [Select](https://github.com/mongodb/leafygreen-ui/tree/main/packages/select)
 - [Side Nav](https://github.com/mongodb/leafygreen-ui/tree/main/packages/side-nav)
 - [Stepper](https://github.com/mongodb/leafygreen-ui/tree/main/packages/stepper)
-- [Syntax](https://github.com/mongodb/leafygreen-ui/tree/main/packages/syntax)
```
brookescarlett, line 39: Can also be removed from global tsconfig

brookescarlett: `build.tsconfig.json`

packages/code/src/index.ts
```diff
@@ -1,4 +1,10 @@
...
 
-export { Language } from '@leafygreen-ui/syntax';
+export { Language, Mode } from './types';
+export type {
+  SyntaxProps,
+  CodeProps,
+  LineHighlightingDefinition,
+} from './types';
+export { variantColors } from './globalStyles';
 export default Code;
```
brookescarlett, line 10: nit, can do export default from

packages/code/src/types.ts
```diff
@@ -33,19 +35,46 @@ export interface SyntaxProps extends React.HTMLAttributes<HTMLElement> {
...
+   *
+   * default: `true`
+   */
+  copyable?: boolean;
+
+  /**
+   * Callback fired when Code is copied via the copy button.
+   *
+   */
+  onCopy?: Function;
```
brookescarlett, line 79: We can still probably be more specific here

brookescarlett:
Another really small nit but while you're updating the website based on these changes, you can set showLineNumbers in the CodeDocs page

packages/code/src/renderingPlugin.tsx
DesignerDave: Hah I spent a little more time with this problem and it brought be back a bit to a different cross-browser issue I found with gradients probably 4 years ago. Transparent in Safari is `rgba(0, 0, 0, 0)`, but transparent in Chrome is `rgba(255, 255, 255, 0)`. Usually, this is not a meaningful distinction, but in gradients it means that the second stop would be lighter than it should in Chrome, and darker in Safari.

I'm pushing a fix for this now. It's worth noting that there are still issues with colors rendering differently across browsers that's good to be aware of, but this was a different case.

README.md
DesignerDave: Nice call

packages/code/src/types.ts
DesignerDave: There isn't really more specificity needed here. It's a function invoked with 0 parameters.

packages/code/src/types.ts
DesignerDave: > Another really small nit but while you're updating the website based on these changes, you can set showLineNumbers in the CodeDocs page

Done

packages/code/src/index.ts
DesignerDave: Done

packages/code/src/renderingPlugin.tsx
DesignerDave: https://jira.mongodb.org/browse/PD-1011

packages/code/src/Code.tsx
DesignerDave: I couldn't reproduce, so I'm removing this change.

packages/code/src/renderingPlugin.tsx
brookescarlett: Looks a lot better!

.changeset/early-lions-turn.md
brookescarlett: sgtm let's just document that this change needs to be made

brookescarlett: https://jira.mongodb.org/browse/PD-1012

DesignerDave: Thanks!

storybook/preview-head.html
```diff
@@ -1,9 +1,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1" />
 
-<link
-  href="https://fonts.googleapis.com/css2?family=Source+Code+Pro&display=swap"
-  rel="stylesheet"
-/>
+<link rel="preconnect" href="https://fonts.gstatic.com" />
```
brookescarlett, line 3: should this be changed in the website head too?

brookescarlett: nvm you already thought of that :)

.changeset/early-lions-turn.md
```diff
@@ -0,0 +1,21 @@
...
+
+Design changes:
+- Updates the syntax highlighting theme to improve contrast between comments and regular code, as well as making several optimizations for red-green colorblindness.
+- Tightens the top and bottom padding for the code component, especially for single line code examples.
+- Significantly tightens the left padding when line numbers are shown.
+- Right-aligns line numbers in their column to improve readability.
+
+Bug fixes:
+- Fixes an issue where iOS browsers would render code examples at a `20px` size, rather than the `13px` font size they're supposed to render as.
+- Sometimes code that's written as multi-line statements wouldn't have line numbers applied to each line of the statement. That should now be fixed.
```
brookescarlett, line 15: This is fixed as far as I can tell, do we know what was causing it in the first place? Is there a test that we can add to make sure we don't introduce a regression later?

DesignerDave: This was technically not a bug. Apparently the specs for browsers to implement are unspecific when it comes to rendering text with `white-space: pre`. iOS specifically applies a font size that can't be overwritten on mobile of `20px`, and unfortunately that means that in those cases, we have to wrap long lines of code instead of overflow scroll handling it to fix this issue. CSS-only fixes are difficult to effectively test without visual diffing, so I don't think there's a good test to add here.

DesignerDave: Ignore my last response – it was in regard to something else.

highlight.js was outputting a nested structure for the parsed code to highlight with. We weren't doing anything with that nested structure, causing it to be output on the same effective line regardless of a line break. I'll add a test for the function I added to flatten that data structure.

brookescarlett: sgtm, thank you. just want to make sure we're holding ourselves accountable to this solution

packages/code/src/renderingPlugin.tsx
```diff
@@ -61,22 +82,20 @@ const cellStyle = css`
...
-
-  &:first-of-type {
-    padding-left: 12px;
-  }
-
-  &:last-of-type {
-    padding-right: 12px;
-  }
+  padding-right: ${spacing[3]}px;
+  padding-left: ${spacing[3]}px;
```
lazytype, line 86: we can collapse some of these
`padding: 0 ${spacing[3]}px;`

packages/code/src/renderingPlugin.tsx
```diff
@@ -151,46 +175,131 @@ export function LineTableRow({
...
+      // If there's a kind, we construct a custom token object with that kind to preserve highlighting.
+      // Without this, the value will simply render without highlighting.
+      const child = kind
+        ? { kind: generateKindClassName(kind), children: [val] }
+        : val;
+
+      return [...acc, child];
+    }
+
+    if (val?.children?.length > 1) {
```
lazytype, line 215: I think it's kinda sketchy to potentially compare `undefined > 1`, even if the behavior is well-defined in JS. how about doing `(val?.children?.length ?? 0) > 1` to make it more explicit

packages/code/src/renderingPlugin.tsx
```diff
@@ -151,46 +175,131 @@ export function LineTableRow({
...
+  if (isArray(children) && children.length === 1 && isString(children[0])) {
+    return true;
+  }
+
+  return false;
+}
+
+// If an array of tokens contains an object with more than one children, this function will flatten that tree recursively.
+function flattenNestedTree(
+  children: Array<string | TokenObject>,
```
lazytype, line 197: we could make this `children: TokenObject['children']`, to avoid some redundancy and make it clear where we expect this to come from

packages/code/src/renderingPlugin.tsx
```diff
@@ -151,46 +175,131 @@ export function LineTableRow({
...
+
+    if (isFlattenedTokenObject(val)) {
+      return [
+        ...acc,
+        { kind: generateKindClassName(kind, val.kind), children: val.children },
+      ];
+    }
+
+    return acc;
+  }, [] as Array<string | FlatTokenObject>);
```
lazytype, line 234: you can explicitly annotate `(acc: Array<string | FlatTokenObject>, ` above instead of doing an assertion here

packages/code/src/renderingPlugin.tsx
```diff
@@ -216,11 +325,26 @@ export function TableContent({ lines }: TableContentProps) {
...
   }
 
+  const lineShouldHighlight = (line: number) => {
+    return highlightLines.some(def => {
+      if (isNumber(def)) {
+        return line === def;
+      }
+
+      if (isArray(def)) {
+        const sortedArr = def.sort();
```
lazytype, line 335: we can avoid potentially unexpected side-effects by not sorting in place,`[...def].sort()`

packages/code/src/renderingPlugin.tsx
DesignerDave: Oh you're right, I have `padding: 0` above. Nice catch.

packages/code/src/renderingPlugin.tsx
DesignerDave: Seems a bit unnecessary, but I don't feel strongly. I'll update it.

packages/code/src/renderingPlugin.tsx
DesignerDave: That's fair. Will do

packages/code/src/renderingPlugin.tsx
DesignerDave: 👍

packages/code/src/Code.story.tsx
```diff
@@ -23,31 +29,37 @@ storiesOf('Code', module).add(
...
+            )}
+            highlightLines={
+              select(
+                'highlight lines',
+                {
+                  none: [],
+                  single: [1],
+                  // @ts-expect-error
+                  // Knobs typing seems to be off – it says that this isn't assignable to a string.
+                  multiple: [[2, 4], 6],
```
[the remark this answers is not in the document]
DesignerDave: Ahh thank you, this was a pain.

packages/code/src/renderingPlugin.tsx
DesignerDave: Oh man, I didn't realize that TS would pick up the typing like that for some reason. This is great.

packages/code/src/Code.story.tsx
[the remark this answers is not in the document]
DesignerDave: Actually, wouldn't this require a consumer of the component to do something like this to implement the component?

```ts
<Code highlightLines={[[2, 3] as const] as const} />
```

If so, that doesn't seem like a great Dx, and I'd rather keep the story file a bit messier.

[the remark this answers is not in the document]
lazytype: it wouldn't since a mutable array is a subtype of an immutable one!

lazytype (approved):

PR #763, opened 2021-01-07
Attempts to prevent FOUC in .design

brookescarlett:
I did a lot of research to try and understand how to prevent the flash of unstyled content that's currently occurring with .design. Ultimately, I based my solution off of a PR to vercel's Next.js repo where they feature examples of using Next with all different libraries/frameworks (see here: https://github.com/vercel/next.js/pull/20228/files). I'm not convinced that this is 100%, but I do think it's a step in the right direction. I'm finding that it works better on dev than prod (`yarn dev` v. `yarn build && yarn start`) and that it works better in Chrome/Safari than Firefox. 

Let me know what you think and if you have ideas at a better approach here. I know for sure that whatever solution will require modifying the`_document.tsx` file as that's explicitly called out in the `emotion` docs (see here: https://emotion.sh/docs/ssr) but have found a few different implementations out there (especially with the recently released emotion 11)

Very open to feedback/thoughts!

### DIFF
--- a/website/pages/_document.tsx
+++ b/website/pages/_document.tsx
@@ -1,0 +1,35 @@
+import Document, { Html, Head, Main, NextScript } from 'next/document';
+import * as React from 'react';
+import { renderStatic } from 'utils/renderer';
+
+export default class AppDocument extends Document {
+  static async getInitialProps(ctx) {
+    const page = await ctx.renderPage();
+    const { css, ids } = await renderStatic(() => page.html);
+    const initialProps = await Document.getInitialProps(ctx);
+    return {
+      ...initialProps,
+      styles: (
+        <React.Fragment>
+          {initialProps.styles}
+          <style
+            data-emotion={`css ${ids.join(' ')}`}
+            dangerouslySetInnerHTML={{ __html: css }}
+          />
+        </React.Fragment>
+      ),
+    };
+  }
+
+  render() {
+    return (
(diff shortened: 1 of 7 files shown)


website/pages/_document.tsx
```diff
@@ -0,0 +1,35 @@
+import Document, { Html, Head, Main, NextScript } from 'next/document';
+import * as React from 'react';
+import { renderStatic } from 'utils/renderer';
+
+export default class AppDocument extends Document {
```
lazytype, line 5: can you add docstring with a link to the Next.js emotion example here

website/utils/renderer.ts
```diff
@@ -0,0 +1,15 @@
+import createEmotionServer from '@emotion/server/create-instance';
+import { cache } from '@emotion/css';
+
+export const renderStatic = async (callback: Function) => {
```
lazytype, line 4: `Function` will make `html` be typed as `any`
```suggestion
export const renderStatic = async (callback: () => string) => {
```

lazytype (approved):

website/utils/renderer.ts
brookescarlett: 👍

website/pages/_document.tsx
brookescarlett: yep

PR #765, opened 2021-01-08
Properly renders Head content in Head not Body

brookescarlett:


website/pages/_app.tsx
```diff
@@ -31,8 +31,6 @@ function MyApp({ Component, pageProps }: AppProps) {
     <LeafyGreenProvider>
       <Global styles={globalStyles} />
       <Head>
-        <html lang="en"></html>
```
brookescarlett, line 34: We set lang in _document.tsx now

DesignerDave (approved):

PR #768, opened 2021-01-11
Quick Fixes

brookescarlett:
Addresses two quick fixes raised this morning by Dave: 
- Overline text animates up on landing page
- Dismissible is now correctly spelled

### DIFF
--- a/website/pages/index.tsx
+++ b/website/pages/index.tsx
@@ -30,10 +30,10 @@ const backdrop = css`
 
 const layoutProperties = css`
   ${mq({
-  width: ['calc(100% + 48px)', '100%', '100%', '1077px'],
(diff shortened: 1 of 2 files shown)


website/pages/index.tsx
```diff
@@ -72,6 +72,7 @@ const previewWrapper = css`
   &:hover {
     & > div {
       opacity: 1;
+      transform: translate3d(0, -${spacing[3]}px, 0) scale(0.95);
```
DesignerDave, line 75: We should always make the transitioned state of an element be the default transform values: `translate3d(0, 0, 0) scale(1)`.

brookescarlett: Got it, fixed!

DesignerDave (approved):

PR #773, opened 2021-01-13
PD-1075, PD-1076: Fixes aria attributes in Tabs

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1075](https://jira.mongodb.org/browse/PD-1075)
🎟 _Jira ticket:_ [PD-1076](https://jira.mongodb.org/browse/PD-1076)

## 🛠 Types of changes

## ✅ Checklist


packages/tabs/src/Tabs.tsx
```diff
@@ -163,105 +173,116 @@ function Tabs({
...
-        const [enabledIndexes, current] = getEnabledIndexes();
-        setSelected(enabledIndexes[(current + 1) % enabledIndexes.length]);
-      } else if (e.keyCode === keyMap.ArrowLeft) {
-        const [enabledIndexes, current] = getEnabledIndexes();
-        setSelected(
-          enabledIndexes[
-            (current - 1 + enabledIndexes.length) % enabledIndexes.length
-          ],
-        );
+  }, [childrenArray, selected]);
```
lazytype, line 189: this will still be re-evaluated on every render because `childrenArray` is not memoized

packages/tabs/src/Tabs.tsx
```diff
@@ -163,105 +173,116 @@ function Tabs({
...
-    }
+  const { tabs, tabpanels } = useMemo(() => {
+    const tabs: Array<React.ReactElement> = [];
 
-    return React.cloneElement(child, {
-      key: index,
-      ariaControl: `tab-${index}`,
-      selected: selected === index,
+    const tabpanels = React.Children.map(children, (child, index) => {
+      if (!isComponentType<'Tab'>(child, 'Tab')) {
```
lazytype, line 214: It would be better to remove `<'Tab'>` altogether, since it expects a `ReactElement` not a string

packages/tabs/src/Tabs.tsx
```diff
@@ -163,105 +173,116 @@ function Tabs({
...
+        return child;
+      }
+
+      const { id: idProp, name, disabled, onClick } = child.props;
+      const tabpanelId = idProp ?? tabpanelIdAllocator.generate();
+      const tabId = idProp ?? tabIdAllocator.generate();
```
lazytype, line 220: when `idProp` is defined, both the `Tab` and `TabTitle` will have the same ID, but IDs should be unique, so it seems like `tabpanelId` should always be generated. if `idProp` is not defined, these will generate new ids on every render. why don't `Tab` and `TabTitle` generate their own IDs that they can memoize? can we get the values for `aria-controls` and `aria-labelledby` from `Ref`s?

lazytype (changes requested):

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
brookescarlett: Good catch on the tabpanelId, but I'm not sure that it necessarily makes more sense to have Tab and TabTitle generate their own ids and then relate them to one another through refs. If we're already mapping through the Tabs' children and creating the TabTitles based on those Tab components, we may as well relate them while we do so.

lazytype: alternatively, can `TabTitle` handle generating all the IDs and setting the props on its child `Tab`? or can we create an `InternalTab` component which creates the `TabTitle` and generates the IDs

brookescarlett: TabTitle doesn't know about its child Tab, it's strictly responsible for rendering the button that appears in the tablist, relating the two anywhere but inside of Tabs gets complicated because it is in Tabs that we create the TabTitle from the Tab.

brookescarlett: If I make an InternalTabs or InternalTab component, would it be responsible for doing anything outside of the logic contained from lines 213-259?

lazytype: I would say 218-259. by being it's only component it would be able to use hooks like memo

brookescarlett: ~30 lines of that are defining the TabTitle component 😱 so it seems like the main benefit we'd be getting here is just in memoizing the ids independently of the large memo that they are currently a part of?

brookescarlett: I'm happy to make the change, I just want to make sure the benefits are worth the tradeoff of breaking up the logic and making the file a little less readable

lazytype: theoretically we could clean up the large memo by wrapping `InternalTab` in `React.memo`

packages/tabs/src/InternalTab.tsx
```diff
@@ -0,0 +1,66 @@
...
+    child,
+    selected,
+    as,
+    tabRef,
+    panelRef,
+    ...tabProps
+  }: InternalTabProps) => {
+    const { id: idProp, name } = child.props;
+
+    const panelId = React.useMemo(() => tabpanelIdAllocator.generate(), []);
```
brookescarlett, line 31: @lazytype upon further work with the component, you were totally right that changing this on every render is a bug.

packages/tabs/src/InternalTab.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const tabId = React.useMemo(() => idProp ?? tabIdAllocator.generate(), [
+      idProp,
+    ]);
+
+    const tab = (
+      // @ts-expect-error
+      <TabTitle
+        {...tabProps}
+        as={as}
```
lazytype, line 39: any reason `as` needs to be separate?

packages/tabs/src/InternalTab.tsx
```diff
@@ -0,0 +1,66 @@
...
+      >
+        {name}
+      </TabTitle>
+    );
+
+    const panel = React.cloneElement(child, {
+      id: panelId,
+      ['aria-labelledby']: tabId,
+      selected: selected,
+    });
```
lazytype, line 52: I suppose we could memo this too

lazytype (approved):

packages/tabs/src/InternalTab.tsx
brookescarlett: 👍

packages/tabs/src/InternalTab.tsx
brookescarlett: 👍

PR #774, opened 2021-01-13
Updates image sourcing in website

brookescarlett:
Added images to our S3 bucket, so fetching from there rather than storing locally. Also, remove usage of Next/Image component as it requires bringing in an external image loader in order to host site with Realm instead of Vercel.


website/components/ComingSoon.tsx
```diff
@@ -26,11 +25,12 @@ function ComingSoon() {
       <Subtitle as="p" className={subtitleStyle}>
         Coming soon!
       </Subtitle>
-      <Image
-        src="/images/coming-soon.gif"
+      <img
+        src={`${process.env.NEXT_PUBLIC_CDN}/images/guidelines/coming-soon.gif`}
```
lazytype, line 29: where does `NEXT_PUBLIC_CDN` come from? does it need to be added to the CSP?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src#violation_cases

brookescarlett: 🤦 I should document this somewhere. I added our CDN URL to `.env.local`
Is there a better place to put this?

lazytype: `process.env` is not available in the browser right? would this be a problem if React on the client needs to hydrate or re-render the component?

brookescarlett: Next makes it available in the browser: https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

lazytype: Is there any benefit to making it an environment variable versus a constant exported from a file?

brookescarlett: Just that we don't have to import it, but I suppose the cost is higher than the benefit here. So i can just add the cdn path to our utils folder

lazytype: do you know what's the state of the Content Security Policy?

website/styles/globals.tsx
```diff
@@ -1,45 +1,49 @@
 import { css } from '@emotion/core';
+import CDN from 'utils/cdnURL';
```
lazytype, line 2: it's unusual to import the default with a name different from the name of the file

website/utils/cdnURL.ts
```diff
@@ -0,0 +1,2 @@
+const CDN = 'https://d2va9gm4j17fy9.cloudfront.net';
+export default CDN;
```
lazytype, line 2: I think this could be reduced to `export default 'https://d2va9gm4j17fy9.cloudfront.net'`

brookescarlett: I think now what might make sense is to have a routes.ts with HOME_PAGE and CDN both being named exports from the same file.

website/pages/index.tsx
```diff
@@ -271,7 +267,12 @@ export default function Home({ updates }: { updates: Array<UpdateProps> }) {
             <div className={largeHeight}>
               <MarketingPreview
                 marketingURL="https://www.mongodb.com/blog/post/meet-our-product-design-team-part-1"
+<<<<<<< HEAD
                 backgroundURL={`${landingURL}/team-thumbnail.png`}
+=======
+                backgroundURL="/images/team-thumbnail.png"
+                isTouchDevice={isTouchDevice}
+>>>>>>> main
```
lazytype:
```suggestion
                backgroundURL={`${landingURL}/team-thumbnail.png`}
                isTouchDevice={isTouchDevice}
```

brookescarlett: whoops not sure how that snuck in there.. it's gone now

lazytype (approved):

PR #776, opened 2021-01-13
Add aria-live to Toast

lazytype:
## ✍️ Proposed changes

https://jira.mongodb.org/browse/PD-1079

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/toast/src/Toast.spec.tsx
+++ b/packages/toast/src/Toast.spec.tsx
@@ -15,7 +15,8 @@ describe('packages/toast', () => {
       });
 
       const toast = queryByRole('status');
-      expect(toast).not.toBeInTheDocument();
+      expect(toast).toBeVisible();
+      expect(toast).toBeEmptyDOMElement();
     });
 
     test(`false, Toast doesn't render`, () => {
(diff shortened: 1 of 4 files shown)


packages/toast/src/Toast.spec.tsx
```diff
@@ -15,7 +15,8 @@ describe('packages/toast', () => {
       });
 
       const toast = queryByRole('status');
-      expect(toast).not.toBeInTheDocument();
+      expect(toast).toBeVisible();
+      expect(toast).toBeEmptyDOMElement();
```
brookescarlett, line 19: Good to know about this query

packages/toast/src/Toast.spec.tsx
```diff
@@ -26,7 +27,8 @@ describe('packages/toast', () => {
       });
 
       const toast = queryByRole('status');
-      expect(toast).not.toBeInTheDocument();
+      expect(toast).toBeVisible();
```
brookescarlett, line 30: Do you think we have to register this as more than a patch, since it broke our own tests?

brookescarlett (approved):

PR #790, opened 2021-01-26
Website Bugs

brookescarlett:
- Finally adds Figma links to components
- Adds type for Components
- Creates variables for sizes used on XLarge and Large desktop
- Fixes bug on home page 
- Cleans up mobile

### DIFF
--- a/website/components/CodeDocs.tsx
+++ b/website/components/CodeDocs.tsx
@@ -75,7 +75,7 @@ const changelogStyles = css`
   }
 `;
 
-const maxWidth700 = css`
+const maxWidth = css`
   max-width: ${largeWidth}px;
 `;
 interface VersionCardProps {
@@ -163,7 +163,7 @@ function DesktopInstall({ component, changelog, version }: InstallProps) {
       <GridContainer
         justify="space-between"
         align="flex-start"
-        className={maxWidth700}
+        className={maxWidth}
       >
         <GridItem md={7} lg={7}>
           <div className={topAlignment}>
@@ -232,7 +232,7 @@ function CodeDocs({ component, readme, changelog }: BaseLayoutProps) {
       <GridContainer
         align="flex-start"
         justify="flex-start"
-        className={maxWidth700}
+        className={maxWidth}
       >
         <GridItem sm={12} md={12} xl={12}>
           <Tabs
--- a/website/components/CodeDocs.tsx
+++ b/website/components/CodeDocs.tsx
@@ -223,12 +223,12 @@ function CodeDocs({ component, readme, changelog }: BaseLayoutProps) {
           changelog={changelog}
         />
       ) : (
-          <DesktopInstall
-            component={component}
-            version={version}
-            changelog={changelog}
-          />
-        )}
+        <DesktopInstall
+          component={component}
+          version={version}
+          changelog={changelog}
+        />
+      )}
       <GridContainer
         align="flex-start"
         justify="flex-start"
--- a/website/components/CodeDocs.tsx
+++ b/website/components/CodeDocs.tsx
@@ -14,7 +14,7 @@ import { uiColors } from '@leafygreen-ui/palette';
 import { spacing, breakpoints } from '@leafygreen-ui/tokens';
 import { useViewportSize } from '@leafygreen-ui/hooks';
 import { BaseLayoutProps } from 'utils/types';
-import { largeWidth } from 'utils/styleConstants';
+import { pageContainerWidth } from 'styles/constants';
 import { GridContainer, GridItem } from 'components/Grid';
 import PropTable, { ReadmeMarkdown } from 'components/PropTable';
 import TypeDefinition from 'components/TypeDefinition';
@@ -76,7 +76,7 @@ const changelogStyles = css`
 `;
 
 const maxWidth = css`
-  max-width: ${largeWidth}px;
+  max-width: ${pageContainerWidth.large}px;
 `;
 interface VersionCardProps {
   version?: string;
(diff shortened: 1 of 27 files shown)


website/components/News.tsx
```diff
@@ -52,10 +52,11 @@ function Update({ date, story, route, updateURL }: UpdateProps) {
   useEffect(() => {
     if (typeof navigator !== 'undefined') {
       setDisplayedDate(
-        // @ts-expect-error typescript complaining that dateStyle is not a valid option, but according to Mozilla docs it is: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
-        new Intl.DateTimeFormat(undefined, { dateStyle: 'long' }).format(
-          new Date(date),
-        ),
+        new Date(date).toLocaleString(undefined, {
```
brookescarlett, line 55: The previous implementation was broken on Safari

website/components/CodeDocs.tsx
```diff
@@ -74,6 +75,9 @@ const changelogStyles = css`
   }
 `;
 
+const maxWidth700 = css`
```
lazytype, line 78: the name of this variable will go out of date if the value of `largeWidth` changes

website/components/PropTable.tsx
```diff
@@ -127,13 +134,49 @@ function getTableData(rows: Table['children']): Array<TableDataInterface> {
...
 function PropTable({
   markdownAst,
   component,
 }: {
   markdownAst: ReadmeMarkdown;
   component: string;
 }) {
+  const viewport = useViewportSize();
+  const isTouchDevice = viewport?.width
+    ? viewport?.width < breakpoints.Tablet
```
lazytype, line 177:
```suggestion
  const isTouchDevice = viewport !== null
    ? viewport.width < breakpoints.Tablet
```

website/utils/styleConstants.ts
```diff
@@ -0,0 +1,3 @@
+// Layout Constants
+export const largeWidth = 700;
+export const largestWidth = 1138;
```
lazytype, line 3: the name of this file and these variables don't indicate where these constants should be used

website/utils/figmaLinks.ts
```diff
@@ -0,0 +1,64 @@
+import { Component } from 'utils/types';
+
+// Not every Component has a corresponding Figma link, but want to be sure we're appropriately keeping track of which components do not
+// @ts-expect-error
+export const figmaLinks: Record<Component, string> = {
```
lazytype, line 5: `Partial<Record<Component, string>>` is more accurate

lazytype, line 5: default export?

lazytype (changes requested):

website/components/CodeDocs.tsx
brookescarlett: True

website/components/PropTable.tsx
brookescarlett: 👍

website/utils/styleConstants.ts
brookescarlett: What do you mean by that?

website/utils/figmaLinks.ts
brookescarlett: Yeah, that's fair.

website/utils/styleConstants.ts
lazytype: when I see these I think to myself "largest width of what?"

brookescarlett: Would `largestLayoutWidth` make more sense to you?

brookescarlett: It's referring to the width of the main page container

lazytype: it then raises the question "layout of what?" 🤔

brookescarlett: The amount of space that component pages are allowed to take up, which changes depending on the size of the viewport

brookescarlett: They're just pretty random-seeming numbers that were used in a bunch of places so I thought it'd be easier to document here, but very open to other suggestions

lazytype: >It's referring to the width of the main page container

wrote my last reply before I saw this comment. seeing `mainPageContainerWidth.largest` would be a lot more descriptive.

brookescarlett: its not just the main page, can I do pageContainerWidth.largest

lazytype: two more questions:
- shouldn't this be under the website/styles/ directory
- "large" and "largest" are relative terms. what determines when something is "large"?

brookescarlett: I can put this under styles. I wasnt sure if it should live there or not because its not using emotion directly

brookescarlett: The problem is, the large/largest don't map directly to our breakpoints. They're more values relative to one and other than anything else. When do we allow content to take up a large space versus when do we allow content to take up the largest space. There's a part of the CodeDocs page that employs both at the same breakpoint.

lazytype: from what I can tell, it seems that the general rule is that when there are large visual elements that have multiple columns such as the grid or a table that display structured data, then we use "largest", otherwise we use "larger". does that seem right?

brookescarlett: yes

lazytype: maybe we can call them something like "pageContentWidth.Default" and "pageContentWidth.DataGraphic"

lazytype (approved):

PR #791, opened 2021-01-27
Fix popover examples when not using Portal

lazytype:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/popover/src/Popover.tsx
+++ b/packages/popover/src/Popover.tsx
@@ -20,7 +20,6 @@ import {
 const rootPopoverStyle = css`
   position: absolute;
   transition: transform 150ms ease-in-out, opacity 150ms ease-in-out;
-  transform: translate3d(0, 0, 0) scale(1);
   opacity: 0;
 `;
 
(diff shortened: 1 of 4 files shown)


brookescarlett:
This isn't working for me when I run locally:
Screen Shot 2021-01-27 at 9 08 42 AM

lazytype:
> This isn't working for me when I run locally:
> Screen Shot 2021-01-27 at 9 08 42 AM

Is the right border getting cut-off? what should we expect to see instead?

brookescarlett:
Yes should expect the  border not to be cut off

packages/popover/src/Popover.tsx
```diff
@@ -20,7 +20,6 @@ import {
 const rootPopoverStyle = css`
   position: absolute;
   transition: transform 150ms ease-in-out, opacity 150ms ease-in-out;
-  transform: translate3d(0, 0, 0) scale(1);
```
brookescarlett, line 23: Curious why this was causing the clipping behavior

lazytype: I believe because it was overriding the scale amount of 0.8 defined in positionUtils.ts

brookescarlett (approved):

lazytype:
oh, guess I need a changeset now

PR #793, opened 2021-01-28
Properly spreads properties in Tab

brookescarlett:
## ✍️ Proposed changes

Fixes bug where props were being overwritten rather than cloned in Tab

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/tabs/src/Tabs.tsx
+++ b/packages/tabs/src/Tabs.tsx
@@ -231,25 +231,31 @@ function Tabs({
     }
 
     const isTabSelected = index === selected;
-    const { disabled, onClick } = child.props;
+    const { disabled, onClick, onKeyDown, className, ...rest } = child.props;
 
     const tabProps = {
-      ...child.props,
-      disabled,
       as,
+      disabled,
       darkMode,
       isAnyTabFocused,
-      onKeyDown: handleArrowKeyPress,
-      className: cx({
-        [modeColors[mode].activeStyle]: isTabSelected,
-        [cx(modeColors[mode].disabledColor, disabledStyle)]: disabled,
-      }),
+      className: cx(
+        {
+          [modeColors[mode].activeStyle]: isTabSelected,
+          [cx(modeColors[mode].disabledColor, disabledStyle)]: disabled,
+        },
+        className,
+      ),
+      onKeyDown: (event: KeyboardEvent) => {
+        handleArrowKeyPress(event);
+        onKeyDown?.(event);
+      },
       onClick: !disabled
         ? (event: React.MouseEvent) => {
             onClick?.(event);
             handleChange(event, index);
           }
         : undefined,
+      ...rest,
     };
 
     return (
--- a/packages/tabs/src/Tabs.tsx
+++ b/packages/tabs/src/Tabs.tsx
@@ -234,6 +234,7 @@ function Tabs({
     const { disabled, onClick } = child.props;
 
     const tabProps = {
+      ...child.props,
       disabled,
       as,
       darkMode,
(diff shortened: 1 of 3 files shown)


packages/tabs/src/Tabs.tsx
```diff
@@ -234,6 +234,7 @@ function Tabs({
     const { disabled, onClick } = child.props;
 
     const tabProps = {
+      ...child.props,
       disabled,
```
lazytype, line 238: we can remove `disabled` since it's already in `child.props`

packages/tabs/src/Tabs.tsx
```diff
@@ -234,6 +234,7 @@ function Tabs({
     const { disabled, onClick } = child.props;
 
     const tabProps = {
+      ...child.props,
```
lazytype, line 237: a few things:
- we should preserve `child.props.onKeyDown`
- we should preserve `child.props.className`
- spread should go last so the child can overwrite `as` and `darkMode` if they wish. not sure whether it makes sense for `disabled` to be overwritable

packages/tabs/src/Tabs.tsx
brookescarlett: 👍

packages/tabs/src/Tabs.tsx
```diff
@@ -231,25 +231,31 @@ function Tabs({
...
 
     const isTabSelected = index === selected;
-    const { disabled, onClick } = child.props;
+    const { disabled, onClick, onKeyDown, className, ...rest } = child.props;
 
     const tabProps = {
-      ...child.props,
-      disabled,
       as,
+      disabled,
```
lazytype, line 238:
```suggestion
```

packages/tabs/src/Tabs.tsx
```diff
@@ -231,25 +231,31 @@ function Tabs({
...
+      className: cx(
+        {
+          [modeColors[mode].activeStyle]: isTabSelected,
+          [cx(modeColors[mode].disabledColor, disabledStyle)]: disabled,
+        },
+        className,
+      ),
+      onKeyDown: (event: KeyboardEvent) => {
+        handleArrowKeyPress(event);
+        onKeyDown?.(event);
```
lazytype, line 250: just to be a bit more consistent with the ordering of the callbacks
```suggestion
        onKeyDown?.(event);
        handleArrowKeyPress(event);
```

packages/tabs/src/Tabs.tsx
```diff
@@ -231,25 +231,31 @@ function Tabs({
     }
 
     const isTabSelected = index === selected;
-    const { disabled, onClick } = child.props;
+    const { disabled, onClick, onKeyDown, className, ...rest } = child.props;
```
lazytype, line 234:
```suggestion
    const { onClick, onKeyDown, className, ...rest } = child.props;
```

lazytype (approved):

packages/tabs/src/Tabs.tsx
brookescarlett: I pulled out disabled from props and am now only spreading rest so if I don't pass it explicitly then it won't be passed

packages/tabs/src/Tabs.tsx
brookescarlett: We use the disabled value a couple of times (in className concat and onClick)

brookescarlett: If you feel strongly though, happy to make these changes

lazytype: no worries

PR #794, opened 2021-01-29
PD-1057: Sets ARIA role presentation on the caret in Select

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1057](https://jira.mongodb.org/browse/PD-1057)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/icon/src/glyphCommon.ts
+++ b/packages/icon/src/glyphCommon.ts
@@ -22,7 +22,7 @@ interface AccessibleFunctionParams {
 
 type AccessibleFunctionReturnType =
   | AccessibleFunctionParams
-  | { 'aria-hidden': true };
+  | { 'aria-hidden': true; alt: '' };
 
 export function generateAccessibleProps(
   role: 'img' | 'presentation',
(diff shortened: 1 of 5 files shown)


packages/icon/src/glyphCommon.ts
```diff
@@ -22,7 +22,7 @@ interface AccessibleFunctionParams {
 
 type AccessibleFunctionReturnType =
   | AccessibleFunctionParams
-  | { 'aria-hidden': true };
+  | { 'aria-hidden': true; alt: '' };
```
lazytype, line 25: does this work?
```suggestion
  | { 'aria-hidden': true; alt: undefined };
```

lazytype (approved):

packages/icon/src/glyphCommon.ts
brookescarlett: we want alt to be an empty string

brookescarlett: not for it to not exist at all

PR #796, opened 2021-01-29
Add and update batch 1 icons

DesignerDave:
## ✍️ Proposed changes

Implements the batch 1 icons from the joint project between Design Systems and Visual Design

## 🛠 Types of changes

## ✅ Checklist


packages/typography/src/Link.tsx
```diff
@@ -59,9 +59,10 @@ const arrowRightIconHover = css`
 `;
 
 const openInNewTabStyles = css`
-  margin-bottom: 4px;
-  margin-left: -1px;
-  margin-right: -2px;
+  position: relative;
```
brookescarlett, line 62: If these styles needed to change, are there other styles that we should be cognizant of updating?

.changeset/dirty-carrots-obey.md
```diff
@@ -0,0 +1,44 @@
+---
+'@leafygreen-ui/icon': minor
+'@leafygreen-ui/typography': patch
```
brookescarlett, line 3: Nit: break out into two changesets so the icon information isn't in the typography changelog and vice versa

DesignerDave: Good point! Done

packages/typography/src/Link.tsx
DesignerDave: Nope – this is specifically because previously this specific icon didn't take up the full 16x16 artboard, and now it does. This is the only place we use this glyph. It's possible we should change to a major version for the icon updates, however 🤔

.changeset/dirty-carrots-obey.md
brookescarlett: Thanks!

packages/typography/src/Link.tsx
brookescarlett: Ah got it

brookescarlett (approved):

PR #802, opened 2021-02-02
Add watch command

lazytype:
Running `yarn watch` will build all the packages *and* pass the `--watch` flag to rollup, which will cache its build results and subsequently take much less time to re-build packages. There is currently no way to make rollup persist this cache to disk so that `yarn build` will execute faster (https://github.com/rollup/rollup/issues/2182), so this is a light-weight alternative (relative to Bazel-ifying everything). 

Some potential follow-up tasks that could speed up the average build process:
- watch the `packages/icon/src/glyphs` directory and run `yarn pre-build` when it changes
  - even further, make `packages/icon/scripts/build.ts` only build for new/removed/modified .svg files
- run `yarn ts:build` when TS files change

## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


package.json
```diff
@@ -25,6 +25,7 @@
     "start": "start-storybook -p 9001 -c storybook -s ./storybook/static --no-version-updates",
     "pre-build": "lerna run --parallel pre-build -- --env production",
     "build": "yarn pre-build && lerna run --parallel build",
+    "watch": "yarn build -- -- --watch",
```
brookescarlett, line 28: Seeing this warning locally: 

```warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.```

lazytype: In the "future version" we'd have only one `--`, but right now we need two, once which yarn ignores (and provides a warning for) and one that gets forwarded.

brookescarlett (approved):

PR #809, opened 2021-02-04
PD-1100: Fix Table SSR issue

DesignerDave:
## ✍️ Proposed changes

There were a few issues causing the Table Rows to not render. This PR fixes those issues.

🎟 _Jira ticket:_ [Table <tbody> contents not rendering properly during SSR](https://jira.mongodb.org/browse/PD-1100)

## 🛠 Types of changes

## ✅ Checklist


packages/table/src/Row.tsx
```diff
@@ -242,17 +244,13 @@ const Row = React.forwardRef(
       return renderedChildren;
     }, [children, rowHasNestedRows, isExpanded, setIsExpanded]);
 
-    // Depending on network speed, will noticeably render columns with incorrect
-    // alignment, would rather wait for proper information before rendering
-    if (!columnInfo) {
```
DesignerDave, line 247: Calling this out pre-emptively.

I removed this block because it was preventing rendering server-side. I tried disabling this block on the server, but it causes a mismatch between the client and server upon hydration.

packages/table/src/Row.tsx
```diff
@@ -117,6 +118,7 @@ const Row = React.forwardRef(
     }: RowProps,
     ref: React.Ref<any>,
   ) => {
+    const { isBrowser } = useSSR();
```
lazytype, line 121: I see this causing problems with SSR testing. The reason is because `useSSR`'s implementation checks `window`, `document`, etc. only once during module load time and then stores the result. This would interact badly with SSR tests because they rely on first mocking `window` etc. to be undefined to simulate server render and then unmocks them to hydrate. Unmocking `window` etc. would have no effect because `useSSR` has already stored the result of the check.

packages/table/src/Row.tsx
brookescarlett: Is there another work around? For tables with large data sets this lag is very visible.

brookescarlett: Also interesting because this never caused a problem with the Table component rendered in our Next app

DesignerDave: In regards to another workaround, I tried a few things, but ultimately the solutions I tried ended up creating a mismatch in the server rendered HTML and the initial hydrated DOM. Leaving this here causes the content to not render at all on the server. Right now, the only thing that changes is alignment, which isn't perfect, but it's better than having the component not work with SSR.

DesignerDave: In regards to your second comment, it is interesting, though I reproduced this with Gatsby, not Next since the docs team uses Gatsby.

packages/table/src/Row.tsx
DesignerDave: Is there a way to update the tests to be more effective in this regard? If this interacts poorly with the tests, then the tests aren't effectively emulating the environments.

packages/table/src/Row.tsx
brookescarlett: Should we consider using the hook here? 

```
if (isBrowser && !columnInfo) {
}
```

packages/table/src/Row.tsx
lazytype: the most straightforward approach I can think of is to try to mock the `use-ssr` module in tests

to truly simulate the environments and avoid these kinds of problems we would need to have a separate server environment node process that the test process could communicate with in order to avoid the two environments from sharing the same set of loaded modules

DesignerDave: Wouldn't mocking `use-ssr` possibly defeat the purpose of our SSR tests?

lazytype: if the purpose of SSR tests is to test the behavior of components rendered in an environment without access to window, document, or DOM apis and the subsequent hydration of those components, then mocking `use-ssr` to produce the behavior it should have in each environment would bring us closer to that purpose

brookescarlett: Perhaps instead of using useSSR we just check if `typeof window !== undefined` like we do elsewhere around the repo, which would mean we don't have to update the jest config and would be more consistent with how we usually handle this same problem

DesignerDave: I'm probably gonna take a stab at mocking this. I think that useSSR is a slightly more robust solution to this sort of problem, and I'd like to use it more, not less.

packages/table/src/Row.tsx
DesignerDave: I tried exactly that, but that also had the issue of creating a mismatch between the server and client-rendered HTML.

packages/table/src/Row.tsx
DesignerDave: Turned out we've actually solved this _exact_ issue before with Emotion! The actual fix to make this work was extremely simple, and I've tested that it detects the environment how we expected it to.

packages/table/src/Row.tsx
brookescarlett: how come?

scripts/jestSSRSetup.js
```diff
@@ -46,6 +46,8 @@ Context.within(new ArtificialServerContext(), () => {
   // Emotion detects whether the document exists at module load time >:(
   // so we have to preload it here so that it doesn't load in a non-server context
   require('@emotion/cache');
+
+  require('use-ssr')
```
brookescarlett, line 50: Do you mind leaving a comment that explains what's going on here and why we're requiring it

DesignerDave: I might just change the comment above to explain both 😄

DesignerDave: Done

brookescarlett (approved):

PR #814, opened 2021-02-05
Fixes and QOL improvements for Code

DesignerDave:
## ✍️ Proposed changes

This PR fixes a few small issues, makes a few quality of life improvements for folks using the components, and makes a couple numbers less magical.

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/.changeset/sweet-suns-breathe.md
+++ b/.changeset/sweet-suns-breathe.md
@@ -1,0 +1,6 @@
+---
+'@leafygreen-ui/code': patch
+---
+
+- Fixes issue where single-line code examples would break when set to wrap.
+- Fixes issue where code examples would be detected as multi-line when a single line example has preceding or subsequent new line characters.
--- a/.changeset/sweet-suns-breathe.md
+++ b/.changeset/sweet-suns-breathe.md
@@ -3,4 +3,5 @@
 ---
 
 - Fixes issue where single-line code examples would break when set to wrap.
+- Fixes issue where many applications had global styles for pre tags that added a second border and border radius to the pre element.
 - Fixes issue where code examples would be detected as multi-line when a single line example has preceding or subsequent new line characters.
--- a/.changeset/sweet-suns-breathe.md
+++ b/.changeset/sweet-suns-breathe.md
@@ -2,6 +2,5 @@
 '@leafygreen-ui/code': patch
 ---
 
-- Fixes issue where single-line code examples would break when set to wrap.
 - Fixes issue where many applications had global styles for pre tags that added a second border and border radius to the pre element.
 - Fixes issue where code examples would be detected as multi-line when a single line example has preceding or subsequent new line characters.
--- a/packages/code/src/Code.tsx
+++ b/packages/code/src/Code.tsx
@@ -1,26 +1,29 @@
 import React, { useRef, useEffect, useState, useMemo } from 'react';
 import PropTypes from 'prop-types';
+import ClipboardJS from 'clipboard';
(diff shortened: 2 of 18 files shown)


packages/code/src/Code.tsx
```diff
@@ -1,29 +1,35 @@
...
+import { Language, CodeProps, Mode } from './types';
+import Syntax from './Syntax';
 import WindowChrome from './WindowChrome';
-import debounce from 'lodash/debounce';
-import { uiColors } from '@leafygreen-ui/palette';
-import ClipboardJS from 'clipboard';
-import facepaint from 'facepaint';
 
 // We use max-device-width to select specifically for iOS devices
 const mq = facepaint([
```
brookescarlett, line 20: If outside the scope of this PR, we should make a ticket to update this to use our breakpoints token

DesignerDave: I did a lot of testing for this back when I made this update, and found that the breakpoints tokens didn't target the right devices for what we needed to do here.

packages/code/src/Code.tsx
```diff
@@ -229,7 +230,7 @@ function Code({
   const [copied, setCopied] = useState(false);
   const [showCopyBar, setShowCopyBar] = useState(false);
   const mode = darkMode ? Mode.Dark : Mode.Light;
-  const isMultiline = useMemo(() => children.includes('\n'), [children]);
+  const isMultiline = useMemo(() => children.trim().includes('\n'), [children]);
```
brookescarlett, line 233: Should we add a test to prove that this works as expected and ensure that we don't introduce any regressions here?

.changeset/sweet-suns-breathe.md
```diff
@@ -0,0 +1,6 @@
+---
```
lazytype, line 1: while we're touching this package could we remove the `console.log` that's in `packages/code/src/renderingPlugin.spec.tsx`

DesignerDave: Nice catch, done.

packages/code/src/Code.tsx
DesignerDave: I can add something small to test this. To be honest, there's not a lot to test here though.

brookescarlett: I think this `Fixes issue where code examples would be detected as multi-line when a single line example has preceding or subsequent new line characters.` could be testable

DesignerDave: I can test that this bit of code checks for that effectively, but testing that the end result is correct is a bit tricky since the only change is CSS.

brookescarlett: can you check the number of lines rendered when that character is appended/prepended

DesignerDave: This code doesn't change the lines rendered at all.

brookescarlett: I thought that was what was called out in the changset

DesignerDave: To clarify, this boolean only changes CSS – there's no change to the DOM to test.

DesignerDave: No – we just weren't applying single line CSS styles when the passed code example had preceding, or subsequent line breaks despite us stripping those lines in the plugin. To be clear, that bit is already tested I believe.

brookescarlett: Got it, think I misunderstood

brookescarlett (approved):

PR #821, opened 2021-02-08
Fix issue with sublanguages

DesignerDave:
## ✍️ Proposed changes

Fixes an issue where sublanguages wouldn't render.

## 🛠 Types of changes

## ✅ Checklist


packages/code/src/renderingPlugin.tsx
```diff
@@ -54,6 +54,18 @@ function isNumber(item: any): item is number {
   return item != null && typeof item === 'number';
 }
 
+function isTokenObject(item: any): item is TokenObject {
+  // kind: string;
```
brookescarlett, line 58: rm?

packages/code/src/renderingPlugin.tsx
```diff
@@ -228,6 +244,10 @@ export function flattenNestedTree(
       ];
     }
 
+    if (isTokenObject(val)) {
+      return [...acc, ...flattenNestedTree(val)];
+    }
+
     return acc;
```
brookescarlett, line 251: We might consider replacing this return acc with `enforceExhaustive(val)` to make sure we're handling all possible cases of val (which I think is what we want to do)

packages/code/src/renderingPlugin.tsx
```diff
@@ -54,6 +54,18 @@ function isNumber(item: any): item is number {
   return item != null && typeof item === 'number';
 }
 
+function isTokenObject(item: any): item is TokenObject {
```
brookescarlett, line 57: should we type this as unknown rather than any?

packages/code/src/highlight.d.ts
```diff
@@ -1,6 +1,7 @@
 interface TokenObject {
   kind: string;
   children: Array<string | TokenObject>;
+  sublanguage?: boolean;
```
brookescarlett, line 4: where do we use this property?

packages/code/src/renderingPlugin.tsx
```diff
@@ -54,6 +54,18 @@ function isNumber(item: any): item is number {
...
+  // kind: string;
+  // children: Array<string | TokenObject>;
+  // sublanguage?: boolean;
+
+  if (item == null || typeof item !== 'object') {
+    return false;
+  }
+
+  return typeof item?.kind === 'string' && item?.children instanceof Array;
```
brookescarlett, line 66: why can't we first assert that item isn't null so that we can do item.kind and item.children

packages/code/src/renderingPlugin.tsx
DesignerDave: We're more likely to not be handling specific data structures that we're unaware of than generic types. This PR is a good example of that.

packages/code/src/renderingPlugin.tsx
DesignerDave: If that's typed as unknown, then TS thinks the `kind` and `children` properties of the object don't exist.

packages/code/src/highlight.d.ts
DesignerDave: We don't, but it can exist on the object. highlight.d.ts is an attempt to accurately type the highlight.js library (or at least, the parts we've come across).

packages/code/src/renderingPlugin.tsx
brookescarlett: perhaps `unknown | TokenObject`

packages/code/src/renderingPlugin.tsx
DesignerDave: That's a good point, we already do that so the optional chaining here is unnecessary.

packages/code/src/renderingPlugin.tsx
brookescarlett: If you test enforceExhaustive(val) here, it seems that we've caught all the possible values. if you delete lines 247-249, enforceExhaustive will error -- which is why I'm suggesting

packages/code/src/renderingPlugin.tsx
DesignerDave: Oops 😅

packages/code/src/highlight.d.ts
brookescarlett: 👍

packages/code/src/renderingPlugin.tsx
DesignerDave: The problem here is that it assumes we have our typing right for what val is (which, though I'm more confident after this PR that we do, but it's still possible). If you look at our main branch and do the same thing, it should also pass.

brookescarlett: Got it

DesignerDave: Since I don't have the same level of confidence here in the type correctness, I feel strongly we still need to `return acc` here to prevent this code from throwing at runtime. If we add enforceExhaustive, it marks `return acc` as unreachable code, which may not be true, and increases the chance someone removes that code in the future.

packages/code/src/renderingPlugin.tsx
DesignerDave: What's the end result you're hoping changing this type will achieve?

brookescarlett: Just being more precise with our typings that's all

DesignerDave: `unknown | TokenObject` has the same issue. Using any here is fine I think. I'd want to be more specific if it were a return type or something similar.

packages/code/src/renderingPlugin.tsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
+import { transparentize } from 'polished';
+import { enforceExhaustive } from '@leafygreen-ui/lib';
```
brookescarlett, line 3: don't need this important anymore

brookescarlett (approved):

PR #822, opened 2021-02-09
Adds useAccessibleFormField hook

brookescarlett:
hook that generates input/label props to accessibly associate the two elements

### DIFF
--- a/.changeset/forty-dodos-dance.md
+++ b/.changeset/forty-dodos-dance.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/hooks': minor
+---
+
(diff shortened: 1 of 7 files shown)


DesignerDave (commented):
Unless I'm missing something, we aren't using this anywhere.

brookescarlett:
> Unless I'm missing something, we aren't using this anywhere.

It handles the same bit of logic that we use across form elements. Didn't want to make this PR unnecessarily big, but will use in TextInput, TextArea, Checkbox, etc. when we have this in. Also, will help with components such as Toggle where we don't ship it with a label, so you can grab the proper label/input props from this hook when implementing.

PR #823, opened 2021-02-09
PD-1006: Adds fill size to select component, so that menu button can take up t…

brookescarlett:
…he width of its container

- We have size variants of the Select component, so instead of adding a net new prop to control the width of the component, I added a `fill` option to the Size prop. Now, if a consumer sets width on the className passed to the select component, the component will take up the width of that container (which I believe is the desired behavior outlined in this ticket)

## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1006](https://jira.mongodb.org/browse/PD-1006)

## 🛠 Types of changes

## ✅ Checklist


packages/select/src/styleSets.ts
```diff
@@ -224,6 +225,22 @@ export const sizeSets: Record<Size, SizeSet> = {
       text: 16,
     },
   },
+  [Size.Fill]: {
```
DesignerDave, line 228: I see a bit of a problem with using the same prop as we use to control the height of the input. I'm also not sure this was the original intent of the ticket. 

If we want to make the input's width more easily controlled, we could consider changing instances of `width` to `min-width`, or removing that altogether. I'd suggest approaching this specific problem the same way we approach TextInput, since the contexts they're used are usually similar.

DesignerDave (changes requested):

.changeset/grumpy-news-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/select': minor
```
DesignerDave, line 2: This is conservatively a breaking change.

.changeset/grumpy-news-sell.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Consuming applications can now set the width of the select dropdown by passing a value to width through the `className` prop.
```
DesignerDave, line 5: This was possible before. This just makes the select fill the available space.

I'm a tad concerned that this PR doesn't address PD-1006 at all...

DesignerDave (changes requested):

brookescarlett:
@DesignerDave might be missing something, but I believe this ticket is a response to several comments coming through about a desire to change the width of the select component. Previously this was explicitly not possible, as setting the width through className did not change the width of the Select dropdown itself, rather it only applied to the outer div. This was specifically called out as a desire from @audroue as something he'd like to see possible in mongodb.design

DesignerDave:
So previously, if you set `width: 300px` on the component, you could override the width. By setting min-width, you may actually be making it harder to explicitly set width.

PD-1006 is to add size variants (think like we have with our Button component.

brookescarlett:
If you set width: 300px, nothing would change... you would be changing the width of the outer div but not the menu button/dropdown itself.

We have size variants (as we have in button) in this component already

DesignerDave (approved):

DesignerDave (approved):

PR #830, opened 2021-02-12
Creates a11y package

brookescarlett:
Initial release of a11y package

### DIFF
--- a/packages/a11y/src/A11y.spec.tsx
+++ b/packages/a11y/src/A11y.spec.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
 import { render } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
 import { axe } from 'jest-axe';
-import { VisuallyHidden } from '.';
+import { VisuallyHidden, useAccessibleForm } from '.';
 
 describe('packages/a11y', () => {
   describe('VisuallyHidden', () => {
@@ -11,6 +12,40 @@ describe('packages/a11y', () => {
       );
       const results = await axe(container);
       expect(results).toHaveNoViolations();
+    });
+  });
+
+  describe('useAccessibleFormField', () => {
+    test('when id is supplied to hook', () => {
+      const inputId = 'text-input-id';
+      const { result } = renderHook(
+        (props: string) => useAccessibleForm(props),
+        { initialProps: inputId },
+      );
+
+      expect(result.current.inputProps).toEqual({
+        id: inputId,
+        ['aria-labelledby']: 'label-0',
+      });
+      expect(result.current.labelProps).toEqual({
+        id: 'label-0',
+        htmlFor: inputId,
+      });
+    });
+
+    test('when no id is supplied to hook', () => {
+      const { result } = renderHook((props: string) =>
+        useAccessibleForm(props),
+      );
+
+      expect(result.current.inputProps).toEqual({
+        id: 'input-0',
+        ['aria-labelledby']: 'label-1',
+      });
+      expect(result.current.labelProps).toEqual({
+        id: 'label-1',
+        htmlFor: 'input-0',
+      });
     });
   });
 });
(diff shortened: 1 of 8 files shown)


packages/a11y/src/useAccessibleForm.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const inputIdAllocator = IdAllocator.create('input');
+const labelIdAllocator = IdAllocator.create('label');
+
+/**
+ * Hook that returns two sets of props that accessibly associate a label and its respective input element
+ * @param id Describes the input element
+ */
+const useAccessibleForm = (idProp: string) => {
+  const inputId = useMemo(() => idProp ?? inputIdAllocator.generate(), [
```
DesignerDave, line 12: On the off chance someone changes the idProp, do we want to update this value? I'd imagine so.

packages/a11y/src/useAccessibleForm.ts
```diff
@@ -0,0 +1,30 @@
...
+    idProp,
+  ]);
+  const labelId = useMemo(() => labelIdAllocator.generate(), []);
```
DesignerDave, line 15: Micro-optimization, but can't we simply do:

```js
const labelId = useMemo(labelIdAllocator.generate, [])
```

packages/a11y/src/A11y.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+      });
+    });
+
+    test('when no id is supplied to hook', () => {
+      const { result } = renderHook((props: string) =>
+        useAccessibleForm(props),
+      );
+
+      expect(result.current.inputProps).toEqual({
+        id: 'input-0',
```
DesignerDave, line 42: Is there a way we could change this test to be less dependent on implementation details of the idAllocator? Potentially just checking that the ID here matches the label's `for` is enough.

packages/a11y/src/VisuallyHidden.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <div {...rest} className={cx(visuallyHidden, className)}>
+      {children}
+    </div>
+  );
+}
+
+VisuallyHidden.displayName = 'VisuallyHidden';
+
+VisuallyHidden.propTypes = {
+  children: PropTypes.string,
```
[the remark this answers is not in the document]
brookescarlett: good catch, thanks

packages/a11y/src/VisuallyHidden.tsx
```diff
@@ -0,0 +1,35 @@
...
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+`;
+
+function VisuallyHidden({
+  children,
+  className,
+  ...rest
+}: HTMLElementProps<'div'>) {
```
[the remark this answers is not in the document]
brookescarlett: I don't think that there's a better element to use here, but if you want I can wrap Box and we can use the as prop.. up to you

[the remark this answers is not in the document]
DesignerDave: I don't feel strongly.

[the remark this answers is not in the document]
DesignerDave: Is there a case where you'd want a screen reader to read this as a label or header? I don't think that's what the component is for, right?

packages/a11y/src/useAccessibleForm.ts
brookescarlett: Yes we would want to update this value, which is the current behavior here. Just added tests to confirm.

packages/a11y/src/useAccessibleForm.ts
brookescarlett: we get a warning from eslint if we don't pass an inline function

packages/a11y/src/A11y.spec.tsx
brookescarlett: yep, updated

packages/a11y/src/VisuallyHidden.tsx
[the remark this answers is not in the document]
brookescarlett: Yeah, I think leaving as a div is fine for now. If a use-case comes up that proves otherwise, happy to make the change

DesignerDave (approved):

PR #832, opened 2021-02-16
PD-443: Traps focus in modal component

brookescarlett:
## ✍️ Proposed changes

Traps focus in Modal component

🎟 _Jira ticket:_ [PD-443](https://jira.mongodb.org/browse/PD-443)

## 🛠 Types of changes

## ✅ Checklist


DesignerDave (commented):
Overall looks good! Only note is that I'm not sure we should be focusing on the close button first when the modal opens. In testing, it doesn't matter if there's other focusable content in the modal, the button is always first in the tabindex.

I imagine the options here are:
- Remove the button from the tabindex. We allow closing via the escape key anyways, and usually modals have a button at the bottom to close as well.
- Move the button to the end of the modal's tabindex. This won't be super intuitive for sighted users navigating via keyboard, but could be fine for screen reader users. The caveat there is if we have a "cancel" button at the bottom, there will be two close buttons in a row in the tabindex.

packages/modal/README.md
```diff
@@ -78,13 +78,14 @@ function ExampleComponent() {
...
+| `contentClassName`     | `string`                                                                                 | Style to be applied to the content div.                                                                                                                                                            |              |
+| `closeOnBackdropClick` | `boolean`                                                                                | Determines whether or not a Modal should close when a user clicks outside the modal.                                                                                                               | `true`       |
+| `initialFocus`         | `string` (Selector that can be passed to document.querySelector() to find the DOM Node.) | By default, when a focus trap is activated the first element in the focus trap's tab order will receive focus. With this option you can specify a different element to receive that initial focus. |              |
```
DesignerDave, line 91: I'd make this bit of documentation a bit less specific to the implementation. I would not refer to the focus trap, and instead say something like: 
> A selector string for the element to move focus to when the modal is opened. The first focusable element in the modal will receive focus by default.

Then you don't need additional detail in the `type` column as well.

.changeset/gentle-tomatoes-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/modal': patch
```
DesignerDave, line 2: I'm inclined to say this is breaking since it affects any existing implementations of initial focus consumers may be using.

DesignerDave (approved):

packages/modal/README.md
brookescarlett: Sounds good, thanks

PR #834, opened 2021-02-22
PD-1136: Adds border to project select component

brookescarlett:
Screen Shot 2021-02-22 at 4 49 25 PM

## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1136](https://jira.mongodb.org/browse/PD-1136)

cc: @nellem23

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/mongo-nav/src/mongo-select/OrgSelect.tsx
+++ b/packages/mongo-nav/src/mongo-select/OrgSelect.tsx
@@ -57,10 +57,8 @@ import {
 const orgButtonStyle = css`
   justify-content: space-between;
   border-radius: 5px 0 0 5px;
-  border: 1px solid ${uiColors.gray.light2};
   width: 180px;
   height: 30px;
-  padding: 3px 5px;
 `;
 
 const orgStandaloneButtonStyle = css`
--- a/.changeset/fifty-poems-nail.md
+++ b/.changeset/fifty-poems-nail.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/mongo-nav': patch
+---
+
+Adds border to ProjectSelect component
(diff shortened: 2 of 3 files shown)


packages/mongo-nav/src/mongo-select/OrgSelect.tsx
```diff
@@ -57,10 +57,8 @@ import {
 const orgButtonStyle = css`
   justify-content: space-between;
   border-radius: 5px 0 0 5px;
-  border: 1px solid ${uiColors.gray.light2};
```
DesignerDave, line 60: Am I missing something, or did we only remove a border, not add one? 🤔

DesignerDave (changes requested):

packages/mongo-nav/src/mongo-select/OrgSelect.tsx
brookescarlett: this is a dupe from here: https://github.com/mongodb/leafygreen-ui/blob/main/packages/mongo-nav/src/mongo-select/styles.ts#L45, and the line that I removed from the project select file was simply overwriting this one.

brookescarlett: if you run locally, you should see both borders as in the screenshot attached

nellem23:
@brookescarlett @DesignerDave 

So I think we need to step back and talk about the desired end result:

For Ops Manager we'd like to have a border on the Project Dropdown. This is because the element is pulled out into the side nav and right now is floating in space and looks weird.

For everything else (Cloud, Atlas, Realm, Charts, Cloud Manager) we want to keep the Project Nav as is and *not* include a border on the Project Dropdown.

nellem23:
OpsManager@2x
Atlas@2x

brookescarlett:
@nellem23 if that's the case then we currently support passing a className to the component itself, so where you want to add a border that should just work

nellem23:
@brookescarlett if I understand correctly then, it is an implementation detail for the engineering team?

brookescarlett:
@nellem23 yep

PR #836, opened 2021-02-24
PD-1198: Initial Ripple Package

brookescarlett:
## ✍️ Proposed changes

Initial release of Ripple package

🎟 _Jira ticket:_ [PD-1198](https://jira.mongodb.org/browse/PD-1198)

## 🛠 Types of changes

## ✅ Checklist


packages/ripple/README.md
```diff
@@ -0,0 +1,43 @@
...
+yarn add @leafygreen-ui/ripple
+```
+
+### NPM
+
+```shell
+npm install @leafygreen-ui/ripple
+```
+
+## Example
```
hswolff, line 21: Probably should show vanilla JS example too!

packages/ripple/src/getRippleGlobalNamespace.ts
```diff
@@ -0,0 +1,37 @@
...
+
+type LGWindow = Window &
+  typeof globalThis & {
+    __LEAFYGREEN_UTILS__: {
+      modules: ModuleType;
+    };
+  };
+
+export function getRippleGlobalNamespace() {
+  if (!window) {
```
hswolff, line 17: Are we worrying about server side rendering at this point?

hswolff: It'd be nice to have our own internal `globalThis` shim that we could rely on across all packages.

packages/ripple/src/getRippleGlobalNamespace.ts
```diff
@@ -0,0 +1,37 @@
...
+      setRippleListener: false,
+      registeredRippleElements: {},
+    },
+  };
+
+  if ((window as LGWindow).__LEAFYGREEN_UTILS__) {
+    (window as LGWindow).__LEAFYGREEN_UTILS__.modules = rippleModule;
+  }
+
+  (window as LGWindow).__LEAFYGREEN_UTILS__ = { modules: rippleModule };
```
hswolff, line 32: I'm confused how this last line won't just always overwrite the previous lines.

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    from {
+      opacity:1;
+    }
+    to {
+      transform: scale(2);
+      opacity: 0;
+    }
+  }
+
+  .ripple {
```
hswolff, line 117: Should we namespace these? 
```suggestion
  .lgui-ripple {
```

hswolff: Ditto the animation name.

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    createRippleEffect(event);
+  }
+}
+
+export function registerRipple(node: HTMLElement, options: Options) {
+  if (!RIPPLE_NAMESPACE) {
+    return;
+  }
+
+  globalOptions = options;
```
hswolff, line 28: Seems less then great that you can only change global options when registering an element.

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    const styles = document.createElement('style');
+    styles.innerHTML = staticRippleStyles;
+    document.head.append(styles);
+
+    RIPPLE_NAMESPACE.setRippleListener = true;
+  }
+}
+
+export function unregisterRipple(node: HTMLElement) {
```
hswolff, line 44: Alternatively can just return this function from `registerRipple` so people can easily remove them. Similar to useEffect behavior, for easy cleanup.

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  if (
+    // @ts-expect-error using HTMLElements to index as it provides a faster lookup when deciding if we should create a ripple effect on a given element
+    RIPPLE_NAMESPACE.registeredRippleElements[event.target]
+  ) {
+    createRippleEffect(event);
+  }
+}
+
+export function registerRipple(node: HTMLElement, options: Options) {
+  if (!RIPPLE_NAMESPACE) {
```
hswolff, line 24: This is implicitly SSR safe due to this. Should we be more explicit about handling SSR?

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    return;
+  }
+
+  globalOptions = options;
+
+  // @ts-expect-error using HTMLElements to index as it provides a faster lookup when deciding if we should create a ripple effect on a given element
+  RIPPLE_NAMESPACE.registeredRippleElements[node] = options;
+
+  if (!RIPPLE_NAMESPACE.setRippleListener) {
+    document.addEventListener('click', rippleEvent, { passive: true });
```
hswolff, line 34: Should we do capture mode so we get first dibs and can't be prevented?

Probably not, but wondering out loud.

hswolff: Should we support a way to remove this event listener?

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    return;
+  }
+
+  const rect = target.getBoundingClientRect();
+  const ripple = document.createElement('span');
+
+  ripple.className = 'ripple';
+  ripple.style.height = ripple.style.width =
+    Math.max(rect.width, rect.height) + 'px';
+  target.appendChild(ripple);
```
hswolff, line 69: Shame we can't move this to after we change the style to reduce a reflow.

packages/ripple/README.md
brookescarlett: Makes sense, will add.

packages/ripple/src/index.ts
brookescarlett: yep will update

packages/ripple/src/index.ts
brookescarlett: You can update the global options by calling register element inside of a hook that has variant, etc. as dependents

packages/ripple/src/index.ts
brookescarlett: 👍

packages/ripple/src/index.ts
brookescarlett: I checked and this does work in our Next app, but added a comment to explain that it was explicitly for handling SSR

packages/ripple/src/getRippleGlobalNamespace.ts
brookescarlett: updated to check for if window is not undefined, because we are worrying about ssr

packages/ripple/src/getRippleGlobalNamespace.ts
brookescarlett: Yep, fixed this

packages/ripple/src/index.ts
brookescarlett: Alternatively, i can expose `removeRippleEvent` if you think that makes more sense

common.tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "pretty": true,
     "strictNullChecks": true,
     "noUnusedLocals": true,
+    "esModuleInterop": true,
```
hswolff, line 9: What prompted the need for this?

packages/ripple/src/getRippleGlobalNamespace.ts
```diff
@@ -0,0 +1,43 @@
...
+    return;
+  }
+
+  if ((window as LGWindow).__LEAFYGREEN_UTILS__) {
+    (window as LGWindow).__LEAFYGREEN_UTILS__.modules[
+      '@leafygreen-ui/ripple'
+    ] = {
+      setRippleListener: false,
+      registeredRippleElements: {},
+    };
```
hswolff, line 28: Can definitely DRY this out.

packages/ripple/src/index.ts
hswolff: Right, but you need to include a `node` to modify options.

packages/ripple/src/index.ts
hswolff: Hard to predict the use case here. If we did it in the returned function we'd only be able to clean up if it was the last ripple element being removed, which takes some cost to calculate.

Maybe let's wait and see if people request it. Curious if we'll have a need for it in testing?

packages/ripple/src/getRippleGlobalNamespace.ts
brookescarlett: Also: The reason I didn't make this more global and include it in lib was because we didn't want React to be a dep of this package (and React is a dep of lib). I can make another external package that handles interacting with our global namespace in a more abstract/reusable way, but figured I'd do that once we needed it again. What are your thoughts?

common.tsconfig.json
brookescarlett: Was seeing this: 

```
import MongoNav
Cannot use JSX unless the '--jsx' flag is provided.ts(17004)
```

All across the repo

packages/ripple/src/getRippleGlobalNamespace.ts
hswolff: Agreed on all thoughts!

brookescarlett: SGTM

packages/ripple/src/index.ts
brookescarlett: Updated the entire implementation around this, but we would need a node to modify the options otherwise we won't know whose options to modify. The logic around this is a lot cleaner now

packages/ripple/src/index.ts
brookescarlett: SGTM

packages/ripple/src/getRippleGlobalNamespace.ts
brookescarlett: Updated

packages/ripple/src/getRippleGlobalNamespace.ts
```diff
@@ -0,0 +1,43 @@
...
+    (window as LGWindow).__LEAFYGREEN_UTILS__.modules[
+      '@leafygreen-ui/ripple'
+    ] = defaultRippleParams;
+  } else {
+    const rippleModule = {
+      '@leafygreen-ui/ripple': defaultRippleParams,
+    };
+
+    (window as LGWindow).__LEAFYGREEN_UTILS__ = { modules: rippleModule };
+  }
```
hswolff, line 38: Fully optional, just thinking aloud, using https://github.com/tc39/proposal-logical-assignment let's us have some fun:

```suggestion
  (window as LGWindow).__LEAFYGREEN_UTILS__ ??= { modules: {} };
  (window as LGWindow).__LEAFYGREEN_UTILS__.modules[
      '@leafygreen-ui/ripple'
    ] = defaultRippleParams;
```

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    document.addEventListener('click', rippleEvent, { passive: true });
+
+    const styles = document.createElement('style');
+    styles.innerHTML = staticRippleStyles;
+    document.head.append(styles);
+
+    RIPPLE_NAMESPACE.setRippleListener = true;
+  }
+
+  return unregisterRipple;
```
hswolff, line 32: We can make this an anonymous function that captures the reference to the node such that an application only has to call the function - they don't have to pass in the node itself. Makes for tight coupling of the unregister method.

hswolff (dismissed):
We need tests!!!

packages/ripple/src/index.ts
brookescarlett: 👍  makes sense to me

packages/ripple/src/getRippleGlobalNamespace.ts
brookescarlett: So much cleaner

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,119 @@
+import { getRippleGlobalNamespace } from './getRippleGlobalNamespace';
```
DesignerDave, line 1: We usually don't write the code within `index.ts` to make the file more findable. Can we move this code to a separate file or files like `registerRipple.ts`?

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+    document.addEventListener('click', rippleEvent, { passive: true });
+
+    const styles = document.createElement('style');
+    styles.innerHTML = staticRippleStyles;
+    document.head.append(styles);
+
+    RIPPLE_NAMESPACE.setRippleListener = true;
+  }
+
+  return () => RIPPLE_NAMESPACE.registeredRippleElements.delete(node);
```
DesignerDave, line 32: Nice 👌

packages/ripple/README.md
```diff
@@ -0,0 +1,54 @@
+# Ripple
```
DesignerDave, line 1: We need documentation somewhere that the element needs `overflow: hidden;`

packages/ripple/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function createRippleEffect(event: MouseEvent) {
+  const target = event.target as HTMLElement | null;
+  const foundNode = RIPPLE_NAMESPACE?.registeredRippleElements.get(
```
DesignerDave, line 37: This is neat. Hows the performance vs the object lookup we were considering?

DesignerDave (changes requested):

packages/ripple/src/index.ts
brookescarlett: 👍

packages/ripple/src/getRippleGlobalNamespace.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type SetWindow = Window &
+  typeof globalThis & {
+    __LEAFYGREEN_UTILS__: {
+      modules: {};
+    };
+  };
+
+export type LGWindow = Window &
```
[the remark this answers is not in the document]
brookescarlett: No, then we get TS errors: 
```
Element implicitly has an 'any' type because expression of type '"@leafygreen-ui/ripple"' can't be used to index type '{} | ModuleType'.
  Property '@leafygreen-ui/ripple' does not exist on type '{} | ModuleType'.ts(7053)
```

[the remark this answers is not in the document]
DesignerDave: Right. Is there a way to do this with less redundancy?

packages/ripple/README.md
brookescarlett: 👍

packages/ripple/src/getRippleGlobalNamespace.ts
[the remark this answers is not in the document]
brookescarlett: I tried, this is the best that I could come up with. I can keep playing with it though

packages/ripple/src/index.ts
brookescarlett: Just sent you a pen offline with perf update, lmk what you think

DesignerDave (approved):

PR #837, opened 2021-02-24
PD-935: Collapsible Side Navigation

DesignerDave:
## ✍️ Proposed changes

Revamps the style, accessibility, and behavior of the side navigation. See the ticket for details.

🎟 _Jira ticket:_ [Collapsible Side Nav](https://jira.mongodb.org/browse/PD-935)

## 🛠 Types of changes

## ✅ Checklist


packages/a11y/src/prefersReducedMotion.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Returns an emotion CSS instance for rendering styles that respect
+ * OS-level preferences for reduced motion.
+ *
+ * Use this function to remove scale, size, and positional transitions
+ * for users with that preference set.
+ * @param styles String of styles to render within media query.
+ */
+export default function prefersReducedMotion(styles: string) {
```
brookescarlett, line 11: This doesn't help us handle suppressing animations that exist in components when prefers-reduced-motion is set. Perhaps we can also add a hook that can be used in a similar way that our `useUsingKeyboardProvider` hook is used to enable us to toggle on/off animations and transitions

brookescarlett: Thinking even further about this, maybe we want to see about injecting global styles that respond to this media query

DesignerDave: I wanted to keep this solution to CSS-only.

On the global styles front, that's quite difficult. We could unset every CSS `transition` on the page, but the problem is that some transitions like opacity and color are appropriate even when the user has "reduce motion" set.

packages/side-nav/src/CollapseToggle.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const { navId } = useSideNavContext();
+
+  const Chevron = collapsed ? ChevronRight : ChevronLeft;
+
+  return (
+    <Tooltip
+      darkMode
+      align="right"
+      justify="middle"
+      open={hideTooltip != null ? !hideTooltip : undefined}
```
brookescarlett, line 123: should we be checking if typeof hideTooltip is a boolean?

packages/side-nav/src/CollapseToggle.tsx
```diff
@@ -0,0 +1,155 @@
...
+            })}
+          >
+            <Chevron role="presentation" />
+          </div>
+        </button>
+      }
+    >
+      <span aria-hidden>
+        {collapsed ? 'Expand' : 'Collapse'}
+        <kbd className={keyboardShortcut}>[</kbd>
```
brookescarlett, line 149: never seen this tag, nice

packages/side-nav/src/CollapseToggle.tsx
DesignerDave: I suppose it's technically better to do that, however unlikely it is that it matters. I can make the change.

packages/side-nav/src/CollapseToggle.tsx
DesignerDave: I knew it existed, but it took some digging to remember what the exact name was haha

packages/side-nav/src/SideNav.tsx
```diff
@@ -35,11 +157,122 @@ interface SideNavProps {
...
+      timeout={collapseDuration}
+    >
+      {state => (
+        <ContextProvider
+          value={{
+            navId,
+            currentPath,
+            collapsed,
+            setCollapsed,
+            portalContainer: portalContainer,
```
brookescarlett, line 220: should just be able to write portalContainer here

DesignerDave: Whoops, you're right.

packages/side-nav/src/SideNav.tsx
```diff
@@ -35,11 +157,122 @@ interface SideNavProps {
...
+                    listStyles,
+                    expandedListStyle,
+                    expandedStateStyles[state],
+                  )}
+                  role="menu"
+                >
+                  {children}
+                </ul>
+
+                <ul
```
brookescarlett, line 251: can we have a self-closing ul?

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -1,67 +1,81 @@
...
+import { spacing } from '@leafygreen-ui/tokens';
+import CollapsedSideNavItem from './CollapsedSideNavItem';
+import { useSideNavContext } from './SideNavContext';
 import {
   ulStyleOverrides,
   sideNavItemSidePadding,
   sideNavWidth,
 } from './styles';
 
+const menuGroupIdAllocator = IdAllocator.create('menu-group');
```
brookescarlett, line 24: we should prob call this sidenavgroup instead of menu group

DesignerDave: Good call

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -168,36 +194,81 @@ function SideNavGroup({
...
+      )}
+
+      <span className={headerText}>{header}</span>
+    </div>
+  );
 
   if (collapsible) {
     return (
-      <li {...rest}>
+      <li role="menuitem" className={cx(listItemStyle, className)} {...rest}>
```
brookescarlett, line 246: can't have li with a role menuitem, role here should be none and then add the menuitem role to it's first child

DesignerDave: Interesting. I wasn't getting any accessibility violations with this. In fact, it fixed one.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -231,9 +320,16 @@ function SideNavGroup({
   }
 
   return (
-    <li {...rest}>
-      <h4 className={sideNavLabelStyle}>{header}</h4>
-      <ul role="menu" className={ulStyleOverrides}>
+    <li role="menuitem" className={cx(listItemStyle, className)} {...rest}>
```
brookescarlett, line 323: ditto here

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -180,28 +164,36 @@ export interface SideNavItemProps {
...
 
-  const rest = omit(props, [
-    'active',
-    'disabled',
-    'ariaCurrentValue',
-    'className',
-    'children',
-  ]);
+  const { currentPath } = useSideNavContext();
+  const active = activeProp != null || currentPath === rest.href;
```
brookescarlett, line 184: Do we need to support bot currentPath matching rest.href and active? I feel like sometimes we try and cast a wider net in favor of attempting to improve the DX, but sometimes this backfires on us. I'm not highly opinionated on this one, but wanted to call that out as something we should think about

DesignerDave: If the consumer uses `as` to render these as React router links with `to`, or something else, an `active` prop provides alternative means to set an element as active.

packages/side-nav/src/SideNav.tsx
brookescarlett: Sorry, I meant more from an accessibility perspective. Why would we have a ul that doesn't contain any li elements.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Since the `currentPath` API is new, I suppose I could see removing it.

packages/side-nav/src/SideNavGroup.tsx
brookescarlett: I can point you to the thread where Deque called this out to us/ have read up on it a bit recently because I just fixed this in our menu component... will slack you more

DesignerDave: Chatted offline – best move was just to remove the roles altogether.

packages/side-nav/src/SideNav.tsx
DesignerDave: We don't. The elements are portaled into the list.

DesignerDave: See `CollapsedSideNavItem` for an example. It's how we render the collapsed state.

brookescarlett: Ah got it

packages/side-nav/src/CollapseToggle.tsx
```diff
@@ -0,0 +1,155 @@
...
+  border-radius: 2px;
+  box-shadow: 0 3px 3px -2px rgba(0, 0, 0, 0.3), 0 0 2px rgba(0, 0, 0, 0.3),
+    inset 0 1px 2px rgba(255, 255, 255, 0.15);
+  border: 1px solid ${uiColors.gray.dark1};
+  line-height: 1em;
+  color: ${uiColors.white};
+  margin-left: ${spacing[2]}px;
+`;
+
+type DetailedElementProps<T> = React.DetailedHTMLProps<
```
brookescarlett, line 97: What is this providing that our HTMLElementProps helper isn't

DesignerDave: Quite honestly, I forgot which approach we had settled on and grabbed this from another component 🤦

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -168,36 +194,81 @@ function SideNavGroup({
...
+    return React.Children.toArray(children).some(child => {
+      return (
+        isComponentType(child, 'SideNavItem') &&
+        (child.props.active || child.props.href === currentPath)
+      );
+    });
+  }, [children, currentPath]);
+
+  const accessibleGlyph =
+    (glyph && isComponentType(glyph, 'Glyph')) || isComponentType(glyph, 'Icon')
```
brookescarlett, line 219: I think there's a reason why this isComponentType won't work on our glyphs (I don't remember what it is) but I know that we exposed the .isGlyph property on glyphs for a reason

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -215,12 +286,29 @@ function SideNavGroup({
...
                 [transitionStyles.exiting]: state === 'exiting',
                 [transitionStyles.exited]: state === 'exited',
               })}
             >
-              <ul ref={ulRef} role="menu" className={ulStyleOverrides}>
+              <ul
+                ref={ulRef}
+                id={menuId}
+                aria-labelledby={menuGroupLabelId}
+                className={cx(
```
brookescarlett, line 299: Should we use prefersReducedMotion to disable these transitions when necessary

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -1,132 +1,109 @@
...
+  text-align: left;
+  text-decoration: none;
+  text-transform: capitalize;
+  color: ${uiColors.gray.dark2};
 
-  &:active:before {
-    transform: scale(1);
-  }
+  // Stateful transitions
+  transition: background-color 150ms ease-in-out;
```
brookescarlett, line 47: ditto here

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: Probably because you need to do this hah. Regardless, this is working as intended,

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: Transitions that don't use motion are fine with the setting, including opacity. 

Example: Turn on "reduce motion", and then open mission control on your mac (three finger swipe up via trackpad, or the f3 key). They used to transform and scale the windows, but switch to doing an opacity transition.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: See my other comment.

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -1,67 +1,80 @@
...
+import { css, cx } from '@leafygreen-ui/emotion';
+import { spacing } from '@leafygreen-ui/tokens';
+import CollapsedSideNavItem from './CollapsedSideNavItem';
 import {
   ulStyleOverrides,
   sideNavItemSidePadding,
   sideNavWidth,
 } from './styles';
 
+const menuGroupIdAllocator = IdAllocator.create('side-nav-group');
```
brookescarlett, line 23: could also change this var name just for consistency

packages/side-nav/src/SideNavGroup.tsx
brookescarlett: Really? With an Icon and a Glyph?

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -180,34 +163,39 @@ export interface SideNavItemProps {
...
     disabled = false,
     ariaCurrentValue = AriaCurrentValue.Page,
     className,
     children,
+    onClick: onClickProp,
+    glyph,
+    ...rest
   } = props;
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+  const active = typeof activeProp === 'boolean' && activeProp;
```
brookescarlett, line 181: if we default to false then we don't have to check for anything here, and we can just use the passed in prop

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -180,34 +163,39 @@ export interface SideNavItemProps {
...
 
-  const rest = omit(props, [
-    'active',
-    'disabled',
-    'ariaCurrentValue',
-    'className',
-    'children',
-  ]);
+  const onClick = disabled
+    ? (e: React.MouseEvent) => {
```
brookescarlett, line 184: Why do we need to do this specifically when the item is disabled? (As opposed to just not calling `onClickProp()`

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -180,34 +163,39 @@ export interface SideNavItemProps {
...
+        e.nativeEvent.stopImmediatePropagation();
+        e.preventDefault();
+      }
+    : onClickProp;
 
-  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+  const accessibleGlyph =
+    (glyph && isComponentType(glyph, 'Glyph')) || isComponentType(glyph, 'Icon')
+      ? React.cloneElement(glyph, { 'aria-hidden': true })
```
brookescarlett, line 192: Does this have a performance implication, as opposed to invoking the <Glyph /> and passing the prop? (I'm not sure)

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: You're right. Fixed this

packages/side-nav/src/SideNavItem.tsx
brookescarlett: ditto here

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Nice call. We couldn't do this before when we supported currentPath.

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: Done

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Yes – we're removing the actual `disabled` attribute usage so that screen reader users can focus on the disabled item. That means we have to manually prevent the click listeners.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Invoking the what? The component takes a rendered ReactNode.

We did research a while back to see if cloneElement had significant performance impacts, and found that wasn't the case IIRC. Even so, the number of items in navigation are not unbounded, so this doesn't have to perform well with 500 on the page (though I think it would be fine if it did).

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Talked offline, and removed the API

packages/side-nav/src/SideNavItem.tsx
brookescarlett: Omg so weird I guess GH ignores `< />` characters
`<Glyph />`

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -1,95 +1,115 @@
...
   text-transform: uppercase;
   color: ${uiColors.green.dark2};
-  margin-top: 12px;
+  min-height: ${spacing[5]}px;
+  margin-top: 0;
   margin-bottom: 0;
   padding: 4px ${sideNavItemSidePadding}px 4px ${sideNavItemSidePadding}px;
-  line-height: 1.3em;
-  position: relative;
+  line-height: 1em;
```
brookescarlett, line 48: A curiosity rather than a comment, but why use em here when we use px everywhere else?

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -168,36 +193,84 @@ function SideNavGroup({
...
+    return React.Children.toArray(children).some(child => {
+      return isComponentType(child, 'SideNavItem') && child.props.active;
+    });
+  }, [children]);
+
+  const accessibleGlyph =
+    glyph && (isComponentGlyph(glyph) || isComponentType(glyph, 'Icon'))
+      ? React.cloneElement(glyph, {
+          className: cx(customIconStyles, glyph.props.className),
+          'aria-hidden': true,
```
brookescarlett, line 220: if you pass role="presentation" aria-hidden will be applied for you

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -1,132 +1,108 @@
 import React, { ReactNode } from 'react';
 import PropTypes from 'prop-types';
-import omit from 'lodash/omit';
-import { css, cx } from '@leafygreen-ui/emotion';
-import { uiColors } from '@leafygreen-ui/palette';
-import { AriaCurrentValue, createDataProp } from '@leafygreen-ui/lib';
+import { transparentize } from 'polished';
+import {
+  AriaCurrentValue,
```
brookescarlett, line 5: should we move this to a11y

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -1,132 +1,108 @@
...
 
-  &:hover,
-  &:focus {
-    &:before {
-      transform: scale(1);
-      background-color: ${uiColors.gray.light2};
-    }
-  }
+  // Typography
+  font-family: Akzidenz, ‘Helvetica Neue’, Helvetica, Arial, sans-serif;
```
brookescarlett, line 36: can use our fontFamilies token here

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -1,132 +1,108 @@
...
+  font-size: 14px;
+  font-weight: normal;
+  line-height: 1em;
```
brookescarlett, line 39: ditto on the em

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -156,6 +132,13 @@ export interface SideNavItemProps {
   children?: ReactNode;
 
   href?: string;
```
brookescarlett, line 134: should we add tsdoc for consistency

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -191,23 +173,28 @@ const SideNavItem: ExtendableBox<
...
-    'ariaCurrentValue',
-    'className',
-    'children',
-  ]);
+  const onClick = disabled
+    ? (e: React.MouseEvent) => {
+        e.nativeEvent.stopImmediatePropagation();
+        e.preventDefault();
+      }
+    : onClickProp;
```
brookescarlett, line 187: I think what I meant was, wouldn't we not want to call this onClick prop when disabled. Instead const onClick = !disabled && () => {}

website/components/News.tsx
```diff
@@ -10,8 +10,8 @@ import { mq } from 'utils/mediaQuery';
 const newsContainer = css`
   display: flex;
   flex-direction: column;
-  justify-content: center;
-  padding-bottom: ${spacing[4]}px;
+  justify-content: top;
+  padding-top: ${spacing[2] - 1}px;
```
brookescarlett, line 14: if we're going to make this a calculation, can we explain where the 1px is coming frrom? otherwise we can just do 7px

website/components/navigation/Navigation.tsx
```diff
@@ -158,16 +145,23 @@ function Navigation() {
...
+        z-index: 1;
+      `}
+    >
       <MDBDesignLogo className={logoStyles} onClick={() => push(HOME_PAGE)} />
-      <SideNav
+      <CollapsedSideNavItem
         className={css`
-          margin-top: ${spacing[3]}px;
+          background-color: #09804c;
+          margin-top: -${spacing[3] + 1}px;
```
brookescarlett, line 157: Ditto re the calc and the 1px

website/layouts/BaseLayout.tsx
```diff
@@ -1,42 +1,54 @@
...
-  align-items: flex-start;
+  height: 100vh;
+  background-color: ${uiColors.white};
+  grid-template-columns: auto 1fr;
+
   ${mq({
-    flexDirection: ['column', 'column', 'row'],
-    paddingLeft: [`${spacing[4]}px`, `${spacing[4]}px`, '0px'],
-    paddingRight: [`${spacing[4]}px`, '0px'],
+    display: ['block', 'block', 'grid', 'grid'],
```
brookescarlett, line 16: can do 'block', 'block', 'grid'

website/layouts/Footer.tsx
```diff
@@ -22,29 +23,37 @@ const footerContainer = css`
 const linksContainer = css`
   display: flex;
   flex-direction: column;
-  color: white;
-  margin-top: 2px;
+  color: ${uiColors.gray.dark2};
+  // Pixel-pushed for baseline alignment with logo
```
brookescarlett, line 27: 👍

website/layouts/Footer.tsx
```diff
@@ -64,18 +73,18 @@ function Footer() {
...
-          About Design at MongoDB
+          About design at MongoDB
         </FooterLink>
         <FooterLink href="https://www.mongodb.com/blog">Blog</FooterLink>
         <FooterLink href="https://www.mongodb.com/blog/channel/events">
           Events
         </FooterLink>
         <FooterLink href="https://www.mongodb.com/careers">Careers</FooterLink>
-        <p className={trademarkStyle}>© 2020 MongoDB, Inc.</p>
+        <p className={trademarkStyle}>© 2021 MongoDB, Inc.</p>
```
brookescarlett, line 87: ah thank you, been meaning to update this..

website/components/navigation/Navigation.tsx
```diff
@@ -158,16 +145,23 @@ function Navigation() {
   }
 
   return (
-    <nav className={navContainer}>
+    <SideNav
```
brookescarlett, line 148: One thing I'm noticing is that when you hover over the collapsed nav it pushes the layout over slightly, even though it's being overlayed on top of the page content. It's particularly noticeable on the home page

brookescarlett: There also might be a better way to transition between the tablet/mobile nav and the desktop nav

website/pages/component/side-nav/example.tsx
```diff
@@ -57,18 +58,33 @@ function DefaultExample({
...
-    </SideNav>
+    <div
+      className={css`
+        display: grid;
+        grid-template-columns: auto 1fr;
+        height: 500px;
+      `}
+    >
+      <SideNav>
+        <SideNavGroup {...collapsibleProps} header={header}>
```
brookescarlett, line 69: I think we should update the toggles to better show off what this side nav can do / achieve parity with storybook

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: By using em, we reduce the maintenance required if we update the font-size. I like em units when you specifically need something to be a whole number multiple of the current font-size.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Maybe, but in a different PR. Where this lives doesn't meaningfully impact the implementation of this PR, so it's out of scope.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Responded to the other comment.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Good call

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: Good catch.

website/components/News.tsx
DesignerDave: Optical alignment where the spacing is still based on our spacing tokens. I'll comment.

brookescarlett: Thank you!

packages/side-nav/src/SideNavItem.tsx
brookescarlett: Makes sense to me, let's just make a ticket to move it then?

packages/side-nav/src/SideNavGroup.tsx
brookescarlett: We use this in a fair number of places (14px font => 20px line-height, 16px font => 24px line-height) maybe we should consider applying this elsewhere. Again, out of scope for this PR but if that's a pattern we want to start applying, it's worth calling out

website/components/navigation/Navigation.tsx
DesignerDave: Turned out to be an issue with the scroll bar. Fixed!

packages/side-nav/src/SideNavItem.tsx
DesignerDave: If this is a link and we didn't pass an onClick when the item is disabled, clicking on the link would still navigate you. That's why we still have an onClick handler when it's disabled.

website/pages/component/side-nav/example.tsx
DesignerDave: I updated the knobs a bit – let me know if there's something specific you'd like to see that's not there.

packages/side-nav/src/SideNavItem.tsx
DesignerDave: Done

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: The point of this is to eliminate line height. 1em line height is used to improve vertical centering usually where line wrapping isn't a major consideration. It's a CSS pattern not specific to our own standards, but general CSS authoring.

packages/side-nav/src/SideNav.tsx
```diff
@@ -35,11 +161,116 @@ interface SideNavProps {
  * @param props.className Class name that will be applied to the root-level element.
  * @param props.children Content that will be rendered inside the root-level element.
  */
-function SideNav({ className, children, ...rest }: SideNavProps) {
+function SideNav({ className, children, id: idProp, ...rest }: SideNavProps) {
+  const { Provider: ContextProvider } = SideNavContext;
+  const [collapsed, setCollapsed] = useState(false);
```
brookescarlett, line 166: Should we offer collapsed as a prop, such that we can have a side nav that is always open?

brookescarlett: Or `collapsible` rather

packages/side-nav/src/SideNav.tsx
```diff
@@ -35,11 +161,116 @@ interface SideNavProps {
...
+    portalContainer,
+    setPortalContainer,
+  ] = useState<HTMLUListElement | null>(null);
+
+  // We visually expand the navigation when a user focuses on an element within the navigation
+  // while navigating via keyboard.
+  const focusExpand = usingKeyboard && focus;
+
+  // Nav element should have a programmatically-determinable label
+  validateAriaLabelProps(rest, 'SideNav');
```
brookescarlett, line 181: I feel like this is a relatively stable component, it might be possible that we just provide teh default label of "Side Navigation." In storybook, I believe the aria-label supplied was "Realm" or something to that effect, and I actually think the function of the component, rather than its context, might be more descriptive here

packages/side-nav/src/SideNavGroup.tsx
```diff
@@ -168,36 +193,84 @@ function SideNavGroup({
...
+
+  const isActiveGroup: boolean = useMemo(() => {
+    return React.Children.toArray(children).some(child => {
+      return isComponentType(child, 'SideNavItem') && child.props.active;
+    });
+  }, [children]);
+
+  const accessibleGlyph =
+    glyph && (isComponentGlyph(glyph) || isComponentType(glyph, 'Icon'))
+      ? React.cloneElement(glyph, {
```
brookescarlett, line 218: does this need to be a ternary?

packages/side-nav/src/SideNav.tsx
DesignerDave: I'll see if I can dig it up, but the guidelines I was following specifically said *not* to include the words "navigation" and to just say at a high level what you're navigating. E.g. "Realm Project" instead of "Realm Project Side Nav".

packages/side-nav/src/SideNavGroup.tsx
DesignerDave: Not technically, but I did prefer this approach largely because of type inference. We don't need to type the variable explicitly since it's part of the same statement.

If you feel strongly, I can update this.

packages/side-nav/src/SideNav.tsx
DesignerDave: No – we want behavior across apps to be universally consistent. That's one of the problems we currently have.

brookescarlett: That's fair, but maybe we can offer something like defaultOpen so that you can control if its collapsed or not on first render?

packages/side-nav/src/CollapsedSideNavItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+        className={cx(
+          collapsedItemStyles,
+          {
+            [collapsedItemActiveStyle]: active,
+          },
+          className,
+        )}
+      >
+        {children}
+      </li>
```
brookescarlett, line 54: I think we need some visually hidden text here. Otherwise, we're rendering a list with empty li tags, and I can imagine that'd be confusing

packages/side-nav/src/CollapsedSideNavItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+    border-top: 1px solid ${uiColors.gray.light2};
+  }
+`;
+
+const collapsedItemActiveStyle = css`
+  background-color: ${uiColors.green.light3};
+`;
+
+interface CollapsedSideNavItemProps {
+  // React Node rendered when the navigation is collapsed.
```
brookescarlett, line 25: are we commenting rather than ts-docing here intentionally?

packages/side-nav/src/SideNav.spec.tsx
```diff
@@ -1,300 +1,52 @@
...
   };
-  let renderedEls: RenderedElements = {};
 
   const className = 'test-class-name';
-  const headerText = 'test-header-text';
-  const headerContent = (
-    <div data-testid={testIds.sideNavHeader}>Header As Content</div>
-  );
+
+  let renderedEls: RenderedElements = {};
```
brookescarlett, line 30: We should add a test for collapsing/expanding the SideNav with a key press

packages/side-nav/src/SideNavGroup.spec.tsx
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+import { render, cleanup, screen, fireEvent } from '@testing-library/react';
+import { SideNavGroup, SideNavItem } from '.';
+import { SideNavGroupProps } from './SideNavGroup';
+
+describe('packages/side-nav', () => {
```
brookescarlett, line 6: can we add a test for when the group label has a glyph, that it's rendered as aria-hidden?

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -191,23 +185,28 @@ const SideNavItem: ExtendableBox<
...
-    'active',
-    'disabled',
-    'ariaCurrentValue',
-    'className',
-    'children',
-  ]);
+  const onClick = disabled
+    ? (e: React.MouseEvent) => {
+        e.nativeEvent.stopImmediatePropagation();
+        e.preventDefault();
```
brookescarlett, line 197: Why do we need to preventDefault here?

DesignerDave: That stops it from redirecting you if it's a link that's clicked.

packages/side-nav/src/SideNav.tsx
DesignerDave: I don't want that behavior to be inconsistent across implementations. We should be the ones controlling that, not the consumer.

packages/side-nav/src/SideNav.spec.tsx
DesignerDave: I really tried but unfortunately there's not a great way to do this without something very hacky. As a workaround, I tested that the states update appropriately on the collapse toggle itself. For the navigation itself, the only thing that changes between states are the classNames for the display.

packages/side-nav/src/SideNavGroup.spec.tsx
DesignerDave: I can test that the role is appropriate, but testing aria-hidden is creating a duplicate test for the icon component, not this component.

brookescarlett: yes wfm

packages/side-nav/src/SideNav.spec.tsx
brookescarlett: if we're already using data-testids here, maybe we do something like data-testid="side-nav-${isOpen ? "open" : "closed"}"

brookescarlett: the naming can definitely be better in my example

DesignerDave: That doesn't test that the state is correct any more than testing that the aria-expanded prop on the button is correct.

brookescarlett: yeah 🤦

packages/side-nav/src/CollapsedSideNavItem.tsx
DesignerDave: For some reason I was thinking tsdoc was compatible with both single-line and multi-line JS comments. Fixed.

packages/side-nav/src/SideNav.tsx
brookescarlett: Interesting

website/components/navigation/Navigation.tsx
```diff
@@ -158,16 +145,26 @@ function Navigation() {
...
+      aria-label="LeafyGreen Design System"
+      className={css`
+        z-index: 1;
```
brookescarlett, line 151: this breaks with our current modal implementation

brookescarlett:
just noticed that im not seeing any changelogs here -- not sure if we want to give the release a beta flag?

brookescarlett (approved):
LGTM

PR #845, opened 2021-03-05
PD-1210: Address focus issue with Tab component

brookescarlett:
Sorry for all of the commits in the history, originally I branched this off of the ripple package by accident.

## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1210](https://jira.mongodb.org/browse/PD-1210)

## 🛠 Types of changes

## ✅ Checklist


packages/tabs/src/TabTitle.tsx
```diff
@@ -141,19 +165,32 @@ const TabTitle: ExtendableBox<BaseTabTitleProps, 'button'> = ({
...
   ...rest
 }: BaseTabTitleProps) => {
   const { usingKeyboard: showFocus } = useUsingKeyboardContext();
   const [showEllipsis, setShowEllipsis] = useState(false);
   const titleRef = useRef<HTMLAnchorElement | HTMLButtonElement>(null);
+  const activeEl = useDocumentActiveElement();
+
   const mode = darkMode ? Mode.Dark : Mode.Light;
 
+  // Checks to see if the current activeElement is a part of the same tab set
```
brookescarlett, line 178: Colocating focus logic. Right now, we are checking for tabsList.indexof(activeEl) inside of the Tabs component, and that useEffect was dependent on activeEl (so would rerun every time activeEl changed) and separately we were setting focus inside of this useEffect (so this useEffect would trigger the other, causing a bad side effect).

website/pages/component/tabs/example.tsx
```diff
@@ -75,3 +75,329 @@ export default function TabsLiveExample() {
     </LiveExample>
   );
 }
+
+// import React, { useState } from 'react';
```
brookescarlett, line 79: This was the example that I created with @sophstad who initially raised the bug. Wanted to leave the code so that you can run locally and check as well, but I will remove before we merge.

packages/tabs/src/TabTitle.tsx
```diff
@@ -125,6 +131,23 @@ const textOverflowStyles = css`
   text-overflow: ellipsis;
 `;
 
+function useDocumentActiveElement() {
+  const [activeEl, setActiveEl] = useState<Element | null>(null);
+
+  const handleFocusIn = useCallback(() => {
+    setActiveEl(document.activeElement);
```
DesignerDave, line 138: Why can't we use `document.activeElement` directly instead of storing it in component state?

If this needs to stay, we should have a comment explaining why it needs to exist here.

packages/tabs/src/TabTitle.tsx
```diff
@@ -125,6 +131,23 @@ const textOverflowStyles = css`
   text-overflow: ellipsis;
 `;
 
+function useDocumentActiveElement() {
+  const [activeEl, setActiveEl] = useState<Element | null>(null);
+
+  const handleFocusIn = useCallback(() => {
```
DesignerDave, line 137: Why do we need `useCallback` here? I don't think it's doing anything because we're only adding an event listener when the component mounts rather than creating one every render.

packages/tabs/src/TabTitle.tsx
```diff
@@ -125,6 +131,23 @@ const textOverflowStyles = css`
...
+    setActiveEl(document.activeElement);
+  }, []);
+
+  useEffect(() => {
+    document.addEventListener('focusin', handleFocusIn);
```
DesignerDave, line 142: This looks a lot like we're adding a new event listener for every single TabTitle on the page. Even though we clean these up, that's eating up more memory than we need to.

packages/tabs/src/TabTitle.tsx
DesignerDave: Remind me why we need to manage focus of the tabs in the first place? I feel like we may be able to solve the problem by simplifying the implementation rather than adding complexity here.

DesignerDave (changes requested):

packages/tabs/src/TabTitle.tsx
brookescarlett: we need to manage focus manually because otherwise keyboard navigation won't focus a tab item.

packages/tabs/src/TabTitle.tsx
brookescarlett: good catch

packages/tabs/src/TabTitle.tsx
brookescarlett: Removed

packages/tabs/src/TabTitle.tsx
brookescarlett: Removed

packages/tabs/src/TabTitle.tsx
DesignerDave: In the future, I think we should explore this pattern a bit more, but sounds good for now.

DesignerDave (approved):
Once the commented code is removed, LGTM!

PR #846, opened 2021-03-08
PD-991: Applies hover styles to active on touch device

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-991](https://jira.mongodb.org/browse/PD-991)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/icon-button/src/IconButton.tsx
+++ b/packages/icon-button/src/IconButton.tsx
@@ -5,8 +5,6 @@ import Box, { ExtendableBox } from '@leafygreen-ui/box';
 import { Either, isComponentType } from '@leafygreen-ui/lib';
 import { uiColors } from '@leafygreen-ui/palette';
 import { isComponentGlyph } from '@leafygreen-ui/icon';
-import { useViewportSize } from '@leafygreen-ui/hooks';
-import { breakpoints } from '@leafygreen-ui/tokens';
 
 const Mode = {
   Light: 'light',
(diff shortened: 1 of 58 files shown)


packages/icon-button/src/IconButton.tsx
```diff
@@ -82,37 +85,43 @@ const iconButtonSizes = {
...
-
-      &:before {
-        background-color: ${uiColors.gray.light2};
+const getIconButtonModeStyle = (
+  isTouchDevice: boolean,
+  mode: 'dark' | 'light',
+) => {
+  if (mode === Mode.Light) {
+    return css`
+      &: ${isTouchDevice ? 'active' : 'hover'} {
```
DesignerDave, line 94: Why not simply do `&:active, &:hover {}` here? Is there a reason we wouldn't want to persist the hover state when it's being pressed on desktop? It could also eliminate the need for this to be a function.

Also, while it _probably_ works, I'd remove the extra space between the colon and interpolated statement.

DesignerDave (changes requested):

packages/icon-button/src/IconButton.tsx
brookescarlett: Updated!

packages/icon-button/src/IconButton.tsx
```diff
@@ -82,37 +85,45 @@ const iconButtonSizes = {
...
 
-const iconButtonMode = {
-  [Mode.Light]: css`
-    &:hover {
-      color: ${uiColors.gray.dark2};
-
-      &:before {
-        background-color: ${uiColors.gray.light2};
+const getIconButtonModeStyle = (
+  isTouchDevice: boolean,
```
DesignerDave, line 89: I might be missing something, but I don't think this is being used anymore. We can just go back to the old structure now in this case.

DesignerDave (approved):

PR #849, opened 2021-03-09
PD-1222: Moves `validateAriaLabelProps` function from lib to a11y package, and update usage

brookescarlett:
## ✍️ Proposed changes

- Removes `validateAriaLabelProps` from lib
- Adds `validateAriaLabelProps` and `validateLabelProps` to a11y
- Updates consumption across components

🎟 _Jira ticket:_ [PD-1222](https://jira.mongodb.org/browse/PD-1222)

## 🛠 Types of changes

## ✅ Checklist


packages/a11y/src/validateProps.tsx
```diff
@@ -11,3 +11,17 @@ export function validateAriaLabelProps(
     );
   }
 }
+
+// Validate that either label or aria-labelledby are present within props.
+export function validateLabelProps(
+  props: Record<string, any>,
+  componentName?: string,
```
DesignerDave, line 18: We should probably make componentName required considering the number of times this being optional has been a pain.

packages/icon-button/src/IconButton.tsx
```diff
@@ -200,11 +201,13 @@ const IconButton: ExtendableBox<
     const mode = darkMode ? 'dark' : 'light';
 
     // We do our own proptype validation here to ensure either 'aria-label' or 'aria-labelledby' are passed to the component.
-    if (!ariaLabel && !ariaLabelledBy) {
-      console.error(
-        'For screen-reader accessibility, aria-label or aria-labelledby must be provided to IconButton.',
-      );
-    }
+    validateAriaLabelProps(
+      {
```
DesignerDave, line 205: The function was designed to accept all of props. Is there a reason we're doing this manually? It's not actually validating the props passed in to the component.

DesignerDave: Ditto for all other instances where this approach was taken.

DesignerDave (changes requested):

packages/a11y/src/validateProps.tsx
brookescarlett: Good call

packages/icon-button/src/IconButton.tsx
brookescarlett: They were already pulled out from rest so I just went with it, but I can go back and clean it up where relevant.

packages/select/src/Select.tsx
```diff
@@ -118,16 +119,11 @@ export default function Select({
     }
   }, [onChange, readOnly, value]);
 
-  if (!label && !ariaLabelledBy) {
-    console.error(
-      'For screen-reader accessibility, label or aria-labelledby must be provided to IconButton.',
-    );
-  }
+  validateLabelProps({ label, 'aria-labelledby': ariaLabelledBy }, 'Select');
```
brookescarlett, line 122: For example, we use the label and arialabelledby props in the component here, so we need to have them separated from rest

DesignerDave: Then we should do this check before destructuring the props, right?

brookescarlett: as long as it is at least checking for the presence of one or the other of the two props I don't see an issue

DesignerDave: My main issue with this is that we aren't actually validating that `aria-labelledby` is spelled correctly, for example (something typescript doesn't check for, though our a11y eslint plugin does in certain cases). We're not testing that the key exists in props, which is one of the reasons for the function.

My point here is, we're only using this function to half of its potential, and also increasing the maintenance burden by setting the keys that we're passing to the function manually in each component.

packages/text-area/src/TextArea.tsx
```diff
@@ -167,11 +167,7 @@ export default function TextArea({
     }
   };
 
-  if (!label && !ariaLabelledBy) {
-    console.error(
-      'For screen-reader accessibility, label or aria-labelledby must be provided to IconButton.',
-    );
-  }
+  validateLabelProps({ label, ...rest }, 'TextArea');
```
DesignerDave, line 170: This function isn't equivalent to the previous check. We only check that `aria-label` or `aria-labelledby` is present with `validateLabelProps`. It will not check whether `label` exists.

This might have to look like:
```js
if (!label) {
  validateLabelProps(rest, 'TextArea');
}
```

packages/text-input/src/TextInput.tsx
```diff
@@ -323,11 +323,7 @@ const TextInput: React.ComponentType<
       }
     }
 
-    if (!label && !ariaLabelledBy) {
-      console.error(
-        'For screen-reader accessibility, label or aria-labelledby must be provided to TextInput.',
-      );
-    }
+    validateLabelProps({ label, ...rest }, 'TextInput');
```
DesignerDave, line 326: Ditto here. Can you validate we're switching this to equivalent usages elsewhere as well?

DesignerDave (changes requested):

packages/text-input/src/TextInput.tsx
brookescarlett: There are two functions, validateLabelProps and validateAriaLabelProps, this is because there's a difference in what's required based on the component's usage. For example, we're requiring that form inputs have a label... if a consumer does not want to use our styled label, then we need to verify that they have another label on the page which is why aria-label doesn't cut it, and we need an aria-labelledby. That's different from an implementation like SideNav, where we don't need an explicit label rendered on the screen, and therefore their options are aria-label or aria-labelledby

DesignerDave: Then I'd ask, does this need to be generalized? This seems like a very different use-case, and very component specific. It's not a check that there's a programmatically-determinable label, but that there is a visual label in the first place which is more of a design consideration. I don't think a utility function here is the right solution, and I'd rather keep this PR to its intent – moving validateAriaLabelProps to the A11y package.

If anything, we could make a component that handles the label and description rendering that we do in a few components. We can make a ticket to address this issue more specifically.

packages/text-input/src/TextInput.tsx
```diff
@@ -323,9 +323,9 @@ const TextInput: React.ComponentType<
       }
     }
 
-    if (!label && !ariaLabelledBy) {
+    if (!label && !ariaLabelledby) {
       console.error(
-        'For screen-reader accessibility, label or aria-labelledby must be provided to TextInput.',
+        'For screen-reader accessibility, label or aria-labelledby must be provided to TextArea.',
```
DesignerDave, line 328: Looks like this is the wrong component name.

packages/text-input/src/TextInput.tsx
```diff
@@ -351,7 +351,6 @@ const TextInput: React.ComponentType<
           >
             <input
               {...rest}
-              aria-labelledby={ariaLabelledBy}
```
DesignerDave, line 354: Was this being removed intentional?

packages/text-input/src/TextInput.tsx
```diff
@@ -298,10 +298,10 @@ const TextInput: React.ComponentType<
       state = State.None,
       type = TextInputType.Text,
       id: propsId,
-      'aria-labelledby': ariaLabelledBy,
       value: controlledValue,
       className,
       darkMode = false,
+      'aria-labelledby': ariaLabelledby,
```
DesignerDave, line 304: Any reason for the casing change?

.changeset/kind-mice-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/a11y': minor
+---
+
+Adds `validateAriaLabelProps` and `validateLabelProps` functions to package.
```
DesignerDave, line 5: This needs an update

packages/text-area/src/TextArea.tsx
```diff
@@ -145,7 +145,7 @@ export default function TextArea({
   id: idProp,
   value: controlledValue,
   onChange,
-  'aria-labelledby': ariaLabelledBy,
+  'aria-labelledby': ariaLabelledby,
```
DesignerDave, line 148: Ditto on the casing change here

DesignerDave (changes requested):

packages/text-input/src/TextInput.tsx
brookescarlett: nope, good catch.

packages/text-input/src/TextInput.tsx
brookescarlett: ty

packages/text-area/src/TextArea.tsx
brookescarlett: more closely mirrors the aria-label

.changeset/kind-mice-travel.md
brookescarlett: yep, done.

packages/text-area/src/TextArea.tsx
DesignerDave: 👍

DesignerDave (approved):

PR #863, opened 2021-03-18
PD-1032: Button updates

brookescarlett:
## ✍️ Proposed changes

- Major version update to Button component. 
- Adds new colors to uiColors

🎟 _Jira ticket:_ [PD-1032](https://jira.mongodb.org/browse/PD-1032)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

- Run mongodb.design, 
- Look at: Button, Select, Copyable, Confirmation Modal, MongoNav


DesignerDave (changes requested):
Looking awesome! Few bits of feedback to start.

Also I noticed the ripple effect still is being applied to the disabled buttons. That seems unintentional.

packages/button/src/Button.tsx
```diff
@@ -1,483 +1,182 @@
...
-            }
-          `]: borderRadius !== undefined,
-        },
-        {
-          [css`
-            &:before {
-              opacity: 1;
-            }
-          `]: forceState.focused === true,
+            margin-right: ${iconSpacing[size]}px;
```
DesignerDave, line 132: I think that I prefer coupling the css call with the iconSpacing enum versus using an inline css call here.

packages/button/src/Button.tsx
```diff
@@ -1,483 +1,182 @@
...
-    height: 25px;
-  `,
-
-  [Size.Normal]: css`
-    height: 32px;
-  `,
+const rippleStyle = css`
+  overflow: hidden;
+  border-radius: 3px;
+  flex-grow: 1;
```
DesignerDave, line 14: Since this is position absolute, I don't think the flex grow property is doing anything here.

packages/button/src/Button.tsx
```diff
@@ -1,483 +1,182 @@
...
+    darkMode,
+    baseFontSize,
+    disabled,
+    showFocus,
+  });
+
+  const isAnchor = typeof rest.href === 'string';
+  let type: JSX.IntrinsicElements['button']['type'];
+
+  // @ts-expect-error rest.as may be defined
```
DesignerDave, line 103: I think the note here meant to say "undefined"

Also, I'm inclined to think we can address this TS error though code by checking whether rest.as is defined before the other checks below.

packages/button/src/Button.tsx
```diff
@@ -1,483 +1,182 @@
...
-    // styles from the the base style and each variant's styles
+  const buttonProps = {
+    type,
+    className: cx(buttonClassName, className),
+    ref: forwardRef,
+    // Provide a default value for the as prop
+    // If consumping application passes a value for as, it will override the default set here
+    as: isAnchor ? 'a' : 'button',
+    // only add a disabled prop if not an anchor
+    ...(typeof rest.href !== 'string' && { disabled }),
```
DesignerDave, line 116: This looks like it would error if `typeof rest.href !== 'string` evaluates to false.

packages/button/src/Button.tsx
```diff
@@ -1,483 +1,182 @@
...
         {
           [css`
-            &:after {
-              opacity: 1;
-            }
-          `]: forceState.active === true,
+            margin-left: ${iconSpacing[size]}px;
+          `]: !isIconOnlyButton || !!leftGlyph,
         },
+        getIconStyle({ variant, size, darkMode, disabled, isIconOnlyButton }),
```
DesignerDave, line 149: I have very mixed feelings about this sort of function. To me, it looks like we should be building a component to handle this since we're effectively handling props as-is.

packages/button/src/Button.tsx
```diff
@@ -1,483 +1,182 @@
...
+  }: ButtonProps,
+  forwardRef,
+) {
+  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+  const rippleRef = useRef<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    if (rippleRef.current != null) {
+      registerRipple(rippleRef.current, { variant, darkMode });
+    }
```
DesignerDave, line 86: We should return a cleanup function here, no?

packages/confirmation-modal/src/ConfirmationModal.tsx
```diff
@@ -46,11 +46,11 @@ const buttonStyle = css`
   margin: 0 2px;
 
   :first-of-type {
```
DesignerDave, line 48: Nit: I know this selector wasn't written in this PR, but we should include the `&` for this sort of selector to indicate if it's being applied to the current className or if it's a descendent selector.

```
& :first-of-type {
  margin: 0 0 0 4px;
}
```

Also, do we need to be defining the margin on all 4 sides here, or would a simple margin-left be sufficient?

packages/ripple/src/registerRipple.ts
```diff
@@ -108,6 +108,7 @@ const staticRippleStyles = `
     transform: scale(0.2);
     opacity: 0;
     pointer-events: none;
+    z-index: -1;
```
DesignerDave, line 111: Could you make a note why we have this?

DesignerDave (changes requested):

packages/button/UPGRADE.md
```diff
@@ -0,0 +1,22 @@
+# v10 to v11
+
+This major version brings many changes to both the component's API and it's design. Below are the main changes that you may need to account for:
+
+Breaking Changes to the API:
+
+- Variant names have changed. We now support: `'default'`, `'primary'`, `'primaryOutline'`, `'danger'`, `'dangerOutline'`. The old variant, `'info'` has been renamed to `'primaryOutline'`, and the `'dark'` variant has been replaced by the addition of the `darkMode` prop.
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/button/UPGRADE.md
```diff
@@ -0,0 +1,22 @@
...
+- We have renamed the `normal` size to `default`. This is to be more consistent with our naming conventions across the library.
+- The `glyph` prop has been replaced by two new props: `leftGlyph` and `rightGlyph`. While positioning a glyph relative on either side of text was possible before, we're adding more bespoke styles to icons passed into the component, so we encourage you to use these new props.
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/button/src/Button.tsx
```diff
@@ -1,483 +1,182 @@
...
-  `,
-} as const;
+const containerChildStyles = css`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+  width: 100%;
+  pointer-events: none;
+  z-index: 0;
```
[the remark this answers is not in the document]
brookescarlett: good catch, thanks

packages/button/src/Button.tsx
brookescarlett: it's easier to do this inline because we use the same spacing values when we need to apply margin-right and margin-left. this allows us to not have to duplicate the iconSpacing const

packages/button/src/Button.tsx
brookescarlett: 👍

packages/button/src/Button.tsx
brookescarlett: what do you mean by error? This just only sets the disabled attribute when the component isn't an anchor because a tags aren't supposed to have disabled set, just aria-disabled.

brookescarlett: afaict this isn't causing any errors (and this is recycled code from our old button implementation)

packages/button/src/Button.tsx
brookescarlett: What do you mean by building a component to handle this?

brookescarlett: I actually think that this makes it really clear what styles are being applied to the button and why (the function itself is very readable) as opposed to having a really messily composed styleset where you can't tell what prop is effecting what behavior

packages/button/src/Button.tsx
brookescarlett: yep 👍

packages/confirmation-modal/src/ConfirmationModal.tsx
brookescarlett: nope, i think simple margin-left is sufficient here

brookescarlett: but added the & selector

packages/ripple/src/registerRipple.ts
brookescarlett: absolutely

packages/button/src/Button.tsx
DesignerDave: You can probably cheat a little here with `:first/last-of-type` selectors so you don't need to duplicate the const.

packages/button/src/Button.tsx
DesignerDave: Can we address this issue through JavaScript? I want to make sure we're taking these TS errors seriously so we're handling edge cases we may not have thought of.

packages/button/src/Button.tsx
brookescarlett: i can't because what if only one is present? then it'll technically be the first of type even if it's on the right side

packages/button/src/Button.tsx
DesignerDave: Is `...(false)` valid? 🤔

brookescarlett: I don't see that in the HTML

packages/button/src/Button.tsx
DesignerDave:
```
&:first-of-type:last-of-type {
  margin-left: 0;
  margin-right: 0;
}
```

or

```
&:first-of-type:not(:last-of-type) {
  margin-right: 8px; // Or whatever it is
}
```

packages/button/src/Button.tsx
DesignerDave: It's very unclear what you're doing to the icon if you're trying to read `Button.tsx` directly. By moving this logic to a component, you can make all that logic related to cloning the glyph and such self-contained, and co-locate styles related to the glyphs with the glyph implementation itself while still keeping `Button.tsx from getting bloated.

packages/button/src/Button.tsx
DesignerDave: I wasn't aware that JavaScript will just ignore it when you spread a boolean. I would've thought it would give you a type error. Learn something new every day.

DesignerDave: (Carry on)

packages/button/src/Button.tsx
brookescarlett: Not really.. this has more to do with our typing of box, but if we want to be more explicit with the component's internals we'll still need an expect-error directive when setting as on the Box component itself.

packages/button/src/Button.tsx
brookescarlett: Created an Icon component, let me know what you think

packages/button/src/Button.tsx
brookescarlett: 👍

website/pages/component/button/example.tsx
```diff
@@ -21,45 +23,58 @@ const knobsConfig: KnobsConfigInterface<{
...
     default: 'Edit',
-    label: 'Glyph',
+    label: 'Left Glyph',
   },
-  children: {
-    type: 'text',
+  rightGlyph: {
+    type: 'select',
+    options: Object.keys(glyphs),
     default: 'Edit',
```
DesignerDave, line 58: Let's default the right glyph to something more realistic, such as CaretDown.

packages/palette/src/ui-colors.less
```diff
@@ -1,35 +1,38 @@
...
+@leafygreen__green--light-3: #e4f4e4;
 
-@leafygreen__gray--dark-3: #21313C;
-@leafygreen__gray--dark-2: #3D4F58;
-@leafygreen__gray--dark-1: #5D6C74;
-@leafygreen__gray--base: #89979B;
-@leafygreen__gray--light-1: #B8C4C2;
-@leafygreen__gray--light-2: #E7EEEC;
-@leafygreen__gray--light-3: #F9FBFA;
+@leafygreen__gray--dark-3: #21313c;
```
DesignerDave, line 8: Did auto formatting do this change?

DesignerDave (changes requested):
Couple more comments.

Also noticed something – the right icon should be aligned to the right of the button when width is set. This issue is super obvious on .design in our select inputs:

Screen Shot 2021-03-26 at 11 27 39 AM

brookescarlett:
> Couple more comments.
> 
> Also noticed something – the right icon should be aligned to the right of the button when width is set. This issue is super obvious on .design in our select inputs:
> 
> Screen Shot 2021-03-26 at 11 27 39 AM

This came from wrapping children in an HTML element, in an effort to clean up the styles around left versus right icon margin. I'm going to revert back to the original approach.

packages/palette/src/ui-colors.less
brookescarlett: yes

brookescarlett:
@DesignerDave responded to your feedback and also added the interaction ring on darkmode per Rob's specs. We might want to sync offline about this as for danger and dangerOutline he offered two options: dark3 and dark2 , and the primary/primaryOutline variants have an interaction ring that is barely visible.

website/pages/component/button/example.tsx
brookescarlett: 👍

packages/button/src/Button.tsx
DesignerDave: Huh... Alright, I don't think we need to shave any yaks here, just usually these errors are resolvable by checking if `rest.as` exists before statements that would rely on that.

DesignerDave (changes requested):

.changeset/tall-onions-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/button': major
+---
+
+Implements complete redesign of button component. For more information on how to upgrade to version 11, see our upgrade guide.
```
DesignerDave, line 5: Can we make this a little more detailed? Just a cliff's notes version like:

```
- Redesign's the button component! 
- Improves accessibility of buttons for visually-impaired users.
- Adds support for right-aligned glyphs.
- Renames glyph prop to `leftGlyph`.
```

That's not comprehensive, but folks should be able to get the main changes from the changelog still.

packages/button/UPGRADE.md
```diff
@@ -0,0 +1,20 @@
...
+
+Minor Changes to the API:
+
+- We now explicitly suppport icon-only buttons (different from our IconButton components). If you pass a glyph to either the `leftGlyph` or `rightGlyph` prop and do not pass any other children to the component, we will style the button accordingly.
+- We now support a `darkMode` prop. Each of the variants mentioned above comes in both light an dark modes.
+
+Breaking Changes to the Design:
+
+- Size:
+  - Small buttons were 25px, they are now 28px.
```
DesignerDave, line 19: 25px what? (Indicate height vs width)

packages/ripple/src/utils.ts
```diff
@@ -38,17 +38,23 @@ export interface Options {
...
     [Variant.Primary]: visualDesignPalette.green5,
-    [Variant.Info]: transparentize(buttonOpacity, visualDesignPalette.green3),
+    [Variant.PrimaryOutline]: transparentize(
+      buttonOpacity,
+      visualDesignPalette.green3,
+    ),
     [Variant.Default]: uiColors.gray.light2,
     [Variant.Danger]: visualDesignPalette.orange4,
-    [Variant.SecondaryDanger]: transparentize(buttonOpacity, uiColors.red.base),
+    [Variant.DangerOutline]: transparentize(buttonOpacity, uiColors.red.base),
```
DesignerDave, line 47: We may use the ripple package for non-buttons in the future. Shouldn't these colors be defined in the Button package instead? It seems strange to define the colors here.

DesignerDave (changes requested):
Couple more comments, then LGTM!

packages/button/UPGRADE.md
brookescarlett: 👍

DesignerDave (approved):

PR #875, opened 2021-03-26
export props used by Tabs

chickenbug:
## ✍️ Proposed changes
This PR exports the types used by `Tabs` so that consumers can use it. It exports AccessibleTabsProps aliased as TabsProps as that is the true type use by Tabs. Lmk if that sounds right or if we should just export the plain `TabsProps` as defined in Tabs or not do the alias.

## 🛠 Types of changes

## ✅ Checklist

Since this is a simple export I don't think we need unit tests. LMK if you think otherwise. I'm happy to add something.

## 🧪 How to test changes
This change can be tested by building the Tabs package and using the newly exported type.

### DIFF
--- a/packages/tabs/src/index.ts
+++ b/packages/tabs/src/index.ts
@@ -1,3 +1,3 @@
-export { default as Tabs } from './Tabs';
+export { default as Tabs, AccessibleTabsProps as TabsProps } from './Tabs';
 export { default as Tab } from './Tab';
 export type { TabProps } from './Tab';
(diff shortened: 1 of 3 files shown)


DesignerDave (approved):

DesignerDave (changes requested):
It appears this change is introducing a circular dependency between files. I'm looking into why this is the case.

packages/tabs/src/index.ts
```diff
@@ -1,3 +1,3 @@
-export { default as Tabs } from './Tabs';
+export { default as Tabs, AccessibleTabsProps as TabsProps } from './Tabs';
```
DesignerDave, line 1: Could you try changing this to:
```ts
export { default as Tabs } from './Tabs';
export type { AccessibleTabsProps as TabsProps } from './Tabs';
```

DesignerDave (approved):
LGTM!

PR #877, opened 2021-03-30
Popover upgrades

DesignerDave:
## ✍️ Proposed changes

This PR does 3 main things:
- Adds support for setting portal containers, including scrollable ones, to Popover and all components based on it through a consistent props interface.
- Improves the accessibility of the focus states for the side navigation.
- Introduces a prop for Popover and all components based on it to set z-indexes directly.

🎟 _Jira ticket:_ [Support zIndex prop for components using a Popover](https://jira.mongodb.org/browse/PD-729)
🎟 _Jira ticket:_ [Allow setting popover scroll containers](https://jira.mongodb.org/browse/PD-1278)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

This list of ways to test this PR isn't comprehensive, but should cover *most* use cases of this new and fixed functionality.

Side navigation focus states:
- Open https://mongodb.design, and tab through the side navigation on the site. All focus states should be higher contrast with the background now.

`zIndex` prop:
- Open storybook.
- For each popover component (Popover, Tooltip, Menu, Select), verify that adjusting the z-index knob updates the z-index CSS property on the popover being rendered.

MongoDB.design popover fix, and scrollContainer provider validation:
- Open MongoDB.design, and expand one of the select knobs.
- When you scroll the page, you should see the select dropdown scroll with it.

Portal container prop:
- For each popover component, with usePortal set to `true`, pass a reference to an HTML element to the popoverContainer prop.
- You should see the popover component render within that container.

LeafyGreenProvider portal container setting:
- When rendering a component using a popover within the provider, set `popover.portalContainer` to an element reference. You should see any popovers within that provider that don't locally have the `portalContainer` prop set be rendered within that container.


packages/popover/package.json
```diff
@@ -24,6 +24,7 @@
     "@leafygreen-ui/hooks": "^6.0.1",
     "@leafygreen-ui/lib": "^6.1.1",
     "@leafygreen-ui/portal": "^3.0.1",
+    "@leafygreen-ui/leafygreen-provider": "^2.0.2",
```
brookescarlett, line 27: should this be a peerDep?

packages/popover/src/Popover.story.tsx
```diff
@@ -115,6 +114,7 @@ function AdvancedExample() {
         usePortal={boolean('usePortal', true)}
         spacing={number('spacing', 10)}
         adjustOnMutation={boolean('adjustOnMutation', false)}
+        popoverZIndex={number('zIndex', 1)}
```
brookescarlett, line 117: should label this with the prop name

packages/popover/src/Popover.story.tsx
```diff
@@ -123,6 +123,65 @@ function AdvancedExample() {
...
+  const buttonRef = useRef(null);
+
+  return (
+    <>
+      <div
+        ref={el => setPortalContainer(el)}
+        className={css`
+          margin: 150px;
+          width: 100%;
+          overflow-y: auto;
```
brookescarlett, line 140: not urgent but this could be a className

packages/popover/src/Popover.tsx
```diff
@@ -73,6 +81,11 @@ function Popover({
   const [contentNode, setContentNode] = useState<HTMLElement | null>(null);
   const [forceUpdateCounter, setForceUpdateCounter] = useState(0);
 
+  let { portalContainer, scrollContainer } = usePopoverPortalContainer();
+
+  portalContainer = portalContainerProp || portalContainer;
```
brookescarlett, line 86: why || over ??

packages/popover/src/Popover.tsx
```diff
@@ -203,7 +217,11 @@ function Popover({
   `;
 
   const Root = usePortal ? Portal : Fragment;
-  const rootProps = usePortal ? { className: portalClassName } : {};
+  const rootProps = usePortal
+    ? portalContainer
+      ? { container: portalContainer }
+      : { className: portalClassName ?? undefined }
```
brookescarlett, line 223: do we need to explicitly write undefined here? if portalClassName is undefined won't it evaluate as such without needing to provide a fallback

packages/select/src/Select.tsx
```diff
@@ -73,18 +74,22 @@ export default function Select({
   darkMode = false,
   size = Size.Default,
   disabled = false,
-  usePortal = true,
   placeholder = 'Select',
   className,
   id: idProp,
   label,
+  'aria-labelledby': ariaLabelledBy,
```
brookescarlett, line 81: We should be consistent about this. I don't feel sooo strongly if the B should or shouldn't be capitalized (I've been using lc because it more closely parallels the html attribute) but just a nit that we should try and standardize

packages/side-nav/src/CollapseToggle.tsx
```diff
@@ -49,7 +49,7 @@ const buttonFocusStyles = css`
     color: ${uiColors.blue.base};
     border-color: transparent;
     box-shadow: 0 3px 4px ${transparentize(0.9, uiColors.black)},
-      0 0 0 3px ${uiColors.blue.light1};
+      0 0 0 3px ${uiColors.focus};
```
brookescarlett, line 52: 👍

website/components/live-example/Knobs.tsx
```diff
@@ -244,34 +238,34 @@ function Area({ onChange, label, value, prop, darkMode }: TextInterface) {
...
-    <div
-      className={cx(
-        knobContainerStyle,
-        css`
-          border-top: 1px solid
-            ${darkMode ? uiColors.gray.dark2 : uiColors.gray.light2};
-        `,
-      )}
-    >
+    <KnobRow darkMode={darkMode}>
```
brookescarlett, line 241: nit: should we just turn the composed classname into a `getClassName` function rather than adding another component?

packages/menu/README.md
```diff
@@ -173,6 +173,11 @@ import { Menu, MenuSeparator, MenuItem } from '@leafygreen-ui/menu';
...
+| `scrollContainer`  | `HTMLElement` \| `null`                  | If the popover portal has a scrollable ancestor other than the window, this prop allows passing a reference to that lement to allow the portal to position properly.                                                                                                    |              |
+| `portalClassName`  | `string`                                 | Passes the given className to the popover's portal container if the default portal container is being used.                                                                                                                                                             |              |
+| `popoverZIndex`    | `number`                                 | Sets the z-index CSS property for the popover.                                                                                                                                                                                                                          |              |
 
 _Any other properties will be spread on the Menu `ul` container_
```
brookescarlett, line 182: did we move this into the prop table for most other components?

packages/popover/package.json
DesignerDave: Nope – we consume a hook from that package.

website/components/live-example/Knobs.tsx
DesignerDave: I don't like getClassName functions in general 😄 I'm applying the same feedback I gave on one of your recent PRs here,

packages/menu/README.md
DesignerDave: Not sure – I didn't change it though.

packages/popover/src/Popover.tsx
DesignerDave: I believe this was to make TS happy.

packages/popover/src/Popover.tsx
DesignerDave: I'd turn the question back around, and ask you why `??` would be better here?

portalContainerProp should always be an HTMLElement, null, or undefined. Since it isn't typed as a string, number, etc, there isn't a risk of coercion causing this to evaluate any differently than `??` would. If portalContainerProp is something other than what it's typed to be, neither operator is likely to evaluate differently. This operator in marginal cases may even be more likely to be correct when the incorrect type is passed here.

Example:
```
const portalContainerProp = false;
portalContainer = portalContainerProp || portalContainer; // Equals portalContainer
portalContainer = portalContainerProp ?? portalContainer; // Equals false
```

In the above unlikely case, `||` leads to the appropriate value while `??` does not. Granted they're both incorrect if the value is `true`.

Another example:
```
const portalContainerProp = 0;
portalContainer = portalContainerProp || portalContainer; // Equals portalContainer
portalContainer = portalContainerProp ?? portalContainer; // Equals 0
```

Both examples should never happen, and ultimately the best case would be for this to throw an error. If we're thinking practically here, using `??` or `||` shouldn't actually matter in this case.

packages/popover/src/Popover.story.tsx
DesignerDave: For css declarations in storybook that aren't shared across pages, I don't mind inline TBH. Let me know if you feel strongly about this though and I can update it.

packages/select/src/Select.tsx
DesignerDave: I didn't change this in this PR, but I think my PR's changes to the props caused your naming change to be reverted. I'll fix.

brookescarlett:
SideNav Focus state doesn't look right to me: 
Screen Shot 2021-04-09 at 1 52 42 PM
Screen Shot 2021-04-09 at 1 52 37 PM

packages/menu/README.md
brookescarlett: Can we update while we're in here?

packages/popover/src/Popover.story.tsx
brookescarlett: that's okay

packages/popover/src/Popover.tsx
brookescarlett: wfm

packages/select/src/Select.tsx
brookescarlett: thank you

website/layouts/BaseLayout.tsx
```diff
@@ -44,16 +47,30 @@ const childrenWrapper = css`
...
+        portalContainer: bodyContainerRef,
+      }}
+    >
+      <LayoutContext.Provider value={bodyContainerRef}>
+        <div className={containerStyle}>
+          <Navigation />
 
-      <div className={layout}>
-        <div className={childrenWrapper}>{children}</div>
+          <div className={layout} ref={el => setBodyContainerRef(el)}>
```
brookescarlett, line 66: nit: setBodyContainerRef

packages/menu/README.md
```diff
@@ -173,6 +173,11 @@ import { Menu, MenuSeparator, MenuItem } from '@leafygreen-ui/menu';
...
 | `adjustOnMutation` | `boolean`                                | Determines whether or not the `<Menu/>` should reposition itself based on changes to `trigger` or reference element position.                                                                                                                                           | `false`      |
+| `usePortal`        | `boolean`                                | Determines if the Menu will be rendered within a portal.                                                                                                                                                                                                                | `true`       |
+| `portalContainer`  | `HTMLElement` \| `null`                  | Sets the container used for the popover's portal.                                                                                                                                                                                                                       |              |
```
brookescarlett, line 177: I think we prefer ',' to \| now

packages/popover/README.md
```diff
@@ -60,16 +60,20 @@ The popover component will be automatically positioned relative to its nearest p
 ## Properties
 
 | Prop               | Type                                                                                           | Description                                                                                                                                                                                                              | Default    |
-| ------------------ | ---------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ---------- |
+| ------------------ | ---------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ---------- | --- |
```
brookescarlett, line 63: did we accidently add an extra column here?

packages/popover/src/Popover.story.tsx
brookescarlett: 'popoverZinex'

DesignerDave: Sure

packages/menu/README.md
DesignerDave: We have 3+ different ways we do this across our components. There isn't currently a standard that appears to be adopted by the majority of components.

I'm gonna leave this out for now since it seems better suited to a blanket update across our components (that we may or may not care about making at all in the existing README's if we're gonna eventually move the content entirely to .design).

packages/popover/README.md
DesignerDave: Yep, fixed 👍

packages/menu/README.md
DesignerDave: Good call, fixed

website/layouts/BaseLayout.tsx
DesignerDave: Done

packages/select/src/Select.tsx
DesignerDave: Done

packages/leafygreen-provider/src/PortalContext.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { createContext, useContext } from 'react';
```
brookescarlett, line 1: Can we get a spec file for PortalContext?

brookescarlett (approved):
Seems to work as intended to me!

PR #882, opened 2021-03-31
PD-1266: Add Language Switcher

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1266](https://jira.mongodb.org/browse/PD-1266)

## 🛠 Types of changes

## ✅ Checklist


DesignerDave (changes requested):
Few pieces of feedback that I can give here:
1. It looks like when we render the language switcher, the code example itself is added to the tabindex and is focusable. This isn't the case with the regular copyable version of the code component.
2. We render a "Select" option in the select for the language switcher, but clicking it doesn't do anything for the use-case (or in storybook for that matter).
3. The copy button looks a little bit far from the right edge. Is that margin to the design spec?
4. Nit, but can we add the LeafyGreen provider to the storybook examples?

brookescarlett:
> Few pieces of feedback that I can give here:
> 
> 1. It looks like when we render the language switcher, the code example itself is added to the tabindex and is focusable. This isn't the case with the regular copyable version of the code component.
- I am not able to reproduce this locally, do you mind sharing a screenshot of what you're seeing?
> 2. We render a "Select" option in the select for the language switcher, but clicking it doesn't do anything for the use-case (or in storybook for that matter).
- This is a bug in the Select component, I agree we should fix this but I do think that it's outside of the scope of this PR
> 3. The copy button looks a little bit far from the right edge. Is that margin to the design spec?
- I will double check
> 4. Nit, but can we add the LeafyGreen provider to the storybook examples?
- Done

DesignerDave (approved):

PR #930, opened 2021-05-03
PD-1122: Makes updates to SideNav to be compatible with Docs

brookescarlett:
## ✍️ Proposed changes

Adds baseFontSize prop to SideNav
Adds manual collapsed/setCollapsed functionality
Exposes sideNavSelector
Supports customizable width
Supports nested SideNavItems

🎟 _Jira ticket:_ [PD-1122](https://jira.mongodb.org/browse/PD-1122)

## 🛠 Types of changes

## ✅ Checklist


packages/side-nav/README.md
```diff
@@ -155,10 +155,14 @@ return (
...
-| `children`  | `node`   | Content that will be rendered inside the root-level element. | `undefined` |
+| Prop            | Type                                            | Description                                                                     | Default     |
+| --------------- | ----------------------------------------------- | ------------------------------------------------------------------------------- | ----------- |
+| `className`     | `string`                                        | Class name that will be applied to the root-level element.                      | `undefined` |
+| `children`      | `node`                                          | Content that will be rendered inside the root-level element.                    | `undefined` |
+| `baseFontSize`  | `14`, `16`                                      | Determines the base font size (in pixels) of the Side Nav                       | `14`        |
+| `widthOverride` | `number`                                        | Width (in pixels) of Side Navigation.                                           | `184`       |
```
DesignerDave, line 163: We should call out these docs-specific props as not being for general use. The last thing we want is for one of the cloud products to override the width.

Also, should we allow this to be set to a percentage? E.g. if they pass a string, we expect the unit to be set (`px`, `%`, `vw`, etc), and if they pass a number, it's interpreted as pixels.

packages/side-nav/src/SideNavItem.tsx
```diff
@@ -61,6 +60,17 @@ const defaultStyle = css`
...
 `;
 
+const typographyStyle = {
+  [14]: css`
+    font-size: 14px;
+    line-height: 1em;
+  `,
+  [16]: css`
+    font-size: 16px;
+    line-height: 1.5em;
```
DesignerDave, line 70: `em` units are proportional, so 1em should work in both cases here. Why do the sizes need different line height definitions?

DesignerDave (changes requested):
Left a couple comments, but outside of those there's one large piece of behavior that seems to be missing here.

When a nested item is active, its descendent items need to become expanded. Otherwise, the currently active navigation item could be hidden.

I'd also double-check to see if the entire tree needs to expand when a parent is active. It might only need to be the direct children of an active item that need to be visible.

packages/side-nav/src/SideNavItem.tsx
brookescarlett: The pattern we've defined is 14px => 20px line height; 16px => 24px line height

packages/side-nav/README.md
brookescarlett: Should I leave them out of the docs? If not, I can write *Docs only*? What do you think?

DesignerDave (approved):

PR #931, opened 2021-05-03
PD-1316: Select updates

brookescarlett:
## ✍️ Proposed changes

- Adds provider to select component
- Updates visuals for focus state
- Updates focus behavior to match wai spec: Tabs moves focus to next focusable element and closes the menu, Escape closes the menu

🎟 _Jira ticket:_ [PD-1316](https://jira.mongodb.org/browse/PD-1316)

## 🛠 Types of changes

## ✅ Checklist


packages/select/src/Option.tsx
```diff
@@ -23,6 +24,20 @@ const optionStyle = css`
   outline: none;
   overflow-wrap: anywhere;
   transition: background-color 150ms ease-in-out;
+  position: relative;
+
+  &:before {
```
DesignerDave, line 29: We need to transition this element in like we do in the side navigation. IIRC it's a scale transform transition on the vertical axis combined with the opacity transition.

DesignerDave (changes requested):
Hey Brooke, thanks for this!

I noted an issue where the focus state doesn't transition consistently when compared to the same focus state in the side navigation.

Additionally, I found a bug with keyboard interactions:
1. Tab to the Select component, and press enter or space to open the dropdown.
2. Use the arrow keys to focus on "Hamster", but don't make the selection.
3. Press the escape key to close the dropdown. "Cat" should still be selected at this point, and the Select component should still be focused.
4. Press enter to open the dropdown again. The dropdown will open, and "Hamster" will become selected. 

Desired behavior: The dropdown should open without selecting a previously focused item.

packages/select/src/Option.tsx
brookescarlett: Done!

DesignerDave (changes requested):
Looks like keyboard navigation now has a different problem:

1. Click the select to open the dropdown.
2. Use the arrow keys to move the focus state down to a currently unselected item.
3. Press enter. Your selection is made and the dropdown closes, but then it immediately re-opens.

Also, it appears you can't currently make a selection in the dropdown with the space key.

DesignerDave (approved):
Noted offline that there's an issue with .design when building. Otherwise this LGTM!

PR #940, opened 2021-05-11
PD-1281: Updates menuitem styles

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1281](https://jira.mongodb.org/browse/PD-1281)

## 🛠 Types of changes

## ✅ Checklist


packages/menu/src/MenuItem.tsx
```diff
@@ -165,6 +165,12 @@ const MenuItem: ExtendableBox<
         ),
       });
 
+    const textStyles = cx(titleTextStyle, {
```
DesignerDave, line 168: What's the value in separating this variable out, rather than incorporating it into the existing `cx()`? We even already have handling for active, disabled, and focus there that we could merge this with.

DesignerDave (changes requested):
Do we have a clear understanding of the implications of this change? The ticket for this was written with some uncertainty in that regard, so I'm a bit cautious about this one.

brookescarlett:
> r understanding of the implications of this change? The ticket for this was written with some uncertainty in that regard, so I'm a bit cautious about this one.

I believe the point was just to make it easier to update the text color of a MenuItem through the className prop, and this Pr achieves against that goal. If there is something that I'm missing let me know!

DesignerDave:
@brookescarlett What's unclear to me is if this will have a downstream impact on application style overrides overall.

PR #953, opened 2021-05-25
Use context instead of cloneElement in RadioBox

jetpacmonkey:
## ✍️ Proposed changes

`RadioBoxGroup` using `cloneElement` to communicate with its child `RadioBox` elements means that those `RadioBox` elements have to be direct children, and it's not possible for consumers to make their own wrapper components. Communicating through context instead removes this limitation and allows the `RadioBox` to be deeper in the React tree.

🎟 _Jira ticket:_ Tangentially related to [CLOUDP-90548](https://jira.mongodb.org/browse/CLOUDP-90548)

## 🛠 Types of changes

## ✅ Checklist

## 💬 Further comments

This is as much a discussion-starter as a specific proposed change. I've argued against `cloneElement` for a while, so we don't have to rehash all of that, but I do think that using context allows us to use React more naturally, without having to worry about the additional constraints brought on by a parent component meddling with its children's props.

### DIFF
--- a/packages/radio-box-group/src/context.ts
+++ b/packages/radio-box-group/src/context.ts
@@ -1,0 +1,19 @@
+import { createContext, ChangeEventHandler, useContext } from 'react';
+import Size from './Size';
+
+export interface RadioBoxGroupContext {
+  value: string | number | undefined;
+  onChange: ChangeEventHandler<HTMLInputElement>;
+  getNextId: () => string;
+  size: Size | undefined;
+  name: string | undefined;
+}
+
+const context = createContext<RadioBoxGroupContext | null>(null);
+
+export const Provider = context.Provider;
+export const Consumer = context.Consumer;
+
+export function useRadioBoxGroupContext() {
+  return useContext(context);
+}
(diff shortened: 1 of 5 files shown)


packages/radio-box-group/src/context.ts
```diff
@@ -0,0 +1,19 @@
...
+  onChange: ChangeEventHandler<HTMLInputElement>;
+  getNextId: () => string;
+  size: Size | undefined;
+  name: string | undefined;
+}
+
+const context = createContext<RadioBoxGroupContext | null>(null);
+
+export const Provider = context.Provider;
+export const Consumer = context.Consumer;
```
brookescarlett, line 15: Is this being used anywhere? I can't find it 👀

jetpacmonkey: It's not, I added this export before I checked if `RadioBox` was a function or a class component

brookescarlett:
This is looking great to me! @DesignerDave did you want to take a look as well

DesignerDave:
@brookescarlett If it looks good to you, let's ship it 😄

packages/radio-box-group/src/context.ts
brookescarlett: We can rm since its not :)

brookescarlett (approved):

jetpacmonkey:
Thanks for the quick review!

PR #974, opened 2021-06-10
PD-1418: Updates Tooltip within Copyable

TheSonOfThomp:
## ✍️ Proposed changes

Resolves Tooltip presentation issues when using a Copyable within a Modal

🎟 _Jira ticket:_ [PD-1418](https://jira.mongodb.org/browse/PD-1418)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

Use a Copyable within a Modal with `z-index > 0`:
Create a temporary Story within `Copyable.story.tsx` 
```js
  .add('Within Modal', () => {
    const darkMode = boolean('Dark mode', false)
    return (
      <Modal
        open={true}
        size='default'
        darkMode={darkMode}
        style={{ zIndex: 1 }}
      >
        <Copyable
          label={text('Label', 'Label')}
          description={text('Description', '') || undefined}
          darkMode={darkMode}
          shouldTooltipUsePortal={false}
        >
          {text('Text', 'npm install @leafygreen-ui/copyable')}
        </Copyable>
      </Modal>
    )
  })
```

### Before
Before

### After
After

### DIFF
--- a/.changeset/cyan-turkeys-run.md
+++ b/.changeset/cyan-turkeys-run.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/copyable': patch
+---
+
+Fixes Copyable Tooltip when used within a Modal
--- a/.changeset/cyan-turkeys-run.md
+++ b/.changeset/cyan-turkeys-run.md
@@ -2,4 +2,4 @@
 '@leafygreen-ui/copyable': patch
 ---
 
-Fixes Copyable Tooltip when used within a Modal
+Tooltips in association with Copyable component now appear as a direct DOM child—not within a portal—in order to support usage within a Modal
--- a/.changeset/cyan-turkeys-run.md
+++ b/.changeset/cyan-turkeys-run.md
@@ -2,4 +2,4 @@
 '@leafygreen-ui/copyable': patch
 ---
 
-Tooltips in association with Copyable component now appear as a direct DOM child—not within a portal—in order to support usage within a Modal
+Exposes a `shouldTooltipUsePortal` prop on Copyable, to control whether the "Copied!" Tooltip is rendered as a direct DOM child, or within a portal, in order to support usage within a Modal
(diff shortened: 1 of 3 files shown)


.changeset/cyan-turkeys-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/copyable': patch
+---
+
+Fixes Copyable Tooltip when used within a Modal
```
brookescarlett, line 5: Let's be a little clearer in the changeset: 

something like: Tooltips that now appear in association with Copyable component are no longer portaled, in order to support usage within a Modal

packages/copyable/src/Copyable.tsx
```diff
@@ -178,6 +178,7 @@ export default function Copyable({
         justify={Justify.Middle}
         trigger={trigger}
         triggerEvent={TriggerEvent.Click}
+        usePortal={false}
```
TheSonOfThomp, line 181: Are there any cases where the copyable tooltip _should_ use a portal?

brookescarlett: We might want to expose a prop here:

shouldPortalTooltip comes to mind but open to others, we can default to `false` as you have here.

brookescarlett: Actually, let's default to `true`, in which case this is only a minor change rather than a breaking one.

brookescarlett:
Also, @TheSonOfThomp, don't forget to run `yarn changeset`

packages/copyable/src/Copyable.tsx
brookescarlett: (that's also more in line with how we default components that consume our Portal component)

brookescarlett:
We just need to document the new prop in the component's readme!

brookescarlett (approved):

PR #980, opened 2021-06-22
PD-1390 tooltip trigger on click

TheSonOfThomp:
## ✍️ Proposed changes

Fixes a bug where event handlers passed into a Tooltip trigger were not called with the correct event object (or at all)

🎟 _Jira ticket:_ [PD-1390](https://jira.mongodb.org/browse/PD-1390)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

Add an `onClick`, `onMouseEnter`, `onMouseLeave`, `onFocus`, or `onBlur` event handler to the `trigger` element of a Tooltip. These functions should run with the correct `event` in local scope.

### DIFF
--- a/packages/tooltip/src/Tooltip.tsx
+++ b/packages/tooltip/src/Tooltip.tsx
@@ -296,7 +296,11 @@ function Tooltip({
 
       function triggerHandler(handler: string, e: MouseEvent): void {
         // call any click handlers already on the trigger
-        if (triggerProps[handler] && typeof triggerProps[handler] == 'function')
+        if (
+          triggerProps &&
+          triggerProps[handler] &&
+          typeof triggerProps[handler] == 'function'
+        )
           triggerProps[handler](e);
       }
     },
--- a/packages/tooltip/src/Tooltip.tsx
+++ b/packages/tooltip/src/Tooltip.tsx
@@ -272,7 +272,7 @@ function Tooltip({
               handleClose();
             }, 35),
             onFocus: (e: MouseEvent) => {
-              triggerHandler('onMouseLeave', e);
+              triggerHandler('onFocus', e);
               setOpen(true);
             },
             onBlur: (e: MouseEvent) => {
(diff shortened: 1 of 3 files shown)


packages/tooltip/src/Tooltip.tsx
```diff
@@ -259,38 +260,48 @@ function Tooltip({
...
+              handleClose();
+            },
+          };
+        case TriggerEvent.Click:
+        default:
+          return {
+            onClick: (e: MouseEvent) => {
+              // ensure that we don't close the tooltip when content inside tooltip is clicked
+              if (e.target !== tooltipNode) {
+                triggerHandler('onClick', e);
```
TheSonOfThomp, line 289: This is the core fix for the bug

packages/tooltip/src/Tooltip.tsx
```diff
@@ -259,38 +260,48 @@ function Tooltip({
...
+              handleClose();
+            }, 35),
+            onFocus: (e: MouseEvent) => {
+              triggerHandler('onFocus', e);
+              setOpen(true);
+            },
+            onBlur: (e: MouseEvent) => {
+              triggerHandler('onBlur', e);
+              handleClose();
+            },
```
TheSonOfThomp, line 281: Calls handlers for all other event types to prevent future bugs

packages/tooltip/src/Tooltip.tsx
```diff
@@ -259,38 +260,48 @@ function Tooltip({
 
   const createTriggerProps = useCallback(
     (triggerEvent: TriggerEvent, triggerProps?: any) => {
-      if (triggerEvent === TriggerEvent.Hover) {
-        return {
-          ref: setTriggerNode,
```
TheSonOfThomp, line 264: This PR removes `ref` prop since it was causing some tests to fail. I may want to look into that deeper in the future

DesignerDave: Everything seems to be working, but lines 317-318 have a comment related to the triggerNode. Are we able to remove the usage of triggerNode if we aren't ever setting it?

TheSonOfThomp: Removed references to `triggerNode` (see 312-320). Doesn't seem like it's needed elsewhere

DesignerDave (approved):
LGTM

PR #982, opened 2021-06-23
PD-1419 Ensures disabled links don't navigate

TheSonOfThomp:
## ✍️ Proposed changes

Fixes a bug where a Button with an `href` and `disabled=true` would still navigate.

🎟 _Jira ticket:_ [PD-1419](https://jira.mongodb.org/browse/PD-1419)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

In Storybook, set the `disabled` prop to true, and set the `href` prop to anything other than `undefined`. Clicking the button should not navigate.

### DIFF
--- a/packages/button/src/Button.spec.tsx
+++ b/packages/button/src/Button.spec.tsx
@@ -103,7 +103,7 @@ describe('packages/button', () => {
         href: 'http://mongodb.design',
         disabled: true,
       });
-      expect(button.tagName.toLowerCase()).toBe('div');
+      expect(button.tagName.toLowerCase()).toBe('button');
     });
 
     test(`renders component inside of a React Element/HTML tag based on as prop`, () => {
--- a/packages/button/src/Button.spec.tsx
+++ b/packages/button/src/Button.spec.tsx
@@ -5,7 +5,6 @@ import { BoxProps } from '@leafygreen-ui/box';
 import Button from './Button';
 import { ButtonProps } from './types';
 
-const onClick = jest.fn();
 const className = 'test-button-class';
 const title = 'Test button title';
 const child = 'Button child';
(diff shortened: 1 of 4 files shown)


packages/button/src/Button.spec.tsx
```diff
@@ -37,89 +36,120 @@ describe('packages/button', () => {
     });
   });
 
-  test(`renders "${className}" in the component's markup`, () => {
-    const { button } = renderButton({
-      className,
+  describe('rendering', () => {
```
TheSonOfThomp, line 39: Re-organized tests to keep like tests together

packages/button/src/Button.spec.tsx
```diff
@@ -37,89 +36,120 @@ describe('packages/button', () => {
...
 
-  test(`renders component inside of a tag when "href" prop is set`, () => {
-    const { button } = renderButton({
-      href: 'http://mongodb.design',
+    test(`renders component inside of \`div\` tag when "href" prop is set, but "disabled" is true`, () => {
+      const { button } = renderButton({
+        href: 'http://mongodb.design',
+        disabled: true,
+      });
+      expect(button.tagName.toLowerCase()).toBe('div');
```
TheSonOfThomp, line 106: New test

packages/button/src/Button.spec.tsx
```diff
@@ -37,89 +36,120 @@ describe('packages/button', () => {
...
+      fireEvent.click(button);
+      expect(onClick.mock.calls.length).toBe(0);
+    });
+
+    test('href attribute exists on a link', () => {
+      const href = 'https://mongodb.design';
+      const { button } = renderButton({
+        href,
+      });
+      expect(button).toHaveAttribute('href', href);
```
TheSonOfThomp, line 151: New tests

packages/button/src/Button.tsx
```diff
@@ -107,13 +107,23 @@ const Button: ExtendableBox<
...
 
+  // If the link is disabled, render as a div
+  const getButtonTag = (
+    isAnchor: boolean,
+    disabled: boolean,
+  ): 'a' | 'button' | 'div' => {
+    if (!isAnchor) return 'button';
+    if (disabled) return 'div';
+    return 'a';
+  };
```
TheSonOfThomp, line 118: Since we can't disable a link with the `disabled` attribute, we have to change the tag type to a `div` to make ensure clicking the button won't navigate

TheSonOfThomp: Are there any a11y concerns with keeping an `href` attribute on a `div` (and expecting it _not_ to work)?

DesignerDave: Hmm... since disabled buttons should still be focusable when disabled for accessibility reasons, I think changing this to a div is probably not the best solution.

I've actually seen some examples of rendering a button instead of a link in these cases, and handling `disabled` in the same way that buttons do. That seems like a stronger approach for accessibility reasons.

TheSonOfThomp: That makes a lot of sense. Putting an `href` attribute on a `button` is still strange, but no more strange than on a `div`

DesignerDave (approved):
LGTM!

PR #989, opened 2021-07-01
PD-1128 Fixes dropdown height

TheSonOfThomp:
## ✍️ Proposed changes

Fixes a bug where a dropdown menu would squish to the space available below the trigger instead of moving above the trigger. Also sets a max-height on the select menu.

🎟 _Jira ticket:_ [PD-1128](https://jira.mongodb.org/browse/PD-1128)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

Add the following story to Storybook:

```
  .add('Alignment test', () => {
    const darkMode = boolean('darkMode', false);
    const alignment = select('Button placement', ['top', 'bottom'], 'bottom')
    const [value, setValue] = useState(null);

    const options = [
      'Apple',
      'Banana',
      'Carrot',
      'Dragonfruit',
      'Eggplant',
      'Fig',
      'Grape',
      'Honeydew',
      'Iceberg',
      'Jelly bean',
      'Kiwi',
      'Lemon',
      'Mango',
      'Nut',
      'Orange',
      'Pear',
      'Quince',
      'Raspberry',
      'Strawberry',
      'Tomato',
      'Uglifruit',
      'Vanilla',
      'Watermelon',
      'Ximenia',
      'Yam',
      'Zucchini',
    ];

    return (
      <div
        className={css`
          position: relative;
          width: 100%;
          height: 100vh;
          background-color: ${darkMode ? uiColors.gray.dark3 : uiColors.white};
          display: flex;
          justify-content: center;
          align-items: ${alignment == 'top' ? 'flex-start' : 'flex-end'};
          align-content: center;
        `}
      >
      <LeafygreenProvider>
        <Select
          aria-labelledby={'Label'}
          onChange={setValue}
          value={value}
          className={css`
            min-width: 144px;
            align-self: ${alignment == 'top' ? 'flex-start' : 'flex-end'};
            margin: 16px auto;
          `}
        >
          {options
            .slice(0, number('Options', 5, {min: 0, max: 25}))
            .map(option => (
            <Option key={option} value={option}>
              {option}
            </Option>
          ))}
        </Select>
      </LeafygreenProvider>
      </div>
    );
  });
```

### DIFF
--- a/packages/select/src/ListMenu.tsx
+++ b/packages/select/src/ListMenu.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useContext } from 'react';
+import React, { useCallback, useContext, useMemo } from 'react';
 import { css, cx } from '@leafygreen-ui/emotion';
 import { useViewportSize } from '@leafygreen-ui/hooks';
 import Popover, { Align, Justify, PopoverProps } from '@leafygreen-ui/popover';
@@ -45,6 +45,11 @@ const ListMenu = React.forwardRef<HTMLUListElement, ListMenuProps>(
   ) {
     const { mode, size, disabled, open } = useContext(SelectContext);
 
+    const { length: optionsCount } = useMemo(
+      () => React.Children.toArray(children),
+      [children],
+    );
+
     const colorSet = colorSets[mode];
     const sizeSet = sizeSets[size];
 
@@ -56,6 +61,8 @@ const ListMenu = React.forwardRef<HTMLUListElement, ListMenuProps>(
       viewportSize === null || ref.current === null
         ? 0
         : viewportSize.height - ref.current.getBoundingClientRect().top - 10;
+
+    const minHeight = sizeSets[size].height * optionsCount;
 
     const onClick = useCallback(
       (event: React.MouseEvent) => {
@@ -98,6 +105,7 @@ const ListMenu = React.forwardRef<HTMLUListElement, ListMenuProps>(
             css`
               font-family: ${fontFamilies.default};
               font-size: ${sizeSet.option.text}px;
+              min-height: ${minHeight}px;
               max-height: ${maxHeight}px;
               background-color: ${colorSet.option.background.base};
               box-shadow: 0 3px 7px 0 ${colorSet.menu.shadow};
--- a/packages/select/src/ListMenu.tsx
+++ b/packages/select/src/ListMenu.tsx
@@ -1,11 +1,12 @@
-import React, { useCallback, useContext, useMemo } from 'react';
+import React, { useCallback, useContext } from 'react';
 import { css, cx } from '@leafygreen-ui/emotion';
-import { useViewportSize } from '@leafygreen-ui/hooks';
 import Popover, { Align, Justify, PopoverProps } from '@leafygreen-ui/popover';
 import { breakpoints, fontFamilies } from '@leafygreen-ui/tokens';
 import SelectContext from './SelectContext';
 import { colorSets, mobileSizeSet, sizeSets } from './styleSets';
 import { useForwardedRef } from './utils';
+
+const MAX_MENU_HEIGHT = 274;
 
 const menuStyle = css`
   position: relative;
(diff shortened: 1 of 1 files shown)


packages/select/src/ListMenu.tsx
```diff
@@ -98,7 +91,8 @@ const ListMenu = React.forwardRef<HTMLUListElement, ListMenuProps>(
             css`
               font-family: ${fontFamilies.default};
               font-size: ${sizeSet.option.text}px;
-              max-height: ${maxHeight}px;
+              min-height: ${sizeSet.height}px;
+              max-height: ${MAX_MENU_HEIGHT}px;
```
TheSonOfThomp, line 95: All that time wrapping my head around how the Select worked, and this is the solution

DesignerDave (commented):
Largely this PR does what it sets out to do, but I'm noticing we are losing one thing in this PR – there's no adjustment for window size at all anymore, so it's easy for the dropdown to appear partially scrolled out of view. Any thoughts around that?

DesignerDave:
Hmm, I'm getting this error trying to run this locally. I've done a full `yarn clean` and `yarn run init` with no luck.

```
Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
Error: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
    at StyleSheet.insert (http://localhost:9001/vendors~main.c00cccda8396c611da88.bundle.js:322:22)
    at toSheet (http://localhost:9001/vendors~main.c00cccda8396c611da88.bundle.js:27:19)
    at Array.forEach (<anonymous>)
    at Function.ruleSheet (http://localhost:9001/vendors~main.c00cccda8396c611da88.bundle.js:89:31)
    at H (http://localhost:9001/vendors~main.c00cccda8396c611da88.bundle.js:4712:24)
    at B (http://localhost:9001/vendors~main.c00cccda8396c611da88.bundle.js:4761:19)
    at Object.insert (http://localhost:9001/vendors~main.c00cccda8396c611da88.bundle.js:152:7)
    at insertStyles (http://localhost:9001/vendors~main.c00cccda8396c611da88.bundle.js:46425:31)
    at css (http://localhost:9001/vendors~main.c00cccda8396c611da88.bundle.js:46274:72)
    at Object../packages/side-nav/src/styles.ts (http://localhost:9001/main.c00cccda8396c611da88.bundle.js:41235:41)
```

packages/select/src/ListMenu.tsx
```diff
@@ -44,18 +48,33 @@ const ListMenu = React.forwardRef<HTMLUListElement, ListMenuProps>(
...
+        // Find out how much space is available above or below the trigger
+        const safeSpace = Math.max(
+          viewportSize.height - triggerBottom,
+          triggerTop,
+        );
+
+        // if there's more than enough space, set to MAX_MENU_HEIGHT
+        // otherwise fill the space available
+        return Math.min(MAX_MENU_HEIGHT, safeSpace - MENU_MARGIN);
+      } else {
```
DesignerDave, line 74: Small nit, but one standard we've applied fairly consistently in LeafyGreen is to use happy-path coding techniques. So rather than what you have here, it would move the else block return out of the conditional:

```js
const maxHeight = useMemo(() => {
    if (viewportSize && ref.current && referenceElement.current) {
      /* code here */
      return Math.min(MAX_MENU_HEIGHT, safeSpace - MENU_MARGIN);
    } 

    return MAX_MENU_HEIGHT;
}, [ref, referenceElement, viewportSize]);
```

packages/select/src/ListMenu.tsx
```diff
@@ -1,11 +1,15 @@
...
-import { useViewportSize } from '@leafygreen-ui/hooks';
 import Popover, { Align, Justify, PopoverProps } from '@leafygreen-ui/popover';
 import { breakpoints, fontFamilies } from '@leafygreen-ui/tokens';
+import { useViewportSize } from '@leafygreen-ui/hooks';
 import SelectContext from './SelectContext';
 import { colorSets, mobileSizeSet, sizeSets } from './styleSets';
 import { useForwardedRef } from './utils';
+import { useMemo } from 'react';
+
+const MAX_MENU_HEIGHT = 274;
```
DesignerDave, line 11: Is there a reason to use snake-case instead of camel-case here?

DesignerDave (changes requested):
A couple smaller stylistic nits, but this PR is now working for me!

packages/select/src/ListMenu.tsx
TheSonOfThomp: None other than I've typically used all caps snake case for config constants like this

packages/select/src/ListMenu.tsx
TheSonOfThomp: Makes sense! I'll make that update

packages/select/src/ListMenu.tsx
DesignerDave: We use camelcase for sizing-related variables in a number of places already like Code, Menu, and Checkbox.

DesignerDave (approved):

PR #1011, opened 2021-08-05
PD-1518 Enables any nested row height

TheSonOfThomp:
Transition styles are never properly added

## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1518](https://jira.mongodb.org/browse/PD-1518)

## 🛠 Types of changes

## ✅ Checklist


packages/table/src/Row.tsx
```diff
@@ -72,42 +72,14 @@ const modeStyles = {
...
-    transition: all 150ms ease-in-out;
+    max-height: unset;
   }
 `;
 
 const hideRow = css`
   opacity: 0;
 `;
 
-const transitionStyles = (mode: Mode) => {
```
TheSonOfThomp, line 83: @DesignerDave None of these styles are being applied

DesignerDave (commented):
I could've sworn that the rows transitioned between the expanded and collapsed states, but it doesn't look like it actually transitions on the deployed storybook. Any chance we could make that happen here?

packages/table/src/Row.tsx
```diff
@@ -70,42 +79,46 @@ const modeStyles = {
...
-      border-top-color: ${mode === Mode.Light
-        ? uiColors.gray.light2
-        : uiColors.gray.dark1};
+const nestedRowInitialStyle = css`
+  transition: all ${transitionTime}ms ease-in-out;
+  transition-property: border-color, transform, opacity;
+  transform-origin: 50% 0%;
+  border-color: var(--lg-table-row-border-color);
+  opacity: 0;
+  transform: translateY(-10%);
```
DesignerDave, line 102: `translate`, `translateX`, and `translateY` don't utilize the GPU like `translate3d` does. We should most always opt for `translate3d` when performing a transition using a translate transform, even though the declaration is a bit more clunky, since the transition will usually perform at a higher FPS.

packages/table/src/Row.tsx
```diff
@@ -70,42 +79,46 @@ const modeStyles = {
 };
 
 const rowStyle = css`
+  border-color: var(--lg-table-row-border-color);
```
DesignerDave, line 82: This seems like a great place to use the `border` property since we're defining each value that the property would handle.

DesignerDave (changes requested):

DesignerDave (changes requested):
Hey! This is almost there. I'm noticing two issues here:

- Borders aren't hidden when sections are collapsed. This is causing table rows with nested rows to have a thicker border bottom than ones that don't have any nested rows.
- The animation looks a lot better! It does look like the animation is cutting to it's final state halfway through the transition though, causing the transition to start smoothly, but end abruptly.

TheSonOfThomp:
@DesignerDave Fixed the border bug, and the animation is a little bit more smooth, but getting it perfect might be a bit out of scope

PR #1012, opened 2021-08-06
Have Portal render synchronously while still not breaking SSR

jetpacmonkey:
## ✍️ Proposed changes

In Portal v2.2.0, there was a change made to support SSR by not rendering until the second render (using `useEffect`). Unfortunately, this had the effect of breaking a bunch of tests in mms when I tried to update the library, because it made multiple components downstream async that had previously been synchronous. This PR fixes the issue by using `useLayoutEffect`, which still allows rendering it in SSR to be a no-op, but triggers the second render immediately while still in the same thread, so it acts as a synchronous render in tests.

🎟 _Jira ticket:_ _none_ (should I file one?)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/.changeset/beige-trains-relax.md
+++ b/.changeset/beige-trains-relax.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/portal': patch
+---
+
+Use useLayoutEffect hook internally to fix tests broken by Portal showing async behavior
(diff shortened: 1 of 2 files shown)


DesignerDave (approved):
LGTM – update makes sense, and everything seems to be running well locally!

PR #1014, opened 2021-08-11
PD-1500 segmented control

TheSonOfThomp:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1500](https://jira.mongodb.org/browse/PD-1500)

## 🛠 Types of changes

## ✅ Checklist


packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,238 @@
...
+      <input
+        type="radio"
+        name={name}
+        value={value}
+        id={id}
+        disabled={disabled}
+        onChange={onChange}
+        checked={checked}
+        className={cx(radioInputStyle)}
+      ></input>
```
DesignerDave, line 232: Input tags are self-terminating.

packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,314 @@
...
+      _onChange: handleChange,
+      ref: setRef(`${name}-${index}`),
+    });
+  });
+
+  const selectedIndex = React.Children.toArray(
+    renderedChildren,
+  ).findIndex(child =>
+    isControlled
+      ? (child as React.ReactElement).props.value === controlledValue
```
DesignerDave, line 250: I _think_ we can type child without typecasting here by typing the output array from `React.Children.toArray`.

packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,314 @@
...
+  );
+
+  /**
+   * Dynamically set the size & position of the selection indicator
+   */
+  const [selectionStyleDynamic, setSelectionStyleDynamic] = useState<string>();
+
+  // Update dynamic styles of the selection indicator
+  useEffect(() => {
+    const selectedElement = (getRef(
```
DesignerDave, line 261: We should improve the typing of the `getRef` function to avoid needing to type the ref on line 263.

packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,314 @@
...
+  }, [selectedIndex, getRef, name, renderedChildren]);
+
+  /**
+   * TODO
+   * - remove hover focus ring
+   */
+
+  return (
+    <div className={cx(wrapperStyle, className)} {...rest}>
+      <Overline className={cx(labelStyle[mode])}>{label}</Overline>
```
DesignerDave, line 281: We shouldn't use `cx()` unless we're composing more than one className together.

DesignerDave: Also, `label` is optional, so we should conditionally render this component.

packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,238 @@
...
+const uncheckedSelector = '[data-checked="false"]';
+
+const optionStyle = ({
+  mode = 'light',
+  size = 'default',
+}: {
+  mode: Mode;
+  size: Size;
+}) => css`
+  ${size === 'small' &&
```
DesignerDave, line 21: Since this is already a function, we don't need to be interpolating this conditional within the return value. You might consider writing the function like this:

```js
const optionStyle = () => {
  if (size === 'small') {
    return css``;
  }
  
  // Etc...
}
```

packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,238 @@
...
+  }: SegmentedControlOptionProps,
+  forwardedRef,
+) {
+  const mode = darkMode ? 'dark' : 'light';
+
+  return (
+    <label
+      htmlFor={id}
+      className={cx(optionStyle({ mode, size }), className)}
+      data-disabled={`${disabled}`}
```
DesignerDave, line 219: I saw this in a selector above as well. We probably shouldn't be using a data prop for this. Let me know if you'd like help brainstorming alternatives.

packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,238 @@
...
+    >
+      <input
+        type="radio"
+        name={name}
+        value={value}
+        id={id}
+        disabled={disabled}
+        onChange={onChange}
+        checked={checked}
+        className={cx(radioInputStyle)}
```
DesignerDave, line 231: We can remove `cx` here.

packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,238 @@
...
+  margin: 0;
+  opacity: 0;
+  pointer-events: none;
+`;
+
+const labelStyle = css`
+  display: inline-flex;
+  align-items: center;
+  gap: calc(var(--font-size) / 2);
+  cursor: inherit;
```
DesignerDave, line 165: Cursor styles should inherit from the parent by default.

packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,238 @@
...
+    content: '';
+    position: absolute;
+    height: var(--divider-height);
+    width: 1px;
+    left: calc(0px - (var(--segment-gap, 1px) + 1px) / 2);
+    background-color: transparent;
+    transition: all 100ms ease-in-out;
+  }
+
+  &${uncheckedSelector} {
```
DesignerDave, line 141: I think this selector needs to be rewritten. Right now, the output selector would be:

```css
.classname[data-checked="false"]:not(:first-child):not([data-checked="true"] + [data-checked="false"]):before
```

Assuming that the `:not([data-checked="true"] + [data-checked="false"])` portion also will have the same class name, you could write something like this instead:

```css
.classname ~ .classname[data-checked="false"]:before {
  background-color: ${uiColors.gray.light1};
}

.classname[data-checked="true"] + [data-checked="false"]:before {
  background-color: unset;
}

```


We can potentially simplify even further if you have the border added to the right of the element, rather than to the left.

```css
.classname[data-checked="false"]:after {
	background-color: ${uiColors.gray.light1};
}

.classname[data-checked="true"],
.classname:last-of-type {
	&:after {
    background-color: unset;
	}
}
```

Disclaimer: This is going from what I'm reading on this first pass – I haven't tested this locally yet.

DesignerDave (changes requested):
Took a first pass! Planning on pulling it down again tomorrow to test some more.

packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,314 @@
...
+  z-index: 0;
+  border-radius: 4px;
+  border-width: 1px;
+  border-style: solid;
+  background-color: var(--indicator-background-color);
+  border-color: var(--indicator-border-color);
+`;
+
+const interactionRingStyle = css`
+  z-index: 3;
```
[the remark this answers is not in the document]
TheSonOfThomp: This ensures the interactionRing shadow appears above the frame and selection indicators

packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,314 @@
...
+  label?: string;
+}
+
+/**
+ * Component
+ */
+const SegmentedControl = React.forwardRef<
+  HTMLDivElement,
+  SegmentedControlProps
+>(function SegmentedControl(
```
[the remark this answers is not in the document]
TheSonOfThomp: I've seen both patterns in the codebase (named functions are used in `Button` and `MenuButton`). The benefit of a named function is that any  runtime errors in the component will have the function name in the stack trace as opposed to 'anonymous'

packages/segmented-control/src/SegmentedControlOption.tsx
TheSonOfThomp: I'll have to think on this. Since the function takes in both `size` and `mode`, the returned css depends on both those parameters. We can't elegantly concatenate the results of two `css\`\`` calls, so I'll have to consider refactoring this a bit.

packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+// Internal “Private” Props
+// _name - the name of the segmented control. Used to group the HTML radio inputs
+// _checked - whether the option is checked. Defined by the parent
+// _onChange - the onChange handler passed in from the SegmentedControl
+export interface SegmentedControlOptionProps {
+  value: string;
+  children: React.ReactNode;
+  disabled?: boolean;
+  as?: any;
```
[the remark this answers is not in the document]
TheSonOfThomp: Thinking more on this, I'm not sure I see a whole lot of value compared to the complexity of supporting an `as` prop on the option.

The main use case I see for the `as` prop is to support routing components like (Next.js `Link`), but since (a) this can be done programmatically via `router.push`, and (b) since we don't want to encourage using the segmented control for navigation, I don't think we should support this functionality

packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -0,0 +1,346 @@
...
+  const vars: {
+    [key: string]: string;
+  } = {};
+
+  if (size == 'small') {
+    vars.height = '100%';
+  } else if (size == 'large') {
+    vars.height = 'calc(100% - 2 * var(--padding))';
+  } else {
+    vars.height = 'calc(100% - 2 * var(--padding))';
```
TheSonOfThomp, line 130: 🥴

packages/segmented-control/src/SegmentedControlOption.tsx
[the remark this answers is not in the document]
DesignerDave: The value is primarily in accessibility and supporting edge cases we didn't anticipate. `router.push` requires a ton of manual work to make it translate well to screen readers, support keybindings like command+click to open in a new tab, etc. We have an entire component built to make the typing and implementation of `as` easier, and I feel strongly we should support this functionality.

DesignerDave:
Testing the component locally, I'm seeing some issues:
- In the "Default" story, one option is disabled, but hovering over the entire control shows the not-allowed cursor. Additionally, clicking on any option doesn't switch what's selected.
- In the "Uncontrolled" story, before I click the component, hovering on the component turns it white. The first time I click to try to switch the selected option, selects the last option (regardless of which I clicked). Clicking any option afterwards does not switch the currently selected option.
- I'm currently seeing a border between the label and the control, but I'm not seeing the borders between options.

packages/segmented-control/src/SegmentedControl.tsx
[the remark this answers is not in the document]
DesignerDave: I see the benefit, but I'd like to see us explore if there's a better if we intend to do this going forward. Is there a better pattern we could use here to avoid the awkwardness of this?

```js
const SegmentedControl = React.forwardRef(function SegmentedControl() {})
```

Maybe we could do something along these lines (with probably a bit of typescript work as well):

```js
function SegmentedControl() {}

SegmentedControl.displayName = "SegmentedControl";

export default React.forwardRef(SegmentedControl);
```

DesignerDave (approved):
Only thing I noticed was that the border to the right of the first option appears when it shouldn't on the first render. When switching between selections, it behaves appropriately.

Screen Shot 2021-09-24 at 1 23 46 PM

Other than that, I think we're good to get this in. Once that small issue is resolved, feel free to merge!

PR #1017, opened 2021-08-19
PD-1540: Removes problematic styles from InteractionRing component

DesignerDave:
## ✍️ Proposed changes

The InteractionRing component was applying some default styles to the children passed to it that were largely unused, and causing issues when TextArea and TextInput were used together.

🎟 _Jira ticket:_ [PD-1540](https://jira.mongodb.org/browse/PD-1540)

## 🛠 Types of changes

- [] Bugfix (non-breaking change which fixes an issue)

## ✅ Checklist

### DIFF
--- a/packages/interaction-ring/src/InteractionRing.tsx
+++ b/packages/interaction-ring/src/InteractionRing.tsx
@@ -96,16 +96,6 @@ function computeStyles({
   };
 }
 
-const baseContentStyle = css`
-  width: 100%;
-  height: 100%;
-
-  // Form elements don't always inherit these by default
-  // https://developer.mozilla.org/en-US/docs/Learn/Forms/Styling_web_forms#Fonts_and_text
-  font-family: inherit;
-  font-size: 100%;
-`;
-
 interface State {
   hovered?: boolean;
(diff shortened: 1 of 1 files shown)


packages/interaction-ring/src/InteractionRing.tsx
```diff
@@ -96,16 +96,6 @@ function computeStyles({
   };
 }
 
-const baseContentStyle = css`
-  width: 100%;
```
zhaodext, line 100: The `TextArea` has CSS `width` and `min-height` defined there, so it should be fine with this change. Does this affect other components? (I don't see this applied to the `leafygreen-card`)

DesignerDave: It's used in a number of other components (not card, however). I'm gonna get a second set of eyes on this to verify this week.

TheSonOfThomp: Doesn't look like this should break anything, and should be a simple fix in case it does

TheSonOfThomp (approved):

PR #1034, opened 2021-10-04
PD-1596 Form footer

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


packages/form-footer/src/FormFooter.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+const flexSpan = css`
+  display: inline-flex;
+  gap: inherit;
+`;
+
+const bannerStyle = css`
+  flex-grow: 0;
+  min-height: unset;
+  padding-block: 7px;
```
DesignerDave, line 42: Why are we changing thee left/right padding of the banner?

packages/form-footer/src/FormFooter.tsx
```diff
@@ -0,0 +1,184 @@
...
+const footerStyle = (sticky: boolean) => {
+  return css`
+    position: ${sticky ? 'fixed' : 'static'};
+    bottom: ${sticky ? '0px' : 'unset'};
+    left: ${sticky ? '0px' : 'unset'};
+    min-height: 92px;
+    width: 100%;
+    padding: 24px;
+    border: 1px solid ${uiColors.gray.light2};
+    box-shadow: 0px -4px 4px 0px rgba(6, 22, 33, 0.1);
```
DesignerDave, line 18: Is this RGBA value one of our palette colors? Can we use polished to programmatically produce this value instead?

packages/form-footer/src/FormFooter.tsx
```diff
@@ -0,0 +1,184 @@
...
+const isPrimaryButtonProps = (testObj: any): testObj is PrimaryButtonProps => {
+  return testObj && testObj.text != null;
+};
+
+export interface FormFooterProps {
+  /**
+   * Defines whether the footer should be "stuck" to the bottom of the frame.
+   * Can also be customized using `className`
+   */
+  sticky?: boolean;
```
DesignerDave, line 65: I worry this will be confused with `position: sticky`. Do we need to handle this ourselves in the first place? It seems simple enough for folks to apply the positioning they need for their own case.

website/pages/component/form-footer/example.tsx
```diff
@@ -0,0 +1,66 @@
...
+  errorMessage: {
+    type: 'text',
+    default: "There's an error on this page.",
+    label: 'Error message',
+  },
+  // darkMode: {
+  //   type: 'boolean',
+  //   default: false,
+  //   label: 'Dark Mode',
+  // },
```
DesignerDave, line 41: Can we remove this please?

packages/form-footer/src/FormFooter.tsx
```diff
@@ -0,0 +1,184 @@
...
+};
+
+const contentStyle = css`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin: 0 auto;
+  gap: 8px;
+
+  > button {
```
DesignerDave, line 29: Can we use the Button's className prop instead of a child selector here?

packages/form-footer/src/FormFooter.tsx
```diff
@@ -0,0 +1,184 @@
...
+    border: 1px solid ${uiColors.gray.light2};
+    box-shadow: 0px -4px 4px 0px rgba(6, 22, 33, 0.1);
+  `;
+};
+
+const contentStyle = css`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin: 0 auto;
```
DesignerDave, line 26: Do we need to be setting the top/bottom margins to 0 here?

DesignerDave (changes requested):
Left a few comments. Also, it looks like we may be missing a changeset?

packages/form-footer/src/FormFooter.tsx
TheSonOfThomp: `padding-block` refers to top/bottom. We're changing it since the banner and button are different heights, re [this thread](https://mongodb.slack.com/archives/G01500NFVPS/p1635181579001300)

packages/form-footer/src/FormFooter.tsx
TheSonOfThomp: That's fair—I was having similar thoughts. I'll remove this prop

packages/form-footer/src/FormFooter.tsx
TheSonOfThomp: I changed the markup so this wasn't actually doing anything. Resolved

packages/form-footer/src/FormFooter.tsx
TheSonOfThomp: Nope. Modified to only target left/right

TheSonOfThomp:
> ... it looks like we may be missing a changeset?

Do new components need a changeset?

packages/form-footer/src/FormFooter.tsx
DesignerDave: Whoops, you're right. Sometimes it's applied on the left and right based on other CSS properties, but not in this case. Sounds good 👍

packages/form-footer/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+```shell
+npm install @leafygreen-ui/form-footer
+```
+
+## Example
+
+```js
+<FormFooter
+  sticky={true}
```
DesignerDave, line 25: We don't need this in the example anymore, right?

website/pages/component/form-footer/example.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import FormFooter from '@leafygreen-ui/form-footer';
+import LiveExample, { KnobsConfigInterface } from 'components/live-example';
+
+const knobsConfig: KnobsConfigInterface<{
+  sticky?: boolean;
```
DesignerDave, line 6: I think we can remove this as well.

website/pages/component/form-footer/example.tsx
```diff
@@ -0,0 +1,50 @@
...
+    label: 'Error message',
+  },
+};
+
+export default function CardLiveExample() {
+  return (
+    <LiveExample knobsConfig={knobsConfig}>
+      {({ primaryButtonText, cancelText, backButtonText, errorMessage }) => (
+        <FormFooter
+          sticky={sticky}
```
DesignerDave, line 39: Ditto here

DesignerDave (changes requested):
Few remaining references to `sticky` I think probably should removed. Otherwise, it's looking good!

DesignerDave (approved):

PR #1059, opened 2021-10-29
[PD-1629 PD-1699] Combobox

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


packages/combobox/src/ComboboxOption.tsx
```diff
@@ -0,0 +1,198 @@
...
+    const renderedIcon = useMemo(() => {
+      if (glyph) {
+        if (isComponentGlyph(glyph) || isComponentType(glyph, 'Icon')) {
+          return glyph;
+        }
+        console.error(
+          '`ComboboxOption` instance did not render icon because it is not a known glyph element.',
+          glyph,
+        );
+      }
```
DesignerDave, line 120: Could we do an explicit default return here?

packages/combobox/src/ComboboxOption.tsx
```diff
@@ -0,0 +1,198 @@
...
+    }, [glyph]);
+
+    const renderedChildren = useMemo(() => {
+      // Multiselect
+      if (multiselect) {
+        const checkbox = (
+          <Checkbox
+            label=""
```
DesignerDave, line 128: I believe it's better to not have the label attribute here than to have it empty. Alternatively, you could use the displayName here as well.

packages/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,923 @@
...
+        </InteractionRing>
+
+        {state === 'error' && errorMessage && (
+          <div className={errorMessageStyle}>{errorMessage}</div>
+        )}
+
+        {/******* /
+          *  Menu  *
+          / *******/}
+        <Popover
```
DesignerDave, line 900: Currently, clicking on an option in the combobox is causing the `:active` pseudo-class of the button to appear.

DesignerDave: Also, in single select mode, clicking on the currently selected option should probably close the dropdown. (Sorry, not sure exactly what lines this comment would be relevant on).

packages/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,923 @@
...
+    updateSelection,
+  ]);
+
+  const renderedInputIcons = useMemo(() => {
+    return (
+      <>
+        {clearable && doesSelectionExist && (
+          <IconButton
+            ref={clearButtonRef}
+            onClick={() => updateSelection(null)}
```
DesignerDave, line 346: This does not appear to cause the example to re-validate your selection.

Repro: 
- Select "Dragonfruit", causing an error to appear
- Clear your current selections
- The error remains until you make a new selection

packages/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,923 @@
...
+
+    const isFocusInComponent = isFocusOnCombobox || isFocusInMenu;
+
+    if (isFocusInComponent) {
+      // No support for modifiers yet
+      if (event.ctrlKey || event.shiftKey || event.altKey) {
+        return;
+      }
+
+      switch (event.keyCode) {
```
DesignerDave, line 740: We could also handle the delete button here. It would feel awesome if you could hit backspace when at the beginning of the input, and it deleted a selection.

packages/combobox/src/Chip.tsx
```diff
@@ -0,0 +1,198 @@
...
+    position: absolute;
+    top: 0;
+    left: 0;
+    height: 100%;
+    width: 1px;
+    background-color: var(--lg-combobox-chip-hover-color);
+  }
+
+  &:hover {
+    background-color: var(--lg-combobox-chip-hover-color);
```
DesignerDave, line 95: Looks like this is missing a transition.

packages/combobox/src/Chip.tsx
```diff
@@ -0,0 +1,198 @@
...
+    };
+
+    const dataProp = createDataProp('combobox-chip');
+
+    return (
+      <span
+        data-test-id="lg-combobox-chip"
+        {...dataProp.prop}
+        ref={forwardedRef}
+        className={chipWrapperStyle({ darkMode, size })}
```
DesignerDave, line 174: It would be rad if you could click on a chip (outside of the "x"), and it would select the chip like you can with the arrow keys.

packages/combobox/src/Combobox.tsx
DesignerDave: Or maybe it selects it instead like the left arrow key does 🤔 That might lead to less accidental deletions.

packages/combobox/src/Chip.tsx
```diff
@@ -0,0 +1,198 @@
...
+      }
+    }, [forwardedRef, isFocused]);
+
+    const handleKeyDown = (e: React.KeyboardEvent) => {
+      if (
+        e.keyCode === 8 || // Delete
+        e.keyCode === keyMap.Enter ||
+        e.keyCode === keyMap.Space
+      ) {
+        onRemove();
```
DesignerDave, line 163: If you remove a selection from a list of multiple chips, it should probably move your focus to the next chip in the list, rather than bringing the input back into focus. Right now, it feels like you lose context when you remove an item.

packages/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,923 @@
...
+          break;
+        }
+
+        case keyMap.ArrowRight: {
+          updateFocusedChip('next');
+          break;
+        }
+
+        case keyMap.ArrowLeft: {
+          updateFocusedChip('prev');
```
DesignerDave, line 797: Currently, this handling seems to break the ability to use arrow keys to move through a string that you've typed. It'll be a challenge, but we need to only move focus to a chip from the input when your cursor is all the way to the left.

I think there's a DOM API like `selectionStart` you might be able to use for this.

DesignerDave (changes requested):
Here's a first pass! It's looking and feeling really great 😁

packages/combobox/src/ComboboxOption.tsx
```diff
@@ -0,0 +1,198 @@
...
+  justify-content: space-between;
+  list-style: none;
+  color: inherit;
+  cursor: pointer;
+  overflow: hidden;
+  font-size: var(--lg-combobox-item-font-size);
+  line-height: var(--lg-combobox-item-line-height);
+  padding: var(--lg-combobox-item-padding-y) var(--lg-combobox-item-padding-x);
+
+  &:before {
```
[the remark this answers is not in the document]
TheSonOfThomp: I had a left-right transition initially & didn't like it. I'll look into the `side-nav` one

packages/combobox/src/ComboboxOption.tsx
```diff
@@ -0,0 +1,198 @@
...
+  display: inline-flex;
+  gap: 8px;
+  justify-content: start;
+  align-items: inherit;
+`;
+
+const displayNameStyle = (isSelected: boolean) => css`
+  font-weight: ${isSelected ? 'bold' : 'normal'};
+
+  > em {
```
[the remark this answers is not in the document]
TheSonOfThomp: Good callout on `em` vs `strong`. Will update

packages/combobox/src/ComboboxOption.tsx
TheSonOfThomp: Curious what that adds?

packages/combobox/src/ComboboxOption.tsx
TheSonOfThomp: `label` is converted into actual text in the DOM. There's a bug in the Checkbox where `label` is still required even if `aria-label` is provided

packages/combobox/src/Combobox.tsx
TheSonOfThomp: At least three Options:
1) `Delete` key removes a chip (gmail)
2) `Delete` key selects the prev chip. Pressing again deletes it
3) `Delete` key removes a chip. and sets the input value to the chip's content (stackoverflow)

packages/combobox/src/Chip.tsx
TheSonOfThomp: That's a good point. Using mouse it made sense at is, but with keyboard it's disorienting

packages/combobox/src/ComboboxOption.tsx
DesignerDave: Do we have a ticket for this? If so, we should leave a comment with a link to the ticket here.

TheSonOfThomp: Created one https://jira.mongodb.org/browse/PD-1681

packages/combobox/src/Combobox.tsx
TheSonOfThomp: >  clicking on an option in the combobox is causing the :active pseudo-class of the button to appear.

I don't think it's the `:active` pseudo-class—I think it's removing the `:focus` pseudo-class

packages/combobox/src/Combobox.tsx
DesignerDave: I'd chat with Alven too about this, but I lean towards #2 to avoid accidental deletions while also supporting the backspace behavior.

packages/combobox/src/Combobox.tsx
DesignerDave: Might be wrong on the issue here 😄 regardless, the symptom is there.

packages/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,953 @@
...
+          (onChange as onChangeType<false>)?.(
+            selection as SelectValueType<false>,
+          );
+          closeMenu();
+        }
+      } else {
+        setInputValue('');
+      }
+    }
+  }, [
```
TheSonOfThomp, line 614: @DesignerDave Should this be handled in `updateSelection`?

DesignerDave (changes requested):
Hey Adam! Spent a bit of time playing with the Combobox component! I found a few issues that are making the component less natural to use right now:
- (Testing single select) Moving focus between the input and the clear (x) button causes the dropdown to close, then re-open.
- (Testing single select) Focusing on the clear (x) button causes the input to also display as focused, which I'm not sure is desirable.
- (Testing multi select) Focusing on one of the chips, then using the right arrow key to move the cursor to the text places the cursor after the first character, not before it as you would expect.
- (Testing multi select) Typing in the input, then clicking outside to make the input lose focus hides the text you typed until you focus on the input again.
- (Testing multi select) You currently can't click to place the cursor anywhere in the string you're filtering with in the input. I'd expect to be able to type "apple", then click between the "p" and "l" to place the cursor there, for example.
- (Testing multi select) The checkbox animation doesn't fire when making a selection in the dropdown.
- (Testing multi select) When a chip is focused, pressing tab skips all other chips and moves directly to the clear (x) button. Pressing shift+tab however will focus on each chip in order rather than skipping them.
- The focus state bar (on the left of a focused option) in the Combobox component is a different color, and has a different border radius than the side nav component's focus states.
- Do you think it would be appropriate to use `type="search"` for the actual text input here? There are pros and cons I think of this approach – namely that some browsers add their own clear buttons to search inputs.

TheSonOfThomp:
> * (Testing multi select) The checkbox animation doesn't fire when making a selection in the dropdown.

Do we want this to fire? I intentionally disabled it since I thought it was to much

TheSonOfThomp:
> * The focus state bar (on the left of a focused option) in the Combobox component is a different color, and has a different border radius than the side nav component's focus states.

I'm using the color/radius as Alven defined in his Figma file. Should we consider changing it?

TheSonOfThomp:
> * Do you think it would be appropriate to use `type="search"` for the actual text input here? There are pros and cons I think of this approach – namely that some browsers add their own clear buttons to search inputs.

AFAIK `type="search"` doesn't add anything semantic over what we're already doing (i.e. `role="combobox"`)

TheSonOfThomp:
> * (Testing single select) Focusing on the clear (x) button causes the input to also display as focused, which I'm not sure is desirable.
> * (Testing multi select) Focusing on one of the chips, then using the right arrow key to move the cursor to the text places the cursor after the first character, not before it as you would expect.
> * (Testing multi select) Typing in the input, then clicking outside to make the input lose focus hides the text you typed until you focus on the input again.
> * (Testing multi select) You currently can't click to place the cursor anywhere in the string you're filtering with in the input. I'd expect to be able to type "apple", then click between the "p" and "l" to place the cursor there, for example.
> * (Testing multi select) The checkbox animation doesn't fire when making a selection in the dropdown.
> * (Testing multi select) When a chip is focused, pressing tab skips all other chips and moves directly to the clear (x) button. Pressing shift+tab however will focus on each chip in order rather than skipping them.
> * The focus state bar (on the left of a focused option) in the Combobox component is a different color, and has a different border radius than the side nav component's focus states.
> * Do you think it would be appropriate to use `type="search"` for the actual text input here? There are pros and cons I think of this approach – namely that some browsers add their own clear buttons to search inputs.

TheSonOfThomp:
> * (Testing single select) Moving focus between the input and the clear (x) button causes the dropdown to close, then re-open.
> * (Testing single select) Focusing on the clear (x) button causes the input to also display as focused, which I'm not sure is desirable.

Should focusing the clear button keep the menu open, or should it be closed?

Also, should the overall focus state still be shown when a Chip is focused? Right now I'm using `:focus-within` to toggle the outline

DesignerDave:
> Should focusing the clear button keep the menu open, or should it be closed?

My main issue here is with it sort of "flickering" closed then open again when that focus changes. I personally feel like the behavior aside from that flickering is probably fine, but it's a good question to run by Alven 😄

> Also, should the overall focus state still be shown when a Chip is focused? Right now I'm using `:focus-within` to toggle the outline

This was the piece of feedback I'm most unsure about, but it didn't feel natural to show both focus states at once IMO

DesignerDave:
> > * (Testing multi select) The checkbox animation doesn't fire when making a selection in the dropdown.
> 
> Do we want this to fire? I intentionally disabled it since it feels delayed, and also when using space/enter to select it's very easy to run into the sprite sheet glitch (where the sprite sheet gets stuck in the middle)

Maybe ask Alven in this case 😄 sounds like a design decision

DesignerDave:
> > * The focus state bar (on the left of a focused option) in the Combobox component is a different color, and has a different border radius than the side nav component's focus states.
> 
> I'm using the color/radius as Alven defined in his Figma file. Should we consider changing it?

Sounds like an inconsistency that should probably be addressed in Figma in this case.

DesignerDave:
> > * Do you think it would be appropriate to use `type="search"` for the actual text input here? There are pros and cons I think of this approach – namely that some browsers add their own clear buttons to search inputs.
> 
> AFAIK `type="search"` doesn't add anything semantic over what we're already doing (i.e. `role="combobox"`)

The largest benefit to the type is some benefits on mobile, and I'd also guess it has an impact on screen reader parsing (though I'm not sure about that). Wanted to make sure we consider it, though I don't feel strongly.

DesignerDave (changes requested):
Looking really good! Noticed two more bugs this time around:
- Single Select: If you make a selection, then backspace to delete characters until you see other options, then select one using the arrow keys and enter, you can no longer type in the input until you leave the input, then re-select it.
- I'm unable to make mouse selections of text within the input in both single and multi select.

Also, we should update the justification of the popover component so that the dropdown animates from the middle.

DesignerDave (changes requested):
Hey! The two issues I noted seem to be working now. I found one other bug:

In the single select combobox, make a selection, then press backspace to remove that selection from the input. From here, even if you type different characters, the dropdown never opens to reveal options.

DesignerDave (approved):
I think it's time we get this into the wild 😄

PR #1063, opened 2021-11-04
[PD-1638, PD-1362] Fixes Select keyboard navigation

TheSonOfThomp:
## ✍️ Proposed changes

🎟 _Jira ticket:_ 
 - [PD-1638](https://jira.mongodb.org/browse/PD-1638)
 - [PD-1362](https://jira.mongodb.org/browse/PD-1362)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

Test with the following story: 
```js
  .add('Tab test', () => {
    return (
      <LeafygreenProvider>
        <div
          className={css`
            display: block;
          `}
        >
          <Select
            label="Label"
            description="Description"
            placeholder="Placeholder"
            name="Name"
            size={Size.Default}
            defaultValue="cat"
          >
            <Option value="dog">Dog</Option>
            <Option value="cat">Cat</Option>
            <OptionGroup label="Less common">
              <Option value="hamster">Hamster</Option>
              <Option value="parrot">Parrot</Option>
            </OptionGroup>
          </Select>
          <TextInput
            label="Email"
            description="Enter your email below"
            placeholder="your.email@example.com"
          />
          <TextInput
            label="Password"
            description="Enter your email below"
            placeholder="your.email@example.com"
          />
          <TextInput
            label="Description"
            description="Enter your email below"
            placeholder="your.email@example.com"
          />
        </div>
      </LeafygreenProvider>
    );
  });
```

### DIFF
--- a/packages/select/src/Select.tsx
+++ b/packages/select/src/Select.tsx
@@ -6,7 +6,7 @@ import {
   useIdAllocator,
   useEventListener,
 } from '@leafygreen-ui/hooks';
-import { OneOf, keyMap } from '@leafygreen-ui/lib';
+import { OneOf } from '@leafygreen-ui/lib';
 import { PopoverProps } from '@leafygreen-ui/popover';
 import { fontFamilies, breakpoints } from '@leafygreen-ui/tokens';
 import { colorSets, mobileSizeSet, Mode, Size, sizeSets } from './styleSets';
--- a/packages/select/src/Select.tsx
+++ b/packages/select/src/Select.tsx
@@ -334,45 +334,57 @@ export default function Select({
         return;
       }
 
-      switch (event.keyCode) {
-        case keyMap.Tab:
-        case keyMap.Escape:
-          onClose();
-          setFocusedOption(undefined);
-          break;
-        case keyMap.Enter:
-        case keyMap.Space:
-          if (open && document.activeElement !== menuButtonRef.current) {
-            // Default behavior is to use these keys to open the dropdown but we handle that manually
-            event.preventDefault();
-          }
+      const isFocusInMenu = listMenuRef.current?.contains(
+        document.activeElement,
+      );
+      const isFocusOnButton = menuButtonRef.current?.contains(
+        document.activeElement,
+      );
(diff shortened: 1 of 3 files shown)


packages/select/src/Select.tsx
```diff
@@ -334,45 +334,57 @@ export default function Select({
...
+        document.activeElement,
+      );
+      const isFocusOnButton = menuButtonRef.current?.contains(
+        document.activeElement,
+      );
+      const isFocusInComponent = isFocusOnButton || isFocusInMenu;
+
+      // We only respond to keypresses if the focus is in the component
+      if (isFocusInComponent) {
+        switch (event.key) {
```
TheSonOfThomp, line 347: Changing the switch parameter since `event.keyCode` [is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)

DesignerDave: Is this what caused the issue? We've kept using keyCode because `KeyboardEvent.key` is only partially supported in IE 11, and keyCode has consistent browser support currently. I don't think _most_ teams support IE 11, but it seems like something we should confirm before making this change.

TheSonOfThomp: It's not the source of the issue, no, but I was getting IDE deprecation warnings. Didn't realize IE had only parital support. Looking into it, it seems IE only has issues regarding the `AltGraph` key, but I may be misreading https://caniuse.com/keyboardevent-key

TheSonOfThomp: I'll revert if that's something we want to stick with

DesignerDave: If it is in fact only AltGraph, then it could be fine though I tend to avoid APIs with inconsistent behavior even if we don't _think_ we use it or will use it. We should do some research to understand the current browser support we offer in all of our properties before moving to `KeyboardEvent.key`, since the `keyCode` property is still widely supported and is consistent.

Let's keep this PR simple so we can get the fix in, and evaluate this separately. If we do make this update, we should make it globally.

DesignerDave (approved):

PR #1069, opened 2021-11-18
PD-1675 Do size mapping before passing into component

khelif96:
## ✍️ Proposed changes

This is similar to https://github.com/mongodb/leafygreen-ui/pull/1068 but handles casing on the type of size internally instead of forcing the implementer to handle it. 

## 🛠 Types of changes

## ✅ Checklist


TheSonOfThomp:
Hey @khelif96, looks good. Could you run `yarn changeset` and make sure any other relevant checklist items are done?

packages/icon/src/createIconComponent.tsx
```diff
@@ -14,10 +14,15 @@ type GlyphObject = Record<string, LGGlyph.Component>;
...
+  const Icon = ({ glyph, size, ...rest }: IconProps) => {
     const SVGComponent = glyphs[glyph];
+    let iconSize = sizeMap[Size.Default];
 
-    return <SVGComponent {...rest} />;
+    if (size !== undefined) {
+      iconSize = typeof size === 'number' ? size : sizeMap[size];
+    }
+
+    return <SVGComponent size={iconSize} {...rest} />;
```
TheSonOfThomp, line 25: It doesn't seem like this works unless the consuming developer handles the own `size` prop in the custom svg themselves
(i.e. 
```
const myGlyphs = {
  MyCustomGlyph: ({size}) => <svg height={size} width={size} />,
};
```
)

I'm not sure this adds much value over requiring the svg be properly sized in the first place. 

I think we should come up with a way to get the `size` prop to work without needing handling within the custom.

TheSonOfThomp (changes requested):

.changeset/pink-ligers-dance.md
```diff
@@ -0,0 +1,18 @@
...
+'@leafygreen-ui/icon': minor
+---
+
+Updates `createIconComponent` to handle both number and string values passed to the `size` prop on custom glyphs. 
+
+e.g.
+```jsx
+  import { createIconComponent, Size} from '@leafygreen-ui/icon'
+
+  const myIcon = ({size}) => <svg height={size} width={size}> ... </svg>
```
TheSonOfThomp, line 11: My initial concern is still valid, but this is still a useful feature to add! We should consider a more versatile solution in the future

TheSonOfThomp (approved):

TheSonOfThomp:
After discussing further with the team, I'm closing this in favor of #1172. Exporting, and encouraging use of, `createGlyphComponent` will ensure more consistency than duplicating that functionality in `createIconComponent`.

PR #1072, opened 2021-12-03
PD-1669 Add size variants to text input

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1669](https://jira.mongodb.org/browse/PD-1669)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/text-input/src/TextInput.spec.tsx
+++ b/packages/text-input/src/TextInput.spec.tsx
@@ -7,7 +7,7 @@ import {
 } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { axe } from 'jest-axe';
-import TextInput, { State } from './TextInput';
+import TextInput, { State, SizeVariant } from './TextInput';
 
 const error = 'This is the error message';
 const validEmail = 'test.email@mongodb.com';
@@ -219,6 +219,21 @@ describe('packages/text-input', () => {
       userEvent.tab(); // focus
       userEvent.type(inputElement, `test`);
       expect(handleValidation).toHaveBeenCalledTimes(5); // blur + keypress * 4
+    });
+  });
+
+  describe('when the "sizeVariant" is "large"', () => {
+    test('check if font-size is 18px', () => {
+      const { label } = renderTextInput({
+        value: validEmail,
+        sizeVariant: SizeVariant.Large,
+        optional: true,
+        ...defaultProps,
+      });
+
+      expect((label)).toHaveStyle({
+        fontSize: '18px',
+      });
     });
   });
 
--- a/packages/text-input/src/TextInput.spec.tsx
+++ b/packages/text-input/src/TextInput.spec.tsx
@@ -7,7 +7,7 @@ import {
 } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { axe } from 'jest-axe';
-import TextInput, { State, SizeVariantType } from './TextInput';
+import TextInput, { State, SizeVariant } from './TextInput';
 
 const error = 'This is the error message';
 const validEmail = 'test.email@mongodb.com';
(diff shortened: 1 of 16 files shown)


packages/text-input/src/TextInput.tsx
```diff
@@ -107,6 +117,8 @@ interface TextInputProps extends HTMLElementProps<'input', HTMLInputElement> {
   handleValidation?: (value: string) => void;
 
   ['aria-labelledby']?: string;
+
+  sizeVariant?: SizeVariantType;
```
TheSonOfThomp, line 121: I think we just call this prop `size` to keep in line with other components

packages/text-input/src/TextInput.tsx
```diff
@@ -39,6 +39,16 @@ const Mode = {
...
 
+export const SizeVariantType = {
+  XSmall: 'xsmall',
+  Small: 'small',
+  Default: 'default',
+  Medium: 'medium',
+  Large: 'large',
+} as const;
+
+export type SizeVariantType = typeof SizeVariantType[keyof typeof SizeVariantType];
```
TheSonOfThomp, line 50: I think we should just call this `Size` to keep in line with e.g. `State`, and other components that have a `State` variant

packages/text-input/src/TextInput.tsx
```diff
@@ -229,6 +241,46 @@ const interactionRingColor: Record<Mode, Record<'valid' | 'error', string>> = {
...
+    inputText: 12,
+    text: 14,
+    lineHeight: 20,
+  },
+  [SizeVariantType.Small]: {
+    inputHeight: 28,
+    inputText: 14,
+    text: 14,
+    lineHeight: 20,
+  },
```
TheSonOfThomp, line 263: Looks like XS and S sizes have a bit too much left/right padding compared to Figma

packages/text-input/src/TextInput.tsx
```diff
@@ -229,6 +241,46 @@ const interactionRingColor: Record<Mode, Record<'valid' | 'error', string>> = {
...
+  [SizeVariantType.Medium]: {
+    inputHeight: 36,
+    inputText: 16,
+    text: 16,
+    lineHeight: 20,
+  },
```
TheSonOfThomp, line 275: Is there a `Medium` variant in Figma?

packages/text-input/src/TextInput.tsx
```diff
@@ -229,6 +241,46 @@ const interactionRingColor: Record<Mode, Record<'valid' | 'error', string>> = {
...
+  [SizeVariantType.Large]: {
+    inputHeight: 48,
+    inputText: 18,
+    text: 18,
+    lineHeight: 22,
+  },
```
TheSonOfThomp, line 281: Looks like the `Large` variant needs more padding

TheSonOfThomp (changes requested):

packages/text-input/src/TextInput.tsx
shaneeza: I tried to use `size` but it looks like since we're using the `HTMLInputElement` interface that already has a property of `size` with a type of `number` so I was getting a TS error. To avoid that I changed it to `sizeVariant`.

packages/text-input/src/TextInput.spec.tsx
```diff
@@ -222,6 +222,21 @@ describe('packages/text-input', () => {
...
+        optional: true,
+        ...defaultProps,
+      });
+
+      expect(label).toHaveStyle({
+        fontSize: '18px',
+      });
+    });
+  });
+
```
TheSonOfThomp, line 239: This works for now, but this type of test is better suited to something like Chromatic or Percy

TheSonOfThomp (approved):

packages/text-input/src/TextInput.spec.tsx
TheSonOfThomp: We could even consider using this package https://www.npmjs.com/package/jest-image-snapshot

PR #1078, opened 2021-12-17
PD-1327 select error message

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [Add error message prop to Select: Dev](https://jira.mongodb.org/browse/PD-1327)

## 🛠 Types of changes

- [X] New feature (non-breaking change which adds functionality)

## ✅ Checklist

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)


packages/select/src/Select.spec.tsx
```diff
@@ -811,5 +813,32 @@ describe('packages/select', () => {
         expect(getByTextFor(combobox, 'selected')).toBeVisible();
       });
     });
+
+    describe('when the "state" is "error"', () => {
+      test('displays error message when input is invalid', () => {
```
TheSonOfThomp, line 818: This is trivial, but I think this test name should mention the `state` prop, and not the input (since it technically has nothing to do with the selected value)

packages/select/src/MenuButton.tsx
```diff
@@ -105,9 +114,44 @@ const MenuButton = React.forwardRef<HTMLElement, Props>(function MenuButton(
...
+            border-color: ${errorColor};
+            box-shadow: 0px 1px 2px rgba(87, 11, 8, 0.3);
+
+            &:hover,
+            &:active {
+              border-color: ${errorColor};
+              box-shadow: ${mode === Mode.Light
+                ? `0px 4px 4px rgba(87, 11, 8, 0.3),
+              0px 0px 0px 3px ${uiColors.red.light3}`
+                : `0px 4px 4px rgba(87, 11, 8, 0.3), 0px 0px 0px 3px ${uiColors.red.dark2}`};
```
TheSonOfThomp, line 128: We should use `polished.transparentize` (or `.opacify`) with `uiColors` (if it exists), instead of hard coding an `rgba` value

packages/select/src/MenuButton.tsx
```diff
@@ -64,6 +71,8 @@ const MenuButton = React.forwardRef<HTMLElement, Props>(function MenuButton(
   const colorSet = colorSets[mode];
   const sizeSet = sizeSets[size];
 
+  const errorColor = mode === Mode.Light ? uiColors.red.base : '#F97216';
```
TheSonOfThomp, line 74: Should the dark mode color be in the palette?

TheSonOfThomp (changes requested):

packages/select/src/MenuButton.tsx
shaneeza: As discussed this will be updated during the reskin

packages/select/src/MenuButton.tsx
shaneeza: As discussed this will be updated during the reskin

TheSonOfThomp (approved):

PR #1079, opened 2022-01-01
Fix license

Eddippla:
Basically just fixed the section in the apache license where the owner of the product should put in their copyright notice and remove the appendix section, look at commit to see changes


DesignerDave (approved):
Thanks!

PR #1094, opened 2022-01-14
PD-1709: Refresh Callout

shaneeza:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/callout/src/Callout.tsx
+++ b/packages/callout/src/Callout.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { css, cx } from '@leafygreen-ui/emotion';
 import { palette } from '@leafygreen-ui/palette';
-import { useBaseFontSize } from '@leafygreen-ui/leafygreen-provider';
 import BulbIcon from '@leafygreen-ui/icon/dist/Bulb';
 import ImportantWithCircleIcon from '@leafygreen-ui/icon/dist/ImportantWithCircle';
 import InfoWithCircleIcon from '@leafygreen-ui/icon/dist/InfoWithCircle';
@@ -166,15 +165,10 @@ interface CalloutProps {
 function Callout({
   variant,
   title,
-  baseFontSize,
+  baseFontSize = 13,
   className,
   children: contents,
 }: CalloutProps) {
-  const providerFontSize = useBaseFontSize();
-  const normalizedProviderFontSize =
-    providerFontSize === 13 || providerFontSize === 16 ? providerFontSize : 13;
-  const fontSize = baseFontSize ?? normalizedProviderFontSize;
-
   const colorSet = colorSets[variant];
   const Icon = headerIcons[variant];
 
@@ -216,9 +210,9 @@ function Callout({
       </div>
       <div className={bodyStyle}>
         {title && (
-          <div className={cx(titleStyle, fontSet[fontSize])}>{title}</div>
+          <div className={cx(titleStyle, fontSet[baseFontSize])}>{title}</div>
         )}
-        <div className={fontSet[fontSize]}>{contents}</div>
+        <div className={fontSet[baseFontSize]}>{contents}</div>
       </div>
     </div>
   );
--- a/packages/callout/src/Callout.tsx
+++ b/packages/callout/src/Callout.tsx
@@ -19,7 +19,7 @@ export const Variant = {
 export type Variant = typeof Variant[keyof typeof Variant];
 
 const baseStyle = css`
-  font-family: Euclid Circular A, ‘Helvetica Neue’, Helvetica, Arial, sans-serif;
+  font-family: Euclid Circular A, ‘Helvetica Neue’, Helvetica, Arial, sans-serif; // TODO: Refresh – remove when fonts are updated
   background-color: ${palette.white};
   border-radius: 12px;
   position: relative;
@@ -61,6 +61,7 @@ const bodyStyle = css`
   display: flex;
   flex-direction: column;
   padding: 16px 24px 16px 54px;
+  font-weight: 300;
 `;
 
 const fontSet = {
--- a/packages/callout/src/Callout.tsx
+++ b/packages/callout/src/Callout.tsx
@@ -37,7 +37,7 @@ const baseStyle = css`
 
 const headerStyle = css`
   font-size: 12px;
-  font-weight: 700;
+  font-weight: 500;
   letter-spacing: 1px;
   line-height: 16px;
   padding: 12px 24px 12px 52px;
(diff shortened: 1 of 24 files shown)


packages/callout/src/Callout.tsx
```diff
@@ -21,33 +19,33 @@ export const Variant = {
 export type Variant = typeof Variant[keyof typeof Variant];
 
 const baseStyle = css`
-  background-color: ${uiColors.white};
-  border-radius: 6px;
-  box-shadow: 0px 2px 5px 0 ${transparentize(0.9, uiColors.gray.dark3)};
-  overflow: hidden;
+  font-family: Euclid Circular A, ‘Helvetica Neue’, Helvetica, Arial, sans-serif; // TODO: Refresh – remove when fonts are updated
```
TheSonOfThomp, line 22: Add note to update to `fontFamilies.default`

packages/callout/src/Callout.tsx
```diff
@@ -21,33 +19,33 @@ export const Variant = {
...
+  font-weight: 700;
+  letter-spacing: 1px;
   line-height: 16px;
-  padding: 12px 24px 12px 52px;
+  padding: 12px 24px 12px 54px;
   position: relative;
   text-transform: uppercase;
   width: 100%;
+  border-top-left-radius: 12px;
+  border-top-right-radius: 11px;
```
TheSonOfThomp, line 48: You should be able to achieve this with `overflow: hidden` on the parent

packages/callout/src/Callout.tsx
```diff
@@ -21,33 +19,33 @@ export const Variant = {
...
 const headerStyle = css`
-  font-family: Akzidenz, ‘Helvetica Neue’, Helvetica, Arial, sans-serif;
   font-size: 12px;
-  font-weight: bold;
-  letter-spacing: 0.8px;
+  font-weight: 700;
+  letter-spacing: 1px;
   line-height: 16px;
-  padding: 12px 24px 12px 52px;
+  padding: 12px 24px 12px 54px;
```
TheSonOfThomp, line 43: Why 54px? I'm seeing 56px in the mocks (unless the extra 2 is somewhere else)

packages/callout/src/Callout.tsx
```diff
@@ -21,33 +19,33 @@ export const Variant = {
...
+    border-radius: 12px 0px 0px 12px;
   }
 `;
 
 const headerStyle = css`
-  font-family: Akzidenz, ‘Helvetica Neue’, Helvetica, Arial, sans-serif;
   font-size: 12px;
-  font-weight: bold;
-  letter-spacing: 0.8px;
+  font-weight: 700;
```
TheSonOfThomp, line 40: I think the mocks are `Medium` (or `500`)

packages/callout/src/Callout.tsx
```diff
@@ -56,20 +54,19 @@ const headerIconStyle = css`
 `;
 
 const titleStyle = css`
-  font-weight: bold;
-  margin-bottom: 6px;
-  line-height: 1;
+  font-weight: 700;
```
TheSonOfThomp, line 57: Looks like the mocks are `500` for this as well

packages/callout/src/Callout.tsx
```diff
@@ -56,20 +54,19 @@ const headerIconStyle = css`
...
 `;
 
 const bodyStyle = css`
   display: flex;
   flex-direction: column;
-  padding: 18px 24px 18px 52px;
+  padding: 16px 24px 16px 54px;
+  font-weight: 300;
```
TheSonOfThomp, line 64: Regular is `400`

packages/callout/src/Callout.tsx
```diff
@@ -182,9 +180,14 @@ function Callout({
         baseStyle,
         css`
           color: ${colorSet.text};
+          border: 1px solid ${colorSet.border};
```
TheSonOfThomp, line 183: Looks like it's 2px in the mocks

packages/callout/src/Callout.tsx
```diff
@@ -182,9 +180,14 @@ function Callout({
...
 
           &:after {
-            background-color: ${colorSet.bar};
+            background: linear-gradient(
+              to left,
+              transparent 5px,
+              ${colorSet.bar} 6px
+            );
```
TheSonOfThomp, line 190: If the parent has `overflow: hidden`, I think this could just be a solid color

TheSonOfThomp (changes requested):

packages/callout/src/Callout.tsx
shaneeza: I avoided `overflow: hidden` on the parent since the right border should not be shown. The colored bar should be fully flush with the edge of the container. It doesn't look like that in figma but Alven and I spoke about it in slack.
Screen Shot 2022-01-24 at 10 47 30 AM

packages/callout/src/Callout.tsx
```diff
@@ -56,20 +54,19 @@ const headerIconStyle = css`
 `;
```
TheSonOfThomp, line 54: Something about this positioning looks off

shaneeza: @TheSonOfThomp Can you share a screenshot?

packages/callout/src/Callout.tsx
shaneeza: actually i'm reverting this back to `52px` since the border width should not be included.

packages/callout/src/Callout.tsx
shaneeza: If I set to `500` it begins to look thinner than the design. The top is `500` and the bottom is `700`:
Screen Shot 2022-01-24 at 11 59 59 AM

Screen Shot 2022-01-24 at 11 59 37 AM

This is from Figma:
Screen Shot 2022-01-24 at 12 02 37 PM

TheSonOfThomp: Try updating from `main`. I think the font definitions have changed slightly

packages/callout/src/Callout.tsx
TheSonOfThomp: Nevermind, you have it right 👍

packages/callout/src/Callout.tsx
shaneeza: @TheSonOfThomp I do have the latest from `main`, maybe we can leave this for visual QA and see what the designers think?

DesignerDave: Figma doesn't use the browser's built in rendering, so some fonts might appear a bit lighter or heavier in one or the other. We should probably resolve this in QA and follow the Figma specifications (even if the rendering is a bit off).

TheSonOfThomp (approved):

PR #1096, opened 2022-01-14
PD-1707 Refresh Text Area

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


packages/text-area/src/TextArea.tsx
```diff
@@ -31,7 +31,7 @@ const textAreaStyle = css`
   width: 100%;
   min-height: ${spacing[6]}px;
   resize: none;
-  border-radius: 4px;
+  /* border-radius: 4px; */
```
shaneeza, line 34: Do you still need this?

shaneeza (commented):
Looking good! Just a few things:

- It does look like the error state is not using the new font, was that intentional?
- It looks like there is a large gap between the label and the description compared to the design. It might be since the line height increased and the label also has a padding bottom.

packages/text-area/src/TextArea.tsx
```diff
@@ -214,14 +214,23 @@ const TextArea: React.ComponentType<
         </Label>
       )}
       {description && (
-        <Description darkMode={darkMode} disabled={disabled}>
+        <Description
+          darkMode={darkMode}
+          disabled={disabled}
+          className={cx({
+            [css`
+              padding-bottom: 4px;
```
shaneeza, line 222: It does look like this component already has `padding-bottom: 4px` for both modes
Screen Shot 2022-01-24 at 5 03 22 PM

shaneeza (approved):

PR #1098, opened 2022-01-19
PD-1709: Refresh Toast

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1709](https://jira.mongodb.org/browse/PD-1709)

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/toast/src/Toast.tsx
+++ b/packages/toast/src/Toast.tsx
@@ -34,14 +34,12 @@ const baseElementStyles: Partial<Record<StyledElements, string>> = {
     width: ${toastWidth}px;
     max-width: calc(100vw - ${spacing[4] * 2}px);
     border-radius: 12px;
-    box-shadow: 0 18px 18px -15px ${transparentize(0.7, palette.black)};
+    box-shadow: 0 16px 32px -14px ${transparentize(0.8, palette.black)};
     overflow: hidden;
     transform: translate3d(0, ${spacing[3]}px, 0) scale(0.95);
     transform-origin: bottom center;
     opacity: 0;
     transition: all ${transitionDuration}ms ease-in-out;
-    font-family: Euclid Circular A, ‘Helvetica Neue’, Helvetica, Arial,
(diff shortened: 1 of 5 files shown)


packages/toast/src/Toast.tsx
```diff
@@ -50,13 +50,13 @@ const baseElementStyles: Partial<Record<StyledElements, string>> = {
   contentWrapper: css`
     display: flex;
     align-items: center;
-    padding: ${spacing[3]}px;
+    padding: 12px 12px 12px 18px;
```
TheSonOfThomp, line 53: Since the icon has built in padding, I think the padding-left should be `16px`

packages/toast/src/Toast.tsx
```diff
@@ -79,153 +79,147 @@ const variantStyles: Record<
 > = {
   [Variant.Success]: {
     toast: css`
-      background-color: ${uiColors.green.light3};
+      background-color: ${palette.green.light3};
+      border: 1px solid ${palette.green.light2};
```
TheSonOfThomp, line 83: Small detail: Might be simpler to just toggle `border-color`, and leave `1px solid` in the base styles

TheSonOfThomp (changes requested):
One other change:

- From the mocks it looks like the icon is `32px`, with a right margin of `13px`

packages/toast/src/Toast.tsx
shaneeza: @TheSonOfThomp I see `28px`, which also happens to be incorrect in my PR lol, where do you see `32px`?

Screen Shot 2022-01-24 at 4 22 54 PM

TheSonOfThomp: The svg itself is `28px`, but the icon component itself is `32px
Screen Shot 2022-01-27 at 11 46 35 AM
`

TheSonOfThomp (approved):
LGTM 👍

PR #1103, opened 2022-01-20
PD-1707 Refresh button

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/button/src/ButtonIcon.tsx
+++ b/packages/button/src/ButtonIcon.tsx
@@ -1,24 +1,27 @@
 import React from 'react';
-import { uiColors } from '@leafygreen-ui/palette';
+import { uiColors, palette } from '@leafygreen-ui/palette';
 import { css, cx } from '@leafygreen-ui/emotion';
 import { ButtonProps, Variant, Size, Mode } from './types';
 
 const iconColor: Record<Mode, Record<Variant, string>> = {
   [Mode.Light]: {
+    [Variant.Default]: css`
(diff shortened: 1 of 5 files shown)


shaneeza (changes requested):
Looking good! A few things:

- The side padding for each size is a few pixels off.
- It looks like for the icon only button, the icon fill should change on hover and focus depending on the variant.
- For `Primary Outline` it looks like the background color should change on focus and hover.

I'm using this [design](https://www.figma.com/file/4h2VwjCuJUbeZ7hzD2J1rq/branch/EafsSYjAX3VYeY832iLiR9/LeafyGreen-Design-System?node-id=4444%3A19384)

packages/button/src/ButtonIcon.tsx
```diff
@@ -90,6 +91,13 @@ const onlyIconColor: Record<Mode, Record<Variant, string>> = {
   },
 };
 
+const onlyIconStyleHover = css`
+  ${ButtonDataProp.selector}:hover &,
```
shaneeza, line 95: Out of curiosity, is this a way to target the parent?

shaneeza (approved):
✨

packages/button/src/ButtonIcon.tsx
TheSonOfThomp: Unfortunately, yes. The other way to do it would be by switching a css variable in the parent styles

PR #1105, opened 2022-01-21
PD-1707 Refresh select

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


shaneeza (changes requested):
A few things:

- Looks like there should be a 1px border around the `<ul>`
- The top and bottom padding for each item might need to be decreased a bit so the height can get down to `32px`. It's about `36px` right now with the padding.

P.S We already spoke, but I wasn't able to check everything as yet since I have to leave early

shaneeza (changes requested):
Some minor things:

- Seems like the placement of the warning icon is a little offScreen Shot 2022-01-27 at 11 24 23 AM
- The red shadow is also missing from the error state on hover
- You can see a little bit of the error message when the dropdown is openScreen Shot 2022-01-27 at 11 27 27 AM

TheSonOfThomp:
@shaneeza Should be fixed now. Let me know if you catch anything else!

shaneeza (approved):
Looks great!!!

PR #1118, opened 2022-02-01
PD-1708 Group 2: Refresh code component

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


shaneeza (changes requested):
Few things:

- [x]  The custom action buttons on the right side are spread out too much

     Screen Shot 2022-02-08 at 3 14 25 PM

       

- [ ]  Not sure if this is a build issue on my end or related to the `select` changes in group 1 but the items in the select look really big: 

    Screen Shot 2022-02-08 at 3 17 49 PM

- [ ] Also related to that image above, there is a gap between the select button and the options list

- [ ] Comments look like they should be `normal`. Currently they are `italic`.

- [ ] Font size looks like it got bumped up to `15px` for the code.

- [ ] Doesn't look like the fake controls are part of the window chrome anymore. 

- [ ] Not sure if this is possible or in scope but it looks like if the code does not wrap then there should be a shadow to indicate scrolling.

TheSonOfThomp:
@shaneeza The item size, and the placement of the menu are build in Select behavior. (the item font size will change with window size). Also, the code font size will be either 12px or 15px depending on the baseFontSize passed into the provider

shaneeza (approved):
Great use of CSS Grid!

PR #1127, opened 2022-02-11
PD-1504 add font sizes to tokens library

TheSonOfThomp:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1504](https://jira.mongodb.org/browse/PD-1504)

## 🛠 Types of changes

## ✅ Checklist


packages/tokens/src/fontSizes.ts
```diff
@@ -0,0 +1,8 @@
+const fontSizes = {
+  defaut: 13,
+  large: 16,
+  legacy: 14,
+  largeCode: 15,
+} as const;
```
TheSonOfThomp, line 6: Not 100% what's best to export. Should we export `font-size` + `line-height` CSS rules instead

DesignerDave: Hmm...

The font sizes we'd probably want to support would be:
- Base (13px)
- Base (16px)
- (Later) Base (18 or 20px)
- Disclaimer
- Code (13px)
- Code (15px)

I don't love names with "large" in them here because it becomes awkward if we need to add another size (which I believe we'll want to do). We should also avoid adding anything that's legacy – we should let those remain hard-coded values until they're updated.

I am a fan of adding line heights to this, if there's a nice way to structure the export. Something like `typeStyles.base13.lineHeight` or something similar could work.

TheSonOfThomp: One issue I have with putting the number in the key is that if we want to change it again, we'll have to update the key, which would be a breaking API change (granted, changing the global font size is a breaking change anyway, but it makes the upgrade smoother)

TheSonOfThomp: In Figma they're defined as `Body1` etc. Using that pattern for now

DesignerDave: I'm with you on the number in the key. That was more _an_ option rather than a good one hah

Sounds good on using body1!

DesignerDave (approved):

PR #1129, opened 2022-02-11
PD-1711: Refresh modals blobs

shaneeza:
## ✍️ Proposed changes

This is a separate PR for the marketing modal blobs.

This is 

## 🛠 Types of changes

## ✅ Checklist


packages/marketing-modal/src/MarketingModal.tsx
```diff
@@ -15,6 +16,14 @@ const Mode = {
 
 type Mode = typeof Mode[keyof typeof Mode];
 
+export const BlobPosition = {
+  TopLeft: 'top left',
+  TopRight: 'top right',
+  BottomRight: 'bottom right',
+} as const;
```
TheSonOfThomp, line 23: No BottomLeft?

packages/marketing-modal/src/MarketingModal.tsx
```diff
@@ -178,6 +191,10 @@ const MarketingModal = ({
       darkMode={darkMode}
       closeIconColor={closeIconColor}
     >
+      {!darkMode &&
+        showBlob &&
+        graphicStyle === GraphicStyle.Center &&
```
TheSonOfThomp, line 196: Why only when GraphicStyle is Center?

packages/marketing-modal/src/blobs.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { BlobPosition } from '.';
+
+type SvgBlobProperties = 'viewBox' | 'path' | 'fill' | 'styles';
+
+const svgBlobPaths: Record<BlobPosition, Record<SvgBlobProperties, string>> = {
+  [BlobPosition.TopLeft]: {
+    viewBox: '0 0 265 501',
+    path:
+      'M-10.0239 -12.0209L154.458 -12.0208C215.515 -12.0208 265 52.2446 265 131.538C265 210.041 215.934 273.615 155.523 273.467L98.8817 273.318C68.886 273.269 44.5623 304.808 44.5623 343.714L44.5623 346.186C44.5623 385.091 20.2765 416.581 -9.64308 416.581L-10.709 416.581C-40.7047 416.581 -64.9905 448.219 -64.9144 487.174C-64.8382 526.129 -89.1241 557.767 -119.12 557.767L-120.604 557.767C-150.524 557.767 -174.772 526.277 -174.81 487.421L-175 201.242C-175.038 162.336 -150.752 130.797 -120.795 130.797L-118.435 130.797C-88.477 130.797 -64.2292 99.2574 -64.2292 60.4015L-64.2292 58.3251C-64.2292 19.5186 -39.9815 -12.0209 -10.0239 -12.0209Z',
+    fill: `${palette.purple.light3}`,
```
TheSonOfThomp, line 13: Should the blob color be customizable?

packages/marketing-modal/src/MarketingModal.story.tsx
```diff
@@ -1,14 +1,20 @@
...
-import MarketingModal, { GraphicStyle } from '.';
+import MarketingModal, { BlobPosition, GraphicStyle } from '.';
 import { CloseIconColor } from '@leafygreen-ui/modal';
 
 function Default() {
   const [open, setOpen] = useState(false);
   const buttonText = text('Button text', 'Okay');
   const linkText = text('Link text', 'Cancel');
   const darkMode = boolean('darkMode', false);
+  const showBlob = boolean('Show Blob', false);
```
TheSonOfThomp, line 12: Can we default this to `true`? 🙃

TheSonOfThomp (approved):
☁ Looks good. Just left a few comments

packages/marketing-modal/src/MarketingModal.tsx
shaneeza: No, there were only 3 options in the design.

packages/marketing-modal/src/MarketingModal.tsx
shaneeza: The designs only show the blob when the graphic is centered. `Fill` is when the image is 100% width of the container so my guess is the modal might become too busy with the blob and the full width image.

packages/marketing-modal/src/blobs.tsx
shaneeza: For now no, but it will probably need to change when `darkMode` is updated

packages/marketing-modal/src/MarketingModal.story.tsx
shaneeza: yes!

packages/marketing-modal/src/MarketingModal.tsx
TheSonOfThomp: That's fair, just wanted to double check

packages/marketing-modal/src/MarketingModal.tsx
TheSonOfThomp: Hmm interesting. I'd double check and make sure that's intentional

packages/marketing-modal/src/MarketingModal.tsx
```diff
@@ -152,6 +161,8 @@ interface MarketingModalProps {
   linkText: string;
   darkMode?: boolean;
   closeIconColor?: CloseIconColor;
```
TheSonOfThomp, line 163: Would be cool if we could programatically set this by checking the color of what's underneath it. Super out-of-scope though 😆

packages/marketing-modal/src/MarketingModal.tsx
shaneeza: Yup, Sooa confirmed that there are only 3.

packages/marketing-modal/src/MarketingModal.tsx
shaneeza: yeah that would be cool

PR #1130, opened 2022-02-11
[PD-1710, PD-1695] group 4 refresh segmented control

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist


packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -279,9 +291,9 @@ const SegmentedControl = React.forwardRef<
   // TODO log warning if defaultValue is set but does not match any child value
   const { usingKeyboard } = useUsingKeyboardContext();
 
-  const [getRef, setRef] = useDynamicRefs<HTMLDivElement>();
+  const getOptionRef = useDynamicRefs<HTMLDivElement>({ prefix: 'option' });
```
shaneeza, line 294: Curious, did you switch over to this so that you could add the prefix?

shaneeza (changes requested):
What a fun PR to look at, I learned so much! A few things:

- [ ] Seems like the option is a little to close to the left edge.

Screen Shot 2022-02-18 at 3 08 10 PM

- [ ] The entire container height of the `large` variant seems a little large, its about `46px` but it looks like it should be `36px` from the design.

packages/segmented-control/src/SegmentedControl.tsx
TheSonOfThomp: Not necessarily. Just switched over to the version in `hooks` instead of the older local version

TheSonOfThomp:
> * [ ]  The entire container height of the `large` variant seems a little large, its about `46px` but it looks like it should be `36px` from the design.

@shaneeza There should be both a `large` variant, and a `16px` font-size variant. Figma seems to be missing the `large` wariant

shaneeza (approved):
LGTM!!

PR #1139, opened 2022-02-18
PD-1762: using Box component for Typography Body

spark3-mdb:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1762](https://jira.mongodb.org/browse/PD-1762)

## 🛠 Types of changes

## ✅ Checklist


packages/typography/src/Typography.tsx
```diff
@@ -30,12 +30,7 @@ const h1 = css`
 
 type H1Props = HTMLElementProps<'h1'>;
 
-const H1: ExtendableBox<H1Props, 'h1'> = ({
-  className,
-  ...rest
-}: {
-  className?: string;
-}) => {
+const H1: ExtendableBox<H1Props, 'h1'> = ({ className, ...rest }: H1Props) => {
```
TheSonOfThomp, line 33: 👍  yeah the old way looked pretty clunky

packages/typography/src/Typography.tsx
```diff
@@ -117,15 +100,19 @@ Subtitle.displayName = 'Subtitle';
 /**
  * Body
  */
-type BodyProps = HTMLElementProps<'div'> & {
+type BodyProps = HTMLElementProps<'p'> & {
```
TheSonOfThomp, line 103: I'd probably keep `BodyProps` using a `div`, to mitigate any breaking style changes (`p` has built in margin that could break some layouts)

packages/typography/src/Typography.tsx
```diff
@@ -117,15 +100,19 @@ Subtitle.displayName = 'Subtitle';
...
    */
   weight?: 'regular' | 'medium';
 };
 
-function Body({ children, className, weight = 'regular', ...rest }: BodyProps) {
+const Body: ExtendableBox<BodyProps, 'p'> = ({
+  className,
+  weight = 'regular',
+  ...rest
+}: BodyProps) => {
```
TheSonOfThomp, line 115: We should add an `as` prop here. My understanding of the ticket is that we want to enable users to write, e.g. `<Body as='"span">` or `<Body as="p">` etc.

TheSonOfThomp (changes requested):
Added a few notes. Let me know if your understanding of thee ticket is different.

Once you add the `as` prop you should add it to the component's `README.md`.
Also, we should add a new knob to Storybook to test the `as` prop.

packages/typography/src/Typography.tsx
spark3-mdb: Having `Body` be a `div` is interesting. It feels like we should make it a `p` and just remove the default margin? Otherwise it feels kinda weird semantically

packages/typography/src/Typography.tsx
spark3-mdb: Makes sense! Should we restrict this to just a few values that make sense for `Body`? Allowing `<Body as="h1" />` feels weird

packages/typography/src/Typography.tsx
TheSonOfThomp: That's a good point. And I just noticed that we're using `sharedStyles` here, which will remove the default margin anyway

packages/typography/src/Typography.tsx
TheSonOfThomp: I feel like that'll get pretty hairy in Typescript. And we're pretty likely to miss one or two tags that do make sense. If someone wants a semantic `h1`  that looks like body text, then they can go ahead 😆

packages/typography/src/Typography.tsx
```diff
@@ -117,33 +100,60 @@ Subtitle.displayName = 'Subtitle';
...
+      ? fontWeights[as][weight]
+      : fontWeights['default'][weight];
 
   const fontWeight = css`
-    font-weight: ${weight === 'regular' ? 400 : 500};
-
-    strong,
-    b {
-      font-weight: ${weight === 'regular' ? 700 : 800};
-    }
```
spark3-mdb, line 138: This wasn't really specced out anywhere, but it's just a minor pet peeve to see magic numbers, so I added this in. Hope that's alright :)

packages/typography/src/Typography.tsx
DesignerDave: What matters in this case is that developers can use any semantic tag regardless of the actual appearance on the page to create a clear page hierarchy for screen readers. Using an `h1` like this may actually be appropriate depending on the circumstance.

packages/typography/src/Typography.tsx
DesignerDave: Unless I'm missing something, the changes here aren't functionally the same. This update doesn't actually style descendent `strong` and `b` tags anymore, where it did before.

In general, I don't actually consider font weight numbers any more magic as using  `font-weight: bold` since the numbers have a very static definition in an application. I think what we previously had here is probably better given the intent and context of the code.

packages/typography/src/Typography.tsx
```diff
@@ -117,33 +100,60 @@ Subtitle.displayName = 'Subtitle';
...
+      medium: 500,
+    },
+    strong: {
+      regular: 700,
+      medium: 800,
+    },
+    b: {
+      regular: 700,
+      medium: 800,
+    },
```
TheSonOfThomp, line 138: 👍  Nice. Might be cleaner to just use an `if` (e.g. if we want to do the same for h# tags), but not critical

packages/typography/src/Typography.tsx
TheSonOfThomp: I think I agree with Dave here, the original is a bit easier to read.
Here's the MDN article on `font-weight`. Using numbers here aren't magic, but actually correspond to weights https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight

packages/typography/src/Typography.tsx
```diff
@@ -117,33 +100,60 @@ Subtitle.displayName = 'Subtitle';
 /**
  * Body
  */
-type BodyProps = HTMLElementProps<'div'> & {
+
+type BodyFontWeight = 'regular' | 'medium';
+type BodyProps = HTMLElementProps<'p'> & {
```
TheSonOfThomp, line 105: There are likely some less-common elements that will have valid attributes that will have invalid props according to TS. (e.g. `<label htmlFor="...">` or `<a href="...">`)

packages/typography/src/Typography.tsx
spark3-mdb: I guess what I meant more when I said magic number wasn't to say that the keyword values themselves were magic, but that they were magic within the context of our design system. For example, the fact that 800 corresponds to `as="strong" + weight="medium"` is just implied with the ternary operator instead of being explicitly defined.

Still going to change it back! Just thought I'd clarify. 😂]

TheSonOfThomp: Not necessarily, since `Body` styles cascade still. A common use case of this component is to write docs etc. so, something like this is pretty common:
```
<Body as="p" weight="medium">
  <strong>All courses</strong> at MongoDB University are <em>free</em>!
</Body>
```

TheSonOfThomp (changes requested):
We should revert back to using CSS for font-weight

packages/typography/src/Typography.tsx
spark3-mdb: Would they be using those tags like `<Body as="label" htmlFor="...">` though? How we define BodyProps shouldn't affect children.

packages/typography/src/Typography.tsx
spark3-mdb: I see. I envisioned the use case for Typography to feel more like the way [Mui does it](https://mui.com/components/typography/)

packages/typography/src/Typography.tsx
TheSonOfThomp: The way the props are typed currently, `<Body as="label" htmlFor="...">` would be invalid (since `htmlFor` is not a valid `div` prop). The solution likely involves some tricky TS generics to do it right
(or we can just allow any other prop)

I'd start with something like:
```
type BodyProps<T extends keyof JSX.IntrinsicElements> = HTMLElementProps<T> & {
```

TheSonOfThomp: I think I have a TS solution if you want to sync at some point

packages/typography/src/Typography.tsx
TheSonOfThomp: As far as I can tell, you'd do it pretty similarly in Mui https://codesandbox.io/s/typographytheme-material-demo-forked-stpn8z?file=/index.js

spark3-mdb: The browser default weight for `<strong>` is coming in that demo, `<Typography>` is setting the weight to 400; it doesn't try to interfere with anything in the children. It's generally fine because there shouldn't be any larger components inside a Body component, but setting top-level CSS here just felt icky.

packages/typography/src/Typography.tsx
```diff
@@ -102,29 +102,50 @@ Subtitle.displayName = 'Subtitle';
...
+      medium: 500,
+    },
+    strong: {
+      regular: 700,
+      medium: 800,
+    },
+    b: {
+      regular: 700,
+      medium: 800,
+    },
```
TheSonOfThomp, line 138: Not really sure it's worth separating `strong` & `b`

spark3-mdb: I agree. I left it initially to demonstrate how we can have 1-to-1 mapping, but it's not that important now we're taking this middle ground route.

packages/typography/src/Typography.tsx
```diff
@@ -117,31 +100,55 @@ Subtitle.displayName = 'Subtitle';
...
+  } = {
+    default: {
+      regular: 400,
+      medium: 500,
+    },
+    strong: {
+      regular: 700,
+      medium: 800,
+    },
+  } as const;
```
TheSonOfThomp, line 135: 👍

TheSonOfThomp (approved):
🚀

PR #1146, opened 2022-02-24
PD-1372: Changing callout heading to Overline as h3

spark3-mdb:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1372](https://jira.mongodb.org/browse/PD-1372)

## 🛠 Types of changes

## ✅ Checklist


packages/callout/src/Callout.tsx
```diff
@@ -208,7 +205,20 @@ function Callout({
...
+        <Overline
+          as="h3"
+          className={cx(
+            css`
+              color: inherit;
+              font-size: 12px;
+              font-weight: 500;
+              letter-spacing: 1px;
+              line-height: 16px;
+            `,
```
spark3-mdb, line 217: Not sure if this text should have one-off styling that isn't based on the theme or other centralized values for font size, weight, etc., but keeping things as is for parity.

packages/callout/src/Callout.tsx
```diff
@@ -208,7 +205,20 @@ function Callout({
           className={headerIconStyle}
           role="presentation"
         />
-        {headerLabels[variant]}
+        <Overline
+          as="h3"
```
TheSonOfThomp, line 209: Should the overline be `h2`, and the title be `h3`?

packages/callout/src/Callout.tsx
TheSonOfThomp: The updated `Callout` design files use the same styles as Typography's `Overline` https://www.figma.com/file/4h2VwjCuJUbeZ7hzD2J1rq/LeafyGreen-Design-System?node-id=5821%3A44855

TheSonOfThomp: This seems like a change that happened after Shaneeza finished the redesign

TheSonOfThomp (changes requested):
I think the semantic hierarchy should be a bit different. Also the styles in Figma have updated

packages/callout/src/Callout.tsx
spark3-mdb: It looks like the font-weight and letter-spacing are still different, so I've just kept those 2 lines now.

TheSonOfThomp: I think font weight should be `700` still (we need to document this somewhere, but `Semibold` in Figma translates to bold/700 in CSS)

TheSonOfThomp (approved):
🎸

PR #1147, opened 2022-02-24
PD-1609: Fixing disabled anchor onClick bug

spark3-mdb:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1609](https://jira.mongodb.org/browse/PD-1609)

## 🛠 Types of changes

## ✅ Checklist


packages/button/src/Button.spec.tsx
```diff
@@ -143,6 +135,17 @@ describe('packages/button', () => {
...
 
+    test('does not fire onClick handler on disabled anchor', () => {
+      const onClick = jest.fn();
+      const { button } = renderButton({
+        disabled: true,
+        as: 'a',
+        onClick,
+      });
+      fireEvent.click(button);
+      expect(onClick.mock.calls.length).toBe(0);
```
TheSonOfThomp, line 146: I realize this is following the pattern of the previous test, but I think `expect(onClick).toHaveBeenCalledTimes(0)` is a bit cleaner

packages/button/README.md
```diff
@@ -80,19 +80,19 @@ npm install @leafygreen-ui/button
...
+| `disabled`     | `boolean`                                                | Disabled the button                                                                                                                                                                                             | `false`                                                                                                                           |
+| `as`           | `'a'                                                     | 'button'`                                                                                                                                                                                                       | Determines the root element. An `a` tags can be supplied to replace `button` from being the DOM element that wraps the component. | `button` |
```
TheSonOfThomp, line 91: Heads up, using pipes in md table text creates a new column. We use commas to represent unions here (or use`\|`)

.changeset/shaggy-cheetahs-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/button': minor
```
TheSonOfThomp, line 2: I'd probably call this a patch change. It's more a bug fix than a new feature

packages/button/src/Button.tsx
```diff
@@ -99,31 +101,15 @@ const Button: ExtendableBox<
...
-    disabled: boolean,
-  ): 'a' | 'button' => {
-    if (isAnchor && !disabled) return 'a';
-    return 'button';
-  };
+  const isAnchor: boolean = (!!rest.href || as === 'a') && !disabled;
 
   const buttonProps = {
-    type,
+    type: type || 'button',
```
TheSonOfThomp, line 107: I think type should still be `undefined` if it's an anchor

packages/button/src/Button.tsx
```diff
@@ -99,31 +101,15 @@ const Button: ExtendableBox<
...
-
-  // Render a disabled link as a button to retain focusability
-  const getButtonTag = (
-    isAnchor: boolean,
-    disabled: boolean,
-  ): 'a' | 'button' => {
-    if (isAnchor && !disabled) return 'a';
-    return 'button';
-  };
+  const isAnchor: boolean = (!!rest.href || as === 'a') && !disabled;
```
TheSonOfThomp, line 104: 👍

TheSonOfThomp (changes requested):
Just a few small things 👍

packages/button/README.md
TheSonOfThomp: i.e. `'a'  | 'button'` created an extra column

packages/button/src/Button.tsx
spark3-mdb: I defaulted the value here to `'button'` because of what used to be on line 107. Should we change this?

packages/button/src/types.ts
```diff
@@ -38,6 +38,8 @@ interface ButtonProps {
   leftGlyph?: React.ReactElement;
   rightGlyph?: React.ReactElement;
   href?: string;
+  type?: JSX.IntrinsicElements['button']['type'];
+  as?: keyof JSX.IntrinsicElements;
```
spark3-mdb, line 42: Might need to restrict these values to just `"a" | "button"`

.changeset/shaggy-cheetahs-smoke.md
spark3-mdb: What should the value be if it's a patch change? Is it just `'@leafygreen-ui/button': patch`?

packages/button/src/Button.tsx
TheSonOfThomp: `type` would still be undefined there if `as=='a'` or if `href` is defined

TheSonOfThomp: My understanding here is that this prop allows devs to set the [button type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type) (e.g. submit, reset) on a LG button

packages/button/src/Button.tsx
```diff
@@ -99,31 +101,15 @@ const Button: ExtendableBox<
...
 
   const buttonProps = {
-    type,
+    type: type || (!isAnchor && 'button'),
```
TheSonOfThomp, line 107: Is this not going to resolve to just a Boolean if type is not set?

TheSonOfThomp (approved):
🚘

PR #1156, opened 2022-03-07
PD-1820: Stepper Redesign

spark3-mdb:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1820](https://jira.mongodb.org/browse/PD-1820)

## 🛠 Types of changes

## ✅ Checklist


packages/stepper/src/EllipseStep.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const EllipseStep = (props: React.PropsWithChildren<EllipseStepProps>) => {
+  return (
+    <Tooltip
+      align={Align.Top}
+      justify={Justify.Middle}
+      trigger={<Step {...props}>{props.children}</Step>}
+      triggerEvent={TriggerEvent.Hover}
+    >
```
spark3-mdb, line 19: TODO: style tooltip

spark3-mdb: TODO: style hover state

packages/stepper/src/NewStepper.tsx
```diff
@@ -0,0 +1,108 @@
...
+import EllipseStep from './EllipseStep';
+
+const Stepper = ({
+  children,
+  currentStep,
+  maxDisplayedSteps = Array.isArray(children) ? children.length : 1,
+  className,
+}: StepperProps) => {
+  // Constants
+  const COMPLETED_STEPS_SHOWN = 1; // Steps shown after prior ellipses and current step
```
spark3-mdb, line 14: This could be a new prop. Needs testing.

packages/stepper/src/EllipseStep.tsx
```diff
@@ -0,0 +1,78 @@
...
+    [StepCompletionStates.CompletedMultiple]: completedMultipleStyles,
+    [StepCompletionStates.UpcomingMultiple]: upcomingMultipleStyles,
+  };
+
+  return (
+    <Tooltip
+      align={Align.Top}
+      justify={Justify.Middle}
+      trigger={
+        <Step
```
spark3-mdb, line 59: Currently, the `<Tooltip>` trigger doesn't work unless wrapped in a div. This is annoying because it either requires 1) the CSS rules for the connector lines to move up the component tree to the Stepper component, or 2) for those rules to be defined twice.

Going to leave the Tooltip unresolved until I can find a workaround.

packages/stepper/src/Stepper.story.tsx
```diff
@@ -13,21 +13,24 @@ storiesOf('Stepper', module)
...
-        <Step>Billing</Step>
-        <Step>Address</Step>
-        <Step>Confirmation</Step>
+        <div>Overview</div>
+        <div>Configuration</div>
+        <div>Update</div>
+        <div>Install</div>
+        <div>Billing</div>
+        <div>Address</div>
+        <div>Confirmation</div>
```
spark3-mdb, line 25: One major-ish change I made is to not require the use of the `<Step>` component, because below is the implementation of the old `<Step>` component:

```
export const Step = function Step({
  children,
}: {
  children: StepContents;
}): StepElement {
  return <>{children}</>;
};
```

I didn't really see the point in that, so I changed it to just wrap each child in the new Step logic.

packages/stepper/src/Stepper.story.tsx
```diff
@@ -13,21 +13,24 @@ storiesOf('Stepper', module)
...
       </Stepper>
     </div>
   ))
   .add('Many', () => {
-    let currentStep = number('Step', 0, { min: 0 });
-    const numSteps = number('Number of steps', 7, { min: 1 });
+    let currentStep = number('Step', 1, { min: 0 });
+    const numSteps = number('Number of steps', 10, { min: 1 });
+    const maxDisplayedSteps = number('Max displayed', 5, { min: 1 });
+    const completedStepsShown = number('Completed steps shown', 3, { min: 1 });
```
spark3-mdb, line 33: This is a new optional prop that changes where the current step is rendered.

Below is what the previous rendering logic would default to. In the new implementation, this is equivalent to `completedStepsShown = 1`, which is the default value.
image

But since the stepper has 5 visible steps, it's probably nicer to render the current step in the center like below. This is `completedStepsShown = 2`
image

shaneeza (commented):
@spark3-mdb I haven't done a full review as yet but from a quick glance locally on my end on the `many` tab, when I change the step the component starts to re-render like every second. I noticed because the style tab in the dev tools kept flickering so I stuck a console log in the component and it logs every second.

packages/stepper/src/Stepper.spec.tsx
```diff
@@ -1,306 +1,142 @@
...
-      {
-        allSteps,
-        maxDisplayedSteps,
-        currentStep: 0,
-      },
-      render,
-    );
-    assertVisibleSteps({ allSteps, currentStep: 0 });
+  describe('basic rendering', () => {
+    test('renders steps', async () => {
```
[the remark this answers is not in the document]
spark3-mdb: I don't think we need it - I needed it for the `axe` call above and I just used that same line. I'll remove it!

packages/stepper/README.md
```diff
@@ -30,42 +39,36 @@ npm install @leafygreen-ui/stepper
...
-  <div class="leafygreen-ui-plizcc">
-    <div
-      class="leafygreen-ui-1cyt2ch"
-      aria-label="Step 1"
-      aria-current="step"
-      step="0"
-    >
-      <div class="leafygreen-ui-7uty9f">
-        <div class="leafygreen-ui-gbxl1s">1</div>
+<ol class="leafygreen-ui-1pvgjhr">
```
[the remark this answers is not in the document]
spark3-mdb: I think `breadcrumbs` is meant more for elements that allow navigation, but maybe I'll use `aria-label="progress"` here!

packages/stepper/src/Stepper.story.tsx
```diff
@@ -26,26 +24,45 @@ storiesOf('Stepper', module)
...
-    const numSteps = number('Number of steps', 7, { min: 1 });
+    let currentStep = number('Step', 1, { min: 0 });
+    const numSteps = number('Number of steps', 10, { min: 1 });
+    const maxDisplayedSteps = number('Max displayed', 5, { min: 1 });
+    let completedStepsShown = number('Completed steps shown', 3, { min: 1 });
+    const darkMode = boolean('Dark mode', false);
 
     // Can't dynamically change the max, so we manually enforce it
-    if (currentStep + 1 > numSteps) {
+    if (currentStep >= numSteps) {
```
TheSonOfThomp, line 34: Looks like we can't set currentStep == numSteps

packages/stepper/src/StepperContext.tsx
```diff
@@ -0,0 +1,45 @@
...
+  darkMode,
+  children,
+}: {
+  darkMode: boolean;
+  children: ReactNode;
+}) => {
+  // TODO: This dark mode state should ideally be managed in a global state.
+  const [isDarkMode, setIsDarkMode] = useState<boolean>(darkMode);
+  const stepIconClassName = createUniqueClassName();
+  const stepLabelClassName = createUniqueClassName();
```
TheSonOfThomp, line 25: These classnames don't really need to be in `context`. They are static values that can just be exported from a constants file. As written we need to generate a new classname for each stepper

packages/stepper/src/InternalStep.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ...rest
+}: PropsWithChildren<InternalStepProps & React.HTMLProps<HTMLDivElement>>) {
+  const {
+    isDarkMode,
+    stepIconClassName,
+    stepLabelClassName,
+  } = useStepperContext();
+  const isCurrent = state === StepStates.Current;
+
+  const baseStyles = css`
```
TheSonOfThomp, line 27: Same comment as before, we shouldn't define styles within render unless absolutely necessary

packages/stepper/src/Step.tsx
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+
+const Step = React.Fragment;
+export default Step;
```
TheSonOfThomp, line 4: If we have a `Step` component, we should consider doing a bit more with it (see `Select/Option`)

packages/stepper/src/StepIcon.tsx
```diff
@@ -0,0 +1,124 @@
...
+  if (state === StepStates.Completed) {
+    return <CheckmarkIcon />;
+  } else if (
+    state === StepStates.UpcomingMultiple ||
+    state === StepStates.CompletedMultiple
+  ) {
+    return <EllipsisIcon />;
+  } else {
+    // if Current (single) or Upcoming (single)
+    // TODO: Use centralized, reusable font-weight value
```
TheSonOfThomp, line 21: Do you mean something like `tokens.fontWeight.medium` etc?

packages/stepper/src/StepIcon.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  return (
+    <div
+      className={cx(
+        stepIconClassName,
+        baseStyles,
+        styles[isDarkMode ? Mode.Dark : Mode.Light][state],
+      )}
+    >
+      <Icon state={state} {...rest} />
```
TheSonOfThomp, line 119: Small nit: It's a bit confusing reading `<Icon>` when we have an LG component with the same name. I'd be a bit more explicit with something like `StepIconGlyph`

packages/stepper/src/Stepper.tsx
```diff
@@ -1,331 +1,107 @@
...
+        {hasPriorSteps && (
+          <EllipsesStep
+            state={StepStates.CompletedMultiple}
+            startingStepIndex={1}
+            tooltipContent={childrenArray.slice(0, firstDisplayedStep)}
+          >
+            {getStepRangeText(1, firstDisplayedStep)}
+          </EllipsesStep>
+        )}
+        {React.Children.map(
```
TheSonOfThomp, line 78: 🔥 this is clean 👍

packages/stepper/src/types.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export interface StepperContextValues {
+  isDarkMode: boolean;
+  setIsDarkMode: Dispatch<React.SetStateAction<boolean>>;
+  stepIconClassName: string;
+  stepLabelClassName: string;
+}
+
+// TODO: replace with common type across all components
+export const Mode = {
```
TheSonOfThomp, line 11: Don't forget to export `type Mode = typeof Mode[keyof typeof Mode];` 🙃

packages/stepper/src/types.ts
```diff
@@ -0,0 +1,63 @@
...
+  Light: 'light',
+  Dark: 'dark',
+} as const;
+
+export enum StepStates {
```
TheSonOfThomp, line 16: We've had some TS issues with `enum` in the past. That's why we prefer the `const X = {...} as const` pattern

TheSonOfThomp (changes requested):

packages/stepper/src/Stepper.story.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Then again, does that make sense?

packages/stepper/src/StepIcon.tsx
spark3-mdb: Yeah. I know we've gone back and forth on this one a bunch though haha.

I think another issue here is the fact that our Typography components' APIs aren't flexible enough for this markup to be clean. In Mui, this would be something like `<Typography variant="body1" weight="medium" />` which feels nicer.

packages/stepper/src/Stepper.story.tsx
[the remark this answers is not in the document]
spark3-mdb: If we have 10 steps, the last step has index 9, so it makes sense to not allow the current step to be 10.

packages/stepper/src/EllipsesStep.tsx
```diff
@@ -0,0 +1,76 @@
...
+const EllipsesStep = ({
+  state,
+  children,
+  tooltipContent,
+  startingStepIndex,
+  ...rest
+}: React.PropsWithChildren<EllipsesStepProps>) => {
+  const { isDarkMode, stepIconClassName } = useStepperContext();
+
+  // TODO: would be good to put these styles inside a baseStyles, but it is currently impossible because the <Tooltip> content is an iframe.
```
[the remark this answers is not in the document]
spark3-mdb: Yeah I'm not sure of the underlying problem, but if I had a baseStyles that was applied to the wrapper of the Tooltip and selected the tooltip using a className, the styles wouldn't be applied.

packages/stepper/src/InternalStep.tsx
```diff
@@ -0,0 +1,146 @@
...
+    }
+  `;
+
+  const styles: Record<StepStates, string> = {
+    [StepStates.CompletedMultiple]: completedMultipleStyles,
+    [StepStates.Completed]: completedStyles,
+    [StepStates.Current]: currentStyles,
+    [StepStates.Upcoming]: upcomingStyles,
+    [StepStates.UpcomingMultiple]: upcomingMultipleStyles,
+  };
```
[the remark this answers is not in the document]
spark3-mdb: That's a great catch - I did this because I wanted to keep this pattern of having just base + variant styles defined here. I'd say the 'correct' workaround here would be to have a `StepLabel` child component that handled these. I'll push this up in a second.

packages/stepper/src/Step.tsx
spark3-mdb: I think the `Step` component isn't really needed - below is what we used to have:
```
export const Step = function Step({
  children,
}: {
  children: StepContents;
}): StepElement {
  return <>{children}</>;
};
```
I've been told though that applications are using the `Step` component in their apps, so just having something exported to pass children through makes sure their apps don't break. 

I've thought about things to add, but because every `Step` label content is styled through children being passed directly through, devs have full flexibility over their styles. Customization of styles on other parts of the component should probably be limited.

packages/stepper/src/Stepper.story.tsx
```diff
@@ -26,26 +24,45 @@ storiesOf('Stepper', module)
...
           currentStep={currentStep}
-          maxDisplayedSteps={number('Max displayed', 5, { min: 1 })}
+          completedStepsShown={completedStepsShown}
+          maxDisplayedSteps={maxDisplayedSteps}
+          darkMode={darkMode}
         >
-          {times(numSteps, count => (
-            <Step key={count}>Step {count + 1}</Step>
+          {times(numSteps, (count: number) => (
+            <div key={count}>Step {count + 1}</div>
```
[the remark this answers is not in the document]
spark3-mdb: Gonna resolve this since we've got a comment thread on the `Step` component!

packages/stepper/src/types.ts
spark3-mdb: Oh yeah the `const` pattern is fine imo. I was more referring to how all this should probably be somewhere globally accessible, not redefined in every component.

packages/stepper/src/InternalStep.tsx
spark3-mdb: I think you meant inside the component, not within render? I do need the `isDarkMode` variable though, so I'll use it here.

packages/stepper/src/types.ts
spark3-mdb: Why do we need this? Not having `type Mode = typeof Mode[keyof typeof Mode];` doesn't seem to interfere with how we write dark mode styles.

Unless the syntax for dark mode styles I wrote in the `StepIcon` component isn't how we want to be writing dark mode styles - lmk if that's the case.

packages/stepper/src/Stepper.tsx
```diff
@@ -1,331 +1,107 @@
...
-
-export function Stepper({
+import React, { PropsWithChildren } from 'react';
+import { css, cx } from '@leafygreen-ui/emotion';
+import Step from './InternalStep';
+import { StepStates, StepperProps } from './types';
+import EllipsesStep from './EllipsesStep';
+import StepperContextProvider from './StepperContext';
+
+const Stepper = ({
```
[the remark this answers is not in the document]
spark3-mdb: Is there a way to add prop validations to check values of props against other props? I'm not personally familiar. 

We could definitely have our own checks to throw our own errors though. I didn't play around with that idea because the previous version of Stepper didn't check for values.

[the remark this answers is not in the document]
TheSonOfThomp: You could get fancy with Typescript prop types, but I'd just recommend a check in the render function that warns in the console (`lib.consoleOnce.warn`) when the props don't make sense

packages/stepper/src/Step.tsx
TheSonOfThomp: I think it's needed to make it clear that we're adding elements & styles to the markup that a user provides. It's not immediately intuitive what something like this would do 
```js
<Stepper>
  <H1>Some step</H1>
  <H2>Some other step</H2>
</Stepper>
```

but this is a bit clearer that what I passed in is going to be rendered as a step
```js
<Stepper>
  <Step>
    <H1>Some step</H1>
  </Step>
  <Step>
    <H2>Some other step</H2>
  <Step>
</Stepper>
```

TheSonOfThomp: On the other side, it doesn't make much sense to use a standalone `<Step/>`, so we should add the restrictions like we have in the [Option component](https://github.com/mongodb/leafygreen-ui/blob/main/packages/select/src/Option.tsx#L302-L304)

packages/stepper/src/types.ts
TheSonOfThomp: It allows you to use `Mode` as both a constant and a type. We may not need it in this component, but it's best practice

packages/stepper/src/Stepper.story.tsx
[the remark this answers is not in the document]
spark3-mdb: Yeah what I mean is that it's by design. If there are 10 steps, the max index is 9. So if you set it to 10 or higher, I've set it to `numSteps - 1` on line 37.

packages/stepper/src/Stepper.story.tsx
```diff
@@ -26,26 +24,45 @@ storiesOf('Stepper', module)
...
     }
 
+    // Can't dynamically change the max, so we manually enforce it
+    if (completedStepsShown >= maxDisplayedSteps - 2) {
+      channel.emit(CHANGE, {
+        name: 'Completed steps shown',
+        value: maxDisplayedSteps - 2,
+      });
+      completedStepsShown = maxDisplayedSteps - 2;
+    }
```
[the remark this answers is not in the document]
spark3-mdb: That only happens when there's more than 1 step, which shouldn't really happen. The line displaying is a minor bug I can fix, but I think we should support `maxDisplayedSteps = 1` because it renders like this when there's just one step:

image

packages/stepper/src/Stepper.tsx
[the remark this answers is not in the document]
spark3-mdb: As discussed, because it's a good amount of extra effort and maintains parity with the pre-redesign Stepper, this work has been added to the backlog here: https://jira.mongodb.org/browse/PD-1964

packages/stepper/src/EllipsesStep.tsx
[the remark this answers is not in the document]
TheSonOfThomp: I think I understand. Styles for tooltips need to be defined directly on the tooltip itself

shaneeza (approved):

packages/stepper/src/EllipsesStep.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const upcomingMultipleStyles = css`
+    &:hover .${stepIconClassName} {
+      // TODO: use centralized box-shadow value
+      box-shadow: 0px 0px 0px 3px
+        ${isDarkMode ? palette.gray.dark2 : palette.gray.light2};
+    }
+  `;
+
+  const stepStyles: Record<EllipsesStepStates, string> = {
```
TheSonOfThomp, line 42: I like this pattern for stateful styles 👍

packages/stepper/src/InternalStep.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { palette } from '@leafygreen-ui/palette';
+import { spacing } from '@leafygreen-ui/tokens';
+import { useStepperContext } from './StepperContext';
+import StepLabel from './StepLabel';
+import { stepIconClassName } from './constants';
+
+const Step = ({
+  children,
+  index,
+  state,
```
TheSonOfThomp, line 14: ~~💡 Could consider using context for `state`~~

packages/stepper/src/InternalStep.tsx
```diff
@@ -0,0 +1,91 @@
...
+        position: absolute;
+        top: ${iconSize / 2}px;
+        left: 50%;
+        z-index: 0;
+        background-color: ${isDarkMode
+          ? palette.gray.light1
+          : palette.gray.base};
+      }
+    `}
+  `;
```
TheSonOfThomp, line 59: ⚠️  Should add this conditionally with `cx` instead of interpolating a boolean here

packages/stepper/src/StepIcon.tsx
```diff
@@ -0,0 +1,125 @@
...
+      [StepStates.UpcomingMultiple]: darkUpcomingStyles,
+    },
+    [Mode.Light]: {
+      [StepStates.CompletedMultiple]: completedStyles,
+      [StepStates.Completed]: completedStyles,
+      [StepStates.Current]: currentStyles,
+      [StepStates.Upcoming]: upcomingStyles,
+      [StepStates.UpcomingMultiple]: upcomingStyles,
+    },
+  };
```
TheSonOfThomp, line 110: 😬 None of this is dependent on state/props, so it should be defined outside the component. (Even the styles that are dependent on props/state should be defined as an object or function outside the component. It's a subjective thing, but a pattern we've established in the codebase)

packages/stepper/src/StepLabel.tsx
```diff
@@ -0,0 +1,62 @@
...
+  `;
+
+  const upcomingMultipleStyles = css`
+    text-decoration-line: underline;
+    text-decoration-style: dotted;
+    text-underline-position: under;
+    color: ${isDarkMode ? palette.gray.light1 : palette.gray.dark1};
+  `;
+
+  const styles: Record<StepStates, string> = {
```
TheSonOfThomp, line 40: 🤔 I'd maybe write this outsize the component as a `Record<StepStates, Record<Mode, string>>` (I know we've discussed potential better ways to do doubly conditional styles)

packages/stepper/src/InternalStep.tsx
TheSonOfThomp: Scratch that. We'd need a state provider for each step...

packages/stepper/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+import { createUniqueClassName } from '@leafygreen-ui/lib';
+
+export const stepLabelClassName = createUniqueClassName();
```
TheSonOfThomp, line 3: 👍

packages/stepper/src/types.ts
TheSonOfThomp: If this is only used internally then it's fine to keep it as an `enum`, but anything we export from `index` would need to be a `const as const`

TheSonOfThomp (commented):
Almost there. Just a few code style changes I've flagged

packages/stepper/src/StepLabel.tsx
spark3-mdb: Yeah it feels related to the comment on the `InternalStep` - it feels kinda verbose to me and a little harder to compose styles.

```
const completedMultipleStyles = css`
  color: ${isDarkMode ? palette.green.base : palette.green.dark2};
`

const styles: Record<StepStates, string> = {
  [StepStates.CompletedMultiple]: completedMultipleStyles,
}
```
vs.

```
const darkModeCompletedMultipleStyles = css`
  color: ${palette.green.base};
`

const lightModeCompletedMultipleStyles = css`
  color: ${palette.green.dark2};
`

const styles: Record<StepStates, string> = {
  [Mode.Dark]: {
    [StepStates.CompletedMultiple]: darkModeCompletedMultipleStyles,
  },
  [Mode.Light]: {
    [StepStates.CompletedMultiple]: lightModeCompletedMultipleStyles,
  }
}
```
Not to mention - what if dark mode and light mode share some styles? Then in the second option, we'd need to have a separate `baseCompletedMultipleStyles` to compose in `cx`.

packages/stepper/src/InternalStep.tsx
TheSonOfThomp: We don't really need to be conservative with cssString variables—they don't add much overhead (and I'm unclear as to how this would drastically increase the number of vars in this case).
We've established a pattern where if we have a block of CSS that is toggled by some prop, it should be added using `cx` to make it clear what's toggling the styles.

spark3-mdb: OH I accidentally thought you were referring to the `isDarkMode` ternary. Using `cx` based on the props makes sense to me. I'll update this.

TheSonOfThomp: From our conversation yesterday, this looks fine 👍. I'd prefer this to be outside the component for performance & maintenance but that doesn't need to block this PR

TheSonOfThomp (approved):
Styles should be written outside the component in the future (for performance & maintenance) but this PRs been open long enough we should get it in soon

PR #1158, opened 2022-03-08
PD-1745 Refresh Checkbox (PD-1681, PD-1881)

TheSonOfThomp:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1745](https://jira.mongodb.org/browse/PD-1745)
Also resolves:
[PD-1681](https://jira.mongodb.org/browse/PD-1681) - adds support for aria-label/aria-labelledby
[PD-1881](https://jira.mongodb.org/browse/PD-1881) - adds description

## 🛠 Types of changes

## ✅ Checklist


packages/checkbox/src/LegacyCheck.tsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
```
shaneeza, line 1: I like that you separated this

packages/checkbox/src/Checkbox.story.tsx
```diff
@@ -32,18 +32,21 @@ storiesOf('Checkbox', module)
...
-        indeterminate={boolean('Indeterminate', false)}
-        label={text('Label', 'I agree to this thing.')}
-        bold={boolean('Bold', false)}
-        animate={boolean('Animate', true)}
-        className={css`
-          padding: 20px;
-          background-color: ${darkMode ? uiColors.gray.dark3 : uiColors.white};
-        `}
-      />
+      <LeafygreenProvider>
```
shaneeza, line 35: Curious why you removed bold?

packages/checkbox/src/Checkbox.tsx
```diff
@@ -26,25 +21,52 @@ const Mode = {
...
-const wrapperStyleAnimated = css`
-  transition: 300ms opacity ease-in-out;
-`;
+const checkBoxSize = 14;
+const checkAnimationDuration = 100;
+const hypotenusePct = 100 * Math.sqrt(2); // relative distance from corner to corner
+const insetPct = 100 - hypotenusePct;
+const flourishScale = 2.25;
+const flourishTransitionScale = 4;
+const flourishTransitionDelay = -flourishTransitionScale / flourishScale;
```
shaneeza, line 30: This is pretty cool. Did you think of this yourself or follow a tutorial?

shaneeza (commented):
I love the animation!

One thing I noticed: 

- [ ] The hover state for darkMode is odd:

Screen Shot 2022-03-17 at 1 41 12 PM

packages/checkbox/src/Checkbox.tsx
TheSonOfThomp: I just kinda made it up 😆

packages/checkbox/src/Checkbox.story.tsx
TheSonOfThomp: Doesn't look like the Figma component supports a `bold` prop anymore

TheSonOfThomp: Seems like there is a wider desire to `unbold` the Label component, but that should be done in a wider ticket
https://jira.mongodb.org/browse/PD-1689

shaneeza (commented):
- [x] The alignment for the description is a little off in `darkMode` compared to `lightMode`

Screen Shot 2022-03-22 at 12 30 52 PM

- [x] In `lightMode` when `Indeterminate` is `true`, you can still see the flourish animation if you click twice on the label

- [x] The flourish animation seems to be missing from `Controlled` and `No Label` in storybook

- [x] Not sure if its a local issue but I don't see any focus styles

packages/checkbox/src/Check.tsx
```diff
@@ -0,0 +1,179 @@
...
+  rippleTransitionScale,
+  insetPct,
+} from './constants';
+import SvgCheck from './SvgCheck';
+import SvgIndeterminate from './SvgIndeterminate';
+import { useUsingKeyboardContext } from '@leafygreen-ui/leafygreen-provider';
+import { CheckProps } from './types';
+
+const checkWrapperBaseStyle = css`
+  --lg-checkbox-border-color: ${palette.gray.dark2};
```
DesignerDave, line 19: This CSS variable is only used in this file. Might be a good time to start discussing the merits of CSS vs JS variables:
```
border: 2px solid var(--lg-checkbox-border-color);
```
vs
```
border: 2px solid ${borderColor};
```

I'm inclined to think we should be using the latter pattern most of the time for a number of reasons:
- Value interpolation over CSS vars keeps the CSS declarations to only property declarations where possible.
- We have to interpolate the value anyways in the variable declaration.
- We're interpolating JS variables elsewhere in this and other blocks.
- Where relevant, we can use polished color transformation functions with JS variables, but not with CSS variables.
- We don't need to worry about the variable being named to avoid collisions (`--lg-checkbox-border-color` vs `borderColor`).
- Finding where a CSS variable is declared can be much harder than finding a JS variable declaration, making it harder to understand what code is doing. JS variables are either declared in the file, or imported from another, making it clear to see what the value is. CSS variables can be effectively declared anywhere, and don't leave a similar trail.
- "Side effects" are built into how CSS variables work. You can't reliably expect the value of a CSS variable to not change, and tracking bugs related to them can be more difficult as a result.

If you're not sold, we could spend some time chatting about this today at the group code review 😄

packages/checkbox/src/Check.tsx
```diff
@@ -0,0 +1,179 @@
...
+      <div
+        className={cx(selector, checkWrapperBaseStyle, {
+          [checkWrapperCheckedStyle]: showCheckIcon,
+          [checkWrapperDisabledStyle]: disabled,
+          [checkWrapperCheckedDisabledStyle]: disabled && showCheckIcon,
+        })}
+      >
+        <Transition
+          in={showCheckIcon}
+          timeout={shouldAnimate ? checkAnimationDuration : 0}
```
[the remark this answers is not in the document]
TheSonOfThomp: The alternative would be `enter={shouldAnimate} exit={shouldAnimate}`, which is probably a bit better, yeah

packages/checkbox/src/Check.tsx
TheSonOfThomp: I think it's very much a case-by-case consideration. I think the rule-of-thumb you presented is a fair one, (i.e.  don't use CSS vars that are only used in one file)

In this particular case, I think I would agree that JS interpolation is probably the clearer choice. In particular because of how it's used in `checkWrapperCheckedDisabledStyle` on Line 78. It's not immediately clear what the value for the var will be in this since it depends on the order in which `checkWrapperCheckedStyle` and `checkWrapperDisabledStyle` are passed into `cx` on Line 152 (affecting the cascade). Since we'll always want "checked + disabled" to be gray, it's probably best to hard code it in this circumstance.

DesignerDave: I think that inheritance is both CSS variables' largest strength, and weakness. If we're utilizing that inheritance heavily, we're simultaneously making it more difficult to understand what's happening – especially if we change the value at any point.

The third alternative, while boring, is to simply share a const between files. A prop or context might just be overcomplicating what is, in this case (I believe), a static variable.

If we wanted to turn this into a case where CSS variables uniquely simplifies a problem, if we were to implement `prefers-reduced-motion`, I believe we could do it universally much more simply here using CSS variables. Which reminds me, we should probably do this! 😄

packages/checkbox/src/Checkbox.tsx
```diff
@@ -66,86 +103,40 @@ const inputStyle = css`
...
-const checkboxStyleAnimated = css`
-  transition: 500ms transform steps(29);
+const labelTextStyle = css`
+  align-self: baseline;
 `;
 
-const checkboxStyleChecked = css`
-  transform: translate3d(${-width + height}px, 0, 0);
+const descriptionStyle = css`
+  grid-area: description;
```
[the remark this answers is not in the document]
TheSonOfThomp: I initially had the check, label & description as siblings, so grid was needed. Not strictly necessary now that there's more nesting. 
As for which is "better", I'm not sure there's really an answer there, but flex is maybe a little clearer

packages/checkbox/src/Checkbox.tsx
```diff
@@ -26,25 +22,56 @@ const Mode = {
...
+  gap: 8px;
+  align-items: center;
+  justify-content: flex-start;
+  cursor: pointer;
+`;
+
+const labelHoverSelector = `
+  &:hover 
+    > .${inputClassName}:not([disabled]):not(:focus) 
+      + .${checkClassName}
```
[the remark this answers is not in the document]
TheSonOfThomp: "When the Label is hovered, and the Input within it is not focused or disabled, select that Input's sibling with XYZ class". 

I suppose that could be rewritten as:
```
&:hover:not(:focus-within) > .input:not([disabled]) + .check
```

or, if we outsource the disabled check to JS, simply:

```
&:hover:not(:focus-within) > .check
```
and conditionally apply the styles to the label
```
[labelHoverStyle]: !disabled
```

packages/checkbox/src/Check.tsx
TheSonOfThomp: There's no universal check that we do for `prefers-reduced-motion`, but there should be!

Reduced motion is the main reason I'm using CSS vars here in [Checkbox.tsx](https://github.com/mongodb/leafygreen-ui/blob/8f64a85536b4201d113b6d1908ee344a210c9783/packages/checkbox/src/Checkbox.tsx#L39)

TheSonOfThomp: It might actually be useful to create a hook than a global var:
```
const prefersReducedMotion = useReducedMotion()
...
cx({
  [animationStyles] : !prefersReducedMotion
})
```
where
```
function useReducedMotion() {
  return window.matchMedia('prefers-reduced-motion: reduce').matches
}

Comment (DesignerDave) [reply packages/checkbox/src/Check.tsx]: We already have a function that generates styles that are applied when the setting is:

```js
<div className={prefersReducedMotion('display: none')} />
```

See `a11y/src/prefersReducedMotion`

TheSonOfThomp: Looking into it more, a context seems like overkill for this since the value will hardly ever change.

I think an imported query selector is a little more usable than the function we have. It follows the Emotion guidance on media queries, and will support syntax highlighting.

e.g.
```js
const reducedMotionQuery = '@media (prefers-reduced-motion: reduce)`
...

const styles = css`
  transition-duration: 100ms;

  ${reducedMotionQuery} {
    transition-duration: 0ms;
  }

`

```

packages/checkbox/src/Checkbox.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Related to the first comment, CSS variables could potentially simplify things here.

```css
/* Checkbox.tsx */
label:hover:not(:focus-within) > .input:not([disabled]) {
  --lg-shadow: 0 0 0 3px ${palette.gray.light2};
}

.input:focus {
  --lg-shadow: 0 0 0 2px ${palette.white}, 0 0 0 4px ${palette.blue.light1};
}


/* Check.tsx */
.checkbox {
  box-shadow: var(--lg-shadow);
}

```

PR #1168, opened 2022-03-18
[PD-1873, PD-1874, PD-1875] Updates to Menu

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

### DIFF
--- a/packages/menu/src/Menu.story.tsx
+++ b/packages/menu/src/Menu.story.tsx
@@ -70,7 +70,7 @@ function Controlled() {
           <MenuItem href="http://mongodb.design" size={size}>
             I am a link!
           </MenuItem>
-          <MenuSeparator />
+          <MenuSeparator margin={0} />
           <MenuItem size={size}>Left out of the MenuGroup</MenuItem>
         </Menu>
       </Button>
--- a/packages/menu/src/MenuSeparator.tsx
+++ b/packages/menu/src/MenuSeparator.tsx
@@ -1,13 +1,30 @@
 import * as React from 'react';
-import { css } from '@leafygreen-ui/emotion';
+import { css, cx } from '@leafygreen-ui/emotion';
 import { palette } from '@leafygreen-ui/palette';
 
 const borderStyle = css`
   border-top: 1px solid ${palette.gray.dark2};
 `;
 
-function MenuSeparator() {
-  return <li role="separator" className={borderStyle} />;
+function MenuSeparator({
+  margin = 0,
+}: {
+  /**
+   * Vertical margin fo the separator
+   */
+  margin?: number;
+}) {
+  return (
+    <li
+      role="separator"
+      className={cx(
+        borderStyle,
+        css`
+          margin: ${margin}px 0;
+        `,
+      )}
+    />
+  );
 }
 
 MenuSeparator.displayName = 'MenuSeparator';


packages/menu/src/SubMenu.tsx
```diff
@@ -242,80 +223,6 @@ const SubMenu: ExtendableBox<
...
-        linkStyle,
-        {
-          [activeMenuItemContainerStyle]: active,
-          [disabledMenuItemContainerStyle]: disabled,
-          [subMenuOpenStyle]: open,
-          [focusedMenuItemContainerStyle]: showFocus,
-        },
-        className,
-      ),
-    };
```
TheSonOfThomp, line 278: Notable differences (that I should likely address):
 
`SubMenu`'s implementation includes `onClick` and `aria-haspopup` (while `MenuItem` does not)

`MenuItem`'s implementation includes `tabIndex`,  `aria-disabled` and `disabled`

TheSonOfThomp: Additionally, `MenuItem` spreads `anchorProps` on the box, while `SubMenu` does not

packages/menu/src/SubMenu.tsx
```diff
@@ -242,80 +223,6 @@ const SubMenu: ExtendableBox<
...
-          <div
-            className={cx(descriptionTextStyle, {
-              [activeDescriptionTextStyle]: active,
-              [disabledTextStyle]: disabled,
-              [focusStyles.descriptionStyle]: showFocus,
-              [linkDescriptionTextStyle]: typeof rest.href === 'string',
-            })}
-          >
-            {description}
-          </div>
```
TheSonOfThomp, line 303: Notably, this node was always rendered, even when no `description` was provided

packages/menu/src/SubMenu.tsx
```diff
@@ -325,9 +232,31 @@ const SubMenu: ExtendableBox<
 
     return (
       <li role="none" className={liStyle}>
-        {renderBox}
+        <MenuItem
```
TheSonOfThomp, line 235: @DesignerDave Here's the issue. The root element of `MenuItem` is a `<li>`, so here we have two nested `<li>`s

packages/menu/src/SubMenu.tsx
```diff
@@ -325,9 +232,31 @@ const SubMenu: ExtendableBox<
...
+          className={cx(
+            subMenuStyle,
+            {
+              [subMenuOpenStyle]: open,
+            },
+            className,
+          )}
+          disabled={disabled}
+          active={active}
+          size={'default'}
```
DesignerDave, line 250: Nit: This should probably either use the size enum for the component, or just be a string (`size='default'`).

packages/menu/src/SubMenu.tsx
DesignerDave: Ahh, yeah nice catch. Now the reasoning why it was the way it was makes sense. Are you gonna try to update this or revert it?

TheSonOfThomp: I'm moving it to a different PR, but I'll try and fix it

TheSonOfThomp:
@DesignerDave This PR is now pretty minimal, and only tackles the Selector change. The MenuItem changes are moved to #1169

packages/menu/src/Menu.story.tsx
```diff
@@ -70,7 +70,7 @@ function Controlled() {
           <MenuItem href="http://mongodb.design" size={size}>
             I am a link!
           </MenuItem>
-          <MenuSeparator />
+          <MenuSeparator margin={0} />
```
DesignerDave, line 73: This breaks a long-standing pattern we've intentionally avoided of directly adding CSS properties as component props (with the exception of `z-index`). Is there any reason we aren't setting a default margin through CSS and allowing a className override here instead of exposing a margin prop?

DesignerDave (changes requested):

packages/menu/src/Menu.story.tsx
TheSonOfThomp: Maybe I misunderstood your comment here? https://jira.mongodb.org/browse/PD-1873

TheSonOfThomp: I feel like we wouldn't want other CSS properties to be overridden, but maybe that's overly restrictive

DesignerDave: I suppose, if we are going to give the ability to override the spacing, it should be through className. The component probably doesn't need it at all though.

TheSonOfThomp: The alternative, in cases where spacing is needed, would be to wrap the separator in a div, and apply the spacing to that. I think it's better to just support that on the separator itself

DesignerDave: Since it's not something that we support now, it might be best to wait until there's a use-case for the override to add an API for one.

TheSonOfThomp: Updated to use `className`

PR #1170, opened 2022-03-21
PD-1874 PD-1875 Updates Menu Item Bold

TheSonOfThomp:
## ✍️ Proposed changes

 - [PD-1874](https://jira.mongodb.org/browse/PD-1874) - Replace pseudoBold with a pseudo element
 - [PD-1875](https://jira.mongodb.org/browse/PD-1875) - Removes `href` from description

## 🛠 Types of changes

## ✅ Checklist


packages/tabs/src/TabTitle.tsx
```diff
@@ -143,6 +140,23 @@ const listTitle = css`
     outline: none;
   }
 
+  // We create a pseudo element that's the width of the bolded text
+  // This way there's no layout shift on hover when the text is bolded.
+  &:before {
+    content: attr(data-text);
+    content: attr(data-text) / '';
```
DesignerDave, line 147: What is this duplicate content attribute doing? Locally, it doesn't seem to be doing much other than overriding the first instance of this property.

packages/tabs/src/TabTitle.tsx
```diff
@@ -143,6 +140,23 @@ const listTitle = css`
...
+    height: 0;
+    font-weight: 700;
+    visibility: hidden;
+    overflow: hidden;
+    user-select: none;
+    pointer-events: none;
+
+    @media speech {
```
DesignerDave, line 155: I actually wasn't sure – is the content CSS property read by screen readers? If so, good to know!

packages/tabs/src/TabTitle.tsx
DesignerDave: Ditto for Menu

DesignerDave (commented):
This is a really great approach to the problem, and it looks really good. I tried to break it, both by adding an inline component, and by adding a margin to a word within the text in a menu item. In both instances, this approach seemed to work well. Great work.

I left two small comments on this PR, but otherwise looks good!

packages/tabs/src/TabTitle.tsx
TheSonOfThomp: It will be read out by screen readers yeah. Just noticed though that `@media screen` is [deprecated in MQ4](https://drafts.csswg.org/mediaqueries/#media-types). I'll update this appropriately

TheSonOfThomp: Looking further `visibility: hidden` will do the same thing that the `@media speech` rule is trying to do

DesignerDave: Ahh perfect! Forgot about that.

packages/tabs/src/TabTitle.tsx
DesignerDave: My understanding of the slash is that it's alt text, not a fallback. Since we're hiding this from screen readers, _I think_ it's safe here to remove this.

TheSonOfThomp: Ah, yeah you're right

packages/menu/src/SubMenu.tsx
```diff
@@ -324,32 +231,53 @@ const SubMenu: ExtendableBox<
...
+          glyph={glyph}
         >
-          <ChevronIcon
-            role="presentation"
-            className={chevronIconStyles}
-            size={14}
-          />
-        </IconButton>
+          {title}
+          <IconButton
```
TheSonOfThomp, line 255: The fact that adding this absolute-ly positioned button as a child of `MenuItem` is a bit lucky. If we add `position: relative;` to anything in the `MenuItem` in the future, this will break.

A more future-proof solution would be
1) Add a `rightGlyph` prop to `MenuItem`
2) separate the item contents into a 3rd internal component that is then consumed by both `MenuItem` and `SubMenu`

DesignerDave: I'm a bit confused why it's lucky. Of course it will break if we change position properties of MenuItem, but since it's dependent on other styles within the same package, we should be able to manage that change fine without exposing a new public API or making the implementation more complicated.

packages/menu/src/SubMenu.tsx
```diff
@@ -410,7 +334,7 @@ const SubMenu: ExtendableBox<
             </ul>
           )}
         </Transition>
-      </li>
+      </>
```
TheSonOfThomp, line 337: The submenu `ul` is no longer a child of the trigger `li`, but rather a sibling. Do we foresee this causing any issues?

TheSonOfThomp: Reverted these changes, but aligned how `SubMenu` and `MenuItem`  implement this behavior. Moved the consolidation to #1176

packages/menu/src/styles.tsx
```diff
@@ -62,6 +62,16 @@ export const menuItemContainerStyle = css`
...
 
+export const menuItemHeight: Record<Size, string> = {
+  [Size.Default]: css`
+    min-height: 34px;
+  `,
+
+  [Size.Large]: css`
+    min-height: 46px;
+  `,
+};
```
spark3-mdb, line 73: Since these are height properties, it feels like these should be multiples of 8px? If you set it to these values directly from Figma, maybe it's worth confirming with design.

packages/menu/src/Menu.story.tsx
```diff
@@ -89,7 +89,7 @@ function SubMenuExample() {
       <Menu trigger={<Button>trigger</Button>}>
         <SubMenu
           title="Menu Item 1"
-          description="https://google.com"
+          description=".design"
```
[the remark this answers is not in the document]
TheSonOfThomp: Description should be the text displayed in the submenu, showing it here as explicitly _not_ the link url

packages/menu/src/styles.tsx
TheSonOfThomp: Default is 32 + 2x2 padding, Large is 32 + 2x7 padding. I'll double check the 7px 👍

packages/menu/src/SubMenu.tsx
```diff
@@ -290,66 +299,72 @@ const SubMenu: ExtendableBox<
...
+            className={cx(iconButtonStyle, {
+              [openIconButtonStyle]: open,
+              [iconButtonFocusedStyle]: showFocus,
+            })}
+            onClick={(e: React.MouseEvent) => {
+              // we stop the event from propagating and closing the entire menu
+              e.nativeEvent.stopImmediatePropagation();
+
+              if (setOpen) {
+                setOpen(!open);
```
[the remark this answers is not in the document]
TheSonOfThomp: No, `setOpen` has the signature `(value: boolean) => void`

packages/menu/src/SubMenu.tsx
```diff
@@ -290,66 +299,72 @@ const SubMenu: ExtendableBox<
...
+              }
+            }}
+          >
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah, that's probably better

packages/menu/src/SubMenu.tsx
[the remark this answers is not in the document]
spark3-mdb: That's weird that that's the signature, because `setOpen` on the Menu component has the signature `React.Dispatch<React.SetStateAction<boolean>>`, which is what I expected here. Should the type signature just be changed?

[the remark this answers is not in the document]
TheSonOfThomp: You'd still call that type the same way. That type just means it was created by calling `useState`.

e.g.
```
const [open,  setOpen] = useState<boolean>(false)
...
setOpen(true) // type is `React.Dispatch<React.SetStateAction<boolean>>`
```

TheSonOfThomp:
> A couple of small things in the comments! A few other things that give me itches -
> 
> * I was a bit confused about the variables being named things like `submenuButtonA` & `submenuButtonB`. Usually when I see these non-descriptive names I wonder if there are things we can do using componentization to get around this? I haven't really dug into the component though, so I didn't leave this as a comment. It also isn't a change that you introduced, so that's another thing :)

Those names are just in the test files as a way to identify he menu items. They just needed some unique name. Could easily have been `firstSubMenuButton` etc.

> * Another thing that you didn't introduce but gave me an itch - the multiple definitions of `fontWeight: 700`, instead of something like `fontWeight: fontWeights.bold`. Maybe this is a topic for our next code review session. These might be kinda painful to hunt down later.

`font-weight: 700` actually has a [semantic meaning ](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight)of `bold`. I agree it's a bit obfuscated, but it isn't a magic number. Abstracting it out into another constant seems a bit overkill to me

packages/menu/src/SubMenu.tsx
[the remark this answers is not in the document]
spark3-mdb: Yeah I understand that. What I'm saying is that the type of `setOpen` on SubMenu should probably match the signature of `setOpen` on Menu.

When it comes to a state dispatch call that depends on the previously stored value, it's recommended that we call it in the form of `setOpen(open => !open`. I've found a quick [thread](https://stackoverflow.com/questions/40359800/how-to-toggle-boolean-state-of-a-react-component#answer-48394541) that also links out to React documentation.

spark3-mdb:
> > A couple of small things in the comments! A few other things that give me itches -
> > 
> > * I was a bit confused about the variables being named things like `submenuButtonA` & `submenuButtonB`. Usually when I see these non-descriptive names I wonder if there are things we can do using componentization to get around this? I haven't really dug into the component though, so I didn't leave this as a comment. It also isn't a change that you introduced, so that's another thing :)
> 
> Those names are just in the test files as a way to identify he menu items. They just needed some unique name. Could easily have been `firstSubMenuButton` etc.
> 
> > * Another thing that you didn't introduce but gave me an itch - the multiple definitions of `fontWeight: 700`, instead of something like `fontWeight: fontWeights.bold`. Maybe this is a topic for our next code review session. These might be kinda painful to hunt down later.
> 
> `font-weight: 700` actually has a [semantic meaning ](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight)of `bold`. I agree it's a bit obfuscated, but it isn't a magic number. Abstracting it out into another constant seems like over-optimization

Yes, but with this suggestion, I'm not talking about what 'bold' means semantically in the world of CSS; I'm talking about it within the context of a design system/theme. For example, some font faces naturally have some thick strokes, so when used in a design system, we may want `fontWeight: 600` to be what's considered bold. In other words, if we don't support this abstractions, we're telling designers that the word 'bold' for them, can only be `font-weight: 700` for the font we're currently using. If we chose to, this is also a pattern that would allow us to do things like `theme.fontWeights.label`, or `theme.fontWeights.[whatever we want]`.

packages/menu/src/SubMenu.tsx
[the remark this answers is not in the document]
TheSonOfThomp: I agree that the props for `SubMenu`, `MenuItem` and `Menu` should be more unified. That'll require some refactoring of the types that's a bit out of scope for this work. I have a draft PR that starts this refactor in #1176

TheSonOfThomp:
> Yes, but with this suggestion, I'm not talking about what 'bold' means semantically in the world of CSS; I'm talking about it within the context of a design system/theme. For example, some font faces naturally have some thick strokes, so when used in a design system, we may want `fontWeight: 600` to be what's considered bold. In other words, if we don't support this abstractions, we're telling designers that the word 'bold' for them, can only be `font-weight: 700` for the font we're currently using. If we chose to, this is also a pattern that would allow us to do things like `theme.fontWeights.label`, or `theme.fontWeights.[whatever we want]`.

I'm not sure I agree. We should discuss at code review

spark3-mdb (approved):
🚀

PR #1180, opened 2022-03-31
PD-1542 right tooltip alignment

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1542](https://jira.mongodb.org/browse/PD-1542)

## 🛠 Types of changes

## ✅ Checklist

#### Fixes a bug that wrapped the content element when the trigger was align to the right of a scroll container or page.
How to replicate:
   - If the placement of the reference element is to the right edge of the scroll container or page
   - `usePortal=true`
   - `justify` is set to `middle`, `start`, or `end`
   - `align` is either `top` or `bottom`
  
Under these conditions the content element would sometimes wrap or position itself off to the right of the container. The actual expected should be that the content element should not wrap and it should not position itself off the screen.

Screen Shot 2022-04-12 at 12 41 33 PM

Fix: 
Screen Shot 2022-04-12 at 12 43 10 PM

---
   
#### Fixes a bug that made the width of the content element too narrow when the justification was `fit` and inside a `scrollContainer`.
How to replicate:
  - Using a `scrollContainer` 
 - `justify` is set to `fit`
  - `align` is either `top` or `bottom`
 
Screen Shot 2022-04-12 at 12 41 49 PM

Under these conditions the content element width would be too narrow. The expected should be that the content element should span the width of the tigger element.

Fix: 
Screen Shot 2022-04-12 at 12 42 42 PM


packages/popover/src/positionUtils.spec.ts
```diff
@@ -2840,6 +3009,7 @@ describe('positionUtils', () => {
             referenceElViewportPos: refElPos.center,
             contentElDocumentPos: contentElPos,
             contentElViewportPos: contentElPos,
+            scrollContainer: scrollContainerNull,
```
TheSonOfThomp, line 3012: Oof. That's a lot of duplication

packages/popover/src/positionUtils.ts
```diff
@@ -122,8 +146,21 @@ export function getElementDocumentPosition(
...
+  } = element.getBoundingClientRect();
+  const { offsetHeight: height, offsetWidth } = element;
+
+  // offsetWidth returns a rounded number of the element's layout width and height.
+  // boundingWidth returns an exact number with the rendered width and height which can include transformations.
+  // Using the exact number is a better indicator of determining if an element will fit in the window, e.g. calcLeft() uses the width to determine the left position and then that number is used to check if it is safetly within the window bounds. In some cases the offsetWidth is rounded up which means that it might not fit in the window bounds when it really can.
+  // With this we check if the difference between the boundingWidth and offsetWidth is less than one, if thats the case then the number was rounded to a whole number and we should use the exact number instead. However if the number is greater than one then that means that the boundingWidth is returning a width that has a transformation applied to it and we don't want that number. We want the untransformed width.
+  const width =
+    Math.abs(boundingWidth - offsetWidth) < 1 ? boundingWidth : offsetWidth;
```
TheSonOfThomp, line 163: 🐞 Small nit: extracting `Math.abs(boundingWidth - offsetWidth) < 1` into a var called something like `wasRounded` would make this a bit clearer

packages/popover/src/positionUtils.ts
```diff
@@ -165,8 +202,17 @@ export function getElementViewportPosition(
...
+    top,
+    bottom,
+    left,
+    right,
+    width: boundingWidth,
+  } = element.getBoundingClientRect();
+  const { offsetHeight: height, offsetWidth } = element;
+
+  const width =
+    Math.abs(boundingWidth - offsetWidth) < 1 ? boundingWidth : offsetWidth;
```
TheSonOfThomp, line 215: 🏜 Can we DRY this?

TheSonOfThomp (approved):
Looks good 👍 . Could you add repro steps to the description?

PR #1199, opened 2022-04-11
PD-1941 dark mode form elements

TheSonOfThomp:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-1941](https://jira.mongodb.org/browse/PD-1941)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/tokens/src/typeScales.ts
```diff
@@ -1,10 +1,17 @@
...
+
 const typeScales = {
   body1: {
-    fontSize: 13,
+    fontSize: BaseFontSize.Body1,
     lineHeight: 20,
   } as const,
   body2: {
-    fontSize: 16,
+    fontSize: BaseFontSize.Body2,
```
spark3-mdb, line 14: 🔥🔥🔥

packages/typography/src/FormElements/Label.tsx
```diff
@@ -0,0 +1,75 @@
...
+  [Mode.Dark]: css`
+    color: ${palette.gray.light1};
+  `,
+};
+
+type LabelProps = HTMLElementProps<'label', never> & {
+  darkMode?: boolean;
+  htmlFor: string;
+  disabled?: boolean;
+};
```
spark3-mdb, line 47: 🤷 Could we move these types for both Description and Label to `types.ts`?

spark3-mdb (changes requested):
I like how you refactored here! The comments themselves are minor, but I'm not able to see the Typography page on Storybook. Is that just me?

Update: the page came back for me. I just had to build again.

packages/typography/src/FormElements/Label.tsx
TheSonOfThomp: In principle, yes, but I feel like these interfaces are small enough that they can live here. If I were to move them I'd prefer to refactor each component into its own directory, for example `FormElements/Label/types.ts` etc

spark3-mdb: I like that idea of having separate directories! I'll leave the call up to you.

spark3-mdb (approved):

packages/typography/src/FormElements/Label.tsx
TheSonOfThomp: Again, I think in principle yes, but for a small component like this I think it's easier to keep it all in one file. The threshold is blurry for when a component should be broken into multiple files

PR #1205, opened 2022-04-13
PD-1941 form elements font size fix

TheSonOfThomp:
## ✍️ Proposed changes

Adds a temporary `useUpdatedBaseFontSize` hook within `Typography` to convert from old base font size values to new ones.

This will get removed once we update `useBaseFontSize` to return the correct values, but that won't happen until after the redesign has been very widely adopted

🎟 _Jira ticket:_ [PD-1941](https://jira.mongodb.org/browse/PD-1941)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/typography/src/useUpdatedBaseFontSize.ts
```diff
@@ -0,0 +1,13 @@
+import { useBaseFontSize } from '@leafygreen-ui/leafygreen-provider';
+import { BaseFontSize } from '@leafygreen-ui/tokens';
+
+/**
+ * A wrapper around the hook useBaseFontSize that returns
+ * a value that is consistent with the updated base font sizes.
+ *
+ * Prefer using {@link useBaseFontSize} when LeafygreenProvider is updated to support the new values.
+ */
+export const useUpdatedBaseFontSize = (): BaseFontSize => {
```
DesignerDave, line 10: Can we document this in a way that clearly lets folks know that this API isn't sticking around in the long term? This will need to be a "public-ish" API so we can use it on other components.

packages/typography/src/useUpdatedBaseFontSize.ts
```diff
@@ -0,0 +1,13 @@
...
+  const baseFontSize = useBaseFontSize();
+  return baseFontSize === 16 ? baseFontSize : 13;
+};
```
DesignerDave, line 13: We'll need to export this from the package's `index.ts`.

DesignerDave (changes requested):
Left a couple notes. Also, this change is missing documentation, as well as a changeset.

packages/typography/src/useUpdatedBaseFontSize.ts
TheSonOfThomp: If it's only public-ish, wouldn't importing it from `/dist/useUpdatedBaseFontSize` be fine?

TheSonOfThomp: Also, I went back & forth on adding this in `typography` vs in `leafygreen-provider` (where the original hook is). Went with typography since I'd need to add `tokens` as  a dependency to the provider, which is a bigger change than I think this warrants. Thought?

DesignerDave: I'd rather us not import that way. It's a bit more brittle if we make build or package structure changes (though I'm aware we do this with Icon).

I don't mind it living in this package because it's temporary.

packages/typography/src/useUpdatedBaseFontSize.ts
DesignerDave: People won't reliably see a TS doc annotation, but might be good for us to get into the habit with it.

PR #1238, opened 2022-04-28
Update Combobox Story CSF

spark3-mdb:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/combobox/src/Combobox.story.jsx
```diff
@@ -0,0 +1,175 @@
+import React from 'react';
```
TheSonOfThomp, line 1: We should stick to using Typescript in stories. We can catch so many more bugs this way.

TheSonOfThomp (changes requested):

packages/combobox/src/Combobox.story.jsx
spark3-mdb: I'm using JSX in some of these stories not because TS is too verbose - there are some potentially painful issues sometimes with having all the controls be dynamically generated when using TS. I don't convert from TS to JS unless it saves me a ton of headaches.

These are 'Docs Only' PRs, so it would be great to just get these merged in so we can deprecate knobs ASAP. There are lot of components I need to get through in the next few weeks. We can come back to converting these to TS later.

TheSonOfThomp: I'm just worried since Combobox is a very TS-heavy component. Bearing the TS headaches now could save worse ones down the road

packages/combobox/src/Combobox.story.tsx
```diff
@@ -1,254 +1,179 @@
...
-  >
-    <LeafygreenProvider>{children}</LeafygreenProvider>
-  </div>
+export default {
+  title: 'Packages/Combobox',
+  component: Combobox,
+  parameters: {
+    controls: { exclude: ['children', 'setError'] },
+  },
+} as ComponentMeta<typeof Combobox>;
```
TheSonOfThomp, line 14: 👍 🙃

TheSonOfThomp (approved):

PR #1243, opened 2022-05-02
PD-1620 Adds `isGlyph` property to generic Icon

TheSonOfThomp:
## ✍️ Proposed changes

Adds the `isGlyph` property to the generic `<Icon />` component, meaning it can now be used in Select and Combobox options. 

Additionally, overloads the `isComponentGlyph` function to accept either a `ReactNode` and `ComponentType`

🎟 _Jira ticket:_ [PD-1620](https://jira.mongodb.org/browse/PD-1620)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/icon/src/isComponentGlyph.ts
```diff
@@ -1,13 +1,34 @@
...
+  if (isValidElement(child)) {
+    return (
+      child != null &&
+      typeof child === 'object' &&
+      'type' in child &&
+      (child.type as any).isGlyph === true
+    );
+  }
+
+  // If we've recieved a component function
```
spark33, line 25: typos!

spark33 (changes requested):
I'm wondering if we can leverage `isComponentType` here. Whether it's  `isComponentType(glyph, 'Icon')`,  `isComponentType(glyph, 'LGIcon')`, `isComponentType(glyph, 'Glyph')` or something along those lines.

Adding a custom prop to check for this feels extraneous and not very "React".

TheSonOfThomp:
> I'm wondering if we can leverage `isComponentType` here. Whether it's `isComponentType(glyph, 'Icon')`, `isComponentType(glyph, 'LGIcon')`, `isComponentType(glyph, 'Glyph')` or something along those lines.
> 
> Adding a custom prop to check for this feels extraneous and not very "React".

I dug a little deeper into this, and while I think this would be a good idea, it breaks how we identify Icon components. 

Currently the `displayName` of any individual icon component is the name of that icon (e.g. `Beaker` etc.). Changing the displayName to `Icon` would break any tests relying on this `displayName`. I think we should file a separate ticket to make this change

spark33 (approved):

PR #1256, opened 2022-05-10
PD-2046 Combobox Backdrop Click

TheSonOfThomp:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-2046](https://jira.mongodb.org/browse/PD-2046)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

```
export const PageTest = (args: any) => {
  return (
    <div
      className={css`
        width: 256px;
      `}
    >
      <Button onClick={() => console.log('Click button')}>Test Button</Button>
      <TextInput label="Test" placeholder="Some input" />
      <Combobox {...args} />
    </div>
  );
};
PageTest.args = Basic.args;
```

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/combobox/src/Combobox.tsx
```diff
@@ -1059,19 +1062,37 @@ export default function Combobox<M extends boolean>({
...
+   * No other click event handlers should fire on backdrop click
+   *
+   * 1. Mousedown event fires
+   * 2. We prevent `mousedown`'s default behavior, to prevent focus from being applied to the body (or other target)
+   * 3. Click event fires
+   * 4. We handle this event on _capture_, and stop propagation before the `click` event propagates all the way to any other element.
+   *  This ensures that even if we click on a button, that handler is not fired
+   * 5. Then we call `closeMenu`, setting `isOpen = false`, and rerender the component
+   */
+  useEventListener('mousedown', (mousedown: MouseEvent) => {
```
DesignerDave, line 1078: Do we see it problematic that we're adding a new event listener for every Combobox component on the page? Usually we're protected by React using event delegation under the hood, but this doesn't use React synthetic events.

packages/combobox/src/Combobox.tsx
```diff
@@ -1059,19 +1062,37 @@ export default function Combobox<M extends boolean>({
...
+    if (!doesComponentContainEventTarget(mousedown) && isOpen) {
+      mousedown.preventDefault(); // Prevent focus from being applied to body
```
DesignerDave, line 1080: Also, @spark3-mdb @shaneeza would love for more than just myself to brainstorm ways this could possibly go wrong 😅 definitely nervous about this change breaking in some way.

packages/combobox/src/Combobox.tsx
TheSonOfThomp: Yeah, that's a fair concern. I think the solution to that would be to do our own event delegation within `useEventHandler`, no?

TheSonOfThomp: Unless there's a way to add a React Synthetic event to the document (or react root) programmatically like this? I don't know of a way to do that though

DesignerDave: Talked offline – this event listener should only be registered when the combobox is open.

DesignerDave (approved):
This looks like it behaves as intended in testing 👍

PR #1286, opened 2022-05-25
PD-2061 popover transition

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-2061](https://jira.mongodb.org/browse/PD-2061)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/popover/src/Popover.story.tsx
```diff
@@ -57,10 +103,37 @@ const Template = ({ buttonText, ...args }: PopoverStoryProps) => {
...
-export const Basic = Template.bind({});
+export const ScrollableContainer = ({
+  refButtonPosition,
+  buttonText,
+  ...args
+}: PopoverStoryProps) => {
+  const [active, setActive] = useState<boolean>(false);
+  const [portalContainer, setPortalContainer] = useState<HTMLElement | null>(
+    null,
+  );
```
spark33, line 114: Could use `useRef` instead?

packages/popover/src/Popover.story.tsx
```diff
@@ -57,10 +103,37 @@ const Template = ({ buttonText, ...args }: PopoverStoryProps) => {
...
+            portalContainer={portalContainer}
+            scrollContainer={portalContainer}
+          >
+            <div className={popoverStyle}>Popover content</div>
+          </Popover>
+        </button>
+      </div>
+    </div>
+  );
+};
```
spark33, line 139:
```
ScrollableContainer.args = {
  usePortal: true,
};
ScrollableContainer.argTypes = {
  usePortal: { control: 'none' },
}
```

packages/popover/src/Popover.story.tsx
shaneeza: Yes!

packages/popover/src/Popover.story.tsx
shaneeza: I was able to remove some controls but I kept running into an error when setting the default `usePortal: true,`

packages/popover/src/Popover.story.tsx
```diff
@@ -40,15 +76,24 @@ export default {
...
+      options: ['centered', 'top', 'right', 'bottom', 'left'],
+      control: { type: 'select' },
+    },
   },
 } as Meta<typeof Popover>;
 
-type PopoverStoryProps = PopoverProps & { buttonText: string };
+type PopoverStoryProps = PopoverProps & { buttonText: string } & {
+  refButtonPosition: string;
+};
```
spark33, line 88: I think we can just use one type here instead of two?
```
type PopoverStoryProps = PopoverProps & { 
  buttonText: string
  refButtonPosition: string;
};
```

spark33: Not sure if this works, but it would be nice to indicate using TSDocs that these props are only for Storybook. Storybook doesn't differentiate these props from the component's props, so it might be confusing. I've been planning to do this myself on other components.

packages/popover/src/Popover.story.tsx
```diff
@@ -40,15 +76,24 @@ export default {
     children: {
       control: false,
     },
+    refButtonPosition: {
+      options: ['centered', 'top', 'right', 'bottom', 'left'],
+      control: { type: 'select' },
+    },
```
spark33, line 82: I just noticed that this arg is only being used in the scrollableContainer story, so it would be better for this arg to be defined just on that story, not here on the story level.

packages/popover/src/positionUtils.ts
```diff
@@ -139,19 +141,12 @@ const defaultElementPosition = {
...
-  // Using the boundingWidth (exact number) is a better indicator of determining if an element will fit in the window, e.g. calcLeft() uses the width to determine the left position and then that number is used to check if it is safetly within the window bounds. In some cases the offsetWidth is rounded up which means that it might not fit in the window bounds when it really can.
-  // With this we check if the difference between the boundingWidth and offsetWidth is less than one, if thats the case then the number was rounded to a whole number and we should use the exact number instead. However if the number is greater than one then that means that the boundingWidth is returning a width that has a transformation applied to it and we don't want that number. We want the untransformed width.
-  const wasRounded = Math.abs(boundingWidth - offsetWidth) < 1;
-
-  return wasRounded ? boundingWidth : offsetWidth;
+const getComputedStyleWidth = (element: HTMLElement): number => {
+  return parseFloat(getComputedStyle(element).width);
 };
 
 export function getElementDocumentPosition(
```
spark33, line 152: An optional comment, but it looks like `getElementDocumentPosition` and `getElementViewportPosition` share a lot of common code. It'd be nice if this were DRY

packages/popover/src/Popover.story.tsx
spark33: If you're talking about the TS error, I think this one is safe to ignore with a `@ts-expect-error`

spark33 (changes requested):
Some comments in Storybook!

packages/popover/src/Popover.story.tsx
shaneeza: This should be used for both stories, i'll updated the other story

packages/popover/src/Popover.story.tsx
shaneeza: i think i figured it out

packages/popover/src/positionUtils.ts
```diff
@@ -139,19 +141,23 @@ const defaultElementPosition = {
...
-  // With this we check if the difference between the boundingWidth and offsetWidth is less than one, if thats the case then the number was rounded to a whole number and we should use the exact number instead. However if the number is greater than one then that means that the boundingWidth is returning a width that has a transformation applied to it and we don't want that number. We want the untransformed width.
-  const wasRounded = Math.abs(boundingWidth - offsetWidth) < 1;
-
-  return wasRounded ? boundingWidth : offsetWidth;
+const getElementPosition = (element: HTMLElement) => {
+  const { top, bottom, left, right } = element.getBoundingClientRect();
+  const { offsetHeight: height } = element;
+  // Returns the unrounded, floating point width of the element which does not include transformations.
+  // `offsetWidth` would not work because this returns a rounded number of the element's layout width and `getBoundingClientRect.width` would also not work because it returns an exact number of the rendered width which can include transformations.
+  const width = parseFloat(getComputedStyle(element).width);
```
TheSonOfThomp, line 151: Hmm, I didn't think `getBoundingClientRect` would be affected by transformations

TheSonOfThomp (approved):
LGTM 👍 Seems to fix the problem!

packages/popover/src/positionUtils.ts
shaneeza: Yeah, theres no mention of it in the MDN docs but there are a few issues on SO that mention this

spark33 (approved):
👌👌

PR #1289, opened 2022-05-27
Updates to storybook files, propTypes

spark33:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/card/src/Card.story.tsx
```diff
@@ -9,6 +9,11 @@ import LGCard, { CardProps } from './Card';
 // https://github.com/storybookjs/storybook/issues/15334
 // eslint-disable-next-line react/jsx-props-no-spreading
 type CardStoryProps = BoxProps<'div', CardProps>;
+
+// TODO: Import below comment directly from component definition.
```
shaneeza, line 13: What does this mean?

packages/radio-group/src/RadioGroup.story.tsx
```diff
@@ -1,43 +1,23 @@
 import React, { useState } from 'react';
 import { Meta } from '@storybook/react';
-import { css, cx } from '@leafygreen-ui/emotion';
-import { palette } from '@leafygreen-ui/palette';
 import { Radio, RadioGroup } from '.';
 import { RadioGroupProps } from './RadioGroup';
+import defaultArgTypes from '../../../stories/defaultArgTypes';
```
shaneeza, line 5: just a thought, is there a way to create an alias for this? its a long path

packages/palette/src/palette.story.tsx
```diff
@@ -54,6 +54,13 @@ function ColorBlock({ color, name }: ColorBlockProps) {
 
 export default {
   title: 'Packages/Palette',
+  argTypes: {
```
TheSonOfThomp, line 57: Is it possible to do something like this on a global level? I feel like we'll want to remove `className` and other props from most other components

TheSonOfThomp (approved):
👨‍🎤

packages/card/src/Card.story.tsx
TheSonOfThomp: My understanding is that the component description on L15 below should be defined in `Card.tsx`, but can't currently for TS trickery reasons?

spark33: There is currently a copy of the component in the `.story.tsx` file. This was defined as a copy because Storybook isn't reading the component's props correctly for some reason (i.e. it's a known issue that Storybook doesn't correctly process props when components use `forwardRef`). 

Because the component was redefined, it also tries to read the TSDoc from the component's copy instead of the original component. So the TSDoc itself also had to be copied over.

Once the issues around `forwardRef` are resolved, this component copy and its TSDoc should be removed so there aren't 2 copies of the same TSDoc.

packages/palette/src/palette.story.tsx
spark33: It is possible - I actually set it to `{ description: "....", control: { type: 'text' }` globally though because I thought we might want it in the future? What I did here in `palette` completely removes `className` from the table, which I don't know if we want globally since that suggests that the className prop doesn't exist on those components.

We can also define it as  `{ description: "....", control: { type: 'none' }`. This would render the row in the prop table along with its description, but without the control. Maybe this is what we want instead. I can make this change some time in the future.

PR #1293, opened 2022-06-02
PD-2125 Refactor combobox css (#1)

TheSonOfThomp:
## ✍️ Proposed changes

This PR removes CSS vars and used a more Emotion idiomatic styling code style

🎟 _Jira ticket:_ [PD-2125](https://jira.mongodb.org/browse/PD-2125)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/combobox/src/ComboboxOption.tsx
```diff
@@ -27,48 +29,71 @@ const comboboxOptionStyle = css`
...
   &:before {
     content: '';
     position: absolute;
     left: 0;
     width: 3px;
-    height: var(--lg-combobox-item-wedge-height);
+    height: 22px;
     background-color: transparent;
     border-radius: 0 2px 2px 0;
     transform: scaleY(0.3);
```
DesignerDave, line 42: Like with transform, I believe the best practice here would be to use `scale3d` instead of `scaleY` to take advantage of hardware accelerated transitions.

packages/combobox/src/Menu.styles.ts
```diff
@@ -0,0 +1,80 @@
...
+`;
+
+export const menuBaseStyle = css`
+  position: relative;
+  width: 100%;
+  margin: 0;
+  padding: 0;
+  font-family: ${fontFamilies.default};
+  border-radius: inherit;
+  overflow: auto;
```
DesignerDave, line 27: Should this be `overflow: auto` for both X and Y?

packages/combobox/src/Menu.styles.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+`;
+
+export const menuBaseStyle = css`
+  position: relative;
+  width: 100%;
+  margin: 0;
+  padding: 0;
+  font-family: ${fontFamilies.default};
+  border-radius: inherit;
```
DesignerDave, line 26: Inherit rules are more brittle than explicit definitions. Is there a reason we can't use a definitive value here? We don't want this to change depending on context.

packages/combobox/src/constants.ts
```diff
@@ -0,0 +1,42 @@
...
+export const maxCharWidth: Record<ComboboxSize, number> = {
+  [ComboboxSize.Default]: typeScales.body1.fontSize,
+  [ComboboxSize.Large]: typeScales.body2.fontSize,
+};
+
+/**
+ * Size of combobox x & y padding (in px)
+ */
+export const comboboxPadding: Record<ComboboxSize, { x: number; y: number }> = {
+  [ComboboxSize.Default]: { x: 7, y: 5 },
```
DesignerDave, line 16: Where is this being used? I don't love the separation of the values here from the styles themselves.

DesignerDave (commented):
Took a first pass and left some feedback 😄 I don't think I'll be able to re-review this in a timely fashion, so leaving this review as non-blocking!

packages/combobox/src/Chip.tsx
```diff
@@ -7,79 +7,49 @@ import { uiColors } from '@leafygreen-ui/palette';
...
-      /* &:focus, */
-      &:focus-within {
-        background-color: var(--lg-combobox-chip-focus-color);
-      }
-    `,
-  );
+const chipWrapperSizeStyle: Record<ComboboxSize, string> = {
+  [ComboboxSize.Default]: css`
+    height: 24px;
+    font-size: ${typeScales.body1.fontSize + 1}px; // TODO: update this;
```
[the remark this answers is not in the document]
TheSonOfThomp: I agree with you in general, but I'll be getting rid of the `+ 1` in the redesign, so I think it's easier to leave this for now

packages/combobox/src/Combobox.tsx
```diff
@@ -1249,25 +1277,29 @@ export default function Combobox<M extends boolean>({
           aria-controls={menuId}
           aria-owns={menuId}
           tabIndex={-1}
```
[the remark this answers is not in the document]
TheSonOfThomp: 👍 Makes sense

packages/combobox/src/Menu.styles.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Menu styles
+ */
+
+export const popoverStyle = (width = 384) => css`
+  border-radius: 4px;
+  width: ${width}px;
+
+  & > * {
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah, this is pretty gross. I'll fix that

packages/combobox/src/Menu.styles.ts
TheSonOfThomp: 👍 Probably just `overflow-y`, yeah

packages/combobox/src/Menu.styles.ts
DesignerDave: Usually, that's handled through hiding overflow. Can we not do that here?

packages/combobox/src/Menu.styles.ts
```diff
@@ -0,0 +1,80 @@
...
+  padding: 0;
+`;
+
+export const menuMessageBaseStyle = css`
+  font-family: inherit;
+  display: inline-flex;
+  align-items: center;
+  gap: 8px;
+  & > svg {
+    width: 1em;
```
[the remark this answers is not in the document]
TheSonOfThomp: `font-size` for this element is being defined in `menuMessageSizeStyle`. We want the icon to be either `13px` or `16px` depending on the Size variant

packages/combobox/src/constants.ts
TheSonOfThomp: I don't love it either, but we need to use these same values in ~7 different locations in `Combobox.styles.ts`

packages/combobox/src/Combobox.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Typescript disagrees though
Screen Shot 2022-06-07 at 3 21 54 PM

packages/combobox/src/Menu.styles.ts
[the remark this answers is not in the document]
DesignerDave: Seems a bit brittle to use em units this way when we don't control the context outside of the component. Though it's a bit more work, I feel we should probably be applying the size CSS specifically based on the variant, and probably also applying it through a className rather than a child selector.

packages/combobox/src/constants.ts
DesignerDave: Why don't those values live in `Combobox.styles.ts`?

packages/combobox/src/Combobox.tsx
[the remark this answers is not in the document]
DesignerDave: Strange since HTML always expects a string value, but okay 😅

packages/combobox/src/Menu.styles.ts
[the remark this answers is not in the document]
TheSonOfThomp: Best we can do here is `& > div` since we need to select the container div that `Popover` injects. Can't get around it with `overflow:hidden` since that'll get rid of the shadow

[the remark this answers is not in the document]
DesignerDave: Shouldn't we apply the style through the className prop on Popover then?

packages/combobox/src/Combobox.styles.ts
```diff
@@ -1,207 +1,165 @@
...
-        --lg-combobox-background-color: ${uiColors.gray.light3};
-        --lg-combobox-background-color-focus: ${uiColors.white};
-        --lg-combobox-background-color-disabled: ${uiColors.gray.light2};
+/**
+ * Size of combobox x & y padding (in px)
+ */
+export const comboboxPadding: Record<Size, { x: number; y: number }> = {
+  [Size.Default]: {
+    y: (36 - inputHeight[Size.Default] - 2) / 2,
+    x: 8 - 1,
```
shaneeza, line 41: Curious, where are `8` and `1` coming from and why can't you use `7`?

packages/combobox/src/Combobox.types.ts
```diff
@@ -1,6 +1,13 @@
 import { ReactElement, ReactNode } from 'react';
 import { Either } from '@leafygreen-ui/lib';
 
+export const Theme = {
```
shaneeza, line 4: 👍🏽

shaneeza (approved):
Looks like you got through most of Dave's feedback, looks good! I left one non-blocking comment

packages/combobox/src/Combobox.styles.ts
TheSonOfThomp: I initially had `spacing[2]`, but decided to hard code it. Adding the `-1` to make it clear that it's the border. I'll change it back to use `spacing`

PR #1310, opened 2022-06-21
PD-2142: Group 2 

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-2142](https://jira.mongodb.org/browse/PD-2142)

[Card Component](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-card--basic&args=darkMode:true)
[Copyable Component](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-copyable--basic)
[Tabs Component](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-tabs--controlled-by-storybook)
[Inline Definition Component](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-inlinedefinition--demo)
[Select](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-select--uncontrolled&args=darkMode:true)
[Menu](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-menu--uncontrolled-template)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


shaneeza:
Update to include Sean's Global darkMode

shaneeza (commented):
~Update to include Sean's Global darkMode`

alvenw:
For Tabs, in Tab 2 (overflow tab), it seems like the text shifts a bit between the selected and unselected states which is something we don't want.

shaneeza:
@alvenw @sooachung [Select](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-select--uncontrolled&args=darkMode:true) is ready for review!

sooachung:
- Size does not change accordingly. The width gets changed by a few pixels for each of them.
- Font size also does not change.

alvenw:
- [ ] Is it intentional that the dropdown option with a long string will cause the input box to expand horizontally?
image
image

shaneeza:
> * [ ]  Is it intentional that the dropdown option with a long string will cause the input box to expand horizontally?
>   image
>   image

@alvenw  I don't see a use case for this in Figma but in the code there is a `max-width` of 400px so it seems that having the select grow is intentional.

shaneeza:
> * [ ]  Size does not change accordingly. The width gets changed by a few pixels for each of them.
> * [ ]  Font size also does not change.

@sooachung I think this might be because your screen size might be small and there is currently some old logic that changes the styles on a smaller screen. If thats the case can you try again on a larger screen? Maybe try without the inspector open if you previously had it open. If thats not the case lets chat over slack because i'm not sure i can reproduce.

shaneeza:
@alvenw @sooachung [Menu](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-menu--uncontrolled-template) is ready for review!

audroue:
Added some further notes (feel free to ignore if these any of these are duplicative with ones above):

**Card**
- we should probably remove the clickable prop (CC. @DesignerDave )
image



**Copyable**
- large prop should have larger (16px) label and description
image

- overflow shadow is always present (should only be visible when content overflows)
image



**Tabs**
- maybe out of scope: can we add overflow ellipsis to tab name that's too long?
image



**Inline definition**
- is the dotted line always the same color as the font? Seems like title dotted line is darker than the green dark 2 (the font color)
image

- positioning of tooltip/arrow seems a bit off center (to the side). I'd consider centering it to the element it's hovering over if possible. 
image



**Select**
- disabled state should have disabled cursor on hover
image



**Menu**
- distance between menu trigger and dropdown seems too large. I'd switch it to 6px
image

- mouse cursor on disabled menu item should be show as disabled 
image

- dark mode menu trigger should be dark mode button not a light mode button
image

DesignerDave:
@audroue Let's hold out on the clickable prop. We need to communicate out that change to relevant teams, and provide an alternative.

sooachung:
> > * [ ]   Size does not change accordingly. The width gets changed by a few pixels for each of them.
> > * [ ]   Font size also does not change.
> 
> @sooachung I think this might be because your screen viewport size might be small and there is currently some old logic that changes the styles on a smaller screen. If thats the case can you try again on a larger viewport? Maybe try without the inspector open if you previously had it open. If thats not the case let's chat over slack because i'm not sure i can reproduce.

It wasn't working for me earlier for some reason but it works now! Thanks:)

shaneeza:
> **Tabs**
> 
> * maybe out of scope: can we add overflow ellipsis to tab name that's too long?
> 
> image

I see that we have code to add the ellipsis but it's not working. I'll try to add it back.

@audroue

shaneeza:
> **Copyable**
> 
> * large prop should have larger (16px) label and description
> 
> image

@audroue right now the label and description size changes based on the global font size of the app. Should the font sizes be based on the font size of the component instead? This means that it will ignore the global font size.


> 
> * overflow shadow is always present (should only be visible when content overflows)
> 
> image

@audroue This functionality has been missing for a while but for the sake of time we should address it in a separate ticket.

shaneeza:
> **Inline definition**
> 
> * is the dotted line always the same color as the font? Seems like title dotted line is darker than the green dark 2 (the font color)
> 
> image

@audroue Are you saying that the dotted line should always be the same color as the font color? I can make that change if thats the case.
> 
> * positioning of tooltip/arrow seems a bit off center (to the side). I'd consider centering it to the element it's hovering over if possible.
> 
> image

@audroue For this issue i'm going to have to do a little more digging to figure this out -- I'll create a separate ticket for this.

shaneeza:
> For Tabs, in Tab 2 (overflow tab), it seems like the text shifts a bit between the selected and unselected states which is something we don't want.

@alvenw this seems to happen when switching between a tab with an icon to a tab without an icon. We're using a little CSS hack to prevent layout shifts when items become bold but the icon is not taken into account so thats why you see that little shift. This ones a little tricky to figure out and we should address it in a separate ticket.

alvenw:
- [ ] This is light mode but it looks like the icon color for the active menu item is off here (should be base green)
image

- [ ] Completely unrelated and is a complaint on the storybook side but I remember before, the button for Dark Mode was already set to a true/false toggle. Now, there's a "Set Boolean" button that I have to click before the toggle appears 
image

packages/copyable/src/Copyable.tsx
```diff
@@ -25,29 +28,13 @@ export const Size = {
...
-      border: uiColors.black,
-    },
-  },
+const codeStyleColor: Record<Theme, string> = {
+  [Theme.Light]: css`
+    color: ${palette.black};
+  `,
+  [Theme.Dark]: css`
+    color: ${palette.gray.light2};
+  `,
```
TheSonOfThomp, line 37: Much better 👍

packages/copyable/src/Copyable.tsx
```diff
@@ -70,28 +57,62 @@ const codeStyle = css`
...
+const labelFontStyle: Record<BaseFontSize, string> = {
+  [BaseFontSize.Body1]: css`
+    font-size: ${typeScales.body1.fontSize}px;
+    line-height: ${typeScales.body1.lineHeight}px;
+  `,
+  [BaseFontSize.Body2]: css`
+    font-size: ${typeScales.body2.fontSize}px;
+    line-height: 20px; // Hardcoding because it does not match body2 lineHeight
+  `,
+};
```
TheSonOfThomp, line 86: These styles should be in the `Label` component, no?

packages/copyable/src/Copyable.tsx
```diff
@@ -160,23 +199,31 @@ interface CopyableProps {
...
 
   const { portalContainer } = usePopoverPortalContainer();
 
+  // TODO: should API change to use number values instead of 'large and default' for component size?
+  // Normalize the size so that we can use the baseFontSize from the provider if the size is not set by the component
+  const normalizedSize = size
+    ? size === Size.Default
+      ? BaseFontSize.Body1
+      : BaseFontSize.Body2
+    : size;
```
TheSonOfThomp, line 224: Looks like we're setting `normalizedSize = undefined` (or something else falsy) if `size` is falsy

packages/copyable/src/Copyable.tsx
```diff
@@ -160,23 +199,31 @@ interface CopyableProps {
...
-  const mode = darkMode ? Mode.Dark : Mode.Light;
-
+  const { theme, darkMode } = useDarkMode(darkModeProp);
   const [copied, setCopied] = useState(false);
   const [showCopyButton, setShowCopyButton] = useState(false);
   const [buttonRef, setButtonRef] = useState<HTMLButtonElement>();
 
   const { portalContainer } = usePopoverPortalContainer();
 
+  // TODO: should API change to use number values instead of 'large and default' for component size?
```
TheSonOfThomp, line 218: I think keeping the API consistent here with other components (i.e. `small | default | large`) makes sense, but I agree that we should have a utility that computes the appropriate baseFontSize based on the size prop

packages/select/src/MenuButton.tsx
```diff
@@ -81,30 +183,28 @@ const menuButtonTextStyle = css`
   max-width: 100%;
 `;
 
-const errorColor: Record<Mode, string> = {
-  [Mode.Light]: palette.red.base,
-  [Mode.Dark]: '#F97216',
+const errorColor: Record<Theme, string> = {
+  [Theme.Light]: palette.red.base,
+  [Theme.Dark]: '#FF6960', // TODO: palette.red.light1 is different in figma(#FF6960) which does not match react(#EF5752) but will be updated with this ticket https://jira.mongodb.org/browse/PD-2206
```
TheSonOfThomp, line 188: I feel like we should write `red.light1` here, and have this automatically update when palette gets updated. Otherwise we may forget to come back and add the token

packages/select/src/MenuButton.tsx
```diff
@@ -81,30 +183,28 @@ const menuButtonTextStyle = css`
...
+  [Theme.Dark]: css`
+    border-color: ${errorColor[Theme.Dark]};
 
     &:hover,
     &:active {
-      border-color: ${errorColor[Mode.Dark]}; // off palette
-      box-shadow: 0px 4px 4px rgba(87, 11, 8, 0.3),
-        0px 0px 0px 3px ${uiColors.red.light3};
+      border-color: ${errorColor[Theme.Dark]};
+      box-shadow: 0px 0px 0px 3px ${palette.yellow.dark3};
```
TheSonOfThomp, line 207: I think this can be `hoverRing.dark.red`

packages/select/src/MenuButton.tsx
```diff
@@ -81,30 +183,28 @@ const menuButtonTextStyle = css`
...
-  [Mode.Light]: css`
-    border-color: ${errorColor[Mode.Light]};
+const menuButtonErrorStyle: Record<Theme, string> = {
+  [Theme.Light]: css`
+    border-color: ${errorColor[Theme.Light]};
     background-color: ${palette.white};
 
     &:hover,
     &:active {
       box-shadow: 0 0 0 3px #f9d5c5; // Between light2 & light3
```
TheSonOfThomp, line 198: This can be `hoverRing.light.red`

packages/select/src/Select.tsx
```diff
@@ -38,8 +54,11 @@ const errorTextStyle = ({
   darkMode: boolean;
   sizeSet: SizeSet;
 }) => css`
-  font-family: ${darkMode ? fontFamilies.legacy : fontFamilies.default};
-  color: ${darkMode ? '#F97216' : palette.red.base};
+  font-family: ${fontFamilies.default};
+  color: ${darkMode
+    ? '#FF6960'
+    : palette.red
+        .base}; // TODO:palette.red.light1 is different in figma(#FF6960) which does not match react(#EF5752)
```
TheSonOfThomp, line 61: Same thing here. Unless we add it to the ticket to update these tokens, my feeling is that we should leave this as the token, and have it magically get fixed

packages/select/src/Select.tsx
```diff
@@ -482,6 +503,13 @@ export default function Select({
           darkMode={darkMode}
           disabled={disabled}
           className={cx(
+            {
+              [largeLabelStyles]: size === Size.Large,
+              [css`
+                font-size: ${baseFontSize}px;
+                line-height: 20px;
```
TheSonOfThomp, line 510: Can use `typeScales['body1' | 'body2'].lineHeight`

packages/tabs/src/TabTitle.tsx
```diff
@@ -68,59 +63,56 @@ const listTitleModeStyles: Record<Mode, ListTitleMode> = {
...
       cursor: not-allowed;
-      color: ${uiColors.gray.dark1};
+      color: ${palette.gray.dark2};
     `,
   },
 };
 
 const listTitleStyles = css`
+  font-family: ${fontFamilies.default};
+  font-size: 13px;
```
TheSonOfThomp, line 114: We should use tokens here

packages/tabs/src/TabTitle.tsx
```diff
@@ -174,16 +165,20 @@ const listTitleStyles = css`
...
 const listTitleChildrenStyles = css`
-  display: flex;
-  align-items: center;
-  gap: 4px;
+  width: 100%;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+
+  // Cannot use flexbox here to center children because it breaks text-overflow: ellipsis
```
TheSonOfThomp, line 176: 😢

TheSonOfThomp (approved):
Looks good! 🌔  Just a few minor, non-blocking comments

packages/copyable/src/Copyable.tsx
shaneeza: They are in the `Label` component but the label component font size is based off of the global font size and doesn't accept a size prop so if the size prop in the copyable component is different than the global font size the label component will not recognize that so I had to duplicate the styles in here. I could export out the `Label` styles then that way I can reuse them in here.

packages/select/src/Select.tsx
shaneeza: the line-height for a label is always `20` when it's `default` which doesn't match whats in `typeScales`.  Maybe we should consider adding a label line height inside of typeScales?

```
body1: {
    fontSize: BaseFontSize.Body1,
    lineHeight: 20,
    labelLineHeignt: 20,
  } as const,
body2: {
    fontSize: BaseFontSize.Body2,
    lineHeight: 28,
    labelLineHeignt: 20,
  } as const,

```

packages/select/src/Select.tsx
shaneeza: true! i'm going to be working on that ticket this week anyways so makes sense

packages/select/src/MenuButton.tsx
shaneeza: Hm looks like the color should be `palette.red.light1`. `hoverRing.light.red` is using `palette.red.light2`

shaneeza: nvm, you're correct!

packages/copyable/src/Copyable.tsx
shaneeza: isn't that ok? override accepts: `BaseFontSize | 14 | undefined`. So if it's undefined then the baseFontSize will be returned?

audroue (approved):
Looks good!

PR #1317, opened 2022-06-23
Removes `bold` from font-face declaration

TheSonOfThomp:
`bold` alias defaults to 700 in Chrome, Safari, FF & Edge

## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -14,7 +14,7 @@
         format('woff2'),
       url('https://d2va9gm4j17fy9.cloudfront.net/fonts/euclid-circular/EuclidCircularA-Semibold.ttf')
         format('truetype');
-    font-weight: 700, bold;
+    font-weight: 700;
     font-style: normal;
   }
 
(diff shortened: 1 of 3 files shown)


.storybook/manager-head.html
```diff
@@ -14,7 +14,7 @@
         format('woff2'),
       url('https://d2va9gm4j17fy9.cloudfront.net/fonts/euclid-circular/EuclidCircularA-Semibold.ttf')
         format('truetype');
-    font-weight: 700, bold;
+    font-weight: 700;
```
TheSonOfThomp, line 17: Removing bold since major browsers map `bold` to 700

spark33 (approved):

.storybook/manager-head.html
spark33: Why is this change necessary? I'm approving the change, but I feel like we should all know why this is being done.

Should this change also be applied to the website? Looks like we've got `700, bold` defined there too.

TheSonOfThomp: Safari and Firefox don't support this way of defining font-face rules. I considered duplicating these definitions with `700` and `bold`, but it seems we don't have to. 
Good callout on the website, I'll fix that

spark33: Makes sense! Thanks for clarifying.

PR #1333, opened 2022-07-01
PD-2066 Radiobox tooltip compatibility 

TheSonOfThomp:
## ✍️ Proposed changes

Improves RadioBox + Tooltip compatibility, resolving an issue where Tooltips would not appear when a RadioBox was used as a trigger.

🎟 _Jira ticket:_ [PD-2066](https://jira.mongodb.org/browse/PD-2066)

## 🛠 Types of changes

## ✅ Checklist

## 🧪 How to test changes

Test with the new Story "Radio Box Group > TooltipTest"


packages/tooltip/src/getTooltipTriggerHandlers.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * e.x.
+ * ```ts
+ * cosnt tooltipHandlers = getTooltipTriggerHandlers(rest)
+ * <label {...tooltipHandlers}>
+ *  <input {...rest} />
+ * </label>
+ * ```
+ */
+export function getTooltipTriggerHandlers<T = any>({
```
TheSonOfThomp, line 17: I figures we may need this logic again elsewhere, so I added this to `Tooltip` instead of having a one-off solution in Radiobox

packages/radio-box-group/src/RadioBox.tsx
```diff
@@ -233,9 +237,14 @@ export default function RadioBox({
...
   });
 
+  const tooltipTriggerHandlers = getTooltipTriggerHandlers<HTMLLabelElement>(
+    rest as HTMLProps<HTMLLabelElement>,
+  );
+
   return (
     <label
       {...radioBoxWrapper.prop}
+      {...tooltipTriggerHandlers}
```
TheSonOfThomp, line 247: The cause of the bug was that we are passing `rest` into the `<input>` element, but the tooltip trigger handlers need to be passed into the component root

packages/tooltip/src/getTooltipTriggerHandlers.ts
spark33: Why do we need this? It looks like it's just putting all the props in an object and returning it?

packages/radio-box-group/src/RadioBox.tsx
spark33: Nice - this sounds like it might've been a hard one to catch lol.

I'm not sure if I'm a fan of the design of this code here, since we're passing a `rest` to a function only for it to explicitly define exactly what parameters it's expecting. I think it would be a cleaner and simpler solution for us to just explicitly pass these into the `label`, or spread them after defining them in a separate object like `const tooltipProps = { onClick, onMouseEnter, ... }; ... <label {...tooltipProps }>`

packages/radio-box-group/src/RadioBox.tsx
```diff
@@ -233,9 +237,14 @@ export default function RadioBox({
     showFocus,
   });
 
+  const tooltipTriggerHandlers = getTooltipTriggerHandlers<HTMLLabelElement>(
+    rest as HTMLProps<HTMLLabelElement>,
+  );
+
   return (
     <label
       {...radioBoxWrapper.prop}
```
spark33, line 246: Without having looked at the rest of the code, I'm confused why this says `...radioBoxWrapper.prop`. Why is it not either `props` or a more descriptive name?

spark33 (changes requested):
A comment on the code design of this fix!

packages/radio-box-group/src/RadioBox.tsx
TheSonOfThomp: `radioBoxWrapper` is the object returned from `createDataProp`

spark33 (approved):

PR #1339, opened 2022-07-06
[PD-2086 PD-2171 PD-2118] Checkbox animation & alignment

TheSonOfThomp:
## ✍️ Proposed changes

🎟 _Jira ticket:_ 
- [Ripple effect not centered on checkbox when clicked](https://jira.mongodb.org/browse/PD-2086)
- [Checkbox and label are off center](https://jira.mongodb.org/browse/PD-2171)
- [Change default label of checkbox from semibold to regular](https://jira.mongodb.org/browse/PD-2118)

*NOTE* This PR currently breaks the alignment of the DarkMode check

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/checkbox/src/Check.tsx
```diff
@@ -150,45 +151,44 @@ export function Check({
...
+          [checkWrapperCheckedStyle]: showCheckIcon,
+          [checkWrapperDisabledStyle]: disabled,
+          [checkWrapperCheckedDisabledStyle]: disabled && showCheckIcon,
+          [disableAnimation]: !shouldAnimate,
+        })}
       >
         <Transition
           in={showCheckIcon}
-          timeout={checkAnimationDuration}
+          timeout={prefersReducedMotion ? 0 : checkAnimationDuration}
```
shaneeza, line 173: Can you explain what this is doing? I set it to 0 and I didn't see a difference.

TheSonOfThomp: If the user has the `prefers-reduced-motion` preference set, then the transition timeout will be 0. This needs to be the same value as the `transition-duration` property in css

shaneeza: I see, however i don't visually see any difference in the animation if `shouldAnimate` is `false` and the `timeout` is `100 or 0` or if `shouldAnimate` is `true` and `timeout` is `100 or 0`.  I can see that when it's `false` it doesn't animate and if it `true` it does. I don't think this is blocking, i'm just curious because I can't see a difference.

shaneeza (approved):

packages/checkbox/src/Check.tsx
TheSonOfThomp: Ah I see. I guess this could just stay as it was before and it'd behave the same. I think it's more "technically correct" as written now even if not entirely necessary

PR #1345, opened 2022-07-18
Blob Generator

TheSonOfThomp:
Demo is here: https://1hqidr.csb.app/


TheSonOfThomp:
Status update: Generates shapes constructed from only small circles (i.e. no 2x2 circles yet)

Screen Shot 2022-07-19 at 4 34 20 PM

TheSonOfThomp:
Update: Generates large (2x2) circles now

Screen Shot 2022-07-20 at 6 56 21 PM

TheSonOfThomp:
Interactive generator story

Screen Shot 2022-07-22 at 4 51 50 PM

packages/brand-shape/src/Blob.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  return (
+    <svg
+      viewBox="0 0 8 8"
+      width="100%"
+      height="100%"
+      className={className}
+      {...rest}
+    >
+      {/* DEBUG */}
```
brookescarlett, line 27: 😂

brookescarlett:
Keep the branch but maybe we close the PR? Unless we're actually looking for feeedback / to merge this in

packages/brand-shape/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+## Example
+
+**Output HTML**
+
+## Properties
+
+| Prop | Type | Description | Default |
+| ---- | ---- | ----------- | ------- |
```
spark33, line 28: Need prop table, and some documentation on how the `o` vs. `O` in the `shape` prop should be expected to behave. 

There are also a few shapes that feel like they should work, but maybe they're not possible due to technical constraints - like:
```
<Blob
shape={[
["o","o","o","o"],
["o"," "," ","o"],
["o"," "," ","o"],
["o","o","o","o"]
]}
/>
```

packages/brand-shape/src/Blob.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+          ['o', ' ', ' ', 'o'],
+        ]}
+      />,
+    );
+
+    const blob = getByTestId(container, 'blob');
+    expect(blob).toBeInTheDocument();
+  });
+
+  // TODO: add more tests
```
spark33, line 24: probably would want to lean on Chromatic more than functional tests - just wouldn't make sense for a package like this

packages/brand-shape/src/Blob.tsx
spark33: If I were to review this component for production purposes, I would propose that we include this logic in a `BlobDebugger` wrapper component that functions similarly to the V11Table's V11Adapter component where these svg elements can be injected into the regular blob component.

packages/brand-shape/src/generateBlobPath.ts
```diff
@@ -0,0 +1,290 @@
...
+          y,
+          b1x: x,
+          b1y: y - bezY,
+          b2x: xNext === x ? x : xNext > x ? xNext - bezX : xNext + bezX, // next v is ? right : left: ;
+          b2y: yNext,
+        };
+      }
+    }
+  });
+}
```
spark33, line 290: These functions should be tested

packages/brand-shape/src/types.ts
```diff
@@ -0,0 +1,136 @@
...
+export const PrevDirection: Record<Direction, Direction> = {
+  top: 'topLeft',
+  topRight: 'top',
+  right: 'topRight',
+  bottomRight: 'right',
+  bottom: 'bottomRight',
+  bottomLeft: 'bottom',
+  left: 'bottomLeft',
+  topLeft: 'left',
+} as const;
```
spark33, line 134: These could probably just be calculated from an array and indexed by `(x,y)` coords. e.g. topLeft would be `-1, -1`.

packages/brand-shape/src/utils.ts
```diff
@@ -0,0 +1,368 @@
...
+    };
+  }
+
+  return {
+    top: [x, y - 1],
+    right: [x + 1, y],
+    bottom: [x, y + 1],
+    left: [x - 1, y],
+  };
+};
```
spark33, line 316: Using `(x,y)` values for the directions would probably help make these functions shorter too.

spark33 (changes requested):
There are few minor comments on code conciseness, but those aren't important.

What probably matters more is:
1) getting some documentation and expected usage guidelines on the README, 
2) testing the functional side of the utilities and `shape` prop, 
3) fixing bugs with the SVG generation,
4) moving the `debug` functionality to an internal component wrapper,
5) and getting clearance from any branding team that would be interested in this.

I think aside from these considerations, the code mostly looks good to me and pretty close to being released. I'm thinking I could put this on the Figma plugin to help designers generate blob shapes quickly.

PR #1347, opened 2022-07-21
Add missing as prop docs to typography

DesignerDave:
## ✍️ Proposed changes

Adds missing prop documentation for `as` in a number of typography components.


TheSonOfThomp (approved):
👍

PR #1348, opened 2022-07-26
Fix props table in Tooltip README

DesignerDave:
## ✍️ Proposed changes

Quick fix to the Tooltip props table in the README.


TheSonOfThomp (approved):
👍

PR #1352, opened 2022-08-01
PD-2168, PD-2149: Refresh Menu

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-2168](https://jira.mongodb.org/browse/PD-2168), [PD-2194](https://jira.mongodb.org/browse/PD-2194)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)

### DIFF
--- a/packages/menu/src/Menu.tsx
+++ b/packages/menu/src/Menu.tsx
@@ -180,6 +180,7 @@ function Menu({
             onExited: () => {
               setClosed(curr => !curr);
             },
+            darkMode
           });
         }
 
--- a/packages/menu/src/Menu.tsx
+++ b/packages/menu/src/Menu.tsx
@@ -10,14 +10,24 @@ import { MenuItemElement } from './MenuItem';
 import { SubMenuElement } from './SubMenu';
 import { MenuProps } from './types';
 import isUndefined from 'lodash/isUndefined';
+import { Mode } from '@leafygreen-ui/tokens';
 
 const rootMenuStyle = css`
   width: 200px;
   border-radius: 12px;
   overflow: auto;
   padding: 14px 0;
-  background-color: ${palette.black};
+  
 `;
+
+const rootMenuThemeStyles: Record<Mode, string> = {
+  [Mode.Light]: css`
+  background-color: ${palette.black};
+  `,
+  [Mode.Dark]: css`
+  background-color: ${palette.gray.light2};
+  `
+};
 
 const scrollContainerStyle = css`
   overflow: auto;
@@ -44,6 +54,7 @@ const scrollContainerStyle = css`
  * @param props.refEl Reference element that Menu should be positioned against.
  * @param props.usePortal Boolean to describe if content should be portaled to end of DOM, or appear in DOM tree.
  * @param props.trigger Trigger element can be ReactNode or function, and, if present, internally manages active state of Menu.
+ * @param props.darkMode Determines whether or not the component will be rendered in dark mode.
  */
 function Menu({
   align = Align.Bottom,
@@ -63,8 +74,13 @@ function Menu({
   scrollContainer,
   popoverZIndex,
   maxHeight = 256,
+  darkMode = false,
   ...rest
 }: MenuProps) {
+
+  // TODO: dark mode context
+  const theme = darkMode ? Mode.Dark : Mode.Light;
+
   const hasSetInitialFocus = useRef(false);
   const hasSetInitialOpen = useRef(false);
 
@@ -325,6 +341,7 @@ function Menu({
       <div
         className={cx(
           rootMenuStyle,
+          rootMenuThemeStyles[theme],
           css`
             max-height: ${maxMenuHeightValue};
           `,
@@ -397,6 +414,7 @@ Menu.propTypes = {
   trigger: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
   open: PropTypes.bool,
   setOpen: PropTypes.func,
+  darkMode: PropTypes.bool,
 };
 
 export default Menu;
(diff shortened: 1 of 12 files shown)


TheSonOfThomp:
Are all sub-items supposed to show the active wedge? This is in Light-mode as well.

Screen Shot 2022-08-02 at 12 02 59 PM

packages/menu/src/Menu.tsx
```diff
@@ -98,7 +112,7 @@ function Menu({
 
     function updateChildren(children: React.ReactNode): React.ReactNode {
       return React.Children.map(children, child => {
-        if (!React.isValidElement(child) || child.props?.disabled) {
+        if (!React.isValidElement(child)) {
```
TheSonOfThomp, line 115: Should we highlight this change as well?

shaneeza:
@TheSonOfThomp i think you might be right, now that i think about it a submenu item would never be active? i'll double check with @audroue

packages/menu/src/MenuItem.tsx
```diff
@@ -63,6 +65,13 @@ interface BaseMenuItemProps {
   children?: React.ReactNode;
 
   href?: string;
+
+  /**
+   * Determines whether or not the component will be rendered in dark mode.
+   *
+   * default: `false`
+   */
+  darkMode?: boolean;
```
TheSonOfThomp, line 74: I think we should use an internal `MenuContext` for this, or use an `InternalMenuItem` component like we do with `Select`. It feels confusing to have a consumer-facing prop like this that doesn't do anything

packages/menu/src/SubMenu.tsx
```diff
@@ -190,6 +315,18 @@ interface SubMenuProps {
   onExited?: ExitHandler<HTMLElement>;
 
   href?: string;
+
+  /**
+   * Determines whether or not the component will be rendered in dark Theme.
+   *
+   * default: `false`
+   */
+  darkMode?: boolean;
```
TheSonOfThomp, line 324: Same concern here. Unless the consumer can toggle DarkMode for individual SubMenus/MenuItems, I don't think we should have this as a prop

packages/menu/src/styles.ts
```diff
@@ -1,12 +1,13 @@
...
 export const svgWidth = 24;
-export const menuItemPadding = 15;
-export const paddingLeft = 52;
-const indentation = 16;
+export const paddingLeftWithGlyph = 54;
+export const paddingLeftWithoutGlyph = 20;
+const indentation = 20;
 const wedgeWidth = 4;
 const minMenuContent = 32;
 const menuBlockPadding: Record<Size, number> = {
```
TheSonOfThomp, line 13: These seem to be unused

TheSonOfThomp (changes requested):
Styles look good. I have a concern about the external API though

packages/menu/src/MenuItem.tsx
shaneeza: I see, thats similar to what the select component. Yeah that makes more sense.

packages/menu/src/Menu.tsx
shaneeza: i'll add it

packages/menu/src/MenuItem.tsx
spark33: We also have an extendable `DarkModeProps` interface now! For all places in this PR that define `darkMode` in interfaces

packages/menu/src/MenuItem.tsx
```diff
@@ -86,6 +96,9 @@ const MenuItem: ExtendableBox<
     const hoverStyles = getHoverStyles(menuItemContainer.selector);
     const focusStyles = getFocusedStyles(menuItemContainer.selector);
 
+    // Darkmode is passed from the `Menu` component which is using the `useDarkMode` hook
+    const theme = darkMode ? Theme.Dark : Theme.Light;
```
spark33, line 100: If we move to using a `MenuContext`, the `useDarkMode` hook also supports:

`const { darkMode, theme } = useDarkMode(darkModeProp);` so all children components could just consume this from `MenuContext`

spark33 (changes requested):
Minor comments around the usage of the recent darkMode utilities

shaneeza:
**note to self: update website**

packages/menu/src/Menu.tsx
shaneeza: Actually i'm going to add this back. If it's disabled it doesn't need to be cloned or have any additional props added to it.

shaneeza:
> Are all sub-items supposed to show the active wedge? This is in Light-mode as well.

@TheSonOfThomp i checked with Rob, submenu items can be active and should show the little wedge. I will clean up the example a bit so there is are not three active state.

TheSonOfThomp (approved):
LGTM! 🎸

spark33 (approved):
noiiice

packages/menu/src/MenuContext.tsx
```diff
@@ -0,0 +1,12 @@
+import { createContext } from 'react';
+import { Theme } from '@leafygreen-ui/lib';
+
+interface MenuData {
+  theme: Theme;
+}
+
+export const MenuContext = createContext<MenuData>({
+  theme: Theme.Light,
+});
```
TheSonOfThomp, line 10: There are probably other props we're passing in that could just be in Context, but that doesn't need to be done right now

shaneeza: Yeah i was thinking about the `size` prop while I was adding this.

PR #1356, opened 2022-08-03
fix to add back the ellipsis

shaneeza:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)

### DIFF
--- a/packages/tabs/src/TabTitle.tsx
+++ b/packages/tabs/src/TabTitle.tsx
@@ -166,10 +166,6 @@ const listTitleStyles = css`
       transform: scaleX(1);
     }
   }
-
-  > * {
-    vertical-align: middle;
-  }
 `;
 
 const listTitleChildrenStyles = css`
@@ -178,8 +174,10 @@ const listTitleChildrenStyles = css`
   overflow: hidden;
   text-overflow: ellipsis;
 
+  // Cannot use flexbox here to center children because it breaks text-overflow: ellipsis
+
   > svg {
-    vertical-align: bottom;
+    vertical-align: text-bottom;
     margin-right: 4px;
   }
 `;
--- a/packages/tabs/src/TabTitle.tsx
+++ b/packages/tabs/src/TabTitle.tsx
@@ -125,9 +125,6 @@ const listTitleStyles = css`
   border: 0px;
   margin: 0;
   text-decoration: none;
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


shaneeza:
Screen Shot 2022-08-03 at 4 20 02 PM

packages/tabs/src/TabTitle.tsx
```diff
@@ -176,9 +173,15 @@ const listTitleStyles = css`
...
-  display: flex;
-  align-items: center;
-  gap: 4px;
+  width: 100%;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+
+  > svg {
+    vertical-align: bottom;
```
TheSonOfThomp, line 182: Technically `vertical-align: text-bottom;` would be more correct, but this works as well

packages/tabs/src/TabTitle.tsx
```diff
@@ -176,9 +173,15 @@ const listTitleStyles = css`
 `;
 
 const listTitleChildrenStyles = css`
-  display: flex;
-  align-items: center;
-  gap: 4px;
+  width: 100%;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
TheSonOfThomp, line 179: I don't like that we can't use flexbox here, but text overflow doesn't work the way we need to use it

packages/tabs/src/TabTitle.tsx
```diff
@@ -176,9 +173,15 @@ const listTitleStyles = css`
 `;
```
TheSonOfThomp, line 173: Unrelated to this PR, but I don't remember why that `vertical-align: middle;` declaration isn't just inside the block below

packages/tabs/src/TabTitle.tsx
TheSonOfThomp: Might be worth adding a comment about why we're not using flex here. Someone (most likely me 🙃) is almost guaranteed to try adding it next time they're in here

TheSonOfThomp (approved):
Works as written, just noted a potentially clearer CSS property value. LGTM 🦅

packages/tabs/src/TabTitle.tsx
shaneeza: oh yeah, I'm going to remove it then. It doesn't seem like its needed.

PR #1370, opened 2022-08-10
PD-2246 Button with NextJS types

TheSonOfThomp:
## ✍️ Proposed changes

Updates typings for Button to allow for NextJs Links in the `as` prop. Additionally, abstracts Button Contents to a separate component

🎟 _Jira ticket:_ [PD-2246](https://jira.mongodb.org/browse/PD-2246)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/button/src/types.ts
```diff
@@ -34,7 +35,11 @@ export const FontSize = {
 
 export type FontSize = typeof FontSize[keyof typeof FontSize];
 
+type AsPropType = keyof JSX.IntrinsicElements | React.ExoticComponent<any>;
```
spark33, line 38: Might be useful in `lib`!

packages/button/src/Button.tsx
```diff
@@ -70,65 +53,61 @@ const Button: ExtendableBox<ButtonProps & { ref?: React.Ref<any> }, 'button'> =
...
+        <>{props.children}</>
+      );
+
+    return (
+      // If WrapperComponent is a Fragment then is ignored, and rootProps gets applied to the Box
+      <WrapperComponent {...rootProps}>
+        <Box {...boxProps}>
+          <ButtonContent {...contentProps}>{children}</ButtonContent>
+        </Box>
+      </WrapperComponent>
```
spark33, line 109: Is there a reason why you're creating a `WrapperComponent` here instead of just passing the `asProp` to the `Box`?

packages/button/src/types.ts
```diff
@@ -92,13 +97,25 @@ interface ButtonProps {
...
+  as?: AsPropType,
+): as is keyof JSX.IntrinsicElements =>
+  !isUndefined(as) && typeof as === 'string';
+
+/** Identifies whether the `as` prop is a Component */
+export const isAsComponent = (
+  as?: AsPropType,
+): as is React.ExoticComponent<any> =>
+  !isUndefined(as) && typeof as !== 'string';
+
```
spark33, line 120: These might be useful in `lib` as well

packages/button/src/Button.tsx
TheSonOfThomp: NextJS Link needs an anchor inside, so we need essentially two wrapper components.
`<Button as={Link} />` returns `<Link href><a>{children}</a></Link>`

spark33: I actually resolved two other issues involving NextJS Link components as the `as` prop, and this is the [workaround](https://jira.mongodb.org/browse/PD-1966?focusedCommentId=4717266&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4717266) that I found.

If this works on your end, I'm wondering if a more graceful solution would either be to 1) add this to our guidelines overall and push engineers to this approach if they want to use NextJS Links, or 2) bake this behavior directly into the `<Box>` so all instances of NextLinks are replaced with the workaround approach.

spark33: It's an issue that we see across many components, so finding one resolution would be nice.

TheSonOfThomp: If I understand correctly, I don't think we'd want to force external developers to create a `Linker` component when the ideal solution would be that LG components could accept the NextJS `Link` component. 
I do think this is functionality that we could add more globally though (i.e. to `Box`)

TheSonOfThomp: I took a swing at it, but I feel like the TS wizardry required to incorporate this into the `Box` component is out of scope of this ticket.

TheSonOfThomp:
@spark33 Chatted with Dave earlier today. We'll do the Box work as a separate ticket https://jira.mongodb.org/browse/PD-2379

TheSonOfThomp:
Closing this PR in favor of the simpler approach in #1395

PR #1384, opened 2022-08-18
PD-2336: Refresh form footer

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [PD-2336](https://jira.mongodb.org/browse/PD-2336)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/form-footer/src/FormFooter.tsx
```diff
@@ -111,23 +131,32 @@ export default function FormFooter({
...
+            darkMode={darkMode}
           >
             {backButtonText}
           </Button>
         )}
         <div className={flexEndContent}>
           {errorMessage && (
-            <Banner className={bannerStyle} variant="danger">
+            <Banner
+              // TODO: REMOVE BEFORE MERGING GROUP7 -- the lastest banner darkMode changes will me merged in by then
```
spark33, line 154: I'm going to leave a blocking review just so we make sure this is resolved!

packages/form-footer/src/FormFooter.tsx
```diff
@@ -136,14 +165,20 @@ export default function FormFooter({
...
+              darkMode={darkMode}
             >
               {cancelButtonText || 'Cancel'}
             </Button>
           )}
           {isComponentType(primaryButton, 'Button') ? (
-            primaryButton
+            React.cloneElement(primaryButton, {
+              darkMode: darkMode,
+            })
```
spark33, line 176: Could we avoid using `cloneElement`? I think the best way forward here is to:
1. Rename the `primaryButton` prop to `PrimaryButton` to indicate that it accepts component types.
2. Replace this `cloneElement` call with a `<PrimaryButton>` call. If the `PrimaryButton` prop is not a component type, we could locally rename it to `primaryButtonProps`, and spread it. This would also help us avoid this messy ternary operator.

spark33 (changes requested):

packages/form-footer/src/FormFooter.tsx
shaneeza: @spark33  Do you mind elaborating on  a few things?

- We can switch to `PrimaryButton`  but there is already a component called `PrimaryButton` that we are importing and using if only props are passed in.  So if we rename to `PrimaryButton` then the imported component would need an alias. 

- I’m not sure what you mean by avoiding the ternary. If i'm not mistaken, we need a check to see if the `primaryButton` passed in a component or props. If it's a component then we can use that component, if not then we would use the imported `PrimaryButton` and pass it those props.

spark33: Sorry - reading this back, I should've said "ternary operator in the render function". We could get rid of the `cloneElement` call in a few different ways, but maybe something like these would be a quick change?
```
const primaryButtonProps = PrimaryButton.props ?? PrimaryButton;

return (
   ...
   <Button {...primaryButtonProps} darkMode={darkMode} />
);
```
```
const primaryButtonProps = isComponentType(PrimaryButton, 'Button') ? {...primaryButton.props, darkMode } : PrimaryButton;

return (
   ...
   <Button {...primaryButtonProps} darkMode/>
);
```

It's a pretty minor refactor here, but I thought I'd bring it up since a major bump would be a good time to explore changes to prop names.

I think the design for this prop is generally kind of weird - even if the `primaryButton` is a component, it can only ever be a Button, which the FormFooter component defaults to. It feels like it should just be `primaryButtonProps`.

spark33: I'm gonna approve the PR anyway if you'd rather prioritize getting this merged in.

spark33 (approved):

packages/form-footer/src/FormFooter.tsx
shaneeza: I'm going to merge this in for now. If we're changing prop names we should address in a separate PR since this PR is for dark mode changes. 

I will bring this up during team code review tho!

PR #1446, opened 2022-10-05
PD-1867:  Pipeline brand refresh

shaneeza:
## ✍️ Proposed changes

[Storybook Link](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-pipeline--basic)

🎟 _Jira ticket:_ [PD-1867](https://jira.mongodb.org/browse/PD-1867)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/pipeline/src/types.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * The DOM node to use as the root node for the intersectionObserver. Defaults to window when null or undefined.
+   **/
+  intersectionNode?: HTMLElement | null;
+
+  /**
+   * Alter the rendered size of the component. Inherited from the parent Pipeline component.
+   */
+  size?: Size;
```
shaneeza, line 61: Was considering moving `size` and `intersectionNode` into the `PipelineContext` but the description of `intersectionNode` in the README -- "The DOM element to observe intersections with. When used with the `Pipeline` component, this prop is automatically set to the element rendered by the `Pipeline`" -- makes it seem like you can use `Stage` outside of `Pipeline`  which I don't think is great. Seeing as this is a `Major` change maybe I should remove this from `Stage` and add to  `PipelineContext` because we shouldn't have consumers be able to pass a `size` to both `<Pipeline>` and `<Stage>` and override the default intersectionNode by passing in a new `intersectionNode`

packages/pipeline/src/Counter.tsx
```diff
@@ -101,29 +34,38 @@ const getStatefulStyles = (state: StateForStyles) => ({
  */
 const Counter = forwardRef(
```
spark33, line 35: I think @TheSonOfThomp would appreciate an `@internal` flag on this component while we're at it

packages/pipeline/src/Counter.tsx
```diff
@@ -101,29 +34,38 @@ const getStatefulStyles = (state: StateForStyles) => ({
...
+        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+        tabIndex={0}
       >
+        {/* The counter is a unique shape with focus and hover state that are not easily accomplished with css alone so we are using an SVG. The reason we have 2 SVGs is to mimic the behavior of a div expanding horizontally when the text increases since an SVG does not stretch horizontally while maintaining its height.*/}
+        <div className={counterSvgBaseStyles}>
+          <div className={counterSvgColStyles}>
+            <Icon className={cx(svgLayer1Styles)} />
+          </div>
+          <div className={counterSvgColStyles}>
+            <Icon className={cx(svgLayer2Styles)} />
```
spark33, line 64: I assume the `cx` was relevant at some point, but it looks like we can remove them now?

packages/pipeline/src/Pipeline.story.tsx
```diff
@@ -16,6 +18,9 @@ export default {
     children: {
       control: false,
     },
+    darkMode: {
+      control: 'boolean',
+    },
```
spark33, line 23: Using `darkMode: storybookArgTypes.darkMode` will also give it the default comment I believe!

packages/pipeline/src/svgs/edges.ts
```diff
@@ -0,0 +1,30 @@
...
+    after:
+      'data:image/svg+xml,%3Csvg width="13" height="36" viewBox="0 0 13 36" fill="none" xmlns="http://www.w3.org/2000/svg"%3E%3Cpath d="M0 0H0.90983C2.80369 0 4.53501 1.07001 5.38197 2.76393L11.882 15.7639C12.5858 17.1716 12.5858 18.8284 11.882 20.2361L5.38196 33.2361C4.535 34.93 2.80369 36 0.90983 36H0V0Z" fill="%23C3E7FE"/%3E%3C/svg%3E%0A',
+  },
+  [Size.Large]: {
+    before:
+      'data:image/svg+xml,%3Csvg width="13" height="48" viewBox="0 0 13 48" fill="none" xmlns="http://www.w3.org/2000/svg"%3E%3Cpath d="M0.223607 1.44721C-0.108843 0.782313 0.374652 0 1.11803 0H13V48H1.11803C0.374652 48 -0.108844 47.2177 0.223607 46.5528L10.382 26.2361C11.0858 24.8284 11.0858 23.1716 10.382 21.7639L0.223607 1.44721Z" fill="%23C3E7FE"/%3E%3C/svg%3E%0A',
+    after:
+      'data:image/svg+xml,%3Csvg width="15" height="48" viewBox="0 0 15 48" fill="none" xmlns="http://www.w3.org/2000/svg"%3E%3Cpath d="M0 0H0.851465C2.77563 0 4.52907 1.10416 5.36055 2.8394L14.4647 21.8394C15.1193 23.2054 15.1193 24.7946 14.4647 26.1606L5.36054 45.1606C4.52907 46.8958 2.77563 48 0.851465 48H0V0Z" fill="%23C3E7FE"/%3E%3C/svg%3E%0A',
+  },
+};
```
spark33, line 30: Why not save these as separate SVGs too for consistency?

spark33 (approved):
Approved with minor comments!

packages/pipeline/src/svgs/edges.ts
shaneeza: I was having trouble adding these as SVG files because they kept importing as react components so I went with this route but If i were to use SVG files I would still need to map them to the correct size and position.
Also If i were to switch over to `clip-path` instead of using a `mask` I would still need the edge object with each path so I just stuck with SVGs and data URIs

packages/pipeline/README.md
```diff
@@ -28,75 +28,27 @@ npm install @leafygreen-ui/pipeline
...
+| Prop        | Type                   | Description                                                                                                                                             | Default     |
+| ----------- | ---------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------- |
+| `className` | `string`               | Adds a className to the class attribute.                                                                                                                | `''`        |
+| `children`  | `node`                 | The content that will render inside of the component.                                                                                                   | `undefined` |
+| `threshold` | `number`               | Either a single number or an array of numbers which indicate at what percentage of the target's visibility, the observer's callback should be executed. | `0.8`       |
```
spark33, line 51: Is this supposed to be `number | number[]`?

packages/pipeline/src/TooltipText.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * ```
+ * <TooltipText />
+ * ```
+ * @internal
+ */
+const TooltipText = ({ hiddenStages }: TooltipTextProps) => {
+  const tooltipChildren = [];
+
+  if (hiddenStages.length > 0) {
+    for (let i = 0; i < hiddenStages.length; i++) {
```
spark33, line 25: Minor - Is `if (hiddenStages.length > 0)` necessary here? I assume it would just be caught by the loop guard.

packages/pipeline/src/TooltipText.tsx
```diff
@@ -0,0 +1,48 @@
...
+          <span key={`icon-${i}`}>
+            <ChevronRight size={12} fill={palette.gray.base} />
+          </span>,
+        );
+        tooltipChildren.push(
+          <span key={`${hiddenStages[i]}-${i}`}>{hiddenStages[i]}</span>,
+        );
+      }
+    }
+  }
```
spark33, line 41: Very minor - I feel like this loop could be cleaner as a `hiddenStages.map(...)`?

packages/pipeline/src/svgs/edges.ts
spark33: Might be a good comment to add to the file!

spark33 (changes requested):
Minor comments!

sooachung (approved):

spark33 (approved):

audroue (approved):
LGTM! Great job 👏

PR #1449, opened 2022-10-06
PD-2492: SearchInput v0.9.0

spark33:
## ✍️ Proposed changes

This does not implement the typeahead search yet - waiting for engineering specs and prioritizing AWS deployment of the website before it. This is a simple `<input type="text">` with the appropriate styles.

🎟 _Jira ticket:_ [PD-2492](https://jira.mongodb.org/browse/PD-2492)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/search-input/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "extends": "../../package.tsconfig.json",
+  "compilerOptions": {
+    "declarationDir": "dist",
+    "outDir": "dist",
+    "rootDir": "src"
+  },
+  "include": ["src/**/*"],
+  "references": [
+    {
+      "path": "../lib"
```
shaneeza, line 12: You're missing a few packages

shaneeza (commented):
Whats the reason for using `input` instead of using `TextInput`?

packages/search-input/tsconfig.json
spark33: Thanks for catching it! I always forget this has to be manual.

spark33:
> Whats the reason for using `input` instead of using `TextInput`?

`TextInput` comes with other functionalities like the `label` and `description` props. There will also be style changes in each component. that are independent of each other. For the styles and other shared characteristics, I'm a fan of implementing something like `BaseTextInput` or something that all text-type `<input>` components can use. Until we have something like that, I think it would be best to keep them all independent of one another.

packages/search-input/tsconfig.json
shaneeza: actually i think Adam added a scrip to add these

shaneeza: `check:deps`

shaneeza (commented):
Are we adding this to the `website`?

spark33:
> Are we adding this to the `website`?

Yes! I was thinking we would just add it to the new website after we published it since we're getting pretty close to launch on that front. This is also a v0.9 that doesn't have all the props we want (i.e. typeahead, loading state, etc), so I wasn't too pressed on getting it on the website for now.

spark33:
> Looks good, I did notice that the icon is really tiny when the size is `xsmall`, from the design it looks like its always the same size for each size variant.
> 
> Screen Shot 2022-10-11 at 4 34 09 PM

Oh you're right - I don't know if that's changed in the meanwhile, or I was tripping. I'll change that!

packages/search-input/tsconfig.json
spark33: Hm that script didn't work for me. I updated the file manually anyway!

.changeset/two-years-pretend.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/search-input': major
+'@leafygreen-ui/text-input': patch
+---
+
+Adds console warning suggestion to use SearchInput component in place of TextInput with type="search"
```
TheSonOfThomp, line 6: Maybe break these out into separate files? So we can have different messages. The patch with this comment, and a second one that releases the search-input

packages/search-input/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "dependencies": {
+    "@leafygreen-ui/emotion": "^4.0.1",
+    "@leafygreen-ui/icon": "^11.10.2",
+    "@leafygreen-ui/lib": "^9.4.2",
+    "@leafygreen-ui/palette": "^3.4.2",
+    "@leafygreen-ui/tokens": "^1.3.4",
+    "@leafygreen-ui/typography": "^13.1.2"
+  },
+  "peerDependencies": {
+    "@leafygreen-ui/leafygreen-provider": "^2.3.3"
```
TheSonOfThomp, line 39: Should bump this to `v3.0.0`

packages/search-input/src/SearchInput.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+      data-testid="search-input"
+      aria-labelledby="test-search-input"
+      {...props}
+    />,
+  );
+  const searchInput = utils.getByTestId('search-input');
+  return { ...utils, searchInput };
+}
+
+describe('packages/search-input', () => {
```
TheSonOfThomp, line 19: Should probably replicate all the `text-input` tests

packages/search-input/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  XSmall: 'xsmall',
+  Small: 'small',
+  Default: 'default',
+  Large: 'large',
+} as const;
+
+export type SizeVariant = typeof SizeVariant[keyof typeof SizeVariant];
+
+export interface SearchInputProps
+  extends HTMLElementProps<'input', HTMLInputElement> {
```
TheSonOfThomp, line 21: SearchInput should accept `aria-label` as well, and not require `aria-labelledby` in that case

packages/text-input/src/TextInput.tsx
```diff
@@ -119,10 +119,15 @@ const TextInput: TextInput = forwardRef<HTMLInputElement, TextInputProps>(
       );
     }
 
-    if (type === 'search' && !rest['aria-label']) {
-      console.error(
-        'For screen-reader accessibility, aria-label must be provided to TextInput.',
+    if (type === 'search') {
+      console.warn(
```
TheSonOfThomp, line 123: Should consider using `lib/consoleOnce`, otherwise this will warn on every render

scripts/package-builder.js
```diff
@@ -141,13 +141,20 @@ export { default } from './${PACKAGE_UC}';
...
 import ${PACKAGE_UC} from '.';
 
-storiesOf('Components/${PACKAGE_UC}', module)
-  .add('Default', () => (
-    <${PACKAGE_UC} />
-  ))
+export default {
+  title: 'Components/${PACKAGE_UC}',
+  component: ${PACKAGE_UC},
+}
```
TheSonOfThomp, line 150: 🔥

packages/search-input/src/SearchInput.spec.tsx
spark33: Some tests aren't relevant for SearchInput (i.e. label, description, baseFontSize, etc.) so there won't be full parity, but I went back and added some more!

packages/search-input/src/types.ts
spark33: Actually, I just saw your PR for the TextInput component. What was the discussion there? It would be nice to get on the same page on it before moving forward.

sooachung:
- The disabled state should not have a hover state.
Screen Shot 2022-11-08 at 11 54 12 AM

- The color of the icon in dark mode seems lighter than it should be. Maybe the box outline as well, can you double-check if the colors are right? The icon should be gray light 1 and the outline should be the gray base.
Screen Shot 2022-11-08 at 12 03 02 PM

sooachung (approved):

spark33:
@TheSonOfThomp This PR feels relevant to PR #1490, so I'll wait for your review before merging this.

packages/search-input/src/SearchInput.spec.tsx
```diff
@@ -30,6 +37,47 @@ describe('packages/search-input', () => {
...
+    ).toBe(true);
+  });
+
+  test('key presses are reflected in component and onChange function is called when value changes', () => {
+    const { searchInput } = renderSearchInput(defaultProps);
+    expect((searchInput as HTMLInputElement).value).toBe('');
+
+    fireEvent.change(searchInput, {
+      target: { value: 'a' },
+    });
```
TheSonOfThomp, line 60: I think we should prefer using `userEvent.type` to more closely simulate real user behavior

TheSonOfThomp, line 60: I think we should prefer using `userEvent.type` to more closely simulate real user behavior

TheSonOfThomp (approved):

packages/search-input/src/types.ts
brookescarlett: Yes definitely valid to accept an `aria-labelledby` prop that relates the label and input field.

spark33:
> > @TheSonOfThomp This PR feels relevant to PR #1490, so I'll wait for your review before merging this.
> 
> As long as there are no conflicts, they can go in together

Cool! Thanks. Merging 👍

PR #1464, opened 2022-10-20
LG-2504, LG-2487, LG-2557: Refresh segmented control

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2504](https://jira.mongodb.org/browse/LG-2504)
🎟 _Jira ticket:_ [LG-2487](https://jira.mongodb.org/browse/LG-2487)
🎟 _Jira ticket:_ [LG-2557](https://jira.mongodb.org/browse/LG-2557)

Screen Shot 2022-10-20 at 12 17 21 PM
Screen Shot 2022-10-25 at 3 36 56 PM

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)

### DIFF
--- a/packages/segmented-control/src/SegmentedControl.spec.tsx
+++ b/packages/segmented-control/src/SegmentedControl.spec.tsx
@@ -16,10 +16,10 @@ const renderNewContainer = () => {
       className={testClassName}
     >
       <SegmentedControlOption value="apple" data-testid="apple">
-        <span>Apple</span>
+        Apple
       </SegmentedControlOption>
       <SegmentedControlOption value="banana" data-testid="banana">
-        <span>Banana</span>
+        Banana
       </SegmentedControlOption>
     </SegmentedControl>,
   );
@@ -126,12 +126,8 @@ describe('packages/segmented-control', () => {
           name="testName"
           defaultValue={'banana'}
         >
-          <SegmentedControlOption value="apple">
-            <span>Apple</span>
-          </SegmentedControlOption>
-          <SegmentedControlOption value="banana">
-            <span>Banana</span>
-          </SegmentedControlOption>
+          <SegmentedControlOption value="apple">Apple</SegmentedControlOption>
+          <SegmentedControlOption value="banana">Banana</SegmentedControlOption>
         </SegmentedControl>,
       );
 
--- a/packages/segmented-control/src/SegmentedControl.spec.tsx
+++ b/packages/segmented-control/src/SegmentedControl.spec.tsx
@@ -16,17 +16,18 @@ const renderNewContainer = () => {
       className={testClassName}
     >
       <SegmentedControlOption value="apple" data-testid="apple">
-        Apple
+        <span>Apple</span>
       </SegmentedControlOption>
       <SegmentedControlOption value="banana" data-testid="banana">
-        Banana
+        <span>Banana</span>
       </SegmentedControlOption>
     </SegmentedControl>,
   );
 
-  const apple = getByText(rendered.container, 'Apple').parentElement as Element;
-  const banana = getByText(rendered.container, 'Banana')
-    .parentElement as Element;
+  const apple = getByText(rendered.container, 'Apple').parentElement
+    ?.parentElement as Element;
+  const banana = getByText(rendered.container, 'Banana').parentElement
+    ?.parentElement as Element;
 
   return {
     ...rendered,
@@ -125,12 +126,17 @@ describe('packages/segmented-control', () => {
           name="testName"
           defaultValue={'banana'}
         >
-          <SegmentedControlOption value="apple">Apple</SegmentedControlOption>
-          <SegmentedControlOption value="banana">Banana</SegmentedControlOption>
+          <SegmentedControlOption value="apple">
+            <span>Apple</span>
+          </SegmentedControlOption>
+          <SegmentedControlOption value="banana">
+            <span>Banana</span>
+          </SegmentedControlOption>
         </SegmentedControl>,
       );
 
-      const banana = getByText(container, 'Banana').parentElement;
+      const banana = getByText(container, 'Banana').parentElement
+        ?.parentElement;
       expect(banana).toHaveAttribute('aria-selected', 'true');
     });
   });
--- a/packages/segmented-control/src/SegmentedControl.story.tsx
+++ b/packages/segmented-control/src/SegmentedControl.story.tsx
@@ -76,16 +76,16 @@ Controlled.args = {
   name: 'fruit',
   children: [
     <SegmentedControlOption key="dragonfruit" value="dragonfruit">
-      Dragonfruit fruit
+      <span>Dragonfruit fruit</span>
     </SegmentedControlOption>,
     <SegmentedControlOption key="eggplant" value="eggplant">
-      Eggplant bananana
+      <span>Eggplant bananana</span>
     </SegmentedControlOption>,
     <SegmentedControlOption key="fig" value="fig">
-      Fig
+      <span>Fig</span>
     </SegmentedControlOption>,
     <SegmentedControlOption key="grape" value="grape">
-      Grape
+      <span>Grape</span>
     </SegmentedControlOption>,
   ],
 };
@@ -96,13 +96,13 @@ WithIcons.args = {
   name: 'language',
   children: [
     <SegmentedControlOption key="json" value="json">
-      <Icon glyph="CurlyBraces"></Icon> JSON
+      <Icon glyph="CurlyBraces"></Icon> <span>JSONNNNN and more</span>
     </SegmentedControlOption>,
     <SegmentedControlOption key="xml" value="xml">
-      <Icon glyph="Code"></Icon> XML
+      <Icon glyph="Code"></Icon> <span>XML</span>
     </SegmentedControlOption>,
     <SegmentedControlOption disabled key="shell" value="shell">
-      <Icon glyph="Shell"></Icon> Shell
+      <Icon glyph="Shell"></Icon> <span>Shell</span>
     </SegmentedControlOption>,
   ],
 };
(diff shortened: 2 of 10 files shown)


packages/segmented-control/src/SegmentedControl.styles.ts
```diff
@@ -0,0 +1,168 @@
...
+      --outer-radius: 6px;
+      --indicator-radius: 4px;
+    `,
+    [Size.Large]: css`
+      --wrapper-padding: 2px;
+      --outer-radius: 6px;
+      --indicator-radius: 4px;
+    `,
+  },
+};
```
TheSonOfThomp, line 104: These shouldn't be different between Light & Dark mode anymore

packages/segmented-control/src/SegmentedControl.styles.ts
```diff
@@ -0,0 +1,168 @@
...
+    --inner-shadow: 0px 1px 2px ${transparentize(0.7, palette.black)} inset;
+    --outer-shadow: 0px 1px 1px ${palette.gray.light2};
+    --hover-background-color: ${palette.white};
+    --indicator-background-color: ${palette.black};
+    --indicator-border-color: ${palette.black};
+  `,
+  [Theme.Dark]: css`
+    --background-color: ${palette.gray.dark4};
+    --border-color: rgba(255, 255, 255, 0);
+    --border-width: 1px;
```
TheSonOfThomp, line 62: Not sure if this difference is needed

packages/segmented-control/src/SegmentedControlOption.styles.ts
```diff
@@ -0,0 +1,211 @@
...
+      --padding-block: 3px;
+    `,
+    [Size.Default]: css`
+      --padding-block: 4px;
+    `,
+    [Size.Large]: css`
+      --padding-block: 5px;
+    `,
+  },
+};
```
TheSonOfThomp, line 89: I think these should be the same now in both modes as well

packages/segmented-control/src/types.ts
```diff
@@ -6,9 +8,149 @@ export const Size = {
 
 export type Size = typeof Size[keyof typeof Size];
 
-export const Mode = {
-  Dark: 'dark',
-  Light: 'light',
-} as const;
+export interface SegmentedControlProps
```
TheSonOfThomp, line 11: 🔥

packages/segmented-control/src/types.ts
```diff
@@ -6,9 +8,149 @@ export const Size = {
...
+   * Identifies the element(s) whose contents/presence is controlled by the segmented control.
+   *
+   * Required as a prop on the control, or on each individual option.
+   */
+  'aria-controls'?: string;
+
+  /**
+   * Styling prop
+   */
+  className?: string;
```
TheSonOfThomp, line 114: Technically don't need to list this explicitly

packages/segmented-control/src/types.ts
```diff
@@ -6,9 +8,149 @@ export const Size = {
...
+
+  /**
+   * Callback that gets called when a user makes a new selection.
+   */
+  onChange?: (value: string) => void;
+
+  /**
+   * Styling prop
+   */
+  className?: string;
```
TheSonOfThomp, line 78: Technically don't need to list this explicitly

TheSonOfThomp (changes requested):
There are some slight shape issues in dark mode (hold over from the old design). Otherwise lgtm!

shaneeza:
@TheSonOfThomp Can you post a screenshot of the issues you're seeing in darkmode?

packages/segmented-control/src/SegmentedControlOption.styles.ts
shaneeza: The paddings are different for each mode

packages/segmented-control/src/SegmentedControl.styles.ts
shaneeza: good catch

TheSonOfThomp:
> @TheSonOfThomp Can you post a screenshot of the issues you're seeing in darkmode?

Just compared to Figma — you've replicated the styles that are there. I'm just curious why the sizes are different between modes. Typically we don't do that

packages/segmented-control/src/SegmentedControl.styles.ts
TheSonOfThomp: Hmm that they are... We don't do that anywhere else though. Might be worth double checking with design?

shaneeza: updated!

packages/segmented-control/src/SegmentedControlOption.styles.ts
shaneeza: updated!

packages/segmented-control/src/SegmentedControl.story.tsx
```diff
@@ -76,16 +76,16 @@ Controlled.args = {
   name: 'fruit',
   children: [
     <SegmentedControlOption key="dragonfruit" value="dragonfruit">
-      Dragonfruit fruit
+      <span>Dragonfruit fruit</span>
```
shaneeza, line 79: @TheSonOfThomp To make the ellipsis work I initially removed `display: flex` from the `span` element because it doesn't mesh well with `text-overflow: ellipsis;`, however `display: flex` is needed to center align the `span`, especially if there is an `Icon`. One fix was to keep flex and have consumers wrap the text in a span. This however would mean that consumers would have to updated all the children once they update to this version.

TheSonOfThomp: Could we just wrap the children in another `span`?

packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -353,23 +94,14 @@ export const SegmentedControlOption = forwardRef<
...
-              className={cx(labelStyle, {
-                // TODO: Refresh - remove darkmode font override
-                [css`
-                  font-family: ${fontFamilies.legacy};
-                `]: mode === 'dark',
-              })}
-            >
-              {children}
-            </span>
+            <span className={labelStyle}>{children}</span>
```
TheSonOfThomp, line 104: Can we add another `span` here so consumers don't have to?

TheSonOfThomp: Oh I see, this affects icons being centered... I still think we should try and find a solution that doesn't require extra work on the consumers' part. Not sure atm what that should be

TheSonOfThomp: Does this help at all? https://css-tricks.com/flexbox-truncated-text/

TheSonOfThomp (commented):
Not sure I'm a fan of needing consumers to add the span themselves. Does this link help at all? https://css-tricks.com/flexbox-truncated-text/

packages/segmented-control/src/SegmentedControlOption.styles.ts
```diff
@@ -0,0 +1,198 @@
...
+  gap: calc(var(--font-size) / 2);
+  display: flex;
+  align-items: center;
+  min-width: 0;
+
+  span {
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+  }
```
TheSonOfThomp, line 193: Also, don't love that it _has_ to be a `span`

shaneeza: I can make it a `div`

packages/segmented-control/src/SegmentedControlOption.tsx
```diff
@@ -79,6 +80,15 @@ export const SegmentedControlOption = forwardRef<
       didComponentMount.current = true;
     }, [focused, followFocus, usingKeyboard, isfocusInComponent]);
 
+    /**
+     * Wraps text in a `<span>`. This span will be used to clip text and add an ellipsis. We do this so that consumers don't have to add the `<span>` themselves. We can't add the ellipsis styles to the parent container because that container has `display: flex` which doesn't work with `text-overflow: ellipsis`.
+     */
+    const renderedChildren = React.Children.map(children, child => {
```
shaneeza, line 86: @TheSonOfThomp What do you think about this? if its text then we wrap it.

TheSonOfThomp: Yeah this makes sense! I'm on the fence whether it makes more sense to assign this, or just put this block inside the JSX, but this looks good 👍 
```jsx
<span className={labelStyle}>
  {React.Children.map(children, child => {
    if (typeof child === 'string')
      return <span className={textEllipsisStyle}>{child}</span>;
    return child;
  })}
</span>
```

packages/segmented-control/src/SegmentedControlOption.styles.ts
TheSonOfThomp: Semantically it makes sense, but the fact that a user would _have to_ use a `span` and nothing else doesn't feel right

TheSonOfThomp (approved):
LGTM! 🔥

packages/segmented-control/src/SegmentedControl.spec.tsx
```diff
@@ -16,17 +16,18 @@ const renderNewContainer = () => {
       className={testClassName}
     >
       <SegmentedControlOption value="apple" data-testid="apple">
-        Apple
+        <span>Apple</span>
```
shaneeza, line 19: Remove these spans

packages/segmented-control/src/SegmentedControlOption.tsx
shaneeza: it might be easier to read inline, i'll move it

packages/segmented-control/src/SegmentedControlOption.styles.ts
shaneeza: oh, that makes sense! I guess it doesn't matter now since we're adding in the component.

PR #1490, opened 2022-11-03
LG-2517 TextInput search type `aria-label`

TheSonOfThomp:
## ✍️ Proposed changes

Fixes a regression when `type="search"`. Providing `aria-label` is sufficient for accessibility.

Updates TS types. Any one of `label`, `aria-label` and `aria-labelledby` is sufficient for all text inputs

🎟 _Jira ticket:_ [LG-2517](https://jira.mongodb.org/browse/LG-2517)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/text-input/src/TextInput.spec.tsx
```diff
@@ -256,11 +256,18 @@ describe('packages/text-input', () => {
     test('TextInput throws error when neither aria-labelledby or label is supplied', () => {
       // @ts-expect-error
       <TextInput />;
+      // @ts-expect-error
+      <TextInput aria-label="some label" />;
```
TheSonOfThomp, line 260: Not sure about this case @DesignerDave

packages/text-input/src/TextInput.tsx
```diff
@@ -57,8 +57,12 @@ import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
  * @param props.sizeVariant determines the size of the text and the height of the input.
  */
 
-type TextInput = React.ForwardRefExoticComponent<TextInputProps>;
-const TextInput: TextInput = forwardRef<HTMLInputElement, TextInputProps>(
+type TextInputComponentType = React.ForwardRefExoticComponent<TextInputProps>;
+/// @ts-expect-error
```
TheSonOfThomp, line 61: This is a temporary fix until the `SearchInput` is released

packages/text-input/src/TextInput.spec.tsx
brookescarlett: This should be allowed AFAICT

.changeset/spicy-donuts-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/text-input': patch
+---
+
+Fixes a regression. Now when `type="search"`, providing `aria-label` is sufficient for accessibility
```
brookescarlett, line 5: Aria-label should always be sufficient for accessibility as it relates to TextInput components

packages/text-input/src/TextInput.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { forwardRef, useState } from 'react';
+import React, { useState } from 'react';
```
brookescarlett, line 1: Is removing this related to the typescript changes?

packages/text-input/src/types.ts
```diff
@@ -38,26 +38,41 @@ export const TextInputFontSize = {
...
+  ['aria-labelledby']?: string;
+}
+
+type AriaLabels = keyof AriaLabelProps;
+
+interface TextInputTypeProp {
+  /**
+   * The input type.
+   *
+   * Requires an `aria-label` if `"search"` is provided
```
brookescarlett, line 63: Aria-labelledby can be sufficient here as well. We don't necessarily need aria-label if search is provided.

packages/text-input/src/TextInput.tsx
TheSonOfThomp: Just removing the explicit import in favor of `React.forwardRef`

packages/text-input/src/TextInput.spec.tsx
TheSonOfThomp: 👍 Updated to require one of `label`, `aria-label`, or `aria-labelledby` in all cases

brookescarlett (approved):

PR #1491, opened 2022-11-07
LG-2554: Refresh modal

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2554](https://jira.mongodb.org/browse/LG-2554)

Screen Shot 2022-11-07 at 1 23 47 PM

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/modal/src/Modal.styles.ts
```diff
@@ -0,0 +1,146 @@
...
+import { palette } from '@leafygreen-ui/palette';
+import { fontFamilies } from '@leafygreen-ui/tokens';
+import facepaint from 'facepaint';
+import { transparentize } from 'polished';
+import { CloseIconColor, ModalSize } from './types';
+
+// breakpoints for different screen sizes
+export const small = '767px'; // mobile screens, from 0px - 767px
+export const medium = '768px'; // tablet screens, from 768px - 1024px
+export const large = '1025px'; // laptops/desktop screens, from 1025px and above
```
spark33, line 12: Unrelated to the PR changes, but we should import breakpoints from our tokens library

packages/modal/src/ModalView.tsx
```diff
@@ -227,72 +85,58 @@ function ModalView({
...
-              [css`
-                background-color: ${transparentize(0.4, uiColors.black)};
-              `]: darkMode,
-            })}
+            className={cx(
+              className,
+              backdropBaseStyle,
+              backdropThemeStyles[theme],
+              {
+                [visibleBackdrop]: state === 'entered',
```
spark33, line 93: Is there no enum for `state`?

spark33 (approved):

packages/modal/src/types.ts
```diff
@@ -0,0 +1,77 @@
...
+   * With this option you can specify a different element to receive that initial focus.
+   * Selector string (which will be passed to document.querySelector() to find the DOM node)
+   */
+  initialFocus?: string;
+
+  /**
+   * Determines if the component will appear in dark mode.
+   * @default false
+   */
+  darkMode?: boolean;
```
spark33, line 70: We should extend `DarkModeProps` from our lib.

spark33 (changes requested):

packages/modal/src/ModalView.tsx
shaneeza: there is no enum, only a string union
```type TransitionStatus = "entering" | "entered" | "exiting" | "exited" | "unmounted"```

File: packages/modal/src/Modal.styles.ts (hunk 1 above)
Comment (shaneeza) [reply packages/modal/src/Modal.styles.ts:12]: Leaving for now because these don't match the breakpoints from the tokens library

Comment (brookescarlett) [reply packages/modal/src/Modal.styles.ts:12]: Could run this by design quickly to see if we're ok to move to the system's breakpoints. Not sure if that makes this a breaking change, though.

File: packages/modal/src/Modal.styles.ts [hunk 4]
```diff
@@ -0,0 +1,146 @@
...
+  font-family: ${fontFamilies.default};
+  transition: all ${transitionDuration.default}ms ease-in-out;
+  margin: auto;
+  max-height: calc(100% - ${defaultVerticalSpacing}px);
+  position: relative;
+  pointer-events: all;
+  transform: translate3d(0, -16px, 0);
+  opacity: 0;
+  border-radius: 24px;
+  padding: 40px 36px;
```
brookescarlett, line 68: nit, non-blocking but do we have tokens for these values?

packages/modal/src/Modal.styles.ts
```diff
@@ -0,0 +1,146 @@
...
+  box-shadow: 0px 8px 20px -8px ${transparentize(0.4, palette.black)};
+
+  &:focus {
+    outline: none;
+  }
+`;
+
+export const modeStyles = css``;
```
brookescarlett, line 76: rm?

packages/modal/src/Modal.tsx
```diff
@@ -136,6 +61,8 @@ Modal.propTypes = {
   shouldClose: PropTypes.func,
   className: PropTypes.string,
   setOpen: PropTypes.func,
+  darkMode: PropTypes.bool,
```
brookescarlett, line 64: not related to this PR specifically but what's the status of propTypes? We have them on some components and not others

packages/modal/src/ModalView.tsx
```diff
@@ -1,172 +1,29 @@
...
-      // so there's already (28 - 16) / 2 = 6px of spacing. 16 - 6 = 10.
-      right: 10px;
-      top: 10px;
-    `,
-  },
-};
-
-const closeClassName = createUniqueClassName();
+import { CloseIconColor, ModalProps, ModalSize } from '././types';
+import {
```
brookescarlett, line 15: nit, non-blocking, curious about our thoughts on doing import * from './styles' so we dont have to name them all out

brookescarlett: cc @spark33 what do you think?

spark33: I like that idea! Especially since we're having a `styles.ts` for every subdirectory, all styles exported from that file should be relevant in the component file.

shaneeza: yeah I don't mind

packages/modal/src/Modal.tsx
shaneeza: I believe we still need them for those who aren't using TS. We should add to those that are missing.

packages/modal/src/Modal.styles.ts
shaneeza: yeah I remember Sooa was working on updating Modal widths, i'll check in with her

packages/modal/src/Modal.styles.ts
shaneeza: Do you mean the spacing tokens?

packages/modal/src/ModalView.tsx
shaneeza: Looks like we would have to use this syntax if we import `*`, `import * as styles from './Modal.styles';`. So for each style we would have to use individual exports,`style.modalContentStyle`

https://bobbyhadz.com/blog/typescript-import-all-exports-from-file

If we have to do this I'd rather name all the exports at the top. Thoughts?

packages/modal/src/Modal.styles.ts
shaneeza: Really looking at this, i don't even think we need this package. This is targeting 3 different breakpoints but the styles are only change at the `large` breakpoint. I think i'm gonna remove it.

packages/modal/src/Modal.styles.ts
brookescarlett: yes

packages/modal/src/ModalView.tsx
brookescarlett: yeah you're probably right. feel free to leave as-is

packages/modal/src/Modal.styles.ts
shaneeza: I guess this is something else we can bring up during the API brainstorm, these numbers don't match the spacing tokens. It would only work if we did something like `spacing[1] * 10`

brookescarlett: Nono that's not right. Leave it as is, but maybe we need a more formal way to notify design when things don't follow their patterns

packages/modal/src/Modal.styles.ts
brookescarlett: sgtm

packages/modal/src/Modal.story.tsx
```diff
@@ -1,13 +1,14 @@
 import React, { useState } from 'react';
 import Button from '@leafygreen-ui/button';
 import { css } from '@leafygreen-ui/emotion';
-import Modal from '.';
+import Modal from './Modal/Modal';
```
brookescarlett, line 4: nit: can we import this from index still?

PR #1492, opened 2022-11-07
LG-2534: PopoverClassName on Select component

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2534](https://jira.mongodb.org/browse/LG-2534)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/select/src/Select.story.tsx
```diff
@@ -81,6 +83,26 @@ export const Controlled = ({
...
+  return (
+    <Uncontrolled
+      {...args}
+      readOnly={false}
+      value={value}
+      onChange={setValue}
+      popoverClassName={css`
+        min-width: 500px;
+      `}
+      justify="end"
```
shaneeza, line 101: Should we use the value from storybook?

packages/select/src/ListMenu.tsx
```diff
@@ -63,6 +63,7 @@ const ListMenu = React.forwardRef<HTMLUListElement, ListMenuProps>(
       scrollContainer,
       portalClassName,
       popoverZIndex,
+      ...rest
```
shaneeza, line 66: I don't know if we should add this yet. Consumers shouldn't be able to change stuff like `adjustOnMutation`, `spacing`, `onClick`, `align`. Not sure why it wasn't omitted. If we do keep this maybe we update the type to omit these props?

shaneeza (commented):
@spark33 I left some comments above ^^

packages/select/src/ListMenu.tsx
spark33: Hm okay. I'll just remove it for now then - since we're not separating Popover's props from the Button's, supporting all props now would be kind of a hassle. That can come as a part of a larger effort later.

.changeset/flat-melons-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/select': patch
+---
+
+popoverClassName prop added. The className is passed to the select options menu
```
brookescarlett, line 5: did we end up doing discovery around what popover pass through props we use/don't across components?

PR #1504, opened 2022-11-17
Reorganize file structure in Stepper

spark33:
## ✍️ Proposed changes

This is a POC PR to demonstrate a subdirectory-driven file structure. It feels nice and readable since no file is over 100 lines, and it is easy to tell what level of componentization a piece of code pertains to.

Another smaller but notable pattern I've used here is using string interpolation in css strings returned from functions (i.e. `(darkMode) => css'...'`) instead of `cx` calls. I did so because using `cx` calls would have meant creating two css string variables when it can be done in one line using string interpolation. It makes things easier to track and understand the logic of when it is all handled in one place.

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


brookescarlett:
Are there any test improvements that need to be made in order to make sure that we're not introducing any regressions with this refactor?

packages/stepper/src/EllipsesStep/index.tsx
```diff
@@ -0,0 +1,47 @@
+import Tooltip, { Align, Justify } from '@leafygreen-ui/tooltip';
```
TheSonOfThomp, line 1: Undecided on whether I prefer using `index.tsx` here, vs `EllipsesStep.tsx`, or both files

I see the benefits of this version (import links are shorter), but it makes it harder to find files and see which file you're working on. In VSCode you can find a specific file (`CMD-T` I think). Having these all named index makes this harder. It's also harder to see which file you have open. E.x. If I have `EllipsesStep/index` and `InternalStep/index` open it takes a bit more attention to see which file I'm in

spark33: Yeah this is the tail end of the discussion your pupper took you away from 😂

Dave and I agreed with you, and I'll be raising POC PRs with that change applied.

packages/stepper/src/EllipsesStep/styles.ts
```diff
@@ -0,0 +1,35 @@
+import { css } from '@leafygreen-ui/emotion';
+import { palette } from '@leafygreen-ui/palette';
+import { spacing } from '@leafygreen-ui/tokens';
+import { stepIconClassName } from '../constants';
+import { StepStates } from '../types';
+import { EllipsesStepStates } from './types';
+
+// TODO: would be good to put these styles inside a baseStyles, but it is currently impossible because the <Tooltip> content is an iframe.
```
TheSonOfThomp, line 8: I think I commented on this in the original PR, but Tooltip is in a Portal, not an iframe. Not sure how that affects our ability to style it

brookescarlett:
I think we agreed that this POC should be closed for now as too many changes were introduced and we wanted to pilot on smaller / more straightforward components without the CSS changes. Let me know if there's FUD here.

spark33:
Closed to be merged at a later date.

PR #1506, opened 2022-11-18
Converts Typography package to use new directory/file structure

brookescarlett:
## ✍️ Proposed changes
Converts Typography package to use new directory/file structure

We *should* make JIRA tickets for all of the package migrations

## 🛠 Types of changes
- [x]  Refactor

## ✅ Checklist

## 💬 Further comments
Things to consider (could be addressed within this PR, could be addressed in a follow up)
- Why do some components inherit from Box and others do not?
- Do we still support propTypes? If not, can we remove propType definitions from components here? 
- I didn't want to touch too many things at once here, but there are definitely opportunities for further refactoring/consistency within these components.

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/typography/src/Description/Description.types.ts
```diff
@@ -0,0 +1,6 @@
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+
+export type DescriptionProps = HTMLElementProps<'p', never> & {
+  darkMode?: boolean;
```
spark33, line 4: Why aren't we extending CommonTypographyProps here?

packages/typography/src/InlineCode/InlineCode.tsx
```diff
@@ -0,0 +1,78 @@
...
+    forwardedRef,
+  ) => {
+    const { usingKeyboard: showFocus } = useUsingKeyboardContext();
+    const baseFontSize = useUpdatedBaseFontSize();
+    const { theme } = useDarkMode(darkModeProp);
+    const whiteSpace =
+      ((typeof children === 'string' && children.match(/./gu)?.length) ?? 0) <=
+      30
+        ? nowrap
+        : normal;
```
spark33, line 34: It feels like this should be in the styles file, then referred to from a `cx` call.

packages/typography/src/InlineCode/InlineCode.tsx
```diff
@@ -0,0 +1,78 @@
...
+          {
+            [codeLinkStyleModes[theme]]: isAnchor,
+            [codeFocusModes[theme]]: showFocus,
+          },
+          className,
+        )}
+      >
+        {children}
+      </code>
+    );
```
spark33, line 54: I think this can just go directly in the render function for better readability

packages/typography/src/Link/Link.tsx
```diff
@@ -0,0 +1,100 @@
...
+          [arrowRightIconHover]: arrowAppearance === ArrowAppearance.Hover,
+          [arrowRightIconPersist]: arrowAppearance === ArrowAppearance.Persist,
+        })}
+      />
+    );
+  }
+
+  const elementProps = href
+    ? ({ as: 'a', href, target, rel } as const)
+    : ({ as: 'span' } as const);
```
spark33, line 78: There's a lot going on in this component. We could reorganize a lot of this (i.e. a `LinkIcon` subcomponent)

packages/typography/src/Description/Description.types.ts
spark33, line 4: Why aren't we extending CommonTypographyProps here?

packages/typography/src/InlineCode/InlineCode.tsx
spark33, line 34: It feels like this should be in the styles file, then referred to from a cx call.

spark33 (changes requested):

packages/typography/src/Description/Description.types.ts
brookescarlett: I didn't modify implementation details just moved code around.

packages/typography/src/InlineCode/InlineCode.tsx
brookescarlett: Agreed but per my above comment, was trying to decouple implementation details from this PR. Just trying to re-org and not do too many things at once.

packages/typography/src/Link/Link.tsx
brookescarlett: Goal of this PR was purely structural -- don't want code changes muddying the ability to merge.

packages/typography/src/Description/Description.types.ts
brookescarlett: updated

spark33 (approved):

PR #1507, opened 2022-11-18
Banner reorganization

spark33:
## ✍️ Proposed changes

Two subcomponents in this reorganization - `BannerIcon` and `BannerDismissButton`. These are quite small subcomponents - especially `BannerDismissButton`. 

I made `BannerIcon` a subcomponent because it has logic that affects the render function. 

For `BannerDismissButton`, it wasn't necessary but I did so because there were 120+ lines of style code that I thought would be nice to break out. Because `BannerDismissButton` only breaks out styling logic, I think a better fix would be to use `createUniqueClassname` to style the dismiss button from inside the top-level Banner's `variantStyles`; I just left it alone for now since I thought that would be a bigger refactor that we want atm, and because I figured this approach would be more useful in facilitating discussion around the file restructure.

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


shaneeza (approved):
Looks good!

PR #1511, opened 2022-11-21
Update Package Dependencies & TSConfigs

TheSonOfThomp:
## ✍️ Proposed changes

1. Adds missing packages used in `.story` or `.spec` files as devDependencies
2. Ensures packages listed as `devDependencies` are truly used as `devDependencies` (i.e. only in `.story` or `.spec` files), and adds them as `dependencies` instead, if appropriate
3. Adds missing packages in `tsconfig` files, (but excludes devDependencies to prevent circular deps)
4. Updates `depcheck` script to automate the above steps

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)
5. Doesn’t meet agreed upon standards (😬 ⚠️ ‼️)
6. Missing an edge case (🛑 🙅 ❌)
7. Missing core functionality (⛔️ ☣️ 🚧 🌋 💀)


packages/guide-cue/tsconfig.json
```diff
@@ -42,6 +41,9 @@
     },
     {
       "path": "../typography"
+    },
+    {
+      "path": "../leafygreen-provider"
```
TheSonOfThomp, line 46: Provider is put at the end since it's a `peerDependency`

packages/popover/package.json
```diff
@@ -27,9 +27,6 @@
     "@leafygreen-ui/tokens": "^1.3.4",
     "react-transition-group": "^4.4.1"
   },
-  "devDependencies": {
-    "@leafygreen-ui/palette": "^3.4.4"
-  },
```
TheSonOfThomp, line 32: This is one of 2 real changes. Removes unused dev dependency

packages/ripple/package.json
```diff
@@ -22,9 +22,6 @@
   "dependencies": {
     "@leafygreen-ui/tokens": "^1.3.4"
   },
-  "devDependencies": {
-    "@leafygreen-ui/lib": "^10.0.0"
-  },
```
TheSonOfThomp, line 27: This is 2 of 2 real changes. Removes unused dev dependency

packages/popover/package.json
brookescarlett: should this be rm'd from the tsconfig as well?

packages/ripple/package.json
brookescarlett: ditto here

brookescarlett (changes requested):
if were rm'ing from package.json should we from tsconfig too?

TheSonOfThomp:
> if were rm'ing from package.json should we from tsconfig too?

Should be good now

packages/popover/package.json
```diff
@@ -27,9 +27,6 @@
     "@leafygreen-ui/tokens": "^1.3.4",
     "react-transition-group": "^4.4.1"
   },
-  "devDependencies": {
-    "@leafygreen-ui/palette": "^3.4.4"
```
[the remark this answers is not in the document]
brookescarlett: We need it in the storybook actually so this should stay

packages/popover/tsconfig.json
```diff
@@ -19,17 +19,14 @@
     {
       "path": "../lib"
     },
-    {
-      "path": "../leafygreen-provider"
-    },
-    {
-      "path": "../palette"
```
brookescarlett, line 26: ditto

packages/ripple/package.json
```diff
@@ -22,9 +22,6 @@
   "dependencies": {
     "@leafygreen-ui/tokens": "^1.3.4"
   },
-  "devDependencies": {
-    "@leafygreen-ui/lib": "^10.0.0"
```
brookescarlett, line 26: does ripple need react? if so this won't work

packages/popover/tsconfig.json
TheSonOfThomp: Updates the depcheck script to add packages back as `devDependencies` if they're used in `.story` or `.spec`

packages/ripple/package.json
TheSonOfThomp: It's only actually needed as a devDependency (in spec files)

brookescarlett:
Will you re lint and re-r? This is a heavy PR for what I'm assuming is a lightweight change :)

.github/workflows/nodejs.yml
```diff
@@ -186,15 +186,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-cache-
       - run: yarn install --frozen-lockfile
-      - uses: actions/cache@v2
```
brookescarlett, line 189: why are we rm'ing this?

TheSonOfThomp: When we run `check:builds` we should be rebuilding and not checking an older cached version

TheSonOfThomp: At least for now until we make sure the builds are caching properly

packages/checkbox/package.json
```diff
@@ -20,17 +20,16 @@
...
     "@leafygreen-ui/tokens": "^1.4.0",
     "@leafygreen-ui/typography": "^15.1.0",
-    "@leafygreen-ui/a11y": "^1.3.4",
-    "@leafygreen-ui/emotion": "^4.0.3"
-  },
-  "devDependencies": {
-    "@leafygreen-ui/palette": "^3.4.4"
+    "react-transition-group": "^4.4.1"
   },
+  "devDependencies": {},
```
brookescarlett, line 32: can we rm if empty?

packages/checkbox/package.json
```diff
@@ -20,17 +20,16 @@
...
     "@leafygreen-ui/lib": "^10.0.0",
+    "@leafygreen-ui/palette": "^3.4.4",
     "@leafygreen-ui/tokens": "^1.4.0",
     "@leafygreen-ui/typography": "^15.1.0",
-    "@leafygreen-ui/a11y": "^1.3.4",
-    "@leafygreen-ui/emotion": "^4.0.3"
-  },
-  "devDependencies": {
-    "@leafygreen-ui/palette": "^3.4.4"
+    "react-transition-group": "^4.4.1"
```
brookescarlett, line 30: nit should this go first? i think above we had a next import statement before lg ones

TheSonOfThomp: I think they go alphabetically by default

packages/ripple/package.json
```diff
@@ -22,9 +22,6 @@
   "dependencies": {
     "@leafygreen-ui/tokens": "^1.4.0"
   },
-  "devDependencies": {
```
brookescarlett, line 25: confused how we don't need this as a dev dep?

packages/ripple/tsconfig.json
```diff
@@ -9,11 +9,8 @@
     "src/**/*"
   ],
   "references": [
-    {
-      "path": "../lib"
```
brookescarlett, line 13: ditto

packages/ripple/package.json
TheSonOfThomp: Since none of the .story or .spec files explicitly import `lib`, we can rely on the global installation of `react` for development purposes

brookescarlett (approved):

PR #1515, opened 2022-11-23
LG-2161 Create `Polymorphic` component

TheSonOfThomp:
## ✍️ Proposed changes

`Polymorphic` is the successor to `Box`

### Basic Usage 
```tsx
interface MyProps {
  someProp: string;
}

const MyComponent = (props: MyProps) => {
  const shouldRenderAs = 'button';
  return <Polymorph as={shouldRenderAs} {...props} />;
};
```

### Extending Polymorphic behavior
```tsx
interface MyProps {
  someProp: string;
}
const MyComponent = Polymorphic<MyProps>(
  ({ as, ...rest }, forwardedRef) => {
    const { Component } = usePolymorphic(as);
    return <Component ref={forwardedRef} {...rest} />;
  },
);

// 

<MyComponent as="span" />
```

### Inferred `as` prop

Components extended using the `Polymorphic` factory function can be made to infer the `as` prop value based on the `href` passed in.
Ensure the custom props are wrapped in `InferredPolymorphicProps`, and use the `useInferredPolymorphic` hook. Make sure to pass both `as` and `rest` (or an object that contains `href`) into the hook.

```tsx
export const MyComponent = Polymorphic<
  InferredPolymorphicProps<MyProps>
>(({ as, ...rest }) => {
  const { Component, ref } = useInferredPolymorphic(as, rest);
  return (
    <Component ref={ref} {...rest}>
      {title}
    </Component>
  );
});

//

<MyComponent href="mongodb.design" />; // renders as <a>
```

### With Emotion `styled` API
```tsx
const StyledPolymorph = styled(Polymorph)`
  color: hotpink;
` as PolymorphicComponentType;

// or

const MyStyledComponent = styled(MyComponent)`
  color: hotpink;
` as PolymorphicComponentType;
```
Note: TSDocs will not compile for styled polymorphs. This can be remedied by creating a wrapper around the `styled` function that explicitly returns a `PolymorphicComponentType`

🎟 _Jira ticket:_ [Refactor Box](https://jira.mongodb.org/browse/LG-2161)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/internal/src/Polymorphic/Polymorphic.tsx
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable react/display-name */
+/**
+ * Based on https://blog.logrocket.com/build-strongly-typed-polymorphic-components-react-typescript/
+ */
+
+import React from 'react';
+
+/** An interface that defines the `as` prop */
+export interface AsProp<T extends React.ElementType> {
```
brookescarlett, line 9: let's follow the new directory structure conventions with this new component
`/Polymorphic/Polymorphic.tsx`,  `/Polymorphic/Polymorphic.types.ts`, `/Polymorphic/Polymorphic.styles.ts`

packages/internal/src/Polymorphic/README.md
```diff
@@ -0,0 +1,10 @@
+# Polymorphic
+
+## Prior art
```
brookescarlett, line 3: love this

packages/internal/src/Polymorphic/Polymorphic.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    return (
+      <Component {...rest} ref={ref}>
+        {children}
+      </Component>
+    );
+  },
+) as PolymorphicComponentType;
+
+// eslint-disable-next-line react/display-name
```
brookescarlett, line 81: rm?

packages/internal/src/Polymorphic/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,326 @@
...
+    },
+  );
+
+  Wrapper.displayName = 'Wrapper';
+
+  return {
+    Wrapper,
+    wrapperDidRender,
+  };
+}
```
brookescarlett, line 326: Can we add a test that confirms this new implementation plays well with Emotion?

```
// We don't want TS to throw an error here
styled(Button)`
  // styles
`
```

packages/internal/src/Polymorphic/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,358 @@
...
+import { Polymorphic, usePolymorphicRef } from '.';
+import {
+  ExampleComponent,
+  ExampleForwardRef,
+  ExampleForwardRefWithHook,
+  ExampleWithHook,
+  RestrictedExample,
+} from './Example';
+
+describe('packages/internal/polymorphic', () => {
```
brookescarlett, line 13: what happens with `href`

packages/polymorphic/src/Example/index.ts
```diff
@@ -0,0 +1,9 @@
+// Uncomment these to debug tsdoc generation
```
brookescarlett, line 1: do we need this comment?

packages/polymorphic/src/Example/Polymorphic.example.tsx
```diff
@@ -0,0 +1,131 @@
...
+  /** An arbitrary title */
+  title?: string;
+  /** Flag for dark mode */
+  darkMode?: boolean;
+}
+
+/**
+ * Uses `usePolymorphic` hook
+ */
+
```
brookescarlett, line 25: nit rm line break

brookescarlett (commented):
Let's use the new dir structure here (see Toggle, Banner, etc)

packages/polymorphic/src/InferredPolymorphic.tsx
```diff
@@ -0,0 +1,37 @@
+import { usePolymorphic } from './Polymorphic.hooks';
+import { AnchorProps, PolymorphicAs } from './Polymorphic.types';
+
+/**
+ * Wrapping props in this type ensures that if `href` is defined,
+ * the `as` type is set to `'a'`, and all anchor props are accepted
+ */
+export type InferredPolymorphicProps<P = {}> =
```
brookescarlett, line 8: this should prob live in the types file

packages/polymorphic/src/Polymorphic.hooks.tsx
```diff
@@ -0,0 +1,36 @@
...
+        ? HTMLElementTagNameMap[E]
+        : unknown)
+    | null
+  >(null);
+};
+
+export function usePolymorphicComponent<T extends PolymorphicAs>(
+  as?: T,
+): React.ElementType {
+  const Component = as || 'div';
```
brookescarlett, line 21: nit can collapse:
`as ? as : 'div'`

packages/polymorphic/src/Polymorph.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Polymorphic component that supports forwarded refs
+ *
+ * With the `as` prop, Polymorphic can dynamically render
+ * as any HTML element or React Component.
+ */
+export const Polymorph: PolymorphicComponentType =
+  React.forwardRef(BasePolymorph);
+Polymorph.displayName = 'Polymorph';
```
brookescarlett, line 34: are we still supporting propTypes?

packages/polymorphic/src/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,505 @@
...
+        <Polymorph as="a" href="mongodb.design" ref={anchorRef} />
+        <Polymorph as="a" href="mongodb.design">
+          some content
+        </Polymorph>
+        {/* Should accept href (and render it as 'a') */}
+        {/* <Polymorph href="mongodb.design" /> */}
+
+        <Polymorph as="input" />
+        {/* TODO: ts-expect-error - Input should not accept children */}
+        {/* <Polymorph as="input">some content</Polymorph> */}
```
brookescarlett, line 51: is this a regression?

packages/polymorphic/src/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,505 @@
...
+
+        <Polymorph as={Wrapper} />
+        <Polymorph as={Wrapper} ref={spanRef} />
+        {/* TODO: ts-expect-error - Must pass the correct ref type */}
+        <Polymorph as={Wrapper} ref={divRef} />
+        <Polymorph as={Wrapper} ref={spanRef} darkMode={true} />
+        {/* @ts-expect-error - Theme is not a prop on Wrapper */}
+        <Polymorph as={Wrapper} ref={spanRef} theme={'dark'} />
```
brookescarlett, line 59: great test suite, but let's add one for styled

packages/polymorphic/src/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,505 @@
...
+  RestrictedExample,
+} from './Example';
+import { makeWrapperComponent } from './utils/Polymorphic.testutils';
+
+describe('packages/polymorphic', () => {
+  describe('Basic Polymorphic Component', () => {
+    /* eslint-disable jest/no-disabled-tests */
+    test.skip('Prop Types behave correctly', () => {
+      const { Wrapper } = makeWrapperComponent();
+      const divRef = usePolymorphicRef<'div'>(); // React.useRef<HTMLDivElement | null>(null);
```
brookescarlett, line 21: do we need this comment?

packages/polymorphic/src/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,505 @@
...
+  ExamplePolymorphic,
+  ExamplePolymorphicWithRef,
+  ExampleInferred,
+  AdvancedPolymorphic,
+  AdvancedPolymorphicWithRef,
+  RestrictedExample,
+} from './Example';
+import { makeWrapperComponent } from './utils/Polymorphic.testutils';
+
+describe('packages/polymorphic', () => {
```
brookescarlett, line 16: should we break this into two files: testing polymorph and polymorphic?

packages/polymorphic/src/Polymorphic.types.ts
```diff
@@ -0,0 +1,94 @@
+/**
+ * Based on https://blog.logrocket.com/build-strongly-typed-polymorphic-components-react-typescript/
+ */
+// TODO: Consider restricting HTML elements to only elements that accept children
```
brookescarlett, line 4: you did this right?

packages/polymorphic/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  PolymorphicPropsWithRef,
+  PolymorphicRef,
+  PolymorphicComponentType,
+} from './Polymorphic.types';
+
+export { useInferredPolymorphic } from './InferredPolymorphic';
+
+export type { InferredPolymorphicProps } from './InferredPolymorphic';
+
+expect.extend({
```
brookescarlett, line 23: should this be exported from test utils instead, or are we intentionally making this external?

scripts/parse-tsdocs.ts
```diff
@@ -1,27 +1,9 @@
 /* eslint-disable no-console */
-import {
-  withCompilerOptions,
```
brookescarlett, line 3: is this intentional?

packages/polymorphic/src/InferredPolymorphic.tsx
TheSonOfThomp: I kept this one here to keep all the "inferred" utilities together

packages/polymorphic/src/Polymorph.tsx
TheSonOfThomp: We've gone back & forth on that. We should bring this up at the next engineers jam

packages/polymorphic/src/Polymorphic.spec.tsx
TheSonOfThomp: No, TS doesn't even catch this for a straight `input`

packages/polymorphic/src/index.ts
TheSonOfThomp: I'm intentionally making this external. I've already written `expect(InputOption).toBePolymorphic()`

scripts/parse-tsdocs.ts
TheSonOfThomp: Yeah, I had to move all the utils to a separate file so I could import them into the `.spec` file

packages/polymorphic/src/Polymorphic.spec.tsx
brookescarlett: So we can rm

packages/polymorphic/src/index.ts
brookescarlett: kk

packages/polymorphic/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "type": "git",
+    "url": "https://github.com/mongodb/leafygreen-ui"
+  },
+  "bugs": {
+    "url": "https://jira.mongodb.org/projects/LG/summary"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {},
```
brookescarlett, line 30: rm?

packages/polymorphic/src/InferredPolymorphic.tsx
brookescarlett: cc @spark33

packages/polymorphic/src/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,505 @@
...
+        <Polymorph as="div" href="mongodb.design" />
+
+        {/* @ts-expect-error - Require href when as="a" */}
+        <Polymorph as="a" />
+        <Polymorph as="a" href="mongodb.design" />
+        <Polymorph as="a" href="mongodb.design" ref={anchorRef} />
+        <Polymorph as="a" href="mongodb.design">
+          some content
+        </Polymorph>
+        {/* Should accept href (and render it as 'a') */}
```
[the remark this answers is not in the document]
TheSonOfThomp: No, only HOCs support this. rm-ing

packages/polymorphic/src/Polymorphic.spec.tsx
TheSonOfThomp: Line ~150 has tests for `styled`, or do you mean type tests for styled?

packages/polymorphic/src/Polymorphic.spec.tsx
TheSonOfThomp: Probably!

packages/polymorphic/src/Polymorphic.types.ts
TheSonOfThomp: No, it currently still accepts any HTML/SVG tag

brookescarlett:
> Let's use the new dir structure here (see Toggle, Banner, etc)

This still hasn't been updated. We should have two directories for Polymorph and Polymorphic

packages/polymorphic/README.md
```diff
@@ -0,0 +1,142 @@
...
+  const shouldRenderAs = 'button';
+  return <Polymorph as={shouldRenderAs} {...props} />;
+};
+```
+
+## Extending Polymorphic behavior
+
+If you want to expose `as` as a prop of your component, use the `Polymorphic` factory function and related hooks.
+
+Note that any inherited props will be indeterminate in the factory function, since the value `as` is not known. (i.e. the attributes of `rest` in the following example are not definitely known).
```
spark33, line 37: I feel like `...the following example are not definitely known` -> `... the following example are unknown` reads better

packages/polymorphic/README.md
```diff
@@ -0,0 +1,142 @@
...
+    <Component ref={ref} {...rest}>
+      {title}
+    </Component>
+  );
+});
+
+//
+
+<MyInferredComponent href="mongodb.design" />; // renders as <a>
+```
```
spark33, line 86: I like this pattern and name! I am curious though if we need this anywhere outside of Button?

brookescarlett: MenuItem, SideNavItem

packages/polymorphic/src/InferredPolymorphic.tsx
spark33: I think in this case, we should have a `InferredPolymorphic` subdir with a types.ts file inside. I could also see each of the other functions being their own files.

packages/polymorphic/src/Polymorph.tsx
spark33: Yeah I think we'd need to support it for devs that don't use typescript

packages/polymorphic/src/Polymorphic.spec.tsx
brookescarlett: type tests

packages/polymorphic/src/Polymorphic.types.ts
brookescarlett: lets rm this comment and make a ticket

packages/polymorphic/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+### With Refs
+
+Accepting a forwarded ref to a Polymorphic component is as simple as passing in a ref into the render function (same as `React.forwardRef`).
+
+```tsx
+interface MyProps {
+  someProp: string;
+}
+const MyComponent = PolymorphicComponent<MyProps>(
```
shaneeza, line 57: Where is `PolymorphicComponent` coming from? Do you mean `Polymorphic`?

packages/polymorphic/src/Polymorphic.hooks.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+import { PolymorphicAs, PolymorphicRef } from './Polymorphic.types';
+
+/**
+ * A wrapper around `React.useRef`
+ * that provides an accurately typed Ref object
+ */
+export const usePolymorphicRef = <E extends PolymorphicAs>(_?: E) => {
+  // By accepting a prop with a generic type, we can type this function using JS-domain vars
```
shaneeza, line 10: What is JS-domain vars?

TheSonOfThomp: This refers to real JavaScript variables (here the `_` arg) as opposed to Typescript generic variables (here the type `E`)

I'll update the comment to clarify, but essentially, by accepting an argument of the same generic type, we can set the type `E` using a JS variable.

i.e. instead of calling `usePolymorphicRef<'div'>()` we can instead call `usePolymorphicRef('div')`

packages/polymorphic/README.md
TheSonOfThomp: Yep, changed that

TheSonOfThomp:
> > Let's use the new dir structure here (see Toggle, Banner, etc)
> 
> This still hasn't been updated. We should have two directories for Polymorph and Polymorphic

I don't fully agree with this, since `Polymorph` is the component, but `Polymorphic` is a factory function that creates a new component

packages/polymorphic/README.md
spark33: I'm still seeing this?

spark33:
> > > Let's use the new dir structure here (see Toggle, Banner, etc)
> > 
> > 
> > This still hasn't been updated. We should have two directories for Polymorph and Polymorphic
> 
> I don't fully agree with this, since `Polymorph` is the component, but `Polymorphic` is a factory function that creates a new component

Why would this distinction impact why we wouldn't want to break the files out into subdirectories? The subdirectory structure doesn't imply that one is a component and another is a factory.

Generally, I'm not a fan of making an edge case out of this. This folder structure looks vastly different from the structure we agreed on, and it not only feels less digestible, it also goes against the very idea of why we're trying to standardize. If the functionalities were so different that it warrants an edge case, I would vote for breaking them out into a separate exports. I don't think that's the case here, so a subdirectory makes sense to me.

TheSonOfThomp:
> Branched off of this branch to try and have Label use Polymorphic and am getting a TS error here:

When assigning a default, I think you have to assert it `as PolymorphicAs`

TheSonOfThomp:
> Why would this distinction impact why we wouldn't want to break the files out into subdirectories? The subdirectory structure doesn't imply that one is a component and another is a factory.
> 
> Generally, I'm not a fan of making an edge case out of this. This folder structure looks vastly different from the structure we agreed on, and it not only feels less digestible, it also goes against the very idea of why we're trying to standardize. If the functionalities were so different that it warrants an edge case, I would vote for breaking them out into a separate exports. I don't think that's the case here, so a subdirectory makes sense to me.

I don't see this as an edge case, merely a (potentially confusing) choice of file name. We could just as easily named the factory function something like `Polymorphic.factory.ts`, `.functions.ts` or `.utilities.ts`. This follows the convention we have of `.styles.ts`, `.types.ts`, `.spec.ts` etc.

Really what we're debating here is the naming convention for exported helper functions. `.hooks.ts` is clear enough, but what do we name other functions, like this factory?

packages/polymorphic/src/Polymorph.tsx
TheSonOfThomp: 👍 Updated types & Added tests to ensure propTypes support

packages/polymorphic/src/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,262 @@
...
+        expect(testRef!).toBeDefined();
+        expect(testRef!.current).toBeDefined();
+      });
+    });
+
+    describe('With Emotion `styled` API', () => {
+      test('Basic styled component', () => {
+        const StyledPolymorph = styled(Polymorph)`
+          color: #ff69b4;
+        `; // as PolymorphicComponentType;
```
brookescarlett, line 152: rm comment?

packages/polymorphic/src/Polymorphic.spec.tsx
```diff
@@ -0,0 +1,262 @@
...
+        expect(getByTestId('styled')).toBeInTheDocument();
+        expect(getByTestId('styled').tagName.toLowerCase()).toBe('span');
+        expect(getByTestId('styled')).toHaveStyle(`color: #ff69b4;`);
+      });
+
+      /* eslint-disable jest/no-disabled-tests */
+      test.skip('TypeScript types are still correct using Styled ', () => {
+        const StyledPolymorph = styled(Polymorph)`
+          color: #ff69b4;
+        ` as PolymorphicComponentType;
```
brookescarlett, line 203: do we need the typecasting? per above it looks like we might not?

packages/polymorphic/src/Example/Polymorphic.example.tsx
```diff
@@ -0,0 +1,132 @@
...
+  InferredPolymorphicProps<ExampleProps>
+>(({ as, title, ...rest }) => {
+  const { Component, ref } = useInferredPolymorphic(as, rest);
+  return (
+    <Component ref={ref} {...rest}>
+      {title}
+    </Component>
+  );
+}, 'ExampleInferred');
+
```
brookescarlett, line 61: nit but should we add an example of passing a default value?

packages/polymorphic/README.md
```diff
@@ -0,0 +1,148 @@
+# Polymorphic
+
+> ### **pol·y·morph** _(noun)_
```
spark33, line 3: General question - should we name this package `Polymorph`? and `Polymorphic` and other resources can be utilities around Polymorph? 

Feels nicer since it'd be a noun and the relationship ^ between Polymorph and Polymorphic feels more intuitive to me with that change.

packages/polymorphic/README.md
```diff
@@ -0,0 +1,148 @@
...
+) => {
+  return (
+    <Polymorph as={as as PolymorphicAs} ref={forwardedRef} {...rest}>
+      {title}
+    </Polymorph>
+  );
+};
+```
+
+# Prior art
```
spark33, line 136: I'm wondering if you meant to say `artifacts` here, but maybe `References` is clearer?

packages/polymorphic/src/Polymorphic/Polymorph.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Polymorphic component that supports forwarded refs
+ *
+ * With the `as` prop, Polymorphic can dynamically render
+ * as any HTML element or React Component.
+ */
+export const Polymorph: PolymorphicComponentType =
+  React.forwardRef(BasePolymorph);
+Polymorph.displayName = 'Polymorph';
```
spark33, line 35: This living in the `Polymorphic` dir is interesting. Not quite sure if there's a clear resolution.

My gut reaction is that it warrants a separate dir, but it doesn't have any accompanying styles or types files. It would feel nice to me if this file and types imported across both Polymorphic and Polymorph lived in a top-level `types.ts` file, but if this current state makes sense to everyone I'm not gonna push on changing it.

spark33: Maybe it would have its own `.stories.tsx` file?

packages/polymorphic/README.md
shaneeza: Prior art is a little unclear

scripts/depcheck.ts
```diff
@@ -30,12 +30,15 @@ const fix: boolean = cli.opts()['fix'];
 const fixTS = cli.opts()['fixTsconfig'];
 const verbose = cli.opts()['verbose'];
 
+const testFilePatterns = [
+  // files matching these patterns will be ignored
+  /.*.spec.tsx?/,
+  /.*.story.tsx?/,
+  /.*.example.tsx?/,
+  /.*.testutils.tsx?/,
+];
```
spark33, line 39: `/.*stories.tsx` should be here. That should be the new standard we build towards. I mention this in the folder structure guidelines PR as well, but that's the standard that Storybook documents.

spark33 (approved):
Non-blocking comments!

packages/polymorphic/README.md
TheSonOfThomp: [Prior art](https://en.wikipedia.org/wiki/Prior_art) is a fairly common industry term afaik

brookescarlett: it is indeed, just dropped something in our team channel 👀

packages/polymorphic/README.md
TheSonOfThomp: That's a fair point. But I think the package name `polymorphic` feels more natural, especially since the `Polymorphic` function will likely be more commonly used than the standalone `Polymorph` component

packages/polymorphic/src/Polymorphic/Polymorph.tsx
TheSonOfThomp: This is a super weird package with respect to the directory structure guidelines. The component `Polymorph` is really secondary to the exported types, hooks and factory function.  And they're all very inter-related so a separate dir for just the component feels strange

TheSonOfThomp: Either that, or if the separation is between component vs package, we have `Polymorph/Polymorph.tsx` and have the rest of the files at the package root (since they're _technically_ not related to the component, but the package)

PR #1527, opened 2022-12-05
Fixes order of import statements across repo

brookescarlett:
- Adds linter rule that determines the order of import statements
- Applies the rule across the repo


packages/icon-button/src/IconButton.tsx
```diff
@@ -1,15 +1,16 @@
...
+import { isComponentGlyph } from '@leafygreen-ui/icon';
 import {
   useDarkMode,
   useUsingKeyboardContext,
 } from '@leafygreen-ui/leafygreen-provider';
-import { isComponentGlyph } from '@leafygreen-ui/icon';
-import { validateAriaLabelProps } from '@leafygreen-ui/a11y';
-import { AccessibleIconButtonProps, IconProps, Size } from './types';
+import { isComponentType } from '@leafygreen-ui/lib';
+
```
spark33, line 13: I don't know how I feel about these line breaks? I think it would be legible enough with the linted groupings.

packages/icon/src/generated/ArrowRight.tsx
```diff
@@ -2,11 +2,13 @@
  * This is a generated file. Do not modify it manually.
  *
  * @script ./node_modules/.bin/ts-node packages/icon/scripts/build.ts
- * @checksum 089a15a9efb4b3a77dda3129c14dd2ae
+ * @checksum 1f595d19570a615b5f15aa38aa9ed64a
```
spark33, line 5: I assume this is just regenerated on every build?

spark33 (changes requested):
Love the change. I'm personally not a fan of the line breaks, but I trust the package has done its job otherwise. A brief scroll through tells me it did, and that works for me

packages/icon/src/generated/ArrowRight.tsx
brookescarlett: yes

PR #1528, opened 2022-12-05
LG-2521 Search input `SearchResult` component. 

TheSonOfThomp:
## ✍️ Proposed changes

Storybook Demo: https://thesonofthomp.github.io/leafygreen-ui/?path=/story/components-searchinput--live-search

🎟 _Jira ticket:_ [LG-2521](https://jira.mongodb.org/browse/LG-2521)
Screen Shot 2023-01-17 at 6 16 29 PM

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/internal-input-option/src/InputOption.style.ts
```diff
@@ -0,0 +1,125 @@
...
+    position: absolute;
+    left: 0;
+    top: 50%;
+    width: ${wedgeWidth}px;
+    height: calc(100% - ${2 * wedgePaddingY}px);
+    background-color: rgba(255, 255, 255, 0);
+    border-radius: 0 6px 6px 0;
+    transform: scale3d(0, 0.3, 0) translateY(-50%);
+    transform-origin: 0%; // 0% since we use translateY
+    transition: 200ms ease-in-out;
```
brookescarlett, line 74: @audroue we use 150ms or 300ms in most places -- can we use one of those values here?

packages/search-input/src/SearchResult/SearchResult.types.ts
```diff
@@ -0,0 +1,23 @@
...
+import { DarkModeProps } from '@leafygreen-ui/lib';
+import { InputOptionProps } from '@leafygreen-ui/internal-input-option';
+
+export type SearchResultProps = AriaLabelProps &
+  DarkModeProps &
+  InputOptionProps & {
+    /**
+     * The value of the result
+     */
+    children: React.ReactNode;
```
brookescarlett, line 12: we should probably have an explicit value prop. no?

packages/search-input/src/SearchResult/SearchResult.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const { theme } = useDarkMode(rest.darkMode);
+
+    return (
+      <InputOption
+        as={as}
+        ref={ref}
+        className={cx(searchResultStyles, className)}
+        {...rest}
+      >
+        <div className={searchResultTitleStyle[theme]}>{children}</div>
```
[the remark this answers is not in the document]
TheSonOfThomp: The styles and children aren't totally consistent (sometimes there's icons/checkboxes etc)

packages/search-input/src/SearchResult/SearchResult.types.ts
TheSonOfThomp: We had a discussion on the RFC , but didn't land anywhere. Feel free to weigh in! https://docs.google.com/document/d/1E_ACiuznG0_oubMDkDyDgwwq8qCJjojTBl96HgIpU9s/edit?disco=AAAAhCNvbbU

brookescarlett:
Check with the designers, but I don't think that the search bar should change width when you start typing to expand to fit the "X." I think the component should remain the same width and the clear icon appears/disappears. This is jarring and a weird UX.

brookescarlett (changes requested):
Blocking feedback:
- Changing the state in storybook does not change the component itself.
- Should typing into the input option change what results are rendered in the menu? If so, that interaction isn't currently happening
- With our Menu component it renders as dark in light mode and light in dark mode, can we confirm that this is intentionally not following that pattern?
- Search Result Result doesn't look right in dark mode (I think you're mixing up disabled styles?)
- SearchResult Result not responding to the `showWedge` prop in storybook
- Search Result Menu story is entirely broken

TheSonOfThomp:
> * Changing the state in storybook does not change the component itself.
`state` only affects the loading indicator. 

> * Should typing into the input option change what results are rendered in the menu? If so, that interaction isn't currently happening
This interaction will need to be implemented by the consuming developer. See the `Live Search` story for an example of this

> * With our Menu component it renders as dark in light mode and light in dark mode, can we confirm that this is intentionally not following that pattern?
This is as intended ([Figma](https://www.figma.com/file/4h2VwjCuJUbeZ7hzD2J1rq/LeafyGreen-Design-System?node-id=14206%3A89891))

> * Search Result Result doesn't look right in dark mode (I think you're mixing up disabled styles?)

Screen Shot 2022-12-19 at 4 29 52 PM

Screen Shot 2022-12-19 at 4 29 41 PM

> * SearchResult Result not responding to the `showWedge` prop in storybook
Shouldn't be in SB. Removed

> * Search Result Menu story is entirely broken
Oops 🙃

TheSonOfThomp:
> Check with the designers, but I don't think that the search bar should change width when you start typing to expand to fit the "X." I think the component should remain the same width and the clear icon appears/disappears. This is jarring and a weird UX.

packages/hooks/src/useBackdropClick.ts
```diff
@@ -0,0 +1,66 @@
...
+      }
+    },
+    {
+      options: { capture: true },
+      enabled,
+    },
+  );
+
+  /**
+   * Returns whether the event target is a Combobox element
```
shaneeza, line 59: Combobox?

shaneeza:
Screen Shot 2022-12-20 at 11 56 42 AM

- [ ] the little active state looks tiny compared to whats in figma

packages/internal-input-option/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@leafygreen-ui/internal-input-option",
+  "version": "0.0.0",
```
shaneeza, line 3: 0.9.0?

packages/internal-input-option/src/InputOption/InputOption.story.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
shaneeza, line 1: I ran `yarn build` but i'm not seeing this in storybook. Am I missing a step?

packages/internal-input-option/src/InputOption/InputOption.style.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export const inputOptionStyles = css`
+  position: relative;
+  list-style: none;
+  outline: none;
+  border: unset;
+  margin: 0;
+  text-align: left;
+  text-decoration: none;
+
```
shaneeza, line 19: maybe we can add a section about grouping CSS in the style guide?

packages/lib/src/validateChildren.tsx
```diff
@@ -0,0 +1,27 @@
+import isUndefined from 'lodash/isUndefined';
+import React from 'react';
+import { consoleOnce, isComponentType } from '.';
+
+export const validateChildren = (
```
shaneeza, line 5: The name is self explanatory but can you can some notes on what this is doing?

packages/search-input/src/SearchInput/SearchInput.story.tsx
```diff
@@ -0,0 +1,129 @@
...
+    state: { control: 'select', options: Object.values(State) },
+    id: { control: 'string' },
+    placeholder: { control: 'string' },
+    value: { control: 'string' },
+    disabled: { control: 'boolean' },
+    ref: { control: 'none' },
+  },
+  parameters: {
+    controls: {
+      exclude: [...storybookExcludedControlParams, 'value', 'id', 'showWedge'],
```
shaneeza, line 30: nice!

packages/hooks/src/useBackdropClick.ts
TheSonOfThomp: 🙃

packages/internal-input-option/src/InputOption/InputOption.style.ts
TheSonOfThomp: Yeah, I like that. This is my preferred one https://github.com/ream88/stylelint-config-idiomatic-order

packages/search-input/src/SearchInput/SearchInput.spec.tsx
```diff
@@ -0,0 +1,436 @@
...
+describe('packages/search-input', () => {
+  describe('a11y', () => {
+    test('does not have basic accessibility issues', async () => {
+      const { container } = renderSearchInput();
+      const results = await axe(container);
+      expect(results).toHaveNoViolations();
+    });
+  });
+
+  describe('Basic rendering', () => {
```
brookescarlett, line 48: We should add tests for the `state` prop, as well as `onChange` properly firing. I believe you test `onSubmit` below but that might be a different component's callback

packages/hooks/src/useBackdropClick.ts
```diff
@@ -0,0 +1,66 @@
+import useEventListener from './useEventListener';
+
+export function useBackdropClick(
+  callback: Function,
+  refEl: React.RefObject<HTMLElement>,
+  enabled?: boolean,
+): void;
+export function useBackdropClick(
+  callback: Function,
+  refEls: Array<React.RefObject<HTMLElement>>,
```
brookescarlett, line 10: is there a reason we can't type this as `refElorEls: React.RefObject<HTMLElement> | Array<React.RefObject<HTMLElement>>` instead of overloading the type?

packages/hooks/src/useControlledValue.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A hook that enables a component to be both controlled or uncontrolled.
+ *
+ * Returns a {@link ControlledValueReturnObject} with the controlled or uncontrolled `value`,
+ * `onChange` & `onClear` handlers, a `setInternalValue` setter, and a boolean `isControlled`
+ */
+export const useControlledValue = <T extends string>(
+  controlledValue?: T,
+  changeHandler?: ChangeEventHandler<any>,
+  // _onClear?: ReactEventHandler<any>,
```
brookescarlett, line 27: rm?

packages/hooks/src/useControlledValue.ts
```diff
@@ -0,0 +1,56 @@
+import { ChangeEventHandler, useEffect, useState } from 'react';
+import isUndefined from 'lodash/isUndefined';
+
+interface ControlledValueReturnObject<T extends string> {
+  /** Whether the value is controlled */
+  isControlled: boolean;
```
brookescarlett, line 6: nit line break between prop definitions

packages/hooks/src/useBackdropClick.ts
TheSonOfThomp: Yeah, got a little too fancy here. This doesn't need to be overloaded

packages/search-input/src/SearchInput/SearchInput.spec.tsx
TheSonOfThomp: Added tests for `state`

I have tests for `onChange` when a character is typed and when the clear button is pressed. We also check that an Option's `onClick` handler is fired on enter key press

shaneeza:
- [ ] Screen Shot 2023-01-18 at 12 26 35 PM
The sides are clipped when the size is `small`

packages/input-option/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@leafygreen-ui/input-option",
+  "version": "0.0.0",
```
shaneeza, line 3: Shouldn't this be `0.9.0` so that It can be bumped to `1.0.0`? I think thats why `Polymorphic` did not initially get bumped in the version PR.

packages/search-input/src/SearchInput/SearchInput.types.ts
```diff
@@ -0,0 +1,67 @@
...
+  extends Omit<ComponentPropsWithoutRef<'form'>, 'onChange'> {
+  /**
+   * The current state of the SearchInput. This can be none, or loading.
+   */
+  state?: State;
+
+  /**
+   * Determines whether the component appears in dark theme.
+   */
+  darkMode?: boolean;
```
shaneeza, line 34: You can extend `DarkModeProps` from `lib`!

packages/input-option/package.json
TheSonOfThomp: I think as long as it's marked in the changeset as a major change, it should bump to 1.0.0 regardless

packages/combobox/src/Combobox.tsx
```diff
@@ -1137,6 +1137,7 @@ export function Combobox<M extends boolean>({
    *
    */
 
+  // TODO: Replace this with `useBackdropClick`
```
spark33, line 1140: Should we add this in now since the `useBackdropClick` PR has gone in now?

packages/input-option/src/InputOption/InputOption.story.tsx
```diff
@@ -0,0 +1,48 @@
...
+    },
+    ...storybookArgTypes,
+  },
+};
+
+const Template = (props: InputOptionProps) => (
+  <InputOption {...props}>Some text</InputOption>
+);
+
+export const Generic = Template.bind({});
```
spark33, line 48: nit: I think we're using the name `Basic` in other components

packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -0,0 +1,132 @@
...
+  outline: none;
+  border: unset;
+  margin: 0;
+  text-align: left;
+  text-decoration: none;
+
+  font-size: ${typeScales.body1.fontSize}px;
+  line-height: ${typeScales.body1.lineHeight}px;
+  font-family: ${fontFamilies.default};
+  padding: ${spacing[2]}px ${spacing[2] + spacing[1]}px;
```
spark33, line 23: `spacing[2] + spacing[1]` feels kind of awkward. I'm wondering if we should evaluate having our spacing units just jump in increments of 4px if that's our base.

packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -0,0 +1,132 @@
...
+  `,
+  [Theme.Dark]: css`
+    color: ${palette.white};
+    background-color: ${palette.gray.dark3};
+  `,
+};
+
+export const inputOptionHoverStyles: Record<Theme, string> = {
+  [Theme.Light]: css`
+    cursor: pointer;
```
spark33, line 47: nit: this can go in our base styles

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -0,0 +1,54 @@
...
+  showWedge?: boolean;
+
+  /**
+   * Whether the component is active, regardless of keyboard navigation
+   */
+  active?: boolean;
+
+  /**
+   * Determines whether to show hover, focus and active styles
+   *
```
spark33, line 46: nit: other TSDocs don't have this line break

packages/lib/src/validateChildren.tsx
```diff
@@ -0,0 +1,41 @@
...
+  /** Any React children */
+  children: React.ReactNode,
+  /** Array of `displayNames` for valid components */
+  validTypes: Array<string>,
+): Array<ReactElement> | undefined => {
+  const validChildren = React.Children.map(children, child => {
+    if (validTypes.some(type => isComponentType(child, type))) {
+      return child as React.ReactElement;
+    }
+  })?.filter(child => !isUndefined(child));
```
spark33, line 27: nit: looks like this is running two loops (`map` and `filter`) when it can be run in one?

packages/search-input/src/SearchInput/SearchInput.story.tsx
```diff
@@ -0,0 +1,136 @@
...
+    `}
+    {...props}
+  />
+);
+
+export const WithResults: ComponentStory<typeof SearchInput> = props => (
+  <SearchInput
+    className={css`
+      width: 200px;
+    `}
```
spark33, line 55: nit: we can use `<Basic>` here so we don't have to duplicate the inline styles

packages/search-input/src/SearchInput/SearchInput.story.tsx
```diff
@@ -0,0 +1,136 @@
...
+      }}
+      description="This is a description"
+    >
+      Apple
+    </SearchResult>
+    <SearchResult>Banana</SearchResult>
+    <SearchResult as="a" href="#" description="This is a link">
+      Carrot
+    </SearchResult>
+    <SearchResult description="This is a very very long description. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor.">
```
spark33, line 73: This might be a nice opportunity to do what I demoed with the new Table component - having a separate story file to allow the user to play with the subcomponent's props using knobs

packages/search-input/src/SearchResult/SearchResult.tsx
```diff
@@ -0,0 +1,72 @@
...
+      className,
+      darkMode,
+      ...rest
+    },
+    ref,
+  ) => {
+    const { theme } = useDarkMode(darkMode);
+    const textContent = getNodeTextContent(children);
+    const ariaLabel =
+      rest['aria-label'] ?? rest['aria-labelledby'] ? '' : textContent;
```
spark33, line 41: Shouldn't this be `rest['aria-label'] ?? (rest['aria-labelledby'] ??  textContent)`

packages/search-input/src/SearchResult/SearchResult.types.ts
spark33: nit: we could use `PropsWithChildren<...>` since we don't need special TSDocs for this prop

packages/search-input/src/SearchResultsMenu/LoadingOption/LoadingOption.styles.ts
```diff
@@ -0,0 +1,24 @@
...
+  to {
+    transform: rotate(360deg);
+  }
+`;
+
+export const loadingOptionIcon = css`
+  height: 16px;
+  width: 16px;
+  animation: ${loadingSpinner} 1.5s linear infinite;
+`;
```
spark33, line 24: A `TODO` comment or ticket to make sure we update this to the new loading animation later might be nice

spark33 (changes requested):

packages/lib/src/validateChildren.tsx
TheSonOfThomp: I guess a `reduce` loop would work here, yeah

packages/combobox/src/Combobox.tsx
TheSonOfThomp: I feel like this should be in a separate body of work

packages/input-option/src/InputOption/InputOption.style.ts
TheSonOfThomp: Definitely agree that `12px` should be in our spacing tokens, since it's used so often

packages/search-input/src/SearchInput/SearchInput.story.tsx
TheSonOfThomp: Technically correct, but I feel like this makes `children` a bit more complicated

packages/search-input/src/SearchInput/SearchInput.story.tsx
TheSonOfThomp: `SearchResults/SearchResults.story.ts`

packages/search-input/src/SearchResult/SearchResult.tsx
TheSonOfThomp: No, the logic is:

> If `aria-label` exists, use that; if `aria-labelledby` exists, then we have no label; if neither exist we set the label to `textContent`

packages/search-input/src/SearchResult/SearchResult.types.ts
TheSonOfThomp: `PropsWithChildren` sets `children` to optional. I feel like trying to get around that is more complex than declaring it here.
(That said, we could update `InputOptionProps` to use that utility type)

packages/search-input/src/SearchResultsMenu/LoadingOption/LoadingOption.styles.ts
TheSonOfThomp: I feel like we'll do an audit with the loading animation project coming up

packages/input-option/src/InputOption/InputOption.style.ts
TheSonOfThomp: Filed a ticket to address this https://jira.mongodb.org/browse/LG-2857

packages/search-input/src/SearchInput/SearchInput.spec.tsx
```diff
@@ -0,0 +1,470 @@
...
+            ...defaultProps,
+          });
+          userEvent.tab();
+          expect(inputEl).toHaveFocus();
+          userEvent.tab();
+          expect(inputEl).not.toHaveFocus();
+        });
+
+        // eslint-disable-next-line jest/no-disabled-tests
+        test.skip('Closes menu when tabbing away', async () => {
```
brookescarlett, line 298: why test.skip?

.changeset/lucky-garlics-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/input-option': major
```
brookescarlett, line 2: need changesets for other impacted components

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,377 @@
...
+        child: React.ReactNode,
+      ): JSX.Element | undefined => {
+        if (isComponentType(child, 'SearchResult')) {
+          resultsCount += 1;
+          const index = resultsCount - 1;
+          return React.cloneElement(child, {
+            ...child.props,
+            id: `result-${index}`,
+            ref: child.props.ref ?? resultRefs?.(`${index}`),
+            focused: index === highlightIndex,
```
brookescarlett, line 116: should we be passing an onClick here, such that when an item is clicked the menu closes?

packages/search-input/src/SearchInput/SearchInput.types.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * The current state of the SearchInput. This can be none, or loading.
+   */
+  state?: State;
+
+  /**
+   * Determines the font size and padding.
+   */
+
+  sizeVariant?: SizeVariant;
```
brookescarlett, line 37: why is this `sizeVariant` when all of the other props with a similar behavior are `size`?

packages/a11y/src/AriaLabelProps.ts
```diff
@@ -0,0 +1,9 @@
+export type AriaLabelProps =
```
brookescarlett, line 1: nice

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -0,0 +1,65 @@
...
+      active,
+      darkMode: darkModeProp,
+      showWedge = true,
+      isInteractive = true,
+      className,
+      ...rest
+    },
+    ref,
+  ) => {
+    const { Component } = usePolymorphic(as);
```
brookescarlett, line 37: For my own edification.. what's the difference between using the ref passed to the function versus the one returned from the hook?

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -0,0 +1,65 @@
...
+  inputOptionActiveStyles,
+  inputOptionDisabledStyles,
+  inputOptionHoverStyles,
+  inputOptionStyles,
+  inputOptionThemeStyles,
+  inputOptionWedge,
+} from './InputOption.style';
+import { InputOptionProps } from './InputOption.types';
+
+export const InputOption = Polymorphic<InputOptionProps>(
```
brookescarlett, line 21: Should this component respond to the "size" prop as well?

packages/search-input/src/SearchInput/SearchInput.styles.ts
```diff
@@ -9,48 +9,129 @@ import {
...
 
 export const baseInputStyle = css`
   font-size: inherit;
   line-height: inherit;
+  color: inherit;
+  background-color: inherit;
   font-family: ${fontFamilies.default};
   width: 100%;
-  height: 36px;
+  height: 1.5em;
```
brookescarlett, line 129: Should we be preferring `em`? Should this be in our style guide?

packages/search-input/src/SearchInput/SearchInput.styles.ts
```diff
@@ -81,23 +162,19 @@ export const baseInputStyle = css`
...
+      -webkit-text-fill-color: inherit;
 
       &:not(:disabled) {
         box-shadow: ${autofillShadowOverride(palette.white)};
 
         &:focus {
           box-shadow: ${autofillShadowOverride(palette.white)},
             ${focusRing.light.input};
-          border-color: ${palette.white};
+          /* border-color: ${palette.white}; */
```
brookescarlett, line 177: why commented out?

packages/search-input/src/SearchInput/SearchInput.styles.ts
```diff
@@ -201,42 +252,13 @@ export const inputThemeStyle: Record<Theme, string> = {
...
-  `,
-  [SizeVariant.Default]: css`
-    height: 36px;
-    padding-left: 36px;
-  `,
-  [SizeVariant.Large]: css`
-    height: 48px;
-    padding-left: 40px;
-  `,
+  [SizeVariant.Small]: css``,
```
brookescarlett, line 256: are we rming sizes?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,377 @@
...
+          if (nestedChildren && nestedChildren.length > 0) {
+            return React.cloneElement(child, {
+              ...child.props,
+              children: nestedChildren,
+            });
+          }
+        }
+      };
+
+      const updatedChildren = React.Children.map(children, processChild);
```
brookescarlett, line 133: should we memoize this value?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,377 @@
...
+
+      return {
+        resultsCount,
+        updatedChildren,
+      };
+    };
+
+    const { updatedChildren, resultsCount } = processChildren();
```
brookescarlett, line 141: memo?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,377 @@
...
+      trackFocusedElement(target);
+      handleOpenMenuAction(e);
+    };
+
+    const handleClearButton: MouseEventHandler<HTMLButtonElement> = e => {
+      // We convert the click event into a Change event,
+      // Update the target & value,
+      // Then fire any `onChange` handler
+      const changeEvent: ChangeEvent<any> = {
+        ...cloneDeep(e),
```
brookescarlett, line 216: Why do we need to clone the event?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,377 @@
...
+          highlight={highlightIndex}
+          resultDynamicRefs={resultRefs}
+        >
+          <form
+            role="search"
+            ref={formRef}
+            className={cx(formStyle, className)}
+            onSubmit={handleSubmit}
+            onBlur={handleBlur}
+            {...rest}
```
brookescarlett, line 300: should this have `rest[aria-label]` and `rest[aria-labelledby]` spread here?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,377 @@
...
+              aria-labelledby={rest['aria-labelledby']}
+            >
+              <MagnifyingGlass
+                className={cx(
+                  searchIconStyle,
+                  searchIconThemeStyle[theme],
+                  searchIconSizeStyle[sizeVariant],
+                  { [searchIconDisabledStyle[theme]]: disabled },
+                )}
+                aria-label="Search Icon"
```
brookescarlett, line 333: nit are aria-label and role="presentation" oxy moronical? as in, i don't think a screen reader will announce the icon if it has role="presentation" set

sooachung:
- [ ] Darkmode hover should be Dark gray 2 (lighter than the menu)
Screen Shot 2023-02-07 at 11 11 47 AM
809-68887fb1-3b5c-4052-ac47-0cc56fc8d7fc.png">

.changeset/lucky-garlics-hear.md
TheSonOfThomp: This is only being used in SearchInput so far. I'll open another PR that refactors Combo/Select

packages/search-input/src/SearchInput/SearchInput.tsx
TheSonOfThomp: Any `onClick` handlers the user has added would be included in `child.props`. Also, since in most cases the search results will navigate to a new page, it's not absolutely necessary to force the menu to close

packages/search-input/src/SearchInput/SearchInput.types.ts
TheSonOfThomp: This was ported over from `TextInput`. @shaneeza Do you remember why we used this prop name?

packages/input-option/src/InputOption/InputOption.tsx
TheSonOfThomp: The hook creates a properly typed ref for you to use instead of struggling with typing `useRef`. If you're forwarding a ref then you can ignore the one in the hook

packages/input-option/src/InputOption/InputOption.tsx
TheSonOfThomp: Not necessarily, since not implementing components will have the same `Size` values (at least not until we have a unified token). It could have a `baseFontSize` prop though

packages/search-input/src/SearchInput/SearchInput.styles.ts
TheSonOfThomp: I would advocate for `em` for most things, yeah, since it makes resizing easier

packages/search-input/src/SearchInput/SearchInput.tsx
TheSonOfThomp: In order to get the proper types & event properties. We can't call `handleChange(e)` since `e` is a `MouseEvent`

TheSonOfThomp: Actually, we don't need to clone, we can just cast & change properties. Fixed this

packages/search-input/src/SearchInput/SearchInput.tsx
TheSonOfThomp: Good catch, I'll pull it out

packages/search-input/src/SearchInput/SearchInput.tsx
TheSonOfThomp: That's a good point. Technically `role` only removes the semantic meaning, but when used on an `svg` it will completely ignore it. rming label

TheSonOfThomp:
> * [ ]  Darkmode hover should be Dark gray 2 (lighter than the unhovered menu items)
> 
> Screen Shot 2023-02-07 at 11 11 47 AM

`Combobox` and `Select` both use a darker menu item on hover. Are we sure we want to have `Search` be different? 

> 
> * [ ]  We need to have a variant for a group label and description.
> 
> Screen Shot 2023-02-07 at 11 14 07 AM

It's there, just didn't have that explicitly in a story. Added now 
Screen Shot 2023-02-07 at 1 10 54 PM

packages/search-input/src/SearchInput/SearchInput.types.ts
shaneeza: yeah i initially added `sizeVariant` cause `input` has a `size` attribute but looking back now I could have just omitted `size`.

TheSonOfThomp: This component inherits from `form` anyway, so doesn't have a native `size` attr. 
In general though, I'm in favor of overriding the `size` attribute. I'd be surprised if anyone uses it, and it will be ignored in favor of any explicit css `width` rule

sooachung:
> `Combobox` and `Select` both use a darker menu item on hover. Are we sure we want to have `Search` be different?
Oh in this case I think we can keep what we have here then. I changed the figma file. 

Everything else looks good!

.changeset/young-cups-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/search-input': minor
```
shaneeza, line 2: Should this be major?

.changeset/lucky-garlics-hear.md
brookescarlett: Looks like we're impacting files in lib, palette, polymorphic too

packages/palette/tsconfig.json
```diff
@@ -7,5 +7,6 @@
   },
   "include": [
     "src/**/*"
-  ]
+  ],
+  "references": []
```
shaneeza, line 11: What is this for?

shaneeza:
It looks like if you tab to the input to focus the search results container will automatically open; I remember combobox did a similar thing but it had to be refactored to only open on enter/space. Should that logic also apply here?

.changeset/young-cups-dream.md
TheSonOfThomp: The only thing making it a major change technically is the renaming of `sizeVariant`. Everything else is backwards compatible

packages/palette/tsconfig.json
TheSonOfThomp: I think this was done by depchek. will remove

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,408 @@
...
+    // The index of the currently highlighted result option
+    const [highlightIndex, setHighlightIndex] = useState<number>(0);
+
+    const closeMenu = () => setOpen(false);
+    const openMenu = () => setOpen(true);
+
+    const formRef = useRef<HTMLFormElement>(null);
+    const searchBoxRef = useRef<HTMLDivElement>(null);
+    const menuRef = useRef<HTMLUListElement>(null);
+    const inputRef = useForwardedRef(forwardRef, null);
```
shaneeza, line 97: does useForwardRef allow you to both use the ref and pass it to the forward ref?

packages/input-option/src/InputOption/InputOption.story.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+import {
+  storybookArgTypes,
+  // @ts-ignore
```
brookescarlett, line 5: why ts-ignore?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,408 @@
...
+                  aria-label="Clear search"
+                  onClick={handleClearButton}
+                  className={clearButtonSizeStyle[size]}
+                >
+                  <XIcon />
+                </IconButton>
+              )}
+            </div>
+            {withTypeAhead && (
+              <SearchResultsMenu
```
shaneeza, line 393: Should we add `aria-describedby` here?

packages/search-input/src/SearchResult/SearchResult.tsx
shaneeza: if i run this in the console `'hi' ?? undefined ? '' : 'hello'` `''` is printed. Is that correct?

.changeset/young-cups-dream.md
brookescarlett: I'm inclined to say we move this to size and make it a major change. MAYBE worth looking into how widely this component is being used currently to know how heavy of a lift this would be as a breaking change. I do think we have a lot of standardization efforts coming up around font size and form size compatibility and this feels like a step in that direction.

packages/input-option/src/InputOption/InputOption.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { render } from '@testing-library/react';
+
+import { InputOption } from '..';
+
+describe('packages/input-option', () => {
+  test('isPolymorphic', () => {
+    expect(InputOption).toBePolymorphic();
+  });
+
+  test('renders', () => {
```
brookescarlett, line 11: the only other prop we could test here is disabled -- should we add a quick test?

packages/input-option/src/InputOption/InputOption.story.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
```
brookescarlett, line 2: Process question -- as we release new components, how are we ensuring that this is compatible with .design?

packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -0,0 +1,131 @@
...
+  font-size: ${typeScales.body1.fontSize}px;
+  line-height: ${typeScales.body1.lineHeight}px;
+  font-family: ${fontFamilies.default};
+  padding: ${spacing[2]}px ${spacing[2] + spacing[1]}px;
+
+  transition: background-color ${transitionDuration.default}ms ease-in-out;
+
+  &:focus,
+  &:focus-visible {
+    outline: none;
```
brookescarlett, line 30: do we want to explicitly unset this when focus-visible?

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -0,0 +1,65 @@
...
+  (
+    {
+      as = 'li' as PolymorphicAs,
+      children,
+      disabled,
+      focused,
```
brookescarlett, line 27: Are we requiring that we set focused in order to know if an option is focused? There's no logic around this internal to the component?

packages/input-option/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export {
+  type BaseInputOptionProps,
```
brookescarlett, line 2: TIL you can combine type / normal export statements

packages/search-input/src/SearchInput/SearchInput.spec.tsx
```diff
@@ -0,0 +1,509 @@
...
+      expect(queryByRole('button')).not.toBeInTheDocument();
+    });
+
+    test('clear button is rendered when there is text', () => {
+      const { queryByRole, inputEl } = renderSearchInput();
+      userEvent.type(inputEl, 'abc');
+      expect(queryByRole('button')).toBeInTheDocument();
+    });
+
+    describe('when the "size" is not "large"', () => {
```
brookescarlett, line 79: nit but why keep these in if they aren't doing anything?

packages/search-input/src/SearchResultGroup/SearchResultGroup.types.ts
```diff
@@ -0,0 +1,13 @@
+import { PropsWithChildren, ReactNode } from 'react';
+
+export type SearchResultGroupProps = PropsWithChildren<{
```
shaneeza, line 3: Do you have to use `PropsWithChildren` since `children` is required below?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,408 @@
...
+      const target =
+        e.target !== clearButtonRef.current
+          ? inputRef.current ?? (e.target as HTMLElement)
+          : clearButtonRef.current;
+      target.focus();
+      trackFocusedElement(target);
+      handleOpenMenuAction(e);
+    };
+
+    const handleClearButton: MouseEventHandler<HTMLButtonElement> = e => {
```
brookescarlett, line 251: I'm not seeing a clear button rendered in storybook ATM. When you click an item, the menu closes but the search bar isn't updated

brookescarlett: Selecting an item is only partially working at the moment

packages/search-input/src/SearchInput/SearchInput.story.tsx
```diff
@@ -0,0 +1,258 @@
...
+    description:
+      'The edible starchy tuber of a climbing plant that is widely grown in tropical and subtropical countries',
+  },
+  {
+    name: 'zucchini',
+    description: 'A green variety of smooth-skinned summer squash',
+  },
+];
+
+export const LiveSearch: ComponentStory<typeof SearchInput> = args => {
```
shaneeza, line 214: Screen Shot 2023-02-08 at 3 52 41 PM

In the live search example, if i try to click on an option that is partially cut off like `eggplant` the text below the input does not update.

brookescarlett:
Noticing a lot of bugs at the moment:
1. Clicking on an item is not reliably setting the actively selected item (the Search bar is not being updated, and in the Live Search story I'm not seeing the actively selected element changing below the component itself)
2. Starting from the Search bar i'm not able to access the Menu via keyboard (sometimes I can if I keep clicking the down arrow a bunch of times, but not reliable).

packages/search-input/src/SearchInput/SearchInput.story.tsx
shaneeza, line 214: When tabbing through menu items, the menu does not scroll to the focus item if its below the fold

packages/input-option/src/InputOption/InputOption.story.tsx
spark33: We import directly from Storybook so I feel like it should mostly be okay, but we don't have anything set up to test with the website overall. We'd need some draft state of these components to be consumed from our staging environment if we want that.

Alpha release from a staging branch that's consumed in our staging environment maybe?

shaneeza:
- [ ] The input is not focusable when disabled. For passwordInput I used `readOnly` instead of `disabled` so that you can't interact with the input but you can focus on it.

packages/search-input/src/SearchInput/SearchInput.tsx
TheSonOfThomp: Exactly, yeah

packages/input-option/src/InputOption/InputOption.story.tsx
TheSonOfThomp: was getting some build errors with the old lib version, but it's gone now

packages/search-input/src/SearchInput/SearchInput.tsx
TheSonOfThomp: What element would this be described by? 
If the user passes a `describedby` prop it'll be spread onto the `form` element which I believe is fine

packages/input-option/src/InputOption/InputOption.spec.tsx
TheSonOfThomp: What would we be testing for though? There are no baked-in functional impacts of the `disabled` prop, only visual

packages/input-option/src/InputOption/InputOption.story.tsx
TheSonOfThomp: Some sort of pre-release, or `yarn link`ing this repo to the design repo.

I agree though, we do need some explicit guidelines on how to prep a story file for use on the website

packages/input-option/src/InputOption/InputOption.style.ts
TheSonOfThomp: Yes, otherwise if `as="button"` you'll get a default focus ring

packages/input-option/src/InputOption/InputOption.tsx
TheSonOfThomp: No this is a visual component only. The implementing component (Select, Combo, Search) will have it's own logic to determine which option is focused

TheSonOfThomp: Worth noting that this is different than the native `:focus` state. It's one of the parameters that determines the blue active state

Screen Shot 2023-02-08 at 6 42 00 PM

TheSonOfThomp: There are cases (like in Combo) where the `active` (or selected) elements, are different than the `focused` elements. Open to changing this terminology though

packages/search-input/src/SearchInput/SearchInput.spec.tsx
TheSonOfThomp: They're still testing for font size. I can remove them though since they'll be replaced by Chromatic soon enough

packages/search-input/src/SearchResultGroup/SearchResultGroup.types.ts
TheSonOfThomp: Nope, good catch 👍

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,408 @@
...
+      };
+    }, [children, highlightIndex, inputRef, resultRefs, setInputValue]);
+
+    const { updatedChildren, resultsCount } = useMemo(
+      () => processChildren(),
+      [processChildren],
+    );
+
+    type Direction = 'next' | 'prev' | 'first' | 'last';
+    const updateHighlight = (direction: Direction) => {
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah good catch. This also doesn't stop at 0/max, or follow the highlighted item

packages/input-option/src/InputOption/InputOption.tsx
TheSonOfThomp: Maybe `selected` & `highlighted`

packages/search-input/src/SearchInput/SearchInput.tsx
TheSonOfThomp: Fixed the underlying issue here. The clear button will only appear if there's text in the input

packages/search-input/src/SearchInput/SearchInput.story.tsx
TheSonOfThomp: This should be fixed now

TheSonOfThomp:
> Noticing a lot of bugs at the moment:
> 
> 1. Clicking on an item is not reliably setting the actively selected item (the Search bar is not being updated, and in the Live Search story I'm not seeing the actively selected element changing below the component itself)
> 2. Starting from the Search bar i'm not able to access the Menu via keyboard (sometimes I can if I keep clicking the down arrow a bunch of times, but not reliable).

1. Resolved this 👍 I was closing the menu on input `blur` before
2. Should be more reliable now

TheSonOfThomp:
> * [ ]  The input is not focusable when disabled. For passwordInput I used `readOnly` instead of `disabled` so that you can't interact with the input but you can focus on it.

Oh interesting! As in `readonly={disabled}` ?

shaneeza:
> > * [ ]   The input is not focusable when disabled. For passwordInput I used `readOnly` instead of `disabled` so that you can't interact with the input but you can focus on it.
> 
> Oh interesting! As in `readOnly={disabled}` ?

yeah! https://github.com/mongodb/leafygreen-ui/pull/1601/files#diff-5020f0b1dac36cd9fd23aa04fc21f50b99522abd9875bb3071b95bc002b0913bR137

packages/search-input/src/SearchInput/SearchInput.tsx
shaneeza: oh sorry, I meant to add this comment on the `input` above. I'm not sure if this applies since its a search input but if `<SearchResultsMenu>` had an id it could be passed to the `aria-describedby` on the input.

.changeset/lucky-garlics-hear.md
TheSonOfThomp: Yep. Those all have separate changeset messages

packages/search-input/src/SearchResult/SearchResult.tsx
TheSonOfThomp: Ah. I think we need parens around the ternary: `'hi' ?? (undefined ? '' : 'hello')`. Not 100% sure why

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,436 @@
...
+    };
+
+    const handleSearchBoxClick: MouseEventHandler<HTMLDivElement> = e => {
+      if (e.target !== clearButtonRef.current) {
+        // Don't open the menu if it was the clear button that was clicked.
+        // Could use `e.stopPropagation` in clear button handler instead,
+        // but users may still want to handle that click event
+        handleOpenMenuAction(e);
+      }
+    };
```
TheSonOfThomp, line 264: @brookescarlett Here I'm just checking the event target instead of stopping propagation

brookescarlett:
I'm not seeing the storybook respond to the "state" prop on the Basic SearchInput story

brookescarlett:
Clicking on the "clear" icon button opens the menu

brookescarlett:
- In the LiveSearch example.. if you make a selection and then reopen the menu the dropdown says "No results found" which isn't true since something is active.

TheSonOfThomp:
> Clicking on the "clear" icon button opens the menu

Weird... I wrote a bad check, so it only worked if the target was the button, not the svg. Reverted to `e.stopPropagation`

TheSonOfThomp:
> * In the LiveSearch example.. if you make a selection and then reopen the menu the dropdown says "No results found" which isn't true since something is active.
> * Typing in LiveSearch is case sensitive relative to value...this should not be the case.

These are just implementation bugs in the story, not the component itself. Fixed

packages/combobox/src/Combobox.spec.tsx
```diff
@@ -1033,6 +1034,10 @@ describe('packages/combobox', () => {
           expect(highlight).toHaveTextContent('Banana');
         });
 
+        test.todo('Down arrow cycles highlight to top');
```
brookescarlett, line 1037: should we add these in?

packages/hooks/src/useControlledValue/useControlledValue.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { ChangeEvent } from 'react';
+import { act } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
+
+// const { act } = TestRenderer;
```
brookescarlett, line 5: rm?

packages/input-option/src/InputOption/InputOption.spec.tsx
brookescarlett: we don't need to set aria-disabled on the container?

packages/input-option/src/InputOption/InputOption.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+    test('Aria labels', () => {
+      /** @ts-expect-error */
+      render(<InputOption />);
+      render(<InputOption aria-label="Some Label" />);
+      render(<InputOption aria-labelledby="some-id" />);
+    });
+
+    // eslint-disable-next-line jest/expect-expect
+    test('as prop', () => {
+      /** @ts-expect-error */
```
brookescarlett, line 28: can you add comments as to why we are expecting the below to break?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,433 @@
...
+          inputRef.current.value = newVal;
+          const nativeChangeEvent = new Event('change', {
+            cancelable: true,
+            bubbles: true,
+          });
+          const syntheticChangeEvent = createSyntheticEvent(
+            nativeChangeEvent,
+            inputRef.current,
+          );
+          handleChange(syntheticChangeEvent);
```
brookescarlett, line 125: why do this instead of just using the event itself?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,433 @@
...
+
+          const onElementClick: MouseEventHandler = e => {
+            child.props.onClick?.(e); // call the child's onClick handler
+
+            changeInputValue(textValue);
+
+            const wasClickedWithMouse = e.detail >= 1;
+
+            if (wasClickedWithMouse && formRef.current && inputRef.current) {
+              // Selecting an option fires the `submit` event
```
brookescarlett, line 156: Why do we do this?

TheSonOfThomp: Do you mean the submit event or the mouse check?

We need the submit event so that users can listen for `onSubmit` instead of `onClick` for every result

We need to do the check since the `Enter` key already triggers a `click` event on the result and `submit` on the form

packages/search-input/src/SearchInput/SearchInput.tsx
TheSonOfThomp: Since `changeInputValue` can be called programatically we need to trigger the `change` event with an event. The handler expects a `React.SyntheticEvent`, and not a native event.

packages/combobox/src/Combobox.spec.tsx
TheSonOfThomp: I wanted to make minimal changes to Combobox in this PR. I'll add these in https://github.com/mongodb/leafygreen-ui/pull/1619

packages/input-option/src/InputOption/InputOption.spec.tsx
TheSonOfThomp: I don't think so. The input itself has the `readOnly` attribute but should still be focusable.

Also, trying to add this to the container gives: 
> The attribute aria-disabled is not supported by the role search

shaneeza:
- [ ] Might want to add the correct cursor when it's disabled but glad to see that readonly worked out!

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,446 @@
...
+      const processChild = (
+        child: React.ReactNode,
+      ): JSX.Element | undefined => {
+        if (isComponentType(child, 'SearchResult')) {
+          resultsCount += 1;
+          const index = resultsCount - 1;
+
+          const textValue = getNodeTextContent(child);
+
+          const onElementClick: MouseEventHandler = e => {
```
brookescarlett, line 152: are we no longer manually closing the menu on child element click?

brookescarlett (approved):

PR #1529, opened 2022-12-06
Reorganize lib

TheSonOfThomp:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/lib/src/utils/createDataProp.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
brookescarlett, line 1: I think this is safe to rm .. it's not used anywhere else in the leafy codebase. What do you think @DesignerDave

TheSonOfThomp: Removing this would technically make this a major change instead of just a patch. Not saying we can't, but that's a consideration

DesignerDave: We gotta remove it at some point 🤷‍♂️

spark33 (changes requested):

packages/lib/src/utils/consoleOnce.ts
```diff
@@ -0,0 +1,7 @@
+import once from 'lodash/once';
+
+export const consoleOnce = {
+  error: once(console.error),
+  warn: once(console.warn),
+  log: once(console.log),
+};
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah, that's fair, there's just a lot of small files

packages/lib/src/aria.ts
```diff
@@ -0,0 +1,22 @@
...
+  Step: 'step', // current value in a step indicator
+  Location: 'location', // current value in a chart or other visual flow
+  Date: 'date', // current value in a calendar or date picker
+  Time: 'time', // current value in a timetable or time picker
+  True: 'true', // (fallback) current value in any set of options/elements
+  Unset: 'false', // equivalent to omitting the aria-current attribute
+} as const;
+
+export type AriaCurrentValue =
+  typeof AriaCurrentValue[keyof typeof AriaCurrentValue];
```
spark33, line 22: Should we just call this file AriaCurrentValue?

packages/lib/src/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+  ArrowLeft: 37,
+  ArrowRight: 39,
+  Backspace: 8,
+  BracketLeft: 91,
+  Delete: 46,
+  Enter: 13,
+  Escape: 27,
+  Space: 32,
+  Tab: 9,
+} as const;
```
spark33, line 14: I'm a fan of calling this just keyMap instead as well

spark33 (changes requested):

PR #1534, opened 2022-12-08
Adds baseFontSize prop to all Typography components that consume baseFontSize from LG provider

brookescarlett:
## ✍️ Proposed changes

Adds baseFontSize prop to all Typography components that consume baseFontSize from LG provider

🎟 _Jira ticket:_ [LG-1741](https://jira.mongodb.org/browse/LG-1741)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/typography/src/Body/Body.tsx
```diff
@@ -28,14 +28,16 @@ const fontWeights: Record<
...
   darkMode: darkModeProp,
   className,
   weight = 'regular',
   as = 'p' as T,
   ...rest
 }: BodyProps<T>) {
   const { theme } = useDarkMode(darkModeProp);
-  const baseFontSize = useUpdatedBaseFontSize();
+  const providerBaseFontSize = useUpdatedBaseFontSize();
+  const baseFontSize = baseFontSizeOverride ?? providerBaseFontSize;
```
shaneeza, line 40: I believe you can pass the override to `useUpdatedBaseFontSize(baseFontSizeOverride)`

brookescarlett: Updated - re r?

shaneeza (commented):
I think we should also update the README for all of these.

brookescarlett:
> I think we should also update the README for all of these.

Great call, on it.

packages/typography/src/types.ts
```diff
@@ -1,3 +1,6 @@
 import { DarkModeProps } from '@leafygreen-ui/lib';
+import { BaseFontSize } from '@leafygreen-ui/tokens';
 
-export type CommonTypographyProps = DarkModeProps;
+export type CommonTypographyProps = DarkModeProps & {
+  baseFontSize?: BaseFontSize;
```
shaneeza, line 5: Other components like `H1` and `H2` use `CommonTypographyProps` but they don't need to support `baseFontSize`

brookescarlett: I thought about that and have two conflicting opinions

1. I can see a world in which we do support "baseFontSize" in h1/h2 -- not that they would be 13/16px but that we change the font size of h1/h2 components based on whatever baseFontSize has been set. In this world, for now this is an optional prop and if you pass a value to the component it'll just be ignored so we're not introducing anything meaningful now and we can update later.
2. I can create a new type ResponsiveTypographyProps ? that inherits from Common and adds baseFontSize.

Which would you prefer to see here?

shaneeza: I prefer option 2, i think it would be a little odd to see a component support a prop but not really do anything with it.

brookescarlett: Sounds good to me! Will make the change.

brookescarlett: Done!

shaneeza (approved):

PR #1538, opened 2022-12-09
Card directory migration

brookescarlett:
## ✍️ Proposed changes

- Migrates Card to new dir structure
- Updates propTypes support 

🎟 _Jira ticket:_ [LG-2624](https://jira.mongodb.org/browse/LG-2624)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/card/src/Card.story.tsx
```diff
@@ -5,7 +5,7 @@ import { Story } from '@storybook/react';
 import { BoxProps } from '@leafygreen-ui/box';
 import { storybookArgTypes } from '@leafygreen-ui/lib';
 
-import { Card as LGCard, CardProps } from './Card';
+import { Card as LGCard, CardProps } from './Card/Card';
```
spark33, line 8: Looks like the subdir is missing an index

packages/card/src/Card/types.ts
```diff
@@ -0,0 +1,28 @@
...
+     */
+    contentStyle?: ContentStyle;
+
+    /**
+     * Determines whether or not the component will appear in dark mode.
+     *
+     * @default false
+     */
+    darkMode?: boolean;
+}
```
spark33, line 28: I know you just copied and pasted this, but we have a DarkModeProps we can extend instead of explicitly defining it here.

spark33 (changes requested):

packages/card/src/Card.story.tsx
brookescarlett: good catch

packages/card/src/Card/types.ts
brookescarlett: will update!

spark33 (approved):

PR #1543, opened 2022-12-13
LG-1453: Allows id prop to be passed to Modal

brookescarlett:
## ✍️ Proposed changes

Previously, we were overwriting the id prop passed to Modal. This is no longer the case.

🎟 _Jira ticket:_ [LG-1453](https://jira.mongodb.org/browse/LG-1453)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)

### DIFF
--- a/packages/modal/src/Modal/ModalView.tsx
+++ b/packages/modal/src/Modal/ModalView.tsx
@@ -36,12 +36,13 @@ function ModalView({
   size = ModalSize.Default,
   setOpen = () => {},
   shouldClose = () => true,
+  closeIconColor = CloseIconColor.Default,
   darkMode: darkModeProp,
+  id: idProp,
   children,
   className,
   contentClassName,
   initialFocus,
-  closeIconColor = CloseIconColor.Default,
   ...rest
 }: ModalProps) {
   const { theme, darkMode } = useDarkMode(darkModeProp);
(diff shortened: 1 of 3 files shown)


packages/modal/src/Modal/types.ts
```diff
@@ -18,6 +18,11 @@ export const ModalSize = {
 
 export type ModalSize = typeof ModalSize[keyof typeof ModalSize];
 
+export type ForwardedRef =
+  | React.ForwardedRef<HTMLDivElement | null>
+  | null
+  | undefined;
+
```
spark33, line 25: Interesting type - I'm wondering if we want this to be reusable from somewhere like lib

packages/modal/src/Modal/ModalView.tsx
```diff
@@ -25,126 +25,133 @@ import {
...
+      initialFocus,
+      ...rest
+    }: ModalProps,
+    forwardedRef: ForwardedRef,
+  ) => {
+    const { theme, darkMode } = useDarkMode(darkModeProp);
 
-  const nodeRef = React.useRef(null);
+    const nodeRef = React.useRef<HTMLDivElement | null>(null);
+    const ref = forwardedRef ?? nodeRef;
```
spark33, line 55: I might be wrong and it's a small nit - could we do `const nodeRef = React.useRef<HTMLDivElement | null>(forwardedRef ?? null);`

spark33 (approved):

PR #1550, opened 2022-12-19
Introduces style-guide.md

brookescarlett:
## ✍️ Proposed changes

This PR introduces style-guide.md a file where we can document our internal style guide with patterns/preferences that we want to standardize across the repo. This will not act as a resource to retro-fit our entire repo, but instead will help keep things clean/consistent moving forwards. This is a first-stab and by no means complete. Off the top of my head things missing are: unit tests/testing guidance, more specific direction on how we name variables (specifically CSS variables), how we define components (arrow functions or named/regular functions), and how we export things from files. Please feel free to add guidance around this!! For now, the workflow should just be a PR (as we normally do, and for bigger changes we can bring topics to Team Code Review on Thursdays). 

Hopefully this is helpful!


DesignerDave (commented):
Nit: Usually the convention for read-only markdown files like this is to name them in all-caps. `STYLEGUIDE.md` or similar.

STYLEGUIDE.md
```diff
@@ -0,0 +1,345 @@
...
+
+#### Prefer
+
+```typescript
+const handleClick = (e) => {
+    e.preventDefault()
+    setState(curr => !curr)
+}
+
+<button onClick={handleClick}>
```
spark33, line 213: Might also be nice to mention matching the event handler names using `on` to `handle`

spark33 (approved):

STYLEGUIDE.md
brookescarlett: Good feedback, added!

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+- [To Contribute](#To-Contribute)
+- [Typing](#Typing)
+- [Functions](#Functions)
+- [Variable naming](#Variable-naming)
+- [Variable placement](#Variable-placement)
+- [Whitespace and line breaks](#Whitespace-and-line-breaks)
+- [JSX/React](#JSX-React)
+- [CSS-in-JS](#CSS-in-JS)
+- [File Structure](https://github.com/mongodb/leafygreen-ui/blob/main/stories/Folder-Structure.stories.mdx)
+- [References](#References)
```
TheSonOfThomp, line 14: I wonder if this would make more sense as a GH Wiki? Or maybe a separate file for each section? This feels a little long & hard to read (a bit TL;DR)

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+Ensures that TypeScript doesn't prevent people from using the values directly rather than consuming the enum.
+
+#### Prefer
+
+```typescript
+export const someConstant = {
+  optionOne = 'option 1',
+  optionTwo = 'option 2',
+} as const;
```
TheSonOfThomp, line 52: Add how to use this in TS
```ts
export type someConstant = typeof someContent[keyof typeof someContent]
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+```typescript
+enum someConstant = {
+ optionOne: 'option 1',
+ optionTwo: 'option 2',
+}
+```
+
+---
+
+### When creating TS entities, place them at the top of the file, above JS
```
TheSonOfThomp, line 66: Why?

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+---
+
+## <a id="Functions">Functions</a>
+
+### Prefer ternaries over simple `if/else` blocks
```
TheSonOfThomp, line 72: But we should prefer `if/else` blocks over complex ternaries, because those are hard to read. 

#### Prefer
```ts
if (isBlah) {
 return 'a'
} else if (isFoo) {
  return 'b'
} else {
  return 'c'
}
```

over

```ts
return isBlah ? 'a' : isFoo ? 'b' : 'c'
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+if (isBlah) {
+  return a;
+} else {
+  return b;
+}
+```
+
+---
+
+### All new (exported) functions should have a JSDoc comment explaining functionality.
```
TheSonOfThomp, line 96: Technically TSDoc

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+};
+```
+
+---
+
+### Prefer creating the function in the highest scope, and passing in all required data.
+
+#### Why
+
+Keeps functions pure.
```
TheSonOfThomp, line 139: Can also consider creating utility closures (e.g. https://github.com/mongodb/leafygreen-ui/blob/943e4191dabb74acf403b2189327216085cc3e81/packages/brand-shape/src/utils.ts#L15)

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+---
+
+## <a id="Variable-naming">Variable Naming</a>
+
+### Acronyms are considered a single entity in camel-casing
+
+#### Why
+
+Some treat each letter in an acronym as its own entity when an acronyme is included in a var name. We'll instead consider the acronym as a single entity, and only capitalize the first letter of it (if that).
```
TheSonOfThomp, line 149: This is opposite of most React variables/types though (e.g. `type React.HTMLElement`)

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Use UPPERCASE for static constants, and refactor to a shared file if needed.
+
+---
+
+### Use camelCase for most javascript functions and variables
+
+---
+
+### Use “be” verbs for boolean types. (e.g. should`/`can`/`does`/`has`)
```
TheSonOfThomp, line 173: `isX`

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+let myWIPValue = 'pending';
+```
+
+---
+
+### Use UPPERCASE for static constants, and refactor to a shared file if needed.
+
+---
+
+### Use camelCase for most javascript functions and variables
```
TheSonOfThomp, line 169: But use `PascalCase` for components/classes etc

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Avoid inline declaration of static variables when they can be moved to top of file or refactored into new file
+
+---
+
+### Prefer placing let/const variables at the beginning of their scope.
+
+---
+
+### Prefer const to let whenever possible.
```
TheSonOfThomp, line 187: "`const` until you `can'tst`" 😆

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+---
+
+## <a id="Whitespace-and-line-breaks">Whitespace and line breaks</a>
+
+### Use a space before every prop defined in an interface
```
TheSonOfThomp, line 193: Can probably lint this one

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+---
+
+## <a id="JSX-React">JSX/React</a>
+
+### Prefer factoring out any more-than-1-line event handlers into their own functions
+
+#### Why
+
+Minimizes inline javascript logic within
```
TheSonOfThomp, line 203: Also creates a named function that'll show up in stack traces

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+return <div></div>;
+```
+
+---
+
+### Prefer using spread operator to avoid mutation of arrays and objects
+
+#### Why
+
+Immutability
```
TheSonOfThomp, line 273: Not really sure what this means

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+---
+
+### Avoid nesting ternary operators
+
+---
+
+### Factor out of functional components any methods that do NOT require state
```
TheSonOfThomp, line 281: Feels hard to parse. Would change to `"Factor out any methods that do NOT require state from functional components"`,  or even `"Prefer utility functions when full component state is not required"`

STYLEGUIDE.md
TheSonOfThomp: Also, missing a tick before "should"

STYLEGUIDE.md
brookescarlett: Chatted offline, leaving this as-is for now. Can re-assess based on usage!

STYLEGUIDE.md
brookescarlett: Helps keeps files organized, and if we want to know where to look for things this will be helpful

STYLEGUIDE.md
brookescarlett: yes absolutely. I can document this as well.

brookescarlett: updated!

STYLEGUIDE.md
brookescarlett: Would you prefer WIP? I can't think of another example where we don't do that other than HTML

STYLEGUIDE.md
brookescarlett: Not mutating props that come in... creating a new object by spreading props.

STYLEGUIDE.md
brookescarlett: We don't currently

STYLEGUIDE.md
brookescarlett: What do you mean here?

brookescarlett:
re-r? @TheSonOfThomp

STYLEGUIDE.md
shaneeza: Hmm, I've never thought about this. It is a little easier to read when the entire acronym is capitalized. Are there examples of how other style guides handle this?

STYLEGUIDE.md
shaneeza: Is there no example for this one?

STYLEGUIDE.md
```diff
@@ -0,0 +1,367 @@
...
+
+#### Avoid
+
+```typescript
+<div className={css`...`} />
+```
+
+---
+
+### Postfix style variable names appropriately
```
shaneeza, line 308: Should we mention theme styles? `myComponentThemeStyle` and `myComponentBaseStyle`?

brookescarlett: Yeah! We should! What's the rule there? Maybe you should make a PR to my branch 😄

STYLEGUIDE.md
brookescarlett: I took this from Atlas' style guide!

STYLEGUIDE.md
TheSonOfThomp: Agh I see. I think an example on this one would help

STYLEGUIDE.md
TheSonOfThomp: Sorry that was pretty unclear. Instead of defining a bunch of utilities that require component context, you can create a single utility function that accepts some component state as args, and returns a bunch of utilities with the component state in the closure. Not always useful, so it's fine to leave it out

STYLEGUIDE.md
TheSonOfThomp: Most built-in examples use all caps (`CSS`, `XML`, `HTML`, `SVG`, `JSX`,  `JSON`, )
Though one example that's not even internally consistent is `XMLTHttpRequest`

brookescarlett: Kk, will leave this out

STYLEGUIDE.md
```diff
@@ -0,0 +1,407 @@
...
+
+### Avoid nesting ternary operators
+
+---
+
+### Prefer utility functions when full component state is not required
+
+---
+
+## <a id="CSS-in-JS">CSS in JS</a>
```
TheSonOfThomp, line 325: @brookescarlett: @shaneeza mentioned documenting CSS property order. I'm personally a fan of the [Idiomatic CSS order](https://github.com/ream88/stylelint-config-idiomatic-order)

TheSonOfThomp (approved):
LGTM. Left a few comments

STYLEGUIDE.md
TheSonOfThomp: I think that's fair. Not sure how much we use acronyms really (the only one that comes to mind is `FedRAMP`)

PR #1551, opened 2022-12-20
Polymorphic default

TheSonOfThomp:
## ✍️ Proposed changes

Fixes issues in Polymorphic. Adds a `DefaultAs` generic prop 

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/polymorphic/src/Polymorphic/index.ts
```diff
@@ -9,15 +8,3 @@ export type {
...
-  toBePolymorphic,
-});
-
-declare global {
-  namespace jest {
-    interface Matchers<R> {
-      toBePolymorphic(): R;
-    }
-  }
-}
```
TheSonOfThomp, line 23: This causes issues when importing the component. We'll need to extend expect in testing lib https://jira.mongodb.org/browse/LG-2733

brookescarlett:
I think we've moved too hard in the other direction... not seeing any TS errors anymore when misusing as/href

packages/polymorphic/src/Example/examples.spec.tsx
```diff
@@ -59,6 +60,15 @@ describe('Polymorphic Higher-order Components', () => {
       expect(getByTestId('styled')).toHaveAttribute('href', 'mongodb.design');
       expect(getByTestId('styled')).toHaveStyle(`color: #ff69b4;`);
     });
+
+    test('Works with a default, and sets default props', () => {
```
brookescarlett, line 64: should probably test that as/href work as expected as well

packages/polymorphic/src/Example/Polymorphic.example.tsx
```diff
@@ -48,16 +52,37 @@ export const ExamplePolymorphicWithRef = Polymorphic<ExampleProps>(
   'ExamplePolymorphicWithRef',
 );
 
-export const ExampleInferred = Polymorphic<
-  InferredPolymorphicProps<ExampleProps>
->(({ as, title, ...rest }) => {
+/**
+ * @example
+ */
+export const ExampleInferred = InferredPolymorphic<ExampleProps>(
```
brookescarlett, line 58: does this need defaults? What happens if there are none?

TheSonOfThomp: This is an example for usage without a default. 

`ExampleInferredDefaultButton` is the example for usage with a default

scripts/utils/tsDocParser.ts
```diff
@@ -58,17 +58,47 @@ const compilerOptions: CompilerOptions = {
 
 const Parser = withCompilerOptions(compilerOptions, TSDocOptions);
 
+interface ParseFunctionOptions {
+  /**
```
brookescarlett, line 62: are these changes related to poly default or should this be its own PR?

packages/polymorphic/README.md
```diff
@@ -83,6 +83,24 @@ export const MyInferredComponent = Polymorphic<
 <MyInferredComponent href="mongodb.design" />; // renders as <a>
 ```
 
+#### With a default `as` value
```
brookescarlett, line 86: Maybe provide a little bit of context as to when/why we'd want to use a default versus not (what value is being provided) -- seeing both examples is confusing without context

scripts/utils/tsDocParser.ts
TheSonOfThomp: Yeah, I made these changes to the parser to make sure we weren't publishing docs for the example compnents

brookescarlett (approved):

packages/polymorphic/src/Polymorphic/index.ts
brookescarlett: This needs to be merged in before my PR that converts typography off of box and onto polymorphic

PR #1553, opened 2023-01-05
LG-2706: Table Enhancements

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2706](https://jira.mongodb.org/browse/LG-2706)

Tests are not written yet and I'm still reviewing this myself, but opening it up for review sooner rather than later. No changeset here since tests are not written yet and this will be an alpha release.

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/table/src/Cell/Cell.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { PropsWithChildren } from 'react';
+import InternalCellWithoutVS from './InternalCellWithoutVS';
+import InternalCellWithVS from './InternalCellWithVS';
+import { CellProps } from './types';
+
+const Cell = <T extends unknown>({
+  cell,
+  toggleExpandedIconProps,
```
spark33, line 8: Fails when not provided. I tried breaking it out into `Cell`, `InternalCellWithVS`, `InternalCellWithoutVS`, and `InternalCellBase` and creating separate types for them, but it broke. Can't remember what broke then. Will need to come back to this.

packages/table/src/Table/Table.stories.tsx
```diff
@@ -0,0 +1,602 @@
...
+  return <>TODO</>
+}
+
+export const Selectable = () => {
+  return <>TODO</>
+}
+
+export const ExpandableContent = () => {
+  const data = makeData(false, 100, 3, 2);
+  const columns = Object.keys(data[0]).filter(x => x !== 'renderExpandedContent' && x !== 'subRows')
```
brookescarlett, line 118: What is the `renderExpandedContent` key doing in the data object? I'm not seeing it used anywhere. I think this is just bloat?

packages/table/src/Table/Table.stories.tsx
```diff
@@ -0,0 +1,602 @@
...
+      {
+        accessorKey: 'progress',
+        header: 'Profile Progress',
+        size: 80,
+      },
+    ],
+    [],
+  );
+
+  const table = useLeafygreenTable<Person>({
```
brookescarlett, line 194: can we make this pattern extend to no virtual rows? we can add an extra arg useVirtualRows or something, so that this API can be compatible either way

packages/table/src/Table/Table.stories.tsx
```diff
@@ -0,0 +1,602 @@
...
+    containerRef: tableContainerRef,
+    data,
+    columns,
+    getCoreRowModel: getCoreRowModel(),
```
brookescarlett, line 198: I don't think anyone should have to import from tanstack so we should wrap this

packages/table/src/Table/Table.stories.tsx
```diff
@@ -0,0 +1,602 @@
...
+      <TableBody>
+        {data.map((row: any) => (
+          <Row>
+            {columns.map((cellKey: string, index: number) => {
+              return <Cell key={`${cellKey}-${index}`}>{row[cellKey]}</Cell>;
+            })}
+            {row.subRows && row.subRows.map((subRow: any) => {
+              let subRowCellKeys = Object.keys(subRow).filter(x => x !== 'renderExpandedContent' && x !== 'subRows')
+              return (
+                <Row>
```
brookescarlett, line 92: Did we decide against nesting Rows within each other?

packages/table/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  rules: {
+    'react/forbid-prop-types': 'off',
```
brookescarlett, line 3: why are we introducing this?

packages/table/src/Cell/Cell.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { CellProps } from './Cell.types';
+import InternalCellBase from './InternalCellBase';
+import InternalCellWithRT from './InternalCellWithRT';
+
+const Cell = <T extends unknown>({
+  cell,
+  ...rest
+}: PropsWithChildren<CellProps<T>>) => {
+  const usesRT = !!cell;
```
brookescarlett, line 12: should this value be coming from context? single source of truth

packages/table/src/Cell/InternalCellBase.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (columnAlignments !== undefined && cellIndex !== undefined) {
+      setAlign(columnAlignments[cellIndex]);
+    }
+  }, [cellIndex, columnAlignments]);
+
+  return (
+    <td className={cx(baseStyles, className)} {...rest}>
+      <div
+        className={cx(cellContentContainerStyles, {
+          [depthPadding(depth)]: cellIndex === 0,
```
brookescarlett, line 40: function calls instead of objects? is this a pattern we're moving towards? if so, can we document in the style guide cc @TheSonOfThomp @shaneeza

packages/table/src/CheckboxCell/Checkbox.styles.tsx
```diff
@@ -0,0 +1,12 @@
+import { css } from '@leafygreen-ui/emotion';
+
+export const baseStyles = css`
+  // break the first-cell padding. would be cleaner to use a className here, but createUniqueClassname could have performance concerns.
```
brookescarlett, line 4: why performance concerns?

packages/table/src/CheckboxCell/Checkbox.styles.tsx
```diff
@@ -0,0 +1,12 @@
...
+  margin-left: -16px;
+  width: 36px;
+`;
+
+export const headerStyles = css`
+  // break the first-cell padding. would be cleaner to use a className here, but createUniqueClassname could have performance concerns.
```
brookescarlett, line 10: why are we breaking the first-cell padding?

packages/table/src/CheckboxCell/CheckboxCell.tsx
```diff
@@ -0,0 +1,31 @@
...
+import PropTypes from 'prop-types';
+
+import Checkbox from '@leafygreen-ui/checkbox';
+import { CheckboxProps } from '@leafygreen-ui/checkbox/src/types';
+import { cx } from '@leafygreen-ui/emotion';
+
+import { baseStyles, headerStyles } from './Checkbox.styles';
+
+const CheckboxCell = (props: CheckboxProps & { isHeader?: boolean }) => (
+  <div className={cx(baseStyles, { [headerStyles]: props.isHeader })}>
```
brookescarlett, line 11: should this not be wrapped in `<td>` as the outer element?

packages/table/src/HeaderCell/HeaderCell.tsx
```diff
@@ -0,0 +1,88 @@
...
+        },
+        className,
+      )}
+      scope="col"
+      {...rest}
+    >
+      <div className={cx(contentContainerStyles, alignmentStyles(align))}>
+        {children}
+        {sortState && onSortIconClick && (
+          <SortIcon
```
brookescarlett, line 77: does this icon not change based on state of sort?

packages/table/src/HeaderCell/SortIcon/SortIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
+    [SortState.Desc]: palette.blue.base,
+    [SortState.Off]: palette.gray.dark1,
+  },
+};
+
+const SortIcon = ({ sortState, onSortIconClick, ...rest }: SortIconProps) => {
+  const { theme } = useDarkMode();
+
+  const handleClick: MouseEventHandler = e => {
+    onSortIconClick && onSortIconClick(e);
```
brookescarlett, line 37: nit: onSortClick?.(e)

packages/table/src/Row/InternalRowBase.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <tr
+      onClick={onClick}
+      aria-disabled={disabled}
+      tabIndex={onClick ? 0 : undefined}
+      className={cx(
+        baseStyles,
+        {
+          [zebraStyles[theme]]: shouldAlternateRowColor,
+          [disabledStyles[theme]]: disabled,
+          [clickableStyles[theme]]: !!onClick,
```
brookescarlett, line 34: can we document this pattern of coercing variables to booleans as well

packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -0,0 +1,76 @@
...
+  className,
+  row,
+  virtualRow,
+  disabled,
+  ...rest
+}: InternalRowWithRTProps<T>) => {
+  const { theme } = useDarkMode();
+  const isNestedRowParent = row.depth === 0 && row.getIsExpanded();
+  const isNestedRowOrParent = row.getIsExpanded() || row.depth > 0;
+  const ExpandedContentRowProp = row?.original.renderExpandedContent;
```
brookescarlett, line 29: nit: lc E?

packages/table/src/Cell/Cell.tsx
spark33: The user is the one calling the hook and we can't know which `cell` object in the `data` prop the `Cell` component needs, so the user does need to pass it in directly to the component

packages/table/.eslintrc.js
spark33: We extend a lot of types from `react-table`, which necessitates the use of `Proptypes.any`

packages/table/src/Cell/InternalCellBase.tsx
spark33: `depth` is a number with an unknown value, so it just makes more sense here. If it's a prop like variant, I think we prefer objects.

packages/table/src/CheckboxCell/Checkbox.styles.tsx
spark33: Actually, maybe since we only need 2 classNames, this would be fine. For some reason, I was thinking we'd need a unique classname for every first cell.

packages/table/.eslintrc.js
brookescarlett: Why does it necessitate that?

packages/table/src/CheckboxCell/Checkbox.styles.tsx
spark33: Because the checkbox column needs to be skinner than other columns, but we DO want a base padding in a regular cell + `table-layout: fixed` as a default behavior. The checkbox cell needs to be passed as a child to Cell, so it's tricky to break the base Cell padding. I could explore an approach with a value created by `createUniqueClassname`, but I'm not 100% sure how that would work off the top of my head.

packages/table/src/CheckboxCell/CheckboxCell.tsx
spark33: It gets passed a child to our Cell component by default

packages/table/src/Row/InternalRowWithRT.tsx
spark33: it's a function component which is why it's capitalized to indicate that

spark33: also this comment was trippy - I thought it was a terrible typo before I decoded it lol

spark33:
This branch is now deprecated and will be replaced with the work in the branch `table-v10-3`.

PR #1555, opened 2023-01-09
LG-2703: language select on click

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2703](https://jira.mongodb.org/browse/LG-2703)

The reason the language selector would not open initially on load when clicked was because of `useUsingKeyboardContext`.
Since the default for the keyboard provider was `true`, whenever a `keydown` would occur the state would update and that would cause the `code` component to re-render. Because of the re-render the dropdown would not open. I decided to remove the hook and switch to `focus-visible`. It's essentially the same exact thing without any additional re-renders.

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)

### DIFF
--- a/packages/code/src/Code.tsx
+++ b/packages/code/src/Code.tsx
@@ -9,7 +9,6 @@ import { css, cx } from '@leafygreen-ui/emotion';
 import { useIsomorphicLayoutEffect } from '@leafygreen-ui/hooks';
 import LeafyGreenProvider, {
   useDarkMode,
-  useUsingKeyboardContext,
 } from '@leafygreen-ui/leafygreen-provider';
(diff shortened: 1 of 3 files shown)


packages/code/src/Code.tsx
```diff
@@ -248,7 +242,6 @@ function Code({
   ...rest
 }: CodeProps) {
   const scrollableElementRef = useRef<HTMLPreElement>(null);
-  const { usingKeyboard: showFocus } = useUsingKeyboardContext();
```
brookescarlett, line 251: Why are we removing this?

brookescarlett (approved):

PR #1560, opened 2023-01-10
Migrates Button component to new directory structure

brookescarlett:
Going to migrate Button off of Box and onto Polymorphic, but wanted to break the PRs up in two so that there's a separation of concerns and the code changes are easier to follow.


packages/button/src/Button/Button.styles.ts
```diff
@@ -0,0 +1,422 @@
...
+const focusBoxShadow = (color: string) => `
+    0 0 0 2px ${color}, 
+    0 0 0 4px ${palette.blue.light1};
+`;
+
+const baseButtonStyles = css`
+  // unset browser default
+  appearance: none;
+  padding: 0;
+  margin: 0;
```
TheSonOfThomp, line 24: Is this also solving https://jira.mongodb.org/browse/LG-2681?

packages/button/src/index.ts
```diff
@@ -1,3 +1,3 @@
-export { Button as default } from './Button';
+export { Button as default } from './Button/Button';
```
TheSonOfThomp, line 1: Nit: Should we be creating a `Button/index.ts` file?

TheSonOfThomp (approved):
🎊

PR #1565, opened 2023-01-10
Renames palette story

TheSonOfThomp:
Renames `Palette.story.tsx` (uppercase) so `.design` can detect it (import statements are case sensitive)

Resolves No. 6 https://docs.google.com/document/d/1zds271-DF_vmsF4-yy7jjV20jvpEtpTsSPIxf5w7xo4/edit


.changeset/tender-birds-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/palette': patch
```
brookescarlett, line 2: Not sure that we need a changeset but won't block on this!

brookescarlett (approved):

.changeset/tender-birds-flow.md
TheSonOfThomp: I think we need a changeset to make sure it publishes

brookescarlett: Duh you're right.. you'll have to add a changeset to your typography PR then as well

TheSonOfThomp: Will do 👍

PR #1589, opened 2023-01-25
LG-2809: add NoFilter icon

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2809](https://jira.mongodb.org/browse/LG-2809)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/icon/src/glyphs/NoFilter.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M2.60553 1.2041C2.21353 0.905115 1.65113 0.934737 1.29289 1.29297C0.902369 1.68349 0.902369 2.31666 1.29289 2.70718L7 8.41429V14.3981C7 14.8277 7.50596 15.0573 7.82925 14.7744L10.779 12.1933L13.2908 14.7051C13.6813 15.0956 14.3145 15.0956 14.705 14.7051C15.0955 14.3146 15.0955 13.6814 14.705 13.2909L12.8869 11.4728L12.8806 11.4791L2.60553 1.2041Z" fill="#001E2B"/>
```
TheSonOfThomp, line 2: I think we need to set the `fill` property to `#000`

TheSonOfThomp (changes requested):
We need to set the `fill` property to some variant of `black` in order for the icon to be colorized.

Not sure why the actions aren't catching it, but there's a failing test

packages/icon/src/glyphs/NoFilter.svg
shaneeza: good catch

shaneeza:
@TheSonOfThomp If i ignore the rebuilt files with the updated imports then I get over 100 failed test. Thoughts on if we should include these new icon files?

TheSonOfThomp:
> @TheSonOfThomp If i ignore the rebuilt files with the updated imports then I get over 100 failed test. Thoughts on if we should include these new icon files?

I think we should still be including those for now

packages/icon/src/glyphs/NoFilter.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M2.60553 1.2041C2.21353 0.905115 1.65113 0.934737 1.29289 1.29297C0.902369 1.68349 0.902369 2.31666 1.29289 2.70718L7 8.41429V14.3981C7 14.8277 7.50596 15.0573 7.82925 14.7744L10.779 12.1933L13.2908 14.7051C13.6813 15.0956 14.3145 15.0956 14.705 14.7051C15.0955 14.3146 15.0955 13.6814 14.705 13.2909L12.8869 11.4728L12.8806 11.4791L2.60553 1.2041Z" fill="#000000"/>
```
TheSonOfThomp, line 2: 👍

TheSonOfThomp (approved):

PR #1590, opened 2023-01-27
Introduces Form Control component

bruugey:
## ✍️ Proposed changes

Creates a useAccessibleField hook that properly relates a label, input, description, and errorMessage

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/form-control/src/hooks/hooks.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { render } from '@testing-library/react';
+import { renderHook } from '@testing-library/react-hooks';
+
+import { ValidationState } from '../types';
+
+import { AccessibleFieldProps } from './types';
+import { useAccessibleField } from '.';
+
+describe('packages/form-control', () => {
+  describe('useAccessibleField', () => {
```
TheSonOfThomp, line 11: This is fine for now, but I think we'd want a new `.spec` file for each hook/component

TheSonOfThomp (approved):
📝  LGTM!

packages/form-control/src/types.ts
```diff
@@ -0,0 +1 @@
+export type ValidationState = 'error' | 'none' | 'valid';
```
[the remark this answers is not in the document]
brookescarlett:
```
export const State = {
  error: 'error',
  none: 'none',
} as const;
export type State = typeof State[keyof typeof State];

export const SearchState = {
  unset: 'unset',
  error: 'error',
  loading: 'loading',
} as const;
export type SearchState = typeof SearchState[keyof typeof SearchState];
```

packages/form-control/src/hooks/useAccessibleField.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { AccessibleFieldProps, LabelProps } from './types';
+
+export const useLabel = ({
+  label,
+  id: idProp,
+  'aria-labelledby': ariaLabelledby,
+  'aria-label': ariaLabel,
+}: LabelProps) => {
+  const id = useIdAllocator({ prefix: 'lg-input', id: idProp });
```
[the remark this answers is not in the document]
brookescarlett: 👍

packages/form-control/src/hooks/hooks.spec.tsx
brookescarlett: probably eventually

packages/form-control/src/hooks/useAccessibleField.ts
```diff
@@ -0,0 +1,74 @@
...
+}: AccessibleFieldProps) => {
+  const { labelProps, fieldProps } = useLabel({
+    label,
+    id,
+    'aria-label': ariaLabel,
+    'aria-labelledby': ariaLabelledby,
+  });
+
+  const descriptionId = useIdAllocator({ prefix: 'lg-description' });
+  const errorMessageId = useIdAllocator({ prefix: 'lg-errorMessage' });
```
shaneeza, line 58: I don't think we are addressing if a custom message container is used, so if the consumer is passing in `aria-describedby`.

packages/form-control/src/hooks/useAccessibleField.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const ariaDescribedby = useMemo(
+    () =>
+      `${description ? descriptionId : ''} ${
+        errorMessage && validationState === 'error' ? errorMessageId : ''
```
shaneeza, line 63: With password input, messages are always visible regardless if they are valid or not. I was also planning on always rendering the message container, even if there are no message, and adding `aria-live` to the container so that it can announce when a message appears. If i were to use this hook it seems like I would always have to pass in a string to `errorMessage` and set `validationState` to `error`  to get the correct text inside `aria-describedby`.

spark33 (approved):
I see Adam and Shaneeza calling out some good details about how the hook behaves. I imagine that they'll catch themselves as we gradually consume the hook in the input components though, so I'm personally not be as concerned on that front. The code looks good to me, so approving!

packages/form-control/src/hooks/useAccessibleField.ts
brookescarlett: ok i can spread that, great call

packages/form-control/src/hooks/useAccessibleField.ts
brookescarlett: Ok so two thoughts here...

Option A: This is contained to the specific use case of handling four distinct elements: label, input, description, error. In the case of Password Input, description might be more accurate for the validation inputs rather than errorMessage, because it isn't exactly that.

Option B: We revert to the original generator function, where we use a hook to check for the presence of ids in the DOM, rather than using errorMessage and validation states as conditionals 

What do you think? cc/ @TheSonOfThomp / @spark33

packages/form-control/src/hooks/useAccessibleField.ts
```diff
@@ -0,0 +1,75 @@
...
+      } ${ariaDescribedbyProp ? ariaDescribedbyProp : ''}`.trim(),
```
shaneeza, line 65: 👍🏽

packages/form-control/src/hooks/useAccessibleField.ts
shaneeza: PasswordInput is a one-off component but it can still use label and input props from the hook. Technically we can use description for the messages but we would always have to pass it a string to get it to show up.

The generator way seems like a lot for so little but that's just because I don't fully understand what it was doing in the first place.

spark33: I personally don't see an issue with the code here since even with password input, I imagine that `errorMessage` would only be one of the errors. 

In other words, even if the password input has 4 error messages to display at all times, the `aria-describedby` should probably only be on the first; the other errors are displayed but aren't used in accessibility props. I'm thinking this because if there are no errors on the password input, it would be inaccurate to say any of the error messages actually describe the input until `validationState === "error"`.

If I'm misunderstanding something here and Brooke's options are the only two options available, then I'd vote for option A.

packages/form-control/src/hooks/types.ts
```diff
@@ -0,0 +1,18 @@
...
+  id?: string;
+}
+
+export interface AccessibleFieldProps extends AriaLabels {
+  id?: string;
+  description?: string;
+  errorMessage?: string;
+  validationState?: ValidationState;
+  ['aria-describedby']?: string;
+}
```
TheSonOfThomp, line 18: Can we add TSDoc to these?

packages/form-control/src/hooks/useAccessibleField.ts
shaneeza: Yeah the aria-describedby would be on the entire message container, not an individual message. And no, there is no description.

TheSonOfThomp: I agree with Brooke, that the entire container should have `aria-describedby`.

@shaneeza What's the anticipated API for the Password requirements? Seems there would be different "messages" for different states: `errorMessage` (i.e. "Please enter matching passwords"), `validMessage` (i.e. "Your passwords match"), and the array of password requirements. 

We could probably generalize this hook api to have `label`, `description`, and `message` props, and just check that `validationState !== 'unset'` before setting describedby to the messageId. Then, regardless of valid/error state, the password can be described by the message below it

shaneeza: For password input there is only a message array for all messages, check out the [RFC](https://docs.google.com/document/d/1b_RsqgOAdlWkyJNFI_PsOXNKbkBerg4DCTXQ_Ybsbcs/edit#). Internally there is a state but even in the case that there isn't a message I was always going to render the message container to take advantage of `aria-live`. The only case where the message container would not render is if the user is using a custom message container. If we change `validationState` values to something else like `set`/`unset` or `visible`/`notVisible` and get rid of `errorMessage` then maybe it could work for all inputs.

brookescarlett: Oh I don't hate the go-forward of

label, input, description, message and checking if `validationState !== undefined`?

brookescarlett: wydt @TheSonOfThomp

brookescarlett:
PTAL @TheSonOfThomp, @shaneeza, @spark33

packages/form-control/src/hooks/types.ts
```diff
@@ -0,0 +1,16 @@
...
+  id?: string;
+}
+
+export interface AccessibleFieldProps extends AriaLabels {
+  id?: string;
+  description?: string;
+  message?: string;
+  isMessageShown?: boolean;
+  ['aria-describedby']?: string;
+}
```
TheSonOfThomp, line 16: Especially now since `message` is a bit ambiguous, we should add TSDoc to this

packages/form-control/src/hooks/types.ts
brookescarlett: yep

packages/form-control/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+  "include": ["src/**/*"],
+  "references": [
+    {
+      "path": "../lib"
+    }, 
+    {
+      "path": "../hooks"
+    }
+  ]
+}
```
shaneeza, line 18: I think that we also need to add this component to `build.tsconfig.json`?

PR #1596, opened 2023-01-30
LG-2181: Pagination component

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2181](https://jira.mongodb.org/browse/LG-2181)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


.changeset/great-beans-remember.md
```diff
@@ -0,0 +1,7 @@
+---
+'@leafygreen-ui/pagination': major
+'@leafygreen-ui/icon-button': patch
+'@leafygreen-ui/select': patch
+---
+
+First release of Pagination component; adding required exports in other relevant packages
```
brookescarlett, line 7: nit but two separate changesets please :)

DEVELOPER.md
```diff
@@ -71,3 +71,10 @@ We use @testing-library/react for writing tests locally. This library helps mock
 #### Linking
 
 We also have a link script, such that you can test components that are in development in environments beyond Storybook. To do so, run `yarn run link -- [path-to-application]`.
+
+## Creating a new component
```
brookescarlett, line 75: thank you!!

packages/pagination/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+## Example
+
+**Output HTML**
+
+## Properties
+
+| Prop | Type | Description | Default |
+| ---- | ---- | ----------- | ------- |
```
brookescarlett, line 28: add props

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,141 @@
...
+    return Math.ceil(numItems / itemsPerPage);
+  };
+
+  return (
+    <LeafyGreenProvider darkMode={darkMode}>
+      <div className={cx(baseStyles, className)} {...rest}>
+        <div className={flexSectionStyles}>
+          {onItemsPerPageOptionChange !== undefined && (
+            <>
+              <Body
```
brookescarlett, line 63: Do we not have a Label component for this?

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const getRangeMaxString = () => {
+    return numTotalItems ? `${numTotalItems} items` : 'many';
+  };
+
+  const getTotalNumPages = (numItems: number) => {
+    return Math.ceil(numItems / itemsPerPage);
+  };
+
+  return (
+    <LeafyGreenProvider darkMode={darkMode}>
```
brookescarlett, line 58: Why do we need to nest a provider here?

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,141 @@
...
+            {getCurrentRangeString()} of {getRangeMaxString()}
+          </Body>
+        </div>
+        <div className={flexSectionStyles}>
+          {onCurrentPageOptionChange !== undefined && numTotalItems ? (
+            <>
+              {/* @ts-expect-error no label element */}
+              <Select
+                onChange={onCurrentPageOptionChange}
+                aria-label="Set current page"
```
brookescarlett, line 98: Could also use a `<VisuallyHidden />` component for this

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const itemsPerPageLabelId = useIdAllocator({
+    prefix: 'lg-pagination-items-per-page-label',
+    id: idProp,
+  });
+  const itemsPerPageSelectId = useIdAllocator({
+    prefix: 'lg-pagination-items-per-page-select',
+    id: idProp,
+  });
+
+  const getCurrentRangeString = () => {
```
brookescarlett, line 42: could these functions accept args and live outside the component itself?

scripts/create-package.js
```diff
@@ -40,7 +40,7 @@ fs.mkdir(newDir, { recursive: true }, err => {
     fs.writeFile(`${newSrcDir}/index.ts`, rootIndex, handleErr);
 
     fs.writeFile(
-      `${newDir}/src/${packageNamePascal}.story.tsx`,
+      `${newDir}/src/${packageNamePascal}.stories.tsx`,
```
brookescarlett, line 43: Is this change going to break anything in the website cc @TheSonOfThomp

packages/pagination/src/Pagination/Pagination.tsx
spark33: to pass the `darkMode` prop to all child components by default

packages/pagination/src/Pagination/Pagination.tsx
spark33: We have a label component, but it has bold text. I think `<Body as="label">` is cleaner and more future-change-proof than `<Label className={...}>`.

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,141 @@
...
+              {/* @ts-expect-error no label element */}
+              <Select
+                onChange={onCurrentPageOptionChange}
+                aria-label="Set current page"
+                value={String(currentPage)}
+                allowDeselect={false}
+                size="xsmall"
+                data-testid="lg-pagination-page-select"
+              >
+                {Array.from(Array(getTotalNumPages(numTotalItems)).keys()).map(
```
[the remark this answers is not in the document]
spark33: `Array(getTotalNumPages())` creates an array that's [undefined, undefined, ...]. Calling `Array.from(someArray.keys())` turns the array of undefineds into [0,1,2,3...]

packages/pagination/src/Pagination/Pagination.tsx
spark33: They could - I can make that change

packages/pagination/src/Pagination/Pagination.tsx
spark33: Isn't this like the most classic case of where `aria-label` should be used though? Using a hidden label feels like a hack

scripts/create-package.js
spark33: discussed with Adam - it might. Changed it back. We'll go back and rename all files to .stories.tsx at some point with a global PR

packages/icon-button/src/index.ts
```diff
@@ -1,2 +1,2 @@
 export { IconButton as default } from './IconButton';
-export { Size } from './types';
+export { type AccessibleIconButtonProps, Size } from './types';
```
brookescarlett, line 2: this is new

packages/pagination/src/Pagination/Pagination.tsx
brookescarlett: Ok

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,133 @@
...
+              <Select
+                aria-label="Set current page"
+                onChange={onCurrentPageOptionChange}
+                value={String(currentPage)}
+                allowDeselect={false}
+                size="xsmall"
+                data-testid="lg-pagination-page-select"
+              >
+                {Array.from(
+                  Array(getTotalNumPages(numTotalItems, itemsPerPage)).keys(),
```
brookescarlett, line 95: Still not clear on why the double array

packages/pagination/src/Pagination/Pagination.tsx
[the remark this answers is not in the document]
brookescarlett: OH

packages/pagination/src/Pagination/Pagination.types.ts
```diff
@@ -0,0 +1,40 @@
+import { AccessibleIconButtonProps } from '@leafygreen-ui/icon-button';
+import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
+import { SelectProps } from '@leafygreen-ui/select';
+
+interface PaginationProps extends HTMLElementProps<'div'>, DarkModeProps {
+  /**
+   * Number of items visible on the current page.
+   */
+  itemsPerPage?: number;
```
shaneeza, line 9: Did we agree on spaces between types?

packages/pagination/src/Pagination/utils.ts
```diff
@@ -0,0 +1,18 @@
+export const getCurrentRangeString = (
+  itemsPerPage: number,
+  currentPage: number,
+  numTotalItems?: number,
+) => {
+  return `${itemsPerPage * (currentPage - 1) + 1} - ${Math.min(
+    itemsPerPage * currentPage,
+    numTotalItems ?? itemsPerPage * currentPage,
+  )}`;
```
TheSonOfThomp, line 9: Might want to double check this for when `itemsPerPage`, `currentPage` and `numTotalItems` don't make sense together

packages/pagination/src/Pagination/Pagination.tsx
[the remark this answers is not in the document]
TheSonOfThomp: `_.range`  would probably be easier to read https://lodash.com/docs/4.17.15#range

packages/pagination/src/Pagination/Pagination.spec.tsx
```diff
@@ -0,0 +1,191 @@
+import React from 'react';
+import { getByText, waitFor } from '@testing-library/dom';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
```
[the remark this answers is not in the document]
brookescarlett: +1

packages/pagination/src/Pagination/utils.ts
spark33: This is just a util to get a formatted string, so the focus should probably be on not allowing incorrect values as props

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const DEFAULT_ITEMS_PER_PAGE_OPTIONS = [10, 25, 50];
+
+function Pagination({
+  id: idProp,
+  className,
+  itemsPerPage = DEFAULT_ITEMS_PER_PAGE_OPTIONS[0],
+  itemsPerPageOptions = DEFAULT_ITEMS_PER_PAGE_OPTIONS,
+  onItemsPerPageOptionChange,
+  currentPage = 1,
```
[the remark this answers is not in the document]
spark33: I assume you used Storybook controls for this? I didn't really check for this in the component because it's impossible to achieve with the controls we provide. This means this is only possible when the user passes in incorrect values as props.

If I'm not mistaken, there isn't a way to use prop-types to validate the value of a prop against the value of another prop even using a custom validator. I could probably `console.error` or throw an error in cases where the prop values are incorrect though. Let me add that

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,133 @@
...
+  getTotalNumPages,
+} from './utils';
+
+const DEFAULT_ITEMS_PER_PAGE_OPTIONS = [10, 25, 50];
+
+function Pagination({
+  id: idProp,
+  className,
+  itemsPerPage = DEFAULT_ITEMS_PER_PAGE_OPTIONS[0],
+  itemsPerPageOptions = DEFAULT_ITEMS_PER_PAGE_OPTIONS,
```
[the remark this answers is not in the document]
spark33: Good catch! Same response as below, but something we should flag with a  console.error at the very least.

packages/pagination/src/Pagination/Pagination.types.ts
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * Options to be shown in the Select to indicate items per page.
+   */
+  itemsPerPageOptions?: Array<number>;
```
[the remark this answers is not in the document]
spark33: Looked into this a little bit - the link below would work for us, but it requires us to use generics in PaginationProps (i.e. `(props: PaginationProps<number>)`, which I don't think makes much sense for us?

https://www.typescriptlang.org/play?ts=4.1.5#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQdwAdjFlAXlnACrIBGqAArEwAHm5wsAD1ZMAJuiYBXEHzYA+OAG86cPXFRYANlnxY5ALjgBRKbiNK5WAGLEQ3AJ5gsASSYE2LCZcLHF1en04GH5UK24AbQBdXX0IJgBlY1MYKwAKaL4rW3tHFzdPbz8AymDQ7nUASjgAXk0ANwhgOToAXwYYLy5ihydXEgrff0DasJa4eO5EhckZIIUUJg84AH44AAY4KyYsNrZE+jone2RKOGvUdABZDwBhEkhjlnEV2XXlVQ0PzW6GweBgADo3uA0kEYOIYsIIGJ6potD0LqJnlCPrCojFmlp4gBGAA0ACYSQBmRI9AxZMxyAkAFlpaUyJnwBIKLVRKT03OagrgRLgAB9RXi+C0hWT6D1aQB6dR0IA

packages/pagination/src/Pagination.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+export default {
+  title: 'Components/Pagination',
+  component: Pagination,
+  args: {
+    onCurrentPageOptionChange: undefined,
+  },
+  argTypes: {
+    currentPage: { control: 'number' },
+    numTotalItems: { control: 'number' },
+    itemsPerPage: { control: 'none' },
```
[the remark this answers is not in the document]
spark33: To prevent the user from inputting an invalid value that would break the component. Like I mentioned in the other thread, there isn't a way through ts or prop-types to validate values that come in as props against other props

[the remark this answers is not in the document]
spark33: Not to mention - there isn't really a good way to sync Storybook args with states. There's the `useArgs` hook, but it seems like there are still some issues with that.

packages/pagination/src/Pagination/Pagination.types.ts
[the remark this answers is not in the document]
TheSonOfThomp: Was playing around with this problem a bit last night too. I think this is what we want https://www.typescriptlang.org/play#code/JYOwLgpgTgZghgYwgAgApQPYAcDOAeAOWQgA9IQATHZEAVwFsAjaAGmQHliyJLq6noAbQC6APmQBvAFDJkCDLXAAuZAQDcM5NjDAMIHCqKlyVDoP7Mow5AH5kggHRP21lSAgA3aBoC+U+fpgyADCGPRYejxBALzIhFwmfAyWbJzGPKYWQmIAFNKy8opgLJrauvolPiro2PgEqaIAlMjR4hI+GlJdoeGR4HlyCsrIAKxsZXoG9gCMbABMbADM1j6NasgA9BvEUJhQ3WER7v0Sg0UqY1pYOpMqgrPIC8iLbCMra5vbGADWB73HYAGhWGlwm+ju1jgSQEUBEyFW6y2yDgABsUcgwAALFAQFEQehRaigK43fTIqAoLJQZGUUbICgYCDUdK0qnIJQbKRAA

packages/pagination/src/Pagination/Pagination.tsx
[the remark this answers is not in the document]
spark33: Would bringing in lodash just to shorten one line be worth the increase in package size though?

packages/pagination/src/Pagination/Pagination.types.ts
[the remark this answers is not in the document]
spark33: Yeah that looks similar to what I posted. I think it has the same issue as mine though

packages/pagination/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "bugs": {
+    "url": "https://jira.mongodb.org/projects/PD/summary"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/emotion": "^4.0.3",
+    "@leafygreen-ui/leafygreen-provider": "^3.1.0",
```
shaneeza, line 32: Should be peer dependency?

spark33: THANK YOU

packages/pagination/src/Pagination/Pagination.types.ts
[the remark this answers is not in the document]
TheSonOfThomp: Not sure what you mean, we're typing props in the component definition anyway. 

It's fine if we're using generics in the component definition, as long as we don't need them in the final api

```ts
interface Props<N extends number, O extends number[]> {
  count: N;
  options: N extends O[number] ? [...O] : never;
}
const Component = <N extends number, O extends number[]>({
  count,
  options,
}: Props<N, O>) => {};

<Component count={5} options={[1,2,3]} />  // error
```

packages/pagination/src/Pagination/Pagination.types.ts
spark33: Added them in!

packages/pagination/src/Pagination/Pagination.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Not if we import it directly: `import range from 'lodash/range'`

packages/pagination/src/Pagination/Pagination.types.ts
```diff
@@ -0,0 +1,51 @@
...
+    DarkModeProps {
+  /**
+   * Number of items visible on the current page.
+   */
+  itemsPerPage?: ExcludeFromTypeInference<T>;
+
+  /**
+   * Options to be shown in the Select to indicate items per page.
+   */
+  itemsPerPageOptions?: Array<T>;
```
TheSonOfThomp, line 16: Fancy 🎩

packages/pagination/src/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,177 @@
...
+  onBackArrowClick: PropTypes.func.isRequired,
+  onForwardArrowClick: PropTypes.func.isRequired,
+  numTotalItems: PropTypes.number,
+  onCurrentPageOptionChange: PropTypes.func,
+  currentPage: PropTypes.number,
+  onItemsPerPageOptionChange: PropTypes.func,
+  itemsPerPageOptions: PropTypes.arrayOf(PropTypes.number),
+  itemsPerPage: PropTypes.number,
+  // casting to unknown ensures TS does not infer types from these prop-types
+} as unknown;
```
TheSonOfThomp, line 173: 👍

TheSonOfThomp (approved):
Looks good 👍

brookescarlett (approved):

PR #1598, opened 2023-01-31
Lg 2824: Select aria-label

spark33:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/select/src/MenuButton.tsx
```diff
@@ -228,12 +228,18 @@ interface MenuButtonProps
   >;
 }
 
+type LabelProp = Pick<
```
brookescarlett, line 231: this should probably live somewhere communal at this point -- maybe in the a11y package?

packages/select/src/Select.tsx
```diff
@@ -101,13 +102,13 @@ export default function Select({
...
   const id = useIdAllocator({ prefix: 'select', id: idProp });
   const labelId = useMemo(
-    () => ariaLabelledby ?? `${id}-label`,
-    [ariaLabelledby, id],
+    () => (ariaLabel && !label ? undefined : ariaLabelledby ?? `${id}-label`),
+    [ariaLabelledby, ariaLabel, label, id],
   );
 
-  if (!label && !ariaLabelledby) {
+  if (!label && !ariaLabelledby && !ariaLabel) {
```
brookescarlett, line 109: just want to confirm that this isn't breaking anything a11y wise? (i don't expect that it would) cc @TheSonOfThomp

spark33: I did this on purpose because an element shouldn't have both `aria-label` AND `aria-labelledby` - it should really just take in one of these props and one only

packages/select/src/MenuButton.tsx
spark33: I thought about this, but I think Select is one of those components that's overdue a major refactor, so I didn't want to touch it up too much. I also feel like this is pretty relevant to the form control / input base component work we'll be doing in Q1, so we'll be doing a deeper dive on how to make cross-package changes soon anyways.

brookescarlett (approved):
Non-blocking feedback

PR #1601, opened 2023-02-02
LG-2508, LG-1676, LG-2781: Password input

shaneeza:
## ✍️ Proposed changes

[Demo](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-passwordinput--basic)

🎟 _Jira ticket:_ [LG-2508](https://jira.mongodb.org/browse/LG-2508)
🎟 _Jira ticket:_ [LG-1676](https://jira.mongodb.org/browse/LG-1676)
🎟 _Jira ticket:_ [LG-2781](https://jira.mongodb.org/browse/LG-2781)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/hooks/src/useControlledValue.tsx
```diff
@@ -0,0 +1,58 @@
+import { ChangeEventHandler, useEffect, useState } from 'react';
```
shaneeza, line 1: From @TheSonOfThomp SearchResults PR. I added it to my PR so that I can use it.

packages/password-input/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Example
+
+```js
+import { PasswordInput } from '@leafygreen-ui/password-input';
+
+// Multiple notifications
+<PasswordInput
+  label="New password"
+  stateNotifications={[
```
brookescarlett, line 29: nit: did we intentionally lean away from verificationState

packages/password-input/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+```
+
+## Special Case
+
+### Aria Labels
+
+Either `label` or `aria-labelledby` or `aria-label` must be provided a string, or there will be a console error. This is to ensure that screenreaders have a description for what the PAssword Input does.
+
+### Aria Describedby
```
brookescarlett, line 115: I wonder if because of this it's worth exporting <Error /> and <Valid /> here?

packages/password-input/src/PasswordInput/PasswordInput.spec.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+      expect(passwordInput.getAttribute('type')).toBe('password');
+      fireEvent.click(toggleButton as HTMLButtonElement);
+      expect(passwordInput.getAttribute('type')).toBe('text');
+      fireEvent.click(toggleButton as HTMLButtonElement);
+      expect(passwordInput.getAttribute('type')).toBe('password');
+    });
+  });
+
+  describe('Password input with stateNotifications array', () => {
```
brookescarlett, line 157: can we add test cases to make sure that the stateNotifications are rendered with the proper icon based on state?

packages/password-input/src/PasswordInput/PasswordInput.styles.ts
```diff
@@ -0,0 +1,325 @@
...
+
+export const inputBaseThemeStyles: Record<Theme, string> = {
+  [Theme.Light]: css`
+    &,
+    &:-webkit-autofill {
+      color: ${palette.black};
+      background: ${palette.white};
+    }
+
+    &:focus-visible {
```
brookescarlett, line 88: 😄

packages/password-input/src/PasswordInput/PasswordInput.styles.ts
```diff
@@ -0,0 +1,325 @@
...
+    font-size: ${typeScales.body1.fontSize}px;
+    line-height: ${typeScales.body1.lineHeight}px;
+    height: 28px;
+    padding: 0 ${spacing[5]}px 0 10px; //TODO: spacing tokens
+  `,
+  [SizeVariant.Default]: css`
+    font-size: ${typeScales.body1.fontSize}px;
+    line-height: ${typeScales.body1.lineHeight}px;
+    height: 36px;
+    padding: 0 ${spacing[2] * 5}px 0 ${spacing[3]}px; //TODO: spacing tokens
```
brookescarlett, line 140: what's the TODO?

packages/password-input/README.md
shaneeza: this is the first i'm hearing about verificationState, it was stateValidations previously.

packages/password-input/src/PasswordInput/PasswordInput.spec.tsx
shaneeza: yes!

packages/password-input/src/PasswordInput/PasswordInput.styles.ts
shaneeza: gonna remove

packages/password-input/src/PasswordInput/PasswordInput.spec.tsx
shaneeza: added to stateNotifications component test

packages/hooks/src/useControlledValue.tsx
TheSonOfThomp: Created a separate PR here https://github.com/mongodb/leafygreen-ui/pull/1603

shaneeza: thanks, i'll take a look!

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+} from './PasswordInput.types';
+import { getStateFromArray } from './utils';
+
+export const PasswordInput = React.forwardRef<
+  HTMLInputElement,
+  PasswordInputProps
+>(
+  (
+    {
+      value: valueProp,
```
brookescarlett, line 41: super nit but i prefer grouping aliased props and props with defaults

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+      autoComplete = 'new-password',
+      id: idProp,
+      'aria-describedby': ariaDescribedbyProp,
+      'aria-labelledby': ariaLabelledbyProp,
+      ...rest
+    }: PasswordInputProps,
+    forwardedRef,
+  ) => {
+    const [showPassword, setShowPassword] = useState<boolean>(false);
+    const prefix = 'passwordinput';
```
brookescarlett, line 58: nit: let's prefix this with 'lg-passwordinput'

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+    useEffect(() => {
+      if (disabled) setShowPassword(false);
+    }, [disabled]);
+
+    const handleTogglePasswordClick = () => setShowPassword(s => !s);
+
+    /**
+     * The overall state of the component
+     */
+    const state: States = Array.isArray(stateNotifications)
```
brookescarlett, line 87: when would stateNotifications not be an array?

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+      ? getStateFromArray(stateNotifications)
+      : stateNotifications;
+
+    /**
+     * Checks if there are any notifications
+     */
+    const hasNotifications: boolean =
+      !ariaDescribedbyProp && Array.isArray(stateNotifications);
```
brookescarlett, line 95: I think you'll want to check the array length rather than using isArray here, since you've set the prop to default to an empty array

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+    return (
+      <LeafyGreenProvider darkMode={darkMode}>
```
brookescarlett, line 98: I'd like us to document this pattern in the style guide if possible. Not blocking here, but I know you had other things that you wanted to add to the guide so noting this as well

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+        <div className={cx(className)} ref={forwardedRef}>
```
brookescarlett, line 99: nit: rm cx

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+    forwardedRef,
+  ) => {
+    const [showPassword, setShowPassword] = useState<boolean>(false);
+    const prefix = 'passwordinput';
+    const inputId = useIdAllocator({ prefix, id: idProp });
+    const ariaDescribedby = useIdAllocator({
+      prefix,
+      id: ariaDescribedbyProp,
+    });
+    const ariaLabelledby = useIdAllocator({
```
brookescarlett, line 64: Does this work here? I think if they pass an aria-labelledby we don't render a label element because their id is pointing to something already in the DOM

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+              {label}
+            </Label>
+          )}
+          <div className={inputWrapperStyles}>
+            <input
+              type={showPassword ? 'text' : 'password'}
+              value={value}
+              id={inputId}
+              autoComplete={autoComplete}
+              aria-describedby={ariaDescribedby}
```
brookescarlett, line 118: confirming this component does not support a description prop?

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+      'aria-describedby': ariaDescribedbyProp,
+      'aria-labelledby': ariaLabelledbyProp,
+      ...rest
+    }: PasswordInputProps,
+    forwardedRef,
+  ) => {
+    const [showPassword, setShowPassword] = useState<boolean>(false);
+    const prefix = 'passwordinput';
+    const inputId = useIdAllocator({ prefix, id: idProp });
+    const ariaDescribedby = useIdAllocator({
```
brookescarlett, line 60: i think we want to call this descriptionId and the below labelId... then it gets passed to the aria-describedby value on the input component, but as an id to the state validations component

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+            <TogglePassword
+              disabled={disabled}
+              showPassword={showPassword}
+              handleTogglePasswordClick={handleTogglePasswordClick}
+              sizeVariant={sizeVariant}
+            />
+          </div>
+          {hasNotifications && (
+            <StateNotifications
+              ariaDescribedby={ariaDescribedby}
```
brookescarlett, line 153: see above comment

packages/password-input/src/StateNotifications/StateNotifications.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <li
+            key={`${index}-${state}`}
+            className={cx(baseStyles, themeStyles[theme][state])}
+            aria-hidden={state === States.None}
+          >
+            <ValidationIcon
+              className={cx(iconBaseStyles, iconThemeStateStyles[theme][state])}
+            />
+            <span>{notification}</span>
+            <div className={srOnly}>
```
brookescarlett, line 56: we have a VisuallyHidden component in @leafygreen-ui/a11y that you might consider using here

packages/password-input/src/StateNotifications/StateNotifications.tsx
```diff
@@ -0,0 +1,67 @@
...
+              {/* TODO: is this message clear enough for screen readers? */}
```
brookescarlett, line 57: how can we validate?

packages/password-input/src/TogglePassword/TogglePassword.tsx
```diff
@@ -0,0 +1,38 @@
...
+        aria-disabled={disabled}
+        disabled={disabled}
+        onClick={handleTogglePasswordClick}
+        className={cx(baseStyles, sizeStyles[sizeVariant])}
+        role="switch"
+        aria-checked={showPassword ? true : false}
+      >
+        <Icon aria-hidden={true} glyph={showPasswordIcon} />
+      </IconButton>
+      <p aria-live="polite" id="password-text" className={srOnly}>
```
brookescarlett, line 31: ditto above VisuallyHidden component

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: i can update this!

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: if `aria-describedby` is passed, instead of an array of message, a string with the input state is expected.

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: Since I want the message container to render even if there are no messages I think this should be ok. The only time it should not render is if a custom message container is used.

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: noted, I started adding a few things and should have a PR up later this week.

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: if they pass in `aria-labelledby` that will be used instead of generating an ID, the label will not render, and the id will be passed to the `input` element with `aria-describedby`.

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  if (typeof props[ariaProp] === 'string') return stateStringType;
+  if (typeof props[ariaProp] === 'undefined') return arrayType;
+
+  return new Error('Error');
+};
+
+PasswordInput.propTypes = {
+  id: PropTypes.string,
+  /// @ts-ignore
```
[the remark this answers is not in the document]
shaneeza: I keep getting these type errors
```
Type 'Requireable<string>' is not assignable to type 'Validator<string>'.
  Types of property '[nominalTypeHack]' are incompatible.
    Type '{ type: string | null | undefined; } | undefined' is not assignable to type '{ type: string; } | undefined'.
      Type '{ type: string | null | undefined; }' is not assignable to type '{ type: string; }'.
        Types of property 'type' are incompatible.
          Type 'string | null | undefined' is not assignable to type 'string'.
            Type 'undefined' is not assignable to type 'string'.

1[8](https://github.com/mongodb/leafygreen-ui/actions/runs/4078324449/jobs/7028400164#step:7:9)1   label: PropTypes.string,
      ~~~~~
```

I have a hunch it's because all the label props are conditional but this error doesn't show up in the `TextInput` component. I added the ignore because its an optional string and consumers will get a warning from the check inside the component if none are used. Have you seen this before?

sooachung:
- [ ] The input text technically should be the gray base for the disabled state. Dark mode = gray dark 1.
Screen Shot 2023-02-06 at 5 44 08 PM

- [ ] Is there a way to check non-error state? default and success

packages/password-input/src/PasswordInput/PasswordInput.tsx
brookescarlett: but you're setting a default for stateNotifications to be an empty array -- so it will always be an array

packages/password-input/src/PasswordInput/PasswordInput.tsx
brookescarlett: thank you!

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: i'm a little confused, stateNotifications can either be an array or a string(one of the `States`). If a string is passed to stateNotifications then that would override the default, right?

packages/password-input/README.md
shaneeza: Are you talking about the `<Error />` component from `Typography`? I don't think we have a `valid` component. But I could make the `StateNotifications` component a separate component so that it could be used in a custom container.

shaneeza:
> * [ ]  Is there a way to check non-error state? default and success

@sooachung  I updated the stories so its easier to view each state https://shaneeza.github.io/leafygreen-ui/?path=/story/components-passwordinput--basic

sooachung:
- [ ] The placeholder text color looks off. Can we have gray light 1 for light mode, and dark 1 for dark mode if it is not already?
- [ ] The text is not on hide when it is in a "none" disabled state.
Screen Shot 2023-02-08 at 1 24 48 PM

shaneeza:
> * [ ]  The text is not on hide when it is in a "none" disabled state.
> 
> Screen Shot 2023-02-08 at 1 24 48 PM

Thats the placeholder text so its always visible

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  [cx(
+                    inputDisabledBaseStyles,
+                    inputDisabledThemeStyles[theme],
+                  )]: disabled,
+                  [inputIconSizeStyles[sizeVariant]]:
+                    !hasNotifications && state !== States.None,
+                },
+              )}
+              onChange={handleChange}
+              readOnly={disabled ? true : false}
```
shaneeza, line 137: use disabled

packages/password-input/src/PasswordInput/PasswordInput.types.ts
```diff
@@ -0,0 +1,123 @@
...
+export const States = {
+  Error: 'error', // red x icon, red text
+  Warning: 'warning', // red warning icon, red text
+  Valid: 'valid', // green checkmark, black text
+  None: 'none', // gray checkmark, gray text
+} as const;
+
+export type States = typeof States[keyof typeof States];
+
+export const SizeVariant = {
```
shaneeza, line 14: use Size, omit size from input props

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+    if (!label && !ariaLabelledbyProp && !rest['aria-label']) {
+      console.warn(
+        'For screen-reader accessibility, label, aria-labelledby, or aria-label must be provided',
+      );
+    }
+
+    // If disabled then hide password
+    useEffect(() => {
+      if (disabled) setShowPassword(false);
+    }, [disabled]);
```
spark33, line 80: Why is this default behavior? I could imagine that an input for ENV variables could be disabled but visible until made explicitly visible

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+    const labelId = useIdAllocator({
+      prefix,
+      id: ariaLabelledbyProp,
+    });
+    const { theme, darkMode } = useDarkMode(darkModeProp);
+    const { value, handleChange } = useControlledValue(valueProp, onChangeProp);
+
+    if (!label && !ariaLabelledbyProp && !rest['aria-label']) {
+      console.warn(
+        'For screen-reader accessibility, label, aria-labelledby, or aria-label must be provided',
```
brookescarlett, line 73: nit but for help in debugging add "must be provided to PasswordInput component"

packages/password-input/src/InputIcon/InputIcon.styles.ts
```diff
@@ -0,0 +1,52 @@
...
+  `,
+  [SizeVariant.Default]: css`
+    right: ${spacing[5]}px;
+  `,
+  [SizeVariant.Large]: css`
+    right: ${spacing[1] * 10}px;
+  `,
+};
+
+export const themeStateStyles: Record<Theme, Record<StateProps, string>> = {
```
brookescarlett, line 29: Should icon color change when disabled?

packages/password-input/src/PasswordInput/PasswordInput.types.ts
```diff
@@ -0,0 +1,123 @@
...
+   * Determines the font size and padding.
+   */
+
+  sizeVariant?: SizeVariant;
+
+  /**
+   * Determines whether the field is currently disabled.
+   * @default false
+   */
+  disabled?: boolean;
```
spark33, line 110: Does this need an explicit TSdoc? I feel like it could just come from HTML input props

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+    const [showPassword, setShowPassword] = useState<boolean>(false);
+    const prefix = 'lg-passwordinput';
+    const inputId = useIdAllocator({ prefix, id: idProp });
+    const descriptionId = useIdAllocator({
+      prefix,
+      id: ariaDescribedbyProp,
+    });
+    const labelId = useIdAllocator({
+      prefix,
+      id: ariaLabelledbyProp,
```
brookescarlett, line 66: ariaLabelledby means that a label exists in the DOM so I actually don't think this is what we want to be setting the default id to

packages/password-input/src/PasswordInput/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Utility function that checks if the values in an array are all equal
+ *
+ * Returns a boolean
+ *
+ * @param arr `Array<States>`
+ */
+export function allEqual(arr: Array<States>): boolean {
+  return new Set(arr).size == 1;
```
spark33, line 11: 🚀

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  )]: disabled,
+                  [inputIconSizeStyles[sizeVariant]]:
+                    !hasNotifications && state !== States.None,
+                },
+              )}
+              onChange={handleChange}
+              readOnly={disabled ? true : false}
+              {...rest}
+            />
+            {/* Visual icons inside the input will only render if aria-describedby is set and the state is not `none`. `none` does not need a visible icon */}
```
brookescarlett, line 140: This comment is misleading

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+>(
+  (
+    {
+      value: valueProp,
+      onChange: onChangeProp,
+      darkMode: darkModeProp,
+      id: idProp,
+      'aria-describedby': ariaDescribedbyProp,
+      'aria-labelledby': ariaLabelledbyProp,
+      sizeVariant = SizeVariant.Default,
```
brookescarlett, line 47: this should be called size not sizeVariant

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: oh good catch, i mixed up the names when i was addressing the first round of feedback

packages/password-input/src/StateNotifications/StateNotifications.tsx
brookescarlett: Rm comment once you validate.

spark33 (changes requested):

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: oh wait, i think this is correct, if they are passing ariaLabelledby then we want to pass that id to the input attribute `aria-labelledby` and idAllocator will not generate a new id. If ariaLabelledby is not passed then we generate an id and pass that to the input attribute `aria-labelledby` and use that as the `<Label>` id.

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: i'm silly, it could also just be `disabled`

shaneeza: but im gonna swap `readOnly` for `disabled`.

packages/password-input/src/PasswordInput/PasswordInput.styles.ts
```diff
@@ -0,0 +1,337 @@
...
+      width: 0;
+      height: 0;
+    }
+    &::-webkit-search-decoration,
+    &::-webkit-search-cancel-button,
+    &::-webkit-search-results-button,
+    &::-webkit-search-results-decoration {
+      display: none;
+    }
+  }
```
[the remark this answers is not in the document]
shaneeza: whoops, no

packages/password-input/src/PasswordInput/PasswordInput.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+    });
+
+    test('input shows error(x) icon', async () => {
+      const { container } = renderPasswordInput({
+        [`aria-describedby`]: defaultProps.ariaDescribedby,
+        stateNotifications: 'error',
+        label: defaultProps.label,
+      });
+      expect(
+        container.querySelector(`[aria-label='${iconAriaLabels.error}']`),
```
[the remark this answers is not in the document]
shaneeza: ok!

packages/password-input/src/InputIcon/InputIcon.styles.ts
shaneeza: it doesn't in `TextInput` but i'll double check with @sooachung

shaneeza: it should gray out, i'll update

packages/password-input/src/PasswordInput/PasswordInput.tsx
spark33: Ah okay. I assumed this ternary was to ensure the value is never undefined, but that works too haha

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: Hm i didn't consider that. I wonder in that case if the toggle button should never be disabled so that visibility can be toggled when the input is disabled.

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: hows this: If a custom message container is used, an icon will render inside the input to represent the state of the input. In the case that stateNotification === `none`, no icon will appear.

packages/password-input/src/StateNotifications/StateNotifications.tsx
shaneeza: it works!

shaneeza: but was suggesting that if you want to test yourself you would have to use `voiceover`

packages/password-input/src/PasswordInput/PasswordInput.styles.ts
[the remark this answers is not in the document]
shaneeza: copypasta

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: @brookescarlett

packages/password-input/src/PasswordInput/PasswordInput.tsx
TheSonOfThomp: As far as the label id goes, this looks correct (as long as the id allocator respects when `ariaLabelledbyProp` is `undefined`)

Got me thinking though, is there any difference between using `id` + `labelledby` vs `htmlFor` + `id`?
i.e. 
```html
<label id="foo"> <input aria-labelledby="foo">
vs
<label for="foo"> <input id="foo">
```

TheSonOfThomp: Actually we shouldn't need this, since the existence of `ariaLabelledbyProp` implies that there is an external element that is labelling the input. It would be an anti-pattern to pass in both `aria-labelled` by and a value for `label`. 
Regardless of what we decide, we should add documentation for this in the README

packages/password-input/src/PasswordInput/PasswordInput.tsx
TheSonOfThomp: Hmm that's a good point. For readonly passwords, I think we should be able to toggle visibility

packages/password-input/src/PasswordInput/PasswordInput.types.ts
```diff
@@ -0,0 +1,117 @@
+import { ChangeEventHandler, ComponentPropsWithoutRef } from 'react';
+
+import { DarkModeProps, Either } from '@leafygreen-ui/lib';
+
+export const States = {
```
TheSonOfThomp, line 5: nit: the name of this enum/type should be singular (e.g. `Size`, `Theme`, `Variant` etc)

It's a bit confusing to see something like `state: States = ...`. Feels like it should be an array

packages/password-input/src/PasswordInput/PasswordInput.tsx
TheSonOfThomp: I think Brooke is asking, when would someone explicitly pass in just a string vs an array? 
I think I understand, but the api is a bit confusing at first glance (passing in a single state string to a prop called `stateNotifications` feels counter intuitive)

e.g. If I want a `valid` password input with an external message, I would have to write:
```tsx
<PasswordInput
  aria-describedby="my-description"
  stateNotifications="valid"
/>
<span id="my-description" />
```

TheSonOfThomp: You could use some TS trickery where if `aria-describedby` is set, then we expect a prop called `state`

packages/password-input/src/PasswordInput/PasswordInput.tsx
TheSonOfThomp: Related, should this component accept `baseFontSize`?

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+              className={cx(labelBaseStyles, {
+                [labelLargeOverrideStyles]: size === Size.Large,
+              })}
+              htmlFor={inputId}
+              disabled={disabled}
+            >
+              {label}
+            </Label>
+          )}
+          <div className={inputWrapperStyles}>
```
TheSonOfThomp, line 112: Does the input never have a `<Description/>`?

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  [cx(
+                    inputDisabledBaseStyles,
+                    inputDisabledThemeStyles[theme],
+                  )]: disabled,
+                  [inputIconSizeStyles[size]]:
+                    !hasNotifications && state !== States.None,
+                },
+              )}
+              onChange={handleChange}
+              readOnly={disabled}
```
TheSonOfThomp, line 137: 👏

packages/password-input/src/PasswordInput/PasswordInput.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Sean & I found a way around this. I you cast the propTypes object `as unknown` it'll stop interfering with TS types

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,208 @@
...
+          )}
+        </div>
+      </LeafyGreenProvider>
+    );
+  },
+);
+
+PasswordInput.displayName = 'PasswordInput';
+
+const stateNotificationCheck = function (
```
TheSonOfThomp, line 165: Should this live in `./utils`?

packages/password-input/src/PasswordInput/PasswordInput.types.ts
```diff
@@ -0,0 +1,117 @@
...
+   * The id reference to the custom notification container
+   */
+  ['aria-describedby']: string;
+
+  /**
+   * Determines what notifications will appear
+   *
+   * @default 'none'
+   */
+  stateNotifications: States;
```
TheSonOfThomp, line 67: I alluded to this above, but the api might be more intuitive if we changed the prop name entirely in this case to `state`.

i.e.
```ts
interface StateOnlyProps {
  ['aria-describedby']: string;
  state: States;
  stateNotifications: never;
}
```

packages/password-input/src/PasswordInput/utils.ts
TheSonOfThomp: Nice 🚀 . Could we add a generic version of this to `lib`?

packages/password-input/src/PasswordInput/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @param stateNotifications `Array<NotificationProps>`
+ */
+export function getStateFromArray(
+  stateNotifications: Array<NotificationProps>,
+): States {
+  if (stateNotifications.length === 0) return States.None;
+
+  const statesArray: Array<States> = (
+    stateNotifications as Array<NotificationProps>
+  ).map((notification: NotificationProps) => notification.state);
```
TheSonOfThomp, line 28: This feels like an ideal use case for `Array.reduce`

packages/password-input/src/StateNotifications/StateNotifications.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const StateNotifications = ({
+  id,
+  notifications,
+  ...rest
+}: StateNotificationsProps) => {
+  const { theme } = useDarkMode();
+
+  return (
+    // We're using aria-polite to announce when a message has changed. In order for aria-polite to work correctly the message wrapper needs to remain on the page even if there are no messages. If a custom message container is specified with aria-describedby then this wrapper will not render.
+    <ul aria-live="polite" className={wrapperStyles} id={id} {...rest}>
```
TheSonOfThomp, line 37: Do we also need `aria-relevant`? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-relevant

packages/password-input/src/StateNotifications/StateNotifications.tsx
```diff
@@ -0,0 +1,61 @@
...
+            className={cx(baseStyles, themeStyles[theme][state])}
+            aria-hidden={state === States.None}
+          >
+            <ValidationIcon
+              className={cx(iconBaseStyles, iconThemeStateStyles[theme][state])}
+            />
+            <span>{notification}</span>
+            <VisuallyHidden aria-live="assertive">
+              {`${notification}: ${state}`}
+            </VisuallyHidden>
```
TheSonOfThomp, line 53: Not really sure how this will behave. Why do we need this in addition to the `aria-live="polite"` on the `ul`?

packages/password-input/src/TogglePassword/TogglePassword.tsx
```diff
@@ -0,0 +1,39 @@
...
+  size,
+}: TogglePasswordProps) => {
+  const showPasswordIcon = showPassword ? 'Visibility' : 'VisibilityOff';
+
+  return (
+    <>
+      <IconButton
+        aria-label="Toggle password visibility"
+        aria-disabled={disabled}
+        disabled={disabled}
```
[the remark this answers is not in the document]
TheSonOfThomp: If we're allowing users to toggle visibility when disabled, then ignore this!

packages/password-input/src/TogglePassword/TogglePassword.tsx
```diff
@@ -0,0 +1,39 @@
...
+        onClick={handleTogglePasswordClick}
+        className={cx(baseStyles, sizeStyles[size])}
+        role="switch"
+        aria-checked={showPassword ? true : false}
+      >
+        <Icon aria-hidden={true} glyph={showPasswordIcon} />
+      </IconButton>
+      <VisuallyHidden aria-live="polite" id="password-text">
+        {showPassword ? 'Password shown' : 'Password hidden'}
+      </VisuallyHidden>
```
TheSonOfThomp, line 34: 👏

packages/password-input/src/TogglePassword/TogglePassword.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { VisuallyHidden } from '@leafygreen-ui/a11y';
+import { cx } from '@leafygreen-ui/emotion';
+import Icon from '@leafygreen-ui/icon';
+import IconButton from '@leafygreen-ui/icon-button';
+
+import { baseStyles, sizeStyles } from './TogglePassword.styles';
+import { TogglePasswordProps } from './TogglePassword.types';
+
+export const TogglePassword = ({
```
TheSonOfThomp, line 11: super nit: I feel like `PasswordToggle` is a more intuitive component name

packages/password-input/src/PasswordInput.story.tsx
```diff
@@ -0,0 +1,144 @@
...
+        'aria-describedby',
+        'aria-label',
+        'id',
+      ],
+    },
+  },
+};
+
+const Template: Story<PasswordInputProps> = props => {
+  return <PasswordInput data-testid="test-id" {...props} />;
```
TheSonOfThomp, line 53: Can we add a story with an external message using `aria-describedby`?

packages/password-input/src/InputIcon/InputIcon.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const InputIcon = ({ state, size, disabled }: InputIconProps) => {
+  const ValidationIcon = validationIcons[state];
+
+  const { theme } = useDarkMode();
+
+  return (
+    <ValidationIcon
+      className={cx(
```
TheSonOfThomp, line 32: do we need `role="presentation"`?

packages/password-input/src/PasswordInput/PasswordInput.types.ts
shaneeza: oh you're right, i'll update this

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: no, i confirmed with sooa

packages/password-input/src/PasswordInput/PasswordInput.tsx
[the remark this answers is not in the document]
shaneeza: oh! this works but then i get another error, my instinct is to just add a ts-ignore since it shouldn't interfere with TS
Screen Shot 2023-02-09 at 1 29 38 PM

packages/password-input/src/InputIcon/InputIcon.tsx
shaneeza: yes!

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: To your first point, I did some [quick research](https://www.w3.org/WAI/tutorials/forms/labels/) and it looks like I can just get away with `htmlFor` and `id` -- I don't need `labelId`. 

This is how it would work if we got rid of `labelId` and `ariaLabelledby`:
- If a label is passed in then the label will show. The input `id` will be used for the labels `for` attribute; that's good enough.
- if no label is passed then its up to the consumer to pass in `ariaLabelledby`. That will get added as an attribute to the `input`.
- if both `label` and `ariaLabelledby` are passed in, I can update the code so that the `label` will not show up but the `ariaLabelledby` attribute will be added to the input.

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: oh yeah, that just clicked -- if its readonly they need something to read lol

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: yeah i can move it there

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: spoke with Dave, i'll keep the button interactive when the input is disabled.

packages/password-input/src/InputIcon/InputIcon.tsx
shaneeza: or not! the icons already have `role='img'` which apparently is correct for SVGs https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role#svg_and_roleimg

packages/password-input/src/StateNotifications/StateNotifications.tsx
shaneeza: When i was testing in voiceover the changes were not being read without this aria-live, but i'll take a second look at this.

shaneeza: i'm removing this

packages/password-input/src/StateNotifications/StateNotifications.tsx
shaneeza: i added it. I did some testing and it seems to work.

packages/password-input/src/PasswordInput/PasswordInput.types.ts
shaneeza: i initially considered this but dave brought up the point that if we do this then we would have two ways of setting state.  That could become confusing. I've been going back and forth on the name `stateNotifications` since I started but if you have any suggestions i'm open to that.

packages/password-input/src/TogglePassword/TogglePassword.tsx
```diff
@@ -20,8 +19,6 @@ export const TogglePassword = ({
     <>
       <IconButton
         aria-label="Toggle password visibility"
-        aria-disabled={disabled}
-        disabled={disabled}
         onClick={handleTogglePasswordClick}
         className={cx(baseStyles, sizeStyles[size])}
```
TheSonOfThomp, line 23: Check with design, but I wonder if the button should get darker when the input is disabled. The button looks disabled even though it isn't

TheSonOfThomp (approved):
LGTM 👁
Just a few small nits

packages/password-input/src/TogglePassword/TogglePassword.tsx
shaneeza: i agree

brookescarlett (approved):

spark33 (approved):

packages/password-input/src/PasswordInput/PasswordInput.tsx
TheSonOfThomp: I think what you describe here makes sense, though I'd think showing `label` would be a better way to go for point 3.

I don't see this implemented though.

As it stands, if a consumer passes in both a `label` and `ariaLabelledby` prop, there will be two elements in the DOM with the same `id`: our label element and whatever external element the consumer adds the id to

packages/lib/src/allEqual.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Utility function that checks if the values in an array are all equal
+ *
+ *
+ * @param arr `Array<any>`
+ * @returns boolean
+ */
+export function allEqual<T>(arr: Array<T>): boolean {
```
TheSonOfThomp, line 8: 👍

packages/a11y/src/AriaLabelProps.ts
```diff
@@ -3,11 +3,33 @@
  */
 export type AriaLabelProps =
   | {
+      /**
+       * Screen reader label text
+       *
+       * Optional if `label` or `aria-labelledby` is provided
+       *
+       */
```
TheSonOfThomp, line 11: 👍

packages/password-input/src/PasswordInput/PasswordInput.tsx
shaneeza: oh i missed this, i'll update

PR #1605, opened 2023-02-06
[LG-2858] Aria label props

TheSonOfThomp:
## ✍️ Proposed changes

Exports `AriaLabelProps` from `a11y`, a union interface requiring _either_ `aria-label` or `aria-labelledby`

Used in `SearchInput` PR: https://github.com/mongodb/leafygreen-ui/pull/1528

## 🛠 Types of changes

## ✅ Checklist


packages/a11y/src/AriaLabelProps.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * A union interface requiring _either_ `aria-label` or `aria-labelledby`
+ */
+export type AriaLabelProps =
```
shaneeza, line 4: Would this only be for a component that doesn't accept the `label` prop?

TheSonOfThomp: Yeah, in that case you'd extend this interface

shaneeza: maybe we should also expose an interface that uses `Pick` for when the three are optional

shaneeza (approved):

packages/a11y/src/AriaLabelProps.ts
TheSonOfThomp: Added `AriaLabelPropsWithLabel`

shaneeza: oops, meant to say `Either` but this also works!

shaneeza (approved):

packages/a11y/src/AriaLabelProps.ts
TheSonOfThomp: I think Either only works with 2 keys. It's easier to export this from here anyway I think

packages/a11y/src/AriaLabelProps.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * A union interface requiring _either_ `label`, `aria-label` or `aria-labelledby`
+ */
+export type AriaLabelPropsWithLabel =
+  | ({
+      label?: string;
+    } & AriaLabelProps)
+  | ({
+      label: string;
+    } & Partial<AriaLabelProps>);
```
shaneeza, line 23: oh nice approach

PR #1611, opened 2023-02-07
LG-733: Unsets `pointer-events:"none"` when Button components are disabled

brookescarlett:
## ✍️ Proposed changes

Unsets `pointer-events:"none"` when Button components are disabled. We do this so that disabled buttons can still be triggers for Tooltips. Right now, the only event we are manually preventing from firing when the Button component is disabled is an `onClick` but if there are other mouse events that we want to filter out, let me know! I'm marking this as a major change because it unsets "disabled" in the DOM, and if tests are relying on this, they will break. (Open to making this a patch, also let me know feedback there)

🎟 _Jira ticket:_ [LG-733](https://jira.mongodb.org/browse/[LG-733)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


TheSonOfThomp:
Screen Shot 2023-02-07 at 11 41 44 AM
Screen Shot 2023-02-07 at 11 43 48 AM

The button still shows a ring on hover, and still gets focused on click

packages/button/src/Button/Button.spec.tsx
```diff
@@ -61,11 +61,10 @@ describe('packages/button', () => {
       expect(button.title).toBe(title);
     });
 
-    test(`renders the disabled and aria-disabled attributes when disabled is set`, () => {
+    test(`renders aria-disabled attribute when disabled is set`, () => {
       const { button } = renderButton({
         disabled: true,
       });
-      expect((button as HTMLButtonElement).disabled).toBe(true);
```
spark33, line 68: I can understand why we might want to remove `pointer-events` from the CSS, but why do we remove the `disabled` attribute?

spark33: Oh I see it plays with why we're no longer checking for the anchor type in the Button component. I see. But shouldn't we still be setting this attribute for <button>s?

brookescarlett: setting aria-disabled instead.

spark33: Those are only the same semantically though. We should make sure we have parity in the ways they're different: 
image

spark33 (approved):

PR #1618, opened 2023-02-08
LG-2826: Don't filter out props being passed to Link component

brookescarlett:
## ✍️ Proposed changes

Does not filter out props passed to Link component

🎟 _Jira ticket:_ [LG-2826](https://jira.mongodb.org/browse/LG-2826)

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


packages/typography/src/Link/Link.spec.tsx
```diff
@@ -155,4 +155,47 @@ describe('packages/typography', () => {
       expect(div!.tagName.toLowerCase()).toBe('div');
     });
   });
+
+  describe('anchor props', () => {
+    test('href is always passed to rendered component when the prop is supplied', () => {
+      renderLink({ href: 'test' });
+      const div = screen.getByText('Link').parentElement;
+      expect(div).toHaveAttribute('href');
```
spark33, line 163: Why do we expect the div to have the href here?

packages/typography/src/Link/Link.spec.tsx
```diff
@@ -155,4 +155,47 @@ describe('packages/typography', () => {
...
+      expect(div?.getAttribute('href')).toBe('test');
+    });
+
+    test('it respects "target" and "rel" values when target is set', () => {
```
spark33, line 167: Not sure if I like these test names since they imply they're both being set from target

packages/typography/src/Link/Link.tsx
```diff
@@ -50,16 +51,21 @@ const Link = Polymorphic<LinkProps>(
...
       target = targetProp;
+      rel = relProp;
     } else {
-      if (hrefHostname === currentHostname) {
-        target = '_self';
-      } else {
-        target = '_blank';
-        rel = 'noopener noreferrer';
+      // Only manually set values if not provided and we are rendering an anchor tag
+      if (Component === 'a') {
```
spark33, line 62: I'm assuming this is also inferring that usePolymorphic is returning `a` as the Component href is set in props

packages/typography/src/Link/Link.spec.tsx
```diff
@@ -155,4 +155,47 @@ describe('packages/typography', () => {
...
+    test('it sets "target" and "rel" values when href is set', () => {
+      renderLink({ href: 'http://mongodb.design' });
+      const anchor = screen.getByText('Link').parentElement;
+      expect(anchor).toHaveAttribute('rel');
+      expect(anchor).toHaveAttribute('target');
+      expect(anchor).toHaveAttribute('href');
+      expect(anchor?.tagName.toLowerCase()).toBe('a');
+    });
+  });
 });
```
spark33, line 201: I don't feel strongly about these, but calling out that we're not testing the values of `target` and `rel` based on the `href` value.

spark33 (approved):

packages/typography/src/Link/Link.spec.tsx
brookescarlett: We still do that above

packages/typography/src/Link/Link.tsx
brookescarlett: confirmed

packages/typography/src/Link/Link.spec.tsx
brookescarlett: the point is that we are no longer being judicious about what can / cannot receive an href prop if one is passed to the component.

packages/typography/src/Link/Link.tsx
```diff
@@ -40,21 +53,22 @@ const Link = Polymorphic<LinkProps>(
...
+    }, [rest, currentHostname]);
 
-    let target, icon, rel;
+    let target, rel, icon;
 
-    if (targetProp) {
-      target = targetProp;
-    } else {
+    // Sets defaults for target and rel props when Component is an anchor tag
+    if (Component === 'a') {
```
TheSonOfThomp, line 71: Nit, but I don't know if I like these uninitialized variables here. I'd prefer see something like below, or even a function that returns these default props. 
```js
const defaultAnchorProps = {
  target: <logic>,
  rel: <logic>
}
```

packages/typography/src/Link/Link.tsx
```diff
@@ -21,15 +26,23 @@ import {
...
-const Link = Polymorphic<LinkProps>(
+type LinkRenderProps = PolymorphicPropsWithRef<'span', LinkProps>;
+
+type AnchorLikeProps = PolymorphicProps<'a', LinkProps>;
+
+const hasAnchorLikeProps = (
+  props: LinkRenderProps | AnchorLikeProps,
+): props is AnchorLikeProps => {
+  return (props as AnchorLikeProps).href !== undefined;
+};
```
TheSonOfThomp, line 37: Should we export `AnchorLike` from `Polymorphic` instead?

packages/typography/src/Link/Link.spec.tsx
```diff
@@ -148,11 +156,85 @@ describe('packages/typography', () => {
...
       expect(div!.tagName.toLowerCase()).toBe('div');
     });
   });
+
+  describe('anchor props', () => {
+    test('href is always passed to rendered component when the prop is supplied', () => {
+      renderLink({ href: 'test' });
+      const a = screen.getByText('Link').parentElement;
+      expect(a).toHaveAttribute('href');
+      expect(a?.getAttribute('href')).toBe('test');
```
TheSonOfThomp, line 173: should also expect `tagName` to be `a`

packages/typography/src/Link/Link.spec.tsx
```diff
@@ -148,11 +156,85 @@ describe('packages/typography', () => {
...
+    });
+
+    test('it respects "target" and "rel" values when target is set', () => {
```
TheSonOfThomp, line 176: Can we group these `respects "target"` tests in a `describe`?

packages/typography/src/Link/Link.spec.tsx
```diff
@@ -148,11 +156,85 @@ describe('packages/typography', () => {
...
+      renderLink({ href: 'http://mongodb.design' });
+      const anchor = screen.getByText('Link').parentElement;
+      expect(anchor).toHaveAttribute('rel');
+      expect(anchor).toHaveAttribute('target');
+      expect(anchor).toHaveAttribute('href');
+      expect(anchor?.tagName.toLowerCase()).toBe('a');
+    });
+  });
+
+  describe('TypeSCript types are correct', () => {
```
TheSonOfThomp, line 211: typo

packages/typography/src/Link/Link.spec.tsx
```diff
@@ -148,11 +156,85 @@ describe('packages/typography', () => {
     test('it renders the correct tag to the DOM', () => {
```
TheSonOfThomp, line 156: Should we add more tests for the `as` prop? It's a bit redundant since `poly` tests it, but it ensures we're properly extending polymorphic

packages/typography/src/Link/Link.tsx
brookescarlett: We could if you think it'll be helpful in other places?

packages/typography/src/Link/Link.spec.tsx
brookescarlett: reworked those tests a little bit, lmk what you think

packages/typography/src/Link/Link.spec.tsx
```diff
@@ -141,100 +118,99 @@ describe('packages/typography', () => {
...
+        expect(anchor!.tagName.toLowerCase()).toBe('a');
+      });
 
-    test('it respects "target" and "rel" values when target is set', () => {
-      renderLink({ href: 'string', target: 'test' });
-      const a = screen.getByText('Link').parentElement;
-      expect(a).toHaveAttribute('target');
-      expect(a).not.toHaveAttribute('rel');
-      expect(a?.getAttribute('target')).toBe('test');
+      test('by default it renders as a span', () => {
```
TheSonOfThomp, line 141: Nit: the test name is a bit confusing. I'd maybe change it to "renders as a span when no `as` or `href` is provided"

packages/typography/src/Link/Link.tsx
```diff
@@ -65,19 +65,30 @@ const Link = InferredPolymorphic<LinkProps, 'span'>(
...
+      JSX.IntrinsicElements['a'],
+      'target' | 'rel'
+    > = {
+      target: undefined,
+      rel: undefined,
+    };
 
-    // Sets defaults for target and rel props when Component is an anchor tag
-    if (Component === 'a') {
+    if ((rest as AnchorLikeProps).target || (rest as AnchorLikeProps).rel) {
```
TheSonOfThomp, line 78: This check is getting a bit verbose. Could consider extracting it to it's own variable
(i.e. `const hasTargetOrRel = (rest as AnchorLikeProps).target || (rest as AnchorLikeProps).rel)`)

packages/typography/src/Link/Link.tsx
```diff
@@ -65,19 +65,30 @@ const Link = InferredPolymorphic<LinkProps, 'span'>(
...
+      defaultAnchorProps.target = (rest as AnchorLikeProps).target;
+      defaultAnchorProps.rel = (rest as AnchorLikeProps).rel;
```
TheSonOfThomp, line 80: It's the same number of lines, so take it or leave it, but I find this cleaner:
```
const {target, rel} = rest as AnchorLikeProps
defaultAnchorProps = {target, rel}
```

TheSonOfThomp (approved):
LGTM

packages/typography/src/Link/Link.spec.tsx
brookescarlett: 👍

packages/typography/src/Link/Link.tsx
brookescarlett: I wanted to leave target and rel in rest so that they override our default settings

PR #1620, opened 2023-02-09
Combobox Fix

brookescarlett:
## ✍️ Proposed changes

Combobox no longer automatically opens when the IconButton "X" is clicked.

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)


spark33 (approved):

packages/combobox/src/Combobox.tsx
```diff
@@ -754,12 +754,11 @@ export function Combobox<M extends boolean>({
       e: React.MouseEvent<HTMLButtonElement, MouseEvent>,
     ) => {
       if (!disabled) {
+        // Prevents triggering the setOpen function called by clicking anywhere within the input wrapper.
```
TheSonOfThomp, line 757: Can we add a test for this?

packages/combobox/src/Combobox.tsx
```diff
@@ -754,12 +754,11 @@ export function Combobox<M extends boolean>({
...
+        e.stopPropagation();
```
TheSonOfThomp, line 758: Do we want to stop propagation here? Will a user potentially want to handle the click event?

brookescarlett: A user can with onClear. If we don't stopPropagation here then the event will bubble and trigger the function called when a user clicks on the wrapper, which forces the menu to open.

packages/combobox/src/Combobox.spec.tsx
```diff
@@ -1299,6 +1299,18 @@ describe('packages/combobox', () => {
       expect(onClear).toHaveBeenCalled();
     });
 
+    test('Clear button does not force the menu to reopen', () => {
```
brookescarlett, line 1302: @TheSonOfThomp added a test! RFAL

brookescarlett: Why would clicking the clear button focus the input?

TheSonOfThomp: I believe that's how native search works, and the recommended w3 Combobox guidelines

TheSonOfThomp: HTML search clear button focuses the input, but [Combobox guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) don't mention a clear button

https://user-images.githubusercontent.com/2414030/218155592-a7116800-43fc-48aa-9daa-5f8143810dce.mov

packages/combobox/src/Combobox.tsx
```diff
@@ -759,9 +759,15 @@ export function Combobox<M extends boolean>({
...
         onFilter?.('');
+        setInputFocus();
       }
     };
 
+    const handleClearButtonKeyDown = (e: React.KeyboardEvent) => {
+      e.stopPropagation();
+      setInputFocus();
+    };
+
```
TheSonOfThomp, line 770: I think this is overriding the `click` behavior when using the keyboard. I think anything other than `stopPropagation` should be handled in the click handler.

(Actually, if we want the arrow key navigation, I'm not even sure we want to stop propagation)


https://user-images.githubusercontent.com/2414030/218544399-a8cca362-986b-4ae3-ad2b-310aca7d4577.mov

packages/combobox/src/Combobox.spec.tsx
```diff
@@ -1288,27 +1288,52 @@ describe('packages/combobox', () => {
...
+
+      test('Clear button clears the value of the input', () => {
+        const value = 'apple';
+        const { inputEl, clearButtonEl } = renderCombobox(select, {
+          value,
+        });
+
+        expect(inputEl).toHaveValue('Apple');
+        userEvent.click(clearButtonEl!);
+        expect(inputEl).toHaveValue('');
```
TheSonOfThomp, line 1323: I don't think this test should be passing. If you provide `value` (i.e. controlled component), then the clear button shouldn't do anything until you update the value prop externally

packages/combobox/src/Combobox.spec.tsx
```diff
@@ -1140,14 +1140,14 @@ describe('packages/combobox', () => {
       });
 
       describe('Right arrow key', () => {
-        test('Does nothing when focus is on clear button', () => {
+        test('Focuses on inputEl when on clear button', () => {
```
TheSonOfThomp, line 1143: I'm not sure about this one. If we do want the focus to cycle around, should it always focus on the input, or on the first chip?

TheSonOfThomp: Also, if this is the behavior we want, pressing left from the first chip in MS should focus the clear button. (and, potentially, `left` from the beginning of the text in SS)

TheSonOfThomp (changes requested):

packages/combobox/src/Combobox.tsx
brookescarlett: stopping prop doesn't prevent the keyboard nav

packages/combobox/src/Combobox.spec.tsx
brookescarlett: Resetting this test, bc per above no longer manually setting focus back to input onKeydown

brookescarlett:
re-r @TheSonOfThomp

packages/combobox/src/Combobox.tsx
TheSonOfThomp: It should though? Since the Arrow Key nav is handled via a `onKeyDown` handler on the wrapper, if we stop prop. on the button, the wrapper won't receive the event

brookescarlett: Those are separate events though. We don't want onClear opening the menu

packages/combobox/src/Combobox.spec.tsx
brookescarlett: Updated the test

packages/combobox/src/Combobox.tsx
```diff
@@ -754,15 +754,19 @@ export function Combobox<M extends boolean>({
...
-        if (!isOpen) {
-          openMenu();
-        }
+        setInputFocus();
       }
     };
 
+    const handleClearButtonKeyDown = (e: React.KeyboardEvent) => {
+      e.stopPropagation();
+    };
```
TheSonOfThomp, line 768: This still seems to be blocking Arrow key nav (in Multiselect story: make a selection, tab over to the clear button. Left arrow key should focus the input)

TheSonOfThomp: The test that checks this is skipped for some reason... https://github.com/mongodb/leafygreen-ui/blob/brooke/fix-combobox/packages/combobox/src/Combobox.spec.tsx#L1102-L1111

packages/combobox/src/Combobox.spec.tsx
```diff
@@ -1099,15 +1099,14 @@ describe('packages/combobox', () => {
           waitFor(() => expect(inputEl).toHaveFocus());
         });
 
-        // eslint-disable-next-line jest/no-disabled-tests
-        test.skip('When focus is on clear button, Left arrow moves focus to input', async () => {
+        test('When focus is on clear button, Left arrow moves focus to input', async () => {
```
TheSonOfThomp, line 1102: 👍

TheSonOfThomp (approved):

PR #1634, opened 2023-02-23
passwordInput: remove ts-ignore from proptypes

shaneeza:
## ✍️ Proposed changes

## 🛠 Types of changes

## ✅ Checklist

## Feedback Levels

1. Praise (👏 🚀 🔥 🍾)

### DIFF
--- a/.changeset/chilled-falcons-ring.md
+++ b/.changeset/chilled-falcons-ring.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/card': patch
+---
+
+Fixes propTypes in Card component, such that contentStyle is accurate
--- a/packages/card/src/Card/Card.tsx
+++ b/packages/card/src/Card/Card.tsx
@@ -49,5 +49,5 @@ Card.displayName = 'Card';
 Card.propTypes = {
   className: PropTypes.string,
   darkMode: PropTypes.bool,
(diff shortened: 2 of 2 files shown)


packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -171,4 +169,4 @@ PasswordInput.propTypes = {
   size: PropTypes.oneOf(Object.values(Size)),
   value: PropTypes.string,
   stateNotifications: stateNotificationCheck,
-} as unknown;
+} as any;
```
shaneeza, line 172: When using `unknown` i kept getting the error 

```Type 'unknown' is not assignable to type 'WeakValidationMap<PasswordInputProps & RefAttributes<HTMLInputElement>> | undefined'.ts(2322)```

but I added `ts-ignore` above to ignore it, but switching to `any` removes that warning. I was using `unknown` before to stop interfering with TS types but I think i could also get away with that using `any`. @TheSonOfThomp @spark33 thoughts?

TheSonOfThomp: Ah yeah, I think `as unknown` only works for regular components, not `forwardRef`. This should work for now

TheSonOfThomp (approved):

PR #1635, opened 2023-02-24
LG-2774 Multi-toast

TheSonOfThomp:
## ✍️ Proposed changes

Demo: https://thesonofthomp.github.io/leafygreen-ui-fork/?path=/story/components-toast--variants

🎟 _Jira ticket:_ [Multi-toast](https://jira.mongodb.org/browse/[LG-2774])

## ✅ Checklist

## 🧪 How to test changes

I recommend reviewing in this order:

1. Start with reviewing `InternalToast` - this is the stateless component
2. Next review ToastContext & Provider - This manages the state of toasts & 
3. Look at `useToastReducer` - The internal state manager
4. `useToast` - the external facing hook for state management
5. `ToastContainer` - The child component rendered inside `ToastProvider`. Responsible for rendering the toast stack
6. Finally look at `ControlledToast` - this is the component that allows consumers to render a toast from JSX (not the recommended approach, but required for backwards compat.)

## Notes:
- The animation when collapsing an expanded stack of 4+ toasts is a little jittery, but barely noticeable, and I feel that since this is already an edge case, this slight imperfection is acceptable. Please comment if you feel otherwise.


.changeset/silly-pets-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/hooks': minor
+---
+
+Creates `useStateRef` - a combination of useState and useRef, that returns the current state, a `setState` function, and a `getState` function that will return the current walue of the state. This is useful to avoid referencing stale state inside callbacks
```
brookescarlett, line 5: will also need a changeset for 🍞  :)

packages/hooks/src/useStateRef.ts
```diff
@@ -0,0 +1,30 @@
...
+  function setState(newVal: T): void {
+    _setState(newVal);
+    ref.current = newVal;
+  }
+
+  function getState(): T {
+    return ref.current;
+  }
+
+  return [state, setState, getState];
```
brookescarlett, line 29: If I'm understanding this correctly, sometimes state and getState will not be in sync?

packages/toast/src/ControlledToast/ControlledToast.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * It's recommended to use the hook `useToast` instead
+ */
+export const ControlledToast = ({ open, ...props }: ControlledToastProps) => {
+  props = defaultsDeep(props, defaultToastProps);
+  const { pushToast, popToast } = useToast();
+  const [toastId, setToastId] = useState<ToastId | null>(null);
+
+  useEffect(() => {
+    if (open && !toastId) {
+      const _id = pushToast({ isControlled: true, ...props });
```
brookescarlett, line 23: nit: why `_id`

brookescarlett:
One thing I'm noticing: When you close a Toast anywhere within the notification bar, the other toasts pop up a little and then come down. This even happens when the top toast is cleared. Not sure if there's a way to make this feel a bit smoother.

packages/toast/src/ToastContext.story.tsx
```diff
@@ -0,0 +1,192 @@
...
+        <Button onClick={() => clearStack()}>Clear all</Button>
+      </div>
+      {progressToasts && progressToasts.length > 0 && (
+        <>
+          <Label htmlFor="progress">Progress: </Label>
+          <InlineCode>{progress}</InlineCode>
+          <br />
+          <input
+            value={progress}
+            onChange={e => {
```
brookescarlett, line 152: should we be defining this not inline

packages/hooks/src/useStateRef.ts
TheSonOfThomp: During a render they will be in sync, but if you want to reference `state` in an event handler, there's a chance it could be stale

packages/toast/src/ControlledToast/ControlledToast.tsx
TheSonOfThomp: just to make it clear it's a temp variable

packages/toast/src/ToastContext.story.tsx
TheSonOfThomp: Technically yes, but it's in the story, so I'm less worried about this one

packages/toast/src/ControlledToast/ControlledToast.story.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <Toast
+        {...args}
+        open={open}
+        onClose={() => {
+          setOpen(false);
+        }}
+      />
+    </ToastProvider>
+  );
+};
```
spark33, line 67: nit: since there's only one story in here, this could probably just live in `Toast.tsx` without needing a story subdir

packages/toast/src/ControlledToast/ControlledToast.types.ts
```diff
@@ -0,0 +1,17 @@
...
+export interface ControlledToastProps extends ToastProps {
+  /**
+   * Required boolean that renders the Toast and makes it visible when true.
+   *
+   * Note: open is not a valid prop when rendering toasts programmatically
+   *
+   * @default false
+   */
+  open: boolean;
+  onClose?: React.EventHandler<any>;
```
spark33, line 16: TSDoc

packages/toast/src/InternalToast/InternalToast.spec.tsx
```diff
@@ -0,0 +1,206 @@
...
+        <InternalToast
+          title="hello"
+          description={bodyElement}
+          variant={Variant.Success}
+        />,
+      );
+
+      const bodySpan = queryByText(bodyText);
+      expect(bodySpan).toBeVisible();
+    });
```
spark33, line 167: nit: maybe test by test-id to check that the span renders too, not just the body text

packages/toast/src/InternalToast/InternalToast.styles.ts
```diff
@@ -12,37 +11,38 @@ import {
 } from '@leafygreen-ui/tokens';
 import { anchorClassName } from '@leafygreen-ui/typography';
 
-import { Variant } from './Toast.types';
+import { TOAST } from '../constants';
```
spark33, line 14: nit: I was kinda confused when I saw this import statement because I didn't know what `TOAST` meant - is it a component? or an instance? Why is it all in caps? Maybe a name like `ToastConstants` is better

packages/toast/src/InternalToast/InternalToast.styles.ts
```diff
@@ -84,7 +84,7 @@ export const toastThemeStyles: Record<Theme, string> = {
     }
   `,
   [Theme.Dark]: css`
-    background-color: ${palette.gray.light2};
+    background-color: ${toastBGColor[Theme.Dark]};
```
spark33, line 87: I see `toastBGColor` is only being referenced here. The colors could just be here instead of being in a separate object

packages/toast/src/InternalToast/InternalToast.tsx
```diff
@@ -0,0 +1,141 @@
...
+      description,
+      className,
+      onClose,
+      action,
+      index = 0,
+      isHovered,
+      isControlled,
+      variant: variantProp,
+      progress: progressProp,
+      dismissible: dismissProp,
```
spark33, line 53: nit: `dismissableProp`

packages/toast/src/InternalToast/InternalToast.tsx
```diff
@@ -0,0 +1,141 @@
+import React from 'react';
+import { defaults } from 'lodash';
```
spark33, line 2: `import defaults from 'lodash/defaults'`?

packages/toast/src/InternalToast/InternalToast.tsx
```diff
@@ -0,0 +1,141 @@
...
+        {...rest}
+      >
+        <div
+          data-testid="lg-toast-content"
+          aria-hidden={!showContent}
+          className={cx(
+            contentWrapperStyle,
+            css`
+              opacity: ${showContent ? 1 : 0};
+            `,
```
spark33, line 86: Move to styles file?

packages/toast/src/InternalToast/defaultProps.ts
```diff
@@ -0,0 +1,10 @@
+import { ToastProps, Variant } from '../Toast.types';
+
+export const defaultToastProps: Required<
+  Pick<ToastProps, 'variant' | 'progress' | 'dismissible' | 'timeout'>
+> = {
+  variant: Variant.Note,
+  progress: 1.0,
+  timeout: 6_000,
+  dismissible: true,
+};
```
spark33, line 10: nit: I assume these are defaultProps for just the story file? It'd be nice if that was indicated.

spark33:
On the `Variant` story, pushing 4 toasts and closing the top one causes a little hiccup in the animation

packages/toast/src/InternalToast/InternalToast.styles.ts
spark33: nvm, I see it's being referenced from other styles files.

packages/toast/src/ToastContainer/ToastContainer.styles.ts
```diff
@@ -0,0 +1,187 @@
...
+      `]: isInteracted,
+
+      [css`
+        // We want auto scroll bars when expanded
+        height: 100vh;
+        overflow: auto;
+      `]: shouldExpand,
+    },
+  );
+}
```
spark33, line 92: nit: it feels like this `cx` call should live on the component itself

spark33: Generally feels like there are few functions in this file that could/should be `cx` calls instead of functions, or functions that are called in `cx` calls inside the component file. Similar to the cx call in [this](https://github.com/mongodb/leafygreen-ui/pull/1668/files#diff-0dfe10c35fa37d766d24e5444f598c29c9ed027ed2821cc557598175064f3205) component

packages/toast/src/ToastContainer/ToastContainer.tsx
```diff
@@ -0,0 +1,282 @@
...
+  const { popToast, getToast } = useToast();
+  const regionId = useIdAllocator({ id: 'lg-toast-region' });
+  const stackSize = stack.size;
+  const toastContainerRef = useRef<HTMLDivElement>(null);
+  const scrollContainerRef = useRef<HTMLDivElement>(null);
+  const getToastRef = useDynamicRefs<HTMLDivElement>({ prefix: 'toast' });
+  const { theme } = useDarkMode();
+  const [isHovered, setHoveredState] = useState(false);
+  const setHovered = () => setHoveredState(true);
+  const setUnhovered = () => setHoveredState(false);
```
spark33, line 60: Are two state variables needed here? Could you have one state typed with `boolean | undefined` instead?

packages/toast/src/ToastContainer/ToastContainer.tsx
```diff
@@ -0,0 +1,282 @@
...
+                              theme,
+                              bottomOffset: notifBarSpacing,
+                              isExpanded, //: shouldExpand,
+                            })]: isInteracted,
+                          },
+                          className,
+                        )}
+                        description={
+                          // TODO: Remove debug strings
+                          toastProps.description + ` (${id} index #${index}.)`
```
spark33, line 257: rm?

shaneeza:
I'm running into a bug in firefox -- when you have a few toasts open, and they auto dismiss there is a gap between the bottom of the page and the last toast. The more toasts you have in the list, the larger the gap becomes. This might be an edge case. 
Screenshot 2023-03-27 at 4 26 36 PM

packages/toast/src/ToastContext/ToastContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { ToastContextProps, ToastProviderProps } from './ToastContext.types';
+import { useToastReducer } from './ToastReducer';
+
+export const ToastContext = React.createContext<ToastContextProps>({
+  pushToast: () => '',
+  popToast: () => undefined,
+  updateToast: () => undefined,
+  getToast: () => undefined,
+  getStack: () => undefined,
```
spark33, line 13: nit: `() => void`?

packages/toast/src/ToastContext/ToastContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+    useToastReducer(initialValue);
+
+  return (
+    <ToastContext.Provider
+      value={{
+        pushToast,
+        popToast,
+        updateToast,
+        getToast,
+        getStack: () => stack,
```
spark33, line 31: this might be a dumb q, but could it not just pass `stack` instead of a `getStack` function?

packages/toast/src/ControlledToast/ControlledToast.spec.tsx
```diff
@@ -0,0 +1,171 @@
...
+          />
+        </ToastProvider>,
+      );
+
+      await waitForElementToBeRemoved(toast);
+      expect(toast).not.toBeInTheDocument();
+    });
+
+    // eslint-disable-next-line jest/expect-expect
+    test('does not close until manually closed', async () => {
```
shaneeza, line 118: If you set a `timeout`, shouldn't it auto-close?

packages/toast/src/InternalToast/InternalToast.styles.ts
shaneeza: I think the all-caps make sense, but I agree that the name can be confusing.

packages/toast/src/InternalToast/InternalToast.styles.ts
```diff
@@ -12,37 +11,38 @@ import {
...
+  bottom: ${TOAST.inset}px;
+  width: ${TOAST.width}px;
+  min-height: ${TOAST.minHeight - 2}px; // -2 for border: ;
 
   display: flex;
   align-items: center;
   justify-content: space-between;
-  padding: ${spacing[2]}px;
+  padding: ${spacing[2] - 1}px; // -1 for border
   padding-left: ${spacing[3]}px;
```
shaneeza, line 33: Can this be combined in the shorthand above?

packages/toast/src/InternalToast/InternalToast.tsx
```diff
@@ -0,0 +1,141 @@
...
+        {dismissible && (
+          <IconButton
+            className={cx(dismissButtonStyle, dismissButtonThemeStyle[theme])}
+            aria-label="Close Message"
+            onClick={onClose}
+            darkMode={!darkMode}
+            data-testid="lg-toast-dismiss-button"
+          >
+            <XIcon />
+          </IconButton>
```
shaneeza, line 122: Should the icon have aria-hidden? The button already has `aria-label` on it.

packages/toast/src/ToastContext/ToastContext.spec.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { render, waitFor } from '@testing-library/react';
+
+import { makeToast, makeToastStack } from './utils/makeToast';
+import { ToastStack } from './ToastContext.types';
+import { ToastProvider } from '.';
+
+/**
+ * This suite exclusively tests the context its
```
shaneeza, line 9: I don't think this sentence is complete

packages/toast/src/ToastContainer/ToastContainer.tsx
```diff
@@ -0,0 +1,282 @@
...
+        aria-relevant="all"
+        data-hovered={isHovered}
+        data-should-expand={shouldExpand}
+        data-expanded={isExpanded}
+        onMouseEnter={setHovered}
+        onMouseLeave={setUnhovered}
+        onFocus={setHovered}
+        onBlur={setUnhovered}
+        className={cx(toastContainerStyles, {
+          [toastContainerVisibleStyles]: stack.size > 0,
```
shaneeza, line 191: nit: you could combine these conditionals since they both check for `stack.size > 0`
```
[cx(
  toastContainerVisibleStyles,
  getContainerStatefulStyles({
    isExpanded,
    isHovered,
    shouldExpand,
    totalStackHeight,
    topToastHeight: toastHeights[0],
    recentToastsLength: recentToasts.length,
    bottomOffset: notifBarSpacing,
  }),
)]: stack.size > 0,
```

packages/toast/src/ControlledToast/ControlledToast.story.tsx
TheSonOfThomp: I left it here since the StoryMeta is quite different, and we may want to add more stories in the future

TheSonOfThomp: We _could_ move the file up one dir level, but I feel like it makes more sense here with the specific component files

packages/toast/src/Toast.types.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Optional action button (only rendered if variant==='progress')
+   *
+   * This should be a shortcut only—and not the _only_ way to perform the action.
+   *
+   * The provided action element should have an `aria-description` attribute
+   * that describes how to alternatively perform the action
+   */
+  action?: React.ReactNode;
```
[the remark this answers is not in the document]
TheSonOfThomp: 👍 Using `actionElement`

packages/toast/src/ToastContainer/ToastContainer.styles.ts
TheSonOfThomp: I'll move this last one (`shouldExpand`) to the component itself, but I felt moving it all to the component might cause a lot of clutter

packages/toast/src/ToastContext/ToastContext.tsx
TheSonOfThomp: Since we don't necessarily want users to access the stack itself and manipulate it directly

packages/toast/src/ControlledToast/ControlledToast.spec.tsx
TheSonOfThomp:
```

```

packages/toast/src/ControlledToast/ControlledToast.story.tsx
spark33: Yeah it's fine if you want to keep it - I only call it out because it's not really best practice. I believe Storybook only considers different configurations of different metas necessary when the story is about a new component altogether (so they can generate the right controls for it), not for when the arguments need to be different, even if they are quite different. Story-specific configurations should really be done through `ControlledToast.args` or `ControlledToast.argTypes`

packages/toast/src/ToastContainer/ToastContainer.tsx
spark33: My fault! I misread. I thought there were two different state variables in use here.

packages/toast/src/ControlledToast/ControlledToast.story.tsx
TheSonOfThomp: It is _technically_ a different component (`ControlledToast`)

packages/toast/src/InternalToast/InternalToast.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const iconThemeStyle = variantIconStyle[variant];
+
+    return (
+      <LeafyGreenProvider darkMode={!darkMode}>
+        <div
+          id={id}
+          ref={forwardedRef}
+          className={cx(baseToastStyle, toastThemeStyles[theme], className)}
+          aria-atomic="true"
+          data-testid="lg-toast"
```
spark33, line 99: nit - this feels like it should be passed in from the test files. feel free to ignore if it adds a lot of overhead to the test suites

packages/toast/src/Toast.story.tsx
```diff
@@ -0,0 +1,210 @@
...
+  component: InternalToast,
+  decorators: [
+    (Story, meta) => (
+      <LeafyGreenProvider darkMode={!!meta.args.darkMode}>
+        <ToastProvider initialValue={meta.args.initialValue}>
+          <Story />
+        </ToastProvider>
+      </LeafyGreenProvider>
+    ),
+  ],
```
spark33, line 28: 🚀

packages/toast/src/ToastContainer/NotificationBar/NotificationBar.styles.ts
```diff
@@ -0,0 +1,76 @@
...
+  entering: css`
+    transform: translate3d(0, 0, 0);
+    opacity: 1;
+  `,
+  exited: css`
+    transform: translate3d(0, ${3 * TOAST_CONSTANTS.yOffset}px, -400px);
+    opacity: 0;
+  `,
+  exiting: css`
+    transform: translate3d(0, ${3 * TOAST_CONSTANTS.yOffset}px, -400px);
```
spark33, line 73: nit - `-400px` could use a comment

packages/toast/src/ToastContainer/ToastContainer.styles.ts
```diff
@@ -0,0 +1,243 @@
...
+
+import { css } from '@leafygreen-ui/emotion';
+import { Theme } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+import { spacing, transitionDuration } from '@leafygreen-ui/tokens';
+
+import { TOAST_CONSTANTS } from '../constants';
+import { toastBGColor } from '../InternalToast';
+
+// Keeping this to ease future debugging
```
spark33, line 12: I like the idea of keeping it for future debugging - it would be nice to have some documentation on how to make use of it for future maintainers

packages/toast/src/ToastContainer/ToastContainer.tsx
```diff
@@ -0,0 +1,342 @@
...
+  // Track hover state
+  const [isHovered, setHoveredState] = useState(false);
+  const setHovered = () => setHoveredState(true);
+  const setUnhovered = () => setHoveredState(false);
+  // Track intended expand state
+  const [shouldExpand, setShouldExpand, getShouldExpand] = useStateRef(false);
+  const expandToasts = () => setShouldExpand(true);
+  const collapseToasts = () => setShouldExpand(false);
+  // Track stack size
+  const stackSize = stack.size;
```
spark33, line 71: nit - doesn't really feel necessary to have another const here

packages/toast/src/InternalToast/InternalToast.tsx
TheSonOfThomp: Since the toast isn't rendered with JSX, and `ToastProps` isn't extensible, it's hard to pass this in as a prop

packages/toast/src/ToastContainer/NotificationBar/NotificationBar.styles.ts
TheSonOfThomp: Removed the magic number entirely based it on a constant

packages/toast/src/ToastContainer/ToastContainer.tsx
TheSonOfThomp: yeah that's fair 👍

PR #1652, opened 2023-03-15
LG-2956: InlineCode component supports Polymorphic

brookescarlett:
## ✍️ Proposed changes

InlineCode component supports Polymorphic

🎟 _Jira ticket:_ [LG-2956](https://jira.mongodb.org/browse/LG-2956)

### For new components


packages/typography/src/InlineCode/InlineCode.tsx
```diff
@@ -20,26 +24,28 @@ import {
 } from './InlineCode.styles';
 import { InlineCodeProps } from './InlineCode.types';
 
-const InlineCode = React.forwardRef<HTMLElement, InlineCodeProps>(
+const InlineCode = InferredPolymorphic<InlineCodeProps>(
```
TheSonOfThomp, line 27: Does `InferredPolymorphic` need a default TS param?

TheSonOfThomp, line 27: Does InferredPolymorphic need a default TS param?

packages/typography/src/InlineCode/InlineCode.tsx
```diff
@@ -60,14 +66,14 @@ const InlineCode = React.forwardRef<HTMLElement, InlineCodeProps>(
       </code>
     );
 
-    if (isAnchor) {
+    if (needsWrapper) {
```
TheSonOfThomp, line 69: Don't love the conditional rendering here. I think we're trying to move away from `renderedXYZ` type variables

TheSonOfThomp: I would recommend something like:
```tsx 
  const { Component } = useInferredPolymorphic(as, rest, Fragment);
  const isFragment = isUndefined(as) // or isEqual(Component, Fragment)
  ...
  <Component {...(isFragment && {
      className: cx(...),
      ...rest
    }
  )}>
    <code>
      ...
    </code>
  </Component>

```

packages/typography/src/InlineCode/InlineCode.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+        const utils = render(<InlineCode href="string">{children}</InlineCode>);
+        const component = utils.container.firstChild;
+        expect((component as HTMLElement).tagName.toLowerCase()).toBe('a');
+      });
+
+      test('it renders as component passed to "as" prop when set', () => {
+        const utils = render(<InlineCode as="div">{children}</InlineCode>);
+        const component = utils.container.firstChild;
+        expect((component as HTMLElement).tagName.toLowerCase()).toBe('div');
+      });
```
TheSonOfThomp, line 27: Should also verify that there's still a `<code>` element in the DOM

packages/typography/src/InlineCode/InlineCode.tsx
brookescarlett: 👍

packages/typography/src/InlineCode/InlineCode.tsx
brookescarlett: Fragments don't accept props though

TheSonOfThomp: Oops I had the check backwards. You'd need to check `!isFragment && ...`. It's a bit of an ugly solution tbf

TheSonOfThomp (approved):

PR #1653, opened 2023-03-17
adam/table-v10-rc

TheSonOfThomp:


packages/table/src/Row/RowCellChildren.tsx
```diff
@@ -1,29 +1,40 @@
...
           <>
             {
               React.createElement(ExpandingCell, {
                 ...(FirstCellChild as ReactElement)?.props,
                 cellIndex: 0,
                 depth: row.depth,
-                isExpanding: isExpandedRow(row.id),
-                toggleIsExpanding: () => toggleExpandedRow(row.id),
+                isExpanding: isExpanded,
+                toggleIsExpanding: () => row.toggleExpanded(),
```
spark33, line 37: can this just be `toggleIsExpanding: row.toggleExpanded` now?

spark33 (approved):

PR #1656, opened 2023-03-20
Table TS

TheSonOfThomp:


packages/table/src/useLeafygreenTable/useLeafygreenTable.types.ts
```diff
@@ -2,57 +2,48 @@ import { RefObject } from 'react';
...
-  scrollToOffset: (index: number, options?: ScrollToOffsetOptions) => void;
-  scrollToIndex: (index: number, options?: ScrollToIndexOptions) => void;
-  measure: () => void;
-}
-
-// Above is copied from react-virtual as their types are not exported
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export type LGRowData = {
+  [key: string]: any;
+};
```
spark33, line 10: Should this extend from tanstack's `RowData` interface?

packages/table/src/useLeafygreenTable/useLeafygreenTable.types.ts
```diff
@@ -2,57 +2,48 @@ import { RefObject } from 'react';
...
-  Cell<LeafygreenTableType<T>, unknown>;
+export type LeafygreenTableCell<T extends LGRowData> = Cell<
+  LGTableDataType<T>,
+  LGRowData
+>;
 
-export type LeafygreenTableRow<T extends unknown> = unknown &
-  Row<LeafygreenTableType<T>>;
+export interface LeafygreenTableRow<T extends LGRowData>
+  extends Row<LGTableDataType<T>> {}
```
spark33, line 22: Everything feels great in this PR, except I may go back and replace all usages of the `Leafygreen` prefix like in `LeafygreenTableRow` with `LG` following your pattern

packages/table/src/useLeafygreenTable/useLeafygreenTable.types.ts
TheSonOfThomp: Technically yes, but I was getting errors. I'll revert this later

PR #1658, opened 2023-03-20
LG-2509: Number input

shaneeza:
## ✍️ Proposed changes

[Demo](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-numberinput--basic)
🎟 _Jira ticket:_ [LG-2509](https://jira.mongodb.org/browse/LG-2509)

Screenshot 2023-03-20 at 1 26 35 PM
Screenshot 2023-03-20 at 1 26 41 PM
Screenshot 2023-03-20 at 1 26 48 PM

### For new components


packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,184 @@
...
+  state = State.None,
+  errorMessage,
+  ...rest
+}: InputProps) {
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const { theme } = useDarkMode();
+
+  const renderErrorIcon = state === State.Error;
+
+  const { value, handleChange: handleChangeProp } = useControlledValue(
```
brookescarlett, line 57: this naming is confusing... the value here is coming from a hook not necessarily from a prop, right? can we rename to controlledValue and controlledHandleChange?

brookescarlett: Alternatively, we can just not wrap the function in our own handleChange since it's not doing anything and we should just be able to call this `handleChange` directly

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,184 @@
...
+    valueProp,
+    onChangeProp,
+  );
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
brookescarlett, line 62: We don't need this fn

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,184 @@
...
+        className={cx(
+          inputBaseStyles,
+          inputThemeStyles[theme],
+          sizeInputStyles[size],
+          {
+            [errorInputStyles[size]]: renderErrorIcon,
+            [inputAnimateStyles]: !disabled,
+          },
+        )}
+        type="number"
```
brookescarlett, line 135: should we add a console warning to our text-input if a user tries to use `type="number"` there?

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+      'aria-labelledby': ariaLabelledbyProp,
+      'aria-label': ariaLabelProp,
+      className,
+      disabled,
+      label,
+      value,
+      description,
+      errorMessage,
+      onChange,
+      onSelectChange = () => {},
```
brookescarlett, line 45: nit: define defaults together

packages/number-input/src/Select/Select.tsx
```diff
@@ -0,0 +1,149 @@
...
+            open={open}
+            {...popoverProps}
+          >
+            {unit?.displayName}
+          </Tooltip>
+          <Button
+            {...props}
+            className={cx(menuBaseStyles, menuThemeStyles[theme], className)}
+            ref={buttonRef}
+            onMouseEnter={(e: React.MouseEvent<HTMLButtonElement>) => {
```
brookescarlett, line 102: should we move this out of defining inline

brookescarlett:
Currently changing the size prop doesn't update the font-size for label, description, errorMessage - i'd assume we want that

packages/number-input/src/Input/Input.tsx
shaneeza: you're right, i don't need it

packages/number-input/src/Input/Input.tsx
shaneeza: yup!

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,184 @@
...
+        value={value}
+        onChange={handleChange}
+        aria-disabled={disabled}
```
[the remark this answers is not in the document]
shaneeza: From our team review last week, it seems we need to do a little more research on if disabled inputs should be focusable or not. I think we should wait until we know for sure before we can this to the styleguide.

shaneeza:
> Overall great work and really excited to see this component launch!
> 
> A couple of notes:
> 
> 1. Currently changing the size prop doesn't update the font-size for label, description, errorMessage - i'd assume we want that

In figma they are all 13px but I think that might be incorrect, I'll double check with Sooa.

> 2. toggling unitOptions in the "basic" story results in an error uniteOptions.find is not a function. We should probably be doing a check like `unitOptions?.find` or something similar to ensure we dont reach this case in prod

Storybook has the incorrect control type but i also should add an array check.

> 3. Might be the Icon itself (worth looking into) but the arrows do not seem to be optically aligned
>    screenshot_2023-03-20_at_2 48 23_pm_720

Def the icon, i'll see what I can do

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,191 @@
...
+      popoverZIndex,
+      usePortal,
+      portalClassName,
+      portalContainer,
+      scrollContainer,
+    } as const;
+
+    return (
+      <LeafyGreenProvider darkMode={darkMode}>
+        <div ref={forwardedRef} className={className}>
```
brookescarlett, line 80: do we usually forward the ref to the wrapper or to the underlying input component?

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,191 @@
...
+                })}
+              >
+                {unit}
+              </Overline>
+            )}
+            {renderSelectOnly && (
+              <Select
+                id={selectId}
+                disabled={disabled}
+                unit={unit as string}
```
brookescarlett, line 131: nit: technically could be undefined (although it shouldn't be), is it safer to default with `""`

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,191 @@
...
+                unitOptions={unitOptions}
+                onChange={onSelectChange}
+                size={size}
+                {...popoverProps}
+              />
+            )}
+          </div>
+          <div
+            className={errorMessageStyles}
```
brookescarlett, line 140: are we consistent with these aria-roles? can we also add to style guide please

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,184 @@
...
+        )}
+      >
+        <WarningIcon aria-hidden="true" />
+      </div>
+
+      <div
+        className={cx(arrowsBaseStyles, { [arrowsAnimateStyles]: !disabled })}
+      >
+        <button
+          aria-label="Increment number"
```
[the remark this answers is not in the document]
shaneeza: If we add `title` a tooltip would show up on hover of the arrows. Not sure if we want that behavior. At least with `aria-label` the string will be read by screenreaders.

shaneeza:
> Currently changing the size prop doesn't update the font-size for label, description, errorMessage - i'd assume we want that

I confirmed with Sooa that this is correct -- they should all be 13px. The `text-input` behaves the same way for sizes xs-default. @brookescarlett

packages/number-input/src/NumberInput/NumberInput.tsx
shaneeza: I started adding these roles in the password input. PIn other input components, we were only rendering the error message container if there was an error which is not ideal for screen readers since that message would not be announced when shown. What I'm doing here is always rendering the error message container and adding `aria-live` and `aria-relevant` as attributes. This way whenever the text changes in this container screen readers will announce it. This is something we should incorporate in our form control component.

shaneeza:
> Hover tooltip cuts off. Should have 4px apart.

I'm not sure I can replicate this. Does this still happen after a hard refresh? @sooachung

shaneeza:
> It is recommended to use the ellipsis to cut off units that are more than 30 characters. Three(3) seems short this is unnecessary here?. Is this something that designers should define?

This is because in the story I set a width on the wrapper container as well as a width on the input. It's up to the consumer to set the width for both because without a width the ellipses would not show up. @sooachung

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { InputProps } from './Input.types';
+
+/**
+ * @internal
+ */
+
+export const Input = React.forwardRef<HTMLInputElement, InputProps>(
+  (
+    {
+      value: valueProp,
```
brookescarlett, line 47: nit but preference for grouping aliases/defaults

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,201 @@
...
+          <WarningIcon aria-hidden="true" />
+        </div>
+
+        <div
+          className={cx(arrowsBaseStyles, {
+            [arrowsAnimateStyles]: !disabled,
+            [arrowDisabledStyles]: disabled,
+          })}
+        >
+          <button
```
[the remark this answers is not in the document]
brookescarlett: Sorry for the bad pseudo code, but you get what i mean

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+                </Description>
+              )}
+            </div>
+          )}
+          <div
+            className={cx(wrapperBaseStyles, wrapperSizeStyles[size], {
+              [wrapperGapStyles]: renderUnitOnly,
+            })}
+          >
+            <Input
```
brookescarlett, line 101: does this component have the potential to live outside of NumberInput and be leveraged as a base elsewhere? Not blocking, but wondering if this was designed with that in mind?

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+      portalContainer,
+      scrollContainer,
+      ...rest
+    }: NumberInputProps,
+    forwardedRef,
+  ) => {
+    const prefix = 'lg-numberinput';
+    const inputId = useIdAllocator({ prefix, id: idProp });
+    const errorMessageId = useIdAllocator({ prefix, id: ariaDescribedbyProp });
+    const descriptionId = useIdAllocator({ prefix });
```
brookescarlett, line 59: cannot wait til we have a helper to handle this for us :)

packages/number-input/src/Select/Select.tsx
```diff
@@ -0,0 +1,169 @@
...
+    if (selectedUnit !== undefined) {
+      onChange(selectedUnit);
+    }
+  };
+
+  /**
+   * Custom unit button with a tooltip.
+   * Tooltip will show up if there is an ellipse.
+   */
+  const CustomMenuButton = React.forwardRef(
```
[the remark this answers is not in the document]
brookescarlett: maybe you could wrap this in a hook that exposes both the component itself and `hasEllipse` and then just use those two pieces of information inside of the Select component itself

packages/number-input/src/Select/Select.tsx
```diff
@@ -0,0 +1,169 @@
...
+        <div className={customMenuButtonWrapperStyles}>
+          <Tooltip
+            enabled={hasEllipse && !disabled}
+            justify="middle"
+            // Using refEl instead of a trigger element because triggerProps by default, such as onMouseEnter, are added to the trigger element inside the tooltip component. OnMouseEnter is triggered by hovering over the trigger or any of its children. In the case of this custom menu button we don't want the tooltip to open when children are hovered so we add our own open logic with onMouseEnter.
+            refEl={buttonRef}
+            open={open}
+            {...popoverProps}
+          >
+            {currentUnitOption?.displayName}
```
brookescarlett, line 114: shouldn't we be confident this is defined at this point? should we have a type guard in place above that asserts this?

packages/number-input/src/Select/Select.tsx
```diff
@@ -0,0 +1,169 @@
...
+            onBlur={handleOnBlur}
+          >
+            {children}
+          </Button>
+        </div>
+      );
+    },
+  );
+
+  CustomMenuButton.displayName = 'CustomMenuButton';
```
brookescarlett, line 139: Yeah I'm feeling pretty strongly that should should be defined outside of the Select component?

packages/number-input/src/Input/Input.tsx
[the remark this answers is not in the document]
spark33: Bump! Feels awkward to have buttons / icons in a component that's called `Input`

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+  scrollContainer:
+    typeof window !== 'undefined'
+      ? PropTypes.instanceOf(Element)
+      : PropTypes.any,
+  portalContainer:
+    typeof window !== 'undefined'
+      ? PropTypes.instanceOf(Element)
+      : PropTypes.any,
+  portalClassName: PropTypes.string,
+} as any;
```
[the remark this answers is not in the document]
spark33: TS tries to infer types from PropTypes without this, which seems to "compete" with the logic we write in TS? I don't understand it well enough to explain it in detail, but this was the workaround that Adam had found.

packages/select/src/ListMenu.tsx
```diff
@@ -99,9 +107,11 @@ const ListMenu = React.forwardRef<HTMLUListElement, ListMenuProps>(
         active={open && !disabled}
         spacing={6}
         align={Align.Bottom}
-        justify={Justify.Middle}
+        justify={dropdownAutoWidth ? Justify.Start : Justify.Middle}
         adjustOnMutation
-        className={className}
+        className={cx(popoverClassName, className, {
+          [autoWidthStyles]: dropdownAutoWidth && !usePortal,
```
spark33, line 113: I don't think we need the `!usePortal` check here since the width will be set to max-content by default in those cases

sooachung:
I don't see the issues for the above comments now so they can be resolved!

Found something new tho, the hover ring and button color look the same here. The fill should be gray light 3 and the hover ring should be gray light 2
Screen Shot 2023-03-23 at 1 50 51 PM

packages/number-input/src/Select/Select.tsx
shaneeza: fixed this

packages/number-input/src/Select/Select.tsx
```diff
@@ -0,0 +1,106 @@
...
+        ref={forwardedRef}
+        disabled={disabled}
+        displayName={currentUnitOption.displayName}
+        {...props}
+      >
+        {children}
+      </SelectButton>
+    ),
+  );
+  CustomMenuButton.displayName = 'CustomMenuButton';
```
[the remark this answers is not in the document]
brookescarlett: No that's fine! There's no issue with the displayName

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+      popoverZIndex,
+      usePortal,
+      portalClassName,
+      portalContainer,
+      scrollContainer,
+    } as const;
+
+    return (
+      <LeafyGreenProvider darkMode={darkMode}>
+        <div className={className}>
```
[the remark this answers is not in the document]
shaneeza: I think it should be here so consumers can add CSS, like the width, on the wrapper, the input, or the select button.

sooachung:
Dark Mode button on hover state needs to be changed to
Screen Shot 2023-03-27 at 12 16 20 PM

On hover
- ring color: gray dark 2 (it should be the same for input filed hover)
- fill: gray dark 2

Inactive:
- fill: gray dark 2

packages/number-input/src/Arrows/Arrow.tsx
```diff
@@ -0,0 +1,50 @@
...
+}: ArrowProps) => {
+  const { theme } = useDarkMode();
+
+  return (
+    <button
+      aria-label={`${direction} number`}
+      onClick={() => handleClick(direction)}
+      onKeyDown={handleArrowKeyDown}
+      className={cx(arrowBaseStyles, arrowThemeStyles[theme])}
+      type="button"
```
brookescarlett, line 34: is this not default?

shaneeza: No, we've had issues in the past where buttons without a default would submit a form.

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const [shouldErrorTransition, setShouldErrorTransition] =
+      useState<boolean>(false);
+    const inputRef = useForwardedRef<HTMLInputElement | null>(forwardRef, null);
+    const { theme } = useDarkMode();
+
+    useEffect(() => {
+      return () => {
+        clearTimeout(translateTimeout);
+      };
+    });
```
brookescarlett, line 62: does this need deps?

shaneeza: i'm not getting any errors

brookescarlett: no errors because this means it'll just run once but I don't think that's the behavior you want, right?

shaneeza: I think that the cleanup is performed when the component unmounts

shaneeza: https://legacy.reactjs.org/docs/hooks-effect.html#example-using-hooks-1

> When exactly does React clean up an effect? React performs the cleanup when the component unmounts.

brookescarlett: But do you want this only performed when the component unmounts? that's what I don't understand

shaneeza: Do you see something that I might be over looking?

packages/number-input/src/Arrows/Arrow.tsx
```diff
@@ -0,0 +1,50 @@
...
+  handleClick,
+  handleArrowKeyDown,
+  direction,
+}: ArrowProps) => {
+  const { theme } = useDarkMode();
+
+  return (
+    <button
+      aria-label={`${direction} number`}
+      onClick={() => handleClick(direction)}
```
spark33, line 31: I think it would be more appropriate to call the prop `onClick` and to have a `const handleClick = () => {...}` defined outside of the render function

packages/number-input/src/ErrorIcon/ErrorIcon.styles.ts
```diff
@@ -0,0 +1,76 @@
...
+  [Size.XSmall]: css`
+    translate: 4px 0;
+  `,
+  [Size.Small]: css`
+    translate: 4px 0;
+  `,
+  [Size.Default]: css`
+    translate: 4px 0;
+  `,
+};
```
spark33, line 19: Looks like this can be all a part of a `baseStyles` that's not dependent on `size`

packages/number-input/src/ErrorIcon/ErrorIcon.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { ErrorIconProps } from './ErrorIcon.types';
+
+/**
+ * @internal
+ */
+
+export const ErrorIcon = ({
+  disabled,
+  size,
+  renderErrorIcon,
```
spark33, line 25: `shouldRenderErrorIcon`?

packages/number-input/src/Input/Input.styles.ts
```diff
@@ -0,0 +1,258 @@
...
+  [Size.XSmall]: css`
+    padding: 0 0 0 ${spacing[1] * 3}px;
+  `,
+  [Size.Small]: css`
+    padding: 0 0 0 ${spacing[1] * 3}px;
+  `,
+  [Size.Default]: css`
+    padding: 0 0 0 ${spacing[1] * 3}px;
+  `,
+};
```
spark33, line 112: Should these values be different based on size?

spark33: Otherwise, they could be all in a `baseStyles`

packages/number-input/src/NumberInput.story.tsx
```diff
@@ -0,0 +1,177 @@
...
+    console.log('story: ref', inputRef.current?.value);
+    setValue(e.target.value);
+  };
+
+  return (
+    <NumberInput
+      {...rest}
+      ref={inputRef}
+      value={value}
+      unit={view === 'unitless' ? '' : view === 'unit' ? unitProp : unit}
```
spark33, line 117: nit: would be cleaner in a `getUnit` helper fn

spark33 (changes requested):

packages/number-input/src/Input/Input.tsx
brookescarlett: ah ok sorry you're right

brookescarlett:
Did you test this in .design?

shaneeza:
@brookescarlett planning on testing after i review seans PR

packages/number-input/src/NumberInput.story.tsx
```diff
@@ -0,0 +1,177 @@
...
+  },
+];
+
+type StoryProps = NumberInputProps & { view: string };
+
+const Template: Story<StoryProps> = args => {
+  const { view, unit: unitProp = 'one', ...rest } = args;
+  const [unit, setUnit] = useState<string>(unitOptions[0].displayName);
+  const [value, setValue] = useState<string>('');
+  const inputRef = useRef<HTMLInputElement | null>(null);
```
TheSonOfThomp, line 96: Noticed that when the input is controlled, exponents (e.g. `2e4`) are not allowed. This doesn't quite match this sandbox (which also behaves somewhat oddly)
https://codesandbox.io/s/quiet-flower-8dcqwd?file=/src/App.tsx

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -0,0 +1,390 @@
...
+  describe('a11y', () => {
+    test('does not have basic accessibility issues', async () => {
+      const { container } = renderNumberInput({ label, ...defaultProps });
+      const results = await axe(container);
+      expect(results).toHaveNoViolations();
+    });
+  });
+
+  describe('unitless number input with default props', () => {
+    test('renders type as "number', async () => {
```
TheSonOfThomp, line 70: "render input `type` as 'number'" - to avoid confusion with actual TS types

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -0,0 +1,390 @@
...
+      });
+
+      expect((numberInput as HTMLInputElement).value).toBe('');
+
+      fireEvent.change(numberInput, {
+        target: { value: '1' },
+      });
+
+      expect((numberInput as HTMLInputElement).value).toBe('1');
+      expect(defaultProps.onChange).toHaveBeenCalledTimes(1);
```
TheSonOfThomp, line 148: should we test this with different values?

packages/number-input/src/Select/Select.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+    },
+    {
+      displayName: 'year',
+      value: 'year',
+    },
+  ],
+};
+
+function renderSelect(props = {}) {
+  // @ts-ignore - missing props
```
TheSonOfThomp, line 26: should be able to type renderSelect props as `Partial<SelectProps>` to avoid ts-error

packages/number-input/src/Select/Select.tsx
```diff
@@ -0,0 +1,111 @@
...
+  selectDisabledStyles,
+  selectStyles,
+  wrapperBaseStyles,
+} from './Select.styles';
+import { SelectProps } from './Select.types';
+
+/**
+ * @internal
+ */
+export function Select({
```
TheSonOfThomp, line 21: Should we call this `UnitSelect` to avoid confusion?

packages/number-input/src/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Checks if the selected option has an ellipse.
+     */
+    const hasEllipse = textNode?.offsetWidth < textNode?.scrollWidth;
+
+    const handleMouseEnter = (e: React.MouseEvent<HTMLButtonElement>) => {
+      const popoverParent = (e.target as HTMLButtonElement).closest(
+        `.${popoverClassName}`,
+      );
+      if (!popoverParent) setOpen(true);
```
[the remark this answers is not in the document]
shaneeza: If there is no popover parent that means that we are not hovering over the popover menu. If that's the case then we should show the tooltip.

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,203 @@
...
+          break;
+        }
+      }
+    };
+
+    /**
+     * Edge case if the user clicks on an arrow button then switches to keyboard click.
+     * By default if focus is in the input then the keyboard clicks will work automatically but since the buttons are custom and outside of the input we are mimicking native behavior.
+     */
+    const handleArrowKeyDown = (e: KeyboardEvent) => {
```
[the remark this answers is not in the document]
shaneeza: correct, since the arrow is not part of the input I'm faking the bahavior

packages/number-input/src/NumberInput.story.tsx
shaneeza: since its just for storybook i think it should be ok

packages/number-input/src/Input/Input.tsx
```diff
@@ -0,0 +1,203 @@
...
+     *
+     * Here we set state to true which conditionally adds the transition to the CSS.
+     * We also clear the timeout that was created during `mouseLeave` or `onBlur` to prevent execution.
+     *
+     * The purpose of this is to only have the error icon animate when the arrows are visible.
+     */
+    const handleSetErrorTransition = () => {
+      setShouldErrorTransition(true);
+      clearTimeout(translateTimeout);
+    };
```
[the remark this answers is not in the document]
shaneeza: oh, I think I forgot to account for being focused but no longer hovering

packages/select/src/types.ts
```diff
@@ -73,6 +73,12 @@ interface BaseSelectProps
    *  determines the base font size if sizeVariant is set to default.
    */
   baseFontSize?: BaseFontSize;
+  /**
+   *  If true, the width of the dropdown will be as wide as the widest option. By default the width of the dropdown will be the same width as the menu button.
+   *
+   * @default false
+   */
+  dropdownAutoWidth?: boolean;
```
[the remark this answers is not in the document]
brookescarlett: agreed - otherwise i think the prop name is a little unclear

packages/number-input/src/Select/Select.tsx
shaneeza: i'll change this

packages/number-input/src/Select/Select.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const { getByRole } = renderSelect({ ...selectProps });
+    const trigger = getByRole('button', { name: selectProps.unit });
+    fireEvent.click(trigger);
+
+    // First option has a checkmark next to it
+    selectProps.unitOptions.slice(0, 1).forEach(lang => {
+      expect(
+        getByRole('option', { name: `Checkmark Icon ${lang.displayName}` }),
+      ).toBeInTheDocument();
+    });
```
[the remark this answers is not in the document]
shaneeza: Yeah, maybe I can remove it since NumberInput has the same test

packages/select/src/types.ts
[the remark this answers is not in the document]
shaneeza: makes sense, i'll update

packages/number-input/src/Input/Input.styles.ts
shaneeza: removing this

packages/number-input/src/ErrorIcon/ErrorIcon.styles.ts
shaneeza: removing

packages/number-input/src/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+      portalContainer,
+      scrollContainer,
+    } as const;
+
+    /**
+     * Gets the text node for the selected option.
+     */
+    const textNode = buttonRef.current?.querySelector(
+      `.${menuButtonTextClassName}`,
+    ) as HTMLElement;
```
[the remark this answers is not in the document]
shaneeza: I think if we use a ref for this, the node would be stale, and we would have to update the node on each render. If this is incorrect let me know!

this is what I ended up having to do to get it to work. 
```
    const textNodeRef = useRef<HTMLElement | undefined | null>(
      buttonRef.current?.querySelector(`.${menuButtonTextClassName}`),
    );

    useEffect(() => {
      textNodeRef.current = buttonRef.current?.querySelector(
        `.${menuButtonTextClassName}`,
      );
    });
```

packages/number-input/src/Select/Select.tsx
```diff
@@ -0,0 +1,111 @@
...
+        size={size}
+        data-testid={dataTestId}
+        {...popoverProps}
+        __INTERNAL__menuButtonSlot__={CustomMenuButton}
+      >
+        {unitOptions?.map(option => (
+          <Option key={option.displayName} value={option.displayName}>
+            {option.displayName}
+          </Option>
+        ))}
```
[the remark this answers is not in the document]
shaneeza: oh, don't need that check, if it's undefined then the select component won't even render.

packages/number-input/src/Input/Input.tsx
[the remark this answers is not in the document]
shaneeza: should be fixed

packages/number-input/src/NumberInput/NumberInput.spec.tsx
shaneeza: i'll add a couple more

packages/number-input/src/Arrows/Arrow.tsx
```diff
@@ -0,0 +1,50 @@
...
+      onClick={() => onClick(direction)}
+      onKeyDown={onKeyDown}
+      className={cx(arrowBaseStyles, arrowThemeStyles[theme])}
+      type="button"
+      tabIndex={-1} // Mimicking native behavior; you cannot focus on an arrow.
+      disabled={disabled}
+    >
+      <Icon
+        className={
+          direction === Direction.Decrement ? downArrowRotateStyles : ''
```
spark33, line 40: nit - either use `cx`, or replace the `''` with `undefined

packages/number-input/src/SelectButton/SelectButton.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Oh I see, didn't realize `menuButtonTextClassName` was the element inside `Select`

packages/number-input/src/Arrows/Arrow.tsx
TheSonOfThomp: It feels verbose but `cx` would be the "correct" way to do this for our library
```
cx({
    [downArrowRotateStyles]: direction === Direction.Decrement
})
```

shaneeza: done

spark33 (approved):
🚀🚀🚀🚀🚀

packages/number-input/src/Select/Select.spec.tsx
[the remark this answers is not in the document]
shaneeza: Do you think it's redundant?

packages/number-input/src/Input/Input.styles.ts
```diff
@@ -0,0 +1,253 @@
...
+import { Size, State } from '../NumberInput/NumberInput.types';
+
+/**
+ * Adds an inset box shadow to hide the UA background styles for autofilled inputs
+ */
+const autofillShadowOverride = (color: string) => `0 0 0 100px ${color} inset`;
+
+export const wrapperClassName = createUniqueClassName('number-input-wrapper');
+
+// TODO: why is xs text not center aligned
```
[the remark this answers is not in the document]
shaneeza: done!

packages/number-input/src/Select/Select.spec.tsx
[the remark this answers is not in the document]
TheSonOfThomp: If it's passing, we can leave it, but it is a bit redundant. Maybe leave a note? I'm just worried that if we make changes that break `Select` that might break these tests (false positive)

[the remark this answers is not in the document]
shaneeza: sounds good

packages/number-input/src/Input/Input.styles.ts
```diff
@@ -30,6 +30,7 @@ export const inputBaseStyles = css`
   box-sizing: border-box;
   transition: ${transitionDuration.default}ms ease-in-out;
   transition-property: background-color;
+  padding: 0 0 0 ${spacing[1] * 3}px;
```
TheSonOfThomp, line 33: Why `x3`? I feel like for numbers like this we should just use the actual number

TheSonOfThomp: If we want to keep using `spacing` here I'd prefer adding instead of multiplying `spacing[2] + spacing[1]` (4 + 2)

packages/number-input/src/Select/Select.spec.tsx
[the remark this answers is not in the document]
shaneeza: done

packages/number-input/src/Input/Input.tsx
```diff
@@ -47,21 +54,25 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
+    const isFocusedRef = useRef<boolean>(false);
     const inputRef = useForwardedRef<HTMLInputElement | null>(forwardRef, null);
     const { theme } = useDarkMode();
+    const isControlled = !isUndefined(valueProp);
 
     useEffect(() => {
+      // On unmount, removes the timeout that is set on `onMouseLeave` and `'onBlur` of the wrapper container.
       return () => {
-        clearTimeout(translateTimeout);
+        clearTimeout(translateTimeout.current as unknown as number);
```
TheSonOfThomp, line 71: 😷 Why do we need `as unknown as number`?

This should work a bit better:
```ts
if (translateTimeout.current) {
  clearTimeout(translateTimeout.current)
}
```

packages/number-input/src/NumberInput/NumberInput.types.ts
```diff
@@ -74,7 +74,9 @@ export interface WithoutSelectProps {
   onSelectChange?: never;
 }
 
-export type SelectProps = WithSelectProps | WithoutSelectProps;
+export type ConditionalUnitSelectProps =
```
TheSonOfThomp, line 77: 👍

TheSonOfThomp (approved):
LGTM, though I really don't love the `as unknown as number`

packages/number-input/src/Input/Input.tsx
shaneeza: yeah this is better, i'll update

packages/number-input/src/Input/Input.styles.ts
shaneeza: i'll use numbers, i think even with adding we'll run into similar issues

PR #1668, opened 2023-03-24
LG-2706: Table v11

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2706](https://jira.mongodb.org/browse/LG-2706)

Demo deployment: https://spark33.github.io/leafygreen-ui/?path=/story/components-table--basic

image

### For new components


.changeset/gold-clouds-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/table': major
+---
+
+Virtual scrolling, selectable rows, and other major features. Refer to upgrade guide for details
```
brookescarlett, line 5: This should probably be a little bit more descriptive. 

"Rewrite of Table component, which includes support for new features: XYZ. See upgrade guide for more details"

packages/table/package.json
```diff
@@ -20,16 +20,23 @@
     "access": "public"
   },
   "dependencies": {
+    "@faker-js/faker": "^7.6.0",
```
brookescarlett, line 23: should this be a dev dep?

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const baseCellStyles = css`
+  padding: 0 8px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+
+  &:focus-visible {
+    box-shadow: inset;
+  }
+  &:last-child {
```
brookescarlett, line 22: nit: line break above

packages/table/src/Cell/HeaderCell/HeaderCell.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const { table } = useTableContext();
+
+  const isFirstCell = cellIndex === 0;
+  const isSelectable = !!table && !!table.hasSelectableRows;
+
+  let columnName, sortState, onSortIconClick;
+
+  if (header && header.column.getCanSort()) {
+    columnName = header.column.columnDef.header as string;
+    const headerSortDirection = header.column.getIsSorted().toString();
```
brookescarlett, line 46: clicking the sort icon three times in a row takes it back to unsorted -- i wouldn't expect this. Clicking should oscillate between sorting a->z and z->a

packages/table/src/ExpandedContent/ExpandedContent.tsx
```diff
@@ -0,0 +1,61 @@
...
+}: ExpandedContentProps<T>) => {
+  const contentRef = useRef<HTMLDivElement>(null);
+  const isExpanded = row.getIsExpanded();
+  const content =
+    row?.original?.renderExpandedContent &&
+    row?.original?.renderExpandedContent(row);
+
+  const { theme } = useDarkMode();
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
brookescarlett, line 30: why are we disabling this check?

packages/table/src/ExpandedContent/ExpandedContent.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const ExpandedContent = <T extends RowData>({
+  row,
+  ...rest
+}: ExpandedContentProps<T>) => {
+  const contentRef = useRef<HTMLDivElement>(null);
+  const isExpanded = row.getIsExpanded();
+  const content =
+    row?.original?.renderExpandedContent &&
+    row?.original?.renderExpandedContent(row);
```
brookescarlett, line 26: `row?.original?.renderExpandedContent?.(row)`

packages/table/src/ExpandedContent/ExpandedContent.tsx
```diff
@@ -0,0 +1,61 @@
...
+  expandedContentTransitionStyles,
+} from './ExpandedContent.styles';
+import { ExpandedContentProps } from './ExpandedContent.types';
+
+const ExpandedContent = <T extends RowData>({
+  row,
+  ...rest
+}: ExpandedContentProps<T>) => {
+  const contentRef = useRef<HTMLDivElement>(null);
+  const isExpanded = row.getIsExpanded();
```
brookescarlett, line 23: if we're sure row exists why do we need optional chaining on the next line?

packages/table/src/ExpandedContent/ExpandedContent.tsx
```diff
@@ -0,0 +1,61 @@
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const contentHeight = useMemo(
+    () => (contentRef.current ? contentRef.current.clientHeight : 0),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [content],
+  );
+
+  return (
+    <InternalRowBase {...rest}>
+      <td colSpan={row?.getVisibleCells().length} className={cx(baseStyles)}>
```
brookescarlett, line 39: ditto here

packages/table/src/HeaderRow/HeaderRow.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { HeaderRowProps } from './HeaderRow.types';
+
+const HeaderRow = ({
+  children,
+  ...rest
+}: PropsWithChildren<HeaderRowProps>) => {
+  return (
+    <tr {...rest}>
+      {React.Children.map(children, (child: ReactNode, index: number) => {
```
brookescarlett, line 13: in the storybook it looks like the consumer is responsible for rendering HeaderCell components inside of HeaderRow components is this not the case?

packages/table/src/Row/InternalRowBase.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { cx } from '@leafygreen-ui/emotion';
+import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+
+import { clickableStyles, disabledStyles, rowBaseStyles } from './Row.styles';
+import { InternalRowBaseProps } from './Row.types';
+
+/**
+ * The base Row component, extended by `InternalRow(With/Without)RT`
+ */
+const InternalRowBase = ({
```
brookescarlett, line 12: should we handle darkmode here, since both with and without RT rely on it?

packages/table/src/Row/InternalRowWithoutRT.tsx
```diff
@@ -0,0 +1,40 @@
...
+      className={cx(
+        {
+          [zebraStyles[theme]]: shouldAlternateRowColor,
+        },
+        className,
+      )}
+      {...rest}
+    >
+      {React.Children.map(children, (child: ReactNode, index: number) => {
+        return <Cell {...(child as ReactElement).props} cellIndex={index} />;
```
brookescarlett, line 34: This pattern is confusing to me: Aren't rows already rendering cells?

packages/table/src/TableBody/TableBody.tsx
```diff
@@ -0,0 +1,49 @@
...
+    paddingTop = virtualRows.length > 0 ? virtualRows?.[0]?.start || 0 : 0;
+    paddingBottom =
+      virtualRows.length > 0
+        ? totalSize - (virtualRows?.[virtualRows.length - 1]?.end || 0)
+        : 0;
+  }
+
+  return (
+    <Polymorph as={bodyAs} {...rest}>
+      {paddingTop > 0 && (
```
brookescarlett, line 32: this could use some comments

packages/table/src/TableContainer/TableContainer.tsx
```diff
@@ -0,0 +1,37 @@
...
+    { children, className, ...rest }: PropsWithChildren<TableContainerProps>,
+    forwardedRef,
+  ) => {
+    return (
+      <div
+        ref={forwardedRef}
+        className={cx(
+          css`
+            overflow: auto;
+            height: 500px;
```
brookescarlett, line 18: where does 500px come from?

packages/table/src/Table.story.tsx
```diff
@@ -1,363 +1,838 @@
...
-        ) : (
-          <Cell>{datum.age}</Cell>
-        )}
-      </Row>
-    )}
-  </Table>
-);
-NestedRowsWithLongContent.args = {
-  withHeaders: true,
+export const KitchenSink: ComponentStory<typeof Table> = args => {
```
brookescarlett, line 695: What happens with content that overflows horizontally as this table demonstrates?

packages/table/src/Table.story.tsx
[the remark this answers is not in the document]
brookescarlett: Might be missing it but also a sticky header example

packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -0,0 +1,106 @@
...
+        disabled={disabled}
+        aria-hidden={!isRowVisible}
+        aria-expanded={isExpanded}
+        {...rest}
+      >
+        <RowCellChildren row={row} disabled={disabled}>
+          {CellChildren}
+        </RowCellChildren>
+      </InternalRowBase>
+      {OtherChildren}
```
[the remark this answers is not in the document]
spark33: Subrows and expandable content. I could check for `isComponentType(child, 'ExpandedContent') || isComponentType(child, 'Row')` instead of `!isComponentType(child, 'Cell')`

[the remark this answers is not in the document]
spark33: But I figured I'd allow any other element in case people want `styled('tr')` or any custom component that still renders a `tr`

alvenw:
- [ ] do we handle truncation behavior? I would expect to either see ellipses or force a line break in this situation 
image
- [ ] Nested rows should no longer have dividing line when expanded 
image
- [ ] I know that expandable rows are unstructured but it would be nice to have it in the example at least aligned with the bounds of the left-most column instead of flush w the edge of the table
image
- [ ] child expandable row content is show even when parent nested rows are collapsed 
image
- [ ] disabled nested rows have white space/line in between each row 
image
- [ ] would be nice to have an option to see the pagination control below the table if it's not too much work
image

packages/table/src/Cell/HeaderCell/HeaderCell.tsx
spark33: This was a design decision that Alven and I discussed, but if it just went a->z to z->a, then there'd be no way to unsort.

packages/table/src/Row/InternalRowWithoutRT.tsx
spark33: This was put in place because the behavior where `Cell`s inherited the `HeaderCell`s' `align` prop depended on a `cellIndex`, but there might actually be a workaround since we changed how that prop is inherited recently. I'll look into this

packages/table/src/Row/Row.stories.tsx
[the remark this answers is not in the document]
spark33: Only one story will be rendered on .design - the KitchenSink story that's marked as the default. I'll have to link the two repos locally to test

packages/table/src/Cell/HeaderCell/HeaderCell.tsx
alvenw: Yeah just to add onto Sean's comment, the inability to unsort was previously brought up as an issue and this was the simplest solution we landed on. Other design systems use this pattern as well, here's [Carbon as an example](https://react.carbondesignsystem.com/?path=/story/components-datatable-sorting--default)

packages/table/src/Row/Row.styles.ts
```diff
@@ -0,0 +1,123 @@
...
+import { palette } from '@leafygreen-ui/palette';
+import {
+  focusRing,
+  hoverRing,
+  transitionDuration,
+} from '@leafygreen-ui/tokens';
+
+export const rowBaseStyles = css`
+  transition: ${transitionDuration.default}ms ease-in-out;
+  transition-property: all;
```
shaneeza, line 12: Maybe set this to only the properties that you need. When you toggle between light and dark mode the background colors linger for a while.

packages/table/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  rules: {
+    'react/forbid-prop-types': 'off',
```
shaneeza, line 3: What does this do?

packages/table/src/ToggleExpandedIcon/ToggleExpandedIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+}: ToggleExpandedIconProps) => {
+  const { theme } = useDarkMode();
+
+  return (
+    <IconButton
+      aria-label={`${isExpanded ? 'Collapse' : 'Expand'} row`}
+      disabled={disabled}
+      onClick={toggleExpanded}
+      className={cx(iconButtonStyles, {
+        [css`
```
shaneeza, line 26: Can you move this to styles?

packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -0,0 +1,106 @@
...
+            [grayZebraRowStyles[theme]]: isOddVSRow && shouldAlternateRowColor,
+            [zebraStyles[theme]]: !virtualRow && shouldAlternateRowColor,
+            [selectedRowStyles[theme]]: isSelected,
+          },
+          className,
+        )}
+        data-selected={isSelected}
+        disabled={disabled}
+        aria-hidden={!isRowVisible}
+        aria-expanded={isExpanded}
```
shaneeza, line 94: Is it possible to add `aria-control` to specify what item is being opened?

packages/table/src/Row/InternalRowWithRT.tsx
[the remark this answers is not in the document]
brookescarlett: Are you enforcing that it any way though? (That the underlying component is a tr?)

packages/table/src/Row/ExpandableContent.spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+              </Row>
+            );
+          })}
+        </TableBody>
+      </Table>
+    </div>
+  );
+};
+
+describe('packages/table/Row/ExpandableContent', () => {
```
brookescarlett, line 63: Don't love how these tests are broken out -- ExpandableContent and NestedRows are not components, so this is confusing. Let's test these functionalities in a test suite that is named after the component that it is directly testing.

packages/table/src/Row/ExpandableContent.spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+  test('renders the correct number of cell children', () => {
+    const { getAllByRole: getAllByRoleLocal } = render(
+      <RowWithExpandableContent />,
+    );
+    const firstRow = getAllByRoleLocal('row')[1];
```
brookescarlett, line 68: first row isn't row[0]?

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+import processColumns from './processColumns';
+import processData from './processData';
+import { V11AdapterProps } from './V11Adapter.types';
+
+// assumes table is first element in children
+// reads columns from columns' keys
+// supports up to one layer of nested rows
+// assumes that the key of a column in the original data === column label header in lowercase; can be overridden by `headerLabelMapping`
+const V11Adapter = <T extends LGRowData>({
```
brookescarlett, line 30: should we be testing this component?

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const {
+    data,
+    columns,
+    children: childrenFn,
+  } = (OldTable as ReactElement).props;
+  const processedColumns: Array<LGColumnDef<T>> = useMemo(
+    () => processColumns(data, columns, headerLabelMapping),
+    [data, columns, headerLabelMapping],
+  );
+  const processedData: Array<LGTableDataType<T>> = useState(() =>
```
brookescarlett, line 49: doesn't useState return a tuple?

packages/table/src/V11Adapter/V11Adapter.types.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { LeafyGreenTableOptions, LGRowData } from '../useLeafyGreenTable';
+import { TableProps } from '..';
+
+export type V11AdapterProps<T extends LGRowData> = PropsWithChildren<
+  Pick<LeafyGreenTableOptions<T>, 'useVirtualScrolling' | 'hasSelectableRows'> &
+    Pick<TableProps<T>, 'shouldAlternateRowColor'> &
+    DarkModeProps & {
+      /**
+       * Mapping of TableHeader label to the key of the field in the Table's data
```
brookescarlett, line 13: headerLabelAccessor?

packages/table/src/utils/getParentRowId.ts
```diff
@@ -0,0 +1,9 @@
+function getParentRowId(childId?: string) {
+  if (childId) {
+    const childIds = childId.split('.'); // ['0']
```
brookescarlett, line 3: nit rm the comment?

packages/table/src/utils/getParentRowId.ts
```diff
@@ -0,0 +1,9 @@
...
+    const parentId = childIds.slice(0, childIds.length - 1).join('.'); // []
```
brookescarlett, line 4: nit ditto here

packages/table/src/index.ts
```diff
@@ -1,6 +1,6 @@
...
-export { default as Row } from './Row';
-export { default as Table } from './Table';
-export { DataType } from './TableContext';
-export { default as TableHeader } from './TableHeader';
+import Table, { TableProps } from './Table';
+
+export * from '@tanstack/react-table';
+export { Table, type TableProps };
+
+export default Table;
```
shaneeza, line 6: I may have missed it, but are you exporting all the necessary components?

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+    }: {
+      table: Table<LGTableDataType<LGRowData>>;
+      row: Row<LGTableDataType<LGRowData>>;
+    }) => (
+      <Checkbox
+        checked={row.getIsSelected()}
+        indeterminate={row.getIsSomeSelected()}
+        onChange={row.getToggleSelectedHandler()}
+        aria-label={`Select row ${row.index}`}
+        // Don't animate if _all_ rows have been checked (usually, if header row is clicked). Not the _best_ check, but it mostly works
```
brookescarlett, line 44: What's this comment?

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+import {
+  LeafyGreenTable,
+  LGColumnDef,
+  LGTableDataType,
+  VirtualizerValues,
+} from '.';
+
+const selectColumnConfig: LGColumnDef<LGRowData> = {
+  id: 'select',
+  size: 14, // match checkbox width
```
brookescarlett, line 19: `const checkboxWidth = 14`

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+`;
+
+const flexAlignment: Record<string, string> = {
+  left: 'start',
+  right: 'end',
+  center: 'center',
+};
+
+export const alignmentStyles = (align: Align = 'left') => css`
```
brookescarlett, line 33: `Align = Align.Left?`

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+  isSelectable,
+}: {
+  depth?: number;
+  isExpandable?: boolean;
+  isSelectable?: boolean;
+}) => {
+  if (depth === 0) {
+    if (isSelectable) {
+      return css`
+        padding-left: 8px;
```
brookescarlett, line 49: spacing[2]

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+        padding-right: 8px;
```
brookescarlett, line 50: ditto here

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+      `;
+    } else {
+      return css`
+        padding-left: ${baseTableSidePadding + (isExpandable ? 0 : 8)}px;
```
brookescarlett, line 54: spacing[2]

packages/table/src/V11Adapter/V11Adapter.tsx
TheSonOfThomp: ^ definitely

packages/table/src/V11Adapter/processColumns.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { ReactElement } from 'react';
+import camelCase from 'lodash/camelCase';
+
+import { TableProps } from '../TableV10/Table';
+import { LGColumnDef, LGRowData } from '../useLeafyGreenTable';
+
+const processColumns = <T extends LGRowData>(
```
TheSonOfThomp, line 7: Can we add TSDoc comments

packages/table/src/V11Adapter/processColumns.tsx
```diff
@@ -0,0 +1,52 @@
...
+  data: Array<T>,
+  columns: TableProps<any>['columns'],
+  headerLabelMapping?: { [key: string]: string },
+) => {
+  const HeaderRow = React.Children.toArray(columns)[0] as ReactElement;
+  const TableHeaders = React.Children.toArray(HeaderRow.props.children);
+  const processedColumns: Array<LGColumnDef<T>> = [];
+  TableHeaders.forEach(TableHeader => {
```
TheSonOfThomp, line 15: Could use `.map` (or `.reduce` if the lengths don't match)

packages/table/src/ExpandedContent/ExpandedContent.tsx
spark33: Without it, you still get the error `Cannot invoke an object which is possibly 'undefined'.ts(2722)`

packages/table/src/V11Adapter/processColumns.tsx
```diff
@@ -0,0 +1,52 @@
...
+      !!headerProps.handleSort ||
+      !!headerProps.compareFn;
+    processedColumns.push({
+      accessorKey:
+        (headerLabelMapping && headerLabelMapping[headerProps.label]) ??
+        camelCase(headerProps.label),
+      header: headerProps.label,
+      align: headerProps.dataType === 'number' ? 'right' : 'left',
+      enableSorting: hasSorting,
+      sortingFn: headerProps.compareFn
```
TheSonOfThomp, line 28: I feel like these sort functions should be defined externally for clarity. This section is a bit tough to read

packages/table/src/V11Adapter/processData.tsx
```diff
@@ -0,0 +1,75 @@
...
+        };
+      },
+      {},
+    );
+
+    const subRowChildren = rowChildren.filter(child =>
+      isComponentType(child, 'Row'),
+    );
+    if (subRowChildren.length > 0) newDatum.subRows = [];
+    subRowChildren.forEach(subRow => {
```
TheSonOfThomp, line 37: I feel like this should be 
```ts
const subRows = subRowChildren.map(...)
newDatum.subRows = subRows
```

packages/table/src/HeaderRow/HeaderRow.tsx
spark33: No because HeaderCell needs to know the rendered order of the cell to determine the padding. If we removed this, we would need the user to pass in the index themselves but I figured we'd prioritize DX here. 

One other approach we could take is to use CSS `:first-of-type`, but it would also have to know if the cell is expandable, or is a checkbox cell, etc. That would take a bit more changes and I think that might be a nice small future improvement?

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -0,0 +1,173 @@
...
+  LGRowData,
+  LGTableDataType,
+} from '../useLeafyGreenTable';
+import Table, { flexRender, getCoreRowModel, getSortedRowModel } from '..';
+
+import processColumns from './processColumns';
+import processData from './processData';
+import { V11AdapterProps } from './V11Adapter.types';
+
+// assumes table is first element in children
```
TheSonOfThomp, line 26: What happens if this is not the case?

packages/table/src/Row/InternalRowBase.tsx
spark33: The theme var is consumed throughout RowWithRT and RowWithoutRT, so it wouldn't make sense to apply all styles here. For example, you can't use CSS `:nth-of-type(even)` with virtual rows, since the zebra style is dependent on the index of the virtual row, not the index of the tr in the DOM.

packages/table/src/Row/InternalRowWithRT.tsx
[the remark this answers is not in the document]
spark33: No, but the browser will flag as a HTML error like 
image

packages/table/src/V11Adapter/index.ts
```diff
@@ -0,0 +1,3 @@
+import V10Adapter from './V11Adapter';
+
+export default V10Adapter;
```
TheSonOfThomp, line 3: V11Adapter?

packages/table/src/Row/InternalRowWithoutRT.tsx
spark33: Looked into it - it's going to be a bit tricky and probably not something we can change quickly now. RT doesn't expose API to know the rendered order of a cell/column because it supports reordering of columns using an array of columns. So to get the rendered order of a cell, it would have to do a search for a nested field in an array of columns for each cell, which is a pretty expensive calculation.

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -0,0 +1,173 @@
...
+  });
+
+  const { rows } = table.getRowModel();
+  let loopedItems;
+
+  if (useVirtualScrolling) {
+    loopedItems = table.virtualRows ?? [];
+  } else {
+    loopedItems = rows;
+  }
```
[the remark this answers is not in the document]
TheSonOfThomp: Also, the name `loopedItems` feels strange

packages/table/src/TableContainer/TableContainer.tsx
spark33: Arbitrary, but the TableContainer needs a height or max-height set to make `position: sticky` work on the header. If developers can change this pretty easily using the `className` if they want, so I didn't think this would be a problem.

packages/table/src/Table.story.tsx
spark33: There's an `overflow: hidden` on the td which you can see with the ID field on other stories. In this story, it should probably not overflow so I'm going to change that.

TheSonOfThomp (commented):
Just need the upgrade guide 👍

TheSonOfThomp (commented):
Just need the upgrade guide 👍

packages/table/src/TableContainer/TableContainer.tsx
spark33: Actually, `max-height` is better here. Gonna change this

packages/table/src/Table.story.tsx
brookescarlett: We need some design pattern around what happens with the Table extends past the edge of the viewport

packages/table/src/V11Adapter/V11Adapter.tsx
brookescarlett: same

packages/table/src/index.ts
brookescarlett: ^

packages/table/src/Cell/Cell.tsx
```diff
@@ -0,0 +1,39 @@
...
+import {
+  alignmentStyles,
+  baseCellStyles,
+  baseTableSidePadding,
+  cellContentContainerStyles,
+  cellContentTransitionStyles,
+} from './Cell.styles';
+import { CellProps } from '.';
+
+const Cell = ({ className, align, children, ...rest }: CellProps) => (
```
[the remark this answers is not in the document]
spark33: The standalone Cell is rendered when the user isn't using RT, so this is basically doing what `InternalCellBase` used to do in the old implementation. And since this component renders a `td` all the time, the browser will surface that the component shouldn't be used outside a `tr` - an error that  says a `td` cannot be a child of a `div` for example.

packages/table/src/HeaderRow/HeaderRow.tsx
spark33: @TheSonOfThomp Good call! Adding that.

packages/table/src/Row/ExpandableContent.spec.tsx
spark33: ^that's right!

packages/table/src/Row/ExpandableContent.spec.tsx
spark33: I did this since the test render functions for each of these use cases can get pretty long, so I thought breaking it out by functionality (since they're pretty clearly separate) would be better than having a 500+ line test file

packages/table/src/Row/InternalRowWithRT.tsx
spark33: And potentially the toggle icon button. Will look into this!

packages/table/src/Row/Row.styles.ts
spark33: Turns out this isn't even needed since the Cell is what ended up handling the transitions!

packages/table/src/Row/Row.types.ts
```diff
@@ -0,0 +1,31 @@
...
+   */
+  row: LeafyGreenTableRow<T>;
+  /**
+   * Virtual row object passed from the `useLeafyGreenTable` hook
+   */
+  virtualRow?: VirtualItem;
+}
+
+export type RowProps<T extends LGRowData> = PropsWithChildren<
+  InternalRowWithoutRTProps & Partial<InternalRowWithRTProps<T>>
```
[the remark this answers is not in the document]
spark33: Because when a user renders a `Row`, they only sometimes pass in `row` and `virtualRow`

packages/table/src/Row/RowCellChildren.tsx
```diff
@@ -0,0 +1,78 @@
...
+interface RowCellChildrenProps<T extends LGRowData>
+  extends PropsWithChildren<{
+    row: LeafyGreenTableRow<T>;
+    disabled?: boolean;
+  }> {}
+
+/**
+ * Renders row cells provided by `useReactTable`
+ */
+const RowCellChildren = <T extends LGRowData>({
```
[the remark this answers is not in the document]
spark33: Haha you'd know better than me since you added this! Is there more to this decision than organizing the markup and logic to improve legibility?

packages/table/src/TableBody/TableBody.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+        );
+      };
+
+      const { container } = render(<DefaultTest />);
+      // would be better to check explicitly if it's a fragment
+      expect(container.querySelector('tbody')).toBeInTheDocument();
+    });
+  });
+
+  test.todo('Virtual scrolling features');
```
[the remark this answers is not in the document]
spark33: I don't believe this is possible using just Jest and RTL since we'd have to simulate scrolling on a set viewport height. I also get warnings from RTL when I try to test anything with VS that tests for server-side rendered markup won't be reliably tested.

packages/table/src/TableContainer/TableContainer.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { forwardRef, PropsWithChildren } from 'react';
+
+import { css, cx } from '@leafygreen-ui/emotion';
+
+import TableContainerProps from './TableContainer.types';
+
+const TableContainer = forwardRef<HTMLDivElement, TableContainerProps>(
```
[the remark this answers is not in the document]
spark33: I think it would make sense for the Table component to render a `<table>`. This component is also only needed for sticky headers and virtual scrolling - the basic use case without RT doesn't need this

packages/table/src/TableHead/TableHead.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { PropsWithChildren } from 'react';
+
+import { cx } from '@leafygreen-ui/emotion';
+import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+
+import { stickyStyles, themeStyles } from './TableHead.styles';
+import { TableHeadProps } from './TableHead.types';
+
+const TableHead = ({
```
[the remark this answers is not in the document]
spark33: Yeah that's an interesting point. It could feel redundant, but I think there are a few reasons why we might want this.

The first is to replicate how HTML markup works. You'd have
```
<thead>
    <tr>...</tr>
</thead>
``` 
in HTML, so it may be redundant markup, but it almost feels more intuitive to have parity to me and it would probably be the same for anyone that's written pure HTML tables.

The other is that although this feature is not currently designed/supported by us, both HTML tables and React-Table support [column groups](https://tanstack.com/table/v8/docs/examples/react/column-groups). In this case, it would be useful to not restrict the thead to just having one row.

packages/table/src/V11Adapter/V11Adapter.tsx
spark33: Since the V11Adapter is used pretty intentionally by the developer, I assume this should be an okay assumption to make. It's just a part of how the V11Adapter is expected to take in, like an API if you will.

packages/table/src/V11Adapter/V11Adapter.tsx
spark33: Yeah I did this because then the type would be: `const [processedData,]: [Array<LGTableDataType<T>>, Dispatch<...>] ` instead of what it is now. It's really a tradeoff of whether we want the `[0]` to be more clear or the type to be easier to decipher. I leaned more to the type clarity personally, but I'll change it and see if y'all disagree

packages/table/src/V11Adapter/V11Adapter.tsx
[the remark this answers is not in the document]
spark33: Going with `iterables` for now - hopefully you like that better haha

packages/table/src/V11Adapter/index.ts
spark33: 🤦‍♂️

packages/table/src/TableContainer/TableContainer.tsx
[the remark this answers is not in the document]
brookescarlett: I feel strongly that we should bake this in. We uphold this pattern with our input components too, and that's fine. It's undo burden on the consumer to know when/when not to use the TableContainer.

packages/table/src/TableBody/TableBody.spec.tsx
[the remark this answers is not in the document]
brookescarlett: leave a comment with this, then

packages/table/src/V11Adapter/V11Adapter.tsx
brookescarlett: we should probably document this better then

packages/table/src/V11Adapter/index.ts
brookescarlett: nit: `export {default} from './`

packages/table/src/TableContext/TableContext.types.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * The `useLeafyGreenTable` return value
+     */
+    table?: LeafyGreenTable<T>;
+  };
+
+export const initialTableContext: TableContextValues<LGRowData> = {
+  getRowById: (_?: string) => undefined,
+  getParentRow: (_?: string) => undefined,
```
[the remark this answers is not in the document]
spark33: I'm thinking I'll just use a partial with the same props interface for the initial state in that case

packages/table/src/TableV10/index.ts
```diff
@@ -0,0 +1,6 @@
+export { default as Cell } from './Cell';
+export { default as HeaderRow } from './HeaderRow';
+export { default as Row } from './Row';
+export { default as Table } from './Table';
+export { DataType } from './TableContext';
+export { default as TableHeader } from './TableHeader';
```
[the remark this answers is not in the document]
spark33: Yeah that should be right, except I removed the test files and the one v10 story that had been broken for as long as I can remember

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  if (useVirtualScrolling) {
+    const { rows } = table.getRowModel();
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    rowVirtualizer = useVirtual({
+      parentRef: containerRef,
+      size: rows.length,
+      overscan: 30,
+    });
+  }
```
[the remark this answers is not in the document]
spark33: I'm a fan of the #2 approach. That way we're not breaking the rule of hooks either

packages/table/src/Cell/Cell.styles.ts
spark33: `Align` is `HTMLElementProps<'td'>['align']`, so that won't work

packages/table/src/utils/makeData.tsx
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
+import { faker } from '@faker-js/faker';
+
+import Code from '@leafygreen-ui/code';
```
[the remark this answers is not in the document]
spark33: Where would the TS error come from? This is not being exported and only being used in Storybook. I'm not seeing any TS errors on my end

packages/table/src/Table.story.tsx
```diff
@@ -1,363 +1,838 @@
...
+            {info.getValue() ? 'Enabled' : 'Not enabled'}
+          </Badge>
+        ),
+      },
+      {
+        accessorKey: 'mdbVersion',
+        header: 'MongoDB Version',
+        enableSorting: true,
+        size: 90,
+      },
```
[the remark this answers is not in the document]
brookescarlett: was this addressed?

packages/table/src/utils/makeData.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Maybe it was just on my end 🤷

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+    ...table,
+    ...(rowVirtualizer && {
+      virtualRows: rowVirtualizer.virtualItems,
+      totalSize: rowVirtualizer.totalSize,
+    }),
+    hasSelectableRows,
+  } as LeafyGreenTable<T>;
+}
+
+useLeafyGreenTable.propTypes = {
```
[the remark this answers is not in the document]
brookescarlett: i don't believe so?

packages/table/src/V11Adapter/V11Adapter.tsx
TheSonOfThomp: You can pass a type into `useState` itself to get the types
```
const processedData = useState<Array<LGTableDataType<T>>>(...)
```

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
TheSonOfThomp: @brookescarlett When a user clicks the "select all" button, we don't want all the other checkboxes to animate. So we disable animation when _all_ the rows are selected

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -0,0 +1,92 @@
+import { TransitionStatus } from 'react-transition-group';
+
+import { css } from '@leafygreen-ui/emotion';
+import { spacing, transitionDuration } from '@leafygreen-ui/tokens';
+
+import { Align } from './HeaderCell/HeaderCell.types';
```
TheSonOfThomp, line 6: nit: I think the `Align` type should live in `Cell.types`

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
brookescarlett: ah ok makes sense

packages/table/src/V11Adapter/V11Adapter.tsx
TheSonOfThomp: ^ and check for errors

packages/table/src/Row/RowCellChildren.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Not really, just makes it cleaner

packages/table/src/Cell/Cell.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+          <Cell {...props} data-testid="lg-test-cell" />
+        </tr>
+      </tbody>
+    </table>,
+  );
+}
+
+describe('packages/table/Cell', () => {
+  describe('a11y', () => {
+    test('does not have basic accessibility issues', async () => {
```
brookescarlett, line 27: should we test that it renders its children?

packages/table/src/Cell/HeaderCell/HeaderCell.types.ts
```diff
@@ -0,0 +1,37 @@
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+
+import { Header } from '../..';
+import { LGRowData } from '../../useLeafyGreenTable';
+
+export const SortState: { [key: string]: string } = {
```
TheSonOfThomp, line 6: Don't need to type this if we add `as const`. That'll get inferred

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
+# v10 to v11
```
TheSonOfThomp, line 1: "Upgrading Table v10 to v11"

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+v11 introduces a large set of new features including virtualized scrolling and sticky header rows. For a functional overview, refer to the [project brief](https://docs.google.com/document/u/1/d/1AaZfYAGi9MCxU-cutWovDwTl_4jViUP34QwMFiWMSxU/edit).
```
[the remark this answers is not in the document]
TheSonOfThomp: "... a large set of new features, and API changes including ... and a composition-based API"

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Overview
+
+The largest change is that the usage API has changed to a set of UI components that wrap HTML Table elements (e.g. `<thead>` equates to the `<TableHead>` component). This should improve the DX of creating basic tables.
```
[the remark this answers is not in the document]
TheSonOfThomp: * i.e. `<thead>` equates...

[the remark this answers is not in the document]
TheSonOfThomp: "We believe this approach will improve the developer experience (DX) of creating basic tables"

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+For more complex functionalities, v11 exports the `useLeafygreenTable` hook which leverages the [react-table](https://tanstack.com/table/v8) and [react-virtual]() libraries. As of v11.0.0, the hook supports:
```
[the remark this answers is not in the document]
TheSonOfThomp: New line before "As of..."

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+- virtualized scrolling
+- nested rows
+- expanded content
+- sortable rows
+- selectable rows
+
+MongoDB developers should not utilize other `react-table` that are not specified in LeafyGreen's Storybook, as the use cases specified in Storybook are the only ones that have been designed for according to guidelines.
```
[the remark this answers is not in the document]
TheSonOfThomp: Maybe link to the published storybook?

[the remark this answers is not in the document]
TheSonOfThomp: "...as other features may not follow our design guidelines"

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Other changes
+
+- **Zebra striping behavior is no longer applied by default for data sets over 10 rows.** Use the `shouldAlternateRowColor` prop on the `Table` component to achieve the same effect. Note that we discourage the usage of zebra stripes in conjunction with nested rows or expandable content.
```
TheSonOfThomp, line 21: "Alternating row colors (or "zebra striping")..."

TheSonOfThomp: add why we discourage zebra with expandable content

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **Disabled cells are no longer supported.** We found in our design audit that this prop was not being used. We recommend applying styles that match our disabled rows if you would like to disable specific cells.
```
TheSonOfThomp, line 23: Maybe leave out the design audit bit? Just say "We recommend [xyz]"

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **Multi-row headers are no longer supported.** We found in our design audit that this prop was not being used. Reach out to the Design Systems team if this prop is crucial to your team's needs.
```
TheSonOfThomp, line 25: Same here

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+### `V11Adapter`
+
+The `V11Adapter` was created to allow v10 Table components to utilize v11 features with minimal effort.
```
TheSonOfThomp, line 29: This feels like it's phrased backwards. I would say something like  "...was created to allow developers to upgrade to v11 without needing to refactor all tables to the new API"

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+
+Given the two versions' significant differences in API, the adapter makes several assumptions about the v10 Table's usage:
+
+- It is assumed that the v10 Table component will be the first child.
```
TheSonOfThomp, line 33: "The v10 Table component must be the first (and only?) child of the adapter"

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+- The v11 columns are read from the v10 columns' labels. If the key of the cells' data does not correspond to the v10 column's label, the user is expected to pass in the labels through the `headerLabels` prop.
```
TheSonOfThomp, line 34: "Column definitions will be read from the v10 Table `column` prop label" (or something like that. Using v11 & v10 a lot is a bit confusing)

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
...
+            return <Cell key={`${cellKey}-${index}`}>{row[cellKey]}</Cell>;
+          })}
+        </Row>
+      ))}
+    </TableBody>
+  </Table>
+</V11Adapter>;
+```
+
+Refer to the LeafyGreen Storybook deployment for more use cases.
```
TheSonOfThomp, line 80: Link out to SB deployment

packages/table/UPGRADE.md
```diff
@@ -0,0 +1,80 @@
+# v10 to v11
+
+v11 introduces a large set of new features including virtualized scrolling and sticky header rows. For a functional overview, refer to the [project brief](https://docs.google.com/document/u/1/d/1AaZfYAGi9MCxU-cutWovDwTl_4jViUP34QwMFiWMSxU/edit).
+
+## Overview
```
TheSonOfThomp, line 5: Should we mention how to migrate from v10 to v11? Maybe two code blocks that render the same table in each version?

spark33: Could be useful, but I think it might be needlessly verbose for an upgrade guide. There are READMEs for both versions and examples of both versions on SB, so I feel like this wouldn't be necessary

TheSonOfThomp: That's fair, as long as both code samples are easily accessible 👍

TheSonOfThomp (approved):

packages/table/src/Cell/Cell.spec.tsx
spark33: No need to at the Cell level - Cell just returns `children` without any processing. All rendering overrides are done at the Row level or above.

packages/table/src/Table.story.tsx
spark33: Going to keep the `overflow: hidden` on the Cell, and ask the developers to implement ellipses behavior on their end because adding `text-overflow: ellipses` requires us have a `width: _px` set on the div, which we can't do because we don't know the width of the Cell

packages/table/src/Row/InternalRowWithRT.tsx
spark33: I added this to the row's checkbox, but it won't be applicable to any other place because `aria-controls` wants the `id` of the element that's controlled (meaning if the element controls multiple elements, `aria-controls` won't be relevant).

alvenw (approved):

PR #1671, opened 2023-03-28
Updates missing dependencies

TheSonOfThomp:
Dependencies were missing from some packages

### DIFF
--- a/packages/palette/package.json
+++ b/packages/palette/package.json
@@ -21,7 +21,10 @@
   },
   "devDependencies": {
     "@leafygreen-ui/emotion": "^4.0.3",
-    "polished": "^4.2.2"
+    "polished": "^4.2.2",
+    "lodash": "^4.17.21",
+    "@leafygreen-ui/tokens": "^2.0.1",
+    "@leafygreen-ui/tooltip": "^9.1.6"
   },
   "gitHead": "dd71a2d404218ccec2e657df9c0263dc1c15b9e0",
   "homepage": "https://github.com/mongodb/leafygreen-ui/tree/main/packages/palette",
--- a/packages/hooks/package.json
+++ b/packages/hooks/package.json
@@ -30,5 +30,8 @@
   },
   "bugs": {
     "url": "https://jira.mongodb.org/projects/PD/summary"
+  },
+  "devDependencies": {
+    "@leafygreen-ui/emotion": "^4.0.3"
   }
 }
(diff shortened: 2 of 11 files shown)


packages/number-input/package.json
```diff
@@ -1,38 +1,46 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/a11y": "^1.4.1",
+    "@leafygreen-ui/button": "^20.0.2",
+    "@leafygreen-ui/emotion": "^4.0.3",
+    "@leafygreen-ui/hooks": "^7.6.0",
+    "@leafygreen-ui/icon": "^11.12.4",
+    "@leafygreen-ui/leafygreen-provider": "^3.1.1",
```
brookescarlett, line 36: why is this a dep and not a peerDep?

brookescarlett (commented):
Does this script update tsconfigs as well?

packages/table/src/Cell/HeaderCell/HeaderCell.spec.tsx
```diff
@@ -6,7 +6,7 @@ import { axe } from 'jest-axe';
 import {
   getTestColumnsProps,
   useTestHookCall,
-} from '../../utils/testHookCalls';
+} from '../../utils/testHookCalls.testutils';
```
brookescarlett, line 9: intentional?

TheSonOfThomp: Yeah, that's the only reliable way to tell depcheck to ignore certain files (without having a giant list of specific files)

packages/number-input/package.json
TheSonOfThomp: Good catch. Will fix

TheSonOfThomp:
> Does this script update tsconfigs as well?

It can yeah. I didn't run that in this PR, but I probably should

packages/table/src/Cell/HeaderCell/HeaderCell.spec.tsx
brookescarlett: ah got it

packages/palette/package.json
```diff
@@ -21,7 +21,10 @@
   },
   "devDependencies": {
     "@leafygreen-ui/emotion": "^4.0.3",
-    "polished": "^4.2.2"
+    "polished": "^4.2.2",
```
brookescarlett, line 24: nit order?

brookescarlett (approved):

PR #1674, opened 2023-03-31
LG-2983 Select padding

TheSonOfThomp:
## ✍️ Proposed changes

Updates padding to the right of the icon

🎟 _Jira ticket:_ https://jira.mongodb.org/browse/LG-2983

Screenshot 2023-03-31 at 3 15 25 PM
Screenshot 2023-03-31 at 3 15 19 PM


packages/select/src/MenuButton.tsx
```diff
@@ -39,28 +40,32 @@ const menuButtonStyleOverrides = css`
   }
 `;
 
+// We don't have 12px or 10px in our spacing vars
+const _12 = spacing[2] + spacing[1];
+const _10 = spacing[2] + spacing[1] / 2;
```
spark33, line 45: Not a big fan of these variants - I think I'd rather see a more descriptive name like `const rightPadding = 10` and see that reused throughout.

spark33 (changes requested):
Not a big fan of these variants - I think I'd rather see a more descriptive name like `const rightPadding = 10` and see that reused throughout.

spark33 (approved):

PR #1680, opened 2023-04-06
GH actions updates

TheSonOfThomp:
## ✍️ Proposed changes

This _should_ make GH actions jobs more reliable


.github/workflows/nodejs.yml
```diff
@@ -9,214 +9,276 @@ concurrency:
...
+        name: Check for build cache
         id: build-cache
         with:
           path: |
             **/dist
             !**/node_modules
           key: ${{ runner.os }}-build-cache-${{ hashFiles('**/rollup.config.js', '**/babel.config.js', '**/src/', '**/scripts/', '!**/node_modules', '!**/dist') }}
+
+      # Only setup & build if there was no build cache hit
+      - name: Use Node.js 14
```
brookescarlett, line 112: will there ever be a cache hit at this point in the workflow? i thought github actions didn't persist caches between workflows, but i could definitely be wrong about that

brookescarlett: nvm i think i am wrong

brookescarlett: but i do think we need to give this cache a key?

.github/workflows/nodejs.yml
```diff
@@ -9,214 +9,276 @@ concurrency:
...
-      - name: tests
-        run: yarn test --ssr --diff --deps --no-cache --ci --runInBand --reporters=default --reporters=jest-junit
-        env:
-          JEST_JUNIT_OUTPUT: 'reports/junit/js-test-ssr-results.xml'
+        if: ${{ steps.build-cache.outputs.cache-hit != 'true' }}
+        run: yarn build:packages
 
-  lints:
-    name: Check lints
+  build-ts:
```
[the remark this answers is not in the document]
bruugey: and if something breaks in building, it'll throw right?

.github/workflows/nodejs.yml
bruugey: i thoguht `key` was required: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#input-parameters-for-the-cache-action

.github/workflows/nodejs.yml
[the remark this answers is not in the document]
TheSonOfThomp: 👍 correct

bruugey (approved):

PR #1683, opened 2023-04-07
Improper usage warning: Toast

TheSonOfThomp:
## ✍️ Proposed changes

Adds a warning when toast is used outside of ToastProvider context

Enhances useToast test suites

### For new components


packages/toast/src/ToastContext/ToastContext.types.ts
```diff
@@ -20,6 +20,8 @@ export interface ToastContextProps {
   getToast: (id: ToastId) => ToastProps | undefined;
   getStack: () => ToastStack | undefined;
   clearStack: () => void;
+  /** @private */
+  _hasProvider?: boolean;
```
spark33, line 24: Do we need this internal flag? Could we detect it by calling `useToastContext` and `console.warn`ing if the right variables aren't defined?

spark33 (changes requested):

packages/toast/src/ToastContext/ToastContext.types.ts
TheSonOfThomp: All the vars will be defined, but I changed it to check whether `getStack` returns something

packages/toast/src/ToastContext/useToast/useToast.ts
```diff
@@ -1,7 +1,18 @@
 import { useContext } from 'react';
+import { isUndefined } from 'lodash';
 
 import { ToastContext } from '../ToastContext';
+import { ToastContextProps } from '../ToastContext.types';
 
-export const useToast = () => {
-  return useContext(ToastContext);
+export const useToast = (): Omit<ToastContextProps, '_hasProvider'> => {
```
spark33, line 7: I don't think this has to be omitted any more!

spark33 (approved):

PR #1692, opened 2023-04-17
LG-2524: SegmentedController size prop from "small" to "xsmall"

brookescarlett:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2524](https://jira.mongodb.org/browse/LG-2524)

### For new components


packages/segmented-control/src/SegmentedControl.story.tsx
```diff
@@ -21,7 +25,7 @@ export default {
     size: {
       control: {
         type: 'radio',
-        options: Object.values(Size),
+        options: [...Object.values(Size), 'small'],
```
spark33, line 28: `options: [...Object.values(Size), ...Object.values(DeprecatedSize)]`?

brookescarlett: ah left that in let me rm

brookescarlett: i was testing but we don't want that in storybook

TheSonOfThomp (approved):

PR #1697, opened 2023-04-18
LG-2198: BaseFontSize controls on Storybook

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2198](https://jira.mongodb.org/browse/LG-2198)

Should be reviewed together with https://github.com/mongodb/design/pull/251

### For new components


packages/side-nav/src/SideNav.story.tsx
```diff
@@ -24,6 +24,10 @@ export default {
   },
   argTypes: {
     darkMode: storybookArgTypes.darkMode,
+    baseFontSize: {
+      control: { type: 'radio' },
+      options: [14, 16],
+    },
```
TheSonOfThomp, line 30: Why aren't we using the `storybookArgTypes` property here?

packages/text-input/src/TextInput.story.tsx
```diff
@@ -50,7 +50,7 @@ export default {
       options: Object.values(SizeVariant),
     },
     baseFontSize: {
-      control: 'select',
+      control: 'radio',
```
TheSonOfThomp, line 53: We should be consistent with how we enable `baseFontSize` in stories

packages/lib/src/storybook/storybookArgTypes.ts
```diff
@@ -1,14 +1,22 @@
 import { InputType } from '@storybook/csf';
 import IntrinsicElements from './IntrinsicElements';
+import { BaseFontSize } from '@leafygreen-ui/tokens';
 export interface StoryArgType extends InputType {
   /**
    * Identify an arg to render a control on Storybook only,
    * and not on `mongodb.design`
```
TheSonOfThomp, line 7: Can we update the docs?

packages/lib/src/storybook/storybookArgTypes.ts
```diff
@@ -1,14 +1,22 @@
...
+  displayedPlatforms?: 'storybookOnly' | 'websiteOnly';
 }
 
 export const storybookArgTypes: { [key: string]: StoryArgType } = {
+  websiteBaseFontSize: {
+    description:
+      'The base font size passed to the LeafyGreenProvider that wraps the component',
+    control: { type: 'radio' },
+    options: Object.values(BaseFontSize),
+    displayedPlatforms: 'websiteOnly',
```
TheSonOfThomp, line 18: Do we a actually only want this on the website? I think it's fine to also have on Storybook itself

packages/lib/package.json
```diff
@@ -24,6 +24,7 @@
     "react": "^17.0.0"
   },
   "dependencies": {
+    "@leafygreen-ui/tokens": "^2.0.1",
```
TheSonOfThomp, line 27: I don't love importing other LG packages from `lib`, since it can introduce circular dependencies, but this one seems fine

TheSonOfThomp (commented):
We should be consistent with how we implement `baseFontSize` as an `argType` (`storybookArgTypes` import vs inline)

packages/lib/package.json
spark33: I did this so we can define the `baseFontSize` storybookArgType as `options: Object.values(BaseFontSize)`, but I don't think it's worth the risk. I've duplicated the font size values here

PR #1700, opened 2023-04-20
Updates Code example

TheSonOfThomp:
## ✍️ Proposed changes

### For new components


packages/lib/src/storybook/StoryMeta.ts
```diff
@@ -37,9 +37,10 @@ export interface StoryMeta<
   component?: T;
 }
 
-const baseMeta: Partial<StoryMeta<any>> = {
+export const baseMeta: Partial<StoryMetaType<any>> = {
   argTypes: {
-    ...storybookArgTypes,
+    darkMode: storybookArgTypes.darkMode,
+    baseFontSize: storybookArgTypes.updatedBaseFontSize,
```
spark33, line 43: Doesn't this mean we have to explicitly exclude `baseFontSize` in every component even if they don't support the prop? I'm assuming that we envision every component's story meta extending this object in the future

bruugey (approved):

PR #1703, opened 2023-04-21
LG-3106, LG-3121, LG-3122, LG-3123: Table 12.0.0

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3106](https://jira.mongodb.org/browse/LG-3106)
🎟 _Jira ticket:_ [LG-3121](https://jira.mongodb.org/browse/LG-3121)
🎟 _Jira ticket:_ [LG-3122](https://jira.mongodb.org/browse/LG-3122)
🎟 _Jira ticket:_ [LG-3123](https://jira.mongodb.org/browse/LG-3123)

### For new components


packages/table/src/Cell/Cell.tsx
```diff
@@ -16,7 +17,16 @@ const Cell = ({ className, align, children, ...rest }: CellProps) => (
...
       css`
-        padding-left: ${baseTableSidePadding}px;
+        &:first-child {
+          ${getCellPadding({
+            depth: 0,
+            isExpandable: false,
+            isSelectable: false,
+          })}
+        }
+        &:last-child {
```
bruugey, line 27: nit: can we move this into baseCellStyles?

packages/table/src/Cell/Cell.tsx
```diff
@@ -5,27 +5,20 @@ import { css, cx } from '@leafygreen-ui/emotion';
...
-      className,
-    )}
-    {...rest}
-  >
+  <td className={cx(baseCellStyles, basicCellStyles, className)} {...rest}>
     <div
       className={cx(
         cellContentContainerStyles,
-        cellContentTransitionStyles['entered'],
+        css`
```
bruugey, line 19: should this be baked into cellContentContainerStyles?

packages/table/src/TableWithVS.story.tsx
```diff
@@ -1,28 +1,24 @@
 import React from 'react';
 import { VirtualItem } from 'react-virtual';
```
bruugey, line 2: I feel like we should export this too? I don't want anyone having to use any external package..

packages/table/src/Row/InternalRowWithoutRT.tsx
```diff
@@ -31,7 +31,9 @@ const InternalRowWithoutRT = ({
       {...rest}
     >
       {React.Children.map(children, (child: ReactNode, index: number) => {
-        return <Cell {...(child as ReactElement).props} cellIndex={index} />;
+        return (
+          <Cell key={`cell-${index}`} {...(child as ReactElement).props} />
```
bruugey, line 35: intentional to rm cellIndex prop?

packages/table/src/Row/Row.stories.tsx
bruugey, line 1: should we be consistent about file names, i'm seeing both .stories and .story here

packages/table/src/Row/InternalRowWithoutRT.tsx
spark33: Yes. `cellIndex` is only used in `InternalCell`, so `InternalRowWithRT`

packages/table/src/Row/Row.stories.tsx
spark33: `*.stories.tsx` is actually the Storybook default/standard, and `*.story.tsx` is an old pattern we need to move away from soon. We're only keeping it now because it's significant overhead to change it all (from the actual name change to making script updates), but we've been using it interchangeably even recently (e.g. Adam's compiled `stories.js` files), so I don't think it's an issue.

bruugey (approved):

PR #1715, opened 2023-04-27
LG-2140: Adds fontWeights to tokens package

bruugey:
## ✍️ Proposed changes

### For new components


packages/typography/src/Body/Body.tsx
```diff
@@ -22,12 +23,12 @@ const fontWeights: Record<
...
 > = {
   default: {
-    regular: 400,
-    medium: 500,
+    regular: fontWeightTokens.regular,
+    medium: fontWeightTokens.medium,
   },
   strong: {
-    regular: 700,
-    medium: 700,
```
spark33, line 30: nit: there's an opportunity here to type the `weight` prop as `keyof fontWeightTokens` and simplify the CSS on line 49 to:

```
font-weight: ${fontWeightTokens[weight]`
strong, b {
   font-weight: ${fontWeightTokens['bold']
}
```

spark33 (approved):

packages/typography/src/Body/Body.tsx
bruugey: can definitely be stricter about the ts but i believe it'd look like: `(typeof fontWeightTokens)[keyof typeof fontWeightTokens]`

bruugey: actually cleaned it up pretty significantly - can you take another look @spark33?

spark33 (approved):

PR #1721, opened 2023-05-01
LG-3108: modal styles updates

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3108](https://jira.mongodb.org/browse/LG-3108)

### For new components


bruugey:
Spoke to @audroue and we actually think in this case it makes sense to just remove the p-specific styling from dark mode. Instead, let's keep the theme defaults as they are and allow consuming applications to override this as they wish by providing a className to whatever typography elements they are passing to Modal components as children. Let me know if you disagree @DesignerDave

DesignerDave:
@brookeyalof I think keeping a sensible default for the mode is something we've heard a number of asks for in the past, and I believe we do this for components like Tooltip.

Also worth noting that I would consider it a breaking change to remove a default styling for dark mode. Existing dark mode modals would be, in theory, broken with the new version if they relied on this styling.

What problem are we solving by removing this?

DesignerDave:
@bruugey We have a pattern for making changes to components within others now. In Popover, for example, we use a context to change how the component behaves based on being rendered in a Modal. We should probably be setting a default using a similar approach, or generally be changing the default color for our Paragraph component based on the LeafyGreenProvider value itself.

Either way, this would be a breaking change if we pursue either approach.

bruugey:
Sounds like what you're talking about @DesignerDave is adaptive defaults (our components smartly rendering themselves based on their context). We have a few tickets that get at this (callouts being placed after headers, tables having proper padding when rendered inside of cards). The scope of that work feels to big for this ticket here. I think either we can remove the p specific styles as we're still assigning the bounding `<div>{children}</div>` or explore the solution proposed in this PR, but I do think that assigning all inheriting p tags a specific color is not the right go forward.

.changeset/angry-teachers-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/modal': patch
```
bruugey, line 2: breaking

.changeset/angry-teachers-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/modal': major
+---
+
+Removes the default `P` styles applied in darkMode that used to override the color of all `P` tags with `#C1C7C6`. Child element colors are now respected and if no colors are explicitly defined for them, they will inherit the color of their parent container. Also reorganizes internal file structure.
```
bruugey, line 5: nit: lc p

packages/modal/package.json
```diff
@@ -36,7 +36,7 @@
   "devDependencies": {
     "@leafygreen-ui/select": "^10.3.1",
     "@leafygreen-ui/button": "^20.0.4",
-    "@leafygreen-ui/typography": "^16.2.1",
+    "@leafygreen-ui/typography": "^16.3.0",
```
bruugey, line 39: why do we have to manually version bump this?

bruugey (approved):

PR #1725, opened 2023-05-02
LG-3057: tab closes menu

shaneeza:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/packages/menu/src/Menu.spec.tsx
+++ b/packages/menu/src/Menu.spec.tsx
@@ -102,18 +102,28 @@ describe('packages/menu', () => {
     });
   });
 
+  type Keys = 'esc' | 'tab';
+  const tests = [['esc'], ['tab']] as Array<Array<Keys>>;
+
   describe('Keyboard Interaction', () => {
-    describe('Escape key', () => {
+    const userEventInteraction = (menu: HTMLElement, key: Keys) => {
+      if (key === 'esc') {
+        userEvent.type(menu, '{esc}');
+      } else {
+        userEvent.tab();
+      }
+    };
+
+    describe.each(tests)('%s key', key => {
       test('Closes menu', async () => {
         const { getByRole, getByTestId } = renderMenu({
           trigger,
         });
         const button = getByRole('button');
-
         userEvent.click(button);
         const menu = getByTestId(menuTestId);
-        userEvent.type(menu, '{esc}');
 
+        userEventInteraction(menu, key);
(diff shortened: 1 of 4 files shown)


packages/menu/src/Menu.spec.tsx
```diff
@@ -102,18 +102,28 @@ describe('packages/menu', () => {
     });
   });
 
+  type Keys = 'esc' | 'tab';
+  const tests = [['esc'], ['tab']] as Array<Array<Keys>>;
```
bruugey, line 106: nit: 
`const tests: Array<Array<Keys>> =`

packages/menu/src/MenuItem/MenuItem.tsx
```diff
@@ -78,7 +78,7 @@ export const MenuItem = InferredPolymorphic<MenuItemProps, 'button'>(
     const baseProps = {
       ref,
       role: 'menuitem',
-      tabIndex: disabled ? -1 : undefined,
+      tabIndex: -1,
```
bruugey, line 81: why always -1?

packages/menu/src/Menu.spec.tsx
```diff
@@ -125,7 +135,8 @@ describe('packages/menu', () => {
         const button = getByRole('button');
         userEvent.click(button);
         const menu = getByTestId(menuTestId);
-        userEvent.type(menu, '{esc}');
+
```
bruugey, line 138: can we add the same test for when usePortal is false?

shaneeza: there is a false test, its just not showing up in this PR since i didn't touch that line

shaneeza: `test('Returns focus to trigger {usePortal: false}', async () => {`

packages/menu/src/MenuItem/MenuItem.tsx
shaneeza: We only need the tab index for tabbing with the tab key, but since we got rid of it, we don't need to be able to tab organically since we're doing it programmatically on arrow click. This change isn't necessary, so I can revert if there are any issues with it.

packages/menu/src/Menu.spec.tsx
bruugey: ah ok

bruugey (approved):

PR #1735, opened 2023-05-09
Menu updates [LG-3190, LG-3161, LG-3193, LG-3168]

TheSonOfThomp:
## ✍️ Proposed changes

🎟 _Jira ticket:_ 
- Arrow keys & Space/Enter keyboard interactions: https://jira.mongodb.org/browse/LG-3161
- Clicking a submenu opens that submenu: https://jira.mongodb.org/browse/LG-3190 (aligns behavior to existing tests)
- Submenu chevron DOM nesting: https://jira.mongodb.org/browse/LG-3168
- Clicking a menu item selects that item: https://jira.mongodb.org/browse/LG-3193

### For new components


packages/menu/src/Menu.spec.tsx
```diff
@@ -65,94 +84,207 @@ describe('packages/menu', () => {
...
+        openMenu();
+        const menu = getByTestId(menuTestId);
+        const options = globalGetAllByRole(menu, 'menuitem');
+
+        userEvent.type(options[0], `{enter}`);
+        expect(triggerButton).toHaveFocus();
+
+        // FIXME: I can verify that this works in the browser,
+        // but this does not work in jest - AT
+        // await waitForElementToBeRemoved(menu);
```
TheSonOfThomp, line 220: I couldn't get these tests to pass even though it works in the browser, and the mouse interaction version of this test is passing (L143)

packages/menu/src/Menu.story.tsx
```diff
@@ -28,7 +28,7 @@ export default StoryMeta({
     },
   },
   args: {
-    open: true,
+    open: undefined,
```
bruugey, line 31: false?

packages/menu/src/Menu/Menu.tsx
```diff
@@ -102,6 +103,10 @@ export function Menu({
     uncontrolledSetOpen;
   const open = controlledOpen ?? uncontrolledOpen;
 
+  const closeMenu = useCallback(() => {
```
bruugey, line 106: I think in this case useCallback is actually less performant than just using setOpen: https://kentcdodds.com/blog/usememo-and-usecallback

packages/menu/src/SubMenu/SubMenu.styles.ts
```diff
@@ -10,12 +16,6 @@ export const chevronClassName = createUniqueClassName('icon-button-chevron');
...
 
-export const subMenuStyle = css`
-  padding-right: ${iconButtonContainerSize + 16}px;
-  align-items: center;
-  justify-content: flex-start;
-`;
-
 export const subMenuThemeStyle: Record<Theme, string> = {
   [Theme.Light]: cx(
     subMenuStyle,
```
bruugey, line 21: you rm'd subMenuStyle but are referencing it

packages/menu/src/SubMenu/SubMenu.tsx
```diff
@@ -201,17 +196,22 @@ export const SubMenu = InferredPolymorphic<SubMenuProps, 'button'>(
...
           {...baseProps}
-          {...anchorProps}
           {...rest}
           className={cx(
             subMenuContainerClassName,
             menuItemContainerStyle,
             menuItemContainerThemeStyle[theme],
             menuItemHeight(size),
             linkStyle,
+            subMenuStyle,
```
bruugey, line 214: ditto here

packages/menu/src/SubMenu/SubMenu.tsx
```diff
@@ -223,29 +223,29 @@ export const SubMenu = InferredPolymorphic<SubMenuProps, 'button'>(
...
+            iconButtonClassName,
+            iconButtonStyle,
+            iconButtonThemeStyle[theme],
+            iconButtonFocusedThemeStyle[theme],
+            iconButtonPosition[size],
+            {
+              [openIconButtonStyle[theme]]: open,
+            },
+          )}
+          onClick={handleChevronClick}
```
bruugey, line 241: need to stop prop here, because this is firing the onClick for the submenu (check .design example)

packages/menu/src/SubMenu/SubMenu.tsx
```diff
@@ -89,22 +90,20 @@ export const SubMenu = InferredPolymorphic<SubMenuProps, 'button'>(
...
 
     const onRootClick = useCallback(
       (
         e: React.MouseEvent<HTMLAnchorElement, MouseEvent> &
           React.MouseEvent<HTMLButtonElement, MouseEvent>,
       ) => {
-        if (iconButtonElement?.contains(e.target as HTMLElement)) {
-          e.preventDefault();
-        } else if (onClick) {
+        if (onClick) {
```
bruugey, line 100: nit: onClick?.(e)

packages/menu/src/SubMenu/SubMenu.tsx
```diff
@@ -153,15 +151,12 @@ export const SubMenu = InferredPolymorphic<SubMenuProps, 'button'>(
       'aria-disabled': disabled,
       // only add a disabled prop if not an anchor
       ...(typeof rest.href !== 'string' && { disabled }),
+      ...(isAnchor && {
```
bruugey, line 154: nit: we can do here what you did with submenu

bruugey: double nit: these components should probably be unified :O

packages/menu/src/Menu.story.tsx
TheSonOfThomp: It should be undefined so it's uncontrolled by default

packages/menu/src/Menu/Menu.tsx
TheSonOfThomp: Ah that's fair. I initially had written some side effects, but removed them

packages/menu/src/SubMenu/SubMenu.tsx
TheSonOfThomp: I think this has been fixed

packages/menu/src/SubMenu/SubMenu.tsx
TheSonOfThomp: Yes they should be... but that's out of scope for this work :upside-down-face:

bruugey:
Arrow keying through a closed submenu still stops on collapsed menu items

bruugey:
Another bug: If I add a click event to the sub menu then clicking anywhere on the submenu component (other than the arrow button) does not open/close the sub menu; however, if i add a click handler on the submenu component, then clicking anywhere on the submenu button does not open/close the submenu component. Not sure which is right, but it should probably be consistent.

packages/menu/src/SubMenu/SubMenu.spec.tsx
```diff
@@ -1,113 +1,288 @@
...
+        data-testid={subMenuTestId[1]}
+        href="https://mongodb.design"
+        title="SubMenu B"
+      >
+        <MenuItem data-testid={menuItemTestId[1]}> Text Content B</MenuItem>
       </SubMenu>
-      <SubMenu data-testid={subMenu2Id} href="mongodb.design">
-        <MenuItem data-testid={menuItem2Id}> Text Content B</MenuItem>
+      <SubMenu
+        active={initialActiveMenu === 1}
```
shaneeza, line 51: should this be `===2`?

packages/menu/src/SubMenu/SubMenu.spec.tsx
```diff
@@ -1,113 +1,288 @@
...
+      });
+    });
+
+    test('Opening a Submenu closes the previous one', async () => {
+      const { getByTestId, queryByTestId, openMenu } = renderSubMenu();
+      openMenu();
+      const subMenuB = getByTestId(subMenuTestId[1]);
+      const chevronB = globalGetByTestId(
+        subMenuB.parentElement!,
+        'lg-sub-menu-icon-button',
```
shaneeza, line 249: nit: this could also go in a variable

TheSonOfThomp: I didn't put these in a variable because the value other ones don't really matter—they could be any string, as long as the string is consistent throughout the file. This id is defined in the component. I feel it's a bit confusing if changing some vars is safe, but changing others is not

packages/menu/src/Menu.spec.tsx
shaneeza: this seems to work `userEvent.click(options[0]);`

PR #1739, opened 2023-05-12
Removes call to use React's useId if found

bruugey:
useIdAllocator not compat with React 18 in current version

Repro case here (install Modal locally versus from NPM): https://codesandbox.io/s/green-moon-0j46x2?file=/src/App.tsx


TheSonOfThomp (approved):
👍

PR #1745, opened 2023-05-15
LG-3064: Fix Inline Props

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3064](https://jira.mongodb.org/browse/LG-3064)

### For new components


.changeset/lemon-rules-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/typography': patch
+---
+
+Fixes `InlineCodeProps` such that it does not result in a complex union type
```
TheSonOfThomp, line 5: I feel like this should be triggering a `@ts-expect-error` somewhere

TheSonOfThomp: If not, can we add a type test?

TheSonOfThomp (approved):

PR #1751, opened 2023-05-16
Chromatic decorator [LG-3233]

TheSonOfThomp:
# Summary

Creates a decorator that generates a story to test all prop combinations of a component.

## Motivation
Storybook & Chromatic provide no way to programmatically define stories for multiple variants. For example, to test the `Button` component with every combination of variant, size, darkMode, icon position etc., we would have to explicitly write each of these stories in the `*.stories.tsx file`, resulting in a lot of manual overhead work. In addition, testing each of these individually would result in hundreds, or even thousands, of stories which would dramatically increase the cost of Chromatic, and generates noise on the sidebar.

Screenshot 2023-05-26 at 2 41 09 PM

Screenshot 2023-06-01 at 1 12 27 PM


.github/workflows/nodejs.yml
```diff
@@ -170,7 +170,6 @@ jobs:
     name: Chromatic
     runs-on: ubuntu-latest
     needs: [install, build]
-    if: ${{ false }} # TODO: re-enable when we're ready for Chromatic
```
TheSonOfThomp, line 173: We're ready for Chromatic now!

.storybook/decorators/GeneratedStory/GeneratedStory.styles.ts
```diff
@@ -0,0 +1,64 @@
+import { css } from '@leafygreen-ui/emotion';
+import { createUniqueClassName } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+import { typeScales } from '@leafygreen-ui/tokens';
+
+const indent = 16;
+
+export const generatedStoryWrapper = css`
+  /* padding: ${indent}px; */
```
shaneeza, line 9: rm?

TheSonOfThomp: I'm going to keep the variable to avoid merge conflicts (since I use it in the next PR), but I'll get rid of the contets

TheSonOfThomp: I'm going to keep the variable to avoid merge conflicts (since I use it in the next PR), but I'll get rid of the rules

.changeset/ninety-peaches-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/button': patch
+---
+
+updates story file and implements GeneratedStory
```
bruugey, line 5: nit: "U"

.github/workflows/nodejs.yml
bruugey: 🎉

.storybook/decorators/ComponentPreview.tsx
```diff
@@ -17,21 +18,17 @@ const darkModeStyle = css`
...
-      baseFontSize={baseFontSize}
-    >
-      <div className={cx(rootStyle, { [darkModeStyle]: darkMode as boolean })}>
-        <div>
-          <Story darkMode={darkMode} {...options} />
-        </div>
+    <LeafyGreenProvider darkMode={darkMode} baseFontSize={baseFontSize}>
+      <div
+        className={cx({
+          [rootStyle]: context.name !== GENERATED_STORY_NAME,
```
bruugey, line 27: what's this doing?

.storybook/decorators/GeneratedStory/GeneratedStory.styles.ts
```diff
@@ -0,0 +1,62 @@
+import { css } from '@leafygreen-ui/emotion';
```
bruugey, line 1: Should we be committing generated files?

.storybook/decorators/ComponentPreview.tsx
TheSonOfThomp: We only want to apply the story wrapper styles for non-generated stories

.storybook/decorators/GeneratedStory/GeneratedStory.styles.ts
TheSonOfThomp: No?

.storybook/decorators/PropCombinations/README.md
```diff
@@ -0,0 +1,275 @@
...
+use `generate.args` to pass static props that are different from the main story args.
+
+```tsx
+const meta: StoryMeta<typeof Tooltip> = {
+  component: Tooltip,
+  parameters: {
+    default: 'LiveExample',
+    generate: {
+      props: { ... },
+      args: {open: true}
```
bruugey, line 103: I'm not entirely clear on the difference between props and args

.storybook/decorators/PropCombinations/README.md
```diff
@@ -0,0 +1,275 @@
...
+  },
+};
+export default meta;
+```
+
+## `decorator`
+
+`(InstanceFn: StoryFn, context?: {args: Args}) => JSX.Element`
+
+The GeneratedStory decorator will _not_ read the decorators value defined in `meta.decorators`. To add a wrapper around each instance, define a decorator within `generate`.
```
bruugey, line 114: Why not?

.storybook/decorators/PropCombinations/utils/shouldExcludePropCombo.spec.ts
bruugey, line 1: Should we have a spec file that confirms we're generating the correct number of stories (with and without excludes?)

.storybook/decorators/PropCombinations/PropCombinationsDecorator.tsx
```diff
@@ -0,0 +1,97 @@
...
+      } = decoratorConfig;
+      if (isGeneratedStory(context)) {
+        // Check for props
+        if (!props || entries(props).length === 0) {
+          return Err('`props` not found in story generation parameters.');
+        }
+
+        // Remove any args that are explicitly defined
+        for (let propName in { ...generatedArgs }) {
+          if (props[propName]) {
```
bruugey, line 59: this looks like we're letting args overwrite props, not the other way around (which the comment implies)

.storybook/decorators/PropCombinations/PropCombinationsDecorator.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return <StoryFn />;
+};
+
+export default PropCombinationsDecorator;
+
+/**
+ * Sorts the darkMode prop to the top
+ */
+function sortDarkMode([propA]: [string, any], [propB]: [string, any]): number {
```
bruugey, line 93: How come?

.storybook/decorators/PropCombinations/utils/shouldExcludePropCombo.ts
```diff
@@ -0,0 +1,144 @@
...
+        /**
+         * Case 2: Conditional props
+         */
+        const checkProp = rule.find(x => typeof x === 'string');
+        const conditions = rule.find(y => typeof y === 'object');
+
+        // Assert types
+        if (
+          isUndefined(checkProp) ||
+          typeof checkProp !== 'string' ||
```
bruugey, line 49: wont checkProp always be a string or undefined based on line 43?

packages/button/src/Button.story.tsx
```diff
@@ -1,24 +1,53 @@
...
 import { Size } from './types';
 import Button, { ButtonProps, Variant } from '.';
 
 const meta: StoryMetaType<typeof Button> = {
   title: 'Components/Button',
   component: Button,
+
+  parameters: {
+    default: 'LiveExample',
+    generate: {
```
bruugey, line 21: Can we only run this if there's an environment variable set to Chromatic?

.storybook/decorators/PropCombinations/README.md
```diff
@@ -0,0 +1,275 @@
+# Prop Combinations Decorator
+
+## Motivation
```
bruugey, line 3: Can we add documentation for argTypes too?

.changeset/ninety-peaches-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/button': patch
+---
+
+Updates story file and implements GeneratedStory
```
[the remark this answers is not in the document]
TheSonOfThomp: I think using button as a POC is ok. They'll be merged into main at the same time

.storybook/decorators/PropCombinations/README.md
TheSonOfThomp: Maybe change `props` to `combine` or `combinations`?
I think `args` needs to stay for consistency with the rest of the storybook api.

@shaneeza @spark33  thoughts?

bruugey: I think it's cleaner to use props for everything and if we're not offering multiple values for a given prop that's okay.

.storybook/decorators/PropCombinations/README.md
TheSonOfThomp: The `meta.decorators` object doesn't exist on the context we get in the decorator function. (My guess is because there would be a self reference in the object)?

.storybook/decorators/PropCombinations/utils/shouldExcludePropCombo.ts
TheSonOfThomp: Yes, but TS doesn't know that

packages/button/src/Button.story.tsx
TheSonOfThomp: Technically yes, but why?

.storybook/decorators/PropCombinations/README.md
TheSonOfThomp: ? There's no `argTypes` property specific to the decorator

.storybook/decorators/PropCombinations/PropCombinationsDecorator.tsx
bruugey: leave that as a comment here

.storybook/decorators/PropCombinations/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+Storybook & Chromatic provide no way to programmatically define stories for multiple variants. For example, to test the `Button` component with every combination of variant, size, darkMode, icon position etc., we would have to explicitly write each of these stories in the `*.stories.tsx file`, resulting in a lot of manual overhead work. In addition, testing each of these individually would result in hundreds, or even thousands, of stories which would dramatically increase the cost of Chromatic, and generates noise on the sidebar.
```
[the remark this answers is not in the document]
DesignerDave: I think I agree with Brooke here - this section increases the ratio of noise to signal when trying to understand how to use this.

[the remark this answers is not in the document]
DesignerDave: All that's needed to understand the docs here is the section that reads:
> This custom decorator reads metadata from the storybook default export, and generates a single story that renders all combinations of props defined in the story metadata.

.storybook/decorators/PropCombinations/PropCombinationsDecorator.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import { generatedStoryWrapper } from './PropCombinations.styles';
+import { Err, PropCombinations } from './components';
+import { isGeneratedStory } from './utils';
+import { PARAM_NAME } from './constants';
+
+type ContextType<T extends React.ComponentType<any>> = StoryContext<T> &
+  StoryMetaType<T>;
+
+const PropCombinationsDecorator: Decorator = (
```
DesignerDave, line 30: I know this is in reference to Chromatic's application of decorators, but I would expect this to be used at some point as a decorator as defined by [TS](https://www.typescriptlang.org/docs/handbook/decorators.html) or [JS](https://github.com/tc39/proposal-decorators) specs/documentation. Can we address this somehow in the documentation to set folks who are new to this functionality, but familiar with the language construct, up for success?

As is, I would expect to be able to prefix something in the story file with `@PropCombinationsDecorator()` to use this (without deep-diving into Storybook's usage of decorators).

.storybook/decorators/PropCombinations/README.md
DesignerDave: I'd expect to be able to specify what properties should remain static as a part of the props definition. Why would it be preferred to specify this through args, rather than simply setting a single value within the prop values arrays? See my example with `darkMode` below:
```js
const meta: StoryMeta<typeof Button> = {
  component: Button,
  parameters: {
    default: 'LiveExample',
    // Define combinations of props to generate
    generate: {
      props: {
        darkMode: [true], // This prop value is unchanging through all instances
        rightGlyph: [undefined, <Icon glyph={'ArrowRight'} />],
        leftGlyph: [undefined, <Icon glyph={'Cloud'} />],
        children: ['MongoDB', undefined],
        size: Object.values(Size),
        variant: Object.values(Variant),
      },
    },
  },
};
export default meta;
```

.storybook/decorators/PropCombinations/README.md
```diff
@@ -0,0 +1,275 @@
...
+    },
+  },
+};
+```
+
+# Options
+
+## `args`
+
+`Record<key in keyof ComponentProps, ComponentProps[key]>`
```
DesignerDave, line 91: Imo showing TS types within documentation isn't that helpful without IntelliSense. Can we include this in a more human-readable format?

.storybook/decorators/PropCombinations/components/PropCombinations.tsx
```diff
@@ -0,0 +1,88 @@
...
+  console.log(
+    `Rendering ${comboCount} prop combinations for component: ${component.displayName}`,
+  );
+
+  return AllCombinations;
+
+  /**
+   * Recursively loop through all prop combinations defined in `variables` and render them
+   */
+  function RecursiveCombinations(
```
DesignerDave, line 38: Nit – I'm not in love with unreachable statements in written code, even if hoisting fixes it under the hood.

We also seem to be passing some values as arguments to this function (like variables), and leveraging others that are in scope where this function is defined (like `args`). Can we take one approach or the other here? Bonus points if we move this function outside of the `PropCombinations` function altogether and make it more testable on its own.

.storybook/decorators/PropCombinations/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+There are three methods of defining `excludeCombinations`.
+
+### Mutually Exclusive combinations
+
+Mutually exclusive combinations check that only one of the defined props exist on the instance. If they are both defined, then the instance is ignored.
+
+For example, a component might accept a `children` prop, or a `title` prop, but never both.
+
+Define mutually exclusive combinations as a tuple of prop keys
```
DesignerDave, line 173: Since this is defined as a tuple instead of an array, that implies we'll never want to define a set of 3 mutually-exclusive props here, which I don't see a whole lot of reason to prevent that here (though ideally we shouldn't have that case much).

Additionally, tuples are usually an ordered pair, while this doesn't appear to care about the order.

.storybook/decorators/PropCombinations/components/PropCombinations.tsx
TheSonOfThomp: For recursive functions I think scoping it within a closure is the best way to handle variables that need to be accessible regardless of recursion level. The alternative is passing the same prop down each iteration, which afaik is not recommended.

The other functions defined below could be defined externally and take a pass through prop

.storybook/decorators/PropCombinations/README.md
TheSonOfThomp: It's not strictly a tuple, no, just an array of length 2.

I never came across a situation where more than 2 props were mutually exclusive. We could update the function if we ever need this

.storybook/decorators/PropCombinations/README.md
TheSonOfThomp: "Object of prop keys"

.storybook/decorators/PropCombinations/PropCombinationsDecorator.tsx
TheSonOfThomp: Is this just a comment on the naming convention? `PropCombinationsStorybookDecorator`?

.storybook/decorators/PropCombinations/README.md
TheSonOfThomp: `argTypes` isn't used in the decorator at all. I've only added a convenience type in this PR

.storybook/decorators/PropCombinations/README.md
TheSonOfThomp: This feels a little odd to me:
```
      props: {
        open: [true],
        darkMode: [false, true],
        baseFontSize: Object.values(BaseFontSize),
        align: Object.values(Align),
        justify: Object.values(Justify),
      },
```

packages/button/src/Button.story.tsx
TheSonOfThomp: AFAIK there's no significant performance bottleneck for adding more stories to the sidebar.

Also, IMO, it's good to have these tests available while developing so we can quickly see if the changes we've made are correct

.storybook/decorators/PropCombinations/README.md
spark33: I think I'm with Adam on this one. `props` and `args` are sometimes used interchangably in vernacular, but by definition, `properties` are what is accepted, and `arguments` are what's actually passed. It's also how Storybook uses the words. I think the DX is cleaner when you separate the two, and the words aren't confusing to me personally.

.storybook/decorators/PropCombinations/components/PropCombinations.tsx
TheSonOfThomp: As for were the function is defined, I see 2 camps, each answering a different question when reading the outer function.

When I read the outer function, I ask:

Q1:  "what is this function doing"?
Q2: "where is this dependency defined"

As written, this function optimizes for answering Q1 while leaving Q2 harder to answer.
Moving the recursive function definition to the top, and return statement to the bottom reverses this and optimizes for Q2, while making Q1 harder to answer.

IMO Q1 is the more important question

.storybook/decorators/PropCombinations/README.md
spark33: Maybe `array of two elements` would be clearer here. `tuple` is a very specific data type, so I agree it can be confusing.

packages/button/src/Button.story.tsx
TheSonOfThomp: @TheSonOfThomp Verify this

bruugey (approved):
Stamping this as the API is largely good to go and want to get ahead of our integration workshop this afternoon

PR #1752, opened 2023-05-16
LG-3213: Allows null to be passed to SegmentedControl 

bruugey:
## ✍️ Proposed changes

Per React standard, null should be allowed to be passed as children to React components. https://legacy.reactjs.org/blog/2014/07/17/react-v0.11.html#rendering-to-null

🎟 _Jira ticket:_ [LG-3213](https://jira.mongodb.org/browse/LG-3213)

### For new components


zhaodext (approved):
Thanks for the change!

PR #1754, opened 2023-05-17
Update Storybook to 7.x LG-3203

TheSonOfThomp:
## ✍️ Proposed changes

Upgrades Storybook to v7.x, and upgrades dependencies & configs accordingly

🎟 _Jira ticket:_ [LG-3203](https://jira.mongodb.org/browse/LG-3203)

### For new components


bruugey:
Tempted to suggest that we merge this directly to main, as its pre-work for Chromatic but not directly related to the integration.

.github/workflows/nodejs.yml
```diff
@@ -104,7 +104,7 @@ jobs:
   build:
     name: Build packages
     runs-on: ubuntu-latest
-    needs: install
+    needs: [install, lints]
```
bruugey, line 107: do we need to lint first?

.github/workflows/nodejs.yml
```diff
@@ -137,7 +137,7 @@ jobs:
   build-ts:
     name: Build TypeScript
     runs-on: ubuntu-latest
-    needs: install
+    needs: [install, lints]
```
bruugey, line 140: ditto here

.github/workflows/nodejs.yml
```diff
@@ -166,6 +166,41 @@ jobs:
         if: ${{ steps.ts-cache.outputs.cache-hit != 'true' }}
         run: yarn build:ts
 
+  chromatic:
```
bruugey, line 169: this could be taken out of this PR, right?

.github/workflows/nodejs.yml
TheSonOfThomp: My thinking is why wast time building if the lints have failed?

.github/workflows/nodejs.yml
TheSonOfThomp: > Tempted to suggest that we merge this directly to main, as its pre-work for Chromatic but not directly related to the integration.

Hypothetically, yes this could go in straight to main. But it's already based off the Chromatic CI branch, and it'll be tough to untangle.

That said, it may be better to update to 7 sooner than later to avoid too many merge conflicts

TheSonOfThomp:
> Bugs on .design:
> * Combobox does not work when usePortal is false
✅

> * Should icon have a dark mode toggle?
Doesn't currently have a toggle.  Can fix this later

> * "background" doesn't work on Logo component
Hmm. Doesn't work on prod either

> * baseFontSize does not work on SegmentedControl
Also doesn't work on prod

> * Can't make selections with Select component
Also broken on prod...?

bruugey (approved):

PR #1755, opened 2023-05-17
Bumps required node version

TheSonOfThomp:
## ✍️ Proposed changes

Node 14 is EOL (end of live) as of April 30 2023: https://nodejs.dev/en/about/releases/
This updates our minimum node version to the oldest LTS (long term stable) version

https://jira.mongodb.org/browse/LG-3218


bruugey:
We have some products still on Node v14... let's talk about the feasibility of this

bruugey (approved):

PR #1758, opened 2023-05-19
LG-3034: Split button

shaneeza:
## ✍️ Proposed changes

[Demo](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-splitbutton--basic)

🎟 _Jira ticket:_ [LG-3034](https://jira.mongodb.org/browse/LG-3034)

localhost_3000_component_split-button_example_ (2)

localhost_3000_component_split-button_documentation_ (3)

### For new components


shaneeza:
menu: keydown focus, wedge

shaneeza:
@sandynguyenn @audroue Clicking the menu open with your mouse and then pressing the arrow key, which highlights the second item in the menu, is correct. Here is an example from [material-ui](https://mui.com/material-ui/react-menu/). When the menu opens on mouse click, the first item is the focused item, but there is no visual indicator since it was a mouse click. When you then click the down arrow, the second item becomes the focused item, and since it was done with keyboard interaction, you can see the focused styles.

shaneeza:
@sandynguyenn, the height of the wedge needs to be updated in the `<Menu />` component.  We can create a separate ticket for this.

packages/lib/src/index.ts
```diff
@@ -87,7 +87,11 @@ export function isComponentType<T = React.ReactElement>(
     element != null &&
     typeof element === 'object' &&
     'type' in element &&
-    (element.type as any).displayName === displayName
+    ((element.type as any).displayName === displayName ||
```
bruugey, line 90: reinforces for me that we should have a more elegant/clever way of checking a component's type (out of scope here)

packages/menu/src/Menu/Menu.tsx
```diff
@@ -286,12 +287,12 @@ export function Menu({
       case keyMap.Tab:
         e.preventDefault(); // Prevent tabbing outside of portal and outside of the DOM when `usePortal={true}`
         handleClose();
-        setFocus((refEl || triggerRef)?.current); // Focus the trigger on close
+        setFocus((triggerRef || refEl)?.current); // Focus the trigger on close
```
bruugey, line 290: why are we changing the order here?

packages/menu/src/Menu/Menu.tsx
```diff
@@ -334,6 +335,7 @@ export function Menu({
         {...popoverProps}
       >
         <div
+          id={id}
```
bruugey, line 338: is this the right element for the id?

packages/menu/src/MenuItem/MenuItem.tsx
```diff
@@ -156,6 +156,7 @@ export const MenuItem = InferredPolymorphic<MenuItemProps, 'button'>(
       </li>
     );
   },
+  'MenuItem',
 );
 
 MenuItem.displayName = 'MenuItem';
```
bruugey, line 162: is this line not relevant now?

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,108 @@
...
+          disabled={disabled}
+          leftGlyph={<Icon glyph="CaretDown" />}
+          variant={variant}
+          size={size}
+          baseFontSize={baseFontSize}
+          className={cx(triggerBaseStyles, triggerSizeStyles[size!], {
+            [triggerThemeStyles(theme, variant!)]: !disabled,
+          })}
+          aria-label="More options" // TODO: should this be a prop for consumers?
+          aria-haspopup={true}
```
bruugey, line 96: should we add this to menu itself?

packages/split-button/src/SplitButton/SplitButton.types.ts
```diff
@@ -0,0 +1,111 @@
...
+  Default: ButtonVariants.Default,
+  Primary: ButtonVariants.Primary,
+  Danger: ButtonVariants.Danger,
+} as const;
+
+export type Variant = typeof Variant[keyof typeof Variant];
+
+type ButtonProps = Omit<
+  ImportedButtonProps,
+  'rightGlyph' | 'href' | 'as' | 'variant'
```
bruugey, line 25: did we decide not to omit href and as?

packages/split-button/src/SplitButton/SplitButton.types.ts
```diff
@@ -0,0 +1,111 @@
...
+>;
+
+export type MenuItemsType = ReactFragment & {
+  props: {
+    children: Array<ComponentType<MenuItemProps>>;
+  };
+};
+
+export const Align = {
```
bruugey, line 34: why are we redefining align and justify?

packages/split-button/src/SplitButton/SplitButton.types.ts
```diff
@@ -0,0 +1,111 @@
...
+  | 'shouldClose'
+  | 'darkMode'
+  | 'onClick'
+  | 'align'
+  | 'justify'
+  | 'className'
+  | 'refEl'
+>;
+
+export interface MenuProps extends OmittedMenuProps {
```
bruugey, line 65: this reads weirdly MenuProps extends OmittedMenuProps should probably rename to SelectedMenuProps (or something similar)

packages/split-button/src/SplitButton/SplitButton.tsx
```diff
@@ -0,0 +1,134 @@
...
+      // only add these props if not an anchor
+      ...(!isAnchor && { disabled }),
+      ...(!isAnchor && { type }),
+    };
+
+    const sharedButtonProps = { variant, size, baseFontSize };
+
+    return (
+      <div
+        id="testetst"
```
bruugey, line 69: do we want to leave this in?

packages/menu/src/Menu/Menu.tsx
shaneeza: Since i'm using both a `triggerRef` and `refEl` inside `SplitButton`, the focus was going to the refEl on close. This way, the focus is on the trigger.

packages/menu/src/MenuItem/MenuItem.tsx
shaneeza: i just checked and i think its safe to remove for now

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,108 @@
...
+          type="button"
+          disabled={disabled}
+          leftGlyph={<Icon glyph="CaretDown" />}
+          variant={variant}
+          size={size}
+          baseFontSize={baseFontSize}
+          className={cx(triggerBaseStyles, triggerSizeStyles[size!], {
+            [triggerThemeStyles(theme, variant!)]: !disabled,
+          })}
+          aria-label="More options" // TODO: should this be a prop for consumers?
```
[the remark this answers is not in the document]
shaneeza: This would have to be called something like `triggerAriaLabel` since aria-label goes to the primaryButton

packages/split-button/src/Menu/Menu.tsx
shaneeza: No, i think this is correct, haspopup should be on the menu trigger

shaneeza: oh as in we add it to to the trigger, yeah we can do that

packages/split-button/src/SplitButton/SplitButton.types.ts
shaneeza: I'm using, `InferredPolymorphic` for `SplitButton`, so i decided to omit these props from `ButtonProps` since the types are different. `<Button />` uses `Box`, so the types don't match up.

packages/split-button/src/SplitButton/SplitButton.types.ts
shaneeza: I was initially doing this `export type Align = Extract<ImportedMenuProps['align'], 'top' | 'bottom'>` but that only gives me a literal union. I couldn't really do Align.Bottom., etc. Same with Justify and Variant. Is there a better way to do this?

packages/menu/src/Menu/Menu.tsx
bruugey: i think this could be a breaking change? is there a way to use the refEl to point to the container, not the buttons themselves?

packages/menu/src/Menu/Menu.tsx
```diff
@@ -378,6 +380,8 @@ export function Menu({
         onClick: triggerClickHandler,
         ref: triggerRef,
         children: popoverContent,
+        ['aria-expanded']: open,
```
bruugey, line 383: 👍

packages/menu/src/Menu/Menu.types.ts
```diff
@@ -45,4 +45,9 @@ export interface MenuProps extends Omit<PopoverProps, 'active'> {
    * default: `false`
    */
   darkMode?: boolean;
+
+  /**
+   * id passed to the menu dropdown.
```
bruugey, line 50: where are most props spread?

packages/split-button/README.md
```diff
@@ -0,0 +1,57 @@
...
+| `variant`         | `'default'` \| `'primary'` \| `'danger'`            | Sets the variant for both Buttons.                                                                                                                                                                                                               | `'default'` |
+| `size`            | `'xsmall'` \| `'small'` \| `'default'` \| `'large'` | Sets the size for both buttons.                                                                                                                                                                                                                  | `'default'` |
+| `align`           | `'top'` \| `'bottom'`                               | Determines the alignment of the menu relative to the component wrapper.                                                                                                                                                                          | `'bottom'`  |
```
bruugey, line 50: why are we exposing only a subset of the available alignments/justifications

packages/split-button/src/SplitButton.story.tsx
```diff
@@ -0,0 +1,99 @@
...
+export default StoryMeta({
+  title: 'Components/SplitButton',
+  component: SplitButton,
+  args: {
+    label: 'label',
+    variant: Variant.Default,
+    align: Align.Bottom,
+    justify: Justify.End,
+    spacing: 6,
+    menuItems: (
```
bruugey, line 19: I'm not convinced this is the best pattern still. I wonder if it should be an array of `<MenuItem />` components. Or at least if we can be opinionated about if it's a fragment or an array and not solve for both. You missed the RFC process @spark33  - what do you think?

packages/split-button/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "scripts": {
+      "build": "../../node_modules/.bin/rollup --config ../../rollup.config.js"
+    },
+    "license": "Apache-2.0",
+    "homepage": "https://github.com/mongodb/leafygreen-ui/tree/main/packages/split-button",
+    "repository": {
+      "type": "git",
+      "url": "https://github.com/mongodb/leafygreen-ui"
+    },
+    "bugs": {
```
bruugey, line 21: nit: is this still something we use?

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,107 @@
...
+  };
+
+  return (
+    <LGMenu
+      data-testid="lg-split-button-menu"
+      align={align}
+      justify={justify}
+      refEl={containerRef}
+      trigger={
+        <Button
```
bruugey, line 85: (if need be, we can turn this into its own component and use this: 
```
useEffect(() => {
    let unregisterRipple: (() => void) | undefined;
    const backgroundColor = rippleColors[theme][variant];

    if (rippleRef.current != null && !disabled) {
      unregisterRipple = registerRipple(rippleRef.current, {
        backgroundColor,
      });
    }

    return unregisterRipple;
  }, [rippleRef, variant, darkMode, disabled, theme]);
```
 to register a ripple

bruugey: Just checked and it looks like the ripple never gets properly deployed when a Button component is passed to the trigger prop of a Menu.

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,107 @@
...
+          type="button"
+          disabled={disabled}
+          leftGlyph={<Icon glyph="CaretDown" />}
+          variant={variant}
+          size={size}
+          baseFontSize={baseFontSize}
+          className={cx(triggerBaseStyles, triggerSizeStyles[size!], {
+            [triggerThemeStyles(theme, variant!)]: !disabled,
```
bruugey, line 93: we shouldn't need to coerce variant.. we can provide a default value to ensure its always here

packages/split-button/src/SplitButton/SplitButton.types.ts
bruugey: will you leave a TODO to update this once button is using polymorphic

packages/split-button/src/SplitButton/SplitButton.types.ts
bruugey: ah got it

packages/split-button/src/Menu/Menu.tsx
shaneeza: oddly, it works for <Tooltip />

packages/split-button/README.md
shaneeza: I'm going to double-check with Rob; I was going off of the specs in Figma which only showed alignment `top` and `bottom` and for the justification `start` and `end`

packages/split-button/package.json
shaneeza: it looks like it correctly links to the latest LG tickets in jira

packages/split-button/src/Menu/Menu.tsx
shaneeza: talked in person, we have a ticket and i will try to not use the trigger prop

packages/menu/src/Menu/Menu.tsx
shaneeza: Removing this

packages/split-button/src/SplitButton.story.tsx
TheSonOfThomp: My issue with an array is that it makes it incompatible with Menu. 

If a developer currently has a Menu that they want to change to a SplitButton, they can just copy/paste their MenuItems as they were in Menu. If we require an array they'll need to change (albeit slightly) how they pass the items in.

packages/menu/src/Menu/Menu.types.ts
shaneeza: i'll keep the id with `...rest` so it spreads on the `ul`

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,162 @@
...
+  return (
+    <>
+      <Button
+        type="button"
+        disabled={disabled}
+        leftGlyph={<Icon glyph="CaretDown" />}
+        variant={variant}
+        size={size}
+        baseFontSize={baseFontSize}
+        className={cx(triggerBaseStyles, triggerSizeStyles[size!], {
```
TheSonOfThomp, line 140: We should make an assertion elsewhere that `size` (and `variant`) exists. We should really only use type assertion in test/story files

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,162 @@
...
+          [triggerThemeStyles(theme, variant!)]: !disabled,
+        })}
+        aria-label="More options" // TODO: should this be a prop for consumers?
```
TheSonOfThomp: Yes, this should be configurable by the consumer. "More options" is a fine default, but there will probably be a better contextual label

packages/lib/src/index.ts
```diff
@@ -7,6 +7,7 @@ import getTheme from './getTheme';
 import { allEqual } from './allEqual';
 export { validateChildren } from './validateChildren';
 export { createSyntheticEvent } from './createSyntheticEvent';
+import isObject from 'lodash/isObject';
```
TheSonOfThomp, line 10: This should have been caught by the linter, but `imports` should come before `exports`

packages/lib/src/index.ts
```diff
@@ -87,7 +88,13 @@ export function isComponentType<T = React.ReactElement>(
     element != null &&
     typeof element === 'object' &&
     'type' in element &&
-    (element.type as any).displayName === displayName
+    // (element.type as any).displayName === displayName
+    ((element.type as any).displayName === displayName ||
+      // TODO: temp solution; Components using InferredPolymorphic have a displayName inside render.
```
TheSonOfThomp, line 93: 👍 Thanks for making the ticket

packages/split-button/package.json
TheSonOfThomp: Should we add a ticket to update this in the template?

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,165 @@
...
+  useEventListener('keydown', handleKeyDown, { enabled: open });
+  useBackdropClick(handleClose, [buttonRef, menuRef], open);
+
+  const renderMenuItems = () => {
+    const onMenuItemClick = (
+      e: MouseEvent,
+      menuItem: ReactElement<MenuItemProps>,
+    ) => {
+      handleClose();
+      // @ts-expect-error - onClick is inferred from inferredPolymorphic
```
TheSonOfThomp, line 84: We can resolve this by typing `menuItem` as:
```
menuItem: ReactElement<InferredPolymorphicProps<PolymorphicAs, MenuItemProps>>
```

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,165 @@
...
+    ) => {
+      if (isComponentType(menuItem, 'MenuItem')) {
+        return React.cloneElement(menuItem, {
+          active: false,
+          key: `menuItem-${index}`,
+          // @ts-expect-error - onClick is inferred from inferredPolymorphic
+          onClick: (e: MouseEvent) => onMenuItemClick(e, menuItem),
+        });
+      } else {
+        console.warn('Please use a LeafyGreen <MenuItem> component.');
```
TheSonOfThomp, line 100: Warning should be a bit more descriptive. Where is this originating? What was found instead?

packages/split-button/src/SplitButton/SplitButton.types.ts
```diff
@@ -0,0 +1,119 @@
...
+export type MenuItemsType = ReactFragment & {
+  props: {
+    children: Array<ReactElement<MenuItemProps>>;
+  };
+};
+
+export const Align = {
+  Top: ImportedAlign.Top,
+  Bottom: ImportedAlign.Bottom,
+};
```
TheSonOfThomp, line 37: `as const` ?

packages/split-button/src/SplitButton/SplitButton.types.ts
```diff
@@ -0,0 +1,119 @@
...
+export type Align = typeof Align[keyof typeof Align];
+
+export const Justify = {
+  Start: ImportedJustify.Start,
+  End: ImportedJustify.End,
+};
+
+export type Justify = typeof Justify[keyof typeof Justify];
+
+export type Test = typeof Justify;
```
TheSonOfThomp, line 48: rm?

packages/split-button/src/SplitButton/SplitButton.spec.tsx
```diff
@@ -0,0 +1,221 @@
...
+        disabled: true,
+        onClick,
+      });
+
+      fireEvent.click(primaryButton);
+      expect(onClick).toHaveBeenCalledTimes(0);
+    });
+  });
+
+  describe('menu', () => {
```
TheSonOfThomp, line 99: Should test Escape, Enter & Arrow Keys as well

TheSonOfThomp: Likely just a copy/paste from `Menu`, but we should test that we're not breaking something that's working in Menu

TheSonOfThomp (commented):
Should we accept a global `onChange` event on `SplitButton` that fires when a menu item is clicked? 

Also, I'm on board with menu items as a Fragment as opposed to an array

packages/split-button/src/SplitButton.story.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <>
+        {/* eslint-disable-next-line no-console */}
+        <MenuItem onClick={(event: MouseEvent) => console.log(event)}>
```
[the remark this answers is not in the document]
shaneeza: I did recently update my demo; maybe it needs a hard refresh?

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,153 @@
...
+  open: controlledOpen,
+  setOpen: controlledSetOpen,
+  ...rest
+}: MenuProps) => {
+  const { theme } = useDarkMode();
+  const [uncontrolledOpen, uncontrolledSetOpen] = useState<boolean>(false);
+  const buttonRef = useRef<null | HTMLButtonElement>(null);
+  const menuRef = useRef<null | HTMLDivElement>(null);
+
+  const setOpen =
```
bruugey, line 45: does the `useControlledValue` hook work here?

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,153 @@
...
+      case keyMap.Escape:
+        handleClose();
+        break;
+    }
+  };
+
+  useEventListener('keydown', handleKeyDown, { enabled: open });
+  useBackdropClick(handleClose, [buttonRef, menuRef], open);
+
+  const renderMenuItems = () => {
```
bruugey, line 74: useMemo

packages/split-button/src/SplitButton.story.tsx
```diff
@@ -0,0 +1,102 @@
...
+    justify: {
+      control: 'select',
+      options: Object.values(Justify),
+    },
+  },
+  parameters: {
+    default: 'Basic',
+    controls: {
+      exclude: [
+        'as',
```
bruugey, line 68: future fun: we should have a lint rule that auto sorts these

packages/split-button/src/SplitButton/SplitButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+      ? PropTypes.instanceOf(Element)
+      : PropTypes.any,
+  portalContainer:
+    typeof window !== 'undefined'
+      ? PropTypes.instanceOf(Element)
+      : PropTypes.any,
+  portalClassName: PropTypes.string,
+} as any;
+
+// TODO: see if its better to use an array or menuItems
```
bruugey, line 139: do we still need these TODOs?

packages/split-button/src/SplitButton.story.tsx
spark33: Yeah I'm with you @bruugey - I think we should stray from parsing children to dictate behavior in all cases. I'm okay with introducing a small migration cost to the user to avoid parsing children. I'd also be okay with accepting an array of "config" for MenuItem props, and we can create the MenuItem instances ourselves

packages/split-button/src/SplitButton.story.tsx
spark33: nit: I think we can add `...storybookExcludedControlParams` here to make this less verbose

packages/split-button/src/SplitButton/SplitButton.types.ts
spark33: and probably a ticket so we don't miss it

packages/menu/src/Menu/Menu.types.ts
```diff
@@ -45,4 +45,9 @@ export interface MenuProps extends Omit<PopoverProps, 'active'> {
...
+   */
+  id?: string;
```
spark33, line 52: Should MenuProps extend the HTMLElement props of the root element (i.e. the `div` inside the Popover) instead? It means our `...rest` on the `ul` might end up being more verbose, but it feels like it would allow the developer more flexibility and for us to not duplicate HTML element props like `id`

packages/split-button/src/SplitButton/SplitButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+// TODO: add test for menu closing
```
spark33, line 140: General component comment - when the variant is `primary` or `danger` and the menu button is hovered, that box shadow seems pretty intrusive. Is that something design specifically wanted?

packages/split-button/README.md
shaneeza: this is correct, we don't want consumers to select middle, center, etc...

packages/split-button/src/SplitButton.story.tsx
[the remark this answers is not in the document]
shaneeza: Where in `Menu` do we update the selection automatically? It looks like the consumer has to pass the `active` prop to the `MenuItem`(which we are removing in this component). But I'm not opposed to an `onChange` handler

packages/split-button/src/Menu/Menu.tsx
shaneeza: no, that's for input values. We could make a new hook for open/setOpen.

packages/split-button/package.json
shaneeza: what would we update it to? i can update it in this PR

packages/split-button/src/Menu/Menu.tsx
shaneeza: I'll update the types so that these are required. Since they are passed down from `SplitButton`, they should exist.

packages/menu/src/Menu/Menu.types.ts
shaneeza: Would it be ok if we're extending a `div` but spreading it on a `ul`? I guess the props would actually be the same.

packages/split-button/src/SplitButton.story.tsx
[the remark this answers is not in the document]
shaneeza: I added an `onChange` prop!

packages/menu/src/Menu/Menu.types.ts
bruugey: Will you file that ticket please

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+          return renderMenuItem(menuItem);
+        },
+      );
+    }
+  }, [handleClose, menuItems, onChange]);
+
+  return (
+    <>
+      <Button
```
bruugey, line 119: why not render this in the SplitButton file. That way you can share props between the two butttons and avoid passing all of these down

shaneeza: It seemed a little neater to keep all the opening and closing logic inside of `Menu` as opposed to having all of that inside `SplitButton`  and passing some of that down to `Menu`.

packages/split-button/src/Menu/Menu.tsx
```diff
@@ -74,58 +79,40 @@ export const Menu = ({
...
-          key: `menuItem-${index}`,
-          // @ts-expect-error - onClick is inferred from inferredPolymorphic
-          onClick: (e: MouseEvent) => onChildClick(e, menuItem),
+          key: `menuItem-${menuItem.key}`,
+          onClick: (e: MouseEvent) => onMenuItemClick(e, menuItem),
         });
       } else {
-        console.warn('Please use a LeafyGreen <MenuItem> component.');
+        console.warn(
+          'Please use a LeafyGreen <MenuItem /> component. Received: ',
```
TheSonOfThomp, line 98: People are probably going to want to use `SubMenu`, but we can deal with that later

TheSonOfThomp (approved):
LGTM

packages/menu/src/Menu/Menu.tsx
```diff
@@ -65,27 +65,30 @@ const scrollContainerStyle = css`
...
+    children,
+    className,
+    refEl,
+    trigger,
+    usePortal = true,
+    portalClassName,
+    portalContainer,
+    scrollContainer,
+    popoverZIndex,
+    maxHeight = 344,
```
bruugey, line 86: nit, but i prefer grouping aliases and defaults

sandynguyenn (approved):
lgtm

PR #1770, opened 2023-05-30
LG-3246: Add loading state to button

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3246](https://jira.mongodb.org/browse/LG-3246)

The designs specify that the button should not change widths when the button is toggled from `isLoading={false}` to `isLoading={true}`, unless the `loadingText` is explicitly defined:

image

[Demo](https://spark33.github.io/leafygreen-ui/?path=/story/components-button--loading-state)

Live Example:
image

### For new components


packages/button/README.md
```diff
@@ -91,7 +91,9 @@ npm install @leafygreen-ui/button
...
+| `rightGlyph`   | `React.ReactElement`                                                                       | Glyph that will appear to the right of text, if there is text provided via the children prop. If no children are supplied to the component, passing an Icon here will render the button as an icon-only button. |
+| `isLoading`    | `boolean`                                                                                  | Indicates whether the Button is in a loading state                                                                                                                                                              | `false`     |
+| `loadingText`  | `string`                                                                                   | String displayed in place of `children` while the button is in a loading state                                                                                                                                  |             |
```
bruugey, line 96: i don't support this API. why not pass through children

packages/button/src/Button/Button.tsx
```diff
@@ -22,13 +22,15 @@ export const Button = React.forwardRef(function Button(
...
     disabled = false,
+    isLoading = false,
     onClick,
     leftGlyph,
     rightGlyph,
     children,
     className,
     as,
     type,
+    loadingText,
```
bruugey, line 33: why are we using this instead of children?

packages/button/src/Button/Button.tsx
```diff
@@ -40,7 +42,7 @@ export const Button = React.forwardRef(function Button(
     size,
     darkMode,
     baseFontSize,
-    disabled,
+    disabled: isLoading || disabled,
```
bruugey, line 45: let's pull this into variable
`notInteractive = isLoading || disbaled`

packages/button/src/ButtonContent/ButtonContent.styles.ts
```diff
@@ -70,3 +70,37 @@ export const buttonContentSizeStyle: Record<Size, string> = {
...
+export const buttonSpinnerSize: Record<Size, number> = {
+  [Size.XSmall]: 16,
+  [Size.Small]: 16,
+  [Size.Default]: 16,
+  [Size.Large]: 20,
+};
+
+export const spinnerStyles: Record<Theme, string> = {
+  [Theme.Dark]: css`
+    path {
```
bruugey, line 97: should we be more specific than this?

bruugey: or provide a better way in the spinner API to change the color?

bruugey: or provide a better way in the spinner API to change the color?

bruugey: `colorOverride`

bruugey:
Also in storybook, why is darkmode an object?

.changeset/beige-books-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/loading-indicator': patch
+---
+
+Updates Storybook config to v7
```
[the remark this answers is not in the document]
spark33: No - these changes are new

packages/button/README.md
spark33: When there is no loading text, we want the button to maintain the width of its children when `isLoading` is true. This means we need some distinction between loadingText and the normally displayed text. We also don't want the user to have `isLoading ? 'some loading text' : 'normal text'` logic in their implementation. I figured this would be a cleaner DX.

packages/button/src/ButtonContent/ButtonContent.styles.ts
spark33: Fine by me either way. Spinner only has one path which makes the selector simple, so it feels like a non-issue. I think it's only going to be overridden internally too. I'm happy to add it if others agree with that approach

packages/button/src/Button/Button.tsx
```diff
@@ -22,13 +22,15 @@ export const Button = React.forwardRef(function Button(
     darkMode: darkModeProp,
     baseFontSize = BaseFontSize.Body1,
     disabled = false,
+    isLoading = false,
```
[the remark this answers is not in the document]
spark33: Searching in regex for `is[A-Z]` across all `*.types.ts` files shows we have more boolean props with the `is_` prefix than not. I think it's also good practice to indicate clearly that it's a boolean state prop and not anything else. 

I would argue that `open` not being `isOpen` is the anti-pattern that should be addressed. It feels like common sense to me that event handlers should be prefixed with `on_`, booleans with `is_`, and component props with capital letters unless the prop name makes it impossible to confuse. With `open`, `open={true}` could mean `isOpen`, `openable`, etc.

sooachung:
- [ ] The spinner inside of the 16font size button looks off-centered and the bottom gets cut off. Both in light and dark mode.
Screenshot 2023-05-31 at 1 39 37 PM

- [ ] Loading state with text's text seems thicker than it should be
Screenshot 2023-05-31 at 1 45 06 PM

- [ ] and the text in the small size button doesn't seem like it is middle aligned.
Screenshot 2023-05-31 at 1 45 50 PM

spark33:
> Also in storybook, why is darkmode an object?

Looks like this was something that was missed in our Storybook migration PR - I see this issue in Table, Stepper, Select, and probably more. I'll address this in Button now, but the other components should be addressed in a separate PR. I'll create a ticket.

packages/button/src/Button/Button.tsx
spark33: For future note - I'd be a fan of using explicit states here the way Adam did with Toasts. So we'd have a `state` prop that's either `disabled`, `loading`, or `none`.

packages/button/src/ButtonContent/ButtonContent.styles.ts
```diff
@@ -70,3 +70,37 @@ export const buttonContentSizeStyle: Record<Size, string> = {
     gap: 8px;
   `,
 };
+
+export const textlessLoadingStyles = (
+  loadingContentRef: React.MutableRefObject<HTMLDivElement | null>,
+) => css`
+  width: ${loadingContentRef.current?.style.width};
```
shaneeza, line 77: I don't see this showing up in the browser, do you need this? Wouldn't the width inherit the child width if there is button text?

packages/button/src/ButtonContent/ButtonContent.tsx
```diff
@@ -47,36 +55,64 @@ export const ButtonContent = ({
...
+              data-testid="lg-button-spinner"
+            />
+            {loadingText}
+          </div>
         )}
+        <div
+          ref={loadingContentRef}
+          className={cx(buttonContentStyle, buttonContentSizeStyle[size], {
+            [css`
+              justify-content: space-between;
```
shaneeza, line 84: Could these styles go inside the style file?

packages/loading-indicator/src/BlobLoader/BlobLoader.spec.tsx
```diff
@@ -1,15 +1,9 @@
...
 import { render } from '@testing-library/react';
 import { axe } from 'jest-axe';
-import { setupJestCanvasMock } from 'jest-canvas-mock';
 
 import BlobLoader from './BlobLoader';
 
-// this could be setup globally in jest.config, but done in this .spec file to pass depcheck
-beforeAll(() => {
-  setupJestCanvasMock();
-});
```
shaneeza, line 11: are you removing because you added this globally?

packages/button/src/ButtonContent/ButtonContent.styles.ts
shaneeza: i like the `colorOverride` approach!

.changeset/beige-books-fetch.md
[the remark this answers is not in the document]
spark33: Actually, we could just remove this changeset since the first release hasn't been done yet.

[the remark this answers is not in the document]
bruugey: let's leave this changeset, because we're releasing tomorrow and not sure if this PR will land by then

packages/button/src/Button/Button.tsx
TheSonOfThomp: Definitely +1 for `state = 'disabled' | 'loading' | 'none'`

packages/button/src/ButtonContent/ButtonContent.styles.ts
spark33: Good call! Thank you.

packages/button/src/Button/Button.tsx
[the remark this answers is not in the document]
TheSonOfThomp: +1 to using `is_` for booleans. Much clearer that it's referring to a boolean. Cases like `open` can be confusing, since it's unclear whether it's a state variable or a function to set the state

packages/button/src/Button/Button.tsx
spark33: https://jira.mongodb.org/browse/LG-3280

packages/button/src/ButtonContent/ButtonContent.tsx
```diff
@@ -47,36 +55,64 @@ export const ButtonContent = ({
...
       >
-        {leftGlyph && (
-          <ButtonIcon
-            glyph={leftGlyph}
-            className={css`
-              justify-self: right;
-            `}
-            {...iconProps}
-          />
+        {isLoading && (
```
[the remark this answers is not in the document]
spark33: My thought was that `isLoading` should take precedence. I'll double check with Sooa!

packages/button/src/Button.story.tsx
```diff
@@ -66,3 +73,14 @@ export const BaseGreen = Playground.bind({});
...
+export const LoadingState = Playground.bind({});
+LoadingState.args = {
+  isLoading: true,
+};
+
+export const LoadingStateWithText = Playground.bind({});
+LoadingStateWithText.args = {
+  isLoading: true,
+  loadingText: 'Saving',
+};
```
TheSonOfThomp, line 86: Just a note, I'm recommending (in the Chromatic adoption) moving away from stories where all that changes is the default args. We should make sure the knobs/controls are set up the way we want them, and allow users to toggle these states in the `Playground`/`LiveExample` story

packages/button/src/ButtonContent/ButtonContent.styles.ts
```diff
@@ -70,3 +74,38 @@ export const buttonContentSizeStyle: Record<Size, string> = {
...
+  [Size.Default]: 16,
+  [Size.Large]: 20,
+};
+
+export const spinnerColor: Record<Theme, string> = {
+  [Theme.Dark]: palette.gray.light1,
+  [Theme.Light]: palette.gray.dark1,
+};
+
+export const hideContentWhileLoading = css`
```
TheSonOfThomp, line 97: this style var should end in `...styles`. Something like `contentHiddenStyles`?

packages/button/src/Button/Button.tsx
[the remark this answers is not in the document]
bruugey: Can someone add this to our API style guide please

packages/button/src/ButtonContent/ButtonContent.tsx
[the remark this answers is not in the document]
spark33: Confirmed by @sooachung - the button is disabled when `isLoading={true}` anyway, so it should be shown.

packages/button/src/Button/Button.tsx
```diff
@@ -22,39 +22,42 @@ export const Button = React.forwardRef(function Button(
...
     type,
+    loadingText,
     ...rest
   }: BoxProps<'button', ButtonProps>,
   forwardRef,
 ) {
   const { darkMode } = useDarkMode(darkModeProp);
 
+  const isAnchor: boolean = (!!rest.href || as === 'a') && !disabled;
+  const isNotInteractive = disabled || isLoading;
```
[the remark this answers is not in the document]
bruugey: `isInteractive = !disabled && !isLoading`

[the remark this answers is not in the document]
bruugey: should `is_` be extended to `disabled`?

TheSonOfThomp:
> Right now the button width is responsive to the children prop when it explicitly should not be. I think we should combine the loadingText and children props, but definitely should not be setting the width based on a prop we're not even leveraging.

AFAIK, When `loadingText` is `false` the button width _should_ resize with respect to children. This is to when `isLoading` toggles to `false`, the button doesn't resize

packages/button/src/Button/Button.tsx
```diff
@@ -22,39 +22,42 @@ export const Button = React.forwardRef(function Button(
...
     className: cx(ButtonClassName, buttonStyles, className),
     ref: forwardRef,
     // Provide a default value for the as prop
     // If consuming application passes a value for as, it will override the default set here
     as: as ? as : ((isAnchor ? 'a' : 'button') as keyof JSX.IntrinsicElements),
-    'aria-disabled': disabled,
-    onClick: !disabled ? onClick : undefined,
-    href: disabled ? undefined : rest.href,
+    'aria-disabled': isNotInteractive,
+    onClick: !isNotInteractive ? onClick : undefined,
```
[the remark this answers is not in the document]
spark33: lol I didn't even realize. Maybe it should just be `const isInteractive = !(disabled || isLoading);`

jest.config.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
   // 1. `react-children-utilities`
   transformIgnorePatterns: ['/node_modules/(?!(react-children-utilities)/)'],
 
-  setupFiles: ['<rootDir>/scripts/jestSetup.js'],
+  setupFiles: ['<rootDir>/scripts/jestSetup.js', 'jest-canvas-mock'],
```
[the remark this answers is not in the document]
spark33: Yes. I realized this will break tests for any components that utilize Spinner (i.e. this PR broke Button tests). We're planning to bake Spinner into more components moving forward, so we should probably just set it up globally.

packages/button/src/Button.story.tsx
spark33: Gotcha - will keep that in mind for future PRs. I think the value of this approach is that it really highlights certain use cases, and in our current process of VD review, this is helpful for designers that aren't fully aware of how our props are set up to work. I'm sure this will be addressed largely by Chromatic itself, so I'm excited to see and discuss how our process would change.

packages/button/src/Button/Button.tsx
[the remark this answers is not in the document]
spark33: Totally with you - that was a miss.

packages/button/src/Button/Button.tsx
spark33: It might be easier to have an epic on Jira

packages/button/src/Button/Button.tsx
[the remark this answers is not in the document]
spark33: It's already [there](https://github.com/mongodb/leafygreen-ui/blob/main/STYLEGUIDE.md#use-be-verbs-for-boolean-types-eg-shouldiscandoeshas)

packages/loading-indicator/package.json
```diff
@@ -25,15 +25,12 @@
...
     "@leafygreen-ui/emotion": "^4.0.4",
-    "@leafygreen-ui/lib": "^10.0.0",
+    "@leafygreen-ui/lib": "^10.3.3",
     "@leafygreen-ui/palette": "^4.0.4",
     "@leafygreen-ui/tokens": "^2.1.0",
     "@leafygreen-ui/typography": "^16.4.0",
     "@lottiefiles/react-lottie-player": "^3.5.3"
   },
-  "devDependencies": {
-    "jest-canvas-mock": "^2.5.1"
```
bruugey, line 35: do we no longer need this as a dev dep?

packages/button/src/Button.story.tsx
bruugey: Would encourage us to adopt these patterns sooner rather than later so we don't need to make larger PRs down the line correcting them. If we know there's a convention that we're working towards, let's use it now.

packages/button/src/ButtonContent/ButtonContent.styles.ts
bruugey, line 1: We should add a transition to button width

packages/loading-indicator/package.json
spark33: No; this PR adds it as a dev dep on our main repo itself since we're moving it to be a global setup job

packages/button/src/Button.story.tsx
spark33: I agree from a process perspective, but it would just add overhead for designers without a Chromatic contingency. Since we're probably going to change a lot of things about our stories to accommodate Chromatic soon anyway, I'd push for an exception here. There are also other bugs I've identified following our migration to Storybook v7 that we'd have to make global bugfixes for anyway.

spark33: I agree from a process perspective, but it would just add overhead for designers without a Chromatic contingency. Since we're probably going to change a lot of things about our stories to accommodate Chromatic soon anyway, I'd push for an exception here. There are also other bugs I've identified following our migration to Storybook v7 that we'd have to make global bugfixes for anyway.

packages/button/src/ButtonContent/ButtonContent.styles.ts
spark33: We will be handling this at a later date

sooachung (approved):

shaneeza (approved):

PR #1771, opened 2023-05-31
LG-3241, LG-3194: controlled toast and notification bar updates

shaneeza:
## ✍️ Proposed changes

[Demo](https://shaneeza.github.io/leafygreen-ui/?path=/story/components-toast--with-initial-toasts)

🎟 _Jira ticket:_ [LG-3241](https://jira.mongodb.org/browse/LG-3241)
🎟 _Jira ticket:_ [LG-3194](https://jira.mongodb.org/browse/LG-3194)

### For new components


packages/toast/src/ControlledToast/ControlledToast.spec.tsx
```diff
@@ -70,6 +70,24 @@ describe('packages/toast/controlled', () => {
       expect(toast).toBeInTheDocument();
     });
 
+    test('does not render when `open` is true and component is unmounted', async () => {
```
TheSonOfThomp, line 73: We should also test that it un-renders when the render condition toggles

TheSonOfThomp (commented):
Just needs the extra test condition

packages/toast/src/ControlledToast/ControlledToast.spec.tsx
```diff
@@ -70,6 +70,58 @@ describe('packages/toast/controlled', () => {
...
+            />
+          )}
+        </ToastProvider>,
+      );
+
+      const toast = await waitFor(() => queryByTestId('test-toast1b'));
+      expect(toast).not.toBeInTheDocument();
+    });
+
+    test('unmounts when `open` is true and component render is toggled to false', async () => {
```
shaneeza, line 91: @TheSonOfThomp I'm not sure if this is the best way to test this. To get it to work i had to change the data-testid for the rerendered Toast so I'm not sure if that is correct. Do you have any suggestions?

packages/toast/src/InternalToast/InternalToast.spec.tsx
```diff
@@ -195,8 +195,9 @@ describe('packages/toast/internal-toast', () => {
 });
 
 // eslint-disable-next-line jest/no-disabled-tests
-describe.skip('TS types', () => {
-  render(
+describe.skip('TS types behave as expected', () => {
```
shaneeza, line 198: Updated this to eliminate some of the warnings in the test console.

packages/toast/src/ControlledToast/ControlledToast.spec.tsx
TheSonOfThomp: I don't think in this case we're checking for the same component. I think after the `rerender` we want to 
```
await waitForElementToBeRemoved(toast)
expect(toast).not.toBeInTheDocument()
```

packages/toast/src/ControlledToast/ControlledToast.spec.tsx
```diff
@@ -88,6 +88,40 @@ describe('packages/toast/controlled', () => {
       expect(toast).not.toBeInTheDocument();
     });
 
+    test('unmounts when `open` is true and component render is toggled to false', async () => {
+      let isOpen = true;
```
TheSonOfThomp, line 92: nit: I'd call this var `shouldRender`, to avoid confusion

TheSonOfThomp (approved):

PR #1772, opened 2023-05-31
LG-3272: Skeleton

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3272](https://jira.mongodb.org/browse/LG-3272)

[Demo](https://spark33.github.io/leafygreen-ui/?path=/story/components-skeletonloader--live-example&args=darkMode:!true)

Live Example:
image

### For new components


packages/skeleton-loader/src/CardSkeleton/CardSkeleton.tsx
```diff
@@ -0,0 +1,30 @@
...
+export function CardSkeleton({
+  darkMode: darkModeProp,
+  className,
+  ...rest
+}: CardSkeletonProps) {
+  const { darkMode } = useDarkMode(darkModeProp);
+  return (
+    <Card
+      aria-busy
+      {...rest}
```
bruugey, line 21: why are we allowing rest to overwrite the aria-busy attribute

bruugey:
Table skeleton the border between the header and the table is not centered

packages/skeleton-loader/README.md
```diff
@@ -0,0 +1,63 @@
...
+<TableSkeleton columnLabels={['Column 1', 'Column 2']} />
+```
+
+## Properties
+
+#### Skeleton
+
+| Name             | Type      | Default | Description                                              |
+| ---------------- | --------- | ------- | -------------------------------------------------------- |
+| `size`           | `boolean` | `false` | Indicates whether the header skeleton should be rendered |
```
shaneeza, line 37: This line doesn't seem correct

packages/skeleton-loader/src/CardSkeleton/CardSkeleton.tsx
```diff
@@ -0,0 +1,30 @@
...
+      darkMode={darkMode}
+      className={cx(rootStyles, className)}
+    >
+      <ParagraphSkeleton withHeader />
```
shaneeza, line 25: Will Cards always show the header?

packages/skeleton-loader/src/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  ...rest
+}: SkeletonProps) {
+  const { theme } = useDarkMode(darkMode);
+  return (
+    <div
+      className={cx(
+        rootStyles,
+        sizeStyles[size],
+        themeStyles[theme],
+        'lg-ui-skeleton',
```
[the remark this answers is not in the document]
shaneeza: could you use `createUniqueClassName`?

packages/skeleton-loader/src/Skeleton/Skeleton.types.ts
```diff
@@ -0,0 +1,17 @@
...
+   * @default Size.Default
+   */
+  size?: Size;
+}
+
+export const Size = {
+  Small: 'small',
+  Default: 'default',
+  Large: 'large',
+};
```
shaneeza, line 15: as const?

packages/skeleton-loader/src/SkeletonLoader.spec.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { render, within } from '@testing-library/react';
+import { axe } from 'jest-axe';
+
+import { CardSkeleton } from './CardSkeleton';
+import { FormSkeleton } from './FormSkeleton';
+import { ParagraphSkeleton } from './ParagraphSkeleton';
+import { Skeleton } from './Skeleton';
+import { TableSkeleton } from './TableSkeleton';
+
+describe('packages/skeleton-loader/Skeleton', () => {
```
shaneeza, line 11: Would it be better to add these tests to each package instead of all to this one file?

packages/skeleton-loader/src/Skeleton/Skeleton.styles.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const themeStyles: Record<Theme, string> = {
+  [Theme.Dark]: css`
+    background: linear-gradient(
+        110deg,
+        ${palette.gray.dark2} 35%,
+        ${palette.gray.dark1},
+        ${palette.gray.dark2} 65%
+      )
+      0 0/ 100vw 100% fixed;
```
shaneeza, line 40: cool

packages/skeleton-loader/src/CardSkeleton/CardSkeleton.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export function CardSkeleton({
+  darkMode: darkModeProp,
+  className,
+  ...rest
+}: CardSkeletonProps) {
+  const { darkMode } = useDarkMode(darkModeProp);
+  return (
+    <Card
+      aria-busy
```
shaneeza, line 20: There are so many conflicting articles about `aria-busy`. This [article I found](https://adrianroselli.com/2020/11/more-accessible-skeletons.html) shows that it should not be on the skeleton loader but rather the component's parent using it. I'm not sure what's correct but maybe we should also add `aria-hidden` to this?

spark33:
> * [ ]  The last column header has a skeleton loader. Was this intentional?
> 
> Screenshot 2023-06-08 at 11 09 08 AM

Yeah this was intentional - I wanted to highlight that having header text for just certain columns was supported.

packages/skeleton-loader/src/CardSkeleton/CardSkeleton.tsx
spark33: That's a bit tricky to achieve without actually baking the Skeleton into each component, since we would want a flip of the aria-busy attribute from false to true on the container element when the content loads. By having `aria-busy` on the component, we would at least make sure something on the screen reader indicates that the page is loading. In a separate component, we can't really know what the appropriate component to pass `aria-busy` to would be.

Maybe Skeletons shouldn't have aria-busy when they're a part of a larger Skeleton component. But this means that the Card in `CardSkeleton` should have aria-busy and the Skeletons inside shouldn't?

packages/skeleton-loader/src/Skeleton/Skeleton.tsx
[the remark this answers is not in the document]
spark33: On second thought, I don't think we even need this. I'll remove it

I initially had it here because I thought every instance of Skeleton would need to share a className (so no createUniqueClassName), but we should be able to achieve the synchronized animation even without it

packages/skeleton-loader/src/CardSkeleton/CardSkeleton.tsx
spark33: Yep!

packages/skeleton-loader/README.md
spark33: Good catch! Thank you

packages/skeleton-loader/src/CardSkeleton/CardSkeleton.tsx
shaneeza: To me, it seems that the `aria-busy` should be left up to the consumer to add and toggle, but i doubt anyone would even use it :(

spark33: But maybe I should add aria-hidden to the basic Skeleton. @bruugey do you know what the best move here would be

packages/skeleton-loader/src/Skeleton/Skeleton.types.ts
```diff
@@ -0,0 +1,17 @@
+import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
+
+export interface SkeletonProps extends DarkModeProps, HTMLElementProps<'div'> {
+  /**
+   * Determines the height of the skeleton
+   * @default Size.Default
```
shaneeza, line 6: should this be the string `default`?

spark33:
> Table skeleton the border between the header and the table is not centered

It looks centered to me by eye and by value on inspect - all cells have a 10px padding and the first row has an additional 3px margin-top on the Skeleton to account for the box-shadow

sooachung (approved):

packages/skeleton-loader/README.md
```diff
@@ -0,0 +1,63 @@
...
+<TableSkeleton columnLabels={['Column 1', 'Column 2']} />
+```
+
+## Properties
+
+#### Skeleton
+
+| Name             | Type      | Default | Description                           |
+| ---------------- | --------- | ------- | ------------------------------------- |
+| `size`           | `boolean` | `false` | Determines the height of the skeleton |
```
shaneeza, line 37: Don't forgot to change the `Type` and `Default` values

spark33: yikes thank you

shaneeza (approved):

PR #1776, opened 2023-06-01
Creates StorybookArgTypes

TheSonOfThomp:
## ✍️ Proposed changes

• Renames `storybookArgTypes` -> `defaultStorybookArgTypes`
      - Refactors all stories to use the new variable name
• Creates enum `StorybookArgTypes` for cleaner argType definition
• Exports new type `StoryType`


packages/badge/src/Badge.story.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { StoryFn } from '@storybook/react';
 
-import { storybookArgTypes, StoryMetaType } from '@leafygreen-ui/lib';
```
spark33, line 4: You know I've pushed for this name change since day 1, so you also know I am a fan 😎

packages/button/src/Button.story.tsx
```diff
@@ -19,36 +22,18 @@ const meta: StoryMetaType<typeof Button> = {
...
-    },
-    href: {
-      control: 'text',
-    },
+    disabled: StorybookArgTypes.Boolean(),
+    leftGlyph: StorybookArgTypes.Select(Object.keys(glyphs)),
+    rightGlyph: StorybookArgTypes.Select(Object.keys(glyphs)),
+    type: StorybookArgTypes.Select(['button', 'submit']),
+    size: StorybookArgTypes.Select(Object.values(Size)),
+    href: StorybookArgTypes.Text(),
```
spark33, line 30: I like this idea a lot - this obviously looks a lot cleaner. My only concern is that because it's another layer of logic to maintain that might be very hard to track down errors in, I'd want these object-transforming utilities to be pretty heavily tested.

packages/lib/src/storybook/StoryMeta.types.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export type StoryMetaType<
+  T extends React.ElementType,
+  XP extends Record<string, any> = {},
+> = Omit<Meta<T>, 'component' | 'argTypes' | 'args'> & {
+  title?: string;
+  component?: T;
+  parameters: StoryParameters<T>;
+  argTypes?: ArgTypes<T, XP>;
+  args?: Partial<ComponentProps<T> | LeafyGreenProviderProps | XP>;
```
spark33, line 83: As I understand, these props should also be defined in `Meta`. Why are they duplicated here?

At the very least, it feels like these should be commented with why the duplication was necessary.

TheSonOfThomp: It's typed as `Record<string, any>`

packages/lib/src/storybook/storybookArgTypes.ts
```diff
@@ -21,61 +21,97 @@ export interface StoryArgType extends InputType {
...
+  } else
+    return (description?: string, config?: Record<string, any>) =>
+      ({
+        description,
+        control: { type, ...config },
+      } as StoryArgType);
+}
+
+export const StorybookArgTypes = {
+  Boolean: argType('boolean'),
```
bruugey, line 60: returning a function? not a type? this verbiage is confusing to me

packages/lib/src/storybook/storybookArgTypes.ts
```diff
@@ -21,61 +21,97 @@ export interface StoryArgType extends InputType {
...
+    'The base font size passed to the LeafyGreenProvider that wraps the component',
+  ),
+  updatedBaseFontSize: StorybookArgTypes.Radio(
+    [13, 16],
+    'The base font size passed to the LeafyGreenProvider that wraps the component. Uses the updated font size values for Euclid Circular A.',
+  ),
+  darkMode: StorybookArgTypes.Boolean('Render the component in dark mode.'),
+  children: StorybookArgTypes.Text('Element rendered inside the component'),
+  as: StorybookArgTypes.Select(
+    IntrinsicElements,
```
bruugey, line 85: we probably want to be a little bit more restrictive here, right now `br` is breaking IconButton, for example

packages/lib/src/storybook/storybookArgTypes.ts
```diff
@@ -21,61 +21,97 @@ export interface StoryArgType extends InputType {
...
-  onDismiss: { control: 'none' },
-  onFilter: { control: 'none' },
-  onSubmit: { control: 'none' },
-  portalClassName: { control: 'none' },
-  portalContainer: { control: 'none' },
-  popoverZIndex: { control: 'none' },
-  ref: { control: 'none' },
-  scrollContainer: { control: 'none' },
-  usePortal: { control: 'none' },
+  'aria-controls': StorybookArgTypes.None,
```
bruugey, line 95: wondering if we want to be more selective with what props we choose to demo, rather than having an exclude list

packages/button/src/Button.story.tsx
shaneeza: I'm with Sean on this

PR #1806, opened 2023-06-15
Adam/tsconfig

TheSonOfThomp:
Updates TSConfig settings

1. Creates package `@lg-tools/config`
2. Extends `@lg-tools/config/package.tsconfig.json` in every package
3. Adds relative paths to each package's tsconfig
4. Updates `build:ts` job to build each package's TS separately (instead of all at once as a single project)

## How to test
Ensure `yarn install` and `yarn build` work as expected, and output `.js` and `.d.ts` files in `dist`


scripts/create-package.js
```diff
@@ -112,13 +112,14 @@ const packageJSON = `
 
 const tsConfig = `
 {
-  "extends": "../../package.tsconfig.json",
+  "extends": "../../tools/config/package.tsconfig.json",
```
bruugey, line 115: what about in cases where this isn't a relative import

TheSonOfThomp: updated

scripts/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "noUnusedLocals": false,
+    "esModuleInterop": true,
+    "strict": true,
+    "allowSyntheticDefaultImports": true,
+    "moduleResolution": "node",
+    "baseUrl": ".",
+    "skipLibCheck": true,
+    "resolveJsonModule": true,
+  }
+}
```
spark33, line 21: Why do we need a tsconfig file in the scripts folder?

spark33 (approved):

scripts/tsconfig.json
TheSonOfThomp: This lets us run `ts-node ./scripts/xyx` without passing any overrides (We previously had a `yarn ts-node` script that added additional options)

This will likely be removed eventually as scripts get removed

TheSonOfThomp: It also makes it clear that this tsconfig applies to the scripts. We could technically use the root one, but that's ambiguous

PR #1807, opened 2023-06-15
Build tooling

TheSonOfThomp:
- Adds Turbo config
- Added separate `tsc` typescript build scripts to each component
  - Turbo uses this script to build TS in parallel
- Moved/updated story files in which there were cyclical dependencies
- Updates CI pipeline


bruugey:
that local build taking 3 seconds is 👩‍🍳 💋

turbo.json
```diff
@@ -0,0 +1,16 @@
...
+  "$schema": "https://turbo.build/schema.json",
+  "pipeline": {
+    "prebuild": {
+      "dependsOn": ["^prebuild"]
+    },
+    "build": {
+      "dependsOn": ["^build"],
+      "outputs": ["dist/**"]
+    },
+    "tsc": {
```
bruugey, line 11: should we add test here?

TheSonOfThomp: I'll add that eventually in a separate PR. Need to mess with the jest config to do that first

bruugey (approved):

PR #1831, opened 2023-07-07
LG-3667: Updates disabled Radio label

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3667](https://jira.mongodb.org/browse/LG-3667)

### For new components

### DIFF
--- a/packages/radio-group/src/Radio/Radio.tsx
+++ b/packages/radio-group/src/Radio/Radio.tsx
@@ -22,6 +22,7 @@ import {
   inputDisplayWrapperClassName,
   inputThemeStyles,
   labelBaseStyle,
+  labelDisabledStyle,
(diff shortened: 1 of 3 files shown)


packages/radio-group/src/Radio/Radio.tsx
```diff
@@ -71,6 +74,7 @@ function Radio({
         className={cx(
           labelBaseStyle,
           {
+            [labelDisabledStyle[theme]]: disabled,
```
shaneeza, line 75: Can you use the disabled prop instead of adding CSS? It looks like the colors are the same.

bruugey: Good catch

shaneeza (approved):

PR #1832, opened 2023-07-07
Adam/react 18 migration fork

TheSonOfThomp:


packages/icon/src/isComponentGlyph.ts
```diff
@@ -2,19 +2,23 @@ import { ComponentType, isValidElement, ReactNode } from 'react';
...
+function isComponentGlyph<P>(
+  node: {} | null | undefined,
+): node is React.ReactElement<P>;
 function isComponentGlyph(node: ReactNode): node is LGGlyph.Element;
 function isComponentGlyph(
   component: ExtendedComponentType,
 ): component is LGGlyph.Component;
 function isComponentGlyph(
-  child: ReactNode | ExtendedComponentType,
+  child: ReactNode | ExtendedComponentType | unknown,
```
shaneeza, line 21: @bruugey im curious why we needed to add this? It created a few ts issues with `isComponentGlyph` in sidenav and combobox

bruugey: there's gotta be a better way to do what i did here, but basically we're passing something with type unknown here in Tooltip - i have a slightly different fix running locally, but there are a couple of different ways we can go about this.

bruugey: either we can add another level of checking in Tooltip before calling this function (maybe preferred) or add stronger typing for supporting an unknown coming in (which is what the validate call is doing from inside the function)

bruugey: Fixed this!

PR #1847, opened 2023-07-14
Remove ts-downlevel

TheSonOfThomp:
All consuming apps use TS>4.0, so we no longer need to downlevel to ts3.4
https://wiki.corp.mongodb.com/display/DESIGN/Application+Requirements


package.json
```diff
@@ -21,8 +21,6 @@
     "fix": "lg-lint --fix",
     "link": "lg-link",
     "lint": "lg-lint",
-    "prepublishOnly": "yarn fix && yarn test && yarn build && yarn prepublish:ts-downlevel",
-    "prepublish:ts-downlevel": "rm -rf packages/*/dist/ts3.4 && echo packages/*/dist | NODE_NO_WARNINGS=1 xargs -I{} -n1 -P8 downlevel-dts {} {}/ts3.4",
     "publish": "yarn changeset publish --public",
```
bruugey, line 24: do we need to explicitly build pre publish?

TheSonOfThomp: Ah, that's a good point

bruugey (approved):

PR #1854, opened 2023-07-24
LG-3314: Prevents disabled buttons from submitting forms

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3314:](https://jira.mongodb.org/browse/LG-3314:)

### For new components


spark33:
Have we actually tested this inside a form element? I thought `onSubmit` handlers were put on the form element, not the submit button, and called from there?

bruugey:
The demo given to me was: https://codesandbox.io/s/eager-water-vrjrfz?file=/src/App.tsx

bruugey:
> Have we actually tested this inside a form element? I thought `onSubmit` handlers were put on the form element, not the submit button, and called from there?

I don't think it's in this component's scope to read from an associated form (although with contextually-aware components we could make that a reality one day) for now I just wanted to solve what was in scope for the ticket. See JIRA for details.

.changeset/few-melons-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/button': minor
+---
+
+Prevents `onSubmit` handler from firing when `disabled="true"` or `isLoading="true"`
```
spark33, line 5: We should update this changelog comment to be more reflective of the new change

spark33 (approved):

.changeset/few-melons-refuse.md
bruugey: Updated, lmk what you think.

PR #1859, opened 2023-07-24
LG-3312: Fix Combobox Props

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3312](https://jira.mongodb.org/browse/LG-3312)

### For new components


packages/combobox/src/Combobox.types.ts
```diff
@@ -132,141 +133,118 @@ export interface ComboboxMultiselectProps<M extends boolean> {
...
-   * When usePortal is `true`, specifies the scrollable element to position relative to.
-   */
-  scrollContainer?: null;
-
-  /**
-   * Number that controls the z-index of the popover element directly.
-   */
-  popoverZIndex?: number;
-}
+export type BaseComboboxProps = Omit<HTMLElementProps<'div'>, 'onChange'> &
```
TheSonOfThomp, line 136: nit: I feel like we should keep this an interface
```ts
interface BaseComboboxProps 
  extends Omit<HTMLElementProps<'div'>, 'onChange'>, PortalControlProps {...}
```

packages/combobox/src/Combobox.types.ts
```diff
@@ -132,141 +133,118 @@ export interface ComboboxMultiselectProps<M extends boolean> {
...
+    /**
+     * Defined the character limit of a multiselect Chip before they start truncating.
+     * Note: the three ellipses dots are included in the character limit.
+     */
+    chipCharacterLimit?: number;
+
+    /**
+     * Number that controls the z-index of the popover element directly.
+     */
+    popoverZIndex?: number;
```
TheSonOfThomp, line 246: is `popoverZIndex` not inherited?

packages/combobox/src/utils/ComboboxTestUtils.tsx
```diff
@@ -199,8 +199,12 @@ export function renderCombobox<T extends Select>(
    * @param newProps
    * @returns
    */
-  const rerenderCombobox = (newProps: renderComboboxProps) =>
-    renderResult.rerender(getComboboxJSX({ ...props, ...newProps }));
+  const rerenderCombobox = (newProps: renderComboboxProps) => {
+    const rerenderProps = { ...props, ...newProps };
+    return renderResult.rerender(
+      getComboboxJSX(rerenderProps as renderComboboxProps),
```
TheSonOfThomp, line 205: We should type the variable not the argument
```ts
const rerenderProps: renderComboboxProps = { ...props, ...newProps };
```

TheSonOfThomp (approved):

packages/combobox/src/Combobox.types.ts
bruugey: Not from PopoverControls

packages/combobox/src/Combobox.types.ts
bruugey: PortalControlProps isn't statically known so we get a TS error

packages/combobox/src/utils/ComboboxTestUtils.tsx
bruugey: we have to cast it, TS still raises an error when we type the variable

TheSonOfThomp: Hmm, I still think asserting at assignment is technically better
```ts
const rerenderProps = {
    ...props,
    ...newProps,
} as renderComboboxProps;
```

PR #1861, opened 2023-07-25
LG-3313: Pass all Row props to V11Adapter rows

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3313](https://jira.mongodb.org/browse/LG-3313)

### For new components


bruugey (changes requested):
This only solves for row. I believe we need to add this to all elements.

packages/table/src/V11Adapter/V11Adapter.story.tsx
```diff
@@ -52,8 +52,8 @@ export const Basic: StoryFn<StoryAdapterProps> = () => {
         }
       >
         {({ datum }: any) => (
-          <Row>
-            <Cell>{datum.name}</Cell>
+          <Row data-test="test-row">
```
bruugey, line 55: confirmed this works, we can clean up storybook.

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -51,8 +52,12 @@ const V11Adapter = <T extends LGRowData>({
   const containerRef = useRef(null);
   const OldTable = flattenChildren(children)[0];
 
+  consoleOnce.warn(
```
bruugey, line 55: are there things we should be explicitly filtering out?

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -142,9 +169,15 @@ const V11Adapter = <T extends LGRowData>({
                       // index by row.index (not the index of the loop) to get the sorted order
                       // @ts-expect-error `processedData` is structured to be indexable by `row.index`
                       processedData[row.index]?.[cell.column.id]?.();
+                    const {
+                      children,
+                      isHeader,
```
bruugey, line 174: Are `isHeader` and `isDisabled` the only props that need filtering out?

spark33: image

bruugey (commented):
we need a changeset here

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -47,17 +48,17 @@ const V11Adapter = <T extends LGRowData>({
...
   const OldTable = flattenChildren(children)[0];
 
   if (!isComponentType(OldTable, 'Table')) {
-    console.error(
+    consoleOnce.error(
       'The first and only child of `Table.V11Adapter` must be a `V10Table` component',
     );
   }
 
   const OldTableProps = (OldTable as ReactElement).props;
```
TheSonOfThomp, line 60: Doesn't `isComponentType` assert that the variable is the right type?

TheSonOfThomp (approved):
LGTM. Tests seems to verify the behavior we want

bruugey (approved):

PR #1873, opened 2023-07-26
Fix Lint task

TheSonOfThomp:
## ✍️ Proposed changes

- Lint task now exits with code `1` if lint errors are found
- Fix lint errors


bruugey:
Should we be bumping everything to 1.0.0 so that there's no confusion on NPM

PR #1875, opened 2023-07-27
Slackbot:release

TheSonOfThomp:
## ✍️ Proposed changes

- Adds `lg slackbot release` to main CLI
- Adds tests to slackbot
- Support for other repos & scopes


tools/slackbot/src/release-bot/release-bot.types.ts
```diff
@@ -4,7 +4,9 @@ export const Channels = {
   'leafygreen-ui-releases': 'C01CBLPFD35',
 } as const;
 
-export interface Opts {
+export type ChangeType = 'major' | 'minor' | 'patch' | 'dependency';
```
shaneeza, line 7: Is the `dependency` type new?

tools/slackbot/src/release-bot/utils/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/* UTILS */
+
+export function exists<T extends string | Array<any>>(
+  arg: T | undefined,
+): arg is T {
+  return !isUndefined(arg) && arg.length > 0;
+}
+
+export function isValidJSON(str?: string): boolean {
```
shaneeza, line 11: Is there a way to share this? It looks like you're also using a few lines above.

shaneeza (approved):

tools/slackbot/src/release-bot/release-bot.types.ts
TheSonOfThomp: It's just a special designation we have we want to format them differently. `dependency` change type is just a patch with nothing other than dependency bumps

tools/slackbot/src/release-bot/utils/utils.ts
TheSonOfThomp: @shaneeza What do you mean share? Like add it to lib/meta?

shaneeza: Yeah, or another appropriate place. Cause right now, you're writing this function in both `tools/slackbot` and `tools/meta`

TheSonOfThomp: 👍

PR #1887, opened 2023-08-01
TS & ESlint configs

TheSonOfThomp:
## ✍️ Proposed changes

Updates TS & ESLint configs so VSCode can validate & auto-import properly

## 🧪 How to test changes

• Intentionally break a lint rule. VSCode should show a red squiggle
• Start typing a local import name & accept auto-import suggestion. Should import from `./` instead of `src/`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,2 @@
-// For VSCode extensions
-const eslint = require('@lg-tools/lint/config/eslint.config');
-module.exports = eslint;
+const config = require('@lg-tools/lint/config/eslint.config');
+module.exports = { ...config };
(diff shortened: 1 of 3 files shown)


bruugey (commented):
Do we need changesets for this?

TheSonOfThomp:
> Do we need changesets for this?

Yeah I guess we should have one for `/build`. Will do 👍

bruugey (commented):
Should we be able to run `tsc` from the root now?

bruugey (commented):
let's link and test in either chat or private

TheSonOfThomp:
> Should we be able to run `tsc` from the root now?

That's not what this fixes. This mainly ensures VSCode imports & highlights properly.

We won't be able to run tsc from root without using turbo, since each package has its own separate config

PR #1911, opened 2023-08-09
LG-3497: Thumb Icons Dark Mode bug

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3497](https://jira.mongodb.org/browse/LG-3497)

### For new components


packages/icon/README.md
```diff
@@ -149,7 +149,7 @@ A `LGGlyph.Component`
 
 (For DS engineers)
 
-- After uploading a new SVG file to `src/glyphs`, replace all `fill` property values with `#000`.
+- After uploading a new SVG file to `src/glyphs`, replace all `fill` property values for strokes with `#000`.
```
TheSonOfThomp, line 152: Is this true? Shouldn't it still be "replace all `fill` values...", but ensure that the root `svg` element does not have a full?

TheSonOfThomp (commented):
Thanks for tackling this!  I don't think the Readme is quite right...

PR #1916, opened 2023-08-09
LG-3313: v10 Row props

spark33:
## ✍️ Proposed changes

### For new components


packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -127,6 +152,8 @@ const V11Adapter = <T extends LGRowData>({
               virtualRow={
                 useVirtualScrolling ? (iterable as VirtualItem) : undefined
               }
+              // @ts-expect-error rowProps is an additional prop passed by the `processData` function
```
bruugey, line 155: why do we need to ts-expect-error here?

TheSonOfThomp: We should fix the types in the `processData` function rather than add expect error

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -100,12 +119,18 @@ const V11Adapter = <T extends LGRowData>({
...
-          {table.getHeaderGroups()[0].headers.map(header => {
+        <HeaderRow {...oldHeaderRow.props}>
+          {table.getHeaderGroups()[0].headers.map((header, i) => {
             return (
-              <HeaderCell key={header.id} header={header}>
+              // @ts-expect-error
+              <HeaderCell
+                key={header.id}
+                header={header}
+                {...oldHeaderCellProps[i]}
```
TheSonOfThomp, line 132: We should update/cast the type of `oldHeaderCellProps` rather than expect error

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -158,18 +191,29 @@ const V11Adapter = <T extends LGRowData>({
...
-                        <Cell key={subRowCell.id}>
-                          {/* @ts-expect-error subRow.original returns the object in the user's defined shape, and should be string indexable */}
-                          {subRow.original[subRowCell.column.id]()}
-                        </Cell>
-                      );
-                    })}
-                  </Row>
-                ))}
+                row.subRows.map(subRow => {
+                  // @ts-expect-error rowProps is an additional prop passed by the `processData` function
```
TheSonOfThomp, line 195: Same here

packages/table/src/V11Adapter/V11Adapter.tsx
```diff
@@ -158,18 +191,29 @@ const V11Adapter = <T extends LGRowData>({
...
+                  const { children, ...subRowProps } = subRow.original.rowProps;
+                  return (
+                    <Row key={subRow.id} row={subRow} {...subRowProps}>
+                      {subRow.getVisibleCells().map(srCell => {
+                        /* @ts-expect-error subRow.original returns the object in the user's defined shape, and should be string indexable */
+                        const subRowCell = subRow.original[srCell.column.id]();
```
TheSonOfThomp, line 201: Seems there are a lot of issues with the TS type of `row.original` vs how it's used.  We should fix those

TheSonOfThomp (commented):
Functionally LGTM. TS types should be updated though

packages/table/src/V11Adapter/V11Adapter.tsx
bruugey: agreeed

TheSonOfThomp (approved):

packages/table/src/V11Adapter/V11Adapter.tsx
bruugey: agree that we should fix htis.

PR #1928, opened 2023-08-10
Adds `bold` prop to RadioGroup

bruugey:
## ✍️ Proposed changes

### For new components


packages/radio-group/src/RadioGroup.story.tsx
```diff
@@ -25,22 +25,11 @@ const meta: StoryMetaType<typeof RadioGroup> = {
...
-        </Radio>
-        <Radio default value="2" description="This is a description">
-          Radio Input 2
-        </Radio>
-        <Radio disabled value="Selection-4">
-          Disabled Option
-        </Radio>
-      </>
-    ),
+    children: <></>,
```
bruugey, line 28: This didn't work previously, because RadioGroup depends on logic where we do not recursively iterate through children, we only check children at the first level.

bruugey, line 28: This didn't work previously, because RadioGroup depends on logic where we do not recursively iterate through children, we only check children at the first level.

spark33 (approved):

packages/radio-group/src/Radio/Radio.tsx
```diff
@@ -75,6 +77,9 @@ function Radio({
             [css`
               font-size: 12px;
             `]: size === Size.XSmall, // TODO: keeping this style until XS is deprecated
+            [css`
```
spark33, line 80: nit: feels more legible to say the bold font-weight should be applied when the `bold === true` than to negate

TheSonOfThomp: On one hand I agree, but this way  implies that `bold === true` is the default

packages/radio-group/src/Radio/Radio.tsx
```diff
@@ -55,6 +56,7 @@ function Radio({
   checked,
   size = Size.Default,
   description,
+  bold,
```
TheSonOfThomp, line 59: We should add the `bold` prop to the single `Radio` story

TheSonOfThomp (changes requested):

packages/radio-group/src/Radio/Radio.tsx
bruugey: Added

bruugey:
@TheSonOfThomp stories are broken bc the components dont properly respond to darkmode/size we have an open ticket to address this (out of scope for this work though)

TheSonOfThomp:
> @TheSonOfThomp stories are broken bc the components dont properly respond to darkmode/size we have an open ticket to address this (out of scope for this work though)

None of the generated stories are showing up though. This is a regression 

Screenshot 2023-08-24 at 3 36 33 PM

packages/radio-group/src/RadioGroup.story.tsx
bruugey: Fixed 'em

bruugey:
Think I'm back to parity @TheSonOfThomp

TheSonOfThomp (approved):

PR #1931, opened 2023-08-10
Explicitly build ts docs

bruugey:
## ✍️ Proposed changes

### For new components


.github/workflows/release.yml
```diff
@@ -34,10 +34,13 @@ jobs:
         if: ${{ steps.build-cache.outputs.cache-hit != 'true' }}
         run: yarn install --frozen-lockfile --prefer-offline
 
-      - name: build
+      - name: Build
         if: ${{ steps.build-cache.outputs.cache-hit != 'true' }}
         run: yarn build
 
+      - name: Generate docs
+        run: turbo run docs
```
TheSonOfThomp, line 42: We're caching the `dist` directory (L20), so this file won't get picked up in the cache restore. 
We should add `tsdoc.json` and `stories.js` to the cache path on L20, and L106

TheSonOfThomp (changes requested):

.github/workflows/release.yml
bruugey: I'm confused bc this says rely on the build step

bruugey: So we add it to the build step

bruugey: And run it without condition

bruugey: Should we not check for a build cache in the build step when releasing, if we know we want the build to look different when releasing than when it's just a normal PR

TheSonOfThomp (approved):

PR #1937, opened 2023-08-15
LG-3508: Add internal props to SearchInput

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3508](https://jira.mongodb.org/browse/LG-3508)

### For new components

### DIFF
--- a/.changeset/spotty-cobras-dream.md
+++ b/.changeset/spotty-cobras-dream.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/search-input': patch
+---
+
+Add internal props for temporary reuse of dropdown behavior in other components
--- a/packages/search-input/src/SearchInput.story.tsx
+++ b/packages/search-input/src/SearchInput.story.tsx
@@ -31,7 +31,16 @@ const meta: StoryMetaType<typeof SearchInput> = {
   parameters: {
     default: 'LiveExample',
     controls: {
-      exclude: [...storybookExcludedControlParams, 'value', 'id', 'showWedge'],
+      exclude: [
+        ...storybookExcludedControlParams,
+        'value',
+        'id',
+        'showWedge',
+        '__INTERNAL__divWrapperProps__',
+        '__INTERNAL__divWrapperSlot__',
+        '__INTERNAL__inputProps__',
+        '__INTERNAL__inputSlot__',
+      ],
     },
     generate: {
       combineArgs: {
@@ -179,7 +188,7 @@ const data = [
 export const LiveExample: StoryFn<SearchInputProps> = (
   args: SearchInputProps,
 ) => {
-  const [currentPage, setPage] = useState<typeof data[0]>();
+  const [currentPage, setPage] = useState<(typeof data)[0]>();
   const [searchResults, setSearchResults] = useState(data);
 
   const handleChange: ChangeEventHandler<HTMLInputElement> = e => {
--- a/packages/search-input/src/SearchInput.story.tsx
+++ b/packages/search-input/src/SearchInput.story.tsx
@@ -1,5 +1,10 @@
 /* eslint-disable no-console */
-import React, { ChangeEventHandler, FormEventHandler, useState } from 'react';
+import React, {
+  ChangeEventHandler,
+  FormEventHandler,
+  forwardRef,
+  useState,
+} from 'react';
 import { StoryFn } from '@storybook/react';
 import { kebabCase, startCase } from 'lodash';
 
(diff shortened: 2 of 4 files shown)


packages/search-input/src/SearchInput/SearchInput.types.ts
```diff
@@ -62,6 +62,33 @@ interface BaseSearchInputProps
    * The placeholder text shown in the input field before the user begins typing.
    */
   placeholder?: string;
+
+  /**
+   * For internal uses, expose a slot to render a custom div to wrap the input
+   * @internal
+   */
+  __INTERNAL__divWrapperSlot__?: React.ForwardRefExoticComponent<
```
bruugey, line 70: I disagree with this approach - I think there should be one slot prop that replaces the entire container.

bruugey (commented):
We need tests for this too

packages/search-input/src/SearchInput/SearchInput.types.ts
spark33: This pattern is maintaining parity with Select. The difference is that unlike Select where only the Button needs to be overridden, this component requires props on both the div and input to be overridden.

bruugey: WE only have ___INTERNAL___MenuButtonSlot - i think we should maintain the same idea here.

bruugey: We should have one line of code that says internal ? <Internal {...receivedProps} /> : <SearchInput />

spark33: I don't think that would work since a lot of the functionality is dependent on both `searchBoxRef` and `inputRef`, unless I'm missing something

spark33: We need to be able to pass in both a custom searchBox and a custom input. If we completely replace the existing searchBox, the input will be completely replaced as well. The other option is that if we replace it completely, we'd be passing in a `searchBoxRef` AND an `inputRef`. I think using two slot props would be better than passing two refs to the override component.

bruugey: you can't have `<Internal searchBoxRef={forwardToSearchBox} inputRef={forwardToInput} />` ?

bruugey: @shaneeza what do you think?

shaneeza: is it easy to forward multiple refs?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -401,16 +411,20 @@ export const SearchInput = React.forwardRef<HTMLInputElement, SearchInputProps>(
               )}
               aria-label={ariaLabel}
               aria-labelledby={ariaLabelledBy}
+              {...__INTERNAL__divWrapperProps__}
```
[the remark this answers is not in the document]
spark33: Yes, and I think it would need to since they'd be rendering something completely different

packages/search-input/src/SearchInput/SearchInput.types.ts
spark33: I don't believe it's easy. It would be kind of hacky from what I've seen

bruugey: ok will defer to you two

.changeset/spotty-cobras-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/search-input': patch
+---
+
+Add internal props for temporary reuse of dropdown behavior in other components
```
shaneeza, line 5: what do you mean by temp?

packages/search-input/src/SearchInput/SearchInput.types.ts
shaneeza, line 70: we should remove these from the story

packages/search-input/src/SearchInput.story.tsx
```diff
@@ -302,4 +307,71 @@ WithResults.parameters = {
...
+
+export const CustomElementsWithResults: StoryFn<SearchInputProps> = (
+  props: SearchInputProps,
+) => (
+  <SearchInput
+    className={css`
+      width: 200px;
+    `}
+    onChange={() => {
+      console.log('SB: Change');
```
shaneeza, line 337: `/* eslint-disable-next-line no-console */`?

packages/search-input/src/SearchInput/SearchInput.tsx
```diff
@@ -245,6 +251,9 @@ export const SearchInput = React.forwardRef<HTMLInputElement, SearchInputProps>(
 
     const handleInputChange: ChangeEventHandler<HTMLInputElement> = e => {
       handleChange?.(e);
+      __INTERNAL__inputProps__ &&
+        __INTERNAL__inputProps__.onChange &&
+        __INTERNAL__inputProps__.onChange(e);
```
shaneeza, line 256: if you have `__INTERNAL__inputProps__.onChange` would `handleInputChange` ever be called since `__INTERNAL__inputProps__.onChange` overrides `onChange` on the input?

shaneeza:
Run `yarn build`, the builds are passing even though you have some errors

.changeset/spotty-cobras-dream.md
spark33: We're planning to create a dropdown component to abstract this behavior in the long run, so we're going to be replacing this pattern pretty soon from what I understand @bruugey

bruugey: Would omit. Let's just say: An internal prop for reuse of dropdown/search behavior in other components.

PR #1942, opened 2023-08-22
LG-3330: password input controlled value

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3330](https://jira.mongodb.org/browse/LG-3330)

### For new components


packages/password-input/src/PasswordInput.story.tsx
shaneeza, line 1: revert these changes

.changeset/late-ants-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/hooks': patch
```
shaneeza, line 2: Should be major since, im changing what is returned

packages/hooks/src/useControlledValue/useControlledValue.spec.ts
```diff
@@ -16,7 +16,7 @@ describe('packages/lib/useControlledValue', () => {
 
     act(() => {
       current.handleChange({ target: { value: 'banana' } } as ChangeEvent<any>);
-      current.setInternalValue('banana');
+      current.setUncontrolledValue('banana');
```
TheSonOfThomp, line 19: We should add more tests to this. 
The current one should be named e.g. "`handleChange` calls provided callback".

We should create another test where we update the value passed into the hook and expect that the current.value changes

TheSonOfThomp (approved):

PR #1944, opened 2023-08-22
Any Decision Records (ADRs)

TheSonOfThomp:
## ✍️ Proposed changes

Initial Any Decision Record (ADR) proposal


docs/decisions/0000-use-adrs.md
```diff
@@ -0,0 +1,48 @@
...
+- [Google Docs](#google-docs)
+-
+
+## Weighing the Options
+
+### Markdown ADRs
+
+- ✅ Decisions remain in context with code
+- ✅ Easier to reference history of decisions
+- ✅ Can solicit feedback via GitHub PR review
```
bruugey, line 38: would much rather comment here than in a google doc, personally.

PR #1951, opened 2023-08-24
Integration: Date Picker [LG-3152]

TheSonOfThomp:
## ✍️ Proposed changes

Creates the Date Picker component

RFC Document: https://docs.google.com/document/d/1HussHpUyylDvQFNlwadxGj8A8ThpO3Uzj_D6HoYY7LQ/

🎟 _Jira ticket:_ https://jira.mongodb.org/browse/LG-3152

### For bug fixes, new features & breaking changes

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] I have run `yarn changeset` and documented my changes

### For new components

- [ ] I have added my new package to the global tsconfig
- [ ] I have added my new package to the Table of Contents on the global README
- [ ] I have verified the Live Example looks as intended on the design website.

PR #1953, opened 2023-08-24
useControlledValue updates [LG-3608]

TheSonOfThomp:
## ✍️ Proposed changes

- Extends `useControlledValue` to accept any type.
  - Adds `initialValue` argument. Used for setting the initial value for uncontrolled components. Without this we may encounter a React error for switching between controlled/uncontrolled inputs
  - Changes signature of `onChange` argument to be a simple setter function (`(value: T) => void`)
  - Changes return object to include only `isControlled`, `value` and an `updateValue` setter function

## 🧪 How to test changes

See spec file


.changeset/soft-berries-obey.md
```diff
@@ -0,0 +1,8 @@
+---
+'@leafygreen-ui/hooks': minor
+---
+
+- Extends `useControlledValue` to accept any type.
+- Adds `updateValue` function in return value. This method triggers a synthetic event to update the value of a controlled or uncontrolled component.
```
shaneeza, line 6: Im a little confused by this. Can you explain what this is solving?

TheSonOfThomp: If a component is using this hook, there's no easy way to programmatically change the value.

Consider the component below. If we want to update the component's value (or trigger a change in the consuming component) on button click, we would need to call `handleChange` and pass in a synthetic change event. The `updateValue` function makes it cleaner to do this.

``` tsx
    const TestComponent = ({ valueProp, handlerProp }) => {
      const inputRef = useRef<HTMLInputElement>(null);
      const { value, handleChange, updateValue } = useControlledValue(
        valueProp,
        handlerProp,
      );

      return (
        <>
          <input
            ref={inputRef}
            value={value}
            onChange={handleChange}
          />
          <button
            onClick={() => updateValue('carrot', inputRef)}
          />
        </>
      );
    };
```

Note: `updateValue` takes an element ref, since we still need to simulate a change event under the hood, and we need to know what element should be the trigger

TheSonOfThomp: One update I'm considering making is to pass the element ref directly into the hook, so the call would be something like

```tsx
const inputRef = useRef<HTMLInputElement>(null);
const { value, handleChange, updateValue } = useControlledValue(
  valueProp,
  handlerProp,
  inputRef
);

...
<button
  onClick={() => updateValue('carrot')}
/>
```

shaneeza: Thanks for explaining! I like the idea of passing the ref directly to the hook.

shaneeza: But there could be confusion as to why `updateValue` won't work if `inputRef` is optional and not passed to the hook.

packages/hooks/src/useControlledValue/useControlledValue.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+      test('user interaction triggers handler', () => {
+        const handler = jest.fn();
+        const result = render(<TestComponent handlerProp={handler} />);
+        const input = result.getByTestId('test-input');
+        userEvent.type(input, 'b');
+        expect(handler).toHaveBeenCalled();
+      });
+
+      test('user interaction does not change the element value', () => {
```
shaneeza, line 344: is this test name correct? user interaction changed the value to banana?

packages/hooks/src/useControlledValue/useControlledValue.ts
```diff
@@ -42,10 +64,36 @@ export const useControlledValue = <T extends string>(
...
+          bubbles: true,
+        }),
+        ref.current,
+      );
+
+      handleChange(synthEvt);
+    }
+  };
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah, I realized this is not actually what we want to be doing, since there are plenty of times we don't want an initialValue. Removing this check

packages/hooks/src/useControlledValue/useControlledValue.spec.tsx
```diff
@@ -0,0 +1,288 @@
...
+  describe('Controlled', () => {
+    test('rendering with a value sets value and isControlled', () => {
+      const {
+        result: { current },
+      } = renderUseControlledValueHook('apple');
+      expect(current.isControlled).toBe(true);
+      expect(current.value).toBe('apple');
+    });
+
+    test('rerendering from initial undefined sets value and isControlled', async () => {
```
TheSonOfThomp, line 101: Do we want this case to log an error?

TheSonOfThomp:
Closing for now and will reopen after #1951 is finished

PR #1960, opened 2023-08-25
Error message font size

TheSonOfThomp:
## ✍️ Proposed changes

Updates `Error` component to read from TypographyProvider to set font size

### DIFF
--- a/packages/typography/src/Error/Error.tsx
+++ b/packages/typography/src/Error/Error.tsx
@@ -8,7 +8,13 @@ import {
   usePolymorphic,
 } from '@leafygreen-ui/polymorphic';
 
-import { errorMessageModeStyle, errorMessageStyle } from './Error.styles';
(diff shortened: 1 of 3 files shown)


packages/typography/src/Error/Error.tsx
```diff
@@ -20,6 +26,7 @@ const Error = Polymorphic<BaseErrorProps>(
     ...rest
   }) => {
```
bruugey, line 27: what was this defaulting to before? and are we sure this is what all the designs call for? i feel like i remember error having its own typography rules but i could be wrong

TheSonOfThomp: It was defaulting to `inherit` before, so 16px unless a parent set a font-size

bruugey:
I'd have more FUD around merging this, but it looks like it doesn't causes breakages in any of our other components (per Chromatic)

bruugey (approved):

PR #1962, opened 2023-08-28
LG-3544: Link inline prop

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3544](https://jira.mongodb.org/browse/LG-3544)

### For new components

### DIFF
--- a/packages/typography/README.md
+++ b/packages/typography/README.md
@@ -176,6 +176,7 @@ All props extend the HTMLElementProps of their root tag, however the below compo
 | `hideExternalIcon` | `boolean`                        | Hides the external icon when the current host name is different from the host of the destination URL                                                                                     | `false`                                       |
 | `darkMode`         | `boolean`                        | Determines if the component renders in dark mode                                                                                                                                         | `false`                                       |
 | `as`               | `keyof JSX.IntrinsicElements`    | Element to render the component as. **Note**: This will supersede the behavior of any other props.                                                                                       | `'a'`                                         |
+| `inline`           | `boolean`                        | Determines if the Link should be inherit font size from the parent                                                                                                                       | `false`                                       |
 | `baseFontSize`     | `13` \| `16`                     | font-size applied to typography element                                                                                                                                                  | Defaults to value set by LeafyGreen Provider. |
 
 # Label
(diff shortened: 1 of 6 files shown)


packages/typography/src/Link/Link.styles.ts
```diff
@@ -94,3 +94,8 @@ export const openInNewTabStyles = css`
   left: -1px;
   height: 12px;
 `;
+
+export const inlineStyles = css`
+  font-size: inherit;
+  line-height: inherit;
```
bruugey, line 100: should this have an explicitly display: inline property as well?

bruugey: `inheritSize`?

packages/typography/README.md
```diff
@@ -176,6 +176,7 @@ All props extend the HTMLElementProps of their root tag, however the below compo
 | `hideExternalIcon` | `boolean`                        | Hides the external icon when the current host name is different from the host of the destination URL                                                                                     | `false`                                       |
 | `darkMode`         | `boolean`                        | Determines if the component renders in dark mode                                                                                                                                         | `false`                                       |
 | `as`               | `keyof JSX.IntrinsicElements`    | Element to render the component as. **Note**: This will supersede the behavior of any other props.                                                                                       | `'a'`                                         |
+| `inline`           | `boolean`                        | Determines if the Link should be inherit font size from the parent                                                                                                                       | `false`                                       |
```
TheSonOfThomp, line 179: This feels like it's going to apply `display: inline`. Prop should be something like `inheritFontSize`

packages/typography/src/Link/Link.styles.ts
```diff
@@ -94,3 +94,8 @@ export const openInNewTabStyles = css`
   left: -1px;
   height: 12px;
 `;
+
+export const inheritSizeStyles = css`
+  font-size: inherit;
+  line-height: inherit;
```
TheSonOfThomp, line 100: I still feel like this should be the default behavior. A link should always be the size of whatever it's inside.
If someone wants to use a Link outside of something that has a font-size set, they can always either provide a `baseFontSize` prop, or add custom CSS

TheSonOfThomp (changes requested):
The prop name definitely needs to change.

I also think this should be the default behavior, not toggled with a prop

PR #1963, opened 2023-08-28
LG-3544: Add inherit styles to Link

spark33:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/packages/typography/src/Link/Link.styles.ts
+++ b/packages/typography/src/Link/Link.styles.ts
@@ -20,6 +20,8 @@ export const linkStyles = css`
   text-decoration: none;
   cursor: pointer;
   line-height: 13px;
+  font-size: inherit;
+  line-height: inherit;
 
   &:focus {
     outline: none;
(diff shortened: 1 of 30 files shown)


packages/typography/src/Link/Link.styles.ts
```diff
@@ -20,6 +20,8 @@ export const linkStyles = css`
   text-decoration: none;
   cursor: pointer;
   line-height: 13px;
```
bruugey, line 22: rm this line

packages/typography/src/Link/Link.tsx
```diff
@@ -112,8 +110,9 @@ const Link = InferredPolymorphic<BaseLinkProps, 'span'>(
         className={cx(
           anchorClassName,
           overwriteDefaultStyles,
-          bodyTypeScaleStyles[baseFontSize],
           linkStyles,
+          // @ts-expect-error baseFontSize will not be used an index when undefined
```
bruugey, line 114: nit: we could have a function above to generate the class name and write out the if statement so we don't need to add an expect error

bruugey:
Per chromatic it looks like we might have been using the contextually-set sizes to determine the icon styling - take a look

bruugey (approved):

PR #1978, opened 2023-09-11
LG-3363: Table Checkboxes color change (CSS Approach)

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3363](https://jira.mongodb.org/browse/LG-3363)

### For new components


packages/table/src/Row/Row.tsx
```diff
@@ -6,20 +6,31 @@ import { LGRowData } from '../useLeafyGreenTable';
...
-  } else {
-    return <InternalRowWithoutRT {...rest} />;
-  }
+  return (
+    <RowContextProvider disabled={disabled}>
+      {row ? (
+        <InternalRowWithRT
+          row={row}
+          virtualRow={virtualRow}
+          disabled={disabled}
```
shaneeza, line 26: Do you need to pass disabled if you can get it from the provider?

packages/table/src/Row/Row.stories.tsx
```diff
@@ -228,4 +238,136 @@ DisabledClickableRows.args = {
   disabled: true,
 };
 
+export const DisabledSelectableRows: StoryFn<
```
shaneeza, line 241: I think this is an unrelated issue, but should you still be able to select/unselect all when the rows are disabled? https://642700aa3e49e32bdbf0b0fc-jevezbjkag.chromatic.com/?path=/story/components-table-row--disabled-selectable-rows&args=disabled:!true

packages/table/src/useLeafyGreenTable/TableCheckbox.tsx
```diff
@@ -0,0 +1,58 @@
...
+  table,
+}: {
+  table: Table<LGTableDataType<LGRowData>>;
+}) => {
+  const { theme } = useDarkMode();
+  const { disabled: rowIsDisabled } = useRowContext();
+  return (
+    <Checkbox
+      className={cx({
+        [disabledTableRowCheckStyles[theme]]: rowIsDisabled,
```
shaneeza, line 23: nit: i don't think you even need to conditionally render these styles with `rowIsDisabled` since you're targeting `[aria-checked='true']`.

packages/table/src/useLeafyGreenTable/TableCheckbox.tsx
```diff
@@ -0,0 +1,58 @@
...
+import Checkbox from '@leafygreen-ui/checkbox';
+import { cx } from '@leafygreen-ui/emotion';
+import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+
+import { useRowContext } from '../Row/RowContext';
+
+import { disabledTableRowCheckStyles } from './useLeafyGreenTable.styles';
+import { LGRowData, LGTableDataType } from '.';
+
+export const TableHeaderCheckbox = ({
```
shaneeza, line 13: Thoughts on making these two separate components?

packages/table/src/Row/Row.stories.tsx
spark33: Created LG-3612 to explore this!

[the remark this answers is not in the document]
spark33: Yeah this is most likely a Storybook issue. This is the same issue that arises when `darkMode` is toggled on larger components.

packages/table/src/Row/Row.tsx
spark33: I guess we don't? But it would have to be consumed at every subcomponent of Row then, which feels pretty verbose. The intention here is to introduce the context so it can be consumed from subcomponents defined outside (like Checkbox), not from within Row (since the prop already handles that). If you feel like removing the props from the subcomponents would feel cleaner, I'll make the change!

shaneeza: It feels like we now have two ways to pass disabled. The name of the provider, `RowContextProvider` itself makes it seem like it should be consumed inside of `Row`. I feel strongly that we should consume it from the provider since it exists now.

spark33: Roger that!

shaneeza (approved):

PR #1979, opened 2023-09-12
LG-3183, LG-3578: Chip

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3183](https://jira.mongodb.org/browse/LG-3183)
🎟 _Jira ticket:_ [LG-3578](https://jira.mongodb.org/browse/LG-3578)

chip-design

### For new components


packages/number-input/README.md
shaneeza, line 1: the table was a little wonky so i fixed it

.changeset/hip-waves-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/combobox': major
```
shaneeza, line 2: the focus states did change but maybe this is only just a patch?

packages/chip/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,132 @@
...
+  chipWrapperDisabledStyle,
+  chipWrapperSizeStyle,
+  chipWrapperThemeStyle,
+} from './Chip.styles';
+import { ChipProps, Size, TruncationLocation, Variant } from './Chip.types';
+
+export const Chip = React.forwardRef<HTMLSpanElement, ChipProps>(
+  (
+    {
+      chipTruncationLocation = TruncationLocation.none,
```
bruugey, line 29: `TruncationLocation.None` (caps N)

packages/chip/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,132 @@
...
+      popoverZIndex,
+      className,
+      dismissButtonAriaLabel,
+      ...rest
+    }: ChipProps,
+    forwardedRef,
+  ) => {
+    const { darkMode, theme } = useDarkMode(darkModeProp);
+
+    const isTruncated =
```
bruugey, line 46: these first two `!!` will never not be true bc you are providing default values. We can definitely clean up this logic.

packages/chip/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,132 @@
...
+      !!chipCharacterLimit &&
+      !!chipTruncationLocation &&
+      chipTruncationLocation !== 'none' &&
+      typeof label === 'string' &&
+      !!label &&
+      (label as string).length > chipCharacterLimit;
+
+    const truncatedName = useMemo(
```
bruugey, line 54: should only need to get this if `isTruncated` is true

packages/chip/src/Chip/Chip.tsx
bruugey: I actually think I also disagree with the below. I don't think we should block on chipTruncationLocation, I think setting a chipCharacterLImit should be enough. We should have a fallback chipTruncationLocation, but double confirm this with design

packages/chip/README.md
```diff
@@ -0,0 +1,59 @@
...
+| ------------------------ | ------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------- |
+| `label`                  | `React.ReactNode`                                                                    | Label rendered in the chip                                                                                                                                                                                                                                |                     |
+| `chipTruncationLocation` | `'end'` \| `'middle'` \| `'none'` \| `'start'`                                       | Defines where the ellipses will appear in a Chip when the label length exceeds the `chipCharacterLimit`. If `none` is passed, the chip will not truncate. **Note**: If there is any truncation, the full label text will appear inside a tooltip on hover | `none`              |
```
bruugey, line 50: I don't know if none should be an option. i think we should set a default and if chipCharacterLimit is true, then we truncate. I don't like a boolean being based on two props, when one should be enough.

packages/chip/src/Chip/Chip.tsx
shaneeza: whoops

packages/chip/src/Chip/Chip.tsx
shaneeza: I think that makes sense. Technically right now, you can either set `none` or set the limit to `0` to prevent truncation. We shouldn't have two ways to prevent truncation.

shaneeza: @alvenw did ask for the default to have no truncation. We can set the default `chipCharacterLImit` to `0` and make the default `chipTruncationLocation` `end`. 

This would be a major change for `Combobox` since the API would be different.

packages/chip/src/Chip/Chip.tsx
shaneeza: A few lines below (L94) there is a check for `isTruncated`. Would this still run with that check?

packages/chip/src/utils/getTruncatedName.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @returns `string`
+ * @internal
+ */
+export const getTruncatedName = (
+  chipCharacterLimit: number,
+  chipTruncationLocation: TruncationLocation,
+  label: string,
+) => {
+  const ellipsis = '…';
+  const chars = chipCharacterLimit - 3; // ellipsis dots included in the char limit
```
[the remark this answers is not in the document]
shaneeza: `ellipsis.length` is `1`. I think it's cause we're using the HTML character, not just three dots. 
Screenshot 2023-09-13 at 4 06 16 PM

packages/chip/src/Chip/Chip.tsx
bruugey: Ah - interesting design decision on the Combobox side. Would get sign off from whoever wrote that API in the first place to ensure they weren't thinking through an edge case I'm missing. If not, I'm happy to set defaults that mimic the current behavior, but still modify the API. Alternatively, we can leave things as is in Combobox, but make the change here in the standalone Chip.

shaneeza: @TheSonOfThomp Is there an edge case we're not seeing?

packages/chip/src/utils/getTruncatedName.ts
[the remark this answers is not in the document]
bruugey: oooooo interesting. leave as is!

packages/chip/src/utils/getTruncatedName.ts
```diff
@@ -0,0 +1,42 @@
+import { TruncationLocation } from '../Chip';
```
shaneeza, line 1: Add test for this

packages/chip/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,132 @@
...
+      className,
+      dismissButtonAriaLabel,
+      ...rest
+    }: ChipProps,
+    forwardedRef,
+  ) => {
+    const { darkMode, theme } = useDarkMode(darkModeProp);
+
+    const isTruncated =
+      !!chipCharacterLimit &&
```
shaneeza, line 47: Fix this so that 0 still truncates

packages/chip/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,132 @@
...
+  chipWrapperSizeStyle,
+  chipWrapperThemeStyle,
+} from './Chip.styles';
+import { ChipProps, Size, TruncationLocation, Variant } from './Chip.types';
+
+export const Chip = React.forwardRef<HTMLSpanElement, ChipProps>(
+  (
+    {
+      chipTruncationLocation = TruncationLocation.None,
+      chipCharacterLimit = 15,
```
shaneeza, line 30: set to undefined

packages/combobox/src/Chip/Chip.tsx
shaneeza, line 1: export this as combobox chip and remove chip alias

packages/chip/src/Chip/Chip.tsx
shaneeza: We chatted about this in person but when the limit is 0 the chip should still truncate, so we need to update the logic to check for `undefined` not `!!`. That means that we would also keep the four locations.

packages/chip/src/Chip.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+const meta: StoryMetaType<typeof Chip> = {
+  title: 'Components/Chip',
+  component: Chip,
+  parameters: {
+    default: 'LiveExample',
+    generate: {
+      storyNames: ['Gray', 'Green', 'Blue', 'Red', 'Purple', 'Yellow'],
+      combineArgs: {
+        darkMode: [false, true],
+        label: ['Chip', 'meow meow meow miaou miao miau'],
```
spark33, line 19: 😂

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -0,0 +1,205 @@
...
+ * Padding on a chip (in px)
+ */
+export const chipWrapperPadding: Record<Size, { x: number; y: number }> = {
+  [Size.Default]: {
+    y: 0,
+    x: 6,
+  },
+  [Size.Large]: {
+    y: 2,
+    x: 6,
```
spark33, line 39: Or here for that matter!

packages/chip/src/Chip/Chip.types.ts
```diff
@@ -0,0 +1,82 @@
...
+export const TruncationLocation = {
+  Start: 'start',
+  Middle: 'middle',
+  End: 'end',
+  None: 'none',
+} as const;
+export type TruncationLocation =
+  (typeof TruncationLocation)[keyof typeof TruncationLocation];
+
+export const Size = {
```
bruugey, line 12: After speaking to Rob, we actually think that Chips should have a `baseFontSize` prop rather than have a `size` prop, since that is the fork not actually the size. We want every Chip in Atlas, for example, to use the 13px Chip rather than the 16px.

shaneeza: i see, ok, that should be easy to change.

packages/chip/src/Chip/Chip.spec.tsx
```diff
@@ -0,0 +1,328 @@
...
+      const { queryByTestId } = renderChip({ disabled: true, onDismiss });
+      const button = queryByTestId('chip-dismiss-button');
+      expect(button?.getAttribute('aria-disabled')).toBe('true');
+    });
+
+    test('does nothing when disabled button is clicked', () => {
+      const onDismiss = jest.fn();
+      const { queryByTestId } = renderChip({ disabled: true, onDismiss });
+      const button = queryByTestId('chip-dismiss-button');
+      expect(() => userEvent.click(button!)).toThrow();
```
spark33, line 244: Curious - what does this do and how does it work?

packages/chip/src/Chip/Chip.spec.tsx
```diff
@@ -0,0 +1,328 @@
...
+    });
+  });
+
+  describe('dismiss button with truncated text', () => {
+    test('renders when onDismiss is passed', () => {
+      const onDismiss = jest.fn();
+      const { queryByTestId } = renderChip({ onDismiss, label: longLabel });
+      const button = queryByTestId('chip-dismiss-button');
+      expect(button).toBeInTheDocument();
+    });
```
spark33, line 254: nit - the description in `describe` doesn't accurately describe this test case

packages/chip/src/Chip/Chip.spec.tsx
```diff
@@ -0,0 +1,328 @@
...
+    test('renders with updated aria-label when dismissButtonAriaLabel is passed', () => {
+      const onDismiss = jest.fn();
+      const { queryByTestId } = renderChip({
+        label: longLabel,
+        onDismiss,
+        dismissButtonAriaLabel: 'new aria label',
+      });
+      const button = queryByTestId('chip-dismiss-button');
+      expect(button?.getAttribute('aria-label')).toBe('new aria label');
+    });
```
spark33, line 282: Looks like the above comment is relevant for the test cases up to here

packages/chip/src/Chip/Chip.spec.tsx
shaneeza: After upgrading to `user-event v13`, you could no longer, in tests, click on buttons with `pointer-events: none`. So instead of clicking on a button you have to test that clicking on the button will throw. However, there are other ways around [this](https://testing-library.com/docs/ecosystem-user-event/#clickelement-eventinit-options)

packages/chip/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Returns true if chipCharacterLimit is defined, the chipTruncationLocation isn't `none`, and the label is a valid string with a length that is greater than the chipCharacterLimit.
+     * @returns `boolean`
+     */
+    const isTruncated =
+      Number.isInteger(chipCharacterLimit) &&
+      chipTruncationLocation !== 'none' &&
+      typeof label === 'string' &&
+      !!label &&
```
spark33, line 55: Is this necessary? `label` is a required prop, right?

packages/chip/src/utils/getTruncatedName/getTruncatedName.spec.tsx
```diff
@@ -0,0 +1,197 @@
...
+    });
+
+    test('and the chipTruncationLocation is `Middle`', () => {
+      const results = getTruncatedName(
+        4,
+        TruncationLocation.Middle,
+        stringToTruncate,
+      );
+      expect(results).toEqual('…y');
+    });
```
spark33, line 110: Looks like a conscious decision to prefer 'End' only one end of 'Middle' is visible. Is this worth noting down? I guess people would just realize when they use it and won't really care?

shaneeza: I don't think we need to go too into the weeds of how middle should work seeing as this test case would never even be an issue inside Chip since we have this check: `(label as string).length > (chipCharacterLimit as number);`. 

Middle is trying to halve the word and add in the ellipsis based on the limit but in this test case the limit is 4 and the label is only 3 letters so the bulk of the truncated word will be the 3 dots and the one left over character.

packages/chip/src/Chip/Chip.spec.tsx
shaneeza: thanks, looks like im missing the truncated limit and location

packages/combobox/src/Combobox/Combobox.styles.ts
```diff
@@ -261,7 +270,7 @@ export const inputWrapperStyle = ({
         gap: ${flexGap}px;
         overflow-x: hidden;
         min-height: ${inputHeight(size)}px;
-        max-height: calc((${getChipHeight(size) * 3}px) + (${flexGap}px * 2));
+        max-height: calc((${inputHeight(size) * 3}px) + (${flexGap}px * 2));
```
spark33, line 273: Not from your PR, but it'd be nice to know what the `3` is for

packages/combobox/src/ComboboxChip/ComboboxChip.styles.ts
```diff
@@ -0,0 +1,77 @@
...
+  [ComboboxSize.Large]: css`
+    font-size: ${fontSize[ComboboxSize.Large]}px;
+    line-height: ${lineHeight[ComboboxSize.Large]}px;
+
+    .${chipTextClassName} {
+      padding-inline-end: 10px;
+      padding-block: ${chipWrapperPaddingY[ComboboxSize.Large]}px;
+    }
+  `,
+};
```
spark33, line 77: nit - This looks like it could be a function that's much more legible and shorter

shaneeza: I thought about this but since `large` has an extra line of CSS that the others don't, I decided against it.

packages/combobox/src/Combobox/Combobox.styles.ts
shaneeza: 😅

alvenw (approved):

packages/chip/src/DismissButton/DismissButton.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+  title: 'Components/Chip/DismissButton',
+  component: DismissButton,
+  parameters: {
+    default: null,
+    generate: {
+      combineArgs: {
+        darkMode: [false, true],
+        label: ['Chip'],
+        disabled: [false, true],
+        variant: [
```
bruugey, line 20: nit but `Object.values(Variant)`

PR #1986, opened 2023-09-18
LG-3592: SearchResultsMenu popoverProps; `footerSlot` prop

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3592](https://jira.mongodb.org/browse/LG-3592)

### For new components

### DIFF
--- a/packages/search-input/src/SearchResultsMenu/SearchResultsMenu.tsx
+++ b/packages/search-input/src/SearchResultsMenu/SearchResultsMenu.tsx
@@ -36,7 +36,7 @@ export const SearchResultsMenu = React.forwardRef<
       portalClassName,
       portalContainer,
       scrollContainer,
-      dropdownFooterSlot,
+      footerSlot,
       ...rest
     }: SearchResultsMenuProps,
     ref,
@@ -98,7 +98,7 @@ export const SearchResultsMenu = React.forwardRef<
             >
               {React.Children.count(children) ? children : <EmptyOption />}
             </ul>
-            {dropdownFooterSlot}
+            {footerSlot}
           </>
         )}
       </Popover>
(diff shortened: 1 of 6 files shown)


shaneeza (approved):

packages/search-input/src/SearchResultsMenu/SearchResultsMenu.tsx
```diff
@@ -58,6 +58,7 @@ export const SearchResultsMenu = React.forwardRef<
         `,
       )}
       refEl={refEl}
+      {...rest}
```
bruugey, line 61: I don't understand how this is different than what was previously here.

bruugey: oh nvm

packages/search-input/src/SearchResultsMenu/SearchResultsMenu.tsx
```diff
@@ -26,7 +26,7 @@ const MAX_MENU_HEIGHT = 256;
 export const SearchResultsMenu = React.forwardRef<
   HTMLUListElement,
   SearchResultsMenuProps
->(({ children, open = false, refEl }: SearchResultsMenuProps, ref) => {
+>(({ children, open = false, refEl, ...rest }: SearchResultsMenuProps, ref) => {
```
bruugey, line 29: how do we know that rest is all popover props? what is rest / should we just pull out the props that we are trying to pass to popover explicitly since those are known

spark33: The type was extending `PortalControlProps` which is also extended by `PopoverProps`. I did notice this also extends `HTMLElementProps<'ul'>`, so I took it out and made it explicit

PR #1989, opened 2023-09-19
useDynamicRefs tests

TheSonOfThomp:
## ✍️ Proposed changes

Adds tests to useDynamicRefs


packages/hooks/src/useDynamicRefs/useDynamicRefs.spec.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const getter1 = result.current;
+    rerender();
+    expect(result.current).toBe(getter1);
+  });
+
+  test('returns identical getters when called with the same prefix', () => {
+    const { result: A } = renderHook(() => useDynamicRefs({ prefix: 'A' }));
+    const { result: B } = renderHook(() => useDynamicRefs({ prefix: 'A' }));
+    expect(A.current).toBe(B.current);
+  });
```
TheSonOfThomp, line 26: Should we add a warning for this case? 
Or enforce uniqueness of `prefix` (and therefore getters and refs)?

Imagine the following:
```jsx
// Select component
const optionRefs = useDynamicRefs({ prefix: 'option' })
...
<div ref={optionRefs('apple')} />

// Combobox component
const optionRefs = useDynamicRefs({ prefix: 'option' })
...
<div ref={optionRefs('apple')} />
```

If both components happen to be rendered at the same time, both components could be sharing a ref object

spark33 (approved):

PR #1998, opened 2023-09-29
Creates `FormField` component

TheSonOfThomp:
## ✍️ Proposed changes

### For new components


packages/form-field/src/FormField/FormField.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <div data-testid="input" />
+      </FormField>,
+    );
+    const input = getByTestId('input');
+    expect(input).toHaveAttribute('id');
+    expect(input).toHaveAttribute('aria-labelledby');
+    expect(input).toHaveAttribute('aria-describedby');
+  });
+
+  test.todo('Error message appears when state changes');
```
bruugey, line 68: let's write these?

packages/form-field/src/FormField/FormField.types.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface FormFieldProps
+  extends Omit<HTMLElementProps<'div'>, 'children'> {
+  children: FormFieldChildren;
+  label?: React.ReactNode;
+  description?: React.ReactNode;
+  state?: FormFieldState;
+  size?: Size;
+  disabled?: boolean;
+  notificationMessage?: React.ReactNode;
```
bruugey, line 29: not errorMessage?

packages/form-field/src/FormField/index.ts
```diff
@@ -0,0 +1,3 @@
+
```
bruugey, line 1: nit rm

packages/form-field/src/FormField/FormField.tsx
```diff
@@ -0,0 +1,121 @@
...
+              className: cx(
+                iconClassName,
+                iconStyles[theme],
+                icon.props.className,
+              ),
+              disabled,
+            })}
+        </div>
+        {state === FormFieldState.Error && (
+          <Error id={errorId}>{notificationMessage}</Error>
```
bruugey, line 114: this should def be errorMessage since it's only showing up when the state is State.Error

TheSonOfThomp: Password shows a message when the state is "valid". Should these be separate strings?

bruugey: yes bc we're only surfacing this when "error" that might be an edge case that's not captured in the initial API

packages/form-field/src/FormField.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+) => (
+  <FormField
+    {...props}
+    inputWrapperProps={{
+      className: css`
+        padding-inline-end: 4px;
+      `,
+      role: 'combobox',
+      tabIndex: -1,
+    }}
```
TheSonOfThomp, line 80: How do we feel about this method of passing props into the wrapper?

TheSonOfThomp: Alternative would be to export the internal `FormFieldInputWrapper` and do require something like:
```tsx
  <FormField
    {...props}
    icon={
      <IconButton aria-label="Icon">
        <Icon glyph={props.glyph} />
      </IconButton>
    }
  >
    <FormFieldInputWrapper
      className={css`
        padding-inline-end: 4px;
      `}
      role='combobox'
      tabIndex={-1}
    >
      <input placeholder="placeholder" />
    </FormFieldInputWrapper>
  </FormField>
```

bruugey:
A couple of notes from inspecting storybook:
- In the `disabled` state the background of the placeholder text is off
- Can we add a darkmode knob in storybook?
- Can we test with placeholder in generated?

packages/form-field/src/FormField.stories.tsx
TheSonOfThomp: The `FormFieldInputWrapper ` version?

audroue:
- There should be no disabled state that includes our error pattern/icon + red text (a disabled input cannot have an error on it)
image

- Add 4px margin/padding to error message so it has the same spacing as text above input
image

- Any reason we're not including the green/success states? 

- There's another edge-case for Combobox where you can have 3 icons — not sure if in scope. (Also means second icon is not always red for error.)
image

- Separate question — are we not capturing 16px + 18px font size variants with this work? Or the different style text within the input?

TheSonOfThomp:
> * There should be no disabled state that includes our error pattern/icon + red text (a disabled input cannot have an error on it)

Removed from the Generated stories. It's more trouble than it's worth IMO to restrict this in code, so it's technically possible, we'll just advise against it

> * Add 4px margin/padding to error message so it has the same spacing as text above input

👍 Done

> * Any reason we're not including the green/success, or focus states?

This is mainly support work for DatePicker, which doesn't have the `Valid` state.

> * There's another edge-case for Combobox where you can have 3 icons — not sure if in scope. (Also means second icon is not always red for error.)

Added the Story `CustomTwoButtons`. This is done by passing two icons a wrapper to the `icon` prop. The spacing of these will need to be handled on a case-by-case basis

> * Separate question — are we not capturing 16px + 18px font size variants with this work? Or the different style text within the input?

Large size should be accurate now. Let me know if it's not right

audroue:
Looking good! What's the difference between state: "unset" vs state: "valid"? They look the same.

packages/form-field/src/FormFieldInput/FormFieldInput.types.ts
```diff
@@ -0,0 +1,8 @@
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+
+import { FormFieldChildren } from '../FormField/FormField.types';
+
+export interface FormFieldInputProps extends HTMLElementProps<'div'> {
```
bruugey, line 5: Should extend HTMLElementProps<'input'>

TheSonOfThomp: No since it's the wrapper around the input (this bit)
Screenshot 2023-10-05 at 1 26 49 PM

TheSonOfThomp:
> Looking good! What's the difference between state: "unset" vs state: "valid"? They look the same.

Yeah "valid" hasn't been built yet. I've deemed it out of scope (since this is mainly DatePicker support work)

I can remove it from SB to avoid confusion

packages/form-field/src/FormField/FormField.types.ts
```diff
@@ -0,0 +1,49 @@
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+import { Size } from '@leafygreen-ui/tokens';
+
+export const FormFieldState = {
+  Unset: 'unset',
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah, I agree. Though in [SearchInput](https://github.com/mongodb/leafygreen-ui/blob/main/packages/search-input/src/SearchInput/SearchInput.types.ts#L10-L13) we have `unset` and `loading` states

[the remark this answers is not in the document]
bruugey: We can add loading later, and probably internally convert unset to none in that component.

packages/form-field/src/FormFieldInput/FormFieldInput.types.ts
bruugey: I think it should be one of those two though, otherwise it's a bit confusing and I'd expect the component to return an input itself

bruugey: the only other option would be to omit the component entirely - render a div that handles these styles 
```
<div className={wrapperStyles}>{children}</div>
```

and then expose a prop to pass the style overrides there (which we discussed previously, and were in favor of the more explicit)

packages/form-field/src/FormField/FormField.types.ts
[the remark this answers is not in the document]
TheSonOfThomp: Here's the Combobox RFC discussion: https://docs.google.com/document/d/1oV_tY_rptHtwr910hhbMRAevXZdkNohSJnLfXyW2up4/edit?disco=AAAAQKue4WU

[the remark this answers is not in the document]
TheSonOfThomp: 👍 Updated to `default`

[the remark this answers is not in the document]
TheSonOfThomp: Oh wait, `none` or `default`? Figma generally uses "Default"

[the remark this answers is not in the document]
bruugey: `None` is what we use in TextInput, TextArea

packages/form-field/src/FormFieldInput/FormFieldInput.types.ts
TheSonOfThomp: I _think_ I prefer `...Container` vs `...Wrapper`. Thoughts?

packages/form-field/src/FormField/FormField.types.ts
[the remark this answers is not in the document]
bruugey: Same with Select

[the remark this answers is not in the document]
bruugey: Same with SearchInput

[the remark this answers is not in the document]
bruugey: Let's use `'none'`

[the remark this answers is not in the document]
TheSonOfThomp: 👍 Will update

bruugey:
Needs support for a `baseFontSize` prop as well

bruugey:
Should export variable Size as well as State

TheSonOfThomp:
> Should export variable Size as well as State

We're exporting State as `FormFieldState`.
Size is getting pulled in from `tokens.Size`

bruugey:
Ah great, aligned!

audroue:
Carets are not aligned vertically for number input: 
https://jmp.sh/fwdEjyEs

Input should not expand horizontally on hover:
https://jmp.sh/n6h1RYRv

Probably out of scope but there should be no focus disabled state on input dropdown menu (we don't show tooltips on this disabled element):
https://jmp.sh/0x8PT6tj

I think there's a 16px "default" size variant that's missing:
image

Right input icon spacing seems tight compared to Figma (Figma pic shown below):
image

I'm seeing a weird bug where the focus state doesn't appear on the hover state but when you hover out of the input it does:
https://jmp.sh/R7BGxjF5

We don't have a disabled optional input in Figma but we can add it. Tbh not sure a disabled input should show it though since it doesn't provide any value when disabled and adds extra noise or potential confusion to the input. 
image


Should this gray cloud icon not be an icon button? 
image

packages/form-field/src/FormFieldContext/FormFieldContext.tsx
```diff
@@ -10,6 +10,7 @@ export interface FormFieldContextProps {
   size: Size;
   state: FormFieldState;
   inputProps?: FormFieldInputElementProps;
+  optional?: boolean;
```
TheSonOfThomp, line 13: Do we want to make this a prop, or leave it as `contentEnd`?

bruugey: Talked offline and I think we're on the same page

bruugey (approved):

PR #1999, opened 2023-10-03
LG-3054: InfoSprinkle ✨

shaneeza:
## ✍️ Proposed changes

Screenshot 2023-10-03 at 4 57 09 PM

🎟 _Jira ticket:_ [LG-3054](https://jira.mongodb.org/browse/LG-3054)

### For new components


packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
```diff
@@ -0,0 +1,61 @@
...
+        baseFontSize={baseFontSize}
+        trigger={
+          <span
+            // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+            tabIndex={0}
+            ref={forwardRef}
+            className={cx(iconBaseStyles, iconThemeStyles(theme))}
+            data-testid="info-sprinkle-icon"
+          >
+            <Icon glyph="InfoWithCircle" size={baseFontSize} aria-hidden />
```
bruugey, line 35: nit: should we import the InfoWithCircle icon directly, rather than importing the entire Icon component?

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+import { cx } from '@leafygreen-ui/emotion';
+import Icon from '@leafygreen-ui/icon';
+import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+import Tooltip from '@leafygreen-ui/tooltip';
+
+import { iconBaseStyles, iconThemeStyles } from './InfoSprinkle.styles';
+import { Align, InfoSprinkleProps, Justify } from './InfoSprinkle.types';
+
+export const InfoSprinkle = React.forwardRef<HTMLDivElement, InfoSprinkleProps>(
```
bruugey, line 12: forwardRef should accept `HTMLSpanElement` right? Since we're spreading the ref on a `<span />`

yarn.lock
```diff
@@ -2258,6 +2258,15 @@
   resolved "https://registry.yarnpkg.com/@juggle/resize-observer/-/resize-observer-3.4.0.tgz#08d6c5e20cf7e4cc02fd181c4b0c225cd31dbb60"
   integrity sha512-dfLbk+PwWvFzSxwk3n5ySL0hfBog779o8h68wK/7/APo/7cgyWp5jcXockbxdk5kFRkbeXWm4Fbi9FrdN381sA==
 
+"@leafygreen-ui/lib@^10.4.0":
```
bruugey, line 2261: if you update to the correct version of lib, this should go away.

packages/info-sprinkle/README.md
```diff
@@ -0,0 +1,20 @@
...
+```shell
+yarn add @leafygreen-ui/info-sprinkle
+```
+
+### NPM
+
+```shell
+npm install @leafygreen-ui/info-sprinkle
+```
+
```
bruugey, line 20: nit: add docs

tools/create/src/templates/component/package.json.ts
```diff
@@ -25,8 +25,8 @@ export const pkgJson = ({
     "access": "public"
   },
   "dependencies": {
-    "@leafygreen-ui/emotion": "^4.0.4",
-    "@leafygreen-ui/lib": "^10.4.0"
+    "@leafygreen-ui/emotion": "^4.0.7",
+    "@leafygreen-ui/lib": "^12.0.0"
```
bruugey, line 29: can we automate this somehow @TheSonOfThomp

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.types.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export { Align, Justify };
+
+type ModifiedTooltipProps = Omit<
+  TooltipProps,
+  'onClick' | 'trigger' | 'triggerEvent' | 'refEl' | 'children'
+>;
+
+export interface InfoSprinkleProps extends ModifiedTooltipProps {
+  children: string;
```
shaneeza, line 11: add doc

alvenw (approved):

.changeset/four-falcons-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-tools/create': patch
+---
+
+Update `lib` and `emotion` versions
```
bruugey, line 5: Thank you!

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const InfoSprinkle = React.forwardRef<
+  HTMLSpanElement,
+  InfoSprinkleProps
+>(
+  (
+    {
+      darkMode: darkModeProp,
+      children,
+      baseFontSize,
```
bruugey, line 20: did we want to add a max-width prop?

shaneeza: I don't think we need a prop; what would it do? I can add a max-width of 400px in the component.

shaneeza: if the consumer wants to change the max width they can use the `className` prop to change it with CSS

bruugey:
Last question - did we want to bake in the spacing? Or expose as a prop?

tools/create/src/templates/component/package.json.ts
TheSonOfThomp: Yes, just need to write the script. It'll involve pinging npm when running `lg create`

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
```diff
@@ -0,0 +1,64 @@
...
+        baseFontSize={baseFontSize}
+        trigger={
+          <span
+            // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+            tabIndex={0}
+            ref={forwardRef}
+            className={cx(iconBaseStyles, iconThemeStyles(theme))}
+            data-testid="info-sprinkle-icon"
+          >
+            <InfoWithCircleIcon size={baseFontSize} aria-hidden />
```
TheSonOfThomp, line 38: I can foresee folks writing in to ask if they can use other icons. What's our answer?

TheSonOfThomp: Also, @alvenw I don't see anything in the guidelines about not using other icons. We can pre-empt developer requests by making sure designers don't use it incorrectly

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.types.ts
```diff
@@ -0,0 +1,15 @@
...
+type ModifiedTooltipProps = Omit<
+  TooltipProps,
+  'onClick' | 'trigger' | 'triggerEvent' | 'refEl' | 'children'
+>;
+
+export interface InfoSprinkleProps extends ModifiedTooltipProps {
+  /**
+   * String rendered inside the tooltip
+   */
+  children: string;
```
TheSonOfThomp, line 14: Why are we restricting `children` to be a string? Can we add the reasoning to the docs

packages/info-sprinkle/src/InfoSprinkle.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+        </div>
+      ),
+    },
+  },
+  args: {
+    children: "I'm LeafyGreen",
+    baseFontSize: BaseFontSize.Body1,
+    darkMode: false,
+    justify: Justify.Start,
+    align: Align.Bottom,
```
TheSonOfThomp, line 50: What's the default behavior of Tooltip? We should set these default args to those default values (looking at Align.Bottom in particular)

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+        const results = await axe(container);
+        expect(results).toHaveNoViolations();
+      });
+    });
+  });
+
+  describe('info sprinkle', () => {
+    test('renders icon when tooltip is closed', async () => {
+      renderInfoSprinkle();
+      const circleIcon = screen.getByLabelText('Info With Circle Icon');
```
TheSonOfThomp, line 56: 👍 Nice I like howe this is inherently testing that it's the correct icon too

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
```diff
@@ -0,0 +1,64 @@
...
+  ) => {
+    const { darkMode, theme } = useDarkMode(darkModeProp);
+    return (
+      <Tooltip
+        darkMode={darkMode}
+        baseFontSize={baseFontSize}
+        trigger={
+          <span
+            // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+            tabIndex={0}
```
TheSonOfThomp, line 33: I believe this should have a `role` attribute (I think `role="button"`)

tools/create/src/templates/component/package.json.ts
bruugey: https://jira.mongodb.org/browse/LG-3663

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
TheSonOfThomp: Should also have `aria-describedby=<tooltip-id>` (unless tooltip does the described by automatically)

TheSonOfThomp (commented):
Just 2 concerns:
1. `aria` attributes on the trigger
2. Additional documentation on the `children` prop

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
shaneeza: tooltip adds `aria-describedby`.

shaneeza: Would having `role=button` imply that it's clickable?

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.types.ts
shaneeza: Based on the guidelines, only text is allowed but I'm actually not sure if there can be multiple lines/paragraphs. @alvenw Can there be multiple lines/paragraphs? 

If that's the case then i can switch this to a reactnode.

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
TheSonOfThomp: Re: zoom sync, we'll use `role="button"` and `aria-disabled` as done here https://cauldron.dequelabs.com/components/TooltipTabstop#help-icon

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.types.ts
TheSonOfThomp: I think I agree that this should just be a string (otherwise we open it up to rendering Links)

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.tsx
alvenw: Our way of enforcing designers to only use the Info icon is that we made the Figma component with only Info icon without the ability to swap it for other icons. We can also add that to the guidelines though but it would be good to make sure developers aren't able to freely swap the icon

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.types.ts
alvenw: Yep the tooltip content should be short and concise. Anything beyond a couple lines of text would warrant having it be surfaced elsewhere in the UI

PR #2007, opened 2023-10-05
Update button size type

TheSonOfThomp:
## ✍️ Proposed changes

Assigns Button Size as const to improve intellisense
https://mongodb.slack.com/archives/CFFA5BS79/p1696528129126939


bruugey (approved):
Needs a changeset but approving

PR #2014, opened 2023-10-10
LG-3672: Memoize Table columns

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3672](https://jira.mongodb.org/browse/LG-3672)

### For new components


packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
```diff
@@ -34,21 +35,24 @@ function useLeafyGreenTable<T extends LGRowData>({
   allowSelectAll = true,
   ...rest
 }: LeafyGreenTableOptions<T>): LeafyGreenTable<T> {
-  let hasSortableColumns = false;
+  const hasSortableColumns = useRef(false);
```
shaneeza, line 38: why did you have to change this to use a ref?

TheSonOfThomp: Looks like we're setting `hasSortableColums` inside the `columns` setter function. If this were a plain `let` variable it would be re-declared on every render, and would then be out of sync with the `columns` object

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
```diff
@@ -34,21 +35,24 @@ function useLeafyGreenTable<T extends LGRowData>({
...
-        align: propColumn.align ?? 'left',
-        enableSorting: propColumn.enableSorting ?? false,
-      } as LGColumnDef<T>;
-    }),
-  ];
+  const columns = React.useMemo<Array<LGColumnDef<T>>>(
+    () => [
+      ...(hasSelectableRows ? [selectColumnConfig as LGColumnDef<T>] : []),
+      ...columnsProp.map(propColumn => {
+        hasSortableColumns.current = propColumn.enableSorting ?? false;
```
TheSonOfThomp, line 46: Thanks to Shaneeza for calling this out.
I don't love having this side-effect within the columns assignment. I think I'd rather see this in a separate assignment.
```ts
const hasSortableColumns = useMemo(() => columnsProp.some(propCol => !!propCol.enableSorting), [...])
```

TheSonOfThomp (changes requested):

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
spark33: Yeah that's fair code-style-wise. I didn't do that since that'd be an additional loop on a hook that's potentially run fairly often, but maybe that's fine since the # of columns should be small.

TheSonOfThomp (approved):

PR #2019, opened 2023-10-11
update `Space` value in keyMap

shaneeza:
## ✍️ Proposed changes

I added tests to components that were using `keyMap.Space`

### For new components


packages/select/src/Select/Select.spec.tsx
```diff
@@ -673,6 +673,37 @@ describe('packages/select', () => {
           expect(button).toHaveValue(optionValue);
         });
 
+        test('by space key', async () => {
```
shaneeza, line 676: Coping the test directly above this

packages/select/src/Select/Select.spec.tsx
```diff
@@ -673,6 +673,37 @@ describe('packages/select', () => {
...
+          });
+
+          const targetOption = getByTextFor(listbox, optionText).closest('li');
+          expect(targetOption).not.toBe(null);
+
+          act(() => targetOption!.focus());
+
+          fireEvent.keyDown(targetOption!, {
+            key: keyMap.Space,
+          });
```
TheSonOfThomp, line 692: Should be using `userEvent` here. This is testing our implementation of `keyMap`, not the behavior of Select

TheSonOfThomp: (should change above as well)

PR #2025, opened 2023-10-12
Select forward ref LG-3681

TheSonOfThomp:
## ✍️ Proposed changes

Allows Select to accept a ref

🎟 _Jira ticket:_ https://jira.mongodb.org/browse/LG-3681


packages/select/src/Select/Select.tsx
```diff
@@ -43,563 +52,576 @@ import { DropdownWidthBasis, SelectProps, Size, State } from './Select.types';
...
-  const onClose = useCallback(() => {
-    setOpen(false);
-    menuButtonRef.current!.focus();
-  }, [menuButtonRef]);
-
-  useEffect(() => {
-    if (!open) {
-      return;
-    }
+    const containerRef = useForwardedRef(fwdRef, null);
```
bruugey, line 108: is this the right place to forward teh ref?

TheSonOfThomp: technically yes, should be on the container. I'd _prefer_ it to be on the button element itself, but then there's no way to  reference the Label/Description

.changeset/loud-trainers-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/select': minor
```
bruugey, line 2: major change. per react docs wrapping something in fwd ref should be a breaking change

TheSonOfThomp: Really? I believe you but can you link the docs?

bruugey (approved):

PR #2034, opened 2023-10-18
Toast tests

TheSonOfThomp:
## ✍️ Proposed changes

Fixes Toast timeout tests


packages/toast/src/Toast.types.ts
```diff
@@ -11,7 +13,8 @@ type Variant = (typeof Variant)[keyof typeof Variant];
 export { Variant };
 
 export interface ToastProps
-  extends Omit<React.ComponentPropsWithoutRef<'div'>, 'title'> {
+  extends Omit<React.ComponentPropsWithoutRef<'div'>, 'title'>,
+    DarkModeProps {
```
shaneeza, line 17: we never had a darkMode type?

shaneeza (approved):

packages/toast/src/Toast.types.ts
TheSonOfThomp: We never did anything with it. It was passed in through `...rest`, but it was applied directly to the DOM, triggering runtime error logs

PR #2038, opened 2023-10-20
LG-3192, 3633, 3642, 3643: Button style fixes for disabled and loading states

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3192](https://jira.mongodb.org/browse/LG-3192)
🎟 _Jira ticket:_ [LG-3633](https://jira.mongodb.org/browse/LG-3633)
🎟 _Jira ticket:_ [LG-3642](https://jira.mongodb.org/browse/LG-3642)
🎟 _Jira ticket:_ [LG-3643](https://jira.mongodb.org/browse/LG-3643)

### For new components


bruugey (commented):
this could be a good candidate for leveraging data-hovered, data-focused states so that we can test these styles in storybook. what do you think @TheSonOfThomp

yarn.lock
bruugey, line 1: this is probably an accident - we should run yarn clean/rebuild

.changeset/strong-bobcats-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/button': patch
+---
+
+- Buttons now correctly disable click events when disabled or isLoading
```
bruugey, line 5: Can we add a test for this please?

spark33: There are already tests to make sure the `onClick` event isn't fired - the updates here are to make sure it doesn't LOOK like an event is fired. I'll update the changeset so it's accurate

bruugey (commented):
looks like you accidentally committed yarn lock changes that are causing builds to fail

bruugey (commented):
Can we please add a test that demonstrates that when the button is focused clicking the space key does not fire the onClick callback?

bruugey:
Also, the loading story on storybook does not work so I can't test that we've fixed the ripple problem with loading buttons

shaneeza:
> Screenshot 2023-11-01 at 11 38 50 AM
> 
> Disabled buttons can still receive focus.

I believe this is correct, when tabbing onto a disabled button the focus styles should show up however, clicking, with a mouse, on a disabled button should not show the focus state

bruugey (approved):

PR #2040, opened 2023-10-20
LG-3285: Align Tabs `inlineChildren` to bottom

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3285](https://jira.mongodb.org/browse/LG-3285)

### For new components


packages/tabs/src/Tabs/Tabs.styles.ts
```diff
@@ -34,6 +34,7 @@ export const tabContainerStyle = css`
 export const inlineChildrenWrapperStyle = css`
   display: flex;
   align-items: center;
+  align-self: flex-end;
```
shaneeza, line 37: Screenshot 2023-10-30 at 4 05 59 PM

It doesn't look like the icons and button line up with the tabs to the right. Another solution could be to wrap this container in another div and inside that div add 
```
display: flex;
border: 1px .....;
```
and then remove `align-self: flex-end` from this element.

shaneeza (changes requested):
left a comment!

packages/tabs/src/Tabs/Tabs.styles.ts
spark33: You were right, but the fix was actually somewhere else haha
image

PR #2042, opened 2023-10-23
Adds utility: `getPackageJson`

TheSonOfThomp:
## ✍️ Proposed changes

adds meta tool `getPackageJson`


tools/meta/src/getPackageJson.ts
```diff
@@ -0,0 +1,46 @@
...
+    return;
+  }
+
+  return findPackageJson(dir);
+}
+
+function findPackageJson(dir: string) {
+  const pkgJsonPath = path.resolve(dir, 'package.json');
+
+  if (dir === '/') {
```
bruugey, line 27: are there any other error cases to handle here?

bruugey: `|| !fse.existsSync(dir)`

bruugey: (for example)

bruugey (approved):

PR #2044, opened 2023-10-26
🦨  ESLint plugin

TheSonOfThomp:
## ✍️ Proposed changes

Creates a new package `@lg-tools/eslint-plugin`

This enables us to create custom ESLint rules for both internal & external use


turbo.json
```diff
@@ -5,7 +5,7 @@
       "dependsOn": ["^prebuild"]
     },
     "build": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["^build", "prebuild"],
```
bruugey, line 8: nit but prebuild first?

tools/lint/config/eslint.config.js
```diff
@@ -1,14 +1,14 @@
...
     'eslint:recommended',
-    'plugin:react/recommended',
-    'plugin:react-hooks/recommended',
     'plugin:jest/recommended',
-    'prettier',
     'plugin:jsx-a11y/recommended',
+    'plugin:react/recommended',
+    'plugin:react-hooks/recommended',
     'plugin:storybook/recommended',
+    'prettier',
```
bruugey, line 11: should we add a basic rule inside of our plugin that we can turn on in this PR?

TheSonOfThomp: Not that wouldn't result in a bunch of lint errors...

TheSonOfThomp: I'd have to write a rule that we 100% follow right now

turbo.json
TheSonOfThomp: I don't think the order matters, since this is just a list of dependencies before this step is run

tools/lint/config/eslint.config.js
TheSonOfThomp: Enabled `"no-indirect-imports"`, which we _have_ to follow otherwise rollup won't build

tools/eslint-plugin/src/rules/boolean-verb-prefix.ts
```diff
@@ -0,0 +1,186 @@
+import { AST_NODE_TYPES, TSESTree } from '@typescript-eslint/utils';
+import upperFirst from 'lodash/upperFirst';
+
+import { createRule } from '../utils/createRule';
+// import { deepOmit } from '../utils/deepOmit';
+import { isTestFile } from '../utils/isTestFile';
+import { RuleContext } from '../utils/RuleContext';
+
+const VERBS = ['is', 'are', 'has', 'should', 'did', 'does', 'will', 'use'];
```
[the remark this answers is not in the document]
TheSonOfThomp: _Technically_, following the strict pattern this establishes, you'd want to write `shouldShow`, `willHide`, or `isShown` etc. That said, I'm not super stuck the technicalities, and we can add whatever verbs we want here

tools/eslint-plugin/src/rules/boolean-verb-prefix.ts
```diff
@@ -0,0 +1,186 @@
...
+const booleanComparators = ['===', '==', '>', '>=', '<', '<=', '!=', '!=='];
```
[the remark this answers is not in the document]
TheSonOfThomp: Yup, probably

packages/lib/src/deepOmit/deepOmit.ts
```diff
@@ -0,0 +1,16 @@
+import isObject from 'lodash/isObject';
+import omit from 'lodash/omit';
+
+export function deepOmit(obj: Record<string, any>, paths: Array<string>) {
```
stephl3: could we add a spec file for some basic expected cases?

tools/eslint-plugin/src/utils/isTestFile.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Returns whether the string is a spec, story or testutils file
+ */
+export const isTestFile = (filename: string): boolean => {
```
stephl3: can we add a spec file for this? maybe also showing that `*.test.*`, `*.story*`, and case sensitivity for `testutils`?

tools/eslint-plugin/src/utils/isTestFile.ts
```diff
@@ -0,0 +1,8 @@
...
+  const testFileRegex = /.+\.((spec)|(story)|(testutils))\.tsx?$/;
```
stephl3: 1. Should be `stories` instead of `story` now right?
2. Also, I see some cases of `testutils` and some cases of `testUtils`. Probably worth updating to 1 standard casing and using that here?
3. Do we also care about `getTestUtils` files? Possibly could even check the `testing` subdir depending on this util's purpose

tools/test/config/polyfills/structuredClone.js
```diff
@@ -0,0 +1,26 @@
+function structuredClone(obj) {
```
stephl3, line 11: I know you have some comments for downstream usage of this, but do you think it's worthwhile to include some comments in this file about when this would be used? Seems like a good practice for all polyfills and can help with identifying when it makes sense to remove them

tools/eslint-plugin/src/utils/isTestFile.ts
TheSonOfThomp: This was for an old rule I had made. I'm just going to rm this file

packages/lib/src/deepOmit/deepOmit.ts
TheSonOfThomp: :+1: gonna move this to `lib`

PR #2046, opened 2023-10-26
Jest custom matchers 🦨

TheSonOfThomp:
## ✍️ Proposed changes

- Creates the package `@lg-tools/jest-matchers` that defines custom matchers for Jest
- Updates `jest.config` to import custom matchers

Note: no existing tests are updated in this PR


bruugey (commented):
I think it'd be good to demo in one of our components?

PR #2047, opened 2023-10-26
Fix devDep logic in `@lg-tools/validate`

TheSonOfThomp:
## ✍️ Proposed changes

Updates the logic to check whether a dependency is used in a development file


tools/validate/src/dependencies/config.ts
```diff
@@ -30,8 +30,15 @@ export const ignoreFilePatterns: Array<RegExp> = [
   /.*\/dist\/.*/,
 ];
 
-// these dependencies will be ignored when listed in a package.json
-export const ignoreMatches = [
+/** If a dependency is flagged as bing imported by one of these files, ignore it */
```
bruugey, line 33: nit: being

tools/validate/src/dependencies/config.ts
```diff
@@ -30,8 +30,15 @@ export const ignoreFilePatterns: Array<RegExp> = [
...
+export const ignoreFilePatterns: Array<RegExp> = [
+  /.*package.json?/,
+  /.*README.md/,
+  /.*CHANGELOG.md/,
+];
+
+/** these dependencies will be ignored when listed in a package.json */
```
bruugey, line 40: how come?

TheSonOfThomp: These are globally available dev dependencies. We don't want every package flagged for not having React or Storybook or Jest explicitly declared in its package.json

bruugey: Leave that in the comment :)

bruugey (commented):
how'd we test this? how can i?

bruugey (approved):

PR #2049, opened 2023-10-30
LG-3721 Disable Chromatic for Popover components

TheSonOfThomp:
## ✍️ Proposed changes

Chromatic tests for these components are flaky. Temporarily disabling them

🎟 _Jira ticket:_ https://jira.mongodb.org/browse/LG-3721


packages/guide-cue/src/GuideCue.story.tsx
```diff
@@ -40,6 +40,10 @@ const meta: StoryMetaType<any> = {
         'onPrimaryButtonClick',
       ],
     },
+    chromatic: {
+      disableSnapshot: true,
+      delay: transitionDuration.slowest,
```
bruugey, line 45: Do we need this on Menu too?

bruugey: And SplitButton?

TheSonOfThomp: Don't need the delay if we're disabling it

bruugey: So let's RM where it's duplicative.

bruugey (approved):

PR #2069, opened 2023-11-03
LG-3740, LG-3741: DP initialOpen + disabled

shaneeza:
## ✍️ Proposed changes

### For new components


packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -102,10 +102,34 @@ describe('packages/date-picker', () => {
         expect(menuContainerEl).not.toBeInTheDocument();
       });
 
+      test('menu is initially closed when rendered with `initialOpen` and `disabled`', async () => {
```
TheSonOfThomp, line 105: Super nitty, but more specific tests should go after more general ones (i.e. this one should go after the basic `initialOpen` test)

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -102,10 +102,34 @@ describe('packages/date-picker', () => {
...
+      });
+
       test('menu is initially open when rendered with `initialOpen`', async () => {
         const { findMenuElements } = renderDatePicker({ initialOpen: true });
         const { menuContainerEl } = await findMenuElements();
-        await waitFor(() => expect(menuContainerEl).toBeInTheDocument());
+        expect(menuContainerEl).toBeInTheDocument();
+      });
+
+      test('open menu closes when `disabled` is set to true and fires handler', async () => {
```
TheSonOfThomp, line 120: I would break this into 2 tests:
"menu closes when disabled is toggled to true", and
"validation handler fires when disabled is toggled to true"

TheSonOfThomp: Could wrap it in a `describe` too:
```
describe('when disabled is toggled to true', () => ... )
```

packages/date-picker/src/DatePicker/DatePickerComponent/DatePickerComponent.tsx
```diff
@@ -29,6 +30,13 @@ export const DatePickerComponent = forwardRef<
 
   useBackdropClick(closeMenu, [formFieldRef, menuRef], isOpen);
 
+  useEffect(() => {
```
TheSonOfThomp, line 33: Can we add a short comment about what this effect is doing

PR #2078, opened 2023-11-13
DatePicker [LG-3771, LG-3774] up/down arrows

shaneeza:
## ✍️ Proposed changes

[LG-3771](https://jira.mongodb.org/browse/LG-3771) - When incrementing/decrementing a segment with up/down arrows, the focus moves to the next segment instead of staying focused on the segment you're incrementing/decrementing.
[LG-3774](https://jira.mongodb.org/browse/LG-3774) - Tabbing into the datepicker and pressing the up arrow shows random number
[LG-3768](https://jira.mongodb.org/browse/LG-3768) - keypresses in the input opens the menu and the input focus is hijacked

### For new components

### DIFF
--- a/packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.spec.tsx
@@ -133,5 +133,57 @@ describe('packages/date-picker/date-picker-input', () => {
         expect(yearInput).toHaveFocus();
       });
     });
+
+    describe('Up Arrow', () => {
+      test('keeps the focus in the current segment', () => {
+        const { monthInput } = renderDatePickerInput();
+        userEvent.click(monthInput);
+        userEvent.keyboard('{arrowup}');
+        expect(monthInput).toHaveFocus();
+      });
+
+      test('keeps the focus in the current segment even if the value is valid', () => {
+        const { monthInput } = renderDatePickerInput();
+        userEvent.click(monthInput);
+        userEvent.keyboard('{arrowup}{arrowup}{arrowup}');
+        expect(monthInput).toHaveValue('03');
+        expect(monthInput).toHaveFocus();
+      });
+
+      test('Resets the value to the min value when the new value is greater than the max value', () => {
+        const { monthInput } = renderDatePickerInput();
+        userEvent.click(monthInput);
+        userEvent.keyboard('{arrowup}');
+        expect(monthInput).toHaveValue('01');
+        userEvent.keyboard(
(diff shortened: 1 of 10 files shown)


packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.spec.tsx
```diff
@@ -133,5 +133,57 @@ describe('packages/date-picker/date-picker-input', () => {
...
+        expect(monthInput).toHaveFocus();
+      });
+
+      test('Resets the value to the min value when the new value is greater than the max value', () => {
+        const { monthInput } = renderDatePickerInput();
+        userEvent.click(monthInput);
+        userEvent.keyboard('{arrowup}');
+        expect(monthInput).toHaveValue('01');
+        userEvent.keyboard(
+          '{arrowup}{arrowup}{arrowup}{arrowup}{arrowup}{arrowup}{arrowup}{arrowup}{arrowup}{arrowup}{arrowup}{arrowup}',
```
TheSonOfThomp, line 159: 😆

TheSonOfThomp: Is this what we want, or do we want it to just not do anything? Worth reaching out to @sandynguyenn

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
```diff
@@ -75,27 +75,39 @@ export const DateInputSegment = React.forwardRef<
     const handleKeyDown: KeyboardEventHandler<HTMLInputElement> = e => {
       const { key, target } = e;
 
+      /** adds in leading zero */
+      const getPaddedValue = (num: number) => String(num).padStart(2, '0');
```
TheSonOfThomp, line 79: There should be a utility to do this

TheSonOfThomp: I believe the formatting happens in the `useDateSegments` hook

packages/lib/src/createSyntheticEvent/createSyntheticEvent.ts
```diff
@@ -12,10 +12,13 @@ export const createSyntheticEvent = <
 >(
   event: NativeEventType,
   target: TargetType,
+  key?: string,
```
TheSonOfThomp, line 15: I would make the 3rd arg an object that we spread onto the return object. More extensible this way

TheSonOfThomp: Ideally we'd convert the `target` arg into an object, but that's a breaking change

PR #2080, opened 2023-11-14
Set dispatch as separate job for better testability

spark33:
## ✍️ Proposed changes

The previous release had the following error, so the array has been wrapped in an object:
`For 'properties/client_payload', [{"name"=>"@leafygreen-ui/banner", "version"=>"7.0.17"}, {"name"=>"@leafygreen-ui/button", "version"=>"21.0.11"}, {"name"=>"@leafygreen-ui/callout", "version"=>"9.0.18"}, {"name"=>"@leafygreen-ui/code", "version"=>"14.2.18"}, {"name"=>"@leafygreen-ui/icon", "version"=>"11.25.1"}, {"name"=>"@leafygreen-ui/lib", "version"=>"13.1.0"}, {"name"=>"@leafygreen-ui/popover", "version"=>"11.2.0"}] is not an object.`

The dispatch step is now broken out as a separate job so it can be run separately until it works with the same payload without requiring a new release.

### For new components

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -136,15 +136,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Dispatch to Website repo
-        uses: peter-evans/repository-dispatch@v2
-        continue-on-error: true
-        with:
-          token: ${{ secrets.DESIGN_REPO_ACCESS_TOKEN }}
-          repository: mongodb/design
-          event-type: release-leafygreen-ui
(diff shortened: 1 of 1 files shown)


TheSonOfThomp (approved):
LGTM
Will `needs.release.outputs.publishedPackages` still be available if this is run as a separate job? I think you might need to reference `jobs.xyz`

spark33:
Yeah I think the needs context is available across the file. The dispatch step used to be under the notify job and still referenced needs.release, so it should still have it in the new job it’s in

PR #2092, opened 2023-11-21
React 17 testing 

TheSonOfThomp:


tools/test/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@lg-tools/test",
-  "version": "0.3.2",
-  "description": "Jest setup & config for LeafyGreen repositories",
+  "version": "0.0.0-react17.0",
```
TheSonOfThomp, line 3: does this feel too hacky? Alternative would be to create a separate `@lg-tools/test-react17` package

shaneeza (approved):

PR #2123, opened 2023-12-08
Date Picker [LG-3852] invalid chevron

shaneeza:
## ✍️ Proposed changes

### For new components


packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -293,36 +293,145 @@ describe('packages/date-picker', () => {
       });
 
       describe('Chevrons', () => {
-        test('Left is disabled if prev. month is entirely out of range', async () => {
-          const { openMenu } = renderDatePicker({
-            min: new Date(Date.UTC(2023, Month.December, 1)),
+        describe('left', () => {
+          describe('is disabled', () => {
+            test('when the year and month is before the min', async () => {
```
TheSonOfThomp, line 298: nit: I would shorten the test name to `when value is before min`

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -293,36 +293,145 @@ describe('packages/date-picker', () => {
...
+            test('when the year and month is the same as the min', async () => {
+              const { openMenu } = renderDatePicker({
+                min: new Date(Date.UTC(2023, Month.December, 10)),
+                value: new Date(Date.UTC(2023, Month.December, 1)),
+              });
+
+              const { leftChevron } = await openMenu();
+              expect(leftChevron).toHaveAttribute('aria-disabled', 'true');
+            });
+            test('when the year is the same as the min and the month is before the min', async () => {
```
TheSonOfThomp, line 316: I think this test case would be covered by the first one (L298), (unless there is logic that makes us think that this could be brittle)

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuHeader/utils/isChevronDisabled/index.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export const isChevronDisabled = (
+  direction: 'left' | 'right',
+  day1: Date,
+  day2: Date,
+): boolean => {
+  if (!day1 || !day2) return false;
+
+  if (direction === 'right') {
+    return day1 >= day2 || isSameUTCMonth(day1, day2);
```
TheSonOfThomp, line 19: I'm not sure comparing dates like this is the most reliable. Though, if your tests are passing, it should be fine!

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -293,36 +293,145 @@ describe('packages/date-picker', () => {
       });
 
       describe('Chevrons', () => {
```
TheSonOfThomp, line 295: Needs interaction tests!

PR #2127, opened 2023-12-08
LG-3697: Expose value type for useLeafyGreenTable

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3697](https://jira.mongodb.org/browse/LG-3697)

### For new components


packages/table/src/useLeafyGreenTable/useLeafyGreenTable.types.ts
```diff
@@ -31,17 +31,22 @@ export type LeafyGreenTableCell<T extends LGRowData> = Cell<
 export interface LeafyGreenTableRow<T extends LGRowData>
   extends Row<LGTableDataType<T>> {}
 
-export type LGColumnDef<T extends LGRowData> = ColumnDef<LGTableDataType<T>> & {
+export type LGColumnDef<
+  T extends LGRowData,
+  V extends unknown = unknown,
```
TheSonOfThomp, line 36: Wondering if we should default this to the value types provided in `T`?
(i.e. `V = T extends object ? T[keyof T] : T`)
This way calling `info.getValue()` will be typed to return some value of `T`

e.g.
```ts

interface MyDataType {
  name: string,
  birthday: Date
}

const columns: Array<LGColumnDef<MyDataType>> = [
  {
    accessorKey: 'birthday',
    header: 'Birthday',
    cell: cell => {
      const val = cell.getValue() // string | Date
      return (val as Date).toLocaleString();
    }
  }
  // ...
]

```

spark33: That's an interesting point. I guess the scenario where that wouldn't be best is if the type of the value is so different from the type in T that they'd have to cast to `any` before they consumed it... This PR is mainly for exposing the second type parameter so I feel like keeping it as `unknown` might actually be preferred? If we were always offering additional type safety by defining this, I'd definitely agree, but I'm a tiny bit concerned we'd be more restrictive than safe in this case.

TheSonOfThomp: Not entirely sure I follow. Are you saying in the case where `V = T` we'll run into trouble? 
We could set it to `V = T extends object ? T[keyof T] : unknown` in that case?

TheSonOfThomp: In most cases `type V = T[keyof T]`, but you're right we can't assume that

TheSonOfThomp (approved):

PR #2130, opened 2023-12-11
LG-3439, 3689, 3785: Nested Row animations

spark33:
## ✍️ Proposed changes

1. Adds `disableAnimations` prop to `Table` component
2. Dynamically sets expanded cell height (`max-height` property) on cell content in order support cells with taller content

🎟 _Jira ticket:_ 
[LG-3439](https://jira.mongodb.org/browse/LG-3439)
[LG-3689](https://jira.mongodb.org/browse/LG-3689)
[LG-3785](https://jira.mongodb.org/browse/LG-3785)

### For new components


packages/table/README.md
```diff
@@ -285,7 +286,7 @@ All HTML `tr` element props
 
 `Cell` accepts HTML `td` element props.
 
-> The `Cell` component does not automatically handle overflowing text content, as `text-overflow` depends on the element having `overflow: hidden` and an explicit pixel width value. Refer to the LeafyGreen [Storybook deployment](https://mongodb.github.io/leafygreen-ui/) for an example.
+> All nested row animations are set at the Cell level, with a `max-height` set to 40vh, which should cover most cases with a relatively smooth animation. For taller content, set `disableAnimation={true}` or override the max-height with a `& > div { max-height: ... }` CSS selector.
```
TheSonOfThomp, line 289: I'm a bit confused by the selector. Where should that be applied?

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -79,9 +79,14 @@ export const cellContentContainerStyles = css`
   display: flex;
   align-items: center;
   text-overflow: ellipsis;
-  transition: ${transitionDuration.default}ms ease-in-out;
   transition-property: min-height, max-height, opacity, transform;
   min-height: ${standardCellHeight}px;
+  transition-duration: ${transitionDuration.faster}ms;
+  transition-timing-function: ease-in-out;
```
TheSonOfThomp, line 85: super-nit: we should try to keep `transition-*` properties together

packages/table/src/Cell/Cell.tsx
```diff
@@ -2,21 +2,33 @@ import React from 'react';
...
 
-const Cell = ({ className, align, children, ...rest }: CellProps) => (
-  <td className={cx(baseCellStyles, basicCellStyles, className)} {...rest}>
-    <div className={cx(cellContentContainerStyles, alignmentStyles(align))}>
-      {children}
-    </div>
-  </td>
-);
+const Cell = ({ className, align, children, ...rest }: CellProps) => {
+  const { disableAnimations } = useTableContext();
```
TheSonOfThomp, line 17: Do we have guidance on whether we prefer `disableXYZ` vs `enableXYZ` prop names? @bruugey

packages/table/src/ExpandedContent/ExpandedContent.spec.tsx
```diff
@@ -91,3 +91,26 @@ describe('packages/table/Row/ExpandableContent', () => {
     expect(queryByText('Expandable content test')).toBeInTheDocument();
   });
 });
+
+describe('packages/table/Row/ExpandableContent/Disabled Animations', () => {
```
TheSonOfThomp, line 95: this describe block should be nested inside `'packages/table/Row/ExpandableContent'`

packages/table/src/ExpandedContent/ExpandedContent.spec.tsx
```diff
@@ -91,3 +91,26 @@ describe('packages/table/Row/ExpandableContent', () => {
...
+    );
+    const expandIconButton = getByLabelText('Expand row');
+    expect(expandIconButton).toBeInTheDocument();
+  });
+  test('rows with expandable content render rows as tbody elements with disabled animations', async () => {
+    const { getAllByRole } = render(
+      <RowWithExpandableContent disableAnimations />,
+    );
+    expect(getAllByRole('rowgroup').length).toBe(4); // 1 for thead, 3 for tbody
+  });
```
TheSonOfThomp, line 115: none of these tests fail if I undo your changes to `Cell` and `InternalCell`

packages/table/src/Row/NestedRows.spec.tsx
```diff
@@ -109,3 +109,22 @@ describe('packages/table/Row/NestedRows', () => {
     expect(queryByText('nested row name')).toBeVisible();
   });
 });
+
+describe('packages/table/Row/NestedRows/Disabled Animations', () => {
```
TheSonOfThomp, line 113: This should be nested inside the previous `describe` block

packages/table/src/Cell/Cell.tsx
TheSonOfThomp: I don't see anything in the [styleguide](https://github.com/mongodb/leafygreen-ui/blob/main/STYLEGUIDE.md#variable-naming). Should we?

bruugey: if the question is should we the answer is yes!!

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -94,7 +99,7 @@ export const cellContentTransitionStyles: Record<TransitionStatus, string> = {
   entered: css`
     opacity: 1;
     min-height: ${standardCellHeight}px;
-    max-height: ${standardCellHeight}px;
+    max-height: 40vh;
```
TheSonOfThomp, line 102: changing this degrades the animation significantly. See the 10% speed clips attached 

https://github.com/mongodb/leafygreen-ui/assets/2414030/c9aea06b-d24a-4a2a-b0c6-adf29fd38212


https://github.com/mongodb/leafygreen-ui/assets/2414030/d36e0fca-17a2-4d8f-a045-e50bf2d73939

TheSonOfThomp (changes requested):
The animation is more choppy than before

packages/table/src/ExpandedContent/ExpandedContent.spec.tsx
spark33: That should be the case, since everything is the same functionally even when this prop is passed - that's really what these tests are for

packages/table/src/Cell/Cell.tsx
spark33: Does anyone know if this is an established standard anywhere? My understanding is that variables should never be set up for the user to have double-negatives (i.e. `isNotEditable` would be a bad variable name since the user might have to use `!isNotEditable` which reads as a double negative), but I'm not personally aware of anything against words like `disabled`.

I went with `disableAnimations` because animations are enabled by default we would generally recommend using animations. Setting `disableAnimations` to `true` feels more like we're disabling something that should be used instead of setting `enableAnimations` to `false`, which feels more like we're just electing to not use something that's optional.

If we like the thought process I had above, I can raise a separate PR to add it to our styleguide!

packages/table/README.md
spark33: Edited!

packages/table/src/Cell/Cell.styles.ts
spark33: Resolved with a little hack with the max-heights right before animations begin - I tested it on 10% speed too and it looks good to me. Let me know if you see the same with the updates

packages/table/src/Cell/Cell.tsx
TheSonOfThomp: You could argue that `!disabled` could also read as a double negative (though it's not quite as clear-cut) 
I'm fine with leaving it as-is, since we use `disabled` as a prop name elsewhere

packages/table/src/ExpandedContent/ExpandedContent.spec.tsx
TheSonOfThomp: Ah ok, I was under the impression that this was some broken behavior you were fixing.

packages/table/src/Cell/Cell.styles.ts
TheSonOfThomp: I'm still seeing a kind-of 2-stage effect going on

https://github.com/mongodb/leafygreen-ui/assets/2414030/c9a4a9e0-2b4c-442d-9fe1-a8c54e451fb3

packages/table/src/ExpandedContent/ExpandedContent.spec.tsx
```diff
@@ -90,27 +90,27 @@ describe('packages/table/Row/ExpandableContent', () => {
...
-    );
-    const expandIconButton = getByLabelText('Expand row');
-    expect(expandIconButton).toBeInTheDocument();
-  });
-  test('rows with expandable content render rows as tbody elements with disabled animations', async () => {
-    const { getAllByRole } = render(
-      <RowWithExpandableContent disableAnimations />,
-    );
-    expect(getAllByRole('rowgroup').length).toBe(4); // 1 for thead, 3 for tbody
+  describe('packages/table/Row/ExpandableContent/Disabled Animations', () => {
```
TheSonOfThomp, line 94: nit: could probably just re-name this `"Disabled animations"`

TheSonOfThomp (changes requested):
See latest animation clip

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -79,25 +79,33 @@ export const cellContentContainerStyles = css`
...
   max-height: 0;
+  padding: 0;
 `;
 
 export const cellContentTransitionStyles: Record<TransitionStatus, string> = {
   entered: css`
     opacity: 1;
     min-height: ${standardCellHeight}px;
-    max-height: ${standardCellHeight}px;
+    max-height: 30vh;
```
TheSonOfThomp, line 102: This is still causing a choppy animation.
1. Can you add a story that reproduces the problem this is solving?
2. If we want the cells to animate to the size of the content, we should probably follow the pattern we use in `ExpandableContent` to make sure the height is correct: https://github.com/mongodb/leafygreen-ui/blob/LG-3439-cell-max-height-2/packages/table/src/ExpandedContent/ExpandedContent.tsx#L40-L45

TheSonOfThomp (changes requested):

packages/table/src/Cell/Cell.styles.ts
```diff
@@ -79,25 +79,33 @@ export const cellContentContainerStyles = css`
   display: flex;
   align-items: center;
   text-overflow: ellipsis;
-  transition: ${transitionDuration.default}ms ease-in-out;
-  transition-property: min-height, max-height, opacity, transform;
   min-height: ${standardCellHeight}px;
+  transition-property: min-height, max-height, opacity, transform, padding;
+  transition-duration: ${transitionDuration.slowest}ms;
```
TheSonOfThomp, line 84: also, I think this should still be `.default`

packages/table/src/Cell/InternalCell.tsx
```diff
@@ -30,6 +30,14 @@ const InternalCell = ({
...
   const isSelectable = !!table && !!table.hasSelectableRows;
   const transitionRef = useRef<HTMLElement | null>(null);
+  const contentRef = useRef<HTMLDivElement>(null);
+
+  const contentHeight = useMemo(
+    () => (contentRef.current ? contentRef.current.clientHeight : 0),
+    // Lint flags `content` as an unnecessary dependency, but we want to update `contentHeight` when the value of `children` changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [children],
+  );
```
TheSonOfThomp, line 40: Non-blocking, but can we add a TODO to:
1. break this out into a reusable hook, and
2. Use a `resize` event listener on `contentRef` element to update the height, instead of listening for changes in `children` (which could be flaky)

TheSonOfThomp (approved):
LGTM 💯 

Can we add a row with long content in Storybook so we can verify this in Chromatic

TheSonOfThomp:
Taking a 2nd look at this:

[Add option to disable animations for nested rows on tables](https://jira.mongodb.org/browse/LG-3785) : ✅ Resolved

[Support custom max-heights in Cell contents](https://jira.mongodb.org/browse/LG-3439): 🚫 Not resolved 

[Table row is cutting the long text](https://jira.mongodb.org/browse/LG-3689): 🚫 Not resolved

TheSonOfThomp (approved):

PR #2145, opened 2023-12-19
DatePicker [LG-3899] arrow key errors

TheSonOfThomp:
### Bug:
- Enter '1996-02-29' > press up/down on year. Results in an invalid date ('1995-02-29' not a leap year)
- Enter '2021-03-31' > down on month > Results in "2021-02-31" (not valid)
- Enter '2020-02-29' > up on day > Results in "2020-02-30" (not valid)

### New behavior:
- Pressing up/down on a month or year input that results in an invalid date will show an error message
- Pressing up/down on a date input will roll-over the date value (e.g. Sep 30 -> up -> Sep 1)

### DIFF
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import {
   fireEvent,
-  prettyDOM,
+  // prettyDOM,
   render,
   waitFor,
   waitForElementToBeRemoved,
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import {
   fireEvent,
+  prettyDOM,
   render,
   waitFor,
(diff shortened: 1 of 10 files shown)


packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1859,25 +1860,48 @@ describe('packages/date-picker', () => {
...
                       break;
                     }
 
                     case 'day': {
-                      // There is no case where decrementing a day results in an invalid date value
+                      test('changing date rolls over to number of days-in-month', async () => {
+                        const result = renderDatePicker({
+                          value: newUTC(2020, Month.February, 1),
+                        });
+                        const input = getRelevantInput(result);
```
shaneeza, line 1887: What input is this?

TheSonOfThomp: Whichever input is being tested by the `.each` (L1359)

Would it be more readable to have the signature `getRelevantInput(segment, renderResult)`?

shaneeza: oh i see, the `.each` makes sense. So if the segment is `year`, pressing `arrowdown` would change the year to 2019?

PR #2150, opened 2024-01-02
Updates to InputOption component

bruugey:
# 

- Changes how input option is rendered based on its rendered context (either form or menu)
- Adds support for action types (specifically, a destructive action type in the menu context)
- Renamed selected prop to checked


packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -91,60 +69,176 @@ export const inputOptionWedge = css`
...
+      }
+    `,
+    {
+      [wedge]: showWedge && (state === 'focus' || state === 'checked'),
+      [hover]: !disabled && isInteractive,
+      [titleHoverOverride]:
+        shouldOverride && !disabled && theme === Theme.Dark && isInteractive,
+    },
+  );
 };
```
spark33, line 244: Curious to hear others' thoughts on this, but there's a lot of repeated code here that feels very verbose in a function format. It looks like the `css` string is also the same across multiple functions, so that isn't necessary. I'd personally prefer a style object approach to writing this CSS (the way we handle mode styles in most components)

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -3,57 +3,78 @@ import React from 'react';
...
+          })
+        : getMenuElementStyle({
+            theme,
+            checked,
+            highlighted,
+            disabled,
+            showWedge,
+            actionType,
+            isInteractive,
+          });
```
spark33, line 62: Feels weird to me there's some styling logic here on the component itself when we have a function dedicated to styling logic in the styles.ts file. I think, if possible, we should use `cx` to split styles on the component and `css` strings to define the styles. If there's a need to define more complex logic, we can use nested style objects using Record<Theme, Record<SomeProp, string>> or inline ternaries inside a simple `() => css` function

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -26,9 +44,9 @@ export interface BaseInputOptionProps {
   highlighted?: boolean;
 
   /**
-   * Whether the component is selected, regardless of keyboard navigation
+   * Whether the component is checked, regardless of keyboard navigation
    */
-  selected?: boolean;
+  checked?: boolean;
```
spark33, line 49: Didn't we align on `selected` to avoid confusion with checkboxes where `checked` is a native HTML prop?

spark33 (changes requested):
I can guess as to why `input-option` needs to be Polymorphic, but it'd be nice to have a record of why that's necessary - ideally in a linked ticket. Some thoughts on the styles as well.

packages/input-option/src/InputOption/InputOption.types.ts
bruugey: we have docs on this and I'm 99% sure that we aligned on checked

packages/input-option/src/InputOption/InputOption.tsx
bruugey: it's just deciding which element styles to use - i can wrap those two functions in another function if you prefer?

packages/input-option/src/InputOption/InputOption.style.ts
bruugey: there's too many conditions to make that work - we have theme.ts but to decide how to apply all of the stlyes, i think there's too much logic.

TheSonOfThomp:
@bruugey Can you add what updates were made to the PR summary?

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -3,6 +3,24 @@ import { PropsWithChildren } from 'react';
...
+} as const;
+
+type ActionType = (typeof ActionType)[keyof typeof ActionType];
+
+export { ActionType };
+
+const RenderedContext = {
+  FormElement: 'formElement',
+  Menu: 'menu',
+} as const;
```
TheSonOfThomp, line 18: Can we a bit more terse here? 
```
const RenderedContext = {
  Form: 'form',
  Menu: 'menu',
} as const;
```

TheSonOfThomp: Also, can we add some comments on what this means, and what it changes about the component?

packages/input-option/src/InputOption/InputOption.types.ts
TheSonOfThomp: `checked`: https://docs.google.com/document/d/1HlzHBlcL5MZG5Lm5m-OIfkI0Dhh4x1XBezLUipV3SW0/edit#heading=h.f9ili2y9gq48

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -3,57 +3,78 @@ import React from 'react';
...
       isInteractive = true,
       className,
+      actionType = ActionType.Default,
+      renderedContext = RenderedContext.FormElement,
       ...rest
     },
     ref,
   ) => {
-    const { Component } = usePolymorphic(as);
+    const { Component } = useInferredPolymorphic(as, rest, 'div');
```
TheSonOfThomp, line 41: Does this need to be inferred? Can we just leave it as basic polymorphic?

TheSonOfThomp: Also, is defaulting to `div` intentional? Why not `li`?

packages/input-option/src/InputOption/InputOption.tsx
TheSonOfThomp: I agree, this feels off. Seems like `renderedContext` should just be an additional prop passed into a single function

packages/input-option/src/InputOption/InputOption.types.ts
TheSonOfThomp, line 49: Needs a major changeset

packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -91,60 +69,178 @@ export const inputOptionWedge = css`
...
-  [Theme.Dark]: css`
-    outline: none;
-    background-color: ${palette.blue.dark3};
-    color: ${palette.blue.light3};
+  &:before {
+    content: unset;
+  }
+`;
 
+export const getFormElementStyle = ({
```
TheSonOfThomp, line 80: How much redundancy can we eliminate in these functions with something like:
```
const contextStyles = renderedContext  === 'menu' ? menuThemeStyles : formThemeStyles;
```

If there's a fair bit of overlap, it might even be worth refactoring `menuThemeStyles` & `formThemeStyles` into a single object, using the `renderedContext` as the key. Then we can write:

```
const contextStyles = renderedContextStyles[renderedContext];
```

packages/input-option/src/InputOption/InputOption.style.ts
TheSonOfThomp: I'm fine with the functional approach if there's a lot of branching. I would like to see these functions combined into one though, since I see a fair bit of overlap

packages/input-option/src/InputOption/themes.ts
```diff
@@ -0,0 +1,169 @@
...
+import { palette } from '@leafygreen-ui/palette';
+
+const State = {
+  Default: 'default',
+  Hover: 'hover',
+  Focus: 'focus',
+  Disabled: 'disabled',
+  Checked: 'checked',
+  Destructive: 'destructive',
+} as const;
```
TheSonOfThomp, line 11: I don't think these states are mutually exclusive. An item can be `checked` + `hovered` at the same time. A user can `focus` on a `destructive` item.

packages/input-option/src/InputOption/themes.ts
```diff
@@ -0,0 +1,169 @@
+import { Theme } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+
+const State = {
+  Default: 'default',
+  Hover: 'hover',
+  Focus: 'focus',
```
TheSonOfThomp, line 7: `highlight` since it may not be true keyboard focus in all cases

packages/search-input/src/SearchResult/SearchResult.tsx
```diff
@@ -40,9 +40,10 @@ export const SearchResult = InferredPolymorphic<SearchResultProps, 'li'>(
       rest['aria-label'] ?? (rest['aria-labelledby'] ? '' : textContent);
 
     return (
+      // @ts-expect-error
```
TheSonOfThomp, line 43: why?

packages/search-input/src/SearchResult/SearchResult.tsx
```diff
@@ -40,9 +40,10 @@ export const SearchResult = InferredPolymorphic<SearchResultProps, 'li'>(
...
       <InputOption
         {...rest}
-        as={as}
+        as={renderedAs}
```
TheSonOfThomp, line 46: SInce we're just passing the `as` prop through, I believe we can just use the [Basic Polymorphic](https://github.com/mongodb/leafygreen-ui/blob/main/packages/polymorphic/README.md#basic-usage) usage

packages/input-option/src/InputOption/InputOption.style.ts
bruugey: there's too much nuance and little rules for each to combine them into one - i can have a wrapper around both functions that just calls its respective if you'd like

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -26,9 +44,9 @@ export interface BaseInputOptionProps {
   highlighted?: boolean;
```
TheSonOfThomp, line 44: Can we update the above `TERMINOLOGY` section

packages/search-input/src/SearchResultGroup/SearchResultGroup.tsx
```diff
@@ -25,10 +25,12 @@ export const SearchResultGroup = ({
 
   return (
     <div>
+      {/* @ts-expect-error li/div incongruency */}
```
TheSonOfThomp, line 28: We should fix this

TheSonOfThomp (changes requested):
Left a few comments.
@bruugey Can you add what updates were made to the PR summary?

packages/input-option/src/InputOption/InputOption.types.ts
bruugey: yes

packages/input-option/src/InputOption/InputOption.tsx
bruugey: div is intentional - sometimes we'll want LI but not always. Divs are more flexible, more adaptable, and can be just as accessible so long as they receive the right aria attributes.

packages/input-option/src/InputOption/InputOption.types.ts
bruugey: Good catch.

packages/input-option/src/InputOption/InputOption.style.ts
bruugey: updated

packages/input-option/src/InputOption/themes.ts
bruugey: this doesn't mean that theyre mutually exclusive (if you look at how hovered, focus styles are applied)

packages/input-option/src/InputOption/themes.ts
bruugey: Updated, you're right

packages/search-input/src/SearchResult/SearchResult.tsx
bruugey:
```
Types of property 'href' are incompatible.
        Type 'NodeUrlLike | ((string | NodeUrlLike) & string) | undefined' is not assignable to type 'string | NodeUrlLike'.
          Type 'undefined' is not assignable to type 'string | NodeUrlLike'.ts(2322)
```

packages/search-input/src/SearchResult/SearchResult.tsx
bruugey: It's not just passing thru, we're setting a default so we have to ask if `href` exists or not

packages/search-input/src/SearchResult/SearchResult.tsx
bruugey: This exists even if we don't destructure as/href

packages/input-option/src/InputOption/InputOption.types.ts
bruugey: yep!

packages/search-input/src/SearchResultGroup/SearchResultGroup.tsx
bruugey: I think this is an issue with Polymorphic 
```
Type '{ children: Element; slot?: string | undefined; style?: CSSProperties | undefined; title?: string | undefined; "aria-label"?: string | undefined; "aria-labelledby": string; ... 263 more ...; as: "li"; }' is not assignable to type 'IntrinsicAttributes & InferredPolymorphicPropsWithRef<"li", InputOptionProps>'.
  Type '{ children: Element; slot?: string | undefined; style?: CSSProperties | undefined; title?: string | undefined; "aria-label"?: string | undefined; "aria-labelledby": string; ... 263 more ...; as: "li"; }' is not assignable to type 'InheritedProps<"li", InputOptionProps>'.
    Types of property 'onClick' are incompatible.
      Type 'MouseEventHandler<HTMLDivElement> | undefined' is not assignable to type 'MouseEventHandler<HTMLLIElement> | undefined'.
        Type 'MouseEventHandler<HTMLDivElement>' is not assignable to type 'MouseEventHandler<HTMLLIElement>'.
          Type 'HTMLDivElement' is missing the following properties from type 'HTMLLIElement': type, valuets(2322)
          ```

File: packages/input-option/src/InputOption/InputOption.tsx (hunk 5 above)
Comment (bruugey) [reply packages/input-option/src/InputOption/InputOption.tsx:40]: I guess we could leave it as base polymorphic, and have the inferred logic happen in the components when they send it down?

Comment (TheSonOfThomp) [reply packages/input-option/src/InputOption/InputOption.tsx]: Yeah, I think stacking too many `Inferred` components will lead to headaches down the line

File: packages/input-option/src/InputOption/InputOption.style.ts [hunk 13]
```diff
@@ -30,43 +40,12 @@ export const inputOptionStyles = css`
...
-  [Theme.Dark]: css`
-    &:hover {
-      outline: none;
-      background-color: ${palette.gray.dark4};
-    }
-  `,
-};
-
 /** in px */
 const wedgeWidth = spacing[1];
```
shaneeza, line 50: new token?

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -3,13 +3,48 @@ import { PropsWithChildren } from 'react';
...
+export { State };
+
+export type FormState = Exclude<State, 'destructive'>;
+
 /**
  * TERMINOLOGY
  *
- * `focused`: The element is "focused" via keyboard navigation
- * (Does not mean `:focus`, since input options are not focused in this sense)
+ * `highlighted`: The element is aria-selected or "focused" via keyboard navigation
```
shaneeza, line 44: This is a little confusing because this says the `element is aria-selected` while `checked` says `The element is selected`.

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -42,6 +77,17 @@ export interface BaseInputOptionProps {
...
   isInteractive?: boolean;
+
+  /**
+   * Styles input based on intended action
+   * @default 'default'
+   */
+  actionType?: ActionType;
+
+  /**
+   * Determines how the items are styled
```
shaneeza, line 88: Can you add a default here?

shaneeza: Can you add the description on [L15](https://github.com/mongodb/leafygreen-ui/pull/2150/files#diff-74aa5a26066cb7cef22f2f104153f3dc93379a26acdb577f026a3199ca61c0abR15) to this? An outsider looking in might not understand this.

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -42,6 +77,17 @@ export interface BaseInputOptionProps {
    * @default true
    */
   isInteractive?: boolean;
+
+  /**
+   * Styles input based on intended action
+   * @default 'default'
+   */
+  actionType?: ActionType;
```
shaneeza, line 85: If I understand correctly, since a form can't be destructive, can we make this conditional so that a type error shows when `renderedContext` is `form` and `actionType` is `destructive`? Similar to what we do in Popover with `usePortal`?

packages/search-input/src/SearchResultGroup/SearchResultGroup.tsx
```diff
@@ -25,10 +25,12 @@ export const SearchResultGroup = ({
 
   return (
     <div>
+      {/* @ts-expect-error Polymorphic-type mismatch */}
       <InputOption
-        aria-label={label}
+        aria-labelledby={label}
```
shaneeza, line 30: Looking at the props, `label` is a title string, so it should be `aria-label`.  `aria-labelledby` is for referencing another element.

shaneeza:
I know that this is internal, but could we still have a README? If you're new to this component, this could be a lot of information to try to decipher.

packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -8,12 +8,22 @@ import {
...
 export const descriptionClassName = createUniqueClassName(
   'input-option-description',
 );
+export const leftGlyphClassName = createUniqueClassName(
+  'input-option-left-glyph',
+);
 
-export const inputOptionStyles = css`
+const hoverSelector = '&:hover, &[data-hover="true"]';
+const focusSelector = '&:focus, &:focus-visible, &[data-focus="true"]';
```
shaneeza, line 24: Do we need both focus and focus-visible?

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -9,51 +9,100 @@ import {
...
+  RenderedContext,
+  State,
+} from './InputOption.types';
 
-export const InputOption = Polymorphic<InputOptionProps>(
+export const InputOption = Polymorphic<InputOptionProps, 'div'>(
   (
     {
-      as = 'li' as PolymorphicAs,
+      as = 'div' as PolymorphicAs,
```
shaneeza, line 33: Why the change from `li` to `div`?

shaneeza, line 33: Do you wanna group these props by defaults and aliases? I know you're a fan of that!

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -9,51 +9,100 @@ import {
...
     },
     ref,
   ) => {
     const { Component } = usePolymorphic(as);
-    const { theme } = useDarkMode(darkModeProp);
+    const { theme, darkMode } = useDarkMode(darkModeProp);
+
+    let state: State = 'default';
+
+    if (disabled) {
```
shaneeza, line 53: Can this be extracted?

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -9,51 +9,100 @@ import {
...
+          getContextStyles(renderedContext, state, theme),
           {
-            [inputOptionWedge]: showWedge,
-            [inputOptionHoverStyles[theme]]: isInteractive,
-            [inputOptionActiveStyles[theme]]:
-              isInteractive && (selected || highlighted),
-            [inputOptionDisabledStyles[theme]]: disabled,
-            [titleSelectionStyles]: selected,
+            [inputOptionWedge]: showWedgeProp,
+            [formWedgeStyles(darkMode)]:
```
shaneeza, line 96: nit: would it be more consistent to stick with [theme]?

packages/input-option/src/InputOption/themes.ts
```diff
@@ -0,0 +1,154 @@
...
+    destructive: {
+      title: palette.red.light1,
+      description: palette.gray.light1,
+      leftGlyph: palette.red.light1,
+      bgColor: palette.gray.dark3,
+    },
+  },
+};
+
+export const contextColors: Record<RenderedContext, any> = {
```
shaneeza, line 151: nit: should we be a little more specific and avoid using `any`?

TheSonOfThomp:
@brookeyalof What's the status of this PR?

PR #2153, opened 2024-01-04
LG-3813: Alternate row color overrides selected row color

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3813](https://jira.mongodb.org/browse/LG-3813)

### For new components


packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -87,7 +87,8 @@ const InternalRowWithRT = <T extends LGRowData>({
         className={cx(
           {
             [grayZebraRowStyles[theme]]: isOddVSRow && shouldAlternateRowColor,
-            [zebraStyles[theme]]: !virtualRow && shouldAlternateRowColor,
+            [zebraStyles[theme]]:
+              !virtualRow && shouldAlternateRowColor && !isSelected,
```
TheSonOfThomp, line 91: @spark33 Should we be applying this condition to the above styles too? (`grayZebraRowStyles`)

packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -86,8 +86,10 @@ const InternalRowWithRT = <T extends LGRowData>({
       <InternalRowBase
         className={cx(
           {
-            [grayZebraRowStyles[theme]]: isOddVSRow && shouldAlternateRowColor,
-            [zebraStyles[theme]]: !virtualRow && shouldAlternateRowColor,
+            [grayZebraRowStyles[theme]]:
+              isOddVSRow && shouldAlternateRowColor && !isSelected,
```
TheSonOfThomp, line 90: nit: should we break `shouldAlternateRowColor && !isSelected` into a separate variable?

TheSonOfThomp: Also, does this style depend on `!virtualRow` as well?

TheSonOfThomp (approved):
Left a few nits, but they're non-blocking

packages/table/src/Row/InternalRowWithRT.tsx
spark33: `isOddVSRow` should imply `!!virtualRow`, so shouldn't be necessary!

PR #2159, opened 2024-01-08
LG-3797: Make Popover props optional in Menu

spark33:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3797](https://jira.mongodb.org/browse/LG-3797)

### For new components


shaneeza (commented):
Can you add some tests in components that use Popover?

spark33:
> Can you add some tests in components that use Popover?

@shaneeza What would we be testing? These props aren't required, so the appropriate tests would be to make sure passing these props won't cause TS errors, and neither will NOT passing these props, and that should be handled by all instances of `render` inside all tests of components that use Popover, no?

package.json
```diff
@@ -43,7 +43,6 @@
     "@types/lodash": "^4.14.170",
     "@types/react": "18.2.23",
     "@types/react-dom": "18.2.8",
-    "@types/react-transition-group": "4.4.8",
```
shaneeza, line 46: why are you removing this? is it because its a devDep?

TheSonOfThomp: We're also installing this within individual packages, and having it here could cause some version discrepancies

TheSonOfThomp:
@spark33 Was playing around with dependencies, and it looks like adding "@types/react-transition-group": "^4.4.8",  to `Popover` dependencies (not dev) should resolve the issue

TheSonOfThomp:
Closing in favor of https://github.com/mongodb/leafygreen-ui/pull/2159

PR #2160, opened 2024-01-09
Proposal: Change styles structure

spark33:
## ✍️ Proposed changes

This PR is a set of proposed edits to #2150.

Several things to note - 

1. I noticed the LiveExample was breaking if `renderedContext === "form" && actionType === "destructive"`. I've restricted this from breaking the component and updated the LiveExample to still be interactive with both renderedContext values.
2. I've tried to migrate the styles over to something more DRY and readable - about halves the lines of style code.

### DIFF
--- a/packages/input-option/src/InputOption/InputOption.story.tsx
+++ b/packages/input-option/src/InputOption/InputOption.story.tsx
@@ -91,19 +91,41 @@ export default meta;
 export const LiveExample: StoryFn<
   InputOptionProps & InputOptionContentProps
 > = (props: InputOptionProps & InputOptionContentProps) => {
-  const { leftGlyph, rightGlyph, description, ...rest } = props;
+  const { leftGlyph, rightGlyph, description, renderedContext, ...rest } =
+    props;
   return (
-    <InputOption {...rest}>
-      <InputOptionContent
-        leftGlyph={leftGlyph ? <Icon glyph={leftGlyph as string} /> : undefined}
-        rightGlyph={
-          rightGlyph ? <Icon glyph={rightGlyph as string} /> : undefined
-        }
-        description={description}
+    <div>
+      <InputOption
+        {...rest}
+        renderedContext={RenderedContext.Form}
+        actionType="default" // input options in forms can only be default
       >
-        Some text
-      </InputOptionContent>
-    </InputOption>
+        <InputOptionContent
+          leftGlyph={
+            leftGlyph ? <Icon glyph={leftGlyph as string} /> : undefined
+          }
+          rightGlyph={
+            rightGlyph ? <Icon glyph={rightGlyph as string} /> : undefined
+          }
+          description={description}
+        >
+          Some text
+        </InputOptionContent>
+      </InputOption>
+      <InputOption {...rest} renderedContext={RenderedContext.Menu}>
+        <InputOptionContent
+          leftGlyph={
+            leftGlyph ? <Icon glyph={leftGlyph as string} /> : undefined
+          }
+          rightGlyph={
+            rightGlyph ? <Icon glyph={rightGlyph as string} /> : undefined
+          }
+          description={description}
+        >
+          Some text
+        </InputOptionContent>
+      </InputOption>
+    </div>
   );
 };
 LiveExample.parameters = {
@@ -112,6 +134,9 @@ LiveExample.parameters = {
 
 export const FormElement = () => <></>;
 FormElement.parameters = {
+  argTypes: {
+    actionType: 'default',
+  },
   generate: {
     args: {
       renderedContext: RenderedContext.Form,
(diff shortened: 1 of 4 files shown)


packages/input-option/src/InputOption/InputOption.story.tsx
```diff
@@ -91,19 +94,47 @@ export default meta;
...
-          rightGlyph ? <Icon glyph={rightGlyph as string} /> : undefined
-        }
-        description={description}
-      >
-        Some text
-      </InputOptionContent>
-    </InputOption>
+    <div style={{ display: 'flex', gap: '16px' }}>
+      <div>
+        <pre>Form</pre>
```
TheSonOfThomp, line 102: nit: Should probably be separate stories, but this is fine for demo

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -60,12 +84,18 @@ export const InputOption = Polymorphic<InputOptionProps, 'div'>(
...
         tabIndex={-1}
         className={cx(
+          getContextStyles(renderedContext, state, theme),
+          getTextStyles(renderedContext, state, theme),
           {
-            [inputOptionWedge]: showWedge,
+            [inputOptionWedge]: showWedgeProp,
+            [getWedgeStyles(renderedContext, state, theme)]: shouldRenderWedge,
+            [getHoverStyles(renderedContext, theme)]:
+              !disabled && isInteractive && state !== State.Highlight,
```
TheSonOfThomp, line 93: nit: I would extract his into a separate `shouldShowHoverStyles` var or something like that

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -60,12 +84,18 @@ export const InputOption = Polymorphic<InputOptionProps, 'div'>(
         aria-checked={checked}
         tabIndex={-1}
         className={cx(
+          getContextStyles(renderedContext, state, theme),
+          getTextStyles(renderedContext, state, theme),
           {
-            [inputOptionWedge]: showWedge,
+            [inputOptionWedge]: showWedgeProp,
+            [getWedgeStyles(renderedContext, state, theme)]: shouldRenderWedge,
```
TheSonOfThomp, line 91: Why are these separate?

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -23,6 +23,21 @@ type RenderedContext = (typeof RenderedContext)[keyof typeof RenderedContext];
...
 export { RenderedContext };
 
+const State = {
+  Default: 'default',
+  Hover: 'hover',
+  Highlight: 'highlight',
+  Disabled: 'disabled',
+  Checked: 'checked',
+  Destructive: 'destructive',
+} as const;
```
TheSonOfThomp, line 33: What's the purpose of this type? It doesn't look like we're exposing it to consumers, and the fact that the states in this type aren't actually mutually exclusive is confusing (e.g. an option can be both "hovered" and "checked" at the same time, so what is `State` in this case?)

TheSonOfThomp: To me there are multiple different "state"-type variables:

`type State =  "default" | "disabled" | "checked" | "highlight"`

`type ActionType = "default" | "destructive"`

and also `isHovered` (which can apply to any combination of `State` & `ActionType`

TheSonOfThomp: Then, in `.style.ts` we have 

`get{Element}StateStyle(context)`, `get{Element}ActionTypeStyle(context)` and `get{Element}HoverStyle(context)`

TheSonOfThomp: You'd also probably also want to simplify `XYZContextColors` to match

TheSonOfThomp (commented):
Broad strokes, I like the pattern of using the emotion `css` object to toggle different styles rather than putting that logic inside a single function.

I do have concerns about how we're defining the styles though, since we have what's logically 4 inter-dependent layers of state (action type, state, hover, darkMode).

It's not perfect, but the way I styled `CalendarCell` may be relevant here: https://github.com/mongodb/leafygreen-ui/blob/adam/date-picker-lg-3152/packages/date-picker/src/shared/components/Calendar/CalendarCell/CalendarCell.styles.ts

packages/input-option/src/InputOption/InputOption.tsx
spark33: Not sure, but these styles have different logic for how they're rendered, and I was just maintaining that. I really just wanted to propose an alternate organizational approach to Brooke's PR

packages/input-option/src/InputOption/InputOption.story.tsx
spark33: We can have separate stories, but if we ever want it on .design, we'd probably want to show both in one story for the LiveExample. Similar case to EmptyState

PR #2161, opened 2024-01-09
React 17 `renderHook`

TheSonOfThomp:
Attempts to make usage of `renderHook` consistent between R17 & R18 test environments

To do this I needed to:

1. Re-export a version `renderHook` from `@leafygreen-ui/testing-lib`, either originating from `@testing-library/react` or `@testing-library/react-hooks`, depending on what packages we have installed (concept lifted from [react-spectrum](https://github.com/adobe/react-spectrum/blob/main/packages/dev/test-utils/src/renderOverride.js))
2. Update imports for all test files that use `renderHook`
3. Update tests to ensure they work using both the `/react-hooks` and `/react` versions of `renderHook`
    a. Update `toastReducer` logic to ensure we return the correct value from `updateToast`
4. Add `@leafygreen-ui/testing-lib` to `@lg-tools/test` dependencies
5. Exclude `@leafygreen-ui/testing-lib` from depcheck validation
6. Update `@leafygreen-ui/testing-lib` build scripts to not require `@lg-tools/cli` (this creates a circular dependency)

**Optional "todo"** : since we're now building `@leafygreen-ui/testing-lib` in the same way we build other `@lg-tools/*` packages, we may want to consider rescoping the package as `@lg-tools/testing-lib`

### DIFF
--- a/packages/hooks/src/hooks.spec.tsx
+++ b/packages/hooks/src/hooks.spec.tsx
@@ -252,10 +252,10 @@ describe('packages/hooks', () => {
       expect(pollHandler).toHaveBeenCalledTimes(0);
     });
 
-    test('when document is not visible', () => {
+    test('when document is not visible', async () => {
       const pollHandler = jest.fn();
 
-      renderHook(() => usePoller(pollHandler));
+      const { rerender } = renderHook(() => usePoller(pollHandler));
 
       expect(pollHandler).toHaveBeenCalledTimes(1);
 
@@ -263,7 +263,6 @@ describe('packages/hooks', () => {
       act(() => {
         document.dispatchEvent(new Event('visibilitychange'));
       });
-
       jest.advanceTimersByTime(30e3);
 
       expect(pollHandler).toHaveBeenCalledTimes(1);
@@ -272,6 +271,7 @@ describe('packages/hooks', () => {
       act(() => {
         document.dispatchEvent(new Event('visibilitychange'));
       });
+      jest.advanceTimersByTime(30e3);
 
       // immediate triggers the pollHandler
       expect(pollHandler).toHaveBeenCalledTimes(2);
@@ -316,9 +316,6 @@ describe('packages/hooks', () => {
 
       rerender();
       expect(result.current).toEqual(123);
-
-      rerender(999);
-      expect(result.current).toEqual(undefined);
     });
   });
 
--- a/packages/hooks/src/hooks.spec.tsx
+++ b/packages/hooks/src/hooks.spec.tsx
@@ -268,12 +268,13 @@ describe('packages/hooks', () => {
       expect(pollHandler).toHaveBeenCalledTimes(1);
 
       mutableDocument.visibilityState = 'visible';
-      rerender(pollHandler);
 
       act(() => {
         document.dispatchEvent(new Event('visibilitychange'));
       });
       jest.advanceTimersByTime(30e3);
+
+      rerender(pollHandler);
 
       // immediate triggers the pollHandler
       expect(pollHandler).toHaveBeenCalledTimes(2);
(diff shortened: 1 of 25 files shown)


packages/testing-lib/src/RTLOverrides.ts
```diff
@@ -0,0 +1,31 @@
...
+  : Fallback;
+
+/**
+ * Re-exports `renderHook` from `"@testing-library/react"` if it exists,
+ * or from `"@testing-library/react-hooks"`
+ *
+ * (used when running in a React 17 test environment)
+ */
+export const renderHook: Exists<typeof RTL, 'renderHook'> =
+  (RTL as any).renderHook ??
```
spark33, line 14: Does `renderHook` only get exported from certain versions of RTL or something? What's the uncertainty that makes use need the `Exist` type?

packages/testing-lib/src/waitForState.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export const waitForState = async <T extends any>(
+  callback: () => T,
+): Promise<T> => {
+  let val: T;
+  await act(() => {
+    val = callback();
+  });
+
+  // @ts-expect-error
```
spark33, line 17: What error are we expecting?

tools/validate/src/dependencies/validateListedDependencies.ts
```diff
@@ -38,16 +38,19 @@ export function validateListedDependencies(
     const listedButOnlyUsedAsDev = listedDependencies.filter(
       listedDepName =>
         !importedPackagesInSourceFile.includes(listedDepName) &&
-        !ignoreDependencies.includes(listedDepName),
+        !externalDependencies.some(glob => {
+          const regexPattern = glob.replace('*', '.+'); // a super rough conversion of glob pattern to regexp
```
shaneeza, line 42: This is not blocking or anything but can you give an example of whats happening here?

packages/testing-lib/src/RTLOverrides.ts
TheSonOfThomp: Yeah. Prior to `@testing-library/react@13.1.0`, `renderHook` was only available in the standalone, `/react-hooks` package.  The `react-hooks` package has since been soft-deprecated for React 18 use-cases, but `/react > 13` is not compatible with React 18, so we need to toggle which version of `renderHook` we import

tools/validate/src/dependencies/validateListedDependencies.ts
TheSonOfThomp: 👍 Moved the logic to a utility function. We're converting a [`glob` ](https://en.wikipedia.org/wiki/Glob_(programming))pattern to RegEx 

`globToRegex('@leafygreen-ui/*') // /@leafygreen-ui\/.+/`

packages/testing-lib/src/waitForState.ts
TheSonOfThomp: Updated. `val` is technically being referenced before it's declared as far as the TS compiler is concerned

packages/hooks/src/hooks.spec.tsx
```diff
@@ -269,6 +271,7 @@ describe('packages/hooks', () => {
       act(() => {
         document.dispatchEvent(new Event('visibilitychange'));
       });
+      jest.advanceTimersByTime(30e3);
```
spark33, line 274: Should this be replaced by the new `waitForState` util?

TheSonOfThomp: Possibly, though I'm not too familiar with this hook, so I used the same logic as on L264

shaneeza: I don't think `waitForState` can replace `advanceTimersByTime`.  We ran into issues with some DatePicker tests where awaiting an act call didn't wait for some popover callbacks to finish. We had to use `advanceTimersByTime`. I could be wrong though.

shaneeza (approved):

PR #2167, opened 2024-01-12
Adds multiple package managers to `link` script

TheSonOfThomp:
Previously the `lg link` script would only install & link packages using yarn. Now it will check the destination app's package manager (via lock file) and link packages using the correct package manager.

Additionally, this PR tests for many `link` utils

### DIFF
--- a/tools/link/src/utils/createLinkFrom.spec.ts
+++ b/tools/link/src/utils/createLinkFrom.spec.ts
@@ -1,0 +1,48 @@
+import { ChildProcess } from 'child_process';
+import xSpawn from 'cross-spawn';
+import fsx from 'fs-extra';
+import { createLinkFrom } from './createLinkFrom';
+import { MockChildProcess } from './mocks.testutils';
+import path from 'path';
+
+describe('tools/link/createLinkFrom', () => {
+  let spawnSpy: jest.SpyInstance<ChildProcess>;
+
+  beforeAll(() => {
+    fsx.emptyDirSync('./tmp');
+    fsx.rmdirSync('./tmp/');
+    fsx.mkdirSync('./tmp/');
+    fsx.mkdirSync('./tmp/test');
+    fsx.mkdirSync('./tmp/test/test-package');
+  });
+
+  beforeEach(() => {
+    spawnSpy = jest.spyOn(xSpawn, 'spawn');
+    spawnSpy.mockImplementation((..._args) => new MockChildProcess());
+  });
+
+  afterEach(() => {
+    spawnSpy.mockRestore();
+    fsx.emptyDirSync('./tmp');
+  });
+
+  afterAll(() => {
+    fsx.rmdirSync('./tmp/');
+  });
+
+  test('calls `npm link` command from package directory', () => {
+    createLinkFrom(path.resolve('./tmp/'), {
+      scopeName: '@example',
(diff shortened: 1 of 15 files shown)


bruugey (commented):
Needs a changeset

tools/link/src/utils/createLinkFrom.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+      scopeName: '@example',
+      scopePath: 'scope',
+      packageName: 'test-package',
+    });
+
+    expect(spawnSpy).toHaveBeenCalledWith(
+      'npm',
+      ['link'],
+      expect.objectContaining({
+        cwd: expect.stringContaining('tmp/scope/test-package'),
```
bruugey, line 45: should this be @example/test-pacakge?

tools/link/src/utils/install.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { ChildProcess } from 'child_process';
+import xSpawn from 'cross-spawn';
+import fsx from 'fs-extra';
+
+// const withLocalTmpDir = require('with-local-tmp-dir')
```
bruugey, line 5: nit rm?

tools/link/src/utils/install.ts
```diff
@@ -1,7 +1,13 @@
...
+import { SupportedPackageManager } from '@lg-tools/meta/src/getPackageManager';
 import { spawn } from 'cross-spawn';
+import fsx from 'fs-extra';
 
+/**
+ * @deprecated use `installPackages` instead
+ */
 export function yarnInstall(path: string) {
-  return new Promise((resolve, reject) => {
+  return new Promise(resolve => {
```
bruugey, line 10: should we just rm this instead of deprecating it?

tools/meta/src/getPackageManager.ts
```diff
@@ -1,11 +1,19 @@
 import fse from 'fs-extra';
+
+export type SupportedPackageManager = `npm` | `yarn` | `pnpm`;
+
 /** Reads the lock file to determine which package manager to use */
-export function getPackageManager(appPath: string): `npm` | `yarn` {
+export function getPackageManager(
+  appPath: string,
+  defaultPackageManager: SupportedPackageManager = 'npm',
```
bruugey, line 8: why is defaultPackageManager an arg?

TheSonOfThomp: Doesn't need to be, but I wanted to add the flexibility of preferring some other default if there's no lock file found

tools/link/src/utils/install.ts
TheSonOfThomp: Probably 👍

tools/link/src/utils/createLinkFrom.spec.ts
TheSonOfThomp: No, `cwd` is the directory where the command is run from. We expect that to be the `test-package` dir

TheSonOfThomp: No, `cwd` is the directory where the command is run from. We expect that to be the `test-package` dir

bruugey (commented):
How did you test this?

bruugey (approved):

PR #2173, opened 2024-01-18
LG-3857: Date Picker Correct focus state when navigating to current select month

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3857](https://jira.mongodb.org/browse/LG-3857)

### For new components

### DIFF
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -1082,42 +1082,26 @@ describe('packages/date-picker', () => {
     });
 
     describe('Keyboard navigation', () => {
-      describe('chaning the month back to current value month focuses the current value', () => {
-        test('after pressing the right chevron', async () => {
-          const { openMenu } = renderDatePicker({
+      describe('focuses the current value', () => {
+        test("when month returns to value's month", async () => {
+          const { openMenu, findAllByRole } = renderDatePicker({
             value: testToday,
           });
 
-          const { rightChevron, queryCellByDate } = await openMenu();
-
+          const { queryCellByDate, monthSelect } = await openMenu();
           expect(queryCellByDate(testToday)).toHaveFocus();
 
-          rightChevron?.focus();
-          userEvent.keyboard(`[Enter]`);
-          // cell, leftChevron
-          tabNTimes(2);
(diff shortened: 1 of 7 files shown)


packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1082,6 +1082,46 @@ describe('packages/date-picker', () => {
...
+
+          rightChevron?.focus();
+          userEvent.keyboard(`[Enter]`);
+          // cell, leftChevron
+          tabNTimes(2);
+          userEvent.keyboard(`[Enter]`);
+          // select, select, rightChevron, cell
+          tabNTimes(4);
+          expect(queryCellByDate(testToday)).toHaveFocus();
+        });
```
TheSonOfThomp, line 1103: Is there a concern that this won't work in one of these cases (L/R chevron)? Otherwise I'd say we test "when month returns to value's month", regardless of _how_ we return to that month

IMO this should be 
- (set value)
- Open menu
- Change month with select
- Change month back
- expect highlight to be value

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenu.tsx
```diff
@@ -92,8 +92,13 @@ export const DatePickerMenu = forwardRef<HTMLDivElement, DatePickerMenuProps>(
...
         return;
       }
+
+      const isSameUTCMonthAsCurrentValue = isSameUTCMonth(newMonth, value);
+
       setDisplayMonth(newMonth);
-      const newHighlight = getNewHighlight(highlight, month, newMonth);
+      const newHighlight = isSameUTCMonthAsCurrentValue
+        ? (value as Date)
+        : getNewHighlight(highlight, month, newMonth);
```
TheSonOfThomp, line 101: I'd say we should update `getNewHighlight` to take in `value` as an arg, and do this logic internally

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
shaneeza: I was being extra cautious but I can reduce this to one test

TheSonOfThomp (approved):

PR #2175, opened 2024-01-18
Table cell overflow [LG-3952]

TheSonOfThomp:
## ✍️ Proposed changes

Adds `overflow` prop to `Cell` component. By default there is no change. When `overflow === 'truncate'`, the styling of the cell is updated (if necessary) to be aligned to the top, with an ellipsis after 2 lines of text.

🎟 _Jira ticket:_ 
[LG-3952](https://jira.mongodb.org/browse/LG-3952)

Screenshot 2024-01-18 at 4 24 16 PM

### DIFF
--- a/packages/table/src/Cell/Cell.styles.ts
+++ b/packages/table/src/Cell/Cell.styles.ts
@@ -1,7 +1,7 @@
 import { TransitionStatus } from 'react-transition-group';
 
 import { css } from '@leafygreen-ui/emotion';
-import { spacing, transitionDuration } from '@leafygreen-ui/tokens';
+import { spacing, transitionDuration, typeScales } from '@leafygreen-ui/tokens';
 
 import { Align } from './Cell.types';
 
@@ -17,7 +17,6 @@ export const standardCellHeight = spacing[5] + spacing[2];
 export const baseCellStyles = css`
   padding: 0 8px;
   overflow: hidden;
-  text-overflow: ellipsis;
 
   &:focus-visible {
     box-shadow: inset;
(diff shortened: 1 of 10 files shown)


packages/table/src/Cell/Cell.styles.ts
```diff
@@ -75,35 +74,43 @@ export const basicCellStyles = css`
...
+    max-height: 0;
+    overflow: hidden;
+  `;
+
   return {
     entered: css`
       opacity: 1;
       min-height: ${standardCellHeight}px;
-      max-height: ${height}px;
+      max-height: ${height + 'px' ?? 'unset'};
```
bruugey, line 113: I think you mean `height ? `${height}px : 'unset'}`

bruugey: Im having trouble formatting that properly sorry

packages/table/src/Cell/Cell.types.ts
```diff
@@ -5,40 +5,63 @@ export type Align = Extract<
...
+   * Any overflowing content will be clipped.
+   *
+   * `Truncate`: The cell height will be fixed to the standard cell height (40px by default),
+   * and include an ellipsis before the content is clipped.
+   *
+   * Note: It's recommended to provide the same value for all cells in a given row.
+   *
+   * @default CellOverflowBehavior.Default
+   */
+  overflow?: CellOverflowBehavior;
```
bruugey, line 38: nit: `overflowBehavior` ? overflow makes me think this will be a boolean

TheSonOfThomp: I had that initially, but wanted to match the similarly-functional prop in Combobox 

https://github.com/mongodb/leafygreen-ui/blob/a19b08de770fe0e4e61aa3a09ea59bae4818c3c1/packages/combobox/src/Combobox.types.ts#L127

packages/table/src/Cell/Cell.styles.ts
TheSonOfThomp: Neither of us had it the first time 😆  

`height ? height + "px" : 'unset'`  is what I was going for

packages/table/src/Table.story.tsx
```diff
@@ -198,7 +199,7 @@ export const LiveExample: StoryFn<StoryTableProps> = args => {
             <Row key={row.id} row={row}>
               {row.getVisibleCells().map(cell => {
                 return (
-                  <Cell key={cell.id}>
+                  <Cell key={cell.id} id={cell.id} overflow="truncate">
```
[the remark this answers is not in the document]
TheSonOfThomp: Don't think we need a whole new story. I've set up the `makeData` function to make sure it generates at least one row of long content

PR #2178, opened 2024-01-22
LG-3839: Date picker switch selects

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3839](https://jira.mongodb.org/browse/LG-3839)

In this PR, I'm flipping the year and month selects based on the locale. To do this, I decided to break the selects into individual components so that I could conditionally render them.

Since I'm switching the HTML order of the selects, I had to refactor some of the tests that rely on tab order. 

### For new components


packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuHeader/DatePickerMenuHeader.tsx
```diff
@@ -101,12 +91,7 @@ export const DatePickerMenuHeader = forwardRef<
       }
     };
 
-  /** Returns whether the provided month should be enabled */
-  const isMonthEnabled = useCallback(
-    (monthName: string) =>
-      shouldMonthBeEnabled(monthName, { month, min, max, locale }),
-    [locale, max, min, month],
-  );
+  const isYearFirstFormat = yearFirstFormats.includes(locale);
```
TheSonOfThomp, line 94: I think checking the order of `formatParts` would be a more reliable way to solve this. It's _probably_ only ISO that formats year first, but I'm not 100% sure

TheSonOfThomp (commented):
Just have concerns around maintaining an array of `yearFirstFormats` when we can check the ordering of `formatParts` instead

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuHeader/DatePickerMenuHeader.tsx
shaneeza: So basically, check if the year comes first in `formatParts`? There are a few other locales that format year first. This brings me back to the convo about `fa-AF` and how we're unsure if flipping the selects makes sense in that language.

shaneeza: Since it's only iso that we went, I think i can just do a check `locale === 'iso8601'`

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuHeader/DatePickerMenuHeader.tsx
```diff
@@ -37,19 +30,17 @@ export const DatePickerMenuHeader = forwardRef<
...
-  const monthOptions = getLocaleMonths(locale);
-  const yearOptions = range(min.getUTCFullYear(), max.getUTCFullYear() + 1);
-
   const updateMonth = (newMonth: Date) => {
     // We don't do any checks here.
     // If the month is out of range, we still display it
     setMonth(newMonth);
   };
 
+  const isIsoFormat = locale === 'iso8601';
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah, that works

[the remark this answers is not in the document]
TheSonOfThomp: I don't think there are any other locales that put the year first

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuSelect/DatePickerMenuSelectMonth.tsx
```diff
@@ -43,7 +43,9 @@ export const DatePickerMenuSelectMonth = ({
   return (
     <Select
       {...selectElementProps}
-      aria-label="Select month"
+      aria-label={`Select month - ${
+        monthOptions[month.getUTCMonth()].long
+      } selected`}
```
TheSonOfThomp, line 48: This is probably a wider issue with Select, right?

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuSelect/DatePickerMenuSelectMonth.tsx
```diff
@@ -43,7 +43,9 @@ export const DatePickerMenuSelectMonth = ({
   return (
     <Select
       {...selectElementProps}
```
TheSonOfThomp, line 45: Will we need to spread `rest` here?

TheSonOfThomp (approved):

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuSelect/DatePickerMenuSelectMonth.tsx
shaneeza: Yeah, if you're using a select without a label and need to use aria-label then it becomes an issue.

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuHeader/DatePickerMenuHeader.tsx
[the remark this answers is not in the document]
shaneeza: There are a few -- `ja-JP` for example

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuSelect/DatePickerMenuSelectMonth.tsx
shaneeza: I can create a ticket for this

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuSelect/DatePickerMenuSelectMonth.tsx
shaneeza: yeah, i think we have to, its an object

PR #2179, opened 2024-01-23
LG-3930: Date Picker extend popover props

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-3930](https://jira.mongodb.org/browse/LG-3930)

### For new components

### DIFF
--- a/.changeset/stale-dryers-unite.md
+++ b/.changeset/stale-dryers-unite.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/date-picker': patch
+---
+
+Extends [Popover props](https://www.mongodb.design/component/popover/documentation/) but omits the following props: `usePortal`, `refEl`, `children`, `className`, `onClick`, and `active`. [LG-3930](https://jira.mongodb.org/browse/LG-3930)
--- a/.changeset/stale-dryers-unite.md
+++ b/.changeset/stale-dryers-unite.md
@@ -1,5 +1,5 @@
 ---
-'@leafygreen-ui/date-picker': patch
+'@leafygreen-ui/date-picker': minor
 ---
 
 Extends [Popover props](https://www.mongodb.design/component/popover/documentation/) but omits the following props: `usePortal`, `refEl`, `children`, `className`, `onClick`, and `active`. [LG-3930](https://jira.mongodb.org/browse/LG-3930)
--- a/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
+++ b/packages/date-picker/src/DatePicker/DatePicker.spec.tsx
@@ -3608,6 +3608,18 @@ describe('packages/date-picker', () => {
         portalClassName=""
         scrollContainer={{} as HTMLElement}
         portalContainer={{} as HTMLElement}
+        align="bottom"
+        justify="start"
+        spacing={10}
+        adjustOnMutation={true}
+        popoverZIndex={1}
+        onEnter={() => {}}
+        onEntering={() => {}}
+        onEntered={() => {}}
+        onExit={() => {}}
+        onExiting={() => {}}
+        onExited={() => {}}
+        contentClassName=""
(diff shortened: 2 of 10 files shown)


.changeset/stale-dryers-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/date-picker': patch
```
TheSonOfThomp, line 2: I think this is technically `minor` since it's adding new behavior

packages/date-picker/src/shared/context/SharedDatePickerContext.utils.ts
```diff
@@ -76,6 +101,21 @@ export const defaultSharedDatePickerContext: SharedDatePickerContextProps = {
...
+  spacing: 10,
+  adjustOnMutation: true,
+  popoverZIndex: 1,
+  onEnter: () => {},
+  onEntering: () => {},
+  onEntered: () => {},
+  onExit: () => {},
+  onExiting: () => {},
+  onExited: () => {},
+  contentClassName: '',
```
TheSonOfThomp, line 118: Do we need to add these to the datepicker provider, or can we use Popover provider to access these values?

packages/date-picker/src/shared/context/SharedDatePickerContext.utils.ts
```diff
@@ -13,16 +13,40 @@ import { BaseFontSize, Size } from '@leafygreen-ui/tokens';
...
+  'adjustOnMutation',
+  'popoverZIndex',
+  'onEnter',
+  'onEntering',
+  'onEntered',
+  'onExit',
+  'onExiting',
+  'onExited',
+  'contentClassName',
+];
```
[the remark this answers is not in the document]
shaneeza: that might a good idea, so we can reuse this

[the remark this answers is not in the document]
shaneeza: Maybe we can chat in person. I'm thinking about how this would make sense in `Popover`. This array of popover prop names is specific to DatePicker only, and its primary purpose is to filter out props using the `pickAndOmit` util. The list is too specific to live in `Popover`.  

My current thought process is whether we have an array of popover props available and whether we can filter them to what we need inside DatePicker.

packages/date-picker/src/shared/context/SharedDatePickerContext.utils.ts
shaneeza: Are you referring to this provider? I don't think we have defaults for `Popover` anywhere.

https://github.com/mongodb/leafygreen-ui/blob/e7d206e6c245b316b6ae8d2a419633612862deb3/packages/leafygreen-provider/src/PopoverContext/PopoverContext.tsx#L37-L53

shaneeza: I made these props optional, so I no longer have to add them here.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -408,6 +408,14 @@ describe('packages/date-picker', () => {
         expect(menuContainerEl).toBeInTheDocument();
       });
 
+      test('appends to the end of the DOM', async () => {
```
TheSonOfThomp, line 411: nit "appends _menu_ ..."

TheSonOfThomp (approved):

packages/date-picker/src/shared/context/SharedDatePickerContext.utils.ts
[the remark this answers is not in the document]
TheSonOfThomp: We _may_ need them elsewhere, but you're right, this is pretty specific to what we need for DatePicker

PR #2180, opened 2024-01-24
Combobox types cleanup

TheSonOfThomp:
## ✍️ Proposed changes

Cleans up Combobox type declaration file. Pre-work for https://jira.mongodb.org/browse/LG-3959.

No functional changes


bruugey (commented):
Needs a changeset

packages/combobox/src/types/Combobox.types.ts
bruugey, line 1: Not sure how I feel about the types directory, since everywhere else we co-locate types with it's respective component

TheSonOfThomp:
> Needs a changeset

Do we need a changeset even though there are no user-facing changes?

packages/combobox/src/types/Combobox.types.ts
TheSonOfThomp: Yeah, that's fair—I considered that. 

So we'd have
src/Combobox/Combobox.types
src/ComboboxOption/ComboboxOption.types

etc?

shaneeza:
> > Needs a changeset
> 
> Do we need a changeset even though there are no user-facing changes?

I think we should add one, we've been adding them each time we restructure a component. There's also a small chance this could break something.

packages/combobox/src/types/Combobox.types.ts
bruugey: I think so - that's more in line with what we do everywhere else

bruugey: We can have a root level .types file if many things are inheriting from one SST

shaneeza (approved):

PR #2186, opened 2024-01-26
LG-3896, LG-3898: Date Picker input selection

shaneeza:
## ✍️ Proposed changes

**_Note_**: Changes are explained in detail in the changeset

🎟 _Jira ticket:_ [LG-3896](https://jira.mongodb.org/browse/LG-3896)
🎟 _Jira ticket:_ [LG-3898](https://jira.mongodb.org/browse/LG-3898)

The issue in [3898](https://jira.mongodb.org/browse/LG-3898) is that when using the left and right arrows, the caret moves to the middle of the value instead of the first and last index and typing doesn't update the value. This could be fixed by [LG-3896](https://jira.mongodb.org/browse/LG-3896), which suggests removing the caret so we won't have caret issues. 

Removing the caret is tricky, but I think I found a good workaround. The caret will still exist and could end up in the middle of an input on click, but once you start typing, the input value will reset. Furthermore, I made some tweaks so that if you tab into an input or initially click on an input, the entire input will become selected, resetting the value once you start typing.

### For new components


packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.tsx
```diff
@@ -73,17 +73,21 @@ export const DatePickerInput = forwardRef<HTMLDivElement, DatePickerInputProps>(
      * Called when the input, or any of its children, is clicked.
      * Opens the menu and focuses the appropriate segment
      */
-    const handleInputClick: MouseEventHandler<HTMLElement> = e => {
+    const handleInputClick: MouseEventHandler<HTMLInputElement> = e => {
       if (!disabled) {
         openMenu(e);
         const { target } = e;
+        e.preventDefault();
```
TheSonOfThomp, line 80: Might be worth just commenting what `default` means in this case

packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.tsx
```diff
@@ -73,17 +73,21 @@ export const DatePickerInput = forwardRef<HTMLDivElement, DatePickerInputProps>(
...
         const segmentToFocus = getSegmentToFocus({
           target,
           formatParts,
           segmentRefs,
         });
 
+        // console.log('🐥 click 🐥');
+
         segmentToFocus?.focus();
+        segmentToFocus?.select();
```
TheSonOfThomp, line 90: Do we need both `.focus` and `.select`? (my assumption is that select automatically sets focus, but maybe not)

shaneeza: I think so, a bunch of test fail without `.focus`

.changeset/eight-beds-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/date-picker': minor
+---
+
+Changes the behavior of segments. Add more details.
```
TheSonOfThomp, line 5: todo

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1415,39 +1420,24 @@ describe('packages/date-picker', () => {
...
               });
               userEvent.click(monthInput);
               userEvent.keyboard('{arrowleft}');
-              expect(monthInput).toHaveFocus();
+              expect(yearInput).toHaveFocus();
             });
 
-            test('moves the cursor when the value starts with 0', () => {
-              const { monthInput } = renderDatePicker({});
+            test('focuses the previous segment when the value starts with 0', () => {
```
TheSonOfThomp, line 1432: don't think we need this test (or the following one) anymore. These are both covered by "...when the segment has a value"

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -3001,27 +2987,26 @@ describe('packages/date-picker', () => {
...
-          test('if the resulting value is not valid, keeps the input as-is', async () => {
+          test('if the resulting value is not valid, clears the input', async () => {
             const { monthInput } = renderDatePicker({});
-            userEvent.type(monthInput, '6');
-            await waitFor(() => expect(monthInput).toHaveValue('06'));
-            userEvent.type(monthInput, '9');
-            await waitFor(() => expect(monthInput).toHaveValue('06'));
+            userEvent.type(monthInput, '0');
+            userEvent.tab();
+            await waitFor(() => expect(monthInput).toHaveValue(''));
```
TheSonOfThomp, line 3009: I think this should test something like month `32` or something like that. Testing `0` is a bit confusing

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -3511,15 +3496,13 @@ describe('packages/date-picker', () => {
...
+            '2020-02- is not a valid date',
           );
 
           userEvent.type(dayInput, '{backspace}{backspace}');
           userEvent.tab();
           errorElement = queryByTestId('lg-form_field-error_message');
-          expect(errorElement).toHaveTextContent(
-            '2020-02- is not a valid date',
-          );
+          expect(errorElement).toHaveTextContent('2020-- is not a valid date');
```
TheSonOfThomp, line 3505: Not for this PR, but we should improve this error message

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -1383,16 +1383,21 @@ describe('packages/date-picker', () => {
         test('deletes any value in the input', () => {
```
TheSonOfThomp, line 1383: We should change the name of this test

packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.tsx
```diff
@@ -108,40 +109,38 @@ export const DatePickerInput = forwardRef<HTMLDivElement, DatePickerInputProps>(
 
       const isSegmentEmpty = !target.value;
 
-      const { selectionStart, selectionEnd } = target;
-
       switch (key) {
         case keyMap.ArrowLeft: {
+          // Prevent the cursor on tab and initial click
+          e.preventDefault();
```
TheSonOfThomp, line 115: Why would `e.preventDefault` in a keyboard handler prevent the cursor on a tab/click event?

packages/date-picker/src/shared/components/DateInput/DateInputBox/DateInputBox.spec.tsx
```diff
@@ -255,17 +255,17 @@ describe('packages/date-picker/shared/date-input-box', () => {
...
 
-      test('value setter is not called when deleting from a single segment', () => {
+      test('value setter is called when pressing backspace in a single segment', () => {
         const setValue = jest.fn();
 
         const { dayInput } = renderDateInputBox({ setValue }, testContext);
         userEvent.type(dayInput, '21');
         userEvent.type(dayInput, '{backspace}');
-        expect(setValue).not.toHaveBeenCalled();
+        expect(setValue).toHaveBeenCalled();
```
TheSonOfThomp, line 268: For completeness, we should test `toHaveBeenCalledWith(...)`

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.spec.tsx
```diff
@@ -204,39 +204,40 @@ describe('packages/date-picker/shared/date-input-segment', () => {
...
+      test('resets the value when the value is complete', () => {
         const { input } = renderSegment({
           value: '26',
           onChange: onChangeHandler,
         });
 
         userEvent.type(input, '6');
-        expect(onChangeHandler).not.toHaveBeenCalled();
+        expect(onChangeHandler).toHaveBeenCalledWith(
+          expect.objectContaining({ value: '6' }),
```
TheSonOfThomp, line 215: nit, since the initial value is `26` can we test with a value that's not 2 or 6 (for clarity)

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
```diff
@@ -95,10 +95,17 @@ export const DateInputSegment = React.forwardRef<
 
     /** Handle keydown presses that don't natively fire a change event */
     const handleKeyDown: KeyboardEventHandler<HTMLInputElement> = e => {
-      const { key } = e as React.KeyboardEvent<HTMLInputElement> & {
+      const { key, target } = e as React.KeyboardEvent<HTMLInputElement> & {
         target: HTMLInputElement;
       };
 
+      // if a number is pressed and the value length is equal to the charsPerSegment, reset the input
+      if (isFinite(Number(key))) {
```
TheSonOfThomp, line 103: Huh, I've never seen this check. Why do we need to check if the key is finite?

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
```diff
@@ -123,17 +130,23 @@ export const DateInputSegment = React.forwardRef<
...
-          if (numChars === 1) {
-            onChange({
-              segment,
-              value: '',
-              meta: { key },
-            });
+          const preVal = target.value;
+
+          // reset the input on backspace
+          target.value = '';
```
TheSonOfThomp, line 136: I don't know if we should manually be setting `target.value` here. The component is controlled, and the value should be cleared in response to the `onChange` event we're firing

TheSonOfThomp (changes requested):
A few nit changes, but overall I like the approach!

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
shaneeza: Can you elaborate a little more on this?

packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.tsx
shaneeza: I should reword this; without this, the input does not select all the text.

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
shaneeza: I'll change this. This should be a check for only numbers.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
shaneeza: I went with `0` to show that it's an invalid value; also, typing any other value auto formats the value to something valid.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
shaneeza: I can create a ticket for this

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
shaneeza: how about `resets the input`?

packages/date-picker/src/DatePicker.stories.tsx
```diff
@@ -112,7 +112,7 @@ export const LiveExample: StoryFn<typeof DatePicker> = props => {
         onDateChange={v => {
           // eslint-disable-next-line no-console
           console.log('Storybook: onDateChange', {
-            value: v!.toUTCString(),
+            value: v && v.toUTCString(),
```
TheSonOfThomp, line 115: `v?.toUTCString` could work too

packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.tsx
```diff
@@ -111,7 +111,7 @@ export const DatePickerInput = forwardRef<HTMLDivElement, DatePickerInputProps>(
 
       switch (key) {
         case keyMap.ArrowLeft: {
-          // Prevent the cursor on tab and initial click
+          // Without this, the input does not select all the text
```
TheSonOfThomp, line 114: I'm still a bit unclear on how preventing default keydown behavior prevents selection

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
TheSonOfThomp: But `32` is also an invalid month, so we should test that as well

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
TheSonOfThomp: is there a case where `key` is not a number? (I guess maybe it could be undefined?)

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
TheSonOfThomp: `DateInputSegment` is always a controlled component—i.e. it accepts `value` & `onChange` props. 
If we manually set `target.value` _without_ calling `onChange` internally, then the controlling component won't know value is updated and could have a stale `value` state. 
Even if we _do_ call `onChange` internally, whatever we set `target.value` to will be reset by the incoming `value` prop

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
shaneeza: Yeah, it could be one of the arrow keys, enter, space, etc...

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
shaneeza: Chatted in person -- we call `onChange` internally inside `DateInputSegment` and inside `DatePickerInput => handleSegmentChange` we update the value of the input using the up-to-date value from `segmentChangeEvent`

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
shaneeza: `32` won't clear the input though. It will format `3` to `03` then auto focus the next segment, populating that segment with `2`. But I can add a test to test this specific behavior.

.changeset/eight-beds-smash.md
```diff
@@ -0,0 +1,65 @@
+---
+'@leafygreen-ui/date-picker': major
+---
+
+Changes the behavior of segments.
+
+## New Behavior
```
shaneeza, line 7: @TheSonOfThomp, if you think there is a better way to organize this, please let me know!

packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.tsx
shaneeza: I'm unsure how to explain this -- without preventDefault, the input ignores `.select()`, and I'm unsure why.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
shaneeza: https://jira.mongodb.org/browse/LG-4028

.changeset/eight-beds-smash.md
TheSonOfThomp: I like this! Maybe move the old/new callouts to each section (i.e. Backspace: old / new  etc.) to make it easier to compare?

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -2632,72 +2605,76 @@ describe('packages/date-picker', () => {
...
                 const { yearInput } = renderDatePicker({
                   onChange,
-                  value: newUTC(2023, Month.December, 25),
                 });
                 userEvent.type(yearInput, '{space}{space}');
                 expect(onChange).not.toHaveBeenCalled();
               });
+            });
 
+            describe('does fire a segment value change', () => {
```
TheSonOfThomp, line 2619: nit "fires" instead of "does fire" (easier to scan)

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -3042,12 +3018,21 @@ describe('packages/date-picker', () => {
...
 
-          test('if the resulting value is not valid, clears the input', async () => {
+          test('if the resulting value is incomplete and not valid, clears the input', async () => {
             const { monthInput } = renderDatePicker({});
             userEvent.type(monthInput, '0');
             userEvent.tab();
             await waitFor(() => expect(monthInput).toHaveValue(''));
           });
+
+          test('if the resulting value is not valid, formats the value', async () => {
```
TheSonOfThomp, line 3028: I think this test name could be more descriptive (maybe call out that the 2nd digit is inputted into the next input)

packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.tsx
```diff
@@ -111,10 +114,10 @@ export const DatePickerInput = forwardRef<HTMLDivElement, DatePickerInputProps>(
 
       switch (key) {
         case keyMap.ArrowLeft: {
-          // Prevent the cursor on tab and initial click
+          // Without this, the input ignores `.select()`
```
TheSonOfThomp, line 117: why?

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.spec.tsx
```diff
@@ -210,34 +210,34 @@ describe('packages/date-picker/shared/date-input-segment', () => {
...
-      test('clears the input when there is a value', () => {
+      test('does not call the onChangeHandler when the value is initially empty', () => {
         const { input } = renderSegment({
-          value: '26',
           onChange: onChangeHandler,
         });
 
         userEvent.type(input, '{backspace}');
-        expect(onChangeHandler).toHaveBeenCalledWith(
+        expect(onChangeHandler).not.toHaveBeenCalledWith(
```
TheSonOfThomp, line 229: not sure we need `.not.toHaveBeenCalledWith`. Could just have `.not.toHaveBeenCalled`

packages/date-picker/src/shared/components/DateInput/DateInputSegment/DateInputSegment.tsx
```diff
@@ -99,8 +102,10 @@ export const DateInputSegment = React.forwardRef<
         target: HTMLInputElement;
       };
 
-      // if a number is pressed and the value length is equal to the charsPerSegment, reset the input
-      if (isFinite(Number(key))) {
+      // If the key is a number and the value length is equal to the charsPerSegment, reset the input.
+      // We also check for space because Number(' ') returns true
+      // We need this check because a key press could be arrow keys, enter, space, etc.
+      if (Number(key) && key !== keyMap.Space) {
```
TheSonOfThomp, line 108: nit: extract to a variable

TheSonOfThomp (approved):
LGTM. Left some nits, but looks good overall!

packages/date-picker/src/DatePicker/DatePickerInput/DatePickerInput.tsx
TheSonOfThomp: Hmm, If you have time, this would be an interesting thing to investigate why

PR #2193, opened 2024-01-29
LG-3876 Combobox input value

TheSonOfThomp:
## ✍️ Proposed changes

- Adds optional `inputValue` and `onInputChange` props to Combobox. These props are used to control the value of the inner text input (not the selected combobox value itself).

- `onChange` callback now fires when the input is blurred and the input contains a valid selection value.

🎟 _Jira ticket:_ 
- [onChange](https://jira.mongodb.org/browse/LG-3978)
- [inputValue](https://jira.mongodb.org/browse/LG-3876)


.changeset/fuzzy-spiders-hug.md
```diff
@@ -0,0 +1,7 @@
+---
+'@leafygreen-ui/combobox': minor
+---
+
+- Adds optional `inputValue` and `onInputChange` props to Combobox. These props are used to control the value of the inner text input (not the selected combobox value itself).
```
bruugey, line 5: I thought I left a comment on here, but I can't find it.
I'm not sure I'm aligned with the prop names... I feel like this doesn't communicate that this is what happens when you type into the input. Would it make more sense just to properly call a consuming app's onKeyDown handler in this case?

bruugey: Breaking change!

bruugey (approved):

PR #2201, opened 2024-01-31
LG-1636: Adds data-testids to confirmation modal

bruugey:
## ✍️ Proposed changes

Adds data-testids to confirmation modal

🎟 _Jira ticket:_ [LG-1636t](https://jira.mongodb.org/browse/LG-1636)

### For new components


packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -59,6 +59,7 @@ export const ConfirmationModal = React.forwardRef(
             // eslint-disable-next-line jsx-a11y/no-autofocus
             autoFocus
             darkMode={darkMode}
+            data-testid="lg-confirmation-modal-confirmation-input"
```
shaneeza, line 62: Is this the pattern we will use in all our components, `lg-[component-name]`?

bruugey: Do we have a different pattern elsewhere?

shaneeza: I remember @TheSonOfThomp having a proposal about naming our test IDs, but that was a while ago. I guess if it's the same everywhere else, then it's fine.

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -99,10 +104,15 @@ export const ConfirmationModal = React.forwardRef(
             disabled={!confirmEnabled || submitDisabled}
             onClick={onConfirm}
             className={buttonStyle}
+            data-testid="lg-confirmation-modal-footer-button"
```
shaneeza, line 107: What about `lg-confirmation-modal-footer-confirm-button` since the button below has `cancel-button`

bruugey: Updated... what do you think?

shaneeza (approved):

PR #2203, opened 2024-01-31
LG-3969 Descendants

TheSonOfThomp:
## ✍️ Proposed changes
 
Creates Descendants package

🎟 _Jira ticket:_ [Name of ticket](https://jira.mongodb.org/browse/LG-3969)

### For new components


packages/descendants/README.md
```diff
@@ -0,0 +1,91 @@
...
+    <div ref={ref} data-testid="leafygreen-item" data-index={index} {...rest}>
+      {children}
+    </div>
+  );
+};
+```
+
+## References
+
+This package heavily references the work of [`pacocoursey/use-descendants`](https://github.com/pacocoursey/use-descendants/tree/v0) and [`@reach-ui/descendants`](https://github.com/reach/reach-ui/tree/dev/packages/descendants). Many thanks to the authors of those packages.
```
bruugey, line 91: Think you should explain why this approach is better than those referenced, as this can be useful for other devs

TheSonOfThomp: Definitely. Working on that now

packages/descendants/src/Descendants.types.ts
```diff
@@ -0,0 +1,12 @@
+import { ComponentProps } from 'react';
+
+/** */
+export interface Descendant<E extends HTMLElement = HTMLElement> {
+  element: E; // Could also be a RefObject
+  id: string;
+  props?: ComponentProps<any>;
```
bruugey, line 7: nice

packages/descendants/src/Descendants.types.ts
```diff
@@ -0,0 +1,12 @@
+import { ComponentProps } from 'react';
+
+/** */
+export interface Descendant<E extends HTMLElement = HTMLElement> {
+  element: E; // Could also be a RefObject
```
bruugey, line 5: i don't have a preference on element or element.current, although element.current makes it feel more like an interactive element? but that might be my own bias

TheSonOfThomp: IMO there's not really a benefit to storing the `ref` here vs the element, and it just makes accessing the element a bit more verbose:

```
const firstEl  = descendants[0].element
// vs
const firstEl  = descendants[0].elementRef.current
```

packages/descendants/src/Descendants.types.ts
TheSonOfThomp: If I can get it to behave, I'd like to pass in the Props type here.

Problem with that though is you'd have to know that type ahead of time when you call `useInitDescendants` so the reducer state gets established with the correct type

packages/descendants/package.json
```diff
@@ -19,7 +19,7 @@
     "lodash": "^4.17.21"
   },
   "devDependencies": {
-    "@faker-js/faker": "^8.3.1",
+    "@faker-js/faker": "^8.0.0",
```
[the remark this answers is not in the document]
TheSonOfThomp: Yeah, that's why I rolled it back, so we'd only have the one version installed

packages/descendants/README.md
```diff
@@ -0,0 +1,200 @@
+---
+last_updated: 2024-02-01
```
shaneeza, line 2: I love how detailed this is

packages/descendants/README.md
```diff
@@ -0,0 +1,200 @@
...
+   * 3. Pass the context, descendants list and setter into the provider
+   *
+   * We need to pass the context value into the Provider
+   * in order for the Parent to have knowledge
+   * of its specific context
+   * (see fly-out menu example in step 1.)
+   */
+  return (
+    <DescendantsProvider
+      context={TestDescendantContext}
```
shaneeza, line 72: Shoud this be `MyDescendantsContext`?

packages/descendants/src/DescendantsReducer.ts
```diff
@@ -0,0 +1,136 @@
+import { ComponentProps, Reducer } from 'react';
+import isEqual from 'lodash/isEqual';
+import isUndefined from 'lodash/isUndefined';
+
+import { findDescendantIndexWithId } from './utils/findDescendantWithId';
```
shaneeza, line 5: not blocking: Is there an advantage of importing these individually over importing everything from one index file?

packages/descendants/src/DescendantsReducer.ts
```diff
@@ -0,0 +1,136 @@
...
+  DescendantsReducerAction<T>
+>;
+
+/**
+ *
+ * Establishes a state with a `descendants` list, and a `dispatch` function to modify the descendants list
+ *
+ * @param state
+ * @param action
+ * @returns
```
shaneeza, line 43: it looks like the return description is missing?

packages/descendants/src/utils/findDescendantWithId.ts
```diff
@@ -0,0 +1,17 @@
+import { DescendantsList } from '../Descendants.types';
+
+/**
+ *
+ * @param descendants
+ * @param id
+ * @returns The index of a descendant with the given id
+ */
+export function findDescendantIndexWithId(
```
shaneeza, line 9: nit: should the file name and function name match?

packages/descendants/src/utils/findDOMIndex.ts
```diff
@@ -0,0 +1,51 @@
+import { DescendantsList } from '../Descendants.types';
+
+import { isElementPreceding } from './isElementPreceding';
+
+/**
+ * Returns the 1 + the index of the registered descendant immediately preceding the search element.
```
shaneeza, line 6: Q for tomorrow: I understand that this util is returning the index for the unregistered element, but I'm not sure i understand how this is doing that.

packages/descendants/src/useDescendant.tsx
```diff
@@ -0,0 +1,80 @@
...
+      dispatch({
+        type: 'remove',
+        id: _id,
+      });
+    };
+  }, [dispatch, ref]);
+
+  // When the props change, update them in the descendants list so the parent has access.
+  // We don't initially register the descendant with props,
+  // so we don't have to remove & re-register it each time the props change
```
shaneeza, line 66: Q for tomorrow: can you elaborate a little more on this line about removing and re-registering

packages/descendants/src/DescendantsReducer.ts
TheSonOfThomp: Not really, no. I just didn't make an index file 🙃  Will fix

packages/descendants/src/utils/findDOMIndex.ts
TheSonOfThomp: Updated the comment to be a bit clearer. Essentially we  crawl the DOM to find the index of the element that's right before our search `element`, then return the index _after_ that (i.e. the element after the element before the search element, is the search element)

packages/descendants/src/useDescendant.tsx
TheSonOfThomp: The alternative to this second `useEffect` would be to do everything inside the same hook. _But_ in doing this, the hook is dependent on too many things, and so we'd be calling `dispatch("register", ...)` and `dispatch("remove", ...)` every time the props change

packages/descendants/README.md
```diff
@@ -0,0 +1,200 @@
...
+  children,
+  ...rest
+}: ComponentProps<'div'>) => {
+  /**
+   * 4. Establish a child component as a descendant
+   *
+   * Pass the context value into the hook
+   * in order to establish this element
+   */
+  const { index, ref } = useDescendant(TestDescendantContext);
```
bruugey, line 91: should this be the same context as used in the Parent? (I'd assume so) If so, let's use the same name

packages/descendants/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+export const MyParent = ({ children, ...rest }: ComponentProps<'div'>) => {
+  /**
+   * 2. Initialize an empty descendants list and setter
+   *
+   * We call this _outside_ the Provider
+   * so we can access the `descendants` object
+   * from the Parent level.
+   */
+  const { descendants, dispatch } = useInitDescendants<HTMLDivElement>();
```
bruugey, line 60: what's the dispatch doing here? do we need to de-structure just to pass it back in? can we do that for the user automatically? when would a consumer want to use this directly?

packages/descendants/src/DescendantProvider.tsx
```diff
@@ -0,0 +1,43 @@
...
+ */
+export const DescendantsProvider = <T extends HTMLElement>({
+  context,
+  children,
+  descendants,
+  dispatch,
+}: PropsWithChildren<DescendantsProviderProps<T>>) => {
+  const Provider = context.Provider;
+
+  // We could init the descendant values here,
```
bruugey, line 30: DO we need to leave this in?

packages/descendants/src/Descendants.spec.tsx
```diff
@@ -0,0 +1,214 @@
...
+          <TestDescendant type="fruit">Banana</TestDescendant>
+          <TestDescendant type="vegetable">Carrot</TestDescendant>
+        </DescendantsProvider>,
+      );
+
+      const appleDescendant = result.current.descendants[0];
+      expect(appleDescendant.props?.type).toEqual('fruit');
+    });
+  });
+});
```
bruugey, line 214: great test suite

packages/descendants/src/Descendants.types.ts
bruugey: happy to leave as-is

packages/descendants/src/Descendants.types.ts
bruugey: i think typing this as ComponentProps<any> is ok

packages/descendants/src/DescendantsContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+export type RegisterDescendantFn<T extends HTMLElement> = (
+  element?: T | null,
+) => () => void;
+
+export interface DescendantsContextProps<T extends HTMLElement> {
+  descendants: DescendantsList<T>;
+  dispatch: Dispatch<DescendantsReducerAction<T>>;
+}
+
+export type DescendantContextType<T extends HTMLElement> = React.Context<
```
bruugey, line 15: do these interfaces belong in .types.ts

packages/descendants/src/DescendantsContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * The returned Context object will be used to define which context a descendant is a part of.
+ */
+export const createDescendantsContext = <T extends HTMLElement = HTMLElement>(
+  displayName: string,
+): DescendantContextType<T> => {
+  const context = createContext<DescendantsContextProps<T>>({
+    descendants: [],
+    dispatch: () => {},
+  });
+  context.displayName = displayName;
```
bruugey, line 30: is this a React pattern?

packages/descendants/src/DescendantsReducer.ts
```diff
@@ -0,0 +1,138 @@
...
+} from './utils/';
+import { Descendant, DescendantsList } from './Descendants.types';
+
+export interface DescendantsState<T extends HTMLElement> {
+  descendants: DescendantsList<T>;
+}
+
+export type DescendantsReducerAction<T extends HTMLElement> =
+  | {
+      type: 'register';
```
bruugey, line 19: do we need a read action?

packages/descendants/src/Descendants.spec.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    expect(apple).toHaveAttribute('data-index', '0');
+    expect(banana).toHaveAttribute('data-index', '1');
+    expect(peppers).toHaveAttribute('data-index', '2');
+    expect(anaheim).toHaveAttribute('data-index', '0');
+    expect(habanero).toHaveAttribute('data-index', '1');
+  });
+
+  describe('internal', () => {
+    test('descendants object has access to child props', () => {
```
bruugey, line 195: can we add an internal test that it responds to changes in descendant props?

packages/descendants/src/useDescendant.tsx
```diff
@@ -0,0 +1,80 @@
...
+import {
+  useForwardedRef,
+  useIsomorphicLayoutEffect,
+} from '@leafygreen-ui/hooks';
+
+import { DescendantContextType } from './DescendantsContext';
+import { findDescendantIndexWithId } from './utils';
+
+/** @deprecated */
+const genId = () => `_${Math.random().toString(36).substr(2, 9)}`;
```
bruugey, line 19: why deprecated? should we use our idAllocator here?

packages/descendants/src/useDescendant.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Registers a component as a descendant of a given context, and calculates its index within the rendered descendant list.
+ */
+export const useDescendant = <T extends HTMLElement>(
+  context: DescendantContextType<T>,
+  fwdRef?: RefObject<T> | ForwardedRef<T>,
+  props?: ComponentProps<any>,
+): UseDescendantReturnObject<T> => {
+  const ref: React.RefObject<T> = useForwardedRef(fwdRef ?? null, null);
```
bruugey, line 35: does this hook actually work? i feel like i've seen it not :/

packages/descendants/src/useDescendant.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const { descendants, dispatch } = useContext(context);
+  const id = useRef(genId());
+
+  // Find the element with this id in the descendants list
+  const index = useMemo(() => {
+    return findDescendantIndexWithId(descendants, id.current);
+  }, [descendants]);
+
+  // On render, register the element as a descendant
+  useIsomorphicLayoutEffect(() => {
```
bruugey, line 45: if dispatch or ref changes do we need to change the type of dispatch to 'update'?

bruugey: (i.e. should we do a lookup before creating?)

packages/descendants/src/useDescendant.tsx
```diff
@@ -0,0 +1,80 @@
...
+    dispatch({
+      type: 'update',
+      id: id.current,
+      props,
+    });
+  }, [dispatch, props]);
+
+  return {
+    ref,
+    index: index,
```
bruugey, line 77: nit: index

packages/descendants/src/useInitDescendants.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/**
+ * Initializes a {@link DescendantsList}
+ */
+export const useInitDescendants = <
+  T extends HTMLElement,
+>(): InitDescendantsReturnObject<T> => {
+  const [state, dispatch] = useReducer<
+    React.Reducer<DescendantsState<T>, DescendantsReducerAction<T>>
```
bruugey, line 23: nit but since you're destructuring everything esle from React, maybe destructure Reducer as well

packages/descendants/README.md
TheSonOfThomp: So that's the unfortunate part about this approach. If we design it so that you initialize the descendants object _outside_ the provider, you need a way to pass the dispatch function into the provider.

You could technically write something like this, but IMO it's not much better (and misleading, since `rest` only refers to one property)

```tsx
const { descendants, ...rest } = useInitDescendants<HTMLDivElement>();

<DescendantsProvider
  context={MyDescendantsContext}
  descendants={descendants}
  {...rest}
>

```

packages/descendants/README.md
TheSonOfThomp: Good catch

packages/descendants/src/DescendantProvider.tsx
TheSonOfThomp: No, this should just be a GH comment

packages/descendants/src/DescendantsContext.tsx
TheSonOfThomp: Potentially... I was imagining `.types.ts` would have shared types, and files like this would have types specific to the use case. 
(If these were more complex files, they should each have their own folder with a `Context.types.ts`, `Context.spec.ts` etc.)

packages/descendants/src/DescendantsContext.tsx
TheSonOfThomp: Which part? Adding the dynamic `displayName`, or the factory function?

packages/descendants/src/DescendantsReducer.ts
TheSonOfThomp: I don't think so. You can just read off of `state.descendants`

TheSonOfThomp: I don't think so. You can just read off of `state.descendants`

packages/descendants/src/useDescendant.tsx
TheSonOfThomp: Ah, I wanted to add a bit of a more robust id generator, but this _should_ be fine.
We may want to add some collision-prevention logic

packages/descendants/src/useDescendant.tsx
TheSonOfThomp: Heh, yeah I'm not sure to be honest

packages/descendants/src/useDescendant.tsx
TheSonOfThomp: When wither of those props change (which they likely won't, but still), the cleanup function will be run, which calls `"remove"`, and then we re-register it

TheSonOfThomp: We also do [a check within the reducer](https://github.com/mongodb/leafygreen-ui/blob/a26750588462d711f81222269708dd24e8da4251/packages/descendants/src/DescendantsReducer.ts#L65) to make sure we're not duplicating

bruugey:
Pref without useIdAllocator

PR #2219, opened 2024-02-13
LG-2342: Description props accept `React.ReactNode` instead of strings

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-2342](https://jira.mongodb.org/browse/LG-2342)

### For new components


packages/select/src/Select/Select.types.ts
```diff
@@ -41,6 +43,11 @@ export interface BaseSelectProps
    * HTML `id` property used to allow Javascript, form, or label to reference the input.
    */
   id?: string;
+
+  /**
+   * Determines whether or not the component appears in dark theme.
+   * @default false
+   */
   darkMode?: boolean;
```
shaneeza, line 51: not blocking: if you want you can extend `DarkModeProps` here

packages/combobox/README.md
```diff
@@ -87,7 +87,7 @@ import { Combobox, ComboboxOption } from '@leafygreen-ui/combobox';
 | `overflow`               | `'expand-y'`, `'expand-x'`, `'scroll-x'`      | Defines the overflow behavior of a multiselect combobox. **expand-y**: Combobox has fixed width, and additional selections will cause the element to grow in the block direction. **expand-x**: Combobox has fixed height, and additional selections will cause the elemenet to grow in the inline direction. **scroll-x**: Combobox has fixed height and width, and additional selections will cause the element to be scrollable in the x (horizontal) direction. | `'expand-y'`             |
 | `label`                  | `string`                                      | An accessible label for the input, rendered in a `<label>` to the DOM                                                                                                                                                                                                                                                                                                                                                                                               |                          |
```
shaneeza, line 88: Should the same be done for `label`?

bruugey: Potentially - this ticket only asked for Description. But WDYT?

shaneeza: Let's create a ticket for the labels to track it. And if it's not too much, we can get it done in this PR

shaneeza (approved):

packages/combobox/README.md
bruugey: Will create a ticket - I think this will actually bea good thing to tackle as a part of CCP.

PR #2221, opened 2024-02-13
Adds `testid` guidelines to Styleguide

TheSonOfThomp:
## ✍️ Proposed changes

Adds `testid` guidelines to Styleguide

Discussed in code review here: https://docs.google.com/document/d/1Vxv2XvPXn5ozAtKSbRQcbC_HgKUBFcn0so6Sm1z28HA/edit#heading=h.lc5w3k6ydn8c


STYLEGUIDE.md
```diff
@@ -190,7 +190,22 @@ Keeps functions pure.
...
+### Follow BEM-ish patterns when hard-coding a `data-testid` or `data-lgid`
+
+1. Prefix with “lg”
+2. Words are separated by underscore `_`
+3. Blocks are separated by dash `-`
+4. Name hierarchy should somewhat match the directory structure
+
+e.g. `data-testid="lg-date_picker-calendar_cell"`
+
+Additionally, a hard-coded `data` attribute should _only_ be placed on a root-level, native HTML element. Components should have their own internal id.
```
bruugey, line 208: we should have a `prefer/not` section

TheSonOfThomp: What do you mean by that?

TheSonOfThomp: As in "not `lg-date-picker-calendar-cell` (with dashes)" ?

STYLEGUIDE.md
```diff
@@ -190,7 +190,22 @@ Keeps functions pure.
...
+---
+
+### Follow BEM-ish patterns when hard-coding a `data-testid` or `data-lgid`
+
+1. Prefix with “lg”
+2. Words are separated by underscore `_`
+3. Blocks are separated by dash `-`
+4. Name hierarchy should somewhat match the directory structure
+
+e.g. `data-testid="lg-date_picker-calendar_cell"`
```
shaneeza, line 206: Should this be `lg-date-picker_calendar-cell`? date-picker is the block

STYLEGUIDE.md
bruugey: https://github.com/mongodb/leafygreen-ui/pull/2221/files/4638ff2c841a9bd9ff9716b030ecaaae90515448#diff-18b19378fcb4038a2ce14f3ef8dd8ffa0bacd8f4e795cef51d08f3178d7e946cL20-L46

bruugey: take a look at our "Contributing to the style guide" guidelines

STYLEGUIDE.md
TheSonOfThomp: No, since "date_picker" is one concept

STYLEGUIDE.md
TheSonOfThomp: I'm a bit confused. How would you write it?

bruugey: Prefer 
this_pattern-pattern

Avoid
this-pattern-pattern

STYLEGUIDE.md
shaneeza: Whats the diff between a block and a concept? I know that we are not fully following BEM but from [docs](https://en.bem.info/methodology/quick-start/#element) it looks like it should be `date-picker` because that's the block.

TheSonOfThomp: BEM uses dashes to separate words within a block/element, and a double underscore to separate blocks/elements/modifiers.

This proposal is to use a single underscore to separate words, and a dash to separate concepts/elements/blocks 

My main gripe with this syntax is that in most text editors a Double click or `Option`+`ArrowKey` presses treat underscores as _one works_ and dashes as a separator. So If I want to replace the `"calendar_cell"` part I can double click it and paste. Or if I want to skip from the end of that string to the previous element I can just press `Option` + `ArrowLeft`. All this is backwards with standard BEM

bruugey (approved):

STYLEGUIDE.md
shaneeza: Can you add this bit to the styleguide? `...use a single underscore to separate words, and a dash to separate concepts/elements/blocks`

shaneeza: oh nvm its there!

PR #2224, opened 2024-02-21
Tokens implementation demo

bruugey:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/packages/tokens/src/borderRadius.ts
+++ b/packages/tokens/src/borderRadius.ts
@@ -1,0 +1,13 @@
+const borderRadius = {
+  0: 0,
+  50: 2,
+  100: 4,
+  150: 6,
+  200: 8,
+  300: 12,
+  400: 16,
+  500: 20,
+  600: 24,
+} as const;
+
+export default borderRadius;
--- a/packages/tokens/src/colors.ts
+++ b/packages/tokens/src/colors.ts
@@ -1,0 +1,110 @@
+import { palette } from '@leafygreen-ui/palette';
+
+const { black, blue, gray, green, red, white, yellow } = palette;
+
+const colors = {
+  accent: {
+    dark: {
+      primary: black,
+      secondary: gray.dark4,
+      inverse: gray.light2,
+      info: blue.dark3,
+      warning: yellow.dark3,
+      success: green.dark3,
+      error: red.dark3,
+      disabled: gray.dark3,
+    },
+    light: {
+      primary: gray.light3,
+      secondary: gray.light3,
+      inverse: black,
+      info: blue.light3,
+      warning: yellow.light3,
+      success: green.light3,
+      error: red.light3,
+      disabled: gray.light2,
+    },
+  },
+
+  bg: {
+    dark: {
+      primary: black,
+      secondary: gray.dark4,
+      inverse: gray.light2,
+      disabled: gray.dark2,
+      focus: blue.light3,
+      hover: gray.dark2,
+    },
+    light: {
+      primary: white,
+      secondary: gray.light3,
+      inverse: black,
+      disabled: gray.light2,
+      focus: blue.dark1,
+      hover: gray.light2,
+    },
+  },
+
+  border: {
+    dark: {
+      primary: gray.base,
+      secondary: gray.dark2,
+      success: green.dark1,
+      error: red.light1,
+      disabled: gray.dark2,
+    },
+    light: {
+      primary: gray.base,
+      secondary: gray.light2,
+      success: green.dark1,
+      error: red.base,
+      disabled: gray.light1,
+    },
+  },
+
+  icon: {
+    dark: {
+      primary: gray.light1,
+      secondary: gray.base,
+      inverse: white,
+      info: blue.light1,
+      warning: yellow.base,
+      success: green.base,
+      error: red.light1,
+      disabled: gray.dark1,
+    },
+    light: {
+      primary: gray.dark1,
+      secondary: gray.base,
+      inverse: white,
+      info: blue.base,
+      warning: yellow.dark2,
+      success: green.dark1,
+      error: red.base,
+      disabled: gray.base,
+    },
+  },
+
+  text: {
+    dark: {
+      primary: gray.light2,
+      secondary: gray.light1,
+      inverse: black,
+      error: red.light1,
+      disabled: gray.dark1,
+      focus: blue.light3,
+      hover: white,
+    },
+    light: {
+      primary: black,
+      secondary: gray.dark1,
+      inverse: white,
+      error: red.base,
+      disabled: gray.base,
+      focus: blue.dark1,
+      hover: black,
+    },
+  },
+} as const;
+
+export default colors;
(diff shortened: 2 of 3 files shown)


packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,110 @@
+import { palette } from '@leafygreen-ui/palette';
+
+const { black, blue, gray, green, red, white, yellow } = palette;
+
+const colors = {
```
bruugey, line 5: make singular!

packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,110 @@
...
+      disabled: gray.dark2,
+      focus: blue.light3,
+      hover: gray.dark2,
+    },
+    light: {
+      primary: white,
+      secondary: gray.light3,
+      inverse: black,
+      disabled: gray.light2,
+      focus: blue.dark1,
```
bruugey, line 43: color.dark.bg.primary.default //

alvenw (approved):
awesome! looks good

packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,300 @@
+import { palette } from '@leafygreen-ui/palette';
+
+import { Mode } from './mode';
+
+const { black, blue, gray, green, red, white, yellow } = palette;
+
+const color = {
+  [Mode.Dark]: {
+    bg: {
```
TheSonOfThomp, line 9: I know it's kindof a long word, but I think this should be `background`. We'll get the autocomplete either way, and the full word makes it much clearer

packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,300 @@
...
+      },
+      disabled: {
+        default: gray.dark1,
+        hover: gray.dark1,
+        focus: gray.dark1,
+      },
+    },
+  },
+  [Mode.Light]: {
+    bg: {
```
stephl3, line 154: this block sorts differently from the others. can we make them consistently `default`, `hover`, `focus`?

packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,300 @@
+import { palette } from '@leafygreen-ui/palette';
+
+import { Mode } from './mode';
+
+const { black, blue, gray, green, red, white, yellow } = palette;
+
+const color = {
+  [Mode.Dark]: {
```
TheSonOfThomp, line 8: It could be easier to read/maintain if we break out at least each dark/light mode object, or even each component level object.

e.g.
```
const darkModeBackgroundColors = {...}

const darkModeColors = {
  background: darkModeBackgroundColors,
  ...
}

const colors = {
  dark: darkModeColors
}
```

packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,300 @@
...
+        hover: gray.light3,
+        focus: blue.light3,
+      },
+      secondary: {
+        default: gray.base,
+        hover: gray.light3,
+        focus: blue.light3,
+      },
+      inverse: {
+        default: white,
```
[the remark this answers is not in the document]
TheSonOfThomp: `colors.dark.icon.inverse` should be dark, yeah.
(normally an icon on a dark bg should be light, but this is inverted)

packages/tokens/src/colors.ts
TheSonOfThomp: (could even have multiple files if that makes it easier)

packages/tokens/src/colors.ts
```diff
@@ -0,0 +1,300 @@
...
+const { black, blue, gray, green, red, white, yellow } = palette;
+
+const color = {
+  [Mode.Dark]: {
+    bg: {
+      primary: {
+        default: black,
+        hover: gray.dark2,
+        focus: blue.dark3,
+      },
```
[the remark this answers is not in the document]
shaneeza: What do you mean by your last comment?

packages/tokens/src/colors.ts
bruugey: Misread this at first - this seems like a good idea. I'll update.

packages/tokens/src/borderRadius.ts
```diff
@@ -0,0 +1,13 @@
+const borderRadius = {
```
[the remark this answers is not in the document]
bruugey: They're scaled so idt they will and we have diff values defined in each (but they'll always match if they overlap)

packages/tokens/src/colors.ts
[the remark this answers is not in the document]
bruugey: Makes sense

packages/tokens/src/colors.ts
bruugey: that's fine i'll update

packages/tokens/src/colors.ts
bruugey: yes updated

packages/tokens/src/colors.ts
bruugey: I think a file for each would be overkill. Maybe a file for darkmode and for light mode?

packages/tokens/src/color/color.types.ts
```diff
@@ -0,0 +1,37 @@
...
+  Disabled: 'disabled',
+} as const;
+
+type Variant = (typeof Variant)[keyof typeof Variant];
+
+export { State, Type, Variant };
+
+export type MapVariantsToStates = {
+  [k in Variant]?: { [k in State]: string };
+};
```
TheSonOfThomp, line 37: I think this approach is fine 👍 Not sure we need to be super explicit here with which variants have what states

TheSonOfThomp: Technically could be `Record<Variant, Record<State, string>>`

bruugey: Kk - are you aligned to merge?

bruugey: It can't be that because its a patial of Variant, and Record<Variant was complaining about missing keys

bruugey: and `Partial<Variant` didn't work either

packages/tokens/src/color/color.types.ts
```diff
@@ -0,0 +1,37 @@
...
+  Icon: 'icon',
+  Text: 'text',
+} as const;
+
+type Type = (typeof Type)[keyof typeof Type];
+
+const Variant = {
+  Primary: 'primary',
+  Secondary: 'secondary',
+  Inverse: 'inverse',
```
bruugey, line 21: nit: should we rm inverse in favor of always using inversePrimary?

packages/tokens/src/color/darkModeColors.ts
```diff
@@ -0,0 +1,159 @@
...
+    [State.Default]: gray.base,
+    [State.Hover]: gray.base,
+    [State.Focus]: blue.light1,
+  },
+  [Variant.Secondary]: {
+    [State.Default]: gray.dark2,
+    [State.Hover]: gray.dark2,
+    [State.Focus]: blue.light1,
+  },
+  success: {
```
stephl3, line 61: some of the keys in this object and the objects below need to get swapped to the enum

bruugey: ah! thank you!

stephl3 (approved):

PR #2225, opened 2024-02-23
LG-3940: NumberInput onBlur runs with arrow buttons

stephl3:
## ✍️ Proposed changes

In the `NumberInput` component, the `onBlur` callback was not being invoked when the arrow buttons were clicked. The `onBlur` was being passed in with the rest of the props directly to the input, and this change will now pass it to both the input and the arrows.

🎟 _Jira ticket:_ [LG-3940](https://jira.mongodb.org/browse/LG-3940)
  _Linked to_: [BAAS-27278](https://jira.mongodb.org/browse/BAAS-27278) 

## 🧪 How to test changes

I added storybook changes, so this can be tested by adding/removing focus on the arrow buttons and reviewing the console log entries.

https://github.com/mongodb/leafygreen-ui/assets/52300529/b3050fb6-565d-4c4d-bf4c-1781f969d999

### DIFF
--- a/.changeset/short-games-double.md
+++ b/.changeset/short-games-double.md
@@ -2,4 +2,5 @@
 '@leafygreen-ui/number-input': patch
 ---
 
-onBlur runs with arrow buttons
+Fixes a bug where onBlur would not get invoked when arrow buttons were blurred.
+[LG-3940](https://jira.mongodb.org/browse/LG-3940)
--- a/.changeset/short-games-double.md
+++ b/.changeset/short-games-double.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/number-input': patch
+---
+
+onBlur runs with arrow buttons
--- a/packages/number-input/src/Arrows/Arrow.tsx
+++ b/packages/number-input/src/Arrows/Arrow.tsx
@@ -21,6 +21,7 @@ export const Arrow = ({
   disabled,
   onClick,
   onKeyDown,
+  onBlur,
   direction,
 }: ArrowProps) => {
   const { theme } = useDarkMode();
@@ -30,6 +31,7 @@ export const Arrow = ({
       aria-label={`${direction} number`}
       onClick={() => onClick(direction)}
       onKeyDown={onKeyDown}
+      onBlur={onBlur}
       className={cx(arrowBaseStyles, arrowThemeStyles[theme])}
       type="button"
       tabIndex={-1} // Mimicking native behavior; you cannot focus on an arrow.
--- a/packages/number-input/src/Arrows/Arrow.tsx
+++ b/packages/number-input/src/Arrows/Arrow.tsx
@@ -21,7 +21,6 @@ export const Arrow = ({
   disabled,
   onClick,
   onKeyDown,
-  onBlur,
   direction,
 }: ArrowProps) => {
   const { theme } = useDarkMode();
@@ -31,7 +30,6 @@ export const Arrow = ({
       aria-label={`${direction} number`}
       onClick={() => onClick(direction)}
       onKeyDown={onKeyDown}
-      onBlur={onBlur}
       className={cx(arrowBaseStyles, arrowThemeStyles[theme])}
       type="button"
       tabIndex={-1} // Mimicking native behavior; you cannot focus on an arrow.
--- a/packages/number-input/src/Arrows/Arrow.tsx
+++ b/packages/number-input/src/Arrows/Arrow.tsx
@@ -34,7 +34,7 @@ export const Arrow = ({
       type="button"
       tabIndex={-1} // Mimicking native behavior; you cannot focus on an arrow.
       disabled={disabled}
-      data-testid={`lg-number_input-${direction}-button`}
+      data-testid={`lg-number_input-${direction}_button`}
     >
       <Icon
         className={cx({
--- a/packages/number-input/src/Arrows/Arrow.tsx
+++ b/packages/number-input/src/Arrows/Arrow.tsx
@@ -34,6 +34,7 @@ export const Arrow = ({
       type="button"
       tabIndex={-1} // Mimicking native behavior; you cannot focus on an arrow.
       disabled={disabled}
+      data-testid={`lg-number_input-${direction}-button`}
     >
       <Icon
         className={cx({
--- a/packages/number-input/src/Input/Input.tsx
+++ b/packages/number-input/src/Input/Input.tsx
@@ -43,6 +43,7 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
     {
       value: valueProp,
       onChange: onChangeProp,
+      onBlur,
       disabled = false,
       size = Size.Default,
       state = State.None,
@@ -162,12 +163,12 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
       }
     };
 
-    const handleOnFocus = () => {
+    const handleFocusContainer = () => {
       isFocusedRef.current = true;
       handleSetErrorTransition();
     };
 
-    const handleOnBlur = () => {
+    const handleBlurContainer = () => {
       isFocusedRef.current = false;
       handleRemoveErrorTransition();
     };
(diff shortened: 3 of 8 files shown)


packages/number-input/src/Input/Input.tsx
```diff
@@ -162,23 +163,23 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
     };
 
-    const handleOnFocus = () => {
+    const handleFocusContainer = () => {
       isFocusedRef.current = true;
       handleSetErrorTransition();
     };
 
-    const handleOnBlur = () => {
+    const handleBlurContainer = () => {
```
stephl3, line 171: made these more explicit to differentiate with `onBlur` prop passed into the api

.changeset/short-games-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/number-input': patch
+---
+
+onBlur runs with arrow buttons
```
TheSonOfThomp, line 5: Can we be more explicit with the change notes? What was the problem before? How does this impact consumers? 
Also, we've been starting to add the JIRA ticket into the change notes as well

packages/number-input/src/Input/Input.tsx
TheSonOfThomp: nitpick: should these be `handleContainerBlur/Focus` instead?

packages/number-input/src/Input/Input.tsx
```diff
@@ -162,23 +163,23 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
       }
     };
 
-    const handleOnFocus = () => {
+    const handleFocusContainer = () => {
```
TheSonOfThomp, line 166: Could we resolve this by listening to the focus/blur event that bubbles up from the input or button, instead of adding a listener to each element?

TheSonOfThomp (changes requested):

packages/number-input/src/Input/Input.tsx
stephl3: thanks for this feedback, I like this approach better

.changeset/short-games-double.md
stephl3: updated!

packages/number-input/src/Input/Input.tsx
```diff
@@ -162,23 +164,26 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
+    const handleFocus = () => {
       isFocusedRef.current = true;
       handleSetErrorTransition();
     };
 
-    const handleOnBlur = () => {
+    const handleBlur = (e: FocusEvent<HTMLDivElement>) => {
       isFocusedRef.current = false;
       handleRemoveErrorTransition();
+      if (onBlur) {
```
shaneeza, line 175: I think we can shorten this to `onBlur?.(e);`

packages/number-input/src/Input/Input.tsx
```diff
@@ -162,23 +164,26 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
       <div
         ref={containerRef}
-        onMouseEnter={() => handleSetErrorTransition()}
-        onMouseLeave={() => handleRemoveErrorTransition()}
-        onFocus={() => handleOnFocus()}
-        onBlur={() => handleOnBlur()}
+        onMouseEnter={handleSetErrorTransition}
+        onMouseLeave={handleRemoveErrorTransition}
+        onFocus={handleFocus}
+        onBlur={e => handleBlur(e)}
```
shaneeza, line 186: I think we can get away with just `onBlur={handleBlur}`

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -243,6 +230,39 @@ describe('packages/number-input', () => {
...
+          label,
+          ...defaultProps,
+        });
+
+        const upArrow = container.querySelector(
+          'button[aria-label="decrement number"]',
+        );
+
+        userEvent.click(upArrow as HTMLButtonElement); // focus
+        expect(defaultProps.onBlur).toHaveBeenCalledTimes(1);
```
shaneeza, line 258: Why is `onBlur` triggered when clicking on the `upArrow`? Was the div in focus prior to the click?

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -243,6 +230,39 @@ describe('packages/number-input', () => {
...
+      });
+
+      test('callback triggers when focus leaves arrow buttons', () => {
+        const { container, numberInput } = renderNumberInput({
+          label,
+          ...defaultProps,
+        });
+
+        const upArrow = container.querySelector(
```
shaneeza, line 253: Maybe we should add testids to the arrows so this its easier for consumers to test?

stephl3: great, added those test ids and updated the relevant specs

shaneeza (approved):

packages/number-input/src/Input/Input.tsx
```diff
@@ -162,23 +164,28 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
+    const handleFocus = () => {
       isFocusedRef.current = true;
       handleSetErrorTransition();
     };
 
-    const handleOnBlur = () => {
+    const handleBlur = (e: FocusEvent<HTMLDivElement>) => {
       isFocusedRef.current = false;
       handleRemoveErrorTransition();
+
```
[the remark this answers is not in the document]
stephl3: added! 
`// If newly focused element is a child of the input container, we do not invoke onBlur`

packages/number-input/src/Arrows/Arrow.tsx
```diff
@@ -34,6 +34,7 @@ export const Arrow = ({
       type="button"
       tabIndex={-1} // Mimicking native behavior; you cannot focus on an arrow.
       disabled={disabled}
+      data-testid={`lg-number_input-${direction}-button`}
```
TheSonOfThomp, line 37: super-nit: `_button`

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -133,8 +135,10 @@ describe('packages/number-input', () => {
     });
 
     test('value change triggers onChange callback', () => {
+      const onChange = jest.fn();
```
TheSonOfThomp, line 138: 👍

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -202,47 +206,64 @@ describe('packages/number-input', () => {
...
+          ...defaultProps,
+        });
+
+        const upArrow = getByTestId(arrowTestId.up);
+
+        userEvent.click(upArrow as HTMLButtonElement); // focus
+        expect(upArrow).toHaveFocus();
+
+        userEvent.tab(); // blur
+        expect(onBlur).toHaveBeenCalledTimes(1);
```
TheSonOfThomp, line 263: 👍

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -202,47 +206,64 @@ describe('packages/number-input', () => {
...
+        });
+
+        userEvent.tab(); // focus
+        expect(numberInput).toHaveFocus();
+        userEvent.tab(); // blur
+
+        expect(onBlur).toHaveBeenCalledTimes(1);
+      });
+
+      test('callback triggers when focus leaves arrow buttons', () => {
```
TheSonOfThomp, line 249: We should test the case(s) where `onBlur` should _not_ be called

TheSonOfThomp (commented):
Just have the one comment on testing negative cases. Otherwise, LGTM!

packages/number-input/src/NumberInput/NumberInput.spec.tsx
stephl3: makes sense, added a test case

TheSonOfThomp (approved):

PR #2226, opened 2024-02-23
LG-3685, LG-2632: adds backButtonProps and cancelButtonProps to FormFooter component

stephl3:
## ✍️ Proposed changes

- Adds `cancelButtonProps` and `backButtonProps` to `FormFooter` component for customizing buttons.
- Marks `cancelButtonText`, `onCancel`, `backButtonText`, and `onBackClick` props as deprecated.
- Migrates package to new file structure

🎟 _Jira ticket:_ [LG-3685](https://jira.mongodb.org/browse/LG-3685) | [LG-2632](https://jira.mongodb.org/browse/LG-2632)

## 🧪 How to test changes

Can see the new prop usage in storybook and added unit test blocks

| light | dark |
| - | - |
| Screenshot 2024-02-27 at 2 53 33 PM | Screenshot 2024-02-27 at 2 54 05 PM |
| Screenshot 2024-02-27 at 2 53 45 PM | Screenshot 2024-02-27 at 2 54 14 PM |


packages/form-footer/src/FormFooter.tsx
```diff
@@ -141,23 +192,41 @@ export default function FormFooter({
...
+
+  const backButton = {
+    text: backButtonProps?.text || backButtonText,
+    onClick: backButtonProps?.onClick || onBackClick,
+    leftGlyph:
+      backButtonProps?.leftGlyph === undefined ? (
+        <ArrowLeftIcon data-testid="lg-form_footer-back-button-icon" />
+      ) : (
+        backButtonProps.leftGlyph ?? undefined
+      ),
```
stephl3, line 204: API works like so:
- if `leftGlyph` is undefined, we default to the `ArrowLeftIcon`
- option to pass an explicit null to hide the icon
- option to pass a custom icon

open to feedback on this if any alternative preferences or if we have any established patterns we should adhere to

packages/form-footer/src/FormFooter.tsx
```diff
@@ -130,6 +179,8 @@ export interface FormFooterProps extends HTMLElementProps<'footer'> {
  */
 export default function FormFooter({
   primaryButton,
+  cancelButtonProps,
+  backButtonProps,
   onCancel,
   cancelButtonText = 'Cancel',
```
stephl3, line 185: Ran into some issues with how we handle hiding the cancel button

I would expect that not defining `cancelButtonText` would lead to NOT rendering the button similar to the approach for the back button. However, it seems that if left undefined, the text will default to `'Cancel'`. To not render the cancel button, `cancelButtonText` must be an empty string

This seems to create some issues with maintaining backwards compatibility with these newly added props. In order to maintain backwards compatibility, I think we should keep these defaults as-is

packages/form-footer/src/FormFooter.tsx
```diff
@@ -170,15 +239,16 @@ export default function FormFooter({
...
               variant="default"
-              onClick={onCancel}
+              onClick={cancelButton.onClick}
               className={buttonStyle}
+              leftGlyph={cancelButton.leftGlyph}
               darkMode={darkMode}
               data-testid="lg-form_footer-cancel-button"
             >
-              {cancelButtonText || 'Cancel'}
+              {cancelButton.text || 'Cancel'}
```
stephl3, line 251: see above note about keeping default cancel button label as-is

packages/form-footer/src/FormFooter.story.tsx
```diff
@@ -41,8 +41,13 @@ const meta: StoryMetaType<typeof FormFooter> = {
     generate: {
       storyNames: ['LightMode', 'DarkMode'],
       combineArgs: {
-        backButtonText: [undefined, 'Back'],
-        cancelButtonText: ['', 'Cancel'],
+        backButtonProps: [
+          undefined,
+          { text: 'Back', leftGlyph: undefined },
+          { text: 'Back', leftGlyph: null },
+        ],
```
stephl3, line 48: Seemed superfluous to include a generated set of stories showing the ability to use a custom `leftGlyph` since most teams will still use the `ArrowLeftIcon`. Open to adding if we prefer to have greater coverage

.changeset/weak-stingrays-live.md
```diff
@@ -0,0 +1,7 @@
+---
+'@leafygreen-ui/form-footer': patch
```
shaneeza, line 2: Since this isn't a bug fix, I think this should be `minor`

packages/form-footer/src/FormFooter.spec.tsx
```diff
@@ -3,10 +3,17 @@ import { render } from '@testing-library/react';
 import { axe } from 'jest-axe';
 
 import Button from '@leafygreen-ui/button';
+import X from '@leafygreen-ui/icon/dist/X';
```
shaneeza, line 6: `X` is a little hard to understand, `XIcon` would be better

packages/form-footer/src/FormFooter.tsx
```diff
@@ -58,6 +58,16 @@ const buttonStyle = css`
...
  * Types
  */
+interface BaseButtonProps {
+  text?: string;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+}
+
+interface BackButtonProps extends BaseButtonProps {
+  leftGlyph?: React.ReactElement | null;
+}
```
TheSonOfThomp, line 68: Should we allow users to fully customize the buttons? The way I read the ticket, the intent is not only to customize the click handler & left glyph, but allow for further customization. 

@bruugey I left a comment in the JIRA ticket, since I'm a bit confused about the AC

bruugey: Responded in JIRA

packages/form-footer/src/FormFooter.spec.tsx
```diff
@@ -3,10 +3,17 @@ import { render } from '@testing-library/react';
...
 
 import { FormFooterProps } from './FormFooter';
 import FormFooter from '.';
 
+const buttonTestId = {
+  back: 'lg-form_footer-back-button',
```
shaneeza, line 12: I believe these should all end with `_button` @TheSonOfThomp ?

packages/form-footer/src/FormFooter.tsx
```diff
@@ -58,6 +58,16 @@ const buttonStyle = css`
 /**
  * Types
  */
+interface BaseButtonProps {
+  text?: string;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+}
```
TheSonOfThomp, line 64: Should we extend `ButtonProps` from the button package?

packages/form-footer/src/FormFooter.spec.tsx
```diff
@@ -39,30 +46,119 @@ describe('packages/form-footer', () => {
       expect(ButtonElement).toBeInTheDocument();
     });
 
-    test('Renders cancel button', () => {
-      const { getByText } = renderFooter({
-        primaryButton: { text: 'Test button' },
+    // remove once deprecated props are removed
```
TheSonOfThomp, line 49: add `TODO:`

packages/form-footer/src/FormFooter.tsx
TheSonOfThomp: FWIW I think if we're going as far as accepting a prop object, I would prefer a slot API instead. That's the route we took for CanvasHeader, and I'm sure other devs would appreciate the customizability

packages/form-footer/src/FormFooter.tsx
```diff
@@ -78,26 +88,65 @@ export interface FormFooterProps extends HTMLElementProps<'footer'> {
...
+   * darkMode is handled internally so you do not have to pass the darkMode prop.
+   */
+  backButtonProps?: BackButtonProps;
+
   /**
    * Text for the cancel button.
    * A cancel button will only appear if this text is defined.
    *
    * @default "Cancel"
+   * @deprecated since version 3.0.17 - use cancelButtonProps instead
```
TheSonOfThomp, line 128: 👍

packages/form-footer/src/FormFooter.tsx
```diff
@@ -141,23 +192,41 @@ export default function FormFooter({
   ...rest
 }: FormFooterProps) {
   const { theme, darkMode } = useDarkMode(darkModeProp);
+
+  const backButton = {
```
TheSonOfThomp, line 196: nit: backButtonProps (or similar. It's not the button itself)

packages/form-footer/src/FormFooter.tsx
TheSonOfThomp: I believe the intent here was to render a default cancel button unless otherwise specified. That definitely results in some interesting backwards compat issues though with the new props

packages/form-footer/src/FormFooter.tsx
```diff
@@ -141,23 +192,41 @@ export default function FormFooter({
...
+      ) : (
+        backButtonProps.leftGlyph ?? undefined
+      ),
+  };
+
+  const cancelButton = {
+    text: cancelButtonProps?.text || cancelButtonText,
+    onClick: cancelButtonProps?.onClick || onCancel,
+    leftGlyph: cancelButtonProps?.leftGlyph ?? undefined,
+  };
```
TheSonOfThomp, line 211: I think we could leverage something like `_.defaults` here

TheSonOfThomp: I think this is where we could add logic for the default text as well

packages/form-footer/src/FormFooter.story.tsx
TheSonOfThomp: This is fine. This is just for visual tests, so we just need to know an icon renders correctly

packages/form-footer/src/FormFooter.story.tsx
```diff
@@ -41,8 +41,13 @@ const meta: StoryMetaType<typeof FormFooter> = {
     generate: {
       storyNames: ['LightMode', 'DarkMode'],
       combineArgs: {
-        backButtonText: [undefined, 'Back'],
-        cancelButtonText: ['', 'Cancel'],
+        backButtonProps: [
+          undefined,
+          { text: 'Back', leftGlyph: undefined },
+          { text: 'Back', leftGlyph: null },
```
TheSonOfThomp, line 47: Do `leftGlyph: undefined` and `leftGlyph: null` render differently? If not, we can omit one of these. This is just for visual testing

packages/form-footer/src/FormFooter.spec.tsx
```diff
@@ -39,30 +46,119 @@ describe('packages/form-footer', () => {
...
+          primaryButton: { text: 'Test button' },
+          backButtonProps: {
+            text: backButtonText,
+          },
+        });
+        const Back = queryByText(backButtonText);
+        expect(Back).toBeInTheDocument();
+      });
+
+      test('does not render', () => {
```
TheSonOfThomp, line 114: when?

packages/form-footer/src/FormFooter.spec.tsx
```diff
@@ -39,30 +46,119 @@ describe('packages/form-footer', () => {
...
+        const { queryByTestId } = renderFooter({
+          primaryButton: { text: 'Test button' },
+        });
+        const Back = queryByTestId(buttonTestId.back);
+        expect(Back).not.toBeInTheDocument();
+      });
+
+      describe('Back button left glyph', () => {
```
TheSonOfThomp, line 122: nit: can omit `"Back button"` here since it's nested in another `describe` block

packages/form-footer/src/FormFooter.spec.tsx
```diff
@@ -39,30 +46,119 @@ describe('packages/form-footer', () => {
...
+        test('Renders ArrowLeftIcon if leftGlyph is undefined', () => {
```
TheSonOfThomp, line 123: 👍

packages/form-footer/src/FormFooter.spec.tsx
```diff
@@ -39,30 +46,119 @@ describe('packages/form-footer', () => {
...
+          const { queryByTestId } = renderFooter({
+            primaryButton: { text: 'Test button' },
+            backButtonProps: {
+              text: 'Back',
+              leftGlyph: undefined,
```
shaneeza, line 128: Do you need to pass `undefined`, or can you omit `leftGlyph`?

packages/form-footer/src/FormFooter.story.tsx
```diff
@@ -41,8 +41,13 @@ const meta: StoryMetaType<typeof FormFooter> = {
...
+        ],
+        cancelButtonProps: [undefined, { text: 'Cancel' }],
+        cancelButtonText: [''], // remove once deprecated props are removed
```
shaneeza, line 50: You might be able to add this inside `args` on L60 and remove it from here

packages/form-footer/src/FormFooter.tsx
shaneeza: I think the diff here is that we don't want consumers to change the button variants.

packages/form-footer/src/FormFooter.tsx
shaneeza, line 1: Maybe we should update the file structure! I can find the ticket

shaneeza: @stephl3 https://jira.mongodb.org/browse/LG-2632 If you're up for it you could update the file structure of this component to match the other components

packages/form-footer/src/FormFooter.tsx
stephl3: yeah from my discussion with design, they only want to introduce button icon customization

so then, customization should be restricted to text + click handler + icon

packages/form-footer/src/FormFooter.spec.tsx
stephl3: ooh gotcha. I was keeping it consistent with what was already in the file, but that makes sense. I'll keep an eye out for inconsistency with what shows up in the style guide and make updates when necessary

packages/form-footer/src/FormFooter.tsx
stephl3: good point, I can use that instead. I will need to make `leftGlyph` nullable though

packages/form-footer/src/FormFooter.tsx
stephl3: ahh yeah did this to avoid naming collision but I can rename to something else

packages/form-footer/src/FormFooter.story.tsx
stephl3: yes, was previously following the api described here: https://github.com/mongodb/leafygreen-ui/pull/2226#discussion_r1505127559

but upon further thought, I don't think I need to do that and will update

packages/form-footer/src/FormFooter.spec.tsx
stephl3: omit is fine too but I made it explicit so it's more clear when reading the test block

packages/form-footer/src/FormFooter.tsx
stephl3: thanks! will check this out!

packages/form-footer/src/FormFooter.spec.tsx
stephl3: dropped since I changed the api

packages/form-footer/src/FormFooter.tsx
stephl3: To make reviewing diffs easier, I'm going to tackle this in a separate PR

stephl3: added!

TheSonOfThomp (approved):

packages/form-footer/src/FormFooter.types.ts
```diff
@@ -0,0 +1,98 @@
+import { ButtonProps } from '@leafygreen-ui/button';
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+
+import { PrimaryButtonProps } from './PrimaryButton';
+
+type CustomButtonProps = Pick<
```
TheSonOfThomp, line 6: 👍

PR #2232, opened 2024-02-28
LG-3915: Add tooltip on hover for Code component copy button

stephl3:
## ✍️ Proposed changes

Add tooltip hover interaction for `Code` component copy button

🎟 _Jira ticket:_ [LG-3915](https://jira.mongodb.org/browse/LG-3915)

### For new components

## 🧪 How to test changes

Changes visible in `Code` component storybook

| mouse interaction | keyboard interaction |
| - | - |
| https://github.com/mongodb/leafygreen-ui/assets/52300529/2040abbb-5837-4e28-b2cc-c7fa21633137 | https://github.com/mongodb/leafygreen-ui/assets/52300529/b28e29d7-a19f-43cd-af73-57aee0e82f81 |


packages/code/src/CopyButton/CopyButton.tsx
```diff
@@ -51,19 +56,45 @@ function CopyButton({ onCopy, contents }: CopyProps) {
...
+      trigger={
+        <IconButton
+          ref={setButtonNode}
+          darkMode={darkMode}
+          aria-label={COPY_TEXT}
+          className={cx(copyButtonThemeStyles[theme], {
+            [copiedThemeStyle[theme]]: copied,
+          })}
+          onClick={handleClick}
+          onMouseEnter={handleMouseEnter}
```
shaneeza, line 85: Would this still work if you removed `onMouseEnter` and `onMouseLeave`? The tooltip should open on hover by default.

stephl3: https://github.com/mongodb/leafygreen-ui/assets/52300529/77788845-4031-4268-97f9-80c305fcfa2e

yeah initially tried to implement with the default api behavior, but I ran into issues with component re-renders because `CopyButton` re-renders when the `copied` state toggles on/off. That resulted in the above inconsistent UX. By setting up state to track the `hovered` state and making the tooltip `open` / `setOpen` controlled by `hovered`, I was able to keep the `Tooltip` open regardless of `CopyButton` re-rendering

shaneeza: I see. Don't forget to add `onFocus` and `onBlur`. Right now, the tooltip remains open if you tab away from the copy button.

packages/code/src/CopyButton/CopyButton.tsx
```diff
@@ -51,19 +56,45 @@ function CopyButton({ onCopy, contents }: CopyProps) {
...
+
+  const shouldClose = () => !hovered;
+
   return (
-    <IconButton
-      ref={setButtonNode}
+    <Tooltip
+      open={hovered}
+      setOpen={setHovered}
       darkMode={darkMode}
```
shaneeza, line 73: This isn't part of your changes but, you should be able to remove `darkMode` from here since we pass `darkMode` to the provider in `code.tsx`

packages/code/src/CopyButton/CopyButton.tsx
stephl3: https://github.com/mongodb/leafygreen-ui/assets/52300529/c3188850-a59e-4854-b664-d48bf22dedf8
I ended up needing to use `handleKeyDown` instead of `onFocus`/`onBlur` because using `onBlur` conflicts with the mouse event handlers I have in place. I have it specifically close out for tab. Nothing else in the `keyMap` seemed to require action although I did consider `Escape` key

@shaneeza would appreciate your feedback on this so I can stay consistent with the codebase. Thanks!

packages/code/src/CopyButton/CopyButton.tsx
stephl3: done!

packages/code/src/CopyButton/CopyButton.tsx
shaneeza: `handleKeyDown` doesn't feel correct, how was `onBlur` conflicting with the mouse event handlers?

shaneeza: Also, should the tooltip still appear on hover after clicking on it? It seems like the tooltip would still be open but the text would change back to `copy`

stephl3: > how was `onBlur` conflicting with the mouse event handlers?
I set up `hovered` state to control the open/close of the tooltip. If I use `onBlur` to toggle off `hovered`, this will happen:
1. mouse enter: hovered set to true and tooltip opens
2. click: copied set to true which makes `trigger` button re-render which causes closed tooltip to re-render
3. mouse enter on newly rendered button: hovered set to true and tooltip re-opens

> Also, should the tooltip still appear on hover after clicking on it? It seems like the tooltip would still be open but the text would change back to copy
ahh yes, the change I made for keyboard interaction was causing it to close, but I just pushed a change that should support both cases now. will update the videos in the PR description

packages/code/src/CopyButton/CopyButton.tsx
```diff
@@ -39,31 +51,70 @@ function CopyButton({ onCopy, contents }: CopyProps) {
     }
 
     return () => clipboard.destroy();
-  }, [buttonNode, contents, copied, portalContainer]);
+  }, [buttonRef, contents, copied, portalContainer]);
```
shaneeza, line 54: `buttonRef` won't trigger a rerender since it's a ref, that's probably why a callback ref was initially used. Although, it doesn't seem like it needs to be a dependency. Was the callback ref causing issues?

packages/code/src/CopyButton/CopyButton.tsx
```diff
@@ -39,31 +51,70 @@ function CopyButton({ onCopy, contents }: CopyProps) {
...
 
   const handleClick = (e: React.MouseEvent) => {
     e.preventDefault();
+    onCopy?.();
+    setCopied(true);
+  };
+
+  const handleKeyDown = (e: React.KeyboardEvent) => {
```
shaneeza, line 62: Can you leave a comment on why you had to use `handleKeyDown`?

packages/code/src/CopyButton/CopyButton.tsx
```diff
@@ -39,31 +51,70 @@ function CopyButton({ onCopy, contents }: CopyProps) {
...
+        closeTooltip();
+        break;
+      }
 
-    if (onCopy) {
-      onCopy();
+      case keyMap.Enter:
+      case keyMap.Space: {
+        e.preventDefault();
+        buttonRef.current?.click();
```
shaneeza, line 73: Does clicking enter not automatically trigger a click?

shaneeza (changes requested):
Looking good! Can you add some tests testing the tooltip functionality?

packages/code/src/CopyButton/CopyButton.tsx
stephl3: no, that wasn't causing issues, but I wasn't able to click the button with the enter/space keys so I defined a separate ref

packages/code/src/CopyButton/CopyButton.tsx
stephl3: Yeah for some strange reason, it doesn't. I originally tried invoking `handleClick`. It would trigger the state change but wouldn't create a clipboard of the contents. I ended up opting for this approach

shaneeza: Is it possible to test that the correct content is copied?

shaneeza (changes requested):

packages/code/src/CopyButton/CopyButton.tsx
stephl3: ahh I thought this was already covered in the original code spec file but I guess it's not. I will add to this spec file

TheSonOfThomp:
Is this PR dependant on `steph/tooltip-tab-bug` (#2240)?

packages/code/src/CopyButton/CopyButton.tsx
shaneeza: no worries, I also assumed the same thing. I'm not sure why its missing.

stephl3:
> Is this PR dependant on `steph/tooltip-tab-bug` (#2240)?

@TheSonOfThomp no these are separate. this one is for adding a tooltip to the code package while the other one is to change tooltip focus behavior

packages/code/src/CopyButton/CopyButton.spec.tsx
```diff
@@ -54,9 +88,21 @@ describe('CopyButton', () => {
...
+    test('invokes onCopy and copies contents to clipboard onClick', () => {
       const onCopy = jest.fn();
-      const { queryByTestId } = renderCopyButton({ onCopy });
+      const { getByTestId } = renderCopyButton({ onCopy });
+      const copyButton = getByTestId(testIds.copyButton);
+
+      userEvent.click(copyButton);
+      expect(onCopy).toHaveBeenCalledTimes(1);
+      expect(ClipboardJS).toHaveBeenCalledWith(copyButton, {
+        text: expect.any(Function),
```
shaneeza, line 99: What does this mean?

packages/code/src/CopyButton/CopyButton.spec.tsx
```diff
@@ -23,19 +27,49 @@ describe('CopyButton', () => {
...
+
+  test(`tooltip displays "${COPIED_TEXT}" text for ${COPIED_SUCCESS_DURATION}ms after trigger is clicked`, async () => {
+    const { getByTestId, queryByTestId } = renderCopyButton({});
+    const copyButton = getByTestId(testIds.copyButton);
+    let tooltip = queryByTestId(testIds.tooltip);
+
+    await waitFor(() => {
+      userEvent.click(copyButton);
+      tooltip = queryByTestId(testIds.tooltip);
+      expect(tooltip).toHaveTextContent(COPIED_TEXT);
```
shaneeza, line 59: 👍🏽

packages/code/src/CopyButton/CopyButton.spec.tsx
stephl3: @shaneeza this tests that when copy button is clicked, a new clipboard object is made

shaneeza (approved):

PR #2239, opened 2024-03-06
LG-3988: expose portalRef from Popover and Portal props

stephl3:
## ✍️ Proposed changes
- expose manual fwdref from `Portal`
- expose `portalRef` from `Popover` 

🎟 _Jira ticket:_ [LG-3988](https://jira.mongodb.org/browse/LG-3988)
[Part 1](https://github.com/mongodb/leafygreen-ui/pull/2239) | [Part 2](https://github.com/mongodb/leafygreen-ui/pull/2241)

### For new components

### DIFF
--- a/packages/popover/src/Popover.types.ts
+++ b/packages/popover/src/Popover.types.ts
@@ -186,6 +186,11 @@ export type PopoverProps = {
    * Number that controls the z-index of the popover element directly.
    */
   popoverZIndex?: number;
+
+  /**
+   * A ref for the portal element
+   */
+  portalRef?: React.MutableRefObject<HTMLElement | null>;
 } & PortalControlProps &
   TransitionLifecycleCallbacks;
 
--- a/packages/popover/src/Popover/Popover.spec.tsx
+++ b/packages/popover/src/Popover/Popover.spec.tsx
@@ -55,6 +55,17 @@ describe('packages/popover', () => {
     expect(ref.current).toBeDefined();
   });
 
+  test('accepts a portalRef', () => {
+    const portalRef = createRef<HTMLElement>();
+    render(
+      <Popover portalRef={portalRef} data-testid="popover-test-id">
+        Popover Content
+      </Popover>,
+    );
+
+    expect(portalRef.current).toBeDefined();
+  });
+
   test('displays popover when the "active" prop is set', () => {
     const { getByTestId } = renderPopover({ active: true });
     expect(getByTestId('popover-test-id')).toBeInTheDocument();
--- a/packages/popover/src/Popover/Popover.spec.tsx
+++ b/packages/popover/src/Popover/Popover.spec.tsx
@@ -55,15 +55,18 @@ describe('packages/popover', () => {
     expect(ref.current).toBeDefined();
   });
 
-  test('accepts a portalRef', () => {
+  test('accepts a portalRef', async () => {
     const portalRef = createRef<HTMLElement>();
-    render(
-      <Popover portalRef={portalRef} data-testid="popover-test-id">
-        Popover Content
-      </Popover>,
-    );
+    waitFor(() => {
+      render(
+        <Popover portalRef={portalRef} data-testid="popover-test-id">
+          Popover Content
+        </Popover>,
+      );
 
-    expect(portalRef.current).toBeDefined();
+      expect(portalRef.current).toBeDefined();
+      expect(portalRef.current).toBeInTheDocument();
+    });
   });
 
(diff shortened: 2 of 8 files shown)


packages/popover/src/Popover/Popover.tsx
```diff
@@ -262,11 +264,12 @@ export const Popover = forwardRef<HTMLDivElement, PopoverComponentProps>(
     `;
 
     const Root = usePortal ? Portal : Fragment;
-    const rootProps = usePortal
-      ? portalContainer
-        ? { container: portalContainer }
-        : { className: portalClassName ?? undefined }
-      : {};
+    const portalProps = {
```
TheSonOfThomp, line 267: 👍 much cleaner

packages/portal/src/Portal/Portal.types.ts
```diff
@@ -4,6 +4,7 @@ import { OneOf } from '@leafygreen-ui/lib';
 
 export type PortalProps = {
   children?: React.ReactNode;
+  _ref?: React.MutableRefObject<HTMLElement | null>;
```
TheSonOfThomp, line 7: not sure if `_ref` is the best prop name (the underscore makes it seem like it's unused or internal)

packages/portal/src/Portal/Portal.tsx
```diff
@@ -15,39 +18,52 @@ export function usePortalContainer(customContainer?: HTMLElement) {
 
   useIsomorphicLayoutEffect(() => {
     if (customContainer) {
+      if (portalRef) {
+        portalRef.current = customContainer;
```
TheSonOfThomp, line 22: Can we add some comments to clarify what's going on here. It's definitely not typical 🙃

packages/popover/src/Popover.types.ts
```diff
@@ -186,6 +186,11 @@ export type PopoverProps = {
    * Number that controls the z-index of the popover element directly.
    */
   popoverZIndex?: number;
+
+  /**
+   * A ref for the portal element
+   */
+  portalRef?: React.MutableRefObject<HTMLElement | null>;
```
TheSonOfThomp, line 193: Can we add tests for the newly added props?

TheSonOfThomp (changes requested):
Code LGTM. Just need to add tests

packages/portal/src/Portal/Portal.types.ts
stephl3: I wanted to name this `ref` but it would create a console warning because the `Portal` component wasn't wrapped in a `forwardRef`. `_ref` seems appropriate since it is internally used and the `portalRef` prop is what we expose to external users. Do you have an alternative name suggestion?

packages/popover/src/Popover/Popover.spec.tsx
```diff
@@ -55,6 +55,17 @@ describe('packages/popover', () => {
...
 
+  test('accepts a portalRef', () => {
+    const portalRef = createRef<HTMLElement>();
+    render(
+      <Popover portalRef={portalRef} data-testid="popover-test-id">
+        Popover Content
+      </Popover>,
+    );
+
+    expect(portalRef.current).toBeDefined();
```
TheSonOfThomp, line 66: We should also test `expect(portalRef.current).toBeInTheDocument()`

packages/portal/src/Portal/Portal.types.ts
stephl3: ok updated it to `portalRef`

packages/popover/src/Popover/Popover.spec.tsx
```diff
@@ -55,15 +55,18 @@ describe('packages/popover', () => {
...
 
-  test('accepts a portalRef', () => {
+  test('accepts a portalRef', async () => {
     const portalRef = createRef<HTMLElement>();
-    render(
-      <Popover portalRef={portalRef} data-testid="popover-test-id">
-        Popover Content
-      </Popover>,
-    );
+    waitFor(() => {
```
TheSonOfThomp, line 60: Do we need to call `waitFor` here? I think `render` does that internally

TheSonOfThomp (approved):

packages/popover/src/Popover/Popover.spec.tsx
stephl3: without the `waitFor` wrapping, `portalRef.Current` wasn't showing up in the document. It takes too long to render before the expect statement runs due to the `useIsomorphicLayoutEffect` in the portal component

PR #2240, opened 2024-03-06
Blur last focused element for tooltip

stephl3:
## ✍️ Proposed changes

### For new components


.changeset/chatty-drinks-itch.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/tooltip': patch
+---
+
+Blur document.activeElement on blur of `Tooltip` component to prevent tooltip unexpectedly reappearing
```
shaneeza, line 5: reappearing when? It might be helpful to add when it reappears

packages/tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -155,6 +155,12 @@ function Tooltip({
             onBlur: (e: MouseEvent) => {
               userTriggerHandler('onBlur', e);
               handleClose();
+
+              // If a user clicks on the tooltip trigger and immediately navigates to a new window/tab and then back, document.activeElement is still the trigger which will cause the
+              // tooltip to reappear even if not hovered. This blurs the last active element to prevent it from unexpectedly reappearing.
+              if (document.activeElement instanceof HTMLElement) {
```
shaneeza, line 161: Can you add a test to ensure that after tabbing to the next focusable item, the focus is on the correct element?

TheSonOfThomp, line 161: Should we be more explicit and check if `document.activeElement` is the trigger itself?

stephl3: good call! will put a ref on the trigger to compare against `document.activeElement` :)

stephl3: This will actually move focus to the `body` element. I wonder if that is acceptable or if we'd rather not address this issue altogether...

| pre-change | post-change |
| - | - |
| 1. click trigger: trigger focused<br/>2. navigate to new browser tab/window and back<br/>3. trigger still focused: tooltip reappears | 1. click trigger: trigger focused<br/>2. navigate to new browser tab/window and back<br/>3. body focused: tooltip does not appear<br/>4. hit tab and trigger is focused|

packages/tooltip/src/Tooltip/Tooltip.spec.tsx
```diff
@@ -694,6 +694,20 @@ describe('packages/tooltip', () => {
       });
     });
 
+    test('onMouseDown events should fire', async () => {
```
stephl3, line 697: added this spec to preserve the existing testing pattern for mouse events. I think it would be ideal to introduce a puppeteer test to check the browser navigation behavior but don't think it makes sense to introduce with this ticket

shaneeza: you can create a ticket for this!

shaneeza: also, can you test that if `onClick` is passed along with `onMouseDown` it's also called?

shaneeza: ignore my comment above!

PR #2256, opened 2024-03-14
Allow `defaultValue` to be used in `TextArea`

ilai2:
## ✍️ Proposed changes

Currently, the way `uncontrolledValue` is handled prevents `defaultValue` from working. I added `defaultValue` as the initial state for `uncontrolledValue` so that an initial value can be set while also allowing future edits from the user.

Not sure if I should have used `rest?.defaultValue` instead of just adding it, but since `className` is also not in `packages/text-area/src/TextArea/TextArea.types.ts` I thought it might be ok. Let me know if you want me to change that :D 

🎟 _Jira ticket:_ N/A

### For new components

## 🧪 How to test changes

### DIFF
--- a/.changeset/chilled-seahorses-pay.md
+++ b/.changeset/chilled-seahorses-pay.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/text-area': patch
+---
+
+Allow defaultValue to be used in TextArea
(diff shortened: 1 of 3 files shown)


bruugey:
This is perfect @ilai2 do you mind just adding a quick test to confirm that passing `defaultValue` sets a default value

bruugey (approved):

PR #2257, opened 2024-03-14
Fix spacing in FormField

bruugey:
## ✍️ Proposed changes

### For new components


packages/form-field/src/FormField/FormField.styles.ts
```diff
@@ -30,9 +30,12 @@ export const getFontSize = ({
   }
 };
 
-export const labelTextContainerStyle = css`
+export const textContainerStyle = css`
   display: flex;
   flex-direction: column;
+`;
+
+export const labelContainerStyle = css`
```
stephl3, line 38: could this instead be changed to `padding-bottom: ...` to avoid adding an extra div?

packages/form-field/src/FormField/FormField.tsx
```diff
@@ -62,17 +63,19 @@ export const FormField = forwardRef<HTMLDivElement, FormFieldProps>(
...
-                <Label
-                  data-testid="lg-form_field-label"
-                  className={getFontSize({ baseFontSize, size })}
-                  htmlFor={inputId}
-                  id={labelId}
-                  disabled={disabled}
-                >
-                  {label}
-                </Label>
+                <div className={labelContainerStyle}>
```
shaneeza, line 68: Could this style be added to the label itself instead of a new div?

packages/form-field/src/FormField/FormField.styles.ts
bruugey: Updated!

packages/form-field/src/FormField/FormField.tsx
bruugey: Updated!

stephl3 (approved):

packages/form-field/src/FormField/FormField.styles.ts
```diff
@@ -30,9 +30,12 @@ export const getFontSize = ({
   }
 };
 
-export const labelTextContainerStyle = css`
+export const textContainerStyle = css`
```
stephl3, line 33: non-blocking system-related question: have we put any thought into having layout components like a `VStack` in this case?

packages/form-field/src/FormField/FormField.styles.ts
```diff
@@ -30,9 +30,12 @@ export const getFontSize = ({
...
   display: flex;
   flex-direction: column;
+`;
+
+export const paddingBottom = css`
```
stephl3, line 38: can we rename this `marginBottom` or `spacingBottom`? I don't like mixing up margin/padding since they can have different behavior

stephl3 (approved):

PR #2261, opened 2024-03-15
LG-3915: Fix CopyButton test mock

stephl3:
## ✍️ Proposed changes

- fixes broken tests

🎟 _Jira ticket:_ [LG-3915](https://jira.mongodb.org/browse/LG-3915)

### For new components

### DIFF
--- a/packages/code/src/CopyButton/CopyButton.spec.tsx
+++ b/packages/code/src/CopyButton/CopyButton.spec.tsx
@@ -10,7 +10,11 @@ import { COPIED_SUCCESS_DURATION, COPIED_TEXT, COPY_TEXT } from './constants';
 import CopyButton from './CopyButton';
 import { CopyProps } from './CopyButton.types';
 
-jest.mock('clipboard');
+jest.mock('clipboard', () => {
+  return jest.fn().mockImplementation(() => ({
+    destroy: jest.fn(),
+  }));
+});
 
 describe('CopyButton', () => {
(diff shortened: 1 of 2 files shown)


packages/code/src/CopyButton/CopyButton.spec.tsx
```diff
@@ -51,11 +58,11 @@ describe('CopyButton', () => {
   test(`tooltip displays "${COPIED_TEXT}" text for ${COPIED_SUCCESS_DURATION}ms after trigger is clicked`, async () => {
     const { getByTestId, queryByTestId } = renderCopyButton({});
     const copyButton = getByTestId(testIds.copyButton);
-    let tooltip = queryByTestId(testIds.tooltip);
 
+    userEvent.click(copyButton);
+
+    let tooltip = queryByTestId(testIds.tooltip);
```
shaneeza, line 64: I'm not 100% certain, but if you wanted to shorten this even more, i think you can get away with
```
let tooltip = await findByTestId(testIds.tooltip);
expect(tooltip).toHaveTextContent(COPIED_TEXT);
```

shaneeza (approved):

packages/code/src/CopyButton/CopyButton.spec.tsx
stephl3: it doesn't complete the re-render with the text update before the expect statement, so I have to keep the `waitFor` wrapping the expect statement

PR #2262, opened 2024-03-18
LG-4095: Improve ExpandableCard title prop type

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [4095](https://jira.mongodb.org/browse/4095)

### For new components

### DIFF
--- a/packages/expandable-card/src/ExpandableCard/ExpandableCard.types.ts
+++ b/packages/expandable-card/src/ExpandableCard/ExpandableCard.types.ts
@@ -12,7 +12,7 @@ export interface ExpandableCardProps
   /**
    * The title of the card
    */
-  title: React.ReactNode;
+  title: ReactNode;
 
   /**
    * Description text below the title
--- a/packages/expandable-card/src/ExpandableCard/ExpandableCard.types.ts
+++ b/packages/expandable-card/src/ExpandableCard/ExpandableCard.types.ts
@@ -1,4 +1,4 @@
-import { ReactNode } from 'react';
+import React, { ReactNode } from 'react';
 
 import { CardProps } from '@leafygreen-ui/card';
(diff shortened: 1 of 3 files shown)


packages/expandable-card/src/ExpandableCard/ExpandableCard.types.ts
```diff
@@ -7,12 +7,12 @@ import { DarkModeProps } from '@leafygreen-ui/lib';
...
  */
 export interface ExpandableCardProps
-  extends Omit<CardProps, 'contentStyle'>,
+  extends Omit<CardProps, 'contentStyle' | 'title'>,
     DarkModeProps {
   /**
    * The title of the card
    */
-  title: string;
+  title: React.ReactNode;
```
stephl3, line 15: do we need to make this `string | React.ReactNode`? otherwise, doesn't this become a breaking change?

stephl3: nit: would make this `ReactNode` instead of ` React.ReactNode` since it's already imported for the `description` prop

bruugey: But I like that nit, will update :)

stephl3: > React.ReactNode includes strings so should be fine
I'll see myself out 😶‍🌫️

stephl3 (approved):

packages/expandable-card/src/ExpandableCard/ExpandableCard.types.ts
bruugey: hahahahhaha

PR #2270, opened 2024-03-22
LG-4113: Wrap `description` value content  in `div` not `p` tag

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4113](https://jira.mongodb.org/browse/LG-4113)

### For new components

### DIFF
--- a/.changeset/chatty-chefs-itch.md
+++ b/.changeset/chatty-chefs-itch.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/expandable-card': patch
+---
+
+Changes description content from being wrapped inside aa `p` tag to be wrapped in a `div`. This makes the nested content more dynamic.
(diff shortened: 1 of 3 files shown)


shaneeza (approved):

.changeset/chatty-chefs-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/expandable-card': patch
+---
+
+Changes description content from being wrapped inside aa `p` tag to be wrapped in a `div`. This makes the nested content more dynamic.
```
stephl3, line 5: nit: "... aa `p` tag..."

stephl3: non-blocking question on this: what does "more dynamic" mean? seems like if it only wraps text content, we'd want to use a `<span/>` or `<p/>` so I'm trying to understand why a `<div/>` makes more sense here

stephl3: oh gotcha, small thing but it seems like a `Box` would make more sense than `Body` as a div to take in dynamic elements. To me, `Body` assumes text content only (although I guess the `as` prop is pretty flexible. maybe something we'd want to make more restrictive)

bruugey: i get that and you're not wrong. we're deprecating box and this would ensure that we don't have to make this a breaking change, but open to feedback/thoguhts

stephl3 (approved):

PR #2276, opened 2024-03-26
Storybook: dark theme

TheSonOfThomp:
## ✍️ Proposed changes

- Creates a dark theme for the Storybook UI
- Adds `storybook-dark-mode` addon to toggle dark SB UI

- Also updates `palette.black` & `palette.white` exports to be defined `as const` to improve intellisense


stephl3 (approved):
😎

PR #2283, opened 2024-04-01
(EAI-267) [UI] Verified Answers in LG Chat

nlarew:
## ✍️ Proposed changes

- Adds support for the `verified` prop  to `@lg-chat/message`. When defined, the message is styled as. "verified answer"
- Adds a new `MessageBanner` component under `@lg-chat/message`. This wraps `@leafygreen-ui/banner` and modifies the styles (i.e. removes the left border colored wedge) to match the design.
  
Screenshot 2024-03-21 at 4 00 01 PM

🎟 _Jira ticket:_ [(EAI-267) [UI] Verified Answers in LG Chat](https://jira.mongodb.org/browse/EAI-267)

### For new components

## 🧪 How to test changes

Storybook > Message > Verified Answer


package.json
```diff
@@ -75,5 +75,8 @@
   },
   "bugs": {
     "url": "https://github.com/mongodb/leafygreen-ui/issues"
+  },
+  "dependencies": {
+    "@leafygreen-ui/banner": "^7.0.20"
```
TheSonOfThomp, line 80: We don't need to list lg dependencies in the root package.json

yarn.lock
```diff
@@ -15094,26 +14608,27 @@ which-typed-array@^1.1.10, which-typed-array@^1.1.11, which-typed-array@^1.1.2,
...
     isexe "^2.0.0"
 
 wordwrap@^1.0.0:
   version "1.0.0"
-  resolved "https://registry.yarnpkg.com/wordwrap/-/wordwrap-1.0.0.tgz#27584810891456a4171c8d0226441ade90cbcaeb"
+  resolved "https://registry.npmjs.org/wordwrap/-/wordwrap-1.0.0.tgz"
   integrity sha512-gvVzJFlPycKc5dZN4yPkP8w7Dc37BtP1yczEneOb4uq34pXZcvrtRTmWV8W+Ume+XCxKgbjM+nevkyFPMybd4Q==
 
 "wrap-ansi-cjs@npm:wrap-ansi@^7.0.0", wrap-ansi@^7.0.0:
+  name wrap-ansi-cjs
```
TheSonOfThomp, line 14629: this line in particular is what's causing issues

package.json
nlarew: Oops that snuck by me! Removed in the latest commit.

chat/message/src/Message/Message.tsx
```diff
@@ -148,3 +160,39 @@ export const Message = forwardRef(
 Message.displayName = 'Message';
 
 export default Message;
+
+function VerifiedAnswerBanner({
+  verifier,
+  verifiedAt,
+  learnMoreUrl,
+}: NonNullable<MessageProps['verified']>) {
```
TheSonOfThomp, line 168: I feel like this should be defined in a separate file, inside `/MessageBanner`

chat/message/src/Message/Message.tsx
```diff
@@ -148,3 +160,39 @@ export const Message = forwardRef(
...
+    if (verifier) {
+      textParts.push(`By ${verifier}`);
+    }
+
+    if (verifiedAt) {
+      const formattedDate = verifiedAt.toLocaleDateString('en-US', {
+        year: 'numeric',
+        month: 'long',
+        day: 'numeric',
+      });
```
TheSonOfThomp, line 181: We should be using the user's locale and not defaulting to US. (just pass undefined as the 1st param and it should use the default)

chat/message/src/Message/Message.tsx
TheSonOfThomp, line 168: Could we just say these props are of type `VerificationInfo`? Do we expect these to potentially be different in the future?

chat/message/src/Message/Message.tsx
```diff
@@ -148,3 +160,39 @@ export const Message = forwardRef(
...
 
 export default Message;
+
+function VerifiedAnswerBanner({
+  verifier,
+  verifiedAt,
+  learnMoreUrl,
+}: NonNullable<MessageProps['verified']>) {
+  const text = useMemo(() => {
+    const textParts = [`Answer Verified`];
```
TheSonOfThomp, line 170: Just want to confirm this is the expected behavior: if someone passes an empty object as the `verified` prop to `Message`, then the banner will only say "Answer Verified"

TheSonOfThomp: Should the word "verified" be lowercase?

chat/message/src/Message/Message.tsx
```diff
@@ -148,3 +160,39 @@ export const Message = forwardRef(
...
+      textParts.push(`By ${verifier}`);
+    }
+
+    if (verifiedAt) {
+      const formattedDate = verifiedAt.toLocaleDateString('en-US', {
+        year: 'numeric',
+        month: 'long',
+        day: 'numeric',
+      });
+      textParts.push(`On ${formattedDate}`);
```
TheSonOfThomp, line 182: Is this the desired order of data? (i.e. "Answer verified by ABC on XYZ"

chat/message/src/Message/Message.tsx
```diff
@@ -148,3 +160,39 @@ export const Message = forwardRef(
...
+function VerifiedAnswerBanner({
+  verifier,
+  verifiedAt,
+  learnMoreUrl,
+}: NonNullable<MessageProps['verified']>) {
+  const text = useMemo(() => {
+    const textParts = [`Answer Verified`];
+
+    if (verifier) {
+      textParts.push(`By ${verifier}`);
```
TheSonOfThomp, line 173: I don't think "By" and "On" should be in caps

chat/message/src/MessageBanner.stories.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
```
TheSonOfThomp, line 1: this file should be inside `/MessageBanner`

chat/message/src/MessageBanner/MessageBanner.styles.ts
```diff
@@ -0,0 +1,63 @@
+import { Variant } from '@leafygreen-ui/banner';
+import { css } from '@leafygreen-ui/emotion';
+import { Theme } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+import { spacing } from '@leafygreen-ui/tokens';
+
+export const baseStyles = css`
+  // Remove left wedge
```
TheSonOfThomp, line 8: can we be more specific in this comment (which left wedge?)

chat/message/src/MessageBanner/MessageBanner.styles.ts
```diff
@@ -0,0 +1,63 @@
...
+`;
+const darkModeWarningMessageBannerStyles = css`
+  border-color: ${palette.yellow.dark2};
+`;
+const darkModeDangerMessageBannerStyles = css`
+  border-color: ${palette.red.dark2};
+`;
+const darkModeSuccessMessageBannerStyles = css`
+  border-color: ${palette.green.dark2};
+`;
```
TheSonOfThomp, line 48: Are these different from the base "Banner" component? Why?

chat/message/src/MessageBanner/MessageBanner.styles.ts
```diff
@@ -0,0 +1,63 @@
...
+    background: transparent;
+  }
+
+  border-width: ${spacing[25]}px;
+  border-radius: ${spacing[600]}px;
+  border-left-color: unset;
+
+  max-width: fit-content;
+  padding: ${spacing[200]}px ${spacing[300]}px;
+`;
```
TheSonOfThomp, line 20: This looks a little off with long text (the icon looks out of place) 
Screenshot 2024-04-02 at 11 39 45 AM

chat/message/src/Message/Message.tsx
TheSonOfThomp: Also, I believe this should be sentence case
Screenshot 2024-04-02 at 11 40 53 AM

chat/message/src/MessageBanner/MessageBanner.types.ts
```diff
@@ -0,0 +1,17 @@
...
+  /**
+   * Determines the color and glyph of the MessageBanner.
+   * @default Variant.Info
+   */
+  variant?: Variant;
+
+  /**
+   * The content inside of the MessageBanner.
+   */
+  children: React.ReactNode | Array<React.ReactNode>;
```
TheSonOfThomp, line 16: `ReactNode` includes `Iterable<ReactNode>`, so we shouldn't need the extra `Array` definition

TheSonOfThomp (commented):
Looks good, just left a few recommended fixes

chat/message/src/MessageBanner/MessageBanner.styles.ts
nlarew: These are a re-implementation of the regular `border-color` in `@leafygreen-ui/banner`.

The original banner also has a different `border-left-color` set which shows up when you remove the wedge. Here's how `MessageBanner` looks without these styles applied:

Screenshot 2024-04-02 at 11 54 23 AM

Here it is without these styles but with `border-left-color: unset;`:

Screenshot 2024-04-02 at 11 49 10 AM

I didn't find a way to override the style so I re-implemented them instead. Obviously not very DRY! Any thoughts on a better way here?

chat/message/src/Message/Message.tsx
nlarew: Ah yeah good catch! I originally had `VerificationInfo` defined inline and meant to port it over here.

chat/message/src/Message/Message.tsx
nlarew: Designs use title case but I agree sentence case is reasonable. I'll run it by Alven and switch.

chat/message/src/MessageBanner.stories.tsx
nlarew: Moved this + moved `Message.stories.tsx` under `/Message`

chat/message/src/Message/Message.tsx
nlarew: Alven says the title case is intentional and wants to stick with it here.

chat/message/src/Message/Message.tsx
nlarew: Title case - see https://github.com/mongodb/leafygreen-ui/pull/2283#discussion_r1548503069

chat/message/src/Message/Message.tsx
nlarew: Also yeah this is the order we use in the design `Answer Verified By <verifier> On <date>`

chat/message/src/Message/Message.tsx
nlarew: > if someone passes an empty object as the verified prop to Message, then the banner will only say "Answer Verified"

Yeah correct though we'll guard against that specific case in our implementation - we ensure every answer has a `verifiedAt` value during import and hard code `verifier: "MongoDB Staff"` for now.

chat/message/src/MessageBanner/MessageBanner.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children,
+  variant = 'info',
+  ...divProps
+}: MessageBannerProps) {
+  const { theme } = useDarkMode(darkModeProp);
+  return (
+    <Banner
+      className={cx(baseStyles, variantStyles[theme][variant], className)}
+      variant={variant}
+      baseFontSize={13}
```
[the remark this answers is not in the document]
nlarew: Hadn't considered this - designs only have 13px but I don't think it's intended to always be the case. I'll remove this hardcode and let it inherit.

chat/message/src/MessageBanner/MessageBanner.styles.ts
nlarew: I missed a small part of the design here - the border radius is supposed to adjust for multiline banners.

Here's the updated version - does it look better aligned?

Screenshot 2024-04-02 at 4 32 46 PM

nlarew: FYI container queries ended up being a bit of a bust here so I handled this in JS instead

.changeset/odd-panthers-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/banner': patch
+---
+
+Banner now forwards refs
```
TheSonOfThomp, line 5: Technically a major change https://legacy.reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

chat/message/src/MessageBanner/MessageBanner.styles.ts
TheSonOfThomp: Honestly it doesn't look like [those border styles](https://github.com/mongodb/leafygreen-ui/blob/main/packages/banner/src/Banner/styles.ts#L32) are even used in `Banner`. Feel free to try and remove those, and if Chromatic passes, then great1

chat/message/src/MessageBanner/MessageBanner.tsx
```diff
@@ -15,11 +22,25 @@ export function MessageBanner({
...
+    }
+  }, [children, baseFontSize]);
   return (
     <Banner
-      className={cx(baseStyles, variantStyles[theme][variant], className)}
+      ref={bannerRef}
+      className={cx(
+        baseStyles,
+        variantStyles[theme][variant],
+        multilineStyles({ isMultiline }),
```
TheSonOfThomp, line 40: for just a single boolean to toggle styles, we tend to go with this pattern:
```js
cx(
  baseStyles,
  {
    [multilineStyles]: isMultiline
  }
)
```

chat/message/src/MessageBanner/VerifiedAnswerBanner.styles.ts
```diff
@@ -0,0 +1,7 @@
+import { css } from '@leafygreen-ui/emotion';
+import { spacing } from '@leafygreen-ui/tokens';
+
+export const verifiedAnswerBannerStyles = css`
+  margin-top: -${spacing[200]}px;
```
TheSonOfThomp, line 5: When using negative margin, we recommend leaving a comment why, since it's not super intuitive

TheSonOfThomp (commented):
Almost LGTM. Just the change type, and some minor code tweaks 👍

.changeset/odd-panthers-ring.md
nlarew: Ah interesting. Are we cool with releasing a major for Banner alongside this?

chat/message/src/MessageBanner/VerifiedAnswerBanner.styles.ts
nlarew: Circled back with Alven and turns out this was from an older design - removed the negative margin entirely!

chat/message/src/MessageBanner/MessageBanner.styles.ts
nlarew: Good call - I removed them from banner and there's no changes in Chromatic so should be good to go.

.changeset/odd-panthers-ring.md
TheSonOfThomp: Yeah, that's fine

TheSonOfThomp (approved):
LGTM !

PR #2301, opened 2024-04-09
LG-4134 | LG-4142: [CCP] NumberInput updates

stephl3:
## ✍️ Proposed changes

- updated styling
  - spacing changes in the `UnitSelect` component are made in Select: https://github.com/mongodb/leafygreen-ui/pull/2313
- adds default `errorMessage`
- adds `valid` state with `successMessage` prop

### For new components

## 🧪 How to test changes

Figma: https://www.figma.com/file/WmbvStSle8GtLg5RqEt2oh/Component-Consistency-Project?type=design&node-id=45%3A1052&mode=design&t=skYRN9GW2RzH3nJk-1


packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -142,16 +144,17 @@ export const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
...
-          >
-            {renderErrorMessage && (
-              <Error className={errorMessageStyles}>{errorMessage}</Error>
-            )}
-          </div>
+          {showError && (
+            <div
+              id={errorMessageId}
+              className={errorContainerStyles}
+              aria-live="polite"
```
shaneeza, line 151: I believe that this div was outside of the conditional check because of `aria-live`. Can you check that the error message is still announced on VoiceOver?

stephl3: Tested and confirmed that it wasn't announcing as the errorMessage rendered. I moved the conditional 1 node down

packages/number-input/src/Input/Input.tsx
```diff
@@ -60,8 +54,8 @@ export const Input = React.forwardRef<HTMLInputElement, InputProps>(
      */
     const translateTimeout = useRef<NodeJS.Timeout>();
     const containerRef = useRef<HTMLDivElement | null>(null);
-    const [shouldErrorTransition, setShouldErrorTransition] =
-      useState<boolean>(false);
+    // const [shouldErrorTransition, setShouldErrorTransition] =
```
bruugey, line 57: rm?

stephl3: oops removed!

packages/number-input/src/NumberInput/NumberInput.styles.ts
```diff
@@ -47,17 +56,23 @@ export const unitDisabledStyles: Record<Theme, string> = {
   `,
 };
 
-export const errorMessageWrapperStyles = css`
-  font-size: ${typeScales.body1.fontSize}px;
-  line-height: ${typeScales.body1.lineHeight}px;
+export const errorContainerStyles = css`
+  margin-top: ${spacing[100]}px;
```
shaneeza, line 60: Since we always show the error container, the margin-top is always there, even if there is no error text. Can we update this so it only shows up when there is text?

packages/number-input/src/NumberInput/NumberInput.tsx
shaneeza: Does `FormField` use this pattern? if not, we should consider adding it.

shaneeza (commented):
Should the input text size also increase when `size` is `large`? It looks like its still `13px` while the `select` next to it is `18px`
Screenshot 2024-04-09 at 6 18 50 PM

stephl3:
> Should the input text size also increase when `size` is `large`? It looks like its still `13px` while the `select` next to it is `18px` Screenshot 2024-04-09 at 6 18 50 PM

ahh yes still waiting to hear back from design on this font. will push an update tomorrow

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -143,13 +146,18 @@ export const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
...
             aria-live="polite"
             aria-relevant="all"
-            id={errorMessageId}
           >
-            {renderErrorMessage && (
-              <Error className={errorMessageStyles}>{errorMessage}</Error>
+            {showError && (
+              <>
+                <Icon glyph="Warning" fill={errorIconThemeStyle[theme]} />
+                <Error>{errorMessage}</Error>
```
shaneeza, line 159: If any empty string is passed, will we only show the icon?

shaneeza (commented):
```
some styling will be addressed by changes made in Select: https://github.com/mongodb/leafygreen-ui/pull/2313
```
what changes?

stephl3:
> ```
> some styling will be addressed by changes made in Select: https://github.com/mongodb/leafygreen-ui/pull/2313
> ```
> 
> what changes?

updated description. The spacing changes for `UnitSelect` are required for `Select` so I made those changes directly to the select component

shaneeza (commented):
Screenshot 2024-04-10 at 5 31 18 PM
is the `Select` component also going to have this font size when its `large`?

stephl3:
> Screenshot 2024-04-10 at 5 31 18 PM is the `Select` component also going to have this font size when its `large`?

@shaneeza yes confirmed this with @alvenw

.changeset/chatty-socks-occur.md
```diff
@@ -0,0 +1,13 @@
...
+
+[LG-4134](https://jira.mongodb.org/browse/LG-4134)
+
+1. Updated styling:
+  - updated spacing for `'small'` and `'xsmall'` size variants
+  - added `'large'` variant
+  - updated placeholder text color
+  - moved error icon from inside the input to underneath the input, alongside the error message
+
+2. A default `errorMessage` of `'This input needs your attention'` will render below number input when state is invalid.
```
bruugey, line 13: Yeah - this belongs somewhere as a const.

.changeset/chatty-socks-occur.md
```diff
@@ -0,0 +1,13 @@
+---
+'@leafygreen-ui/number-input': major
+---
+
+[LG-4134](https://jira.mongodb.org/browse/LG-4134)
+
+1. Updated styling:
```
bruugey, line 7: Not for now, but we should write some guidelines for changesets. (Tense consistency, when to use ols/uls, etc.) 

Maybe you have a POV here

packages/number-input/README.md
```diff
@@ -62,28 +62,28 @@ or
...
-| `unit`            | `string`                                               | The string unit that appears to the right of the input if using a single unit. </br> </br>Required if using `unitOptions`. When using `unitOptions` this value becomes the controlled value of the select input.                                 | `default` |
-| `unitOptions`     | `Array<{displayName: string; value: string}>`          | The options that appear in the select element attached to the right of the input.                                                                                                                                                                | `default` |
+| Prop              | Type                                                   | Description                                                                                                                                                                                                                                      | Default                             |
```
bruugey, line 65: Is there a way to use our tsdocs to generate this table for us?

packages/number-input/src/Arrows/Arrows.types.ts
```diff
@@ -16,8 +16,14 @@ export interface ArrowsProps {
    */
   onKeyDown: (e: React.KeyboardEvent) => void;
 }
+export interface ArrowsProps extends BaseArrowProps {
+  /**
+   * Determines horizontal positioning of arrows
```
bruugey, line 21: can we just add size  to arrow props?

packages/number-input/src/Input/Input.styles.ts
```diff
@@ -43,36 +42,16 @@ export const inputBaseStyles = css`
...
-  .${wrapperClassName}:focus-within & {
-    padding-right: ${spacing[1] * 6}px;
-  }
-`;
-
 export const inputThemeStyles: Record<Theme, string> = {
   [Theme.Light]: css`
     &::placeholder {
-      color: ${palette.gray.light1};
+      color: ${palette.gray.base};
```
bruugey, line 54: this should be a token

packages/number-input/src/NumberInput/NumberInput.styles.ts
```diff
@@ -47,17 +56,26 @@ export const unitDisabledStyles: Record<Theme, string> = {
...
 `;
 
-export const labelDescriptionStyles = css`
-  margin-bottom: ${spacing[1]}px;
-  display: flex;
-  flex-direction: column;
-`;
+export const errorIconThemeStyle: Record<Theme, string> = {
+  [Theme.Light]: palette.red.base,
```
bruugey, line 70: tokens

packages/number-input/src/NumberInput/NumberInput.tsx
```diff
@@ -143,13 +146,18 @@ export const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
             )}
           </div>
           <div
-            className={errorMessageWrapperStyles}
+            id={errorMessageId}
+            className={cx(errorContainerStyles, {
+              [errorContainerMarginTop]: showError,
+            })}
```
bruugey, line 152: do the below aria-attributes match what we have in FormField? If not let's make them match, bc we want to make the eventual backport to use that component as seamless as possible.

.changeset/chatty-socks-occur.md
stephl3: I'm sure there's a good one written somewhere. I'll take a look around and share with the team at the next dev sync. I've mostly been writing based on how content reads/appears in the generated changelog files

packages/number-input/README.md
stephl3: yeah agreed, it would be nice to rely on the generated props table that exists on .design so we don't need to maintain this readme. I'm not clear on what the level of effort looks like to do so but would be happy to look further into this if we can allocate time to do so for all packages

packages/number-input/src/Arrows/Arrows.types.ts
stephl3: `size` is used to position the group of arrows rather than positioning each one separately which I think is less code, so I prefer to keep the prop on `Arrows` rather than drill further to each singular`Arrow`

packages/number-input/src/NumberInput/NumberInput.tsx
stephl3: made a shareable `FormFieldFeedback` component that will make sure the aria-attributes match across inputs: https://github.com/mongodb/leafygreen-ui/pull/2318/files#diff-226086e5a2e4eeb4bcf2ebfbbfc394693b53ea7cf5eff665b5afcd846637741aR55-R56

shaneeza:
@stephl3 it looks like the UI Review and UI Tests are failing

packages/number-input/src/UnitSelect/UnitSelect.styles.ts
```diff
@@ -11,23 +11,20 @@ export const wrapperBaseStyles = css`
...
+  }
+`;
 
 export const selectStyles = css`
   > div {
     display: flex;
   }
+
+  button {
+    font-size: 12px;
```
shaneeza, line 27: Do we have tokens for this and the line-height?

packages/number-input/src/NumberInput/NumberInput.types.ts
```diff
@@ -18,17 +19,15 @@ export const Direction = {
 
 export type Direction = (typeof Direction)[keyof typeof Direction];
 
-export const State = {
-  Error: 'error',
-  None: 'none',
-} as const;
+export const State = FormFieldState;
 
 export type State = (typeof State)[keyof typeof State];
```
shaneeza, line 24: Instead of doing this, could you import `type FormFieldState` and use that on line 102?

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -8,7 +8,6 @@ import { NumberInput } from '.';
 
 const label = 'This is the label text';
 const description = 'This is the description text';
-const errorMessage = 'error message';
 const arrowTestId = {
   up: 'lg-number_input-increment_button',
```
shaneeza, line 12: Could we make these consts?

shaneeza (approved):

packages/number-input/src/NumberInput/NumberInput.types.ts
stephl3: I think I'll leave this as-is for now since it's part of an existing pattern (albeit a slightly more verbose one that may need another look in the future)

packages/number-input/src/NumberInput/NumberInput.spec.tsx
stephl3: assuming the question is about the arrow test ids, I think I'll leave this as-is for now since it requires a further refactor of the Arrows/Arrow components. Maybe can get a refactor when adding test harnesses to this package

PR #2302, opened 2024-04-09
Run CI PR workflow on Merge Groups

bruugey:
Related Docs:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,8 @@
 name: Release
 
 on:
+  merge_group:
+    types: [checks_requested]
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,6 +1,8 @@
 name: Release
 
 on:
+  merge_group:
+    types: [checks_requested]
```
TheSonOfThomp, line 5: Do we still want to filter by branch here?

TheSonOfThomp: Nevermind, just read the docs

PR #2310, opened 2024-04-09
LG-4135 | LG-4151: [CCP] PasswordInput updates

stephl3:
## ✍️ Proposed changes

- updated styling
- added `errorMessage` and `successMessage` props

### For new components

## 🧪 How to test changes

Figma: https://www.figma.com/file/WmbvStSle8GtLg5RqEt2oh/Component-Consistency-Project?type=design&node-id=59%3A15339&mode=design&t=SrDvmdmTT2ixXqVC-1


packages/password-input/src/PasswordInput.stories.tsx
```diff
@@ -98,7 +98,6 @@ const meta: StoryMetaType<typeof PasswordInput> = {
     generate: {
       combineArgs: {
         darkMode: [false, true],
-        value: [undefined, 'password'],
```
stephl3, line 101: The generated story variants were reaching chromatic's limits. Please feel free to provide feedback if we should take a different approach
Screenshot 2024-04-09 at 5 40 21 PM

shaneeza: You can add a new story testing only the value. Is that what you wanted to do with this change?

.changeset/neat-kiwis-know.md
```diff
@@ -0,0 +1,18 @@
...
+[LG-4135](https://jira.mongodb.org/browse/LG-4135)
+
+1. Updated styling:
+  - updated spacing for `'small'` and `'large'` size variants
+  - added `'xsmall'` variant
+  - updated placeholder text color
+  - removed validation icon from inside the input
+
+2. Added `errorMessage` and `successMessage` props
+  - `errorMessage` will render a default of `'This input needs your attention'` and can be customized
```
shaneeza, line 14: Screenshot 2024-04-10 at 1 09 09 PM
the custom container story is missing the icon and default message, is that correct?

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -123,30 +135,32 @@ export const PasswordInput = React.forwardRef<
...
-            {!hasNotifications && state !== State.None && (
-              <InputIcon state={state} size={size} disabled={disabled} />
-            )}
             <PasswordToggle
               showPassword={showPassword}
               handlePasswordToggleClick={handlePasswordToggleClick}
               size={size}
             />
           </div>
-          {hasNotifications && (
```
shaneeza, line 144: This was only hidden if a custom container was used, which is why getting rid of `aria-live` didn't matter. Since we will always show a message, even if the consumer uses a custom container, we should keep a div with `aria-live` on the page.

packages/password-input/src/PasswordInput/PasswordInput.tsx
```diff
@@ -123,30 +135,32 @@ export const PasswordInput = React.forwardRef<
...
             <StateNotifications
               id={descriptionId}
               notifications={stateNotifications as Array<NotificationProps>}
             />
           )}
+          {showStateFeedback && (
+            <StateFeedback
+              id={descriptionId}
+              state={state}
+              errorMessage={errorMessage}
```
shaneeza, line 160: What happens if the consumer passes a blank string?

packages/password-input/src/PasswordInput.stories.tsx
stephl3: Once I added the general error/valid message below the input container, the number of unique combinations generated exceeded the limit. I don't think we needed to test the different variations with the password text pre-populated (similar to the pattern of other generated stories), so I went ahead and dropped that key

.changeset/neat-kiwis-know.md
stephl3: yes, this is intentional if a custom error is defined outside of the component. I've updated the story to include this

packages/password-input/src/PasswordInput/PasswordInput.tsx
stephl3: consumer can pass an empty string if they don't want an error message to render. In that case, only the warning icon will render

packages/password-input/src/StateFeedback/StateFeedback.tsx
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  return (
+    /**
+     * We're using aria-polite to announce when a message has changed. In order
+     * for aria-polite to work correctly the message wrapper needs to remain on
+     * the page even if there are no messages. If a custom message container is
+     * specified with aria-describedby then this wrapper will not render.
+     */
+    <div
```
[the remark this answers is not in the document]
stephl3: gotcha, I updated the logic to have them share a container

packages/password-input/src/PasswordInput/PasswordInput.tsx
stephl3: updated!

packages/password-input/src/PasswordInput.stories.tsx
```diff
@@ -175,12 +176,15 @@ export const CustomContainer = ({
 }: PasswordInputProps) => {
   const ref = useRef<HTMLInputElement>(null);
   return (
-    <PasswordInput
-      {...rest}
-      stateNotifications={stateNotifications as State}
-      aria-describedby={'my-id'}
-      ref={ref}
-    />
+    <div style={{ display: 'flex', flexDirection: 'column', gap: '4px' }}>
```
bruugey, line 179: 4px is spacing[1]

packages/password-input/src/StateFeedback/StateFeedback.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const containerStyles = css`
+  padding-top: ${spacing[100]}px;
+  display: flex;
+  align-items: center;
+  gap: ${spacing[100]}px;
+`;
+
+export const errorIconThemeStyle: Record<Theme, string> = {
+  [Theme.Light]: palette.red.base,
```
bruugey, line 14: tokens

packages/password-input/src/StateFeedback/StateFeedback.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  [Theme.Dark]: palette.red.light1,
+};
+
+export const validIconThemeStyle: Record<Theme, string> = {
+  [Theme.Light]: palette.green.dark1,
```
bruugey, line 19: tokens

packages/password-input/src/StateFeedback/StateFeedback.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  [Theme.Dark]: palette.green.base,
+};
+
+export const validMessageThemeStyle: Record<Theme, string> = {
+  [Theme.Light]: css`
+    color: ${palette.green.dark1};
```
bruugey, line 25: tokens

packages/password-input/src/StateFeedback/StateFeedback.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  `,
+  [Theme.Dark]: css`
+    color: ${palette.green.base};
```
bruugey, line 28: tokens

.changeset/neat-kiwis-know.md
shaneeza: i'm curious, why we don't want to show the icon anymore?

packages/password-input/src/PasswordInput/PasswordInput.spec.tsx
```diff
@@ -231,49 +231,34 @@ describe('packages/password-input', () => {
...
 
-    test('input shows error(x) icon', async () => {
-      const { container } = renderPasswordInput({
-        [`aria-describedby`]: defaultProps.ariaDescribedby,
+  describe('without stateNotifications array or aria-describedby', () => {
+    test(`it renders a general error message when stateNotifications is 'error'`, async () => {
+      const { queryByText } = renderPasswordInput({
         stateNotifications: 'error',
         label: defaultProps.label,
+        errorMessage: defaultProps.errorMessage,
```
shaneeza, line 241: Is there anything preventing a consumer from passing the `successMessage` even if `stateNotification` is `error`? The test name implies that only the error message will render if the state is error.

shaneeza (changes requested):
We've spoken in person about this but I'll leave a note on this here as well, but its seems a little odd that if `stateNotifications` is a `string`, instead of an `array`, you can now either have a custom container or show the built-in `error/success` prop. 
I think the assumption is that if a team isn't using the built-in `stateNotification` array then it means that they are using a custom message container, which could contain an ordered/unordered list of messages. 

But if we're going with this then we should try to update the TS conditionals so that aria-describedby is required if the state is a string and error/success message is empty

packages/password-input/src/PasswordInput/PasswordInput.spec.tsx
stephl3: no, a consumer could additionally pass a `successMessage`, and there's always a default set for both. the state will determine which message will render

stephl3:
> We've spoken in person about this but I'll leave a note on this here as well, but its seems a little odd that if `stateNotifications` is a `string`, instead of an `array`, you can now either have a custom container or show the built-in `error/success` prop. I think the assumption is that if a team isn't using the built-in `stateNotification` array then it means that they are using a custom message container, which could contain an ordered/unordered list of messages.
> 
> But if we're going with this then we should try to update the TS conditionals so that aria-describedby is required if the state is a string and error/success message is empty

@shaneeza let me know if the updated types and api for `PasswordInputFeedback` makes sense

packages/password-input/src/PasswordInput/PasswordInput.types.ts
```diff
@@ -43,7 +43,22 @@ interface StateOnlyProps {
   stateNotifications: State;
 }
 
-// using default notification container, no aria-describedby
+// Using general notification container, no aria-describedby
+interface GeneralNotificationProps {
```
shaneeza, line 47: Can you add types tests for these in the spec file? We already have some type checks at the bottom of the file. I'm curious what TS will say if you pass a string to `stateNotifications`

shaneeza (changes requested):
@alvenw left a comment in the UI test

packages/password-input/src/PasswordInput/PasswordInput.types.ts
stephl3: https://github.com/mongodb/leafygreen-ui/pull/2310/commits/16e2dcab8c5df02878f77f38faa96157026ba188

stephl3:
> @alvenw left a comment in the UI test
ahh sorry @alvenw I must have missed that. This component has a special case in which we allow consumers to define an error message outside of the conventional icon+message format. For example, it could be a banner that has an id, and that id is passed to the password input through the `aria-describedby` prop

I wanted to use a `Banner` with `variant="danger"` to better illustrate this storybook instance, but it didn't seem worthwhile to bump up the package's bundlesize just to add a single banner to the storybook. Instead, I opted to render just an error string which is why there is no icon

I updated the readme to indicate what a custom container could look like in code using a banner component: https://github.com/mongodb/leafygreen-ui/pull/2310/files#diff-de451f127d6661845f8a9fd8307e82b2af30059df5cb835cd8f59f5fa9277fd4R70-R79

cc: @shaneeza

packages/password-input/src/PasswordInput/PasswordInput.spec.tsx
```diff
@@ -303,6 +303,21 @@ describe('packages/password-input', () => {
...
+      <PasswordInput
+        data-attribute="data test"
+        aria-labelledby="label"
+        stateNotifications="valid"
+      />
+      <PasswordInput
+        data-attribute="data test"
+        aria-labelledby="label"
+        stateNotifications="warning"
+      />
```
shaneeza, line 320: I'm not sure if this is the correct behavior. If someone passes `stateNotifications` and uses a custom container, TS should require `aria-describedby`. 
It looks like i never actually added a test to test this but I was able to get the error in `main` with this test.

Screenshot 2024-04-24 at 2 51 40 PM

Does that make sense?

stephl3: @shaneeza yes that makes sense. I think previously ts would error if `aria-describedby` wasn't defined and `stateNotifications` was defined as a `State` value. In this iteration of the password input, ts shouldn't error since the new default will be the `FormFieldFeedback` with the icon + error/success message

This change still retains backwards compatibility for cases that use a custom container with `aria-describedby` or `stateNotifications` with an array of notifications. Do you think we need to make further adjustments to the types and if so, what does that look like?

shaneeza: i think that if we no longer require it, then no ones really going to use `aria-describedby`, unfortunately. Do you wanna bring this up during dev sync tomorrow and get some other opinions?

stephl3: @shaneeza If `aria-describedby` isn't defined, the default `FormFieldFeedback` will render in addition to the custom container. The additional default seems like it would provide enough visual feedback for the consumer to understand that `aria-describedby` needs to be used in order to remove the default. I feel good with this approach as-is but can discuss in dev sync if you think it warrants broader discussion

shaneeza: I see, that clears things up a bit. Can you make it clear in the README that you need `aria-describedby` to remove the default message?

shaneeza (approved):

PR #2317, opened 2024-04-17
`color` token use `satistfies`

TheSonOfThomp:
## ✍️ Proposed changes

Adds `satisfies` to ensure consistent & correct `colors` token structure


stephl3 (approved):
LGTM

PR #2313, opened 2024-04-10
LG-4138 | LG-4150: [CCP] Select updates

stephl3:
## ✍️ Proposed changes

- updates styling
- adds default `errorMessage`
- adds `valid` state with `successMessage` prop

🎟 _Jira ticket:_ [LG-4138](https://jira.mongodb.org/browse/LG-4138)

### For bug fixes, new features & breaking changes

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `yarn changeset` and documented my changes

### For new components

- [ ] I have added my new package to the global tsconfig
- [ ] I have added my new package to the Table of Contents on the global README
- [ ] I have verified the Live Example looks as intended on the design website.

## 🧪 How to test changes

Figma: https://www.figma.com/file/WmbvStSle8GtLg5RqEt2oh/Component-Consistency-Project?type=design&node-id=59%3A15314&mode=design&t=Wf5VD9qqs9EmPGXy-1
Review storybook and chromatic

PR #2318, opened 2024-04-19
Integration: LG-2930: [CCP] input components

stephl3:
## ✍️ Proposed changes

tokens
- Fixes `darkModeBackgroundColors` disabled colors
- Fixes `lightModeBorderColors` variable typo

typography
- Updates typography components to use tokens where possible

form-field
- Adds and exports `FormFieldFeedback` component
- Adds and exports `DEFAULT_MESSAGES` constant
- Updates `FormField` and `FormFieldInputContainer` components to use tokens where possible
- Refactors `FormField` to use `FormFieldFeedback`

date-picker
- Updates test id for test specs involving error message

combobox
- Updated styling
- New default `errorMessage`
- New `successMessage` prop with default
- Disabled state changes

number-input
- Updated styling
  - Spacing changes in the `UnitSelect` component are made in Select: https://github.com/mongodb/leafygreen-ui/pull/2313
- Adds default `errorMessage`
- Adds `valid` state with `successMessage` prop

password-input
- Updated styling
- Added `errorMessage` and `successMessage` props

select
- Updates styling
- Adds default `errorMessage`
- Adds `valid` state with `successMessage` prop

🎟 _Jira ticket:_ [LG-2930](https://jira.mongodb.org/browse/LG-2930)

### For new components

## 🧪 How to test changes


packages/form-field/src/FormField/FormField.tsx
```diff
@@ -46,8 +39,8 @@ export const FormField = forwardRef<HTMLDivElement, FormFieldProps>(
       state = FormFieldState.None,
       size = Size.Default,
       disabled = false,
-      errorMessage,
-      successMessage,
+      errorMessage = DEFAULT_MESSAGES.error,
+      successMessage = DEFAULT_MESSAGES.success,
```
stephl3, line 43: I previously thought we might want to customize a different default `errorMessage` or `successMessage` depending on the input component, but upon further reflection, I think it's fine to set this broadly at the form field level. Open to feedback if we'd prefer to define at the wrapping component level (text input, combobox, etc) since I'm also exporting `DEFAULT_MESSAGES` from the form field package

packages/form-field/src/FormFieldFeedback/FormFieldFeedback.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+  gap: ${spacing[100]}px;
+`;
+
+export const spacingTop = css`
+  padding-top: ${spacing[100]}px;
+`;
+
+export const hideContainerStyle = css`
+  opacity: 0;
+`;
```
stephl3, line 15: this is conditionally added for cases where a popover may render below an input container but not fully cover the feedback message. adding opacity: 0 will effectively erase the feedback when the menu is open

packages/form-field/src/FormFieldInputContainer/FormFieldInputContainer.styles.ts
```diff
@@ -92,30 +92,24 @@ export const inputWrapperModeStyles: Record<Theme, string> = {
       }
     }
   `,
+  /** token exceptions: background-color value was designated prior to token system */
   [Theme.Dark]: css`
-    color: ${palette.gray.light3};
+    color: ${color.dark.text.primary.default};
     background-color: ${palette.gray.dark4};
```
stephl3, line 98: at this point in standardizing components, we'll intentionally continue to leave the `palette.gray.dark4` even though it's not in the token system until design decides on how to proceed

.changeset/proud-bugs-smile.md
```diff
@@ -0,0 +1,8 @@
+---
+'@leafygreen-ui/tokens': minor
+---
+
+[LG-2930](https://jira.mongodb.org/browse/LG-2930)
+
+- Fixes `darkModeBackgroundColors` disabled colors
```
stephl3, line 7: made the assumption that this is still a minor change because it's an aesthetic change but please feel free to correct me if this should instead be major

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -53,6 +52,10 @@ import { DatePicker } from '.';
  * Generally, this type of tests should _only_ exist in a test file for user-facing components.
  */
 
+const TEST_IDS = {
+  errorMessage: 'lg-form_field-feedback-error_message',
```
shaneeza, line 56: could we export this from `form-field`?

packages/form-field/src/FormField.spec.tsx
```diff
@@ -10,6 +10,10 @@ import {
   useFormFieldContext,
 } from '.';
 
+const TEST_IDS = {
+  feedback: 'lg-form_field-feedback',
```
shaneeza, line 14: same here

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
stephl3: The idea of introducing a pattern of exporting internal test ids from each package seems iffy to me. Is there an internal directory we can rely on that doesn't involve exporting directly from the `form-field` package?

What do you think @bruugey @TheSonOfThomp ?

shaneeza: i'd imagine that we would want to use this test id inside a test harness package. We're currently importing the test ids from form-field inside other packages.

packages/form-field/src/FormFieldFeedback/FormFieldFeedback.tsx
```diff
@@ -0,0 +1,86 @@
...
+    >
+      {showFormFieldFeedback && (
+        <>
+          <div className={cx(iconWrapperStyles, getIconWrapperHeight(size))}>
+            <Icon {...iconProps} aria-hidden />
+          </div>
+          {isErrorState ? (
+            <Error
+              data-lgid={LGIDS_FORM_FIELD.errorMessage}
+              data-testid="lg-form_field-feedback-error_message"
```
shaneeza, line 66: `data-lgid` and `data-testid` should be the same. I missed updating`data-testid` to be `LGIDS_FORM_FIELD.errorMessage` while i was working on test-harnesses. The difference between the two is that data-testid was there first and is probably used by consumers and `data-lgid` is used inside our test-harnesses. If we're sticking with exporting our test ids, we should update `LGIDS_FORM_FIELD.errorMessage` to be `lg-form_field-feedback-error_message`.

packages/date-picker/src/DatePicker/DatePicker.spec.tsx
stephl3: oh gotcha, I think I can use the one that's already being exported in place of this

packages/form-field/src/FormFieldFeedback/FormFieldFeedback.tsx
stephl3: thanks! this is helpful context. I will keep them consistent with what was previously being used since consumers may be relying on that in their tests

.changeset/afraid-doors-knock.md
```diff
@@ -0,0 +1,10 @@
+---
+'@leafygreen-ui/form-field': minor
+---
+
+[LG-2930](https://jira.mongodb.org/browse/LG-2930)
+
+- Adds and exports `FormFieldFeedback` component
+- Adds and exports `DEFAULT_MESSAGES` constant
```
TheSonOfThomp, line 8: I don't love this name, but can't think of anything better

.changeset/proud-bugs-smile.md
TheSonOfThomp: I'd even say patch. It's just a bug fix, and doesn't affect API or layout

packages/form-field/src/FormField.spec.tsx
```diff
@@ -294,123 +295,128 @@ describe('packages/form-field', () => {
...
-      const errorIcon = queryByTitle('Error');
-      const errorText = queryByText('This is an error message');
-      expect(errorIcon).toBeNull();
-      expect(errorText).toBeNull();
+      const input = getByTestId('input');
+      const description = queryByText('Description');
+      const feedback = getByTestId(LGIDS_FORM_FIELD.feedback);
+      expect(input).toHaveAttribute(
+        'aria-describedby',
+        description?.id + ' ' + feedback?.id,
```
TheSonOfThomp, line 353: If we can test this in a way where the order of ids doesn't matter, that'd be better (this is technically testing a subtle implementation detail)

packages/form-field/src/FormField.spec.tsx
```diff
@@ -294,123 +295,128 @@ describe('packages/form-field', () => {
...
-  describe(`${FormFieldState.Valid} state`, () => {
-    test('Success icon and message are not shown by default', () => {
-      const { queryByText, queryByTitle } = render(
+  describe(`when state is not ${FormFieldState.None}`, () => {
+    test('form field feedback has an id', () => {
+      const { getByTestId } = render(
         <FormField
           label="Label"
           successMessage="Success"
+          state={FormFieldState.Valid}
```
TheSonOfThomp, line 304: Isn't the original `describe` title still accurate here?

packages/form-field/src/FormField/FormField.tsx
TheSonOfThomp: Seems the designs all have the same messages. We can update the default messages per-component if we need to as well

packages/form-field/src/FormField/FormField.tsx
```diff
@@ -56,16 +49,20 @@ export const FormField = forwardRef<HTMLDivElement, FormFieldProps>(
...
-    const isSuccessState = state === FormFieldState.Valid;
-    const showStateFeedback = (isErrorState || isSuccessState) && !disabled;
+    const formFieldFeedbackProps = {
+      disabled,
+      errorMessage,
+      id: feedbackId,
+      size,
+      state,
+      successMessage,
+    } as const;
```
TheSonOfThomp, line 65: Maybe a topic for later, but it may be worth considering adding most of these props to a FormFieldContext?

TheSonOfThomp: Actually, most of these are already on the provider. Could we pull these from the provider in `FormFieldFeedback` instead of passing them as props?

TheSonOfThomp: (That does make it harder to test though, to take it or leave it)

packages/form-field/src/FormField/useFormFieldProps.ts
```diff
@@ -32,16 +32,17 @@ export const useFormFieldProps = ({
   const descriptionId = useIdAllocator({
     prefix: 'lg-form-field-description',
   });
-  const errorId = useIdAllocator({ prefix: 'lg-form-field-description' });
+  const feedbackId = useIdAllocator({ prefix: 'lg-form-field-description' });
```
TheSonOfThomp, line 35: `lg-form_field-description`

packages/form-field/src/FormField/useFormFieldProps.ts
```diff
@@ -32,16 +32,17 @@ export const useFormFieldProps = ({
...
   const generatedInputId = useIdAllocator({ prefix: 'lg-form-field-input' });
```
TheSonOfThomp, line 36: `lg-form_field-input`

packages/form-field/src/FormFieldFeedback/FormFieldFeedback.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const iconWrapperStyles = css`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`;
+
+export const getIconWrapperHeight = (size: Size) => {
+  return css`
+    height: ${size === Size.Large ? 24 : typeScales.body1.lineHeight}px;
```
TheSonOfThomp, line 25: Is `typeScales.body2.lineHeight` not `24`?

packages/form-field/src/FormField/FormField.styles.ts
```diff
@@ -1,6 +1,4 @@
 import { css } from '@leafygreen-ui/emotion';
-import { Theme } from '@leafygreen-ui/lib';
-import { palette } from '@leafygreen-ui/palette';
 import { BaseFontSize, Size, spacing, typeScales } from '@leafygreen-ui/tokens';
 
 export const getFontSize = ({
```
TheSonOfThomp, line 4: I know this isn't changed in this PR, but this Fn name should be `getFontSizeStyles` (as-is it seems like the func would return just a number)

packages/form-field/src/FormFieldFeedback/FormFieldFeedback.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const baseFontSize = useUpdatedBaseFontSize();
+  const fontStyles = getFontSize({ baseFontSize, size });
+
+  const isErrorState = state === FormFieldState.Error;
+  const isValidState = state === FormFieldState.Valid;
+  const showFormFieldFeedback = (isErrorState || isValidState) && !disabled;
+
+  const iconProps = {
+    glyph: isErrorState ? 'Warning' : 'Checkmark',
+    fill: color[theme].icon[isErrorState ? 'error' : 'success'].default,
```
TheSonOfThomp, line 42: Could we just refer directly to the `state` property? 
 `color[theme].icon[state].default`

packages/form-field/src/FormFieldInputContainer/FormFieldInputContainer.styles.ts
```diff
@@ -55,33 +56,32 @@ export const inputWrapperBaseStyles = css`
...
-    border: 1px solid ${palette.gray.base};
-
-    &:hover,
-    &:active {
-      &:not(:focus) {
-        box-shadow: ${hoverRing.light.gray};
-      }
-    }
+    color: ${color.light.text.primary.default};
+    background: ${color.light.background.primary.default};
```
TheSonOfThomp, line 60: maybe not worth rewriting in this PR, but now that we have the `colors` token, we can consolidate this into a single function:

```ts
export const getInputWrapperModeStyles = (theme: Theme) => css`
    color: ${color[theme].text.primary.default};
    background: ${color[theme].background.primary.default};
`
```

packages/form-field/src/FormFieldInputContainer/FormFieldInputContainer.styles.ts
TheSonOfThomp: Hmmm. Might be premature to tokenize these then. This could potentially change to specific `input` color tokens

TheSonOfThomp (approved):

packages/form-field/src/FormField/FormField.tsx
stephl3: I think I'll leave it as-is for now since `FormFieldFeedback` will at times be used outside of a `FormFieldContext`

packages/form-field/src/FormFieldInputContainer/FormFieldInputContainer.styles.ts
stephl3: I started applying this pattern where I could. Unfortunately, until we get design to align on how to handle dark mode backgrounds using gray.dark4 which is currently the value of `color.dark.background.secondary.default` rather than `color.dark.background.primary.default`, I think it'll be better to retain the `Theme` object style for these cases

We've got this design ticket to figure out how to handle those background tokens: https://jira.mongodb.org/browse/LG-4226

packages/form-field/src/FormField.spec.tsx
stephl3: https://github.com/mongodb/leafygreen-ui/pull/2318/commits/3b444b2d490b74792fc8d9972616b1d6681d4db1#diff-96f34579a072b062c22857b1e854bfefef73318f69c1d41de310cfb2d77bdb45L353-R357

packages/form-field/src/FormField/useFormFieldProps.ts
stephl3: https://github.com/mongodb/leafygreen-ui/pull/2318/commits/3b444b2d490b74792fc8d9972616b1d6681d4db1

packages/form-field/src/FormField/useFormFieldProps.ts
stephl3: https://github.com/mongodb/leafygreen-ui/pull/2318/commits/3b444b2d490b74792fc8d9972616b1d6681d4db1

packages/form-field/src/FormFieldFeedback/FormFieldFeedback.styles.ts
stephl3: thanks for syncing offline on this. I've corrected the `typeScales` token and swapped this with the corrected `typeScales` token

packages/form-field/src/FormFieldFeedback/FormFieldFeedback.tsx
stephl3: https://github.com/mongodb/leafygreen-ui/pull/2318/commits/3b444b2d490b74792fc8d9972616b1d6681d4db1#diff-c731bce236af5cb475ed9909a79e3033cc4fd1d9119accb38c1d426b377f70eaR41-R52
couldn't just directly key into `state` because `FormFieldState` uses `'valid'` while color variant uses `'success'`. made a style util that converts the `FormFieldState` into a color variant

bruugey (approved):

PR #2328, opened 2024-04-30
(EAI-71) [UI] Rich Links in LG Chat

nlarew:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [(EAI-71) [UI] Rich Links in LG Chat](https://jira.mongodb.org/browse/EAI-71)

_Figma:_ https://www.figma.com/file/L41rvEwQmJJ8VzO4hCPc7Y/LeafyGreen-Chat-Library?type=design&node-id=1%3A492&mode=design&t=rznAQHxznYGhO56X-1

### For new components


.changeset/angry-hornets-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-chat/message': minor
```
shaneeza, line 2: Can you add a changeset for rich links?

chat/message/README.md
```diff
@@ -42,16 +42,18 @@ return (
...
+| `messageBody`        | `string`                                                             | Message body text passed to LGMarkdown                                            |                                                                                                                           |
+| `verified`           | `{ verifier?: string; verifiedAt?: Date; learnMoreUrl?: string; }`   | Sets if an answer is "verified" and controls the content of the message banner.   |                                                                                                                           |
+| `links`              | `{ url: string; text: string; imageUrl?: string; variant: string; }` | A list of links to show in a section at the end of the message.                   |                                                                                                                           |
```
shaneeza, line 57: Can you list what the variants are?

chat/message/src/Message/Message.stories.tsx
```diff
@@ -130,43 +134,97 @@ VerifiedAnswer.args = {
     verifier: 'MongoDB Staff',
     verifiedAt: new Date('2023-08-24T16:20:00Z'),
   },
-  // @ts-ignore onChange is passed in the story itself
-  children: <MessageFeedbackStory />,
+  children: <MessageFeedback />,
+};
+
+function useCorrectBaseFontSize() {
```
shaneeza, line 140: We also have a [util](https://github.com/mongodb/leafygreen-ui/blob/main/packages/typography/src/utils/useUpdatedBaseFontSize.ts) to help with this.

chat/message/src/Message/Message.styles.ts
```diff
@@ -11,7 +11,7 @@ export const avatarClassName = createUniqueClassName('lg-message-avatar');
 
 export const baseStyles = css`
   display: flex;
-  gap: ${spacing[2]}px;
+  gap: ${spacing[200]}px;
```
shaneeza, line 14: 👍🏽

chat/message/src/MessageContainer/MessageContainer.styles.ts
```diff
@@ -14,6 +14,10 @@ export const baseStyles = css`
   box-shadow: 0px 4px 10px -4px ${palette.black}4D; // 4D is 30% opacity
 
   position: relative;
+
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
```
shaneeza, line 20: Is there a token for this?

chat/message/src/MessageLinks/MessageLinks.styles.ts
```diff
@@ -0,0 +1,23 @@
+import { css } from '@leafygreen-ui/emotion';
+import { Theme } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+import { spacing } from '@leafygreen-ui/tokens';
+
+export const baseStyles = css`
+  container-type: inline-size;
```
shaneeza, line 7: Do you know what browsers we want to support and if this is compatible with those browsers?

chat/message/src/MessageLinks/MessageLinks.styles.ts
```diff
@@ -0,0 +1,23 @@
...
+  [Theme.Dark]: css`
+    border: 1px solid ${palette.gray.dark2};
+  `,
+  [Theme.Light]: css`
+    border: 1px solid ${palette.gray.light2};
+  `,
+};
+
+export const linksHeadingStyles = css`
+  margin-bottom: 8px;
```
shaneeza, line 21: Token?

chat/message/src/MessageLinks/MessageLinks.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import {
+  baseStyles,
+  dividingLineStyles,
+  linksHeadingStyles,
+} from './MessageLinks.styles';
+import { type MessageLinksProps } from './MessageLinks.types';
+
+export function MessageLinks({
+  darkMode: darkModeProp,
```
shaneeza, line 16: If this component is nested inside another component that is wrapped in `<LeafyGreenProvider darkMode={darkMode}>` then you don't need `darkMode` as a prop. The `useDarkMode` hook on L21 should read the darkMode value from `LeafyGreenProvider`

chat/rich-links/README.md
```diff
@@ -0,0 +1,19 @@
+# Rich Links
```
shaneeza, line 1: Can you add the props to the README?

chat/rich-links/src/RichLink/RichLink.spec.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import { RichLinks } from '.';
+
+describe('packages/rich-links', () => {
```
shaneeza, line 6: Is there anything to test?

chat/rich-links/src/RichLink/RichLink.styles.ts
```diff
@@ -0,0 +1,57 @@
...
+    background-color: ${palette.gray.light3};
+
+    &:hover {
+      box-shadow: 0 0 0 3px ${palette.gray.light2};
+    }
+  `,
+};
+
+export const badgeAreaStyles = css`
+  padding-bottom: calc(18px + 2 * ${spacing[200]}px);
```
shaneeza, line 42: Can you explain the calculation?

chat/message/src/MessageLinks/MessageLinks.tsx
shaneeza: However, if this component is intended to also be used outside of `<Message>`, as a standalone component, then that's understandable

chat/rich-links/src/RichLink/RichLink.types.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Variant as RichLinkBadgeVariantName } from './RichLinkBadge/RichLinkBadge.types';
+import { RichLinkVariantName } from './RichLinkVariants';
+
+export interface BaseRichLinkProps
+  extends HTMLElementProps<'a', HTMLAnchorElement>,
+    DarkModeProps {
+  /**
+   * The URL that the rich link points to
+   */
+  url: string;
```
shaneeza, line 12: Does this component only support external links?

chat/rich-links/src/RichLink/RichLink.types.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * A URL for the background image of the rich link
+   */
+  imageUrl?: string;
+
+  /**
+   * Props to be spread on the anchor element
+   */
+  anchorProps?: Pick<JSX.IntrinsicElements['a'], 'target' | 'rel'>;
```
shaneeza, line 27: I don't fully understand this prop. On L7, you're extending anchor props, which means you can pass `target` or `rel` outside of this prop. Can you explain the reason for this?

chat/rich-links/src/RichLink/RichLink.types.ts
nlarew: For our use case we'll be point to URLs that may be on the docs, dev center, youtube, etc

Was there an alternative you were thinking of here?

chat/rich-links/src/RichLink/RichLink.tsx
shaneeza, line 1: If I understand correctly, you can either pass the variant or skip the variant and pass the badge props; can you explain the reasoning for this?

chat/rich-links/src/RichLink/RichLinkBadge/RichLinkBadge.tsx
```diff
@@ -0,0 +1,134 @@
...
+}: {
+  children: React.ReactNode;
+  color?: string;
+}) => {
+  return <div style={{ position: 'relative', color }}>{children}</div>;
+};
+
+export interface RichLinkBadgesProps {}
+
+export const RichLinkBadges = (props: RichLinkBadgesProps) => {
```
shaneeza, line 24: Is this used or exported anywhere?

chat/rich-links/src/RichLink/RichLink.styles.ts
```diff
@@ -0,0 +1,57 @@
+import { css } from '@leafygreen-ui/emotion';
+import { Theme } from '@leafygreen-ui/lib';
+import { createUniqueClassName } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+import { spacing } from '@leafygreen-ui/tokens';
+
+export const richLinkTextClassName = createUniqueClassName('lg-chat-rich-link');
+
+export const baseStyles = css`
```
shaneeza, line 9: will the height always be 64px?

chat/rich-links/src/RichLink/RichLinkBadge/RichLinkBadge.styles.ts
```diff
@@ -0,0 +1,139 @@
+import { css } from '@leafygreen-ui/emotion';
+import { Theme } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+import { fontFamilies, fontWeights } from '@leafygreen-ui/tokens';
+
+import { Variant } from './RichLinkBadge.types';
+
+export const baseStyles = css`
```
shaneeza, line 8: Can you add tokens where applicable?

chat/rich-links/src/RichLink/RichLinkBadge/RichLinkBadge.styles.ts
```diff
@@ -0,0 +1,139 @@
...
+  font-family: ${fontFamilies.default};
```
shaneeza, line 9: You might not need to specify the font family and font weight, those should already be defined in `Card`.

chat/rich-links/src/RichLink/RichLinkBadge/RichLinkBadge.types.ts
```diff
@@ -0,0 +1,41 @@
...
+  Gray: 'gray',
+  Blue: 'blue',
+  Green: 'green',
+  Purple: 'purple',
+  Red: 'red',
+  Yellow: 'yellow',
+} as const;
+export type Variant = (typeof Variant)[keyof typeof Variant];
+
+export interface RichLinkBadgeProps
```
shaneeza, line 16: Could you extend the props from `RichLink.types.ts` instead of rewriting them?

chat/rich-links/src/RichLink/RichLink.types.ts
shaneeza: I'm not sure what other teams might use this component, but there is a chance that if someone wants to use an internal link and they're using nextjs, we would need to support the `as` prop.

shaneeza (changes requested):
Looking good. There were some choices that were unclear to me, so I left a few questions.

alvenw (approved):

chat/message/src/MessageLinks/MessageLinks.styles.ts
nlarew: Container queries should work on all (updated) modern browsers + we will fall back to the two column layout when it's unsupported

https://caniuse.com/css-container-queries

chat/rich-links/src/RichLink/RichLinkBadge/RichLinkBadge.tsx
nlarew: Good catch - leftover from early debugging

chat/rich-links/src/RichLink/RichLink.styles.ts
nlarew: Added an explanation + slightly rewrote the calc for clarity

chat/rich-links/src/RichLink/RichLink.types.ts
nlarew: Good point! I'm not too familiar with as/polymorphism but I can take a look at some other components and let you know if I have any questions on how to implement.

chat/rich-links/src/RichLink/RichLink.styles.ts
nlarew: I think it depends on `baseFontSize` but otherwise we aren't currently wrapping text or expanding the height at all

chat/message/src/MessageLinks/MessageLinks.tsx
nlarew: I wouldn't say it's "intended" but I figured the flexibility would be nice to have should the need ever arise

.changeset/angry-hornets-pump.md
nlarew: Sure added one. I went with "patch" but wasn't sure given this is an initial release.

chat/rich-links/README.md
nlarew: Will add once I've resolved the anchorProps/Card/as TS error

chat/message/README.md
nlarew: Added a new section with them

chat/rich-links/src/RichLink/RichLink.types.ts
nlarew: I think you're right - a relic of working around some TS error I think. Removed this now I'm just picking off these named props and spreading rest on to `...anchorProps`.

```
const {
  url,
  text,
  imageUrl,
  badgeGlyph,
  badgeLabel,
  badgeVariant,
  ...anchorProps
} = props;
```

However that's revealing some weird TS errors. I think the gist of these is that `Card` is not properly typing based on the `as` prop. A bit stumped atm but I'll keep looking at it. Any ideas?

```
Type '{ children: (Element | null)[]; key?: Key | null | undefined; download?: any; href: string; hrefLang?: string | undefined; media?: string | undefined; ping?: string | undefined; ... 270 more ...; as: "a"; }' is not assignable to type 'IntrinsicAttributes & ((Omit<BoxAnchorDefault<CardProps>, "ref"> | Omit<BoxIntrinsic<keyof IntrinsicElements, CardProps>, "ref"> | Omit<...> | Omit<...>) & RefAttributes<...>)'.
  Types of property 'as' are incompatible.
    Type '"a"' is not assignable to type 'undefined'.ts(2322)
```

```
 '{ children: (Element | null)[]; ref: ForwardedRef<HTMLAnchorElement>; key?: Key | null | undefined; download?: any; hrefLang?: string | undefined; ... 272 more ...; href: string; } | { ...; }' is not assignable to type 'IntrinsicAttributes & ((Omit<BoxAnchorDefault<CardProps>, "ref"> | Omit<BoxIntrinsic<keyof IntrinsicElements, CardProps>, "ref"> | Omit<...> | Omit<...>) & RefAttributes<...>)'.
  Type '{ children: (Element | null)[]; ref: ForwardedRef<HTMLAnchorElement>; key?: Key | null | undefined; download?: any; hrefLang?: string | undefined; ... 272 more ...; href: string; }' is not assignable to type 'IntrinsicAttributes & ((Omit<BoxAnchorDefault<CardProps>, "ref"> | Omit<BoxIntrinsic<keyof IntrinsicElements, CardProps>, "ref"> | Omit<...> | Omit<...>) & RefAttributes<...>)'.
    Property 'href' does not exist on type 'IntrinsicAttributes & Omit<BoxIntrinsic<keyof IntrinsicElements, CardProps>, "ref"> & RefAttributes<unknown>'. Did you mean 'ref'?ts(2322)
```

.changeset/wicked-ladybugs-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-chat/rich-links': patch
```
shaneeza, line 2: I think we can make this `major`. This will be the first major release.

chat/rich-links/src/RichLink/RichLink.types.ts
shaneeza: The [Link](https://github.com/mongodb/leafygreen-ui/tree/main/packages/typography/src/Link/Link) component might be helpful

chat/rich-links/src/RichLink/RichLink.types.ts
nlarew: Ended up removing the extends and going with the explicit `anchorProps` prop. Still need to use an `@ts-ignore` (☹️) because `Card` still thinks it's an `HTMLDivElement` instead of `HTMLAnchorElement`.

chat/rich-links/src/RichLink/RichLink.types.ts
nlarew: Took some time to look into this but it's not quite the same use case as Link and TS is misbehaving with the Card `as` prop which complicates things further. Given that we won't be using this on the docs chatbot, can we leave polymorphism as a future feature?

nlarew:
@shaneeza Ready for another look - thanks!

chat/rich-links/src/RichLink/RichLink.types.ts
nlarew: https://jira.mongodb.org/browse/LG-4259

chat/rich-links/src/RichLink/RichLink.types.ts
```diff
@@ -0,0 +1,61 @@
+import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
+
+import { RichLinkBadgeProps } from './RichLinkBadge/RichLinkBadge.types';
+import { RichLinkVariantName } from './RichLinkVariants';
+
+export interface BaseRichLinkProps extends DarkModeProps {
+  /**
+   * The text that shows on the rich link
+   */
+  text: string;
```
shaneeza, line 10: I think this might be better as `children` instead of `text.` That's the pattern we use for the `link` component. [Material UI](https://mui.com/material-ui/react-link/#basic-links) also uses this pattern. So, instead of passing the string to the `text` prop, we pass it as children.

chat/rich-links/src/RichLink/RichLink.types.ts
shaneeza: Thanks for creating the ticket! 

If you don't add the `anchorProps` prop, would you still have to add `@ts-ignore`? Adding a prop just for anchor props isn't a pattern I'm used to seeing, and it sorta feels like an extra step for consumers. @bruugey Do you have any thoughts on this?

chat/message/README.md
```diff
@@ -42,16 +42,35 @@ return (
...
+| `messageBody`        | `string`                                                               | Message body text passed to LGMarkdown                                            |                                                                                                                           |
+| `verified`           | `{ verifier?: string; verifiedAt?: Date; learnMoreUrl?: string; }`     | Sets if an answer is "verified" and controls the content of the message banner.   |                                                                                                                           |
+| `links`              | `{ url: string; text: string; imageUrl?: string; variant: string; }[]` | A list of links to show in a section at the end of the message.                   |                                                                                                                           |
```
shaneeza, line 57: I'm starting to think that maybe these would be better as a slot where you pass the `<RichLink>` component instead of an object. I believe you mentioned on another comment that consumers could potentially use `RichLink`'s independently. If that's the case, then we should keep the usage of `RichLink` consistent. And this way, it would be easier to swap `RichLink`'s between different components. Thoughts?

chat/rich-links/src/RichLink/RichLink.types.ts
shaneeza: I pulled the `rich-links-polymorphic` [branch](https://github.com/mongodb/leafygreen-ui/blob/rich-links-polymorphic/chat/rich-links/src/RichLink/RichLink.tsx), and I'm not running into any build errors. I'm honestly not sure why it's giving you issues. 

Also, `Card` is tricky because it uses `Box`. We haven't gotten around to updating it to use `Polymorphic`, so I think it's okay to ignore the errors.

@bruugey Do you think we should proceed without polymorphic links since this is causing a few issues?

chat/rich-links/src/RichLink/RichLink.types.ts
nlarew: Do you mean the explicit `BaseRichLinkProps.anchorProps` or passing additional anchor props to the link in general?

RE the explicit prop, I also played around with the rich link props extending `HTMLElementProps<'a', never>` but was still hitting TS errors. No matter how I get them from the user, there's still a mismatch between what Card and I expect the props to be.

In general, I think we'd want folks to be able to control link things like `target`, `rel`, etc if they need to

chat/message/README.md
nlarew: I like the idea of component composition in general. Here though I think it's a bit tricky because we'd want to enforce that you're passing `RichLink` components to this slot. Consumers would also need to wrap all of the rich links they pass in a fragment which is a little more friction.

The way I think of it (as set up here) the `links` prop is data that gets passed to the `MessageLinks` component. If you want a custom look for this slot you'd define `componentOverrides.MessageLinks`.

The case for independent rich links is foggy future but would be something like rendering links in the message markdown as rich links, or creating a custom component that also uses rich links.

chat/rich-links/src/RichLink/RichLink.types.ts
shaneeza: Yeah, I was referring to the explicit `anchorProps` prop. Are you're saying that,
```
export interface BaseRichLinkProps extends DarkModeProps, Omit<HTMLElementProps<'a', never>, 'children' | 'href'> {}
```
causes more TS errors?

chat/rich-links/src/RichLink/RichLink.tsx
```diff
@@ -0,0 +1,85 @@
...
+        darkMode={darkMode}
+        ref={ref}
+        className={cx(baseStyles, themeStyles[theme], {
+          [badgeAreaStyles]: showBadge,
+          [imageBackgroundStyles(imageUrl ?? '')]: showImageBackground,
+        })}
+        as="a"
+        href={href}
+        target="_blank"
+        {...anchorProps}
```
shaneeza, line 67: this is not an ideal solution but what if we assert that anchorProps are typed as a div just to get rid of the TS error 
Screenshot 2024-05-09 at 3 38 47 PM

nlarew: Good idea looks like this cast works as expected!

chat/rich-links/src/RichLink/RichLink.tsx
```diff
@@ -62,9 +64,8 @@ export const RichLink = forwardRef<HTMLAnchorElement, RichLinkProps>(
           [imageBackgroundStyles(imageUrl ?? '')]: showImageBackground,
         })}
         as="a"
-        href={href}
-        target="_blank"
-        {...anchorProps}
+        // Cast to div props to get around Card's Box typing
```
shaneeza, line 67: Can you add a link here, in the comment, to the Jira ticket you created about this?

shaneeza:
@nlarew not sure if you saw my comment here https://github.com/mongodb/leafygreen-ui/pull/2328#discussion_r1596860809

chat/rich-links/src/RichLink/RichLink.types.ts
nlarew: @shaneeza Good call - updated!

chat/rich-links/src/RichLink/RichLink.tsx
```diff
@@ -0,0 +1,86 @@
...
+      ...badgeDefaults,
+      ...props,
+    };
+
+    const showBadge = badgeLabel !== undefined;
+
+    const showImageBackground = (imageUrl?.length ?? -1) > 0;
+
+    return (
+      <Card
```
shaneeza, line 59: I just noticed that there are no focus styles, and I think it's because the [`contentStyle`](https://github.com/mongodb/leafygreen-ui/blob/main/packages/card/README.md#properties) prop is missing. Can you add it?

nlarew: Ah that's my bad - in the refactor I forgot to pass `href` to the card so the focus selector wasn't appearing. Fixed the href and now focus states are working properly again.



https://github.com/mongodb/leafygreen-ui/assets/15657698/fbc95ee4-8528-4f6c-aa98-a272b8d88e82

shaneeza (approved):
✨

PR #2329, opened 2024-04-30
Refactor FormField

bruugey:
## ✍️ Proposed changes

### For new components


packages/form-field/src/FormFieldInputContainer/FormFieldInputContainer.styles.ts
```diff
@@ -194,83 +162,56 @@ export const inputWrapperSizeStyles: Record<Size, string> = {
...
 
-export const inputWrapperDisabledStyles: Record<Theme, string> = {
-  [Theme.Light]: css`
+export const getInputWrapperDisabledThemeStyles = (theme: Theme) => {
+  return css`
     cursor: not-allowed;
-    background-color: ${color.light.background.disabled.default};
-    border-color: ${color.light.border.disabled.default};
+    color: ${color[theme].text.disabled
+      .default}; // was not defined in light mode previously
```
stephl3, line 214: do we need this comment?

bruugey: Just a note to myself until Chromatic passes

packages/form-field/src/FormFieldInputContainer/FormFieldInputContainer.styles.ts
```diff
@@ -54,76 +54,46 @@ export const inputWrapperBaseStyles = css`
...
+            ${hoverRing[theme].gray};
         }
       }
 
       &::placeholder {
-        color: ${palette.gray.dark1};
         font-weight: ${fontWeights.regular};
+        color: ${isDarkMode
+          ? color.dark.text.primary.default
+          : palette.gray.base};
```
stephl3, line 93: I think the placeholder tokens should be left as-is. They match the tokens used for disabled but shouldn't use the disabled tokens

I don't see placeholder samples in [the latest figma file](https://www.figma.com/file/4h2VwjCuJUbeZ7hzD2J1rq/LeafyGreen-Design-System?type=design&node-id=30296%3A38718&mode=design&t=I5aKuIOtc2exjJ9N-1). @alvenw can we add variants to the inputs that provide specs for the placeholder colors?

packages/form-field/src/FormFieldInputContainer/FormFieldInputContainer.styles.ts
```diff
@@ -54,76 +54,44 @@ export const inputWrapperBaseStyles = css`
...
-            ${hoverRing.dark.gray};
+          box-shadow: ${autofillShadowOverride(backgroundColor)},
+            ${hoverRing[theme].gray};
         }
       }
 
       &::placeholder {
-        color: ${palette.gray.dark1};
         font-weight: ${fontWeights.regular};
+        color: color[theme].text.disabled.default;
```
stephl3, line 91: hey sorry if my previous comment wasn't clear but can we avoid using the disabled tokens for placeholder? I don't think we should introduce a pattern of doing that since it's a misnomer. can we instead continue using the palette.gray.base and palette.gray.dark1?

packages/typography/src/Error/Error.styles.ts
```diff
@@ -8,33 +8,33 @@ import {
...
 
-  // Unsets browser default margins
-  margin-block-start: 0;
-  margin-block-end: 0;
-`;
+  return css`
+    font-family: ${fontFamilies.default};
+    font-weight: ${fontWeights.regular};
+    font-size: inherit;
+    line-height: inherit;
```
stephl3, line 29: is there still a need to include these `inherit` values if we set these attributes further down?

stories/forms/utils.ts
```diff
@@ -0,0 +1,200 @@
+export const countries = [
```
stephl3, line 1: nit: might name this file `constants.ts` and capitalcase `COUNTRIES`

packages/tokens/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { default as borderRadius } from './breakpoints';
+export { default as borderRadius } from './borderRadius';
```
stephl3, line 1: 😅 wow good catch

packages/tokens/src/color/darkModeColors.ts
```diff
@@ -129,6 +129,11 @@ const darkModeTextColors = {
     [State.Hover]: gray.light2,
     [State.Focus]: blue.light3,
   },
+  [Variant.Placeholder]: {
+    [State.Default]: gray.dark1,
+    [State.Hover]: gray.dark1,
+    [State.Focus]: gray.dark1,
+  },
```
stephl3, line 136: nit: would sort this the same way it is done for `lightModeTextColors`

stephl3 (approved):

stories/forms/utils.ts
bruugey: its just for our shared stories - i can, but we could also easily rm and use another example

PR #2337, opened 2024-05-10
[LG-4208, LG-4232]: button slot pattern

shaneeza:
## ✍️ Proposed changes

### For new components


packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.spec.tsx
```diff
@@ -7,6 +7,7 @@ import {
 } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { axe } from 'jest-axe';
+import { truncate } from 'lodash';
```
shaneeza, line 10: remove

packages/form-footer/src/FormFooter.tsx
```diff
@@ -20,97 +22,92 @@ import PrimaryButton, { PrimaryButtonProps } from './PrimaryButton';
...
-    ),
-  };
-  const _cancelButtonProps = {
-    children: cancelButtonProps?.children || cancelButtonText,
-    onClick: cancelButtonProps?.onClick || onCancel,
-    leftGlyph: cancelButtonProps?.leftGlyph ?? undefined,
-  };
+  const showBackButton = backButtonProps;
+  const showCancelButton = cancelButtonProps;
+  const showDepricatedPrimaryButton = primaryButton;
```
jetpacmonkey, line 37: quick spelling nitpick while I'm here: "deprecated" 😁

packages/form-footer/src/FormFooter.types.ts
```diff
@@ -83,16 +71,4 @@ export interface FormFooterProps extends HTMLElementProps<'footer'> {
    * Useful for setting left and right margins, or max-width
    */
   contentClassName?: string;
-
-  /**
-   * Styling prop
-   */
-  className?: string;
```
stephl3, line 90: do we no longer need this documentation? it seems like it could be helpful to differentiate from `contentClassName`. The current description isn't helpful but maybe could be used to specify which element it applies to?

packages/form-footer/src/FormFooter.tsx
```diff
@@ -20,97 +22,92 @@ import PrimaryButton, { PrimaryButtonProps } from './PrimaryButton';
...
+FormFooter.propTypes = {
+  contentClassName: PropTypes.string,
+  errorMessage: PropTypes.string,
+  darkMode: PropTypes.bool,
+  // eslint-disable-next-line react/forbid-prop-types
+  primaryButtonProps: PropTypes.object,
+  // eslint-disable-next-line react/forbid-prop-types
+  cancelButtonProps: PropTypes.object,
+  // eslint-disable-next-line react/forbid-prop-types
+  backButtonProps: PropTypes.object,
```
stephl3, line 112: do we need these eslint disables? I'm not seeing it get flagged in editor. were they failing in CI?

packages/form-footer/src/FormFooter.tsx
```diff
@@ -20,97 +22,92 @@ import PrimaryButton, { PrimaryButtonProps } from './PrimaryButton';
...
+                <PrimaryButton
+                  darkMode={darkMode}
+                  data-testid="lg-form_footer-primary_button"
+                  {...(primaryButton as PrimaryButtonProps)}
+                />
+              )
+            ) : (
+              <Button
+                variant="primary"
+                data-testid="lg-form_footer-primary_button"
```
stephl3, line 90: nit: even if not adding test harnesses with this change, could be worth making a constants file to store the `data-testid`s

packages/form-footer/src/FormFooter.tsx
```diff
@@ -20,97 +22,92 @@ import PrimaryButton, { PrimaryButtonProps } from './PrimaryButton';
...
+                {...primaryButtonProps}
+              />
```
stephl3, line 92: wonder if this can also use the existing `PrimaryButton` or at least apply the same approach to defining the variant prop. It seems more explicit that `variant` is overwritten whereas this current approach makes it seem like it's hardcoded to `"primary"`

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.types.ts
```diff
@@ -7,6 +8,14 @@ export const Variant = {
 
 export type Variant = (typeof Variant)[keyof typeof Variant];
 
+interface CustomButtonOnClick {
+  onClick?: () => void;
+}
+type CustomConfirmButtonProps = Omit<ButtonProps, 'variant' | 'onClick'> &
+  CustomButtonOnClick;
+type CustomCancelButtonProps = Omit<ButtonProps, 'onClick'> &
+  CustomButtonOnClick;
```
stephl3, line 17: Same question about housing this in the button package. The types are different but maybe it can even be a util that uses `lodash/omit`to narrow the options? It would help with pattern enforcement and make it simpler to identify where custom buttons are being used in the various packages

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -112,17 +122,19 @@ export const ConfirmationModal = React.forwardRef(
...
+            disabled={!confirmEnabled || isConfirmDisabled}
+            className={cx(buttonStyle, confirmButtonProps?.className)}
             variant={variant}
-            disabled={!confirmEnabled || submitDisabled}
             onClick={handleConfirm}
-            className={buttonStyle}
-            data-testid={LGIDS_CONFIRMATION_MODAL.confirm}
           >
-            {buttonText}
+            {buttonText || confirmButtonProps?.children || 'Confirm'}
```
stephl3, line 132: should this follow the same fallback logic order as above?
`confirmButtonProps?.children || buttonText || 'Confirm'`

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -112,17 +122,19 @@ export const ConfirmationModal = React.forwardRef(
...
           </Button>
           <Button
+            {...cancelButtonProps}
             onClick={handleCancel}
-            className={buttonStyle}
+            className={cx(buttonStyle, cancelButtonProps?.className)}
             data-testid={LGIDS_CONFIRMATION_MODAL.cancel}
           >
             Cancel
```
stephl3, line 140: does cancel button text/children get overwritten? if it does, can we make this more explicit like with confirm button?

.changeset/dull-bikes-help.md
```diff
@@ -0,0 +1,30 @@
+---
+'@leafygreen-ui/confirmation-modal': minor
+---
+
+- Adds new prop `confirmButtonProps`. This props is an object that accepts all `Button` props except for the `variant` prop. The variant is controlled by the `variant` prop.
```
stephl3, line 5: small typo: `... This prop* is an object that accepts all...`

packages/form-footer/src/FormFooter.types.ts
shaneeza: I don't think so. I'll update the documentation of `contentClassName` to specify that this is added to the first `div` child of `footer`

packages/form-footer/src/FormFooter.tsx
shaneeza: Screenshot 2024-05-21 at 12 42 25 PM
 I get a warning without them

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
shaneeza: I think that `buttonText` should be first since we don't want to break the current functionality.

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
shaneeza: I don't think this text is customizable, so I'll double-check with Rob.

packages/form-footer/src/FormFooter.types.ts
```diff
@@ -1,14 +1,23 @@
...
+import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
 
 import { PrimaryButtonProps } from './PrimaryButton';
 
-type CustomButtonProps = Pick<
-  ButtonProps,
-  'children' | 'leftGlyph' | 'onClick'
->;
+type ButtonPropsWithOmits = Omit<ButtonProps, 'variant'>;
+type ButtonPropsWithChildren = Required<Pick<ButtonProps, 'children'>>;
```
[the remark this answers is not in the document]
shaneeza: I'm not so sure. Doing so would start to bloat `Button` with types not used in the `Button` component.

packages/form-footer/src/FormFooter.tsx
shaneeza: I like this approach because it avoids destructuring and extra lines of code. I don't think less explicit is a bad thing since it's still going to be overwritten.

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
shaneeza: Confimed with Rob that this should always be `Cancel`

stephl3: It feels possible to me that a product team will want to customize this if they can customize the confirm text but maybe a case of YAGNI 😂

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -146,4 +159,8 @@ ConfirmationModal.propTypes = {
   variant: PropTypes.oneOf(Object.values(Variant)),
   requiredInputText: PropTypes.string,
   darkMode: PropTypes.bool,
+  // eslint-disable-next-line react/forbid-prop-types
+  confirmButtonProps: PropTypes.objectOf(PropTypes.any),
+  // eslint-disable-next-line react/forbid-prop-types
+  cancelButtonProps: PropTypes.objectOf(PropTypes.any),
```
stephl3, line 165: our usage of `PropTypes` feels pretty inconsistent that it makes me wonder if it's actually serving a purpose or if it makes sense to move towards typescript-only support. If we want to type these better, I think it makes sense to export the button proptypes from button similar to how we export the typescript `ButtonProps` interface. can we avoid adding the lint disable if it's just a warning to flag that it's something we should clean up/remove in the future?

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.types.ts
stephl3: I think this should be fine. we can revisit if it comes up again in the future

packages/form-footer/src/FormFooter.tsx
```diff
@@ -20,97 +23,90 @@ import PrimaryButton, { PrimaryButtonProps } from './PrimaryButton';
...
+FormFooter.propTypes = {
+  contentClassName: PropTypes.string,
+  errorMessage: PropTypes.string,
+  darkMode: PropTypes.bool,
+  // eslint-disable-next-line react/forbid-prop-types
+  primaryButtonProps: PropTypes.objectOf(PropTypes.any),
+  // eslint-disable-next-line react/forbid-prop-types
+  cancelButtonProps: PropTypes.objectOf(PropTypes.any),
+  // eslint-disable-next-line react/forbid-prop-types
+  backButtonProps: PropTypes.objectOf(PropTypes.any),
```
stephl3, line 111: same note as above about intentionally leaving these disables out for future cleanup

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
shaneeza: oh good catch, i'll swap those

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
shaneeza: We can't do TS only support because not everyone using our packages use TS. However, i can remove the comments for now.

stephl3 (approved):
LGTM!

packages/form-footer/src/FormFooter.tsx
TheSonOfThomp: PropTypes doesn't play nice with more complex TS types

packages/confirmation-modal/src/ConfirmationModal/ConfirmationModal.tsx
TheSonOfThomp: Jeremy mentioned that he wanted to use this button as a "delete" action. It might be worth opening that up for customization as well

PR #2347, opened 2024-05-22
Extend errorMessage and successMessage types in number input, text area, and text input

stephl3:
## ✍️ Proposed changes

### For new components


packages/number-input/src/NumberInput/NumberInput.types.ts
```diff
@@ -136,12 +136,12 @@ export interface BaseNumberInputProps
   /**
    * The message shown below the input element if the value is invalid.
    */
-  errorMessage?: string;
+  errorMessage?: React.ReactNode;
```
TheSonOfThomp, line 139: Do we not have a shared type for these props?

TheSonOfThomp (approved):

packages/number-input/src/NumberInput/NumberInput.types.ts
stephl3: not currently. do you think it's worth it if it's a simpler type like `ReactNode`?

PR #2348, opened 2024-05-22
LG-4251: remove portal and use descendants pattern in tabs

stephl3:
## ✍️ Proposed changes

- refactors tabs to use `@leafygreen-ui/descendants` infra
- removes `@leafygreen-ui/portal` dependency

🎟 _Jira ticket:_ [LG-4251](https://jira.mongodb.org/browse/LG-4251)

### For new components


packages/tabs/src/TabPanel/TabPanel.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import {
+  useDescendant,
+  useDescendantsContext,
+} from '@leafygreen-ui/descendants';
+
+import {
+  TabPanelsDescendantsContext,
+  TabsDescendantsContext,
+} from '../Tabs/Tabs';
```
TheSonOfThomp, line 11: nit: it might make more sense to either lift these up one directory level, or move them into e.g. `/TabContext`

packages/tabs/src/TabTitle/TabTitle.tsx
```diff
@@ -15,36 +30,60 @@ import {
...
+  const selected = index === selectedIndex;
 
   // Checks to see if the current activeElement is a part of the same tab set
   // as the current TabTitle. If so, and the current TabTitle is not disabled
   // and is selected, we manually move focus to that TabTitle.
   useEffect(() => {
-    const tabsList = Array.from(parentRef?.children ?? []);
+    const tabList = tabDescendants.map(
+      descendant => descendant.element.parentElement,
+    );
```
TheSonOfThomp, line 60: I feel like there's a better way to do this logic now

packages/tabs/src/TabTitle/TabTitle.tsx
```diff
@@ -62,12 +101,15 @@ const TabTitle: ExtendableBox<BaseTabTitleProps, 'button'> = ({
...
     role: 'tab',
     tabIndex: selected ? 0 : -1,
+    ['aria-controls']: relatedTabPanel?.id,
     ['aria-selected']: selected,
-    name: nodeText,
     ['data-text']: nodeText,
-    disabled,
   } as const;
 
   if (typeof rest.href === 'string') {
```
TheSonOfThomp, line 115: If you're feeling up to it, you could take this opportunity to replace `Box` with `Polymorphic` 😊

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -11,19 +17,26 @@ import { BaseFontSize } from '@leafygreen-ui/tokens';
...
 } from './Tabs.styles';
 import { AccessibleTabsProps } from './Tabs.types';
 
+export const TabsDescendantsContext = createDescendantsContext<HTMLDivElement>(
+  'TabsDescendantsContext',
+);
+
+export const TabPanelsDescendantsContext =
+  createDescendantsContext<HTMLDivElement>('TabPanelsDescendantsContext');
+
```
TheSonOfThomp, line 39: Maybe declare these in a separate file?

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -35,54 +48,54 @@ import { AccessibleTabsProps } from './Tabs.types';
...
+  const { descendants: tabDescendants, dispatch: tabDispatch } =
+    useInitDescendants<HTMLDivElement>();
+  const { descendants: tabPanelDescendants, dispatch: tabPanelDispatch } =
+    useInitDescendants<HTMLDivElement>();
+
+  const [uncontrolledSelected, setUncontrolledSelected] = useState(0);
+  const [selected, setSelected] = [
+    controlledSelected ? controlledSelected : uncontrolledSelected,
+    controlledSelected ? setControlledSelected : setUncontrolledSelected,
+  ];
```
TheSonOfThomp, line 92: does `useControlledValue` help here?

packages/tabs/src/TabTitle/TabTitle.tsx
TheSonOfThomp: Or even just consolidating this to 
```jsx
<Box 
  as={typeof rest.href === 'string' ? 'a' : 'button'}
   ...
>
```

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -35,54 +48,54 @@ import { AccessibleTabsProps } from './Tabs.types';
...
-    childrenArray.findIndex(child => child.props.default || 0),
-  );
-  const selected = isControlled ? controlledSelected : uncontrolledSelected;
-  const setSelected = isControlled
-    ? setControlledSelected
-    : setUncontrolledSelected;
-
   const handleChange = useCallback(
     (e: React.SyntheticEvent<Element, MouseEvent>, index: number) => {
       setSelected?.(index);
```
TheSonOfThomp, line 101: nit: I know this isn't your new code, but this handler name is super vague. Handle what change?

packages/tabs/src/TabTitle/TabTitle.tsx
TheSonOfThomp: We're essentially responding to changes to the `selected` prop here, so it would be better to run these side effects when we change that state as opposed to firing a complex effect like this.

One quick solution would be in `Tabs.tsx` to write:
```jsx
  const updateSelection = useCallback(
    (index: number) => {
      setSelected?.(index);
      tabDescendants[index].element.parentElement?.focus();
    },
    [setSelected, tabDescendants],
  );
]
```
and then call `updateSelection` in place of `setSelected`. 

One thing this doesn't account for is when `selected` is controlled, so we may need a custom hook to account for both these possibilities (e.g. a smaller `useEffect` in `Tabs.tsx` that responds to changes in the `controlledSelected` prop state)

packages/tabs/src/TabTitle/TabTitle.tsx
stephl3: I've got a separate ticket to handle this so will knock it out with that change! https://jira.mongodb.org/browse/LG-2728

packages/tabs/src/Tabs/Tabs.tsx
stephl3: I don't think it does for this case

packages/tabs/src/TabTitle/TabTitle.tsx
stephl3: We can't access the `tabDescendants` in `Tabs.tsx` because it's outside of the provider. I think it fits appropriately where it is, but I rewrote it to make it more easily readable

stephl3: ok I think this is looking better now. let me know what you think!

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -74,62 +67,81 @@ const Tabs = (props: AccessibleTabsProps) => {
...
+  } = useCallback(() => {
+    const enabledIndices = tabTitleElements
+      .filter(tabTitleEl => !tabTitleEl.hasAttribute('disabled'))
+      .map(tabTitleEl => tabTitleElements.indexOf(tabTitleEl));
+
+    return {
+      activeIndex: enabledIndices.indexOf(selected),
+      enabledIndices,
+    };
+  }, [selected, tabTitleElements]);
```
TheSonOfThomp, line 111: nit, I think we can make this a bit more "pure-functional". I'd pull out this fn into a separate file, and pass in `tabTitleElements` and `selected` as args

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -74,62 +67,81 @@ const Tabs = (props: AccessibleTabsProps) => {
...
-        } else if (e.key === keyMap.ArrowLeft) {
-          const [enabledIndexes, current] = getEnabledIndexes();
-          setSelected?.(
-            enabledIndexes[
-              (current - 1 + enabledIndexes.length) % enabledIndexes.length
-            ],
-          );
-        }
-      }
+      if (e.metaKey || e.ctrlKey) return;
```
TheSonOfThomp, line 115: should `shift` technically move in the opposite direction?

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -74,62 +67,81 @@ const Tabs = (props: AccessibleTabsProps) => {
...
+    if (
+      !isControlled ||
+      tabTitleElements.length === 0 ||
+      tabTitleElements[controlledSelected] === undefined
+    ) {
+      return;
+    }
+
+    tabTitleElements[controlledSelected].focus();
+  }, [controlledSelected, isControlled, tabTitleElements]);
```
TheSonOfThomp, line 143: This will technically fire if/when `tabTitleElements` or `isControlled` changes. 

I can imagine a case where some external event causes tabs to be added/removed, which would trigger this effect

packages/tabs/src/Tabs/Tabs.tsx
stephl3: moved to a separate file and added tests!

packages/tabs/src/Tabs/Tabs.tsx
TheSonOfThomp: If I comment out this `useEffect` locally, changing `selected` in Storybook does not update the focus (only the green "selected" indicator)

TheSonOfThomp:
Noticed that setting the controlled state to a disabled index still highlights the tab
Screenshot 2024-05-28 at 3 55 21 PM

packages/tabs/src/Tabs/Tabs.tsx
stephl3: hmm good question. I don't see that pattern adopted with tabs implemented by [MUI](https://mui.com/material-ui/react-tabs/) and [react aria](https://react-spectrum.adobe.com/react-aria/examples/category-tabs.html). My guess is that because arrow keys have an implied direction, it's superfluous to create additional `shift` + arrow key combinations

packages/tabs/src/Tabs/Tabs.tsx
stephl3: @TheSonOfThomp to avoid having the effect run based off external events when tabs are added/removed, I opted to leave out the `useEffect`. When testing on the `Controlled` story instance, focus seems to update as intended. Let me know if you think otherwise, and I can add it back in

packages/tabs/src/Tabs.stories.tsx
```diff
@@ -110,6 +110,26 @@ export const LiveExample: StoryFn<TabsProps> = ({
...
+  );
+};
+Controlled.parameters = {
+  chromatic: { disableSnapshot: true },
+};
+Controlled.parameters = {
+  chromatic: {
+    disableSnapshot: true,
+  },
+};
```
TheSonOfThomp, line 131: duplicating the same params here

packages/tabs/src/Tabs.stories.tsx
```diff
@@ -110,6 +110,26 @@ export const LiveExample: StoryFn<TabsProps> = ({
...
 );
 
+export const Controlled: StoryFn<TabsProps> = (args: TabsProps) => {
+  const [selectedTab, setSelectedTab] = useState(0);
+
+  return (
+    <LiveExample
+      {...args}
+      selected={selectedTab}
+      setSelected={setSelectedTab}
```
TheSonOfThomp, line 120: Since we're controlling this value, can we remove it from the SB "controls" 
Screenshot 2024-05-30 at 1 41 00 PM

packages/tabs/src/TabPanel/TabPanel.tsx
```diff
@@ -0,0 +1,33 @@
...
+    return tabDescendants.find(tabDescendant => tabDescendant.index === index);
+  }, [tabDescendants, index]);
+
+  return (
+    <div ref={ref}>
+      {React.cloneElement(child, {
+        id,
+        selected: !child.props.disabled && index === selectedIndex,
+        ['aria-labelledby']: relatedTab?.id,
+      })}
```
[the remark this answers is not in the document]
stephl3: I'll address this in a follow-up

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -35,164 +42,166 @@ import { AccessibleTabsProps } from './Tabs.types';
...
     if (!isComponentType(child, 'Tab')) {
       return child;
     }
 
-    const isTabSelected = index === selected;
-    const { disabled, onClick, onKeyDown, className, ...rest } = child.props;
+    const { disabled, onClick, onKeyDown, name, ...rest } = child.props;
 
     const tabProps = {
       as,
```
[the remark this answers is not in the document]
stephl3: nope, I can move it into a context in this PR https://github.com/mongodb/leafygreen-ui/pull/2357

packages/tabs/src/utils/getActiveAndEnabledIndices/getActiveAndEnabledIndices.ts
```diff
@@ -0,0 +1,12 @@
+export function getActiveAndEnabledIndices(
```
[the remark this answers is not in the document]
stephl3: can address in a follow-up

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -35,164 +42,166 @@ import { AccessibleTabsProps } from './Tabs.types';
...
+      const { activeIndex, enabledIndices } = getActiveAndEnabledIndices(
+        tabTitleElements,
+        selected,
+      );
+      const numberOfEnabledTabs = enabledIndices.length;
+      const indexToUpdateTo =
+        enabledIndices[
+          (e.key === keyMap.ArrowRight
+            ? activeIndex + 1
+            : activeIndex - 1 + numberOfEnabledTabs) % numberOfEnabledTabs
```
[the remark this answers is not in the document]
stephl3: can address and make more explicit in a follow-up

PR #2357, opened 2024-05-28
LG-4147: forceRenderAllTabPanels prop

stephl3:
## ✍️ Proposed changes

`Tabs` changes
- adds `forceRenderAllTabPanels` prop with default value of `false` to `Tabs` component

`Tab` changes
- adds private `_shouldRender` prop to `Tab` component
  - in `Tab`, conditionally renders entire `tabpanel` instead of `children`
  - `_shouldRender` is true if:
    - tab is not disabled, and
    - either `forceRenderAllTabPanels` is propagated to children or tab is selected. if the former, tab is hidden with css

Test utils changes
- adds `getAllTabPanelsInDOM test util
- updates how `getSelectedPanel` queries for selected tab panel

🎟 _Jira ticket:_ [LG-4147](https://jira.mongodb.org/browse/LG-4147)

### For new components

### DIFF
--- a/packages/tabs/src/Tab/Tab.tsx
+++ b/packages/tabs/src/Tab/Tab.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import { useTabsContext } from '../context';
+
 import { TabProps } from './Tab.types';
 
 /**
@@ -20,14 +22,20 @@ import { TabProps } from './Tab.types';
  * @param props.to Destination when name is rendered as `Link` tag.
  *
  */
-function Tab({ selected, children, ...rest }: TabProps) {
+function Tab({ children, disabled, selected, ...rest }: TabProps) {
   // default and name are not an HTML properties
   // onClick applies to TabTitle component, not Tab component
   delete rest.default, delete rest.name, delete rest.onClick, delete rest.href;
 
+  const { forceRenderAllTabPanels } = useTabsContext();
+
+  const shouldRender = !disabled && (forceRenderAllTabPanels || selected);
+
+  if (!shouldRender) return null;
+
   return (
     <div {...rest} role="tabpanel">
-      {selected ? children : null}
+      {children}
     </div>
   );
 }
--- a/packages/tabs/src/Tabs.spec.tsx
+++ b/packages/tabs/src/Tabs.spec.tsx
@@ -74,6 +74,49 @@ describe('packages/tabs', () => {
 
       expect(getByTestId('inline-children')).toBeInTheDocument();
     });
+
+    describe('forceRenderAllTabPanels', () => {
+      test('renders only selected panel in DOM when prop is false', () => {
+        const { getAllTabsInTabList, getAllTabPanelsInDOM, getSelectedPanel } =
+          renderTabs({
+            forceRenderAllTabPanels: false,
+          });
+        const tabs = getAllTabsInTabList();
+        const tabPanels = getAllTabPanelsInDOM();
+        expect(tabs).toHaveLength(3);
+        expect(tabPanels).toHaveLength(1);
+
+        const selectedPanel = getSelectedPanel();
+        const hiddenPanels = tabPanels.filter(
+          panel => panel.id !== selectedPanel?.id,
+        );
+
+        expect(selectedPanel).toBeVisible();
+        expect(hiddenPanels).toHaveLength(0);
+      });
+
+      test('renders all tab panels in DOM but only selected panel is visible when prop is true', () => {
+        const { getAllTabsInTabList, getAllTabPanelsInDOM, getSelectedPanel } =
+          renderTabs({
+            forceRenderAllTabPanels: true,
+          });
+        const tabs = getAllTabsInTabList();
+        const tabPanels = getAllTabPanelsInDOM();
+        expect(tabs).toHaveLength(3);
+        expect(tabPanels).toHaveLength(3);
+
+        const selectedPanel = getSelectedPanel();
+        const hiddenPanels = tabPanels.filter(
+          panel => panel.id !== selectedPanel?.id,
+        );
+
+        expect(selectedPanel).toBeVisible();
+        hiddenPanels.forEach(panel => {
+          expect(panel).not.toBeVisible();
+          expect(panel).toBeInTheDocument();
+        });
+      });
+    });
   });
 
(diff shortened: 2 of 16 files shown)


packages/tabs/src/Tab/Tab.tsx
```diff
@@ -20,14 +20,16 @@ import { TabProps } from './Tab.types';
  * @param props.to Destination when name is rendered as `Link` tag.
  *
  */
-function Tab({ selected, children, ...rest }: TabProps) {
+function Tab({ children, _shouldRender, ...rest }: TabProps) {
```
shaneeza, line 23: Instead of a private prop, it might be better to use context to store the value of `forceRender`. Then, you can read that value inside this component.

stephl3: @shaneeza I prefer not to reach for the context API in this case since it's only 1 nesting level deep. I also like keeping all the rendering/visibility logic in `TabPanel`

shaneeza: I'm not convinced we should expose a prop we don't want consumers to use. What is the con of adding the rendering logic in this component?

packages/tabs/src/Tabs.spec.tsx
```diff
@@ -74,6 +74,50 @@ describe('packages/tabs', () => {
...
     });
+
+    describe('forceRenderAllTabPanels', () => {
+      test('renders only selected panel in DOM when prop is false', () => {
+        const { getAllTabsInTabList, getAllTabPanelsInDOM, getSelectedPanel } =
+          renderTabs({
+            forceRenderAllTabPanels: false,
+          });
+        const tabs = getAllTabsInTabList();
+        const tabPanels = getAllTabPanelsInDOM();
```
shaneeza, line 85: In this test, would this return only 1 panel?

packages/tabs/src/utils/getTestUtils/getTestUtils.ts
```diff
@@ -56,18 +55,50 @@ export const getTestUtils = (
...
+
+    if (!visibleTabPanel) return null;
+
+    return visibleTabPanel;
   };
 
   return {
     getAllTabsInTabList: () => getAllTabsInTabList(),
     getTabUtilsByName: (name: string) => getTabUtilsByName(name),
+    getAllTabPanelsInDOM: () => getAllTabPanelsInDOM(),
```
shaneeza, line 101: Can you add this to the README?

packages/tabs/src/utils/getTestUtils/getTestUtils.ts
```diff
@@ -56,18 +55,50 @@ export const getTestUtils = (
...
+  };
+
+  /**
+   * Gets all tab panels and filters for the displayed tab panel. Returns null if selected panel is not found.
+   */
+  const getSelectedPanel = () => {
+    const allTabPanels = getAllTabPanelsInDOM();
+
+    const visibleTabPanel = allTabPanels.find(
+      tabPanel => getComputedStyle(tabPanel).display !== 'none',
```
shaneeza, line 90: Should a selected tab panel have `aria-selected`?

shaneeza: looks like im wrong. `aria-selected` is for `tab`, not `tabpanel`.

shaneeza: Another alternative could be to use the `hidden` attribute like [mdn docs](https://developer.mozilla.org/en-US/play), and you can search for that attribute?

packages/tabs/src/Tab/Tab.tsx
stephl3: Agree that we probably don't want consumers to use the prop. I created a separate `TabsContext` to store that and the `selectedIndex`

For more context on why I was trying to avoid this initially. We have a weird existing pattern of taking a consumer defined `Tab` and breaking it into a `TabTitle` + `TabPanel`. Ideal future state: we have 2 separate components that the consumer defines separately. We are blocked on completely pursuing that due to the refactor burden it would place on consumers. I think it's better if there is less logic in `Tab` for that future state

packages/tabs/src/Tabs.spec.tsx
stephl3: yes I rewrote the expect lines to make it more explicit

packages/tabs/src/utils/getTestUtils/getTestUtils.ts
stephl3: done!

packages/tabs/src/utils/getTestUtils/getTestUtils.ts
stephl3: from what I could find, this was the recommended approach over `hidden` attribute

source: https://stackoverflow.com/questions/6708247/what-is-the-difference-between-the-hidden-attribute-html5-and-the-displaynone

we also have to use an approach that satisfies search engine crawling. It's still not 100% bulletproof, will depend on how it fares in the docs team implementation

packages/tabs/src/Tabs.spec.tsx
```diff
@@ -74,6 +74,50 @@ describe('packages/tabs', () => {
...
+        expect(tabs.length).not.toEqual(tabPanels.length);
+
+        const selectedPanel = getSelectedPanel();
+        const hiddenPanels = tabPanels.filter(
```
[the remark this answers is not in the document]
stephl3: updated!

packages/tabs/src/Tabs.spec.tsx
```diff
@@ -74,6 +74,49 @@ describe('packages/tabs', () => {
...
+        const tabs = getAllTabsInTabList();
+        const tabPanels = getAllTabPanelsInDOM();
+        expect(tabs).toHaveLength(3);
+        expect(tabPanels).toHaveLength(3);
+
+        const selectedPanel = getSelectedPanel();
+        const hiddenPanels = tabPanels.filter(
+          panel => panel.id !== selectedPanel?.id,
+        );
+
```
shaneeza, line 112: Do you think it would make sense to add a test that the length of `hiddenPanels` should be `2`?

stephl3: sure doesn't hurt! I want to keep building on this so I will add it in a follow-up

PR #2359, opened 2024-05-29
LG-2728: migrate from Box to Polymorphic in tabs package

stephl3:
## ✍️ Proposed changes

- add `@leafygreen-ui/polymorphic` dependency to tabs package
- in `TabTitle` component, migrate from `Box` to `InferredPolymorphic`
- remove `@leafygreen-ui/box` dependency from tabs package

🎟 _Jira ticket:_ [LG-2728](https://jira.mongodb.org/browse/LG-2728)

### For new components

### DIFF
--- a/packages/tabs/src/TabTitle/TabTitle.tsx
+++ b/packages/tabs/src/TabTitle/TabTitle.tsx
@@ -1,9 +1,12 @@
-import React, { RefObject, useCallback, useMemo, useRef } from 'react';
+import React, { useCallback, useMemo } from 'react';
 
-import Box, { ExtendableBox } from '@leafygreen-ui/box';
 import { useDescendant } from '@leafygreen-ui/descendants';
 import { cx } from '@leafygreen-ui/emotion';
 import { getNodeTextContent, Theme } from '@leafygreen-ui/lib';
+import {
+  InferredPolymorphic,
+  useInferredPolymorphic,
+} from '@leafygreen-ui/polymorphic';
 import { BaseFontSize } from '@leafygreen-ui/tokens';
 import { useUpdatedBaseFontSize } from '@leafygreen-ui/typography';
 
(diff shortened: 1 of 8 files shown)


packages/tabs/src/TabTitle/TabTitle.tsx
```diff
@@ -20,89 +23,78 @@ import {
...
+    } as const;
 
-  if (typeof rest.href === 'string') {
     return (
-      <Box
-        as="a"
-        ref={titleRef as RefObject<HTMLAnchorElement>}
-        {...sharedTabProps}
-      >
+      <Component {...componentProps}>
```
TheSonOfThomp, line 90: technically i believe this now allows `Tab` to render "as" any other element, not just `a` or `button`. Should add this to the changeset (minor)

packages/tabs/src/TabTitle/TabTitle.types.ts
```diff
@@ -1,4 +1,7 @@
+import { PolymorphicAs } from '@leafygreen-ui/polymorphic';
+
 export interface BaseTabTitleProps {
+  as?: PolymorphicAs;
```
TheSonOfThomp, line 4: Can we add a control for this prop in Stortybook?

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -48,11 +49,11 @@ import { AccessibleTabsProps } from './Tabs.types';
  * @param props.selected Index of the Tab that should appear active. If value passed, component will be controlled by consumer.
  * @param props.setSelected Callback to be executed when Tab is selected. Receives index of activated Tab as the first argument.
  */
-const Tabs = (props: AccessibleTabsProps) => {
+const Tabs = Polymorphic<AccessibleTabsProps>(props => {
```
TheSonOfThomp, line 52: is `Tabs` itself polymorphic? IIRC the wrapper should always be the same type of element, and only the `TabTitle` can be different

packages/tabs/src/TabTitle/TabTitle.types.ts
TheSonOfThomp, line 4: Do we need to explicitly declare this prop? I believe it gets automatically added by `InferredPolymorphic`

stephl3: oops yeah this is a remnant from earlier

packages/tabs/src/Tabs/Tabs.tsx
stephl3: updated!

TheSonOfThomp (approved):
👍 Just needs a changeset

PR #2362, opened 2024-05-31
Enhance Color types

TheSonOfThomp:
## ✍️ Proposed changes

`lib`: 
 - Adds `RecursiveRecord` type

`tokens`:
 - Improves semantics of color token segment type names (e.g. `Type` -> `Property`, `State` -> `InteractionState`).
 - Adds TSDoc to color types
 - Deprecates `tokens.Mode` in favor of `lib.Theme`


packages/lib/src/types/RecursiveRecord.d.ts
```diff
@@ -0,0 +1,34 @@
...
+ *     >>
+ *   >>
+ * >>
+ * ```
+ * */
+export type RecursiveRecord<
+  Keys extends Array<any>,
+  Strict extends boolean = true,
+> = Keys extends [
+  infer Key, // the current union of keys
```
shaneeza, line 28: Can you ELI5 what's happening on lines 27-30?

TheSonOfThomp: > If the type `Keys` is an array, then de-structure that type. The first index in the array is an inferred `Key` type, and then a spread the remaining indexes into an inferred `Rest` type.

TheSonOfThomp: Technically `Keys` is not an array, but a [Variadic Tuple](https://typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types)

shaneeza: That was very helpful; thanks for the breakdown!

PR #2358, opened 2024-05-29
Menu descendants

TheSonOfThomp:
## ✍️ Proposed changes

Implements `@leafygreen-ui/descendants` in Menu. This is pre-work for several bugs in the linked Epic

🎟 _Jira ticket:_ [Menu Refactor](https://jira.mongodb.org/browse/LG-4120)

### For bug fixes, new features & breaking changes

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] I have run `yarn changeset` and documented my changes

PR #2365, opened 2024-05-31
➡ Updates Input Option

TheSonOfThomp:
- ➡ https://github.com/mongodb/leafygreen-ui/pull/2365

### API changes
- Renames `selected` prop to `checked` (this is done to avoid confusion with the `aria-selected` attribute, witch is conditionally applied via the `highlighted` prop)
 - `checked` applies the `aria-checked` attribute
 - `checked` _does not_ apply any styles. Any "checked" styles must be applied by the consuming component
- Adds `preserveIconSpace` prop to `InputOptionContent` to determine whether menu items should preserve space for a left glyph, or left align all text content. Use this prop in menus where some items may or may not have icons/glyphs, in order to keep text across menu items aligned.
- Extends `AriaLabelPropsWithChildren` in `InputOptionProps`

### Styling changes

- Updates `InputOption` and `InputOptionContent` styles to use updated `color` and `spacing` tokens
- Exports `inputOptionClassName`, and `inputOptionContentClassName`.

#### Spacing overview
 - block padding: 8px
 - inline padding: 12px
 - icon/text/chevron gap: 8px
 - label & description font-size: 13px
 - label & description line-height: 16px

#### Colors overview
 - Left & right icon color: `color.[theme].icon.primary` tokens 
 - Label & Description: use default `Label` & `Description` colors from `typography`
 - Background uses `color[theme].background.primary` tokens (including hover & focus states)
 - Wedge uses `palette.blue.base` for all modes
 - The `highlight` prop uses the `.focus` state color for Icon, Text & Background colors

image

### Internal updates

- Establishes internal `InputOptionContext` to track `disabled`,  `highlighted`, & `checked` attributes.


.changeset/spotty-dogs-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/split-button': patch
+---
+
+Updates Menu. Resolves failing tests due to Menu update
```
bruugey, line 5: What were the failing tests due to menu update - will this cause user tests to fail?

TheSonOfThomp: These were caused by moving to use `onEntered`. 
Overall this will be a major change to `Menu` and part of this epic will be to add test harnesses, so I'm not too concerned about consumer's tests at this point

packages/a11y/src/AriaLabelProps.ts
```diff
@@ -55,3 +55,15 @@ export type AriaLabelPropsWithLabel =
        */
       label: ReactNode;
     } & Partial<AriaLabelProps>);
+
+/**
+ * Conditionally requires {@link AriaLabelProps}
+ * if `children` is not provided
+ */
+export type AriaLabelPropsWithChildren =
```
bruugey, line 63: hm where is this relevant?

packages/input-option/src/InputOptionContent/InputOptionContent.tsx
```diff
@@ -1,54 +1,93 @@
...
+        >
           {leftGlyph}
         </div>
       )}
-      <div className={textWrapper}>
-        <div className={cx(titleClassName, titleBaseStyles)}>{children}</div>
+      <div className={textContainerStyles}>
+        <Label
+          as="div"
+          htmlFor={''}
```
bruugey, line 66: is this correct?

bruugey: maybe we should just not use the label component here?

bruugey: I guess it is a label IF it's an input option, but if we're using it _not_ as one then this doesn't make sense

TheSonOfThomp: Yeah I was using this to ensure we got styles consistent with the `Label` component. Might be more correct to just re-write them here

packages/a11y/src/AriaLabelProps.ts
TheSonOfThomp: I'm using it in InputOption. It can be useful when we want the text read by VO to be whatever the text content (children) is, instead of insisting on a `label` prop. 

(previously `InputOption` without a `label` prop would result in a TS error, which IMO is not correct)

packages/input-option/src/InputOption/InputOption.tsx
```diff
@@ -36,31 +34,46 @@ export const InputOption = Polymorphic<InputOptionProps>(
...
+          darkMode,
+          disabled,
+          highlighted,
+        }}
       >
-        {children}
-      </Component>
+        <Component
+          ref={ref}
+          role="option"
```
bruugey, line 49: is this correct when it's a menuitem? we might need to fork on that and change some things here..

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -19,16 +17,28 @@ export interface BaseInputOptionProps {
...
+   * Defines the currently selected/active element, regardless of interaction state.
+   *
+   * Functionally similar to a checkbox/radio's `checked` attribute,
+   * this identifies an option as currently selected.
+   *
+   * Note: There are no styling changes applied by this prop.
+   * `Checked` styles must be applied by the implementing component
    */
-  selected?: boolean;
+  checked?: boolean;
```
bruugey, line 41: Does this change not require us to update other components (we could test code mods here ! )

TheSonOfThomp: It doesn't seem like these props have been used anywhere yet

.changeset/big-wasps-fix.md
```diff
@@ -0,0 +1,34 @@
+---
+'@leafygreen-ui/input-option': major
+---
+
+### API changes
+- Renames `selected` prop to `checked` (this is done to avoid confusion with the `aria-selected` attribute, witch is conditionally applied via the `highlighted` prop)
```
shaneeza, line 6: A _witch_ you say??

.changeset/big-wasps-fix.md
```diff
@@ -0,0 +1,34 @@
...
+ - `checked` applies the `aria-checked` attribute
+ - `checked` _does not_ apply any styles. Any "checked" styles must be applied by the consuming component
```
shaneeza, line 8: If I understand this correctly, `input-option` used to apply checked/selected styles, but it will no longer do that?

.changeset/big-wasps-fix.md
```diff
@@ -0,0 +1,34 @@
...
+- Adds `preserveIconSpace` prop to `InputOptionContent` to determine whether menu items should preserve space for a left glyph, or left align all text content. Use this prop in menus where some items may or may not have icons/glyphs, in order to keep text across menu items aligned.
+- Extends `AriaLabelPropsWithChildren` in `InputOptionProps`
```
shaneeza, line 10: Can you expand on what this means?

packages/a11y/src/AriaLabelProps.ts
shaneeza: Can you add the bit about VO in the doc above? The context is really helpful

packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -2,77 +2,97 @@ import { css } from '@leafygreen-ui/emotion';
...
+  isInteractive?: boolean;
+}
+
+export const getInputOptionStyles = ({
+  theme,
+  disabled,
+  highlighted,
+  isInteractive,
+}: InputOptionStyleArgs) => {
+  const ixnState = highlighted ? State.Focus : State.Default;
```
shaneeza, line 30: What's `ixn`?

bruugey: interaction lol

packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -2,77 +2,97 @@ import { css } from '@leafygreen-ui/emotion';
...
+        }
+
+        .${leftGlyphClassName} {
+          color: ${color[theme].icon.primary.hover};
+        }
+      }
+
+      /* Focus (styling handled by highlighted prop) */
+      &:focus,
+      &:focus-visible {
```
shaneeza, line 79: why do we need both `focus` and `focus-visible`?

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -19,16 +17,28 @@ export interface BaseInputOptionProps {
   disabled?: boolean;
 
   /**
-   * Defines the currently highlighted option element for keyboard navigation.
-   * Not to be confused with `selected`, which identifies the currently selected option
+   * Defines the currently highlighted option,
+   * and applies the relevant styles and aria attributes.
+   *
+   * Functionally similar to `:focus` state, however highlight not always implemented with true browser focus state
```
shaneeza, line 23: I think there is a word missing after `highlight`?

packages/input-option/src/InputOption/InputOption.types.ts
```diff
@@ -19,16 +17,28 @@ export interface BaseInputOptionProps {
   disabled?: boolean;
 
   /**
-   * Defines the currently highlighted option element for keyboard navigation.
-   * Not to be confused with `selected`, which identifies the currently selected option
+   * Defines the currently highlighted option,
+   * and applies the relevant styles and aria attributes.
```
shaneeza, line 21: Maybe it's worth mentioning that the aria attribute is `aria-selected`?

packages/input-option/src/InputOptionContent/InputOptionContent.styles.ts
```diff
@@ -1,42 +1,112 @@
...
+import {
+  color,
+  spacing,
+  State,
+  transitionDuration,
+  Variant,
+} from '@leafygreen-ui/tokens';
 
+export const inputOptionContentClassName = createUniqueClassName(
+  'input_option-content',
```
shaneeza, line 12: Thoughts on adding `input_option` to a constants file?

packages/input-option/src/InputOptionContent/InputOptionContent.styles.ts
```diff
@@ -1,42 +1,112 @@
...
-export const descriptionBaseStyles = css`
-  max-height: ${spacing[3] * 3}px;
-  overflow: hidden;
-  font-size: inherit;
-  text-overflow: ellipsis;
-`;
+  return css`
+    display: grid;
+    grid-template-columns: ${spacing[400]}px 1fr ${spacing[400]}px;
+    grid-template-areas: '${col1Name} text ${col3Name}';
```
shaneeza, line 35: nice idea

packages/input-option/src/InputOptionContent/InputOptionContent.tsx
```diff
@@ -1,54 +1,90 @@
...
+      <div className={textContainerStyles}>
+        <div
+          className={cx(titleClassName, getTitleStyles({ theme, highlighted }))}
+        >
+          {children}
+        </div>
         {description && (
           <Description
-            className={cx(descriptionClassName, descriptionBaseStyles)}
+            darkMode={darkMode}
```
shaneeza, line 71: nit: Do we need to pass `darkMode`? Can we wrap `InputOption` in a `LeafyGreenProvider`?

packages/input-option/src/InputOption/InputOption.stories.tsx
```diff
@@ -95,4 +92,15 @@ LiveExample.parameters = {
...
 
-export const Generated = () => {};
+export const Generated = {
+  render: () => <></>,
+  parameters: {
+    generate: {
+      combineArgs: {
+        highlighted: [false, true],
+        checked: [false, true],
+        disabled: [false, true],
```
shaneeza, line 102: I may be overthinking this, but can an option be highlighted/checked and disabled at the same time?

shaneeza (commented):
`InputOption` is used in `Chat` but I don't think we have any stories with an opened dropdown; does everything look good there?

packages/input-option/src/InputOption/InputOption.style.ts
TheSonOfThomp: b/c `:focus-visible` styles don't get applied when programmatically setting focus (i.e. with `element.focus()`)

packages/input-option/src/InputOptionContent/InputOptionContent.styles.ts
TheSonOfThomp: Was planning on tackling that with test harnesses 👍

packages/input-option/src/InputOptionContent/InputOptionContent.tsx
TheSonOfThomp: We technically could internally wrap this in the provider, but I'd be a little concerned about the performance of dozens of provider instances in a dropdown for not much gain IMO

packages/input-option/src/InputOption/InputOption.stories.tsx
TheSonOfThomp: `checked`+`disabled`, yes

`highlighted`+`disabled`, not sure, but possibly?

.changeset/big-wasps-fix.md
TheSonOfThomp: 🧙‍♀️

packages/input-option/src/InputOption/InputOption.style.ts
shaneeza: doesn't that mean you can get rid of `focus-visible`? `focus` would override `focus-visible`?

TheSonOfThomp: oh true. yeah, probably

TheSonOfThomp:
> `InputOption` is used in `Chat` but I don't think we have any stories with an opened dropdown; does everything look good there?

Chat `InputBar` uses the `SearchResultsMenu`, so that should be tested here
https://mongodb.chromatic.com/component?appId=642700aa3e49e32bdbf0b0fc&csfId=components-searchinput-searchresultsmenu--generated&buildNumber=2773&k=6662321d911b895a5f4dbc8f-1200px-interactive-true&h=17&b=-4

shaneeza (approved):

PR #2366, opened 2024-05-31
Integration Menu Refactor [LG-4120, LG-4236, LG-4233, LG-3263, LG-3259, LG-3190 LG-3188, LG-3168]

TheSonOfThomp:
## Integration Branch for Menu Refactor

PR Chain:
- ➡️ https://github.com/mongodb/leafygreen-ui/pull/2366
- ~~https://github.com/mongodb/leafygreen-ui/pull/2399~~ Not Doing
- https://github.com/mongodb/leafygreen-ui/pull/2404


packages/menu/src/MenuGroup/MenuGroup.tsx
shaneeza, line 1: it looks like the focus state is cut off a little
Screenshot 2024-06-26 at 1 06 50 PM

packages/menu/src/Menu/Menu.types.ts
```diff
@@ -67,4 +67,11 @@ export interface MenuProps extends Omit<PopoverProps, 'active'> {
    * id passed to the menu dropdown.
    */
   id?: string;
+
+  /**
+   * Whether the menu should always render dark, regardless of the theme context
+   *
+   * @default {true}
+   */
+  renderDarkMenu?: boolean;
```
shaneeza, line 76: When `renderDarkMenu` is true

Screenshot 2024-06-26 at 1 22 35 PM

TheSonOfThomp: Since `MenuGroup` didn't exist before, and we're deprecating the dark menus soon, I didn't spend time getting this prop combo to work

stephl3: also noticed that the collapse/expand icon button coloring for the submenu doesn't have enough contrast
Screenshot 2024-06-26 at 12 01 30 PM

stephl3:
https://github.com/mongodb/leafygreen-ui/assets/52300529/d7a23642-eea4-43a4-a251-9b1745e844f8
when a user clicks the submenu expand/collapse button, it seems like it might be triggering a full re-render and sending the user back to the top of the menu. keyboard navigation seems like it has the correct behavior and does not do this

.changeset/blue-crews-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/menu': minor
```
bruugey, line 2: note: none of these add up to a major for Menu - should we consolidate the changelogs and release this all under one big major for safety's sake?

packages/menu/src/MenuGroup/MenuGroup.tsx
TheSonOfThomp: Technically this state is not possible when this is in a menu, since the focus will remain on the menu item

.changeset/blue-crews-hope.md
TheSonOfThomp: changesets will handle that, and put the change notes in the appropriate section in the changelog

TheSonOfThomp:
> https://github.com/mongodb/leafygreen-ui/assets/52300529/d7a23642-eea4-43a4-a251-9b1745e844f8 when a user clicks the submenu expand/collapse button, it seems like it might be triggering a full re-render and sending the user back to the top of the menu. keyboard navigation seems like it has the correct behavior and does not do this

Hmm this is happening since the first item is the element in focus, and we run a check to make sure the "highlighted" element has focus. 
I'd assume the fix would be to set highlight/focus on the submenu element, but I'm not sure.  @bruugey, what would you say is the correct behavior should be?

.changeset/blue-crews-hope.md
bruugey: none of them are major though, so i'm saying to we batch them all and call it "major" even if we've identitifed these all individually as minor/patch


I do think this is a BIG change

TheSonOfThomp: Ah you're right. This one should be major then

shaneeza (approved):

PR #2372, opened 2024-06-05
Creates `waitForTransition` test utility

TheSonOfThomp:
## ✍️ Proposed changes

Creates a `waitForTransition` test utility. Fires the `transitionend` event on a given element to ensure event handlers are called after CSS transitions complete. 

This is useful for testing `onEntered` or `onExited` callbacks on `Transition` or `Popover` elements.

### DIFF
--- a/packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
+++ b/packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
@@ -1,0 +1,117 @@
+import React, { createRef } from 'react';
+import { Transition } from 'react-transition-group';
+import { render, waitFor } from '@testing-library/react';
+
+import { css } from '@leafygreen-ui/emotion';
+
+import { waitForTransition } from '.';
+
+const ref = createRef<HTMLElement>();
+
+const callbacks = {
+  onEnter: jest.fn(),
+  onEntered: jest.fn(),
+  onEntering: jest.fn(),
+  onExit: jest.fn(),
+  onExited: jest.fn(),
+  onExiting: jest.fn(),
+};
+
+const renderTransitionElement = (in0: boolean) => {
+  const result = render(
+    <Transition
+      nodeRef={ref}
+      in={in0}
+      timeout={0}
+      onEntered={callbacks.onEntered}
+      onExited={callbacks.onExited}
+    >
+      {state => (
+        <div
+          data-testid="test-div"
+          className={css`
+            opacity: ${state === 'entered' ? 1 : 0};
+          `}
+        >
+          {state}
+        </div>
+      )}
+    </Transition>,
+  );
+
+  const rerender = (in1: boolean) =>
+    result.rerender(
+      <Transition
+        nodeRef={ref}
+        in={in1}
+        timeout={0}
+        onEntered={callbacks.onEntered}
+        onExited={callbacks.onExited}
+      >
+        {state => (
+          <div
+            data-testid="test-div"
+            className={css`
+              opacity: ${state === 'entered' ? 1 : 0};
+            `}
+          >
+            {state}
+          </div>
+        )}
+      </Transition>,
+    );
+
+  return {
+    ...result,
+    rerender,
+  };
+};
+
+describe('packages/testing-lib/waitForTransition', () => {
+  afterEach(() => {
+    Object.values(callbacks).forEach(cb => cb.mockReset());
+  });
+
+  describe('triggers react-transition-group handlers', () => {
+    test('onEntered', async () => {
+      const renderResult = renderTransitionElement(false);
+      const testDiv = await renderResult.findByTestId('test-div');
+
+      /** ENTER */
+      renderResult.rerender(true);
+
+      // onEnter will be called immediately
+      await waitFor(() => {
+        // onEntering will be called after a timeout
+        // but onEntered is not called until CSS transitions are finished
+        expect(callbacks.onEntered).not.toHaveBeenCalled();
+      });
+
+      await waitForTransition(testDiv);
+
+      // onEntered is finally called after we wait for the transition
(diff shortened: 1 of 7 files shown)


packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+  describe('triggers react-transition-group handlers', () => {
+    test('onEntered', async () => {
+      const renderResult = renderTransitionElement(false);
+      const testDiv = await renderResult.findByTestId('test-div');
+
+      /** ENTER */
+      renderResult.rerender(true);
+
+      // onEnter will be called immediately
+      await waitFor(() => {
```
bruugey, line 84: why would we have to await this call?

packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+    Object.values(callbacks).forEach(cb => cb.mockReset());
+  });
+
+  describe('triggers react-transition-group handlers', () => {
+    test('onEntered', async () => {
+      const renderResult = renderTransitionElement(false);
+      const testDiv = await renderResult.findByTestId('test-div');
+
+      /** ENTER */
+      renderResult.rerender(true);
```
bruugey, line 81: why do we have to await this in the test below but not here

packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+      await waitFor(() => /** ENTER */ renderResult.rerender(true));
+
+      /** EXIT */
+      renderResult.rerender(false);
+
+      // onExit will be called immediately
+      await waitFor(() => {
+        // onExiting will be called after a timeout
+        // but onExited is not called until CSS transitions are finished
```
bruugey, line 108: should we expect onExiting to have been called instead of leaving the comment? ditto above?

packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
TheSonOfThomp: The `await` in the test below is essentially combining L81-88 here into one test

packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
TheSonOfThomp: The sequence of calls is `onEnter` -> `onEntering` -> `onEntered`. 
`onEntering` doesn't fire until after the `timeout` has elapsed. 
This `await` is essentially showing that "we've waited up to when `onEntering` should have been called"

TheSonOfThomp: it's clearer when I have `expect(onEntering).toHaveBeenCalled()`, but testing that is not really the purpose of this suite, so I left it out.
I can put it back if you think it makes it clearer

bruugey: ah ok im tracking now.

packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
TheSonOfThomp: Ditto as above. This suite isn't testing the implementation of `Transition`, so I didn't think it was necessary to add that `expect`

packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
bruugey: (it is clearer, but if thats not the prupose of the test you can leave it out)

packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
bruugey: i get it now

packages/testing-lib/src/waitForTransition/waitForTransition.ts
```diff
@@ -0,0 +1,21 @@
+import { fireEvent, waitFor } from '@testing-library/react';
+
+import { act } from '../RTLOverrides';
+
+/**
+ * Fires the `transitionEnd` event on the provided element,
+ * triggering any event handlers attached to this CSS transition events.
+ *
+ * Ensures `onEnter`,`onEntering`,`onEntered`, `onExit`, `onExiting` and `onExited` handlers are fired when
```
shaneeza, line 9: From the comments in the test file above, `onEntered` and `onExited` are the only ones not called until `fireEvent.transitionEnd`. How does this ensure that the other handlers like `onEnter`, `onEntering`, etc, are fired?

TheSonOfThomp: Ah, good catch. It only ensures the `*ed` handlers are fired. Updating the comment

packages/testing-lib/src/waitForTransition/waitForTransition.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+};
+
+describe('packages/testing-lib/waitForTransition', () => {
+  afterEach(() => {
+    Object.values(callbacks).forEach(cb => cb.mockReset());
+  });
+
+  describe('triggers react-transition-group handlers', () => {
+    test('onEntered', async () => {
+      const renderResult = renderTransitionElement(false);
```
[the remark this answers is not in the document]
TheSonOfThomp: It means it will trigger a state change from `entered` -> `exiting` -> `exited`

shaneeza (approved):

PR #2375, opened 2024-06-06
Rename "AllIcons" to "LiveExample" for proper rendering on .design

bruugey:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/packages/icon/src/Icon.stories.tsx
+++ b/packages/icon/src/Icon.stories.tsx
@@ -17,7 +17,7 @@ const meta: StoryMetaType<typeof Icon> = {
   title: 'Components/Icon',
   component: Icon,
   parameters: {
-    default: 'AllIcons',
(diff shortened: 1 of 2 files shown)


packages/icon/src/Icon.stories.tsx
```diff
@@ -76,7 +76,7 @@ Single.argTypes = {
 };
 Single.parameters = { chromatic: { disableSnapshot: true } };
 
-export const AllIcons: StoryFn<IconProps> = (
+export const LiveExample: StoryFn<IconProps> = (
```
shaneeza, line 79: I think Ln20 in this file should also be updated

shaneeza (approved):

packages/icon/src/Icon.stories.tsx
bruugey: good catch

PR #2378, opened 2024-06-06
Descendants: adds `getDescendants` 

TheSonOfThomp:
## ✍️ Proposed changes

Adds `getDescendants` accessor function, returned from `useInitDescendants`. Use this method when referencing descendants from withing effects/callbacks where the `descendants` object may have updated, and could be stale.

### Usage

Sometimes you'll need to access descendants from within an event handler or `useEffect` callback, where the descendants may have updated between renders, resulting in a `descendants` object with references to stale/non-existent DOM nodes. In this case, use the `getDescendants()` function to access the most up-to-date descendants object.

```typescript
const { descendants, dispatch, getDescendants } =
  useInitDescendants<HTMLDivElement>();

const handleTransition = () => {
  console.log(descendants); // this list will be empty, or contain references to old DOM nodes
  console.log(getDescendants()); // this call will return a list of the updated descendants list
};

return (
  <DescendantsProvider
    context={MyDescendantsContext}
    descendants={descendants}
    dispatch={dispatch}
  >
    <Transition onEntered={handleTransition}>
      <div {...rest}>{children}</div>
    </Transition>
  </DescendantsProvider>
);
```


packages/descendants/src/useInitDescendants.tsx
```diff
@@ -1,15 +1,22 @@
...
 
 export interface InitDescendantsReturnObject<T extends HTMLElement> {
+  /** The descendants list in the current render */
   descendants: DescendantsList<T>;
+
+  /** A setter function for the descendants state */
   dispatch: Dispatch<DescendantsReducerAction<T>>;
+
+  /** Accessor function for the most up-to-date descendants list */
+  getDescendants: () => DescendantsList<T>;
```
bruugey, line 19: perhaps it makes sense only to return the getter, rather than both?

packages/descendants/src/Descendants.spec.tsx
```diff
@@ -312,5 +285,58 @@ describe('packages/descendants', () => {
...
+      }: ComponentProps<'div'> & { open: boolean }) => {
+        const { descendants, dispatch, getDescendants } =
+          useInitDescendants<HTMLDivElement>();
+
+        const handleTransition = () => {
+          logDescendants(getDescendants());
+        };
+
+        return (
+          <StrictMode>
```
shaneeza, line 305: Why do you need to add `StrictMode`? I thought tests were already using `StrictMode`

packages/descendants/src/useInitDescendants.tsx
TheSonOfThomp: If you want a component to update in response to changing descendants, you'd still need the list object, but we could recommend preferring `getDescendants `.

packages/testing-lib/src/useTraceUpdate.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable no-console */
+import { useEffect, useRef } from 'react';
+
+/**
+ * Used to log & debug what props caused a component to re-render
```
shaneeza, line 5: Can you provide an example of how to use this?

TheSonOfThomp: 👍 (I'll add this in the docs as well)

```jsx
const  myComponent = ({prop1, prop2, prop3, ...rest}: SomeProps) => {

  // When debugging a rendering issue, use `useTraceUpdate` to log which prop caused a component to rerender
  useTraceUpdate({prop1, prop2, prop3, ...rest})

  return <...>
}
```

packages/descendants/README.md
```diff
@@ -100,6 +100,32 @@ export const TestDescendant = ({
 };
 ```
 
+### Handling Sale Descendants
```
bruugey, line 103: nit: Stale

packages/descendants/src/DescendantsReducer.ts
```diff
@@ -38,26 +36,27 @@ export type DescendantsReducerType<T extends HTMLElement> = Reducer<
...
- * Establishes a state with a `descendants` list, and a `dispatch` function to modify the descendants list
- *
- * @param state {@link DescendantsState} The current reducer state
+ * @param currentState {@link DescendantsState} The current reducer state
  * @param action {@link DescendantsReducerAction} The reducer action type
  * @returns A modified {@link DescendantsState} object
  */
 export const descendantsReducer = <T extends HTMLElement>(
-  state: DescendantsState<T>,
+  currentState: DescendantsState<T>,
```
bruugey, line 48: is the fact that we changed the type here going to have any other downstream consequences?

bruugey: (i.e. do we need to update usage in any other component)

packages/descendants/src/useDescendant.tsx
```diff
@@ -44,13 +44,16 @@ export const useDescendant = <T extends HTMLElement>(
   // On render, register the element as a descendant
   useIsomorphicLayoutEffect(() => {
     const _id = id.current;
+    const refExists = document.contains(ref.current);
```
bruugey, line 47: smart

yarn.lock
```diff
@@ -7536,9 +7536,9 @@ camelcase@^7.0.0:
   integrity sha512-xlx1yCK2Oc1APsPXDL2LdlNP6+uu8OCDdhOBSVT279M/S+y75O30C2VuD8T2ogdePBBl7PfPF4504tnLgX3zfw==
 
 caniuse-lite@^1.0.30001406, caniuse-lite@^1.0.30001517, caniuse-lite@^1.0.30001585, caniuse-lite@^1.0.30001587:
-  version "1.0.30001628"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001628.tgz#90b6cd85ddc2e9f831de0225f4ca5a130c8d8222"
-  integrity sha512-S3BnR4Kh26TBxbi5t5kpbcUlLJb9lhtDXISDPwOfI+JoC+ik0QksvkZtUVyikw3hjnkgkMPSJ8oIM9yMm9vflA==
+  version "1.0.30001629"
```
bruugey, line 7539: is this intentional?

packages/descendants/src/DescendantsReducer.ts
TheSonOfThomp: No, this is internal, and shouldn't be used on it's own
The Provider pulls this property off and uses it

yarn.lock
TheSonOfThomp: no, caniuse has been a little fickle recently

bruugey (approved):

PR #2387, opened 2024-06-12
(LG-4339) RichLink variants should handle unknown/undefined gracefully

nlarew:
## ✍️ Proposed changes

Refactors & simplifies how we handle rich link variants inside of the RichLink component. Now we simply look up the props and merge them in, rather than short circuit rendering a separate RichLinkVariant component that had its own constructor/factory.

This also covers two edge cases:

- unknown variant names are ignored
- explicitly undefined variants (i.e. `variant={undefined}`) are ignored

🎟 _Jira ticket:_ [(LG-4339) RichLink variants should handle unknown/undefined gracefully](https://jira.mongodb.org/browse/LG-4339)

### For new components


chat/rich-links/src/RichLink/RichLink.spec.tsx
```diff
@@ -66,5 +68,32 @@ describe('@lg-chat/rich-links', () => {
       expect(screen.getByText('Link to Video')).toBeInTheDocument();
       expect(screen.getByText('Link to Website')).toBeInTheDocument();
     });
+
+    it('ignores unknown variants', () => {
```
shaneeza, line 72: This test name is a little unclear. What does ignoring the unknown variant imply? Maybe you can add that it still renders?

chat/rich-links/src/RichLink/RichLink.tsx
```diff
@@ -13,30 +13,20 @@ import {
...
-      const RichLinkVariant = richLinkVariants[variant];
-      return <RichLinkVariant darkMode={darkMode} {...variantProps} />;
-    }
-
-    const badgeDefaults: Partial<RichLinkBadgeControlProps> = {
-      badgeGlyph: undefined,
-      badgeLabel: undefined,
-      badgeColor: undefined,
-    };
+    const richLinkVariantProps =
```
shaneeza, line 26: I like this solution; is so much cleaner

.changeset/warm-lizards-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-chat/rich-links': minor
```
shaneeza, line 2: It could be a patch change since you're fixing a few edge cases without introducing new functionality.

shaneeza (approved):
Looks good. I left a few minor comments.

PR #2389, opened 2024-06-12
LG-4325: Exposes `onCopy` callback so that consuming applications can respond to clipboard events

bruugey:
## ✍️ Proposed changes

LG-4325: Exposes `onCopy` callback so that consuming applications can respond to successful clipboard events

🎟 _Jira ticket:_ [LG-4325](https://jira.mongodb.org/browse/[LG-4325)

### For new components

### DIFF
--- a/packages/copyable/src/Copyable/Copyable.tsx
+++ b/packages/copyable/src/Copyable/Copyable.tsx
@@ -38,14 +38,15 @@ import {
 import { CopyableProps, Size } from './Copyable.types';
 
 export default function Copyable({
-  darkMode: darkModeProp,
   children,
-  label,
-  description,
   className,
   copyable = true,
-  size: SizeProp,
+  darkMode: darkModeProp,
+  description,
+  label,
+  onCopy,
   shouldTooltipUsePortal = true,
+  size: SizeProp,
 }: CopyableProps) {
   const { theme, darkMode } = useDarkMode(darkModeProp);
   const [copied, setCopied] = useState(false);
@@ -83,6 +84,10 @@ export default function Copyable({
     const clipboard = new ClipboardJS(buttonRef, {
       text: () => children,
       container: portalContainer,
+    });
+
+    clipboard.on('success', (event: React.ClipboardEvent<HTMLDivElement>) => {
+      onCopy?.(event);
     });
 
     if (copied) {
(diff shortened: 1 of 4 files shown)


packages/copyable/src/Copyable/Copyable.tsx
```diff
@@ -85,6 +86,10 @@ export default function Copyable({
       container: portalContainer,
     });
 
+    clipboard.on('success', (event: React.ClipboardEvent<HTMLDivElement>) => {
+      onCopy?.(event);
+    });
```
stephl3, line 91: Screenshot 2024-06-12 at 1 50 25 PM
The ticket says I can't view but was looking to verify if we only want to fire this on `'success'`. `onCopy` suggests this callback should run any time a user attempts to copy (e.g. if a product team added analytics to capture button clicks)

If the ticket specifies we only want to fire this on `'success'`, can we call the prop `onCopySuccess`? That said, I think an `onCopy` prop makes sense to introduce as well. It would be good to retain parity with the implementation of the `CopyButton` component in the `Code` package
https://github.com/mongodb/leafygreen-ui/blob/9afcc3bdbfe966ae894903fdeb990b3a3c799ac9/packages/code/src/CopyButton/CopyButton.tsx#L65-L69

stephl3 (commented):
can we add to the types docs and README?

bruugey:
> can we add to the types docs and README?

It's actually part of the types by default bc we're extendign HTML Div elements and they have a native onCopy handler.

packages/copyable/src/Copyable/Copyable.tsx
bruugey: Added on fail too - good call

packages/copyable/src/Copyable/Copyable.tsx
```diff
@@ -85,6 +86,14 @@ export default function Copyable({
...
+
+    clipboard.on('error', (event: React.ClipboardEvent<HTMLDivElement>) => {
+      onCopy?.(event);
+    });
```
stephl3, line 95: at this point, is there a reason for doing it this way vs calling `onCopy` in the `onClick` handler?

stephl3 (approved):
1 implementation question but this looks fine too

packages/copyable/src/Copyable/Copyable.tsx
bruugey: so we can pass the consumer the same event information that we get ... the onClick won't have the text/trigger args

bruugey: now we're passing the clipboard event not the onClick event

PR #2394, opened 2024-06-13
LG-4341: Add 72px spacing token

stephl3:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4341](https://jira.mongodb.org/browse/LG-4341)

### For new components

### DIFF
--- a/.changeset/sweet-rice-unite.md
+++ b/.changeset/sweet-rice-unite.md
@@ -1,0 +1,7 @@
+---
+'@leafygreen-ui/tokens': minor
+---
+
+Adds 72px spacing token
+
+[LG-4341](https://jira.mongodb.org/browse/LG-4341)
--- a/packages/tokens/src/spacing.ts
+++ b/packages/tokens/src/spacing.ts
@@ -15,6 +15,7 @@ const spacing = {
   1200: 48,
   1400: 56,
   1600: 64,
+  1800: 72,
 
   /**  @deprecated use 100 instead */
   1: 4,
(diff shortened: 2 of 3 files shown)


bruugey (approved):

packages/tokens/src/spacing.ts
```diff
@@ -15,6 +15,7 @@ const spacing = {
   1200: 48,
   1400: 56,
   1600: 64,
+  1800: 72,
```
bruugey, line 18: is this the right number... it's only 6 away from 1600 so i think it shoudl be 1750 not 1800

bruugey (changes requested):
i don't think the number is correct

packages/tokens/src/spacing.ts
TheSonOfThomp: This should be correct: 18 x 4 = 72

TheSonOfThomp (approved):

.changeset/sweet-rice-unite.md
```diff
@@ -0,0 +1,7 @@
+---
+'@leafygreen-ui/tokens': minor
+---
+
+Adds 72px spacing token
```
TheSonOfThomp, line 5: nit: maybe add what the token key is here

packages/tokens/src/spacing.ts
bruugey: ugh you're right

stephl3: @bruugey it's 8 away

1800 / 100 = 72 / 4 = 18

bruugey (approved):

PR #2380, opened 2024-06-07
Restructure SubMenu & Implement new Light-in-Light mode [LG-4236, LG-4060, LG-3263, LG-3168, LG-3190]

TheSonOfThomp:
## ✍️ Proposed changes

Part of the Menu Refactor PR chain:
- https://github.com/mongodb/leafygreen-ui/pull/2366
- https://github.com/mongodb/leafygreen-ui/pull/2358
- https://github.com/mongodb/leafygreen-ui/pull/2365
- https://github.com/mongodb/leafygreen-ui/pull/2369
- ➡ https://github.com/mongodb/leafygreen-ui/pull/2380
- https://github.com/mongodb/leafygreen-ui/pull/2399

PR #2369, opened 2024-06-04
Update Menu item styling [LG-3188]

TheSonOfThomp:
Part of the Menu Refactor PR chain:
- https://github.com/mongodb/leafygreen-ui/pull/2366
- https://github.com/mongodb/leafygreen-ui/pull/2358
- https://github.com/mongodb/leafygreen-ui/pull/2365
- ➡ https://github.com/mongodb/leafygreen-ui/pull/2369
- https://github.com/mongodb/leafygreen-ui/pull/2380

## ✍️ Proposed changes

- Updates `MenuItem` to use `InputOption`
- Aligns spacing & colors with other dropdown menus
- Creates `MenuItem` stories

PR #2399, opened 2024-06-14
Menu Props [LG-3259]

TheSonOfThomp:
- ➡  https://github.com/mongodb/leafygreen-ui/pull/2399
- https://github.com/mongodb/leafygreen-ui/pull/2404

## ✍️ Proposed changes
- Updates Menu props to extend `div` props
- Moves `...rest` prop spread onto the `Popover` div (previously props were spread on the inner `ul` element)


packages/menu/src/Menu.spec.tsx
```diff
@@ -348,7 +364,7 @@ describe('packages/menu', () => {
 
       expect(firstItem).toHaveFocus();
 
-      userEvent.keyboard('[Enter]');
+      userEvent.type(menuEl!, '[Enter]');
```
shaneeza, line 367: I'm just curious: why are we switching this to `.type`, but on L234, we're switching to `.keyboard`?

packages/menu/src/Menu/Menu.types.ts
```diff
@@ -12,7 +12,9 @@ export type SubMenuType = ReactElement<
   InferredPolymorphicPropsWithRef<PolymorphicAs, SubMenuProps>
 >;
 
-export interface MenuProps extends Omit<PopoverProps, 'active'> {
+export interface MenuProps
+  extends Omit<PopoverProps, 'active'>,
+    Omit<ComponentPropsWithoutRef<'div'>, 'onClick'> {
```
shaneeza, line 17: If I'm reading this correctly, this omits the `onClick` from `div`, but `PopoverProps` will still include `onClick`?

TheSonOfThomp: Yeah we omit the `onClick` from div props since it collides with the one defined in `PopoerProps`

packages/menu/src/Menu/Menu.tsx
```diff
@@ -185,14 +188,19 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
...
           justify={justify}
           refEl={refEl}
           adjustOnMutation={adjustOnMutation}
+          // Need to stop propagation, otherwise Menu will closed automatically when clicked
+          onClick={e => e.stopPropagation()}
+          onKeyDown={handleKeyDown}
           onEntered={handlePopoverOpen}
           data-testid={LGIDs.root}
           data-lgid={LGIDs.root}
+          {...rest}
```
shaneeza, line 203: Could this rest override the `onClick` above?

TheSonOfThomp: Oo good catch!

.changeset/two-deers-matter.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/menu': major
+---
+
+- Updates Menu props to extend `div` props
```
bruugey, line 5: Wondering if we leave this as-is for now, and revisit when we revisit how we handle Popover props holistically throughout the system. We can discuss at dev sync today if preferred

bruugey: If we omit this from the queue does it break everything that comes after?

PR #2402, opened 2024-06-18
LG-4338: Normalizes Modal size property

bruugey:
## ✍️ Proposed changes

[LG-4338](https://jira.mongodb.org/browse/LG-4338)

### For new components

### DIFF
--- a/packages/modal/src/Modal/ModalView.tsx
+++ b/packages/modal/src/Modal/ModalView.tsx
@@ -43,7 +43,7 @@ const ModalView = React.forwardRef(
   (
     {
       open = false,
-      size = ModalSize.Default,
+      size: sizeProp = ModalSize.Default,
       setOpen = () => {},
       shouldClose = () => true,
       closeIconColor = CloseIconColor.Default,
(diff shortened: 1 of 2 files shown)


packages/modal/src/Modal/ModalView.tsx
```diff
@@ -87,6 +87,20 @@ const ModalView = React.forwardRef(
           fallbackFocus: `#${closeId}`, // tests fail without a fallback. (https://github.com/focus-trap/focus-trap-react/issues/91)
         };
 
+    let size: ModalSize = ModalSize.Default;
```
shaneeza, line 90: Could you do something like, 
```
size = ModalSize[size] ?? ModalSize.Default
```

shaneeza: nvm, i don't think thats correct

shaneeza: or maybe something like this so you don't have to list every size

```
const exists = Object.values(ModalSize).includes(size);
const size = exists ? size : ModalSize.Default
```

bruugey: good call - i can do this

stephl3 (approved):
LGTM

PR #2403, opened 2024-06-20
Enhances MenuGroup [LG-4346]

TheSonOfThomp:
Part of the Menu Refactor PR chain:
- https://github.com/mongodb/leafygreen-ui/pull/2366
- https://github.com/mongodb/leafygreen-ui/pull/2358
- https://github.com/mongodb/leafygreen-ui/pull/2365
- https://github.com/mongodb/leafygreen-ui/pull/2369
- https://github.com/mongodb/leafygreen-ui/pull/2380
- https://github.com/mongodb/leafygreen-ui/pull/2399
- ➡️ https://github.com/mongodb/leafygreen-ui/pull/2403
- https://github.com/mongodb/leafygreen-ui/pull/2404

## ✍️ Proposed changes

https://jira.mongodb.org/browse/LG-4346

PR #2404, opened 2024-06-20
Updates `FocusableMenuItem` component for descendants

TheSonOfThomp:
- ~~https://github.com/mongodb/leafygreen-ui/pull/2399~~ Not Doing
- ➡️ https://github.com/mongodb/leafygreen-ui/pull/2404

## ✍️ Proposed changes

A replacement for `FocusableMenuItem`, that allows adding misc. elements (like TextInput, Button) as menu children

Screenshot 2024-06-20 at 5 28 16 PM

Resolves: [LG-4202](https://jira.mongodb.org/browse/LG-4202)


packages/menu/src/MenuDescendant/MenuDescendant.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const MenuDescendant = React.forwardRef<
+  HTMLElement,
+  PropsWithChildren<{}>
+>(({ children }, fwdRef) => {
+  const { theme } = useMenuContext();
+  const { ref, id } = useDescendant(MenuDescendantsContext, fwdRef, {});
+  const child = React.Children.only(children) as ReactElement;
+
+  return child ? (
```
bruugey, line 19: do we want to force this coercion to an InputOption component? i don't think so - i think instead we expose the hook that registers an element to the Descendant Context, or just clone the element and pass it the ref/id

bruugey: we can probably do this in focusable menuitem directly

TheSonOfThomp: You're right - we shouldn't use InputOption here, just a div with padding (to solve the focus issues)

TheSonOfThomp: Also, (not to self) I need to document that the single child of MenuDescendant needs to be focusable

packages/menu/src/MenuDescendant/MenuDescendant.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <div
+      id={id}
+      className={cx(getMenuDescendantWrapperStyles(theme))}
+      {...rest}
+    >
+      {React.cloneElement(child, {
+        ref,
```
bruugey, line 37: does the id and ref need to be on the same element for the descendant pattern to work correctly or no? also do we want to pass an aria-role while we're at it?

bruugey (commented):
Small question in the PR - last thing I'll say:

Should we just update FocusableMenuItem to do this? Why worry about deprecating a component when they're essentially doing the same thing just updated for a new pattern in the system. I know the name isn't the best, but I think it might make more sense that way. Open to discussing.

packages/menu/src/MenuDescendant/MenuDescendant.tsx
TheSonOfThomp: - `id` doesn't matter for tracking descendants, only the ref. The `id` is just used internally as an identifier. 
- The role would change depending on the element, no? If we're baking a lot of things in, why not keep it as an `InputOption`?

bruugey: InputOption enforces way too much style. This is just about making something usable/accessible by Menu specifically without enforcing any stylistic rules.

TheSonOfThomp:
> Should we just update FocusableMenuItem to do this?

Potentially. I had created the new component when I was baking in the `InputOption`. Doesn't need to be a new component now

bruugey (approved):

packages/menu/src/FocusableMenuItem/FocusableMenuItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const FocusableMenuItem = React.forwardRef<
+  HTMLElement,
+  ComponentPropsWithRef<'div'>
+>(({ children, className, ...rest }, fwdRef) => {
+  const { theme } = useMenuContext();
+  const { ref } = useDescendant(MenuDescendantsContext, fwdRef, {});
+  const child = React.Children.only(children) as ReactElement;
+
+  return child ? (
+    <div className={cx(getFocusableMenuItemWrapperStyles(theme))} {...rest}>
```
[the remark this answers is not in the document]
TheSonOfThomp: yes... but I forgot to pass `className` in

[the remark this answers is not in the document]
bruugey: lol good catch

PR #2412, opened 2024-06-27
LG-4361 : Avoid most infinite render loops in ControlledToast

jetpacmonkey:
## ✍️ Proposed changes

The `ControlledToast` component was using a `props` object that was newly-created on every render, and then including it in a dependencies array for a `useEffect`. The `useEffect`'s cleanup callback was calling `popToast`, which triggers a re-render, so any time a component with an open controlled toast had a second render it would trigger an infinite render loop. This change is one possible way to avoid this, using `updateToast` instead of pop and push, storing some objects in refs (with deep `isEqual` comparisons to detect updates) to avoid triggering unnecessary re-renders, and separating the "unmount" effect from the "update" one.

🎟 _Jira ticket:_ [LG-4361](https://jira.mongodb.org/browse/LG-4361)

## 🧪 How to test changes

I added a simple unit test that would hang forever with the existing code, that should be enough to see the difference.

### DIFF
--- a/packages/toast/src/ControlledToast/ControlledToast.spec.tsx
+++ b/packages/toast/src/ControlledToast/ControlledToast.spec.tsx
@@ -69,6 +69,22 @@ describe('packages/toast/controlled', () => {
       expect(toast).toBeInTheDocument();
     });
 
+    test('handles rerender when `open` is true without hanging', async () => {
+      const { findByTestId, rerender } = render(
+        <Toast open title="Test Rerender" data-testid="test-toast-rerender" />,
+        {
+          wrapper: ({ children }) => <ToastProvider>{children}</ToastProvider>,
+        },
+      );
+
+      rerender(
+        <Toast open title="Test Rerender" data-testid="test-toast-rerender" />,
+      );
+
+      const toast = await findByTestId('test-toast-rerender');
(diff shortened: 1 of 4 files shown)


packages/toast/src/ControlledToast/ControlledToast.spec.tsx
```diff
@@ -69,6 +69,22 @@ describe('packages/toast/controlled', () => {
       expect(toast).toBeInTheDocument();
     });
 
+    test('handles rerender when `open` is true without hanging', async () => {
```
TheSonOfThomp, line 72: I assume this test fails when we don't use `useStableControlledToastProps`?

TheSonOfThomp (approved):
I like it 👍 

TBF, the ControlledToast is here as a stop-gap while teams can migrate to the `useToast` hook, (but it shouldn't break 😅)

packages/toast/src/ControlledToast/ControlledToast.spec.tsx
jetpacmonkey: Yup! Or technically it just hangs in an infinite loop, but it would fail in Evergreen from a timeout eventually.

PR #2421, opened 2024-07-03
Menu exports

TheSonOfThomp:
## ✍️ Proposed changes

Exports the following from `Menu`:
- `MenuDescendantsContext`
  - Used to register and consume Menu descendants
  - This context value allows you to register and track custom menu items
- Context hooks `useMenuContext`, `useMenuGroupContext`, & `useSubMenuContext`, along with types types `MenuContextData`, `MenuGroupContextData`, & `SubMenuContextData`,
  - use these to read context data from custom menu item components
- `menuColor`: Custom color tokens used within `Menu` and related components
- `MenuInteractionState`: Enumerates interaction states on a menu item
- `LGIDs`: Unique ids for menu elements
- `menuItemClassName`: The unique class name for menu item components
- `subMenuContainerClassName` & `subMenuToggleClassName`
  - Unique classnames for submenu elements

Also, adds missing `useHighlightContext` export from `Descendants`

### DIFF
--- a/packages/menu/src/MenuItem/MenuItem.types.ts
+++ b/packages/menu/src/MenuItem/MenuItem.types.ts
@@ -1,7 +1,5 @@
 import { ReactElement, ReactNode } from 'react';
 
-import { Size } from '../types';
-
 const Variant = {
   Default: 'default',
   Destructive: 'destructive',
@@ -54,5 +52,5 @@ export interface MenuItemProps {
    * @deprecated - Size no longer has any effect
    */
   // TODO: codemod to remove `size` props from existing implementations
-  size?: Size;
+  size?: 'default' | 'large';
 }
--- a/packages/menu/src/testUtils/withMenuContextDecorator.testutils.tsx
+++ b/packages/menu/src/testUtils/withMenuContextDecorator.testutils.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable react/jsx-key, react/display-name, react-hooks/rules-of-hooks */
 import React from 'react';
 import { InstanceDecorator } from '@lg-tools/storybook-utils';
+import noop from 'lodash/noop';
 
 import { css } from '@leafygreen-ui/emotion';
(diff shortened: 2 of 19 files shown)


packages/menu/src/MenuItem/MenuItem.types.ts
```diff
@@ -54,5 +52,5 @@ export interface MenuItemProps {
    * @deprecated - Size no longer has any effect
    */
   // TODO: codemod to remove `size` props from existing implementations
-  size?: Size;
+  size?: 'default' | 'large';
```
shaneeza, line 55: Why the switch from `Size`?

packages/menu/src/testUtils/withMenuContextDecorator.testutils.tsx
```diff
@@ -15,11 +16,20 @@ import { MenuItem } from '../MenuItem';
...
+        highlight,
+        moveHighlight,
+        setHighlight,
+      },
     } = ctx ?? {
       args: {
         darkMode: false,
         renderDarkMenu: false,
+        highlight: undefined,
+        moveHighlight: noop,
```
shaneeza, line 31: What's the diff between `noop` and `undefined`?

shaneeza (approved):

packages/menu/src/testUtils/withMenuContextDecorator.testutils.tsx
TheSonOfThomp: `moveHighlight` has to be defined, so here we use `noop`, which is a function that returns undefined

packages/menu/src/MenuItem/MenuItem.types.ts
TheSonOfThomp: We've deprecated the `size` prop, and I wanted to remove the type def as well. Having the inline types will avoid breaking TS, but let's us remove old files

PR #2423, opened 2024-07-08
Creates highlight management utilities in `descendants`  [LG-4389]

TheSonOfThomp:
## ✍️ Proposed changes

See [RFC document](https://docs.google.com/document/d/1NLwi1gOdsHQ1_p8iQUCqIKx0QP9Cz9SyS3jIHgXjXfA/edit)

### For new components


packages/descendants/src/Highlight/README.md
```diff
@@ -0,0 +1,115 @@
...
+  );
+};
+```
+
+```tsx
+/** HighlightItem.tsx */
+
+const HighlightItem = ({ children }: PropsWithChildren<{}>) => {
+  const { ref, id } = useDescendant(TestDescendantContext);
+  const { highlight } = useHighlightContext(TestHighlightContext);
```
bruugey, line 44: nit: highlight or isHighlighted?

packages/descendants/src/Highlight/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+## `options.onInit`: Setting the initial highlight
+
+To set the initial highlight (or run side-effects once on init), provide an `onInit` callback to the `options` argument of `useHighlight`.
+
+```ts
+const { getDescendants, dispatch } = useInitDescendants<HTMLDivElement>();
+const { highlight, setHighlight } = useHighlight(getDescendants, {
+  // Fired once when the descendants are initialized
+  onInit: () => {
```
bruugey, line 67: is this too many renders? can we jsut bake this in?

packages/descendants/src/Highlight/README.md
TheSonOfThomp: it's a `Descendant` object, so `highlight` is correct

packages/descendants/src/Highlight/README.md
TheSonOfThomp: We need to wait for the first render for `descendants` to actually refer to a DOM element, only then can we accurately set the initial highlight

packages/descendants/src/Highlight/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+## `options.filter`: Filtering descendants
+
+Occasionally some descendants should not be able to be highlighted. In these cases, provide a `filter` function to the `options` argument of `useHighlight`.
+
+```ts
+const { highlight, setHighlight } = useHighlight(getDescendants, {
+  // Exclude descendants with the `disabled` prop from receiving highlight
+  filter: (descendant: Descendant) => !descendant.props.disabled,
```
bruugey, line 96: out of curiosity... why can't we do 
```
useHighlight(getDescendants.filter(d => !d.disabled)
```

TheSonOfThomp: Because then we don't have the correct total number of descendants to compute when to overflow

packages/descendants/src/Highlight/README.md
bruugey: below we say isHighlighted so just consistent

packages/descendants/src/Highlight/README.md
bruugey: wait why not?

packages/descendants/src/Highlight/README.md
TheSonOfThomp: Ah, I accidentally removed the line where I computed `isHighlighted`...

packages/descendants/src/Highlight/README.md
bruugey: it's relative to the number of enabled indexes not to the number of total elements
you just `% enabledIndexes`

bruugey: the overflow doesn't care about the total it's only concerned with the total number that can receive a highlight

packages/descendants/src/Highlight/reducer/makeHighlightReducerFunction.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Creates a new reducer function for closure for a given `descendants` value
+ */
+export const makeHighlightReducerFunction = <T extends HTMLElement>(
+  getDescendants: () => DescendantsList<T>,
+  options?: ReducerFunctionOptions<T>,
+): HighlightReducerFunction<T> => {
+  return (currentHighlight, action) => {
+    const descendants = getDescendants();
```
bruugey, line 27: because this is relative to what ever you pass in and not the canonical number of descendants, you should be able to filter on the component side which gets passed as the arg (line 23)

packages/descendants/src/Highlight/reducer/makeHighlightReducerFunction.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    return currentHighlight;
+  };
+};
+
+/**
+ * Returns the provided value if it passes the filter function,
+ * otherwise returns the current value
+ */
+function filteredResult<T extends HTMLElement>(
```
bruugey, line 59: i think this is over engineered :O

packages/descendants/src/Highlight/README.md
TheSonOfThomp: Screenshot 2024-07-10 at 6 19 08 PM

Short answer: because the `index` is an attribute on a `Descendant` object, and  if we filter these, the indexes won't be aligned.

If I call `moveHighlight(1)` in the case above, we want it to skip from `Eliane` to `George`. Currently we do this with the following logic:
- The current `highlight` object has `highlight.index == 4` 
- Increment this index by `1` (`i = 5`)
- Check if the descendant with the incremented index is valid (index 5 "Fred" is disabled)
- Keep incrementing by 1 until we find a non-disabled one (`i = 6`)
- Set the highlight to the element with index `6` ("George")

TheSonOfThomp: If we filter the array, we can get into a case where the `index` property on the `highlight` object doesn't match the index in the array. 
Consider the inverse case:
- "George" is selected (`index === 6`)
- Call `moveHighlight(-1)`
- Check if `descendants[5]` is valid
- If not, check `descendants[4]` (etc.)
- Set the highlight to `descendants[4]` (`"Eliane"`)

If we were to pre-filter the descendants list, then we would decrement the index of "George" (`6`) down to `5`, and then check `filteredDescendants[5]`, which would still be "George"

packages/descendants/src/Highlight/reducer/makeHighlightReducerFunction.ts
TheSonOfThomp: See https://github.com/mongodb/leafygreen-ui/pull/2423#discussion_r1673122693

packages/descendants/src/Highlight/reducer/makeHighlightReducerFunction.ts
TheSonOfThomp: I disagree. I think we need this filter for the reasons outlined here https://github.com/mongodb/leafygreen-ui/pull/2423#discussion_r1673122693

packages/descendants/src/Highlight/README.md
TheSonOfThomp: A potential alternative if we really don't want this `options.filter`:

We expect a pre-filtered list of descendants. 
To move the highlight relative to the current one we 
- Loop through the `filteredDescendants` array to find the new index of the current element
- update this index
- return the `filteredDescendant[x]`

Not a lot of effort, but it includes a redundant search through an array, as well as a `filter` loop every render

bruugey: I keep typing and resetting - I think what you have is OK to move forward with, we can get into the nuances offline.

I think there's a lot of assumptions we're baking in by colocating these two contexts, which are discrete and completely independent but become mentally associated by this colocation. 

For example: the moveHighlight function gets complicated if you're being asked to keep track of two different index values in a given component (one from descendants, one from highlihgt) but theoretically if these weren't packaged together you could easily return an index from this context.

All this to say - comfortable going forward with as-is, but an interesting discussion to reason about in my head :)

TheSonOfThomp: Let's chat tomorrow. 

Note, the arg passed into `moveHighlight` is _relative to the current_ highlight, not an absolute index. 

To set an absolute index, you'd use `setHighlight`. 

I decided that integrating the relative updating into the utility was better than having the consumer compute that externally

packages/lib/src/index.ts
```diff
@@ -8,10 +8,13 @@ export * from './helpers';
...
 export { createSyntheticEvent } from './createSyntheticEvent';
 export type {
+  ExclusiveUnion,
   Mutable,
-  RecursiveRecord,
+  Only,
   Optional,
   PartialRequired,
+  RecursiveRecord,
+  ValuesOf,
```
stephl3, line 17: nit: could make this `export * ...`

packages/menu/src/Menu/Menu.tsx
```diff
@@ -169,11 +169,7 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
   };
 
   const popoverContent = (
-    <DescendantsProvider
-      context={MenuDescendantsContext}
-      descendants={descendants}
-      dispatch={dispatch}
-    >
+    <MenuDescendantsProvider>
```
stephl3, line 172: 💥

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -161,16 +159,8 @@ const Tabs = (props: AccessibleTabsProps) => {
...
-        descendants={tabDescendants}
-        dispatch={tabDispatch}
-      >
-        <DescendantsProvider
-          context={TabPanelDescendantsContext}
-          descendants={tabPanelDescendants}
-          dispatch={tabPanelDispatch}
-        >
+      <TabDescendantsProvider>
+        <TabPanelDescendantProvider>
```
stephl3, line 163: 💥

.changeset/swift-points-sip.md
```diff
@@ -0,0 +1,37 @@
...
+'@leafygreen-ui/descendants': major
+---
+
+Updates `useInitDescendants` signature to require a Context value, and return a `Provider` component.
+
+Eliminates the need to destructure `descendants` and `dispatch` from the hook's return value just to pass into the provider. Instead, the hook will construct a pre-populated provider unique to the Context value given.
+
+Note: `descendants`, `dispatch` and `getDescendants` are still returned by the hook for use in the parent component if necessary.
+
+Before:
```
stephl3, line 11: nice! I like this before/after setup

packages/descendants/src/Descendants.stories.tsx
```diff
@@ -77,6 +78,12 @@ export const Basic = () => {
     setItems(shuffle(items));
   };
 
+  // 2. Initialize an empty descendants data structure
```
stephl3, line 81: is there another numbered list item elsewhere in this file?

packages/descendants/src/Descendants.stories.tsx
```diff
@@ -150,6 +159,9 @@ export const Nested = () => {
     },
   ]);
 
+  // 2. Initialize an empty descendants data structure
```
stephl3, line 162: same question on this one

packages/descendants/src/Descendants/useInitDescendants.tsx
```diff
@@ -41,9 +51,26 @@ export const useInitDescendants = <
...
   );
 
+  // Create a new Provider for the given context
+  const Provider = useMemo(() => {
+    // eslint-disable-next-line react/display-name
+    return ({ children }: PropsWithChildren<{}>) => {
+      return (
+        <DescendantsProvider
+          context={context}
+          descendants={getDescendants()}
```
stephl3, line 61: should this be using the `descendants` instead of `getDescendants()`?

packages/descendants/src/Highlight/Highlight.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+      '`onInit` option is called once when descendants are initialized',
+    );
+    test.todo('`onChange` option is called when the highlight object changes');
+    test.todo(
+      '`filter` option prevents relevant descendants from being highlighted',
+    );
+  });
+
+  describe('useHighlightContext', () => {
+    test.todo('returns the current highlight object');
```
stephl3, line 15: is there some reason we're unable to add these at this time?

packages/descendants/src/Highlight/Highlight.stories.tsx
```diff
@@ -0,0 +1,199 @@
...
+  'Adam',
+  'Brooke',
+  'Chris',
+  'Dave',
+  'Eliane',
+  'Fred',
+  'George',
+  'Harry',
+  'Irena',
+  // 'Jeremy',
```
stephl3, line 34: rm or uncomment?

packages/descendants/src/Highlight/Highlight.stories.tsx
```diff
@@ -0,0 +1,199 @@
...
+];
+
+const TestHighlightContext = createHighlightContext('TestHighlight');
+
+const HighlightItem = ({ children }: PropsWithChildren<{}>) => {
+  const isDisabled = children === 'Fred'; // Fred is on sabbatical
```
stephl3, line 40: 😂

packages/descendants/src/Highlight/Highlight.stories.tsx
```diff
@@ -0,0 +1,199 @@
...
+      default:
+        break;
+    }
+  };
+
+  useEffect(() => {
+    document.body.addEventListener('keydown', handleKeyDown);
+    () => document.body.removeEventListener('keydown', handleKeyDown);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
stephl3, line 115: should this be baked into `useHighlight` or is it only for these storybook instances?

packages/descendants/src/Highlight/README.md
```diff
@@ -0,0 +1,117 @@
+# Highlight Management
+
+A suite of utilities for managing the highlight state of Descendants.
```
stephl3, line 3: can you add some detail around how highlight state is similar to or different from active/focus/hover?

packages/descendants/src/Highlight/reducer/makeHighlightReducerFunction.ts
```diff
@@ -0,0 +1,76 @@
...
+        delta,
+        currentHighlight,
+        descendants,
+        options?.filter,
+      );
+      return nextHighlight; // getRelativeDescendant returns a `filter`ed value
+    } else if (isDefined(action.index) || isDefined(action.position)) {
+      // Explicitly set the highlight to the desired position or index
+      const index =
+        action.index ?? getIndexFromPosition(action.position!, descendants);
```
stephl3, line 42: don't think we need to assert `action.position` here

packages/descendants/src/Descendants/useInitDescendants.tsx
TheSonOfThomp: I thought so too, but no. We can't use the descendants object here since it would require us to add `descendants` as a dependency of this `useMemo`, resulting in a _different_ Provider component every time the descendants change. We need this to remain stable across renders. 
(Also, if we just omit `descendants` from the deps array, then it will refer to a stale vlaue)

packages/descendants/src/Highlight/Highlight.spec.ts
TheSonOfThomp: ...I just forgot...

packages/descendants/src/Highlight/Highlight.stories.tsx
TheSonOfThomp: No, we don't want any interaction patterns to be baked into `useHighlight`. And interactions for changing the highlight should be implemented separately

packages/descendants/src/Descendants/useInitDescendants.tsx
TheSonOfThomp: See the [latest changes](https://github.com/mongodb/leafygreen-ui/blob/cd49f1ddb3abf0c818ac53625bb83a9ec4623f54/packages/descendants/src/Highlight/useHighlight.tsx#L104-L113) to `useHighlight` where I have to do something similar to ensure the Provider stays stable across changes to `highlight`

stephl3: separately, I wonder if we could use a different naming between `descendants` and `getDescendants` to make it more explicit when to use which. Maybe something like `getStableDescendants`? Or maybe it even makes sense to only surface the latter; I'm not sure when `descendants` would make more sense over ` getDescendants`

packages/descendants/src/Highlight/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+# Background
+
+Keeping track of which element in a list should be focused/highlighted is a common pattern across the UI library. This package creates a common pattern for implementing this behavior.
```
stephl3, line 7: fantastic!

packages/descendants/src/Highlight/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+# Glossary of terms
+
+## Highlight
+
+An element that currently shows in a _highlighted_ state. This is distinct from the “checked” state. Technically implemented using either DOM activeElement (focus), or the aria-selected attribute. (see Accessibility)
```
stephl3, line 13: did you mean to have the `see Accessibility` link somewhere?

stephl3 (approved):
small question on readme but LGTM!

packages/descendants/src/Descendants/useInitDescendants.tsx
TheSonOfThomp: I'll ad guidance to prefer using `getDescendants`, since in most cases that's what would guarantee you the accurate value. In some cases you can get away with just referencing `descendants`, but I'm not sure if that would actually gain you anything

PR #2432, opened 2024-07-16
LG-4102: `Chip` icon

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4102](https://jira.mongodb.org/browse/LG-4102)

### For new components


packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -47,167 +55,278 @@ export const chipWrapperPadding: Record<
...
+        [InteractionState.Default]: palette.yellow.dark2,
+      },
+    },
+  },
+} as const satisfies RecursiveRecord<
+  [Variant, Theme, Property, InteractionState, string],
+  false
+>;
+
+const chipTextTruncateDisabled = {
```
shaneeza, line 220: Think of another way to organize this

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -47,167 +55,278 @@ export const chipWrapperPadding: Record<
...
+    },
+  },
+  [Theme.Light]: {
+    [VariantToken.Disabled]: {
+      [InteractionState.Focus]: palette.gray.light1,
+    },
+  },
+};
+
+const chipWrapperDisabled = {
```
shaneeza, line 233: Think of another way to organize this

.changeset/khaki-ants-chew.md
```diff
@@ -0,0 +1,12 @@
+---
+'@leafygreen-ui/chip': minor
+---
+
+Adds optional `glyph` prop. The glyph will appear to the right of the chip text.
```
shaneeza, line 5: left

packages/chip/src/Chip.stories.tsx
```diff
@@ -41,7 +45,7 @@ const meta: StoryMetaType<typeof Chip> = {
   args: {
```
stephl3, line 45: can we update the `glyph` control to match the select control we use for `leftGlyph`/`rightGlyph` in the button stories? https://github.com/mongodb/leafygreen-ui/blob/c86227a67d34bae8b5c858b2a72bb348fa1332df/packages/button/src/Button.stories.tsx#L56-L74

packages/chip/src/Chip/Chip.spec.tsx
```diff
@@ -245,6 +247,22 @@ describe('packages/chip', () => {
     });
   });
 
+  describe('glyph', () => {
+    test('renders when glyph is passed', () => {
+      const { queryByTestId } = renderChip({
+        glyph: <Icon glyph="Wizard" data-testid="chip-glyph" />,
+      });
+      const icon = queryByTestId('chip-glyph');
```
stephl3, line 255: nit: `queryByTestId` is useful for asserting that an element is not present, but `getByTestId` makes more sense for this case to get a descriptive error if the expected element is not present

https://testing-library.com/docs/queries/about/#types-of-queries

packages/chip/src/Chip/Chip.tsx
```diff
@@ -70,31 +64,25 @@ export const Chip = React.forwardRef<HTMLSpanElement, ChipProps>(
...
-          chipWrapperThemeStyle(variant, theme),
-          chipWrapperSizeStyle(baseFontSize),
-          {
-            [cx(
-              chipWrapperDisabledStyle[theme],
-              chipWrapperBaseDisabledStyles,
-            )]: disabled,
-          },
+          getChipWrapperStyles(baseFontSize, variant, theme, disabled),
           className,
```
stephl3, line 68: could take this 1 step further and pass `className` as an arg so the `@leafygreen-ui/emotion` dependency is completely separated in `*.styles.ts` files

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -39,175 +46,270 @@ export const chipWrapperPadding: Record<
 > = {
   [BaseFontSize.Body1]: {
     y: 0,
-    x: 6,
+    x: 4,
   },
   [BaseFontSize.Body2]: {
     y: 2,
-    x: 6,
+    x: 4,
```
stephl3, line 53: could update to use spacing tokens

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -39,175 +46,270 @@ export const chipWrapperPadding: Record<
...
-  Record<
-    Theme,
-    {
-      color: string;
-      bgColor: string;
-      focusBgColor: string;
-    }
-  >
-> = {
+export const variantColor = {
```
stephl3, line 57: does it make sense to use color tokens for these values? if we don't know if there are intentional differences, are there gaps we should surface to design?

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -39,175 +46,270 @@ export const chipWrapperPadding: Record<
...
+
+const truncateDisabledColor = {
+  [Theme.Dark]: palette.gray.dark1,
+  [Theme.Light]: palette.gray.light1,
+};
+
+const wrapperDisabledColor = {
+  [Theme.Dark]: palette.gray.dark2,
+  [Theme.Light]: palette.gray.base,
 };
```
stephl3, line 227: these look like they should use tokens

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -39,175 +46,270 @@ export const chipWrapperPadding: Record<
...
+    `,
+    {
+      [css`
+        cursor: not-allowed;
+        background-color: ${color[theme].background.secondary.default};
+        color: ${wrapperDisabledColor[theme]};
+      `]: isDisabled,
+      [css`
+        box-shadow: inset 0 0 1px 1px ${palette.gray.dark2};
+      `]: isDisabled && theme === Theme.Dark,
```
stephl3, line 260: these are easier to parse when the css block is broken out into a human readable variable

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -39,175 +46,270 @@ export const chipWrapperPadding: Record<
...
-      background-color: ${palette.gray.dark1};
-    }
-  `,
-  [Theme.Light]: css`
-    &:focus-within {
-      background-color: ${palette.gray.light1};
-    }
-  `,
-};
+export const getChipTextStyles = (
```
stephl3, line 267: same here, it feels harder to read/maintain a big block of css vs breaking it up into human readable variables

packages/chip/src/Chip/Chip.styles.ts
shaneeza: As in create new tokens in `Tokens` for chip?

packages/chip/src/Chip/Chip.styles.ts
shaneeza: I'll double-check with rob because they currently don't match any of our tokens

packages/chip/src/Chip/Chip.styles.ts
shaneeza: I just spoke with Rob, and the current tokens are almost exclusively for input which chips are not a part of. We don't have component-specific tokens yet.

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -39,156 +47,249 @@ export const chipWrapperPadding: Record<
 > = {
   [BaseFontSize.Body1]: {
     y: 0,
-    x: 6,
+    x: spacing[100],
   },
   [BaseFontSize.Body2]: {
     y: 2,
-    x: 6,
+    x: spacing[100],
```
stephl3, line 54: nit: could also swap the `y` value to a token

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -39,156 +47,249 @@ export const chipWrapperPadding: Record<
...
+) => css`
   padding-inline: ${chipWrapperPadding[baseFontSize].x}px;
   padding-block: ${chipWrapperPadding[baseFontSize].y}px;
-`;
 
-export const chipTextDismissSizeStyle = css`
-  padding-inline-end: 2px;
-`;
+  display: flex;
+  gap: 2px;
```
stephl3, line 284: nit: token

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -39,156 +47,249 @@ export const chipWrapperPadding: Record<
...
+  svg {
+    align-self: center;
+    color: ${variantColor[variant][theme].icon.default};
+  }
 
-export const chipTextStyles = (variant: Variant, theme: Theme) => css`
   &:focus-within {
-    background-color: ${variantColor[variant][theme].focusBgColor};
+    background-color: ${variantColor[variant][theme].background.focus};
   }
```
stephl3, line 293: Does this selector ever get used? The text span wrapping the glyph/text doesn't seem to ever have `:focus-within` unless it's forcibly done through dev tools
Screenshot 2024-07-18 at 2 47 57 PM
I'm seeing that there is a `label` that can get passed in. Is this selector included here if a consumer passes in a focusable label?

packages/chip/src/Chip/Chip.styles.ts
```diff
@@ -199,15 +300,32 @@ export const chipTextStyles = (variant: Variant, theme: Theme) => css`
...
-};
+export const textDisabledStyles = (theme: Theme) => css`
+  svg {
+    color: ${color[theme].icon.disabled.default};
+  }
+
+  // truncated + disabled + focused styles (a truncated disabled chip is still focusable)
+  &:focus-within {
+    background-color: ${truncateDisabledColor[theme]};
+  }
```
stephl3, line 311: similar note as above: does this ever get used?

packages/chip/src/Chip/Chip.styles.ts
shaneeza: This is used when the chip truncates.

packages/chip/src/Chip/Chip.styles.ts
shaneeza: whoops

packages/chip/src/Chip/Chip.styles.ts
stephl3: oh gotcha, I see now

stephl3 (approved):

PR #2434, opened 2024-07-17
LG-4397: `Callout` style update

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4397](https://jira.mongodb.org/browse/LG-4397)

### For new components


packages/callout/src/Callout/Callout.styles.ts
```diff
@@ -0,0 +1,100 @@
...
+interface ColorSet {
+  headerText: string;
+  bar: string;
+}
+
+export const calloutColor: Record<Theme, Record<Variant, ColorSet>> = {
+  [Theme.Dark]: {
+    [Variant.Note]: {
+      headerText: palette.blue.light2,
+      bar: palette.blue.light1,
```
TheSonOfThomp, line 26: should we use `color` tokens here?

TheSonOfThomp: eh, maybe not since there's no "Tip" semantic token, etc.

.changeset/spotty-days-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/typography': patch
```
TheSonOfThomp, line 2: Technically I think new exports is `minor` (non-breaking new features)

TheSonOfThomp (approved):

packages/callout/src/Callout/Callout.styles.ts
```diff
@@ -0,0 +1,100 @@
...
+      left: 0;
+      border-radius: 2px;
+      background-color: ${calloutColor[theme][variant].bar};
+    }
+  `;
+
+export const getHeaderStyles = (theme: Theme, variant: Variant) =>
+  css`
+    width: 100%;
+    margin-block-end: ${spacing[100]}px;
```
stephl3, line 91: what is the reason we would use `margin-block-end` vs `margin-bottom` here?

packages/callout/src/Callout/Callout.styles.ts
```diff
@@ -0,0 +1,100 @@
...
+    color: ${calloutColor[theme][variant].headerText};
+  `;
+
+export const getContentStyles = (theme: Theme) => css`
+  a {
+    ${linkStyles}
+    ${linkModeStyles[theme]}
+  }
```
stephl3, line 99: is this something we need to support vs expecting the consumer to use LG typography components to handle the link styling? if not, is there something we should actually patch in the LG link components?

packages/callout/src/Callout/Callout.spec.tsx
```diff
@@ -27,15 +27,12 @@ describe('packages/callout', () => {
 
   for (const key of Object.keys(Variant)) {
     const variant = Variant[key as keyof typeof Variant];
-    const icon = headerIcons[variant];
     const label = headerLabels[variant];
 
     describe(`for variant ${variant}`, () => {
-      test(`renders icon "${icon.displayName}" and label "${label}" in header"`, () => {
+      test(`renders label "${label}" in header"`, () => {
```
stephl3, line 33: I think it was here before your change but there's an extra `"` at the end of this test statement

packages/callout/src/Callout/Callout.styles.ts
shaneeza: Unlike `margin-bottom`, `margin-block-end` depends on the [element's writing mode, directionality, and text orientation](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-end). Our components don't really support this, but I add it anyway.

packages/callout/src/Callout/Callout.styles.ts
shaneeza: This is only specific to this component. In cases where the consumer won't have control over the content returned from a fetch call, we want to make sure that if an anchor tag is rendered, it will have LG link styles.

PR #2436, opened 2024-07-19
Ensures that event handlers do not fire when input is disabled

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4411](https://jira.mongodb.org/browse/LG-4411)

### For new components


packages/radio-group/src/Radio/Radio.tsx
```diff
@@ -90,7 +96,7 @@ function Radio({
           id={id}
           name={name}
           type="radio"
-          onChange={onChange}
+          onChange={handleChange}
           value={value}
           aria-checked={checked}
           aria-disabled={disabled}
```
TheSonOfThomp, line 102: @bruugey should we just be using the disabled attribute in this case?

bruugey: Trying to understand why we don't do this currently..

bruugey: Maybe bc folks wanted to put a tooltip on these elements?

packages/radio-box-group/src/RadioBoxGroup.spec.tsx
```diff
@@ -66,6 +66,30 @@ describe('packages/RadioBox', () => {
...
+
+  const radioBoxGroupContainer = container.firstChild;
+
+  if (!typeIs.element(radioBoxGroupContainer)) {
+    throw new Error('Could not find radio box group container element');
+  }
+
+  const option2 = radioBoxGroupContainer?.children[1];
+  expect(option2.getAttribute('aria-checked')).not.toBeTruthy();
+  expect(onChange).not.toHaveBeenCalled();
```
TheSonOfThomp, line 89: @bruugey is it just the `onChange` or should we be checking for the `checked` attribute as well?

bruugey: see above - checked both

packages/radio-group/src/Radio/Radio.tsx
TheSonOfThomp: I think it's just an oversight, because a tool tip would show up on the label, not the input element itself

packages/radio-box-group/src/RadioBoxGroup.spec.tsx
TheSonOfThomp: Not just `aria-checked`, but the actual checked attribute itself

packages/radio-group/src/Radio/Radio.tsx
bruugey: Not with a RadioBoxGroup or Button - so perhaps for consistency?

packages/radio-box-group/src/RadioBoxGroup.spec.tsx
TheSonOfThomp: (or whatever attribute we use to define the checked styling)

bruugey: Updated

TheSonOfThomp (commented):
We should not allow focus on disabled inputs, but that should be resolved in this ticket https://jira.mongodb.org/browse/LG-4408

packages/radio-group/src/Radio/Radio.tsx
bruugey: the focus is so that if it's a tooltip trigger it can still be used as such, but this feels/looks bad

TheSonOfThomp (approved):

PR #2446, opened 2024-08-02
LG-4171: `Table` test harnesses

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4171](https://jira.mongodb.org/browse/LG-4171)

### For new components


packages/table/README.md
```diff
@@ -325,3 +325,105 @@ To add expandable content to your Table, ensure the `renderExpandedContent` prop
 ### React 18
 
 For cases where you anticipate having more than 20 rows with nested rows, it is advisable to opt for virtual scrolling to improve performance of row transitions.
+
+# Test Harnesses
+
+## getTestUtils()
```
TheSonOfThomp, line 331: nit: use `\``

packages/table/README.md
```diff
@@ -325,3 +325,105 @@ To add expandable content to your Table, ensure the `renderExpandedContent` prop
...
+### Test Utils
+
+```tsx
+const {
+  getTable,
+  getAllHeaders,
+  getHeaderByIndex: { getElement, getSortIcon },
+  getSelectAllCheckbox,
+  getAllVisibleRows,
+  getRowByIndex: {
```
TheSonOfThomp, line 393: technically `getRowByIndex: () => {...}`, right?

packages/table/src/ExpandedContent/ExpandedContent.spec.tsx
```diff
@@ -63,16 +64,17 @@ const RowWithExpandableContent = args => {
...
 describe('packages/table/Row/ExpandableContent', () => {
   test('renders the correct number of cell children', () => {
-    const { getAllByRole: getAllByRoleLocal } = render(
-      <RowWithExpandableContent />,
-    );
-    const firstRow = getAllByRoleLocal('row')[1];
-    expect(getAllByRole(firstRow, 'cell').length).toBe(6);
+    render(<RowWithExpandableContent />);
+    const { getRowByIndex } = getTestUtils();
+    expect(getRowByIndex(0)?.getAllCells().length).toBe(6);
```
TheSonOfThomp, line 69: nit: not sure if this is better or not:
```
expect(getRowByIndex(0)?.getAllCells()).toHaveLength(6);
```

packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -96,6 +96,7 @@ const InternalRowWithRT = <T extends LGRowData>({
         )}
         data-selected={isSelected}
         aria-hidden={!isRowVisible}
+        data-expanded={isExpanded}
```
TheSonOfThomp, line 99: do we need this?

packages/table/src/Table/Table.spec.tsx
```diff
@@ -87,35 +88,40 @@ describe('packages/table/Table', () => {
...
     test('clicking checkbox adds row index to rowSelection state', async () => {
-      const { getAllByRole, getByTestId } = render(
+      const { getByTestId } = render(
         <TableWithHook hookProps={{ hasSelectableRows: true }} />,
       );
-      const firstCheckbox = getAllByRole('checkbox')[1];
-      fireEvent.click(firstCheckbox);
+      const { getRowByIndex } = getTestUtils();
+      const firstCheckbox = getRowByIndex(0)?.getCheckbox();
+      fireEvent.click(firstCheckbox!);
```
TheSonOfThomp, line 96: `userEvent`?

packages/table/src/Table/Table.tsx
```diff
@@ -26,6 +27,7 @@ const Table = forwardRef<HTMLDivElement, TableProps<any>>(
       darkMode: darkModeProp,
       table,
       disableAnimations = false,
+      'data-lgid': dataLgId = LGIDS.root,
```
TheSonOfThomp, line 30: nit: `lgidProp`?

packages/table/src/constants.ts
```diff
@@ -0,0 +1,12 @@
...
+export const LGIDS = {
+  root: LGID_ROOT,
+  cell: `${LGID_ROOT}-cell`,
+  row: `${LGID_ROOT}-row`,
+  header: `${LGID_ROOT}-header`,
+  selectAllCheckbox: `${LGID_ROOT}-select_all`,
+  checkbox: `${LGID_ROOT}-checkbox`,
+  sortIcon: `${LGID_ROOT}-sort_button`,
+  expandButton: `${LGID_ROOT}-expand_button`,
+} as const;
```
TheSonOfThomp, line 12: `as const` doesn't have any effect here when you construct the values using `LGID_ROOT`.
These are all still just typed as `string`, not the literal value

packages/table/src/utils/getTestUtils/getTestUtils.tsx
```diff
@@ -0,0 +1,112 @@
...
+    if (!allHeaders.length)
+      throw new Error('Unable to find any <th> elements.');
+
+    return Array.from(allHeaders);
+  };
+
+  const getHeaderByIndex = (index: number) => {
+    const allHeader = getAllHeaders();
+    const header = allHeader[index];
+    if (!header) return null;
```
TheSonOfThomp, line 30: should we log something here if no header is found at `index`?

packages/table/src/utils/getTestUtils/getTestUtils.tsx
```diff
@@ -0,0 +1,112 @@
...
+      );
+
+    return {
+      getElement: () => header,
+      getSortIcon: () => sortIcon(),
+    };
+  };
+
+  const getSelectAllCheckbox = () => {
+    const checkbox = queryBySelector<HTMLInputElement>(
```
TheSonOfThomp, line 45: what happens if there is no checkbox?

packages/table/src/utils/getTestUtils/getTestUtils.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const getHeaderByIndex = (index: number) => {
+    const allHeader = getAllHeaders();
+    const header = allHeader[index];
+    if (!header) return null;
+
+    const sortIcon = () =>
+      queryBySelector<HTMLButtonElement>(
+        header,
+        `[data-lgid=${LGIDS.sortIcon}]`,
+      );
```
TheSonOfThomp, line 36: nit: should rename this to `getSortIcon`

packages/table/src/utils/getTestUtils/getTestUtils.tsx
```diff
@@ -0,0 +1,112 @@
...
+      `[data-lgid=${LGIDS.row}][aria-hidden="false"], [data-lgid=${LGIDS.row}]:not([aria-hidden])`,
+    );
+
+    return Array.from(allRows);
+  };
+
+  const getRowByIndex = (index: number) => {
+    const allRows = getAllVisibleRows();
+    const row = allRows[index];
+    if (!row) return null;
```
TheSonOfThomp, line 64: Same here, should we log something if no row is found?

packages/table/src/utils/getTestUtils/getTestUtils.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    const isExpanded = row.matches(`[data-expanded="true"]`);
+    const isSelected = row.matches(`[data-selected="true"]`);
+    const isDisabled = row.matches(`[aria-disabled="true"]`);
+
+    return {
+      getElement: () => row,
+      getAllCells: () => allCells,
+      getCheckbox: () => checkbox,
+      getExpandButton: () => getExpandButton,
```
TheSonOfThomp, line 88: These values are already calculated, so if something changes in the table in between `getRow`, calls these values could be stale.

These functions (`getCheckbox` etc) should run `queryBySelector` when called instead of returning a pre-calculated value

packages/table/src/Row/InternalRowWithRT.tsx
shaneeza: Yes. This data attribute is the only way i can check if a row is expanded in the DOM.

packages/table/src/utils/getTestUtils/getTestUtils.tsx
shaneeza: That makes sense, I'll update.

packages/table/src/constants.ts
shaneeza: Am I misinterpreting this? The literal values show up when using `as const`
Screenshot 2024-08-13 at 12 35 07 PM
Screenshot 2024-08-13 at 12 35 21 PM

packages/table/src/utils/getTestUtils/getTestUtils.tsx
shaneeza: queryBySelector returns null if nothing is found. I have a test for this.

packages/table/src/utils/getTestUtils/getTestUtils.tsx
shaneeza: I believe returning null should be sufficient and this is noted in the README. However, I made a change to all test harnesses so that an error is thrown if `querySelectorAll` returns a length of 0.

packages/table/src/utils/getTestUtils/getTestUtils.tsx
shaneeza: As noted above,  i think that `null` should be enough, and its noted in the README.

TheSonOfThomp (approved):

PR #2449, opened 2024-08-07
LG-3327: Prevent re-render in `useViewportSize`

tsck:
## ✍️ Proposed changes

- Initialize viewport size with value instead of setting it in `useEffect` callback to prevent additional re-render.

🎟 _Jira ticket:_ [LG-3327](https://jira.mongodb.org/browse/LG-3327)

### For new components

## 🧪 How to test changes
- Checked all live examples using the `useViewportSize()` hook
- Checked all live examples using the `useAvailableSpace()` hook, which uses `useViewportSize()`
- All previous tests passing

### DIFF
--- a/.changeset/honest-drinks-change.md
+++ b/.changeset/honest-drinks-change.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/hooks': patch
+---
+
+Fixes re-render after initial render of useViewportSize hook
--- a/.changeset/honest-drinks-change.md
+++ b/.changeset/honest-drinks-change.md
@@ -2,4 +2,4 @@
 '@leafygreen-ui/hooks': patch
 ---
 
-Fixes re-render after initial render of useViewportSize hook
+Fixes re-render after initial render of `useViewportSize` hook
--- a/packages/hooks/src/hooks.spec.tsx
+++ b/packages/hooks/src/hooks.spec.tsx
@@ -95,33 +95,51 @@ describe('packages/hooks', () => {
   // Difficult to test a hook that measures changes to the DOM without having access to the DOM
   describe.skip('useMutationObserver', () => {}); //eslint-disable-line jest/no-disabled-tests
 
-  test('useViewportSize responds to updates in window size', async () => {
-    const { result, rerender } = renderHook(() => useViewportSize());
+  describe('useViewportSize', () => {
+    test('returns correct window dimensions on initial render', () => {
+      const mutableWindow: { -readonly [K in keyof Window]: Window[K] } =
+        window;
+      const initialHeight = 360;
+      const initialWidth = 480;
 
-    const mutableWindow: { -readonly [K in keyof Window]: Window[K] } = window;
-    const initialHeight = 360;
-    const initialWidth = 480;
+      mutableWindow.innerHeight = initialHeight;
+      mutableWindow.innerWidth = initialWidth;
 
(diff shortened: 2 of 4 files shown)


tsck:
Should this be a version bump, since it is a technical improvement? I wasn't sure.

packages/hooks/src/useViewportSize.ts
```diff
@@ -13,14 +13,12 @@ function getViewportSize(): ViewportSize {
   };
 }
 
-export default function useViewportSize(): ViewportSize | null {
-  const [viewportSize, setViewportUpdateVal] = useState<ViewportSize | null>(
-    null,
+export default function useViewportSize(): ViewportSize {
+  const [viewportSize, setViewportUpdateVal] = useState<ViewportSize>(
+    getViewportSize(),
```
TheSonOfThomp, line 18: is `window` always defined on initial render? IIRC we ran into issues with this, but they may not be a concern anymore. 
Could you add tests to verify this?

tsck: `window` was defined on initial render when manually testing. Will add a test to verify it shortly

.changeset/honest-drinks-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/hooks': patch
+---
+
+Fixes re-render after initial render of useViewportSize hook
```
TheSonOfThomp, line 5: nit: use `` for function names/ vars in MD files

packages/hooks/src/hooks.spec.tsx
```diff
@@ -95,33 +95,51 @@ describe('packages/hooks', () => {
...
-      expect(result?.current?.width).toBe(updateWidth);
+      mutableWindow.innerHeight = initialHeight;
+      mutableWindow.innerWidth = initialWidth;
+
+      window.dispatchEvent(new Event('resize'));
+      rerender();
+      await waitFor(() => {
+        expect(result?.current?.height).toBe(initialHeight);
+        expect(result?.current?.width).toBe(initialWidth);
+      });
```
TheSonOfThomp, line 130: probably could replace these lines with just the 2 expect statements from above

TheSonOfThomp (approved):

packages/hooks/src/hooks.spec.tsx
tsck: Wonder if there's even value in asserting the initial response in this test now given that it's tested above. Perhaps it just makes sense to assert the one resize here?
```ts
    test('responds to updates in window size', async () => {
      const { result, rerender } = renderHook(() => useViewportSize());

      const mutableWindow: { -readonly [K in keyof Window]: Window[K] } =
        window;

      const updateHeight = 768;
      const updateWidth = 1024;

      mutableWindow.innerHeight = updateHeight;
      mutableWindow.innerWidth = updateWidth;

      window.dispatchEvent(new Event('resize'));
      await waitFor(() => {
        expect(result?.current?.height).toBe(updateHeight);
        expect(result?.current?.width).toBe(updateWidth);
      });
    });
```

tsck: Pushed the change for this but happy to revert if you feel opposed to it.

tsck: Since this was a nit and already approved, I'm going to merge. If more is needed here, happy to do so in a followup.

PR #2454, opened 2024-08-09
LG-4444: OverlayContext and useOverlay

stephl3:
## ✍️ Proposed changes

- Adds `OverlayContext`, `OverlayProvider`, `useOverlayContext`, and `useOverlay`
- Includes `OverlayProvider` in `LeafyGreenProvider`

🎟 _Jira ticket:_ [LG-4444](https://jira.mongodb.org/browse/LG-4444)

## 🧪 How to test changes

Added unit tests

### DIFF
--- a/packages/leafygreen-provider/src/OverlayContext/OverlayContext.spec.tsx
+++ b/packages/leafygreen-provider/src/OverlayContext/OverlayContext.spec.tsx
@@ -1,5 +1,6 @@
 import React, { createRef, useState } from 'react';
-import { fireEvent, render } from '@testing-library/react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 
 import { renderHook } from '@leafygreen-ui/testing-lib';
 
@@ -50,7 +51,7 @@ describe('OverlayProvider', () => {
 
     expect(getByText('Overlays count: 0')).toBeInTheDocument();
 
-    fireEvent.click(getByText('Register overlay'));
+    userEvent.click(getByText('Register overlay'));
 
     expect(getByText('Overlays count: 1')).toBeInTheDocument();
     expect(getByText('Top most overlay ID: 1')).toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


packages/leafygreen-provider/src/OverlayContext/OverlayContext.tsx
```diff
@@ -0,0 +1,79 @@
...
+  useCallback,
+  useContext,
+  useMemo,
+  useState,
+} from 'react';
+import PropTypes from 'prop-types';
+
+interface OverlayItem<E extends HTMLElement = HTMLElement> {
+  element: E;
+  ref: RefObject<E>;
```
tsck, line 14: Nit: Ditto on single letter generic types

packages/leafygreen-provider/src/OverlayContext/OverlayContext.tsx
```diff
@@ -0,0 +1,79 @@
...
+const OverlayProvider = ({ children }: { children: ReactNode }) => {
+  const [overlays, setOverlays] = useState<Array<OverlayItem>>([]);
+
+  const topMostOverlay = useMemo(
+    () => overlays[overlays.length - 1],
+    [overlays],
+  );
+
+  const registerOverlay = useCallback((overlay: OverlayItem) => {
+    if (overlays.findIndex(item => item.id === overlay.id) > -1) return;
```
tsck, line 51: Since all we're doing is checking if any overlay exists with this ID, it may be more readable to just use `Array.prototype.some()` here.
```suggestion
    if (overlays.some(item => item.id === overlay.id)) return;
```

packages/leafygreen-provider/src/OverlayContext/OverlayContext.spec.tsx
```diff
@@ -0,0 +1,98 @@
+import React, { createRef, useState } from 'react';
+import { fireEvent, render } from '@testing-library/react';
```
tsck, line 2: Since we're only firing click events in our tests, we may want to consider using `userEvent` instead of `fireEvent` to better replicate actual user interaction. A bit more reasoning here: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event

packages/leafygreen-provider/src/OverlayContext/OverlayContext.spec.tsx
```diff
@@ -0,0 +1,98 @@
...
+  );
+}
+
+function renderProviderWithChildren() {
+  return render(
+    <OverlayProvider>
+      <TestComponent />
+    </OverlayProvider>,
+  );
+}
```
tsck, line 45: Love the use of test and render functions for nice and clean test blocks! 😍

tsck (commented):
A few small, mostly nit comments but overall looks great!

tsck (approved):
Thanks for the changes, looks great!

LGTM

PR #2443, opened 2024-07-26
Flip ExpandableCard icon and add animation

stephl3:
## ✍️ Proposed changes

- Updates chevron in `ExpandableCard` to properly indicate if interaction will expand or collapse
- Adds animation

🎟 _Jira ticket:_ [Name of ticket](https://jira.mongodb.org/browse/[name-of-ticket])

### For bug fixes, new features & breaking changes

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `yarn changeset` and documented my changes

### For new components

- [ ] I have added my new package to the global tsconfig
- [ ] I have added my new package to the Table of Contents on the global README
- [ ] I have verified the Live Example looks as intended on the design website.

## 🧪 How to test changes

- [ ] Toggle expand/collapse in chromatic storybook instance

PR #2455, opened 2024-08-13
LG-4319: Add expand/collapse prop to code component in React

tsck:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4319](https://jira.mongodb.org/browse/LG-4319)

### For new components

## 🧪 How to test changes
- Unit tests
- Check `Code` for unexpected regressions


packages/code/src/Code/Code.styles.ts
```diff
@@ -98,6 +124,22 @@ export const singleLineCodeWrapperStyle = css`
   padding-bottom: ${(singleLineComponentHeight - lineHeight) / 2}px;
 `;
 
+export function getExpandableCodeWrapperStyle(
```
shaneeza, line 127: Should there be a transition?

packages/code/src/Code/Code.spec.tsx
```diff
@@ -237,4 +237,75 @@ describe('packages/Code', () => {
...
+    `;
+
+    test('shows no expand button when <= 5 lines of code', () => {
+      render(
+        <Code expandable={true} language="javascript">
+          {codeSnippet}
+        </Code>,
+      );
+
+      expect(screen.queryByRole('button', { name: /expand/i })).toBeNull();
```
shaneeza, line 258: We should use a testid here. This would also make it easier for consumers to target this button.

packages/code/src/Code/Code.styles.ts
```diff
@@ -98,6 +124,22 @@ export const singleLineCodeWrapperStyle = css`
...
+  expanded: boolean,
+  baseFontSize: number,
+) {
+  const codeLineHeight = baseFontSize === BaseFontSize.Body2 ? 24 : 20;
+  const numOfLinesWhenCollapsed = 5;
```
shaneeza, line 132: Will this always be 5?

packages/code/src/Code/Code.styles.ts
```diff
@@ -112,6 +154,47 @@ export function getCodeWrapperVariantStyle(theme: Theme): string {
...
+    Code wrapper's border radius is 12px. Matching that creates a very 
+    slight gap between the button and the code wrapper. Decreasing by
+    1px removes gap without making focused state look off.
+  */
+  border-radius: 0 0 11px 11px;
+  border-width: 1px 0 0 0;
+  border-style: solid;
+  display: flex;
+  font-family: ${fontFamilies.default};
+  font-size: ${typeScales.body1.fontSize}px;
```
shaneeza, line 170: Should this font size change when the `baseFontSize` changes?

packages/code/src/Code/Code.styles.ts
```diff
@@ -112,6 +154,47 @@ export function getCodeWrapperVariantStyle(theme: Theme): string {
   `;
 }
 
+export const expandButtonStyle = css`
```
shaneeza, line 157: Can we add transitions to this button?

packages/code/src/Code/Code.styles.ts
```diff
@@ -112,6 +154,47 @@ export function getCodeWrapperVariantStyle(theme: Theme): string {
...
+    slight gap between the button and the code wrapper. Decreasing by
+    1px removes gap without making focused state look off.
+  */
+  border-radius: 0 0 11px 11px;
+  border-width: 1px 0 0 0;
+  border-style: solid;
+  display: flex;
+  font-family: ${fontFamilies.default};
+  font-size: ${typeScales.body1.fontSize}px;
+  gap: 4px;
```
shaneeza, line 171: Whenever possible, can you use spacing tokens?

packages/code/src/Code/Code.styles.ts
tsck: I didn't see anything communicated about a transition on expansion/collapse. I've reached out to Rob to get his thoughts.

packages/code/src/Code.stories.tsx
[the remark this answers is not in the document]
tsck: Will do!

packages/code/src/Code/Code.spec.tsx
tsck: Good point! I have to get into the design system mindset. I generally try to avoid test ID's to avoid bloating markup for test purposes, but what you're saying makes sense given how consumers might use it. Will update!

packages/code/src/Code/Code.styles.ts
tsck: Yup. If there are less than 5 lines of code there won't be an expand/collapse button, even if `expandable` is `true`. So the collapsed version will always be 5 lines when visible.

packages/code/src/Code/Code.styles.ts
tsck: Good catch! I would think that it should. I'll update this to include that.

packages/code/src/Code/Code.styles.ts
tsck: Oh absolutely, will update shortly

packages/code/src/Code/Code.styles.ts
```diff
@@ -112,6 +154,47 @@ export function getCodeWrapperVariantStyle(theme: Theme): string {
...
+  const colors = variantColors[theme];
+
+  return css`
+    background-color: ${colors[0]};
+    border-color: ${colors[1]};
+    color: ${colors[2]};
+    &:hover {
+      background-color: ${colors[1]};
+    }
+    &:focus {
```
shaneeza, line 190: Using `focus-visible` instead of `focus` is preferable here, as it prevents focus styles from appearing on mouse clicks.

packages/code/src/Code/Code.styles.ts
tsck: Is there a transition that we typically use for buttons? I'll have a look and see if I can find something, but if there's a standard here that you could share, that would be helpful.

packages/code/src/Code/Code.styles.ts
```diff
@@ -112,6 +154,47 @@ export function getCodeWrapperVariantStyle(theme: Theme): string {
...
+      background-color: ${
```
shaneeza, line 191: Would any of our color tokens work here? https://github.com/mongodb/leafygreen-ui/blob/main/packages/tokens/src/color/color.ts

packages/code/src/Code/Code.tsx
```diff
@@ -115,12 +128,20 @@ function Code({
     }
   }, []);
 
+  useIsomorphicLayoutEffect(() => {
+    if (!expandable) return;
+
+    const codeContent = document.getElementById('code-content');
```
shaneeza, line 134: Is it possible to use a `ref` here instead of an `id`?

packages/code/src/Code/Code.tsx
```diff
@@ -5,9 +5,12 @@ import PropTypes from 'prop-types';
 
 import { cx } from '@leafygreen-ui/emotion';
 import { useIsomorphicLayoutEffect } from '@leafygreen-ui/hooks';
+import ChevronDown from '@leafygreen-ui/icon/dist/ChevronDown';
+import ChevronUp from '@leafygreen-ui/icon/dist/ChevronUp';
```
shaneeza, line 9: If we add transitions, then we could mimic Stephen's changes in #2443

packages/code/src/Code/Code.styles.ts
tsck: Just caught this with Rob actually! Will do

packages/code/src/Code/Code.styles.ts
tsck: Discussed with Rob and will be adding a 300-500ms animation on expansion/collapse here

packages/code/src/Code/Code.tsx
tsck: I like it, will do!

packages/code/src/Code/Code.tsx
```diff
@@ -163,6 +188,13 @@ function Code({
       : { usePortal }),
   } as const;
 
+  const collapsedLinesOfCode = 5;
```
shaneeza, line 191: not: since this will always be 5, maybe we should make this a global constant since you're also using it in styles?

packages/code/src/Code/Code.styles.ts
shaneeza: Can you add this to the tsdoc and README?

shaneeza: Also, remember to add a changeset

packages/code/src/Code/Code.tsx
```diff
@@ -219,6 +258,19 @@ function Code({
               {...popoverProps}
             />
           )}
+
+          {showExpandButton && (
+            <button
```
shaneeza, line 263: nit: you could break this out into a new component.

shaneeza (changes requested):
Looks good so far! Remember to update the README and add a changeset.

packages/code/src/Code/Code.styles.ts
tsck: I was able to find one for the background, but I'm not sure I'm seeing one for the color. Feel free to let me know if I'm missing that though.

packages/code/src/Code/Code.tsx
tsck: Yup, definitely is, and it's helpful for calculating the height now that we'll be transitioning. Great point!

packages/code/src/Code/Code.tsx
tsck: I considered this, but really there wasn't much logic going on here so it didn't seem very helpful. If we decide to rotate the chevron, there may be more that will justify this though. I'll update shortly when I have more thoughts 🙂

packages/code/src/Code/Code.tsx
tsck: I'm not sure the usage here is quite the same. This is primarily because both the text and the icon change on expansion/collapse. So rotating the icon alone doesn't inherently make the transition look less jarring. I have an idea and want to play around with a few possible solutions. I'll push those on separate commits before requesting another review, and probably run them by design to see what looks best here.

packages/code/src/Code/Code.tsx
tsck: This makes sense to m. Could you point me to where we store global constants. I see `lib/` seems to be similar but I'm not quite sure that's correct.

shaneeza: You'll notice that some components have a `constants.ts` file — you can add a constants.ts file inside `code/src`.

packages/code/src/Code.stories.tsx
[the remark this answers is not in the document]
tsck: I chose not to do this because we don't follow the same pattern for `copyable`, which feels like it would be a similar use case here. If this is something we should still do please let me know, happy to discuss!

packages/code/src/Code/Code.styles.ts
```diff
@@ -98,6 +124,18 @@ export const singleLineCodeWrapperStyle = css`
...
 `;
 
+export function getExpandableCodeWrapperStyle(
+  expanded: boolean,
+  codeHeight: number,
+  collapsedCodeHeight: number,
+) {
+  return css`
+    max-height: ${expanded ? codeHeight : collapsedCodeHeight}px;
+    overflow-y: scroll;
```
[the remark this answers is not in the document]
tsck: This is a really good point! I don't think there is! Let me remove the scrollbar. I'll have Rob give another look anyway before merging but I think this is a good catch.

packages/code/src/Code/Code.styles.ts
shaneeza: Was there an issue with adding transitions?

packages/code/README.md
```diff
@@ -115,6 +115,7 @@ const SomeComponent = () => <Code>{codeSnippet}</Code>;
...
 | `copyable`                | `boolean`                                                                                                                                                                                                                                                                 | When true, allows the code block to be copied to the user's clipboard                                                                                                                                                                           | `true`  |
+| `expandable`              | `boolean`                                                                                                                                                                                                                                                                 | When true, allows the code block to be expanded and collapsed.                                                                                                                                                                                  | `false` |
```
shaneeza, line 118: Can you add that this will happen if there are more than five lines of code?

packages/code/src/Code/Code.tsx
```diff
@@ -37,6 +48,10 @@ export function hasMultipleLines(string: string): boolean {
   return string.trim().includes('\n');
 }
 
+function getHorizontalScrollbarHeight(element: HTMLElement): number {
```
shaneeza, line 51: nit: you could create a `utils` folder inside `code/src` and add this there. `Combobox` is a good example.

packages/code/src/Code/Code.styles.ts
tsck: Oops, nope! I think I just accidentally removed it when I removed the transition for the icon rotation. Will add it back now.

packages/code/src/types.ts
```diff
@@ -20,7 +20,11 @@ export type LineHighlightingDefinition = ReadonlyArray<
   number | readonly [number, number]
 >;
 
-export interface SyntaxProps extends HTMLElementProps<'code'> {
+interface HTMLElementProps<T extends HTMLElement> extends HTMLAttributes<T> {
+  ref?: React.Ref<T>;
+}
+
+export interface SyntaxProps extends HTMLElementProps<HTMLPreElement> {
```
shaneeza, line 27: Instead of `HTMLElementProps`, could you use `ComponentPropsWithRef`?

packages/code/src/Code/Code.styles.ts
tsck: Just a note - Rob has indicated that the button `font-size` should always be `13px`. Will be changing this back shortly.

packages/code/src/types.ts
```diff
@@ -101,10 +110,17 @@ export type CodeProps = Omit<
   /**
    * When true, allows the code block to be copied to the user's clipboard by clicking the rendered copy button.
    *
-   * default: `true`
+   * @default: `true`
    */
   copyable?: boolean;
 
+  /**
+   * Makes long code blocks expandable
```
shaneeza, line 118: Could you add that this only works if there are 5 or more lines of code?

packages/code/src/Code/Code.tsx
tsck: So the goal in my mind when moving this into a named function wasn't so much to make it reusable, but rather to make it clear what `element.offsetHeight - element.clientHeight` is calculating. I'm not sure if this is a utility method that will be used elsewhere, so I think there's at least some benefit to keeping it close to its usage. Happy to discuss further if this is a well established pattern we follow!

.changeset/fresh-comics-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/code': minor
+---
+
+Adds `expandable` prop to `Code` component
```
shaneeza, line 5: Can you add more details on what the expandable prop does and link the JIRA ticket? [Here's an example of that.](https://github.com/mongodb/leafygreen-ui/blob/main/packages/tabs/CHANGELOG.md#1300)

packages/code/src/Code/Code.styles.ts
```diff
@@ -112,6 +150,48 @@ export function getCodeWrapperVariantStyle(theme: Theme): string {
...
+    Code wrapper's border radius is 12px. Matching that creates a very 
+    slight gap between the button and the code wrapper. Decreasing by
+    1px removes gap.
+  */
+  border-radius: 0 0 11px 11px;
+  border-width: 1px 0 0 0;
+  border-style: solid;
+  display: flex;
+  font-family: ${fontFamilies.default};
+  font-size: 13px;
```
shaneeza, line 166: let's use a [token](https://github.com/mongodb/leafygreen-ui/blob/main/packages/tokens/src/typeScales.ts#L1) here

packages/code/src/Code/Code.styles.ts
```diff
@@ -112,6 +150,48 @@ export function getCodeWrapperVariantStyle(theme: Theme): string {
...
+  gap: ${spacing[100]}px;
+  grid-area: expandButton;
+  justify-content: center;
+  transition: all ${transitionDuration.default}ms ease-in-out;
+  z-index: 2;
```
shaneeza, line 171: Whats the z-index for?

packages/code/src/Code/Code.tsx
shaneeza: I think it's okay that it's not reusable; it's more to avoid bloating the file, but I don't feel too strongly about this.

packages/code/src/Code/Code.styles.ts
shaneeza: From what I've seen, we usually wait until a feature is fully supported before we use it, so I would hold off on using `auto`

packages/code/src/Code/Code.styles.ts
tsck: It moves the button above the shadows. You'll see this done throughout the component in other styles. I've added a comment to indicate this.

packages/code/src/Code/Code.spec.tsx
```diff
@@ -237,4 +238,77 @@ describe('packages/Code', () => {
...
+      ).join('\n');
+
+    test(`shows no expand button when <= ${numOfCollapsedLinesOfCode} lines of code`, () => {
+      render(
+        <Code expandable={true} language="javascript">
+          {getCodeSnippet(numOfCollapsedLinesOfCode - 1)}
+        </Code>,
+      );
+
+      expect(screen.queryByTestId('expand-button')).toBeNull();
```
shaneeza, line 256: I forgot to mention this earlier, but we recently decided to follow a [BEM-ish pattern](https://github.com/mongodb/leafygreen-ui/blob/main/STYLEGUIDE.md#follow-bem-ish-patterns-when-hard-coding-a-data-testid-or-data-lgid) for testids. Following the pattern, we can change this to: `lg-code-expand_button`.

packages/code/src/types.ts
```diff
@@ -50,6 +50,11 @@ export interface SyntaxProps extends HTMLElementProps<'code'> {
    * An array of lines to highlight. The array can only contain numbers corresponding to the line numbers to highlight, and / or tuples representing a range (e.g. `[6, 10]`);
    */
   highlightLines?: LineHighlightingDefinition;
+
+  /**
+   * A ref to the underlying `code` element.
+   */
+  ref?: React.Ref<HTMLPreElement>;
```
shaneeza, line 57: You can remove this line since you're using `ComponentPropsWithRef`

packages/code/src/Code/Code.tsx
tsck: I can see the trade-off here. Given that it's just a one line function created for readability, and is similar with the other util method above it, I don't love the idea of putting it further away from the usage, causing navigation between files to be necessary in order to gain context. I think I'd see benefit in just in-lining it in that case to be more in line with the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

packages/code/src/types.ts
tsck: On second thought, this is no longer needed at all. I was originally adding a `ref` to access a `Sytnax` child, but found a way around this by using another reference. I'm going to revert these changes entirely.

audroue (approved):
LGTM!

shaneeza (approved):
✨

PR #2467, opened 2024-09-04
[LG-4473]: Tab selection with name

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4473](https://jira.mongodb.org/browse/LG-4473)

### For new components


packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -105,8 +119,8 @@ const Tabs = (props: AccessibleTabsProps) => {
 
       const enabledIndices = getEnabledIndices(tabTitleElements);
       const numberOfEnabledTabs = enabledIndices.length;
-      const activeIndex = enabledIndices.indexOf(selected);
-
+      const normalizedSelectedState = getNormalizedSelectedState(selected);
```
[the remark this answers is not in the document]
shaneeza: makes sense

packages/tabs/src/Tabs.stories.tsx
```diff
@@ -109,7 +109,7 @@ const meta: StoryMetaType<typeof Tabs> = {
     },
   },
   // TODO: Add subcomponent controls for Tab when supported by Storybook
-  subcomponents: { tab: Tab },
+  // subcomponents: { tab: Tab },
```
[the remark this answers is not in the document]
shaneeza: It was giving a TS error.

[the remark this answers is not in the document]
TheSonOfThomp: Yeah we can rm.
This was a placeholder until we're on SB8 https://storybook.js.org/docs/writing-stories/stories-for-multiple-components#subcomponents

packages/tabs/src/utils/getSelectedIndex/getSelectedIndex.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export const getSelectedIndex = (
+  selected: number | string,
+  tabTitleElements: Array<HTMLElement>,
+) => {
+  if (typeof selected === 'number') return selected;
+
+  return tabTitleElements.findIndex(
+    element => element.dataset.text === selected,
+  );
```
TheSonOfThomp, line 14: What happens if there are multiple tabs with the same name? 
I'm fine with the default `.findIndex` behavior (resolve to the first one), but we should add a unit test for this case.

packages/tabs/src/utils/getSelectedIndex/getSelectedIndex.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @param tabTitleElements
+ * @returns Returns the corresponding index for a selected string
+ */
+export const getSelectedIndex = (
+  selected: number | string,
+  tabTitleElements: Array<HTMLElement>,
+) => {
+  if (typeof selected === 'number') return selected;
+
+  return tabTitleElements.findIndex(
```
TheSonOfThomp, line 12: Also, what happens if we don't find anything?

packages/tabs/src/Tabs.stories.tsx
```diff
@@ -129,14 +129,28 @@ export const LiveExample: StoryFn<TabsProps> = ({
 );
 
 export const Controlled: StoryFn<TabsProps> = (args: TabsProps) => {
-  const [selectedTab, setSelectedTab] = useState(0);
+  const [selectedTab, setSelectedTab] = useState<string | number>(
```
TheSonOfThomp, line 132: Can a consumer switch between using strings & numbers to reference tabs?

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -90,12 +90,13 @@ const Tabs = (props: AccessibleTabsProps) => {
 
   const handleClickTab = useCallback(
     (e: React.SyntheticEvent<Element, MouseEvent>, index: number) => {
-      setSelected?.(index);
+      (setSelected as React.Dispatch<number>)?.(index);
```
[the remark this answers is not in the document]
TheSonOfThomp: This is a bit tricky. Currently it seems like regardless if a consumer passes in a string, their state value will be set to a number when the tab is changed

packages/tabs/src/utils/getSelectedIndex/getSelectedIndex.ts
shaneeza: i'll add a test.

packages/tabs/src/utils/getSelectedIndex/getSelectedIndex.ts
shaneeza: It's the same as passing an index that doesn't exist—nothing will be selected. I can document this.

packages/tabs/src/Tabs.stories.tsx
shaneeza: Yes, they can.

packages/tabs/src/Tabs/Tabs.tsx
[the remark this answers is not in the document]
shaneeza: Do you think it makes more sense if the consumer passes a string, then we should also return a string? That should be possible.

TheSonOfThomp (commented):
My main concern is that the `selected` data provided by the consumer won't be the same as what's passed into `setSelected`.

We should add checks to ensure the data passed to the setter is the same type that the consumer expects.

The following code is valid, and won't raise any obvious errors, but could cause issues
```tsx
  const [selectedTab, setSelectedTab] = useState<string>('Tab 1');

  useEffect(() => {
    console.log( selectedTab ); // logs some some number, but we expect it to be a string
  }, [selectedTab]);

  return (
    <Tabs 
      selected={selectedTab}
      setSelected={setSelectedTab}
   />
  )
```

packages/tabs/src/Tabs/Tabs.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Yeah, we should add a check that ensures that the type of `selected` matches the type passed into `setSelected`

packages/tabs/src/Tabs.spec.tsx
```diff
@@ -163,12 +170,35 @@ describe('packages/tabs', () => {
...
+      const { getSelectedPanel } = renderTabs({
+        setSelected,
+        selected: 'Third',
+      });
+      const selectedPanel = getSelectedPanel();
+      expect(selectedPanel).toHaveTextContent('Content 3');
+    });
+
     test('clicking a tab fires setSelected callback', () => {
+      const setSelected = jest.fn();
```
TheSonOfThomp, line 201: we should assert that it gets called with the correct data

packages/tabs/src/Tabs/Tabs.types.ts
```diff
@@ -64,16 +64,16 @@ export interface TabsProps
...
+  selected?: number | string;
 
   /**
-   * Callback to be executed when Tab is selected. Receives index of activated Tab as the first argument.
+   * Callback to be executed when Tab is selected. Receives index or name of activated Tab as the first argument.
    *
    * @type (index: number) => void
    */
-  setSelected?: React.Dispatch<number>;
+  setSelected?: React.Dispatch<number> | React.Dispatch<string>;
```
TheSonOfThomp, line 76: Can we update these with a type param. so that they're always in sync (i.e. `selected: T` and `setSelected: Dispatch<T>`

packages/tabs/src/Tabs.spec.tsx
```diff
@@ -163,12 +170,35 @@ describe('packages/tabs', () => {
...
     });
+
+    test('selected tab panel is active on first render using a string', () => {
+      const setSelected = jest.fn();
+      const { getSelectedPanel } = renderTabs({
+        setSelected,
+        selected: 'Second',
+      });
+      const selectedPanel = getSelectedPanel();
+      expect(selectedPanel).toHaveTextContent('Content 2');
```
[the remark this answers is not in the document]
shaneeza: Panels don't have testids on them and consumers can't add testids to them but I updated panels to reuse the testid of the corresponding `Tab` child. For example, if`Tab` has a testid of `second-content`, the panel will get a testid of `second-content-panel`. I kept the content check but also added a check for for the testid.

packages/tabs/src/Tabs/Tabs.tsx
```diff
@@ -83,20 +85,35 @@ const Tabs = (props: AccessibleTabsProps) => {
...
   const accessibilityProps = {
     ['aria-label']: ariaLabel,
     ['aria-labelledby']: ariaLabelledby,
   };
 
+  /** If the controlled state is a string then return a string, else return a number */
+  const setSetSelectedValue = useCallback(
+    (index: number) => {
+      if (typeofSelected === 'string') {
+        const indexString = tabTitleElements[index].dataset.text!;
```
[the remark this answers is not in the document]
shaneeza: I think that even `text` would work

packages/tabs/src/Tabs.spec.tsx
```diff
@@ -60,6 +56,30 @@ describe('packages/tabs', () => {
   });
 
   describe('rendering', () => {
+    test('consoles errors if multiple tabs have the same name', () => {
```
TheSonOfThomp, line 59: What if they have different HTML, but the same text content (i.e. `<span>Third</span>` vs `<div>Third</div>`? I think this should work, but can we add a test?

TheSonOfThomp (approved):

PR #2468, opened 2024-09-05
Terrence/fix undefined window

tsck:
## ✍️ Proposed changes

When server side rendering is used, `window` is not defined. This is causing build issues on the server where we access `window` in `useViewportSize`. To fix this, this PR adds a hook to check if we're on the server or client. Placing this logic in a hook allows for the `useEffect` callback to be re-run on the client and properly updated when server rendered.

Note to reviewer: Would love a 2nd opinion on rendering here. I don't believe there should be an issue, since this is just essentially reverting to a previous state where we do an initial set to `null`, but adds logic to update that IF rendered on the client. However, I want to make sure I'm not missing something that will cause a disruption on render.

- [ x ] I have run `yarn changeset` and documented my changes

### For new components

### DIFF
--- a/.changeset/short-eels-deliver.md
+++ b/.changeset/short-eels-deliver.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/hooks': patch
+---
+
+When server side rendering is used, `window` is not defined. This is causing build issues on the server where we access `window` in `useViewportSize`. To fix this, this change adds a hook, `useIsSsr`, that checks the rendering environment and can be used before attempting to access `window`. It adds a check of this to `useViewportSize` to fix the current build issue.
--- a/packages/hooks/src/useIsSsr.ts
+++ b/packages/hooks/src/useIsSsr.ts
@@ -1,0 +1,15 @@
+import { useEffect, useState } from 'react';
+
+export default function useIsSsr() {
+  const [isSsr, setIsSsr] = useState(typeof window === 'undefined');
+
+  useEffect(() => {
+    // When rendered on server, this won't run until we're on the client. Therefore,
+    // isSsr should be true when server rendered, and only be set to false on subsequent client render.
+    // When rendered directly on the client, isSsr should already be false, so
+    // this update shouldn't trigger a re-render.
+    setIsSsr(false);
+  }, []);
+
+  return isSsr;
+}
--- a/packages/hooks/src/index.ts
+++ b/packages/hooks/src/index.ts
@@ -16,3 +16,4 @@ export { default as usePrevious } from './usePrevious';
 export { useStateRef } from './useStateRef';
 export { default as useValidation } from './useValidation';
 export { default as useViewportSize } from './useViewportSize';
+export { default as useIsSsr } from './useIsSsr';
--- a/packages/hooks/src/useViewportSize.ts
+++ b/packages/hooks/src/useViewportSize.ts
@@ -1,5 +1,6 @@
 import { useEffect, useState } from 'react';
 import debounce from 'lodash/debounce';
+import useIsSsr from './useIsSsr';
 
 interface ViewportSize {
   width: number;
@@ -14,10 +15,10 @@ function getViewportSize(): ViewportSize {
 }
 
 export default function useViewportSize(): ViewportSize | null {
-  const isRenderingServerSide = typeof window === 'undefined';
+  const isSsr = useIsSsr();
 
   const [viewportSize, setViewportUpdateVal] = useState<ViewportSize | null>(
-    isRenderingServerSide ? null : getViewportSize(), // window undefined on server
+    isSsr ? null : getViewportSize(), // window undefined on server
   );
 
   useEffect(() => {
--- a/packages/hooks/src/useViewportSize.ts
+++ b/packages/hooks/src/useViewportSize.ts
@@ -13,9 +13,11 @@ function getViewportSize(): ViewportSize {
   };
 }
 
-export default function useViewportSize(): ViewportSize {
-  const [viewportSize, setViewportUpdateVal] = useState<ViewportSize>(
-    getViewportSize(),
+export default function useViewportSize(): ViewportSize | null {
+  const isRenderingServerSide = typeof window === 'undefined';
+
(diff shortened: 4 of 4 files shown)


packages/hooks/src/useIsSsr.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react';
+
+export default function useIsSsr() {
```
tsck, line 3: Could just make this `useRenderingEnv` or something like similar, but this seemed easier.

shaneeza: is `use` referring to the `user`?

TheSonOfThomp: I think that's just that hook syntax

TheSonOfThomp: I personally don't like the `useIs...` naming for hooks. I like `useRenderEnv` or something

TheSonOfThomp, line 3: Can we add some tests? 

I wonder if this is what we're after: https://testing-library.com/docs/react-testing-library/api/#hydrate

tsck: I hate it as well lol. I think the only downside of some `useEnv` alternative is that I would expect that to return an env enum-like string as opposed to just a boolean. How about `useSsrCheck`? The might read better `const isSsr = useSsrCheck()`

tsck: Unfortunately, it seems like [this wouldn't work](https://github.com/testing-library/react-testing-library/issues/1340). It also seems that SSR testing is a long-standing issue both for [hook](https://github.com/testing-library/react-testing-library/issues/1120) and [component](https://github.com/testing-library/react-testing-library/issues/561) testing, neither of which appear to be getting resolved anytime soon based on the issues.

Going off of some threaded workarounds I found, I've added a `renderHookServer` method to `@leafygreen/testing-lib`. It appears to solve the issue for SSR tests in a JSDOM env. Take a look when you get a chance, would love to get your thoughts!

tsck:
Tests failing in React 17 possibly due to usage of `hydrateRoot`. The React 17 equivalent I believe is `hydrate` ([reference](https://react.dev/reference/react-dom/hydrate)). I believe that `@testing-library/react-hooks/server` exports its own `render` method that works the same as this React 18+ version that I added (see initial question [here](https://github.com/testing-library/react-testing-library/issues/1120#issue-1364219273)). Fix should be to use that pre-react 18, and the shim post. Will look into this shortly.

packages/hooks/src/useIsSsr.ts
TheSonOfThomp: I like `useSSRCheck` 👍

packages/hooks/src/hooks.spec.tsx
```diff
@@ -368,4 +369,13 @@ describe('packages/hooks', () => {
       expect(result.current.onChange).toBeDefined();
     });
   });
+
+  describe('useSsrCheck', () => {
+    it('should return true when server-side rendered and false after hydration', () => {
```
TheSonOfThomp, line 374: nit: I _think_ we prefer `test()` vs `it()`. Not sure if there's any difference other than stylistic

packages/testing-lib/src/renderHookServer.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+ */
+export function renderHookServer<Hook extends () => any>(
+  useHook: Hook,
+  {
+    wrapper: Wrapper,
+  }: {
+    wrapper?: ({ children }: { children: ReactNode }) => JSX.Element;
+  } = {},
+): { result: { current: ReturnType<Hook> }; hydrate: () => void } {
```
TheSonOfThomp, line 38: nit: can we define these types separately?

packages/testing-lib/src/renderHookServer.tsx
```diff
@@ -0,0 +1,88 @@
...
+  ) : (
+    <Component useHook={useHook} />
+  );
+
+  // Running tests in an environment that simulates a browser (like Jest with jsdom),
+  // the window object will still be available even when server rendered. To ensure
+  // that window is not available during SSR we need to explicitly mock or remove the
+  // window object.
+  // @ts-ignore Type 'undefined' is not assignable to type 'Window'.
+  jest.spyOn(global, 'window', 'get').mockImplementation(() => undefined);
```
TheSonOfThomp, line 67: eww lol

TheSonOfThomp:
>... Will look into this adding this in `RTLOverrides` shortly

I like this approach. LMK if you need extra eyes adding this shim

packages/hooks/src/hooks.spec.tsx
tsck: Oh good call out, totally missed this. Will update!

packages/testing-lib/src/renderHookServer.tsx
tsck: Lol honestly I hate this too. I was going to specifically flag this as a like "should we do this?" or "is there a better way?" but I'm not sure

tsck: Alternative could be to not do this here and do it in the test.

tsck:
Failing test unrelated to previous rollup issue. Looking into this now.

tsck:
Couldn't reproduce the failing `Popover` test locally, and now it's passing. Perhaps it was just flaky.

.changeset/mighty-clouds-impress.md
```diff
@@ -0,0 +1,15 @@
+---
+'@leafygreen-ui/testing-lib': patch
```
TheSonOfThomp, line 2: I think this should be a `minor`

.changeset/short-eels-deliver.md
```diff
@@ -0,0 +1,7 @@
+---
+'@leafygreen-ui/hooks': patch
```
TheSonOfThomp, line 2: Also should be a minor

TheSonOfThomp (approved):
Just bump both changesets to `minor` and this should be good to go!

PR #2475, opened 2024-09-10
LG-4507: fix autocomplete styles in FormFieldInputContainer

stephl3:
## ✍️ Proposed changes
We received a report of a UI bug in form field components that are using the `FormFieldInputContainer`. When the descendant input element is autocompleted,`:-webkit-autofill` pseudo-selector is applied to the element. Because we use `FormFieldInputContainer` to handle the styling of the form field, some of the styles are being duplicated in the input element`. The issue is especially observable in `TextInput` because the input element doesn't fully stretch the space in `FormFieldInputContainer`.

We also bump `@leafygreen-ui/text-area` and `@leafygreen-ui/date-picker` which similarly use `FormFieldInputContainer` although there is less impact because there aren't uses of those components with the `autoComplete` prop.

🎟 _Jira ticket:_ [LG-4507](https://jira.mongodb.org/browse/LG-4507)

## 🧪 How to test changes

- [ ] test autocomplete functionality in `TextInput` storybook instance
- [ ] test autocomplete functionality in `TextArea` storybook instance
- [ ] test autocomplete functionality in `DatePicker` storybook instance

### DIFF
--- a/.changeset/late-houses-work.md
+++ b/.changeset/late-houses-work.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/form-field': patch
+---
+
+[LG-4507](https://jira.mongodb.org/browse/LG-4507): fix `FormFieldInputContainer` autocomplete styling used in `DateInputSegment`, `TextArea` and `TextInput`. A double-border issue occurs when the auto-completed input has focus or hover state applied


.changeset/late-houses-work.md
```diff
@@ -0,0 +1,8 @@
+---
+'@leafygreen-ui/form-field': patch
+'@leafygreen-ui/date-picker': patch
```
shaneeza, line 3: I don't think that you need to include date-picker, text-area, and text-input since you're not manually bumping the dependency version. The version PR should automatically bump the version and add a changeset.

shaneeza (approved):
Looks good - I left a minor comment about the changeset.

PR #2482, opened 2024-09-18
LG-4445: add renderMode and top layer

stephl3:
## ✍️ Proposed changes

- adds `renderMode`, `dismissMode`, and `onToggle` props to `Popover` component
  - `usePortal` currently overrides `renderMode` prop but will be dropped in an upcoming PR
- adds support for `renderMode="top-layer"`
- changesets will be consolidated in upcoming PR

🎟 _Jira ticket:_ [LG-4445](https://jira.mongodb.org/browse/LG-4445)

## 🧪 How to test changes

- [ ] Review `Popover` stories
- [ ] Review UI Tests
- [ ] Check for regressions
  - [ ] `Combobox/LiveExample`
  - [ ] `DatePicker/LiveExample`
  - [ ] `GuideCue/LiveExample`
  - [ ] `NumberInput/LiveExample`
  - [ ] `SearchInput/LiveExample`
  - [ ] `Chat/InputBar/WithDropdown`


packages/popover/src/Popover.styles.ts
```diff
@@ -78,33 +95,91 @@ const transformOriginStyles: Record<ExtendedPlacement, string> = {
   `,
 };
 
-const getTransformStyles = (placement: ExtendedPlacement, spacing: number) =>
-  cx({
-    [css`
+const getClosedStyles = (placement: ExtendedPlacement, spacing: number) => {
+  if (placement.startsWith('top')) {
```
TheSonOfThomp, line 99: I think breaking `placement` into its constituent `align` and `justify`, and then using a `switch` here would make this a bit more readable, and would likely be easier to test

packages/popover/src/Popover.styles.ts
```diff
@@ -78,33 +95,91 @@ const transformOriginStyles: Record<ExtendedPlacement, string> = {
...
+
+const getTransitionStyles = (
+  placement: ExtendedPlacement,
+  spacing: number,
+) => ({
+  exited: getClosedStyles(placement, spacing),
+  entering: getClosedStyles(placement, spacing),
+  entered: getOpenStyles(placement),
+  exiting: getClosedStyles(placement, spacing),
+  unmounted: getClosedStyles(placement, spacing),
```
TheSonOfThomp, line 181: I noticed a slight delay after clicking before the popover started transitioning. I wonder if `entering` should have `getOpenStyles`?

packages/popover/src/Popover.styles.ts
```diff
@@ -122,16 +197,14 @@ export const getPopoverStyles = ({
...
     transformOriginStyles[placement],
+    getTransitionStyles(placement, spacing)[state],
+
     {
-      [getTransformStyles(placement, spacing)]:
-        state === 'exiting' || state === 'entering',
-      [closedStyles]: state === 'exiting' || state === 'exited',
-      [openStyles]: state === 'entering' || state === 'entered',
       [css`
         z-index: ${popoverZIndex};
```
TheSonOfThomp, line 206: should z-index have a default?

packages/popover/src/Popover.hooks.tsx
```diff
@@ -4,14 +4,80 @@ import {
...
+import { getRenderMode } from './utils/getRenderMode';
 import { getElementDocumentPosition } from './utils/positionUtils';
 import {
   PopoverProps,
+  RenderMode,
   UseContentNodeReturnObj,
   UseReferenceElementReturnObj,
 } from './Popover.types';
 
+export function usePopoverContextProps(
```
TheSonOfThomp, line 18: can we add some TSDoc

packages/popover/src/Popover.hooks.tsx
```diff
@@ -4,14 +4,80 @@ import {
...
+    onEnter,
+    onEntering,
+    onEntered,
+    onExit,
+    onExiting,
+    onExited,
+    popoverZIndex,
+    spacing,
+    ...rest
+  } = props;
```
TheSonOfThomp, line 51: why not destructure these in the function signature?

packages/popover/src/Popover.hooks.tsx
TheSonOfThomp, line 1: All these `Popover.xxx` files should live in `Popover/` (except maybe the stories file)

packages/leafygreen-provider/src/PopoverContext/PopoverContext.types.ts
TheSonOfThomp, line 1: @shaneeza What was the initial reasoning for having PopoverContext live in `leafygreen-provider` instead of within the `popover` package?

packages/leafygreen-provider/src/PopoverContext/PopoverContext.types.ts
```diff
@@ -15,68 +15,117 @@ type TransitionLifecycleCallbacks = Pick<
...
+  Manual: 'manual',
+} as const;
+export type DismissMode = (typeof DismissMode)[keyof typeof DismissMode];
+
+/** Local implementation of web-native `ToggleEvent` until we use typescript v5 */
+export interface ToggleEvent extends Event {
+  type: 'toggle';
+  newState: 'open' | 'closed';
+  oldState: 'open' | 'closed';
+}
```
TheSonOfThomp, line 29: I think these types should all live in the popover package

TheSonOfThomp: Why are they duplicated here?

packages/popover/src/Popover.types.ts
```diff
@@ -11,6 +11,39 @@ type TransitionLifecycleCallbacks = Pick<
   'onEnter' | 'onEntering' | 'onEntered' | 'onExit' | 'onExiting' | 'onExited'
 >;
 
+/**
+ * Options to render the popover element
+ * @param Inline will render the popover element inline with the reference element
```
TheSonOfThomp, line 16: `inline` can have different meanings. Can we be clear that we mean "in the DOM where it's written" and not "display: inline`

packages/popover/src/Popover.types.ts
```diff
@@ -11,6 +11,39 @@ type TransitionLifecycleCallbacks = Pick<
...
+ * @param Portal will render the popover element in a provided `portalContainer` or
+ * in a new div appended to the body
```
TheSonOfThomp, line 18: nit: should lead with the default case (in a div appended to the body)

packages/popover/src/Popover.types.ts
```diff
@@ -11,6 +11,39 @@ type TransitionLifecycleCallbacks = Pick<
...
+ * @param Manual will require that the consumer handle dismissal manually
+ */
+export const DismissMode = {
+  Auto: 'auto',
+  Manual: 'manual',
+} as const;
+export type DismissMode = (typeof DismissMode)[keyof typeof DismissMode];
+
+/** Local implementation of web-native `ToggleEvent` until we use typescript v5 */
+export interface ToggleEvent extends Event {
```
TheSonOfThomp, line 41: I don't think this extends `Event`, does it? (i.e. it doesn't have `e.target` etc. iirc)

packages/popover/src/Popover.types.ts
```diff
@@ -70,65 +103,115 @@ export interface ChildrenFunctionParameters {
...
-       * A ref for the portal element
-       */
-      portalRef?: undefined;
-
-      /**
-       * When usePortal is `true`, specifies the scrollable element to position relative to.
-       */
-      scrollContainer?: null;
-    };
+/** @deprecated - use {@link RenderTopLayerProps} */
```
TheSonOfThomp, line 106: Could potentially add more on why we recommend TopLayerProps

packages/popover/src/Popover.types.ts
```diff
@@ -70,65 +103,115 @@ export interface ChildrenFunctionParameters {
...
+  scrollContainer?: never;
+
+  /** Not used in this `renderMode` */
+  usePortal?: never;
+}
+
+export type PopoverRenderModeProps =
+  | RenderInlineProps
+  | RenderPortalProps
+  | RenderTopLayerProps;
```
TheSonOfThomp, line 214: I don't believe the `deprecated` flags above will be seen by consumers, since we're grouping them into this type here

packages/popover/src/Popover/Popover.tsx
```diff
@@ -57,42 +63,51 @@ export const Popover = forwardRef<HTMLDivElement, PopoverComponentProps>(
...
-      scrollContainer: scrollContainerCtxVal,
+      /** style props */
+      popoverZIndex,
+      spacing = spacingToken[100],
+      /** deprecated props */
+      isPopoverOpen: _,
       setIsPopoverOpen,
-    } = usePopoverContext();
+      ...restProps
+    } = usePopoverContextProps(rest, popoverContext);
```
TheSonOfThomp, line 102: does this need to be a hook?

packages/popover/src/Popover/Popover.tsx
```diff
@@ -57,42 +63,51 @@ export const Popover = forwardRef<HTMLDivElement, PopoverComponentProps>(
...
-    // show a warning if the portalContainer is not inside of the scrollContainer.
-    // Note: If no portalContainer is passed the portalContainer will be undefined and this warning will show up.
-    // By default if no portalContainer is passed the <Portal> component will create a div and append it to the body.
+    /**
+     * When usePortal is true and a scrollContainer is defined, log a warning if the
+     * portalContainer is not inside of the scrollContainer.
+     * Note: If no portalContainer is passed the portalContainer will be undefined, and
+     * this warning will show up. By default if no portalContainer is passed the <Portal>
+     * component will create a div and append it to the body.
+     */
```
TheSonOfThomp, line 110: nit:
```ts
/**
 * When `usePortal` is true and a `scrollContainer` is defined,
 * log a warning if the `portalContainer` is not inside of the `scrollContainer`.
 *
 * Note: If no `portalContainer` is provided,
 * the `Portal` component will create a `div` and append it to the body.
 */
```

packages/popover/src/Popover/Popover.tsx
```diff
@@ -190,16 +231,24 @@ export const Popover = forwardRef<HTMLDivElement, PopoverComponentProps>(
             )}
```
TheSonOfThomp, line 231: do we still need this placeholder?

packages/popover/src/Popover/Popover.tsx
```diff
@@ -190,16 +231,24 @@ export const Popover = forwardRef<HTMLDivElement, PopoverComponentProps>(
...
                 })}
+                // @ts-expect-error - `popover` attribute is not typed in current version of `@types/react` https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69670
+                // eslint-disable-next-line react/no-unknown-property
+                popover={
+                  renderMode === RenderMode.TopLayer ? dismissMode : undefined
+                }
+                {...restProps}
               >
                 {/* We need to put `setContentNode` ref on this inner wrapper because
                 placing the ref on the parent will create an infinite loop in some cases
```
TheSonOfThomp, line 254: do we need to use `setContentNode`? Can we use a standard `useRef` RefObject?

packages/popover/src/Popover/Popover.tsx
```diff
@@ -160,14 +175,37 @@ export const Popover = forwardRef<HTMLDivElement, PopoverComponentProps>(
...
+      if (renderMode !== RenderMode.TopLayer || !onToggle) {
+        return;
+      }
+
+      // @ts-expect-error - `toggle` event not supported pre-typescript v5
+      refs.floating.current?.addEventListener('toggle', onToggle);
+      return () =>
+        // @ts-expect-error - `toggle` event not supported pre-typescript v5
+        refs.floating.current?.removeEventListener('toggle', onToggle);
+    }, [onToggle, renderMode]);
```
TheSonOfThomp, line 188: might need a eslint-ignore here to stop the linter complaining about `refs.floating`

packages/popover/src/Popover/Popover.tsx
```diff
@@ -160,14 +175,37 @@ export const Popover = forwardRef<HTMLDivElement, PopoverComponentProps>(
...
+        return;
+      }
+
+      if (context.open) {
+        // @ts-expect-error - Popover API not currently supported in react v18 https://github.com/facebook/react/pull/27981
+        refs.floating.current?.showPopover?.();
+      } else {
+        // @ts-expect-error - Popover API not currently supported in react v18 https://github.com/facebook/react/pull/27981
+        refs.floating.current?.hidePopover?.();
+      }
```
TheSonOfThomp, line 201: nit: might want to check whether `context.open` has changed, otherwise we'll be calling show/hide unnecessarily. Shouldn't be a big deal but idk how this API works under the hood

packages/popover/src/Popover/Popover.tsx
TheSonOfThomp, line 188: looks like we're updating the event listener on each toggle

packages/popover/src/Popover/Popover.tsx
```diff
@@ -160,14 +175,37 @@ export const Popover = forwardRef<HTMLDivElement, PopoverComponentProps>(
...
+    }, [context.open, renderMode]);
```
TheSonOfThomp, line 202: It looks like each time the toggle button is clicked, the Popover fully unmounts and re-mounts. This first effect is fired on every toggle, but the second is never fired

packages/popover/src/Popover.stories.tsx
```diff
@@ -111,23 +118,28 @@ const meta: StoryMetaType<typeof Popover> = {
...
           >
-            <Button>
+            <Button onClick={() => setActive(active => !active)}>
               Button Text
               <Instance
+                buttonText={undefined}
                 className={css`
                   background-color: ${color.light.background.primary.default};
                 `}
+                dismissMode="manual"
```
TheSonOfThomp, line 142: I don't remember what the answer was last time: does this need to be _inside_ the Button?

packages/popover/src/Popover.stories.tsx
```diff
@@ -111,23 +118,28 @@ const meta: StoryMetaType<typeof Popover> = {
...
               />
```
TheSonOfThomp, line 143: this should be an `arg`

packages/popover/src/Popover.stories.tsx
```diff
@@ -111,23 +118,28 @@ const meta: StoryMetaType<typeof Popover> = {
       },
       // eslint-disable-next-line react/display-name
       decorator: Instance => {
+        // eslint-disable-next-line react-hooks/rules-of-hooks
+        const [active, setActive] = useState<boolean>(false);
```
TheSonOfThomp, line 122: how is this working?

TheSonOfThomp: Oh I see this is just the generated decorator?
can we do 
```
const [_, setActive] = useState<boolean>(false);
```

or just cast a void function as a click handler to Button?

packages/popover/src/Popover.stories.tsx
```diff
@@ -177,21 +189,42 @@ export const LiveExample: StoryFn<PopoverStoryProps> = ({
...
       >
         {buttonText}
-        <Popover {...args} active={active}>
-          <div className={popoverStyle}>Popover content</div>
-        </Popover>
       </Button>
+      <Popover
+        {...rest}
+        active={active}
+        renderMode={RenderMode.TopLayer}
```
TheSonOfThomp, line 222: technically we should probably open up the `renderMode` as a control in the LiveExample

packages/popover/src/Popover.stories.tsx
```diff
@@ -235,16 +267,61 @@ export const ScrollableContainer: StoryFn<PopoverStoryProps> = ({
...
 };
-ScrollableContainer.args = {
-  usePortal: true,
+RenderModePortalInScrollableContainer.argTypes = {
+  renderMode: { control: 'none' },
+  portalClassName: { control: 'none' },
+  refEl: { control: 'none' },
+  className: { control: 'none' },
+  active: { control: 'none' },
+};
```
TheSonOfThomp, line 281: nit, could probably move to StoryObj for the rest of these stories as well

packages/popover/src/Popover/Popover.tsx
TheSonOfThomp: (See below comment)

tsck (commented):
I'm seeing some behavior that seems strange with `Select` in Storybook where the popover is rendering like in the corner of the page. Does this have anything to do with this? https://642700aa3e49e32bdbf0b0fc-qztpposnja.chromatic.com/?path=/story/components-select--live-example

Screenshot 2024-10-03 at 3 14 57 PM

tsck (approved):
LeLGTM

PR #2501, opened 2024-10-10
LG-4526: refactor popover-consuming components

stephl3:
## ✍️ Proposed changes

- Replaces `usePortal` prop with `renderMode`
- Refactors `Popover`-consuming components

🎟 _Jira ticket:_ [LG-4526](https://jira.mongodb.org/browse/LG-4526)

### Terrence section
- [x] Review `Combobox` live example
  - [x] Try different `renderMode` settings
  - [x] Toggle on `multiselect`, select long text options, and hover chips
- [x] Review `GuideCue` live example
  - [x] Review `Multistep Demo` story
- [x] Review `Menu` live example
  - [ ] Try different `renderMode` settings
- [ ] Review `SplitButton` live example
  - [ ] Try different `renderMode` settings
- [x] Review `SearchInput` live example

### Adam section
- [x] Review `Select` live example
  - [x] Try different `renderMode` settings
- [x] Review `Code` live example
  - [x] Interact with `CopyButton`
  - [x] Review `With Language Switcher` story and interact with `LanguageSwitcher`
- [x] Review `DatePicker` live example
  - [x] Interact with opening/closing menu
  - [x] Interact with month/year dropdowns
- [x] Review `NumberInput` live example
  - [x] Interact with `UnitSelect`
  - [x] Select "Astronomical units" option and hover `UnitSelectButton`
- [x] Review `Pagination` live example
  - [x] Review `With Current Page Options` story

### Shaneeza section
- [x] Review `Tooltip` live example
  - [x] Try different `renderMode` settings
- [x] Review `Copyable` live example
- [x] Review `InfoSprinkle` live example
- [x] Review `InlineDefinition` live example
- [x] Review `Pipeline` live example
  - [x] Need to review on Large mobile breakpoint
- [x] Review `SideNav` live example
  - [x] Hover `CollapseToggle`
- [x] Review `Stepper` live example
  - [x] Hover `EllipsesStep`

### DIFF
--- a/packages/select/src/Select.stories.tsx
+++ b/packages/select/src/Select.stories.tsx
@@ -65,6 +65,19 @@ const childrenArray = [
 const meta: StoryMetaType<typeof Select> = {
   title: 'Components/Select',
   component: Select,
+  decorators: [
+    (StoryFn, _ctx) => (
+      <LeafyGreenProvider darkMode={_ctx?.args?.darkMode}>
+        <div
+          className={css`
+            height: 100vh;
+          `}
+        >
+          <StoryFn />
+        </div>
+      </LeafyGreenProvider>
+    ),
+  ],
   parameters: {
     default: 'LiveExample',
     controls: {
--- a/packages/menu/src/Menu.stories.tsx
+++ b/packages/menu/src/Menu.stories.tsx
@@ -47,7 +47,13 @@ export default {
   decorators: [
     (StoryFn, _ctx) => (
       <LeafyGreenProvider darkMode={_ctx?.args?.darkMode}>
-        <StoryFn />
+        <div
+          className={css`
+            height: 100vh;
+          `}
+        >
+          <StoryFn />
+        </div>
       </LeafyGreenProvider>
     ),
   ],


packages/date-picker/src/DatePicker/DatePicker.spec.tsx
```diff
@@ -408,27 +408,6 @@ describe('packages/date-picker', () => {
         expect(menuContainerEl).toBeInTheDocument();
       });
 
-      test('appends to the end of the DOM', async () => {
```
TheSonOfThomp, line 411: was going to comment that maybe we want to test something else instead, but that would just be re-testing portal functionality. I think this is correct, and we don't want to re-test that here.

TheSonOfThomp: OTOH, it may be useful to test that a component extends Popover correctly (e.g. a custom expect? `expect(X).toExtendPopover()`)?

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenu.types.ts
```diff
@@ -3,6 +3,14 @@ import { PopoverProps } from '@leafygreen-ui/popover';
...
-  'children' | 'renderMode' | 'usePortal'
+  | 'children'
+  | 'dismissMode'
+  | 'onToggle'
+  | 'popoverZIndex'
+  | 'portalClassName'
+  | 'portalContainer'
+  | 'portalRef'
+  | 'renderMode'
+  | 'scrollContainer'
```
TheSonOfThomp, line 14: Could we simplify this as an exported type from Popover?

packages/date-picker/src/shared/components/MenuWrapper/MenuWrapper.tsx
```diff
@@ -28,7 +39,8 @@ export const MenuWrapper = forwardRef<HTMLDivElement, MenuWrapperProps>(
         ref={fwdRef}
         className={cx(menuStyles[theme], className)}
         {...props}
-        usePortal
+        dismissMode={DismissMode.Manual}
+        renderMode={RenderMode.TopLayer}
```
TheSonOfThomp, line 43: Aren't these the defaults?

TheSonOfThomp:
I feel like all the transitions are too fast

packages/number-input/src/NumberInput.stories.tsx
```diff
@@ -33,6 +33,17 @@ const unitOptions = [
 const meta: StoryMetaType<typeof NumberInput> = {
   title: 'Components/NumberInput',
   component: NumberInput,
+  decorators: [
+    StoryFn => (
+      <div
+        className={css`
+          height: 100vh;
```
TheSonOfThomp, line 40: don't we already have this in the default decorator?

packages/number-input/src/NumberInput/NumberInput.spec.tsx
```diff
@@ -418,54 +403,6 @@ describe('packages/number-input', () => {
         id="1"
         size={Size.Default}
       />
-
-      {/* @ts-expect-error - portalClassName should be undefined */}
```
TheSonOfThomp, line 422: Are there other TS tests we should add for the new/updated props?

packages/select/src/ListMenu/ListMenu.types.ts
```diff
@@ -1,11 +1,8 @@
 import { HTMLElementProps } from '@leafygreen-ui/lib';
-import { PopoverProps } from '@leafygreen-ui/popover';
 
 import { DropdownWidthBasis } from '../Select/Select.types';
 
-export interface ListMenuProps
-  extends HTMLElementProps<'ul', HTMLLIElement>,
-    Omit<PopoverProps, 'active' | 'refEl' | 'onClick'> {
+export interface ListMenuProps extends HTMLElementProps<'ul', HTMLLIElement> {
```
TheSonOfThomp, line 5: Should be `HTMLULElement` (mistake from previous).

Probably cleaner to use `ComponentPropsWithoutRef<'ul'>` though

packages/select/src/Select/Select.spec.tsx
```diff
@@ -696,7 +697,7 @@ describe('packages/select', () => {
 
           await waitForElementToBeRemoved(listbox);
 
-          expect(getByTextFor(button, optionText)).toBeVisible();
```
TheSonOfThomp, line 699: This was implicitly testing that the button had the text value of `optionText`. I think we should still include that assertion here

packages/select/src/Select/Select.spec.tsx
```diff
@@ -961,160 +962,163 @@ describe('packages/select', () => {
...
-    });
-
-    describe('fires onChange when selecting an option', () => {
-      test('on mouse click', async () => {
-        const onChange = jest.fn();
 
-        const { getInput, getOptionByValue } = renderSelect({
-          usePortal: false,
-          onChange: onChange,
+      test('menu renders as a child of button', async () => {
```
TheSonOfThomp, line 975: Confused me for a sec (forgot TopLayer was still a child). This is correct 👍

packages/select/src/Select/index.ts
```diff
@@ -1,6 +1,8 @@
 export { Select } from './Select';
```
TheSonOfThomp, line 1: Not sure what's causing this (might be related to what we chatted about a few weeks ago), but there seems to be a slight delay before the menu opens https://jmp.sh/XXm7a3hF

packages/select/src/Select/Select.types.ts
```diff
@@ -35,7 +37,7 @@ export interface BaseSelectProps
       HTMLElementProps<'button', HTMLButtonElement>,
       'onChange' | 'onClick'
     >,
-    Omit<PopoverProps, 'active' | 'spacing'>,
+    Omit<PopoverProps, 'active' | 'dismissMode' | 'onToggle' | 'spacing'>,
```
TheSonOfThomp, line 40: Not really sure what's causing the duplicated docs in SB. (the VSCode TSDoc highlights docs are fine, so not super pressing)
Screenshot 2024-10-14 at 1 20 13 PM

packages/code/src/CopyButton/CopyButton.tsx
```diff
@@ -109,11 +113,13 @@ function CopyButton({ onCopy, contents }: CopyProps) {
 
   return (
     <Tooltip
```
TheSonOfThomp, line 115: It seems the 100ms delay on the Tooltip hover is no longer applied. (Same for the NumberInput tooltips)

packages/code/src/LanguageSwitcher/LanguageSwitcher.tsx
```diff
@@ -100,7 +83,7 @@ function LanguageSwitcher({
         value={language?.displayName}
         className={selectStyle}
         allowDeselect={false}
-        {...popoverProps}
+        renderMode={RenderMode.TopLayer}
```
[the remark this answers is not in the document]
TheSonOfThomp: Can we add some migration notes for this in case there are instances where it breaks?

packages/code/src/LanguageSwitcher/LanguageSwitcher.tsx
```diff
@@ -28,22 +28,13 @@ function isLeafyGreenIcon(element: React.ReactNode) {
...
-  language,
-  languageOptions,
-  onChange,
-  usePortal,
-  portalClassName,
-  portalContainer,
-  scrollContainer,
-  popoverZIndex,
-}: Props) {
+function LanguageSwitcher({ language, languageOptions, onChange }: Props) {
```
TheSonOfThomp, line 37: Haven't checked whether this is new or now, but I noticed that hovering over the menu would also apply the hover styles to the trigger 
Screenshot 2024-10-14 at 1 27 33 PM

packages/popover/README.md
```diff
@@ -90,16 +90,15 @@ In pre-v12 versions, a popover can be rendered in 2 ways using the `usePortal` p
...
+| `portalRef` (deprecated)       | `string`                                  | Passes a ref to forward to the portal element. This will only apply when `usePortal` is `true`                                                                                                                                                                                                                                                                                        |               |
+| `scrollContainer` (deprecated) | `HTMLElement` \| `null`                   | If the popover portal has a scrollable ancestor other than the window, this prop allows passing a reference to that element to allow the portal to position properly. This will only apply when `usePortal` is `true`                                                                                                                                                                 |               |
```
TheSonOfThomp, line 100: Update the last sentence in each of these to "This will only apply when `renderMode` is `portal`"

packages/popover/src/utils/getPopoverRenderModeProps.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @internal
+ */
+export const getPopoverRenderModeProps = ({
+  dismissMode,
+  onToggle,
+  portalClassName,
+  portalContainer,
+  portalRef,
+  renderMode,
+  scrollContainer,
```
TheSonOfThomp, line 14: nit, would it make sense to pull out the arg type? Don't think we have guidelines, but imo if the arg object has more than 2-3 properties, I prefer to extract the object into something like `interface GetPopoverRenderModePropsArgs` (preferably less verbose...)

packages/code/src/LanguageSwitcher/LanguageSwitcher.tsx
tsck: This can also be seen on `Menu` 
Screenshot 2024-10-15 at 12 48 05 PM

packages/stepper/src/EllipsesStep/EllipsesStep.tsx
```diff
@@ -20,8 +20,9 @@ export const EllipsesStep = ({
   return (
     <Tooltip
       align={Align.Top}
-      justify={Justify.Middle}
       darkMode={darkMode}
+      justify={Justify.Middle}
+      renderMode={RenderMode.TopLayer}
```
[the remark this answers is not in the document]
shaneeza: ?

packages/copyable/src/Copyable/Copyable.tsx
```diff
@@ -183,7 +187,7 @@ export default function Copyable({
                 </Button>
               }
               triggerEvent={TriggerEvent.Click}
-              usePortal={shouldTooltipUsePortal}
+              renderMode={RenderMode.TopLayer}
```
shaneeza, line 190: It looks like the delay is missing when the tooltip appears

packages/inline-definition/src/InlineDefinition.tsx
```diff
@@ -51,7 +51,19 @@ const triggerElementModeStyles: Record<Theme, string> = {
   `,
 };
 
-export interface InlineDefinitionProps extends Partial<TooltipProps> {
+export interface InlineDefinitionProps
```
shaneeza, line 54: nit: can we move this to a types file?

packages/inline-definition/src/InlineDefinition.tsx
```diff
@@ -89,10 +101,11 @@ function InlineDefinition({
 
   return (
     <Tooltip
-      justify="middle"
-      spacing={9}
       className={tooltipClassName}
       darkMode={darkMode}
+      justify="middle"
+      renderMode={RenderMode.TopLayer}
```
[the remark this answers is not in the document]
shaneeza: nit: I'm curious about the GitHub search. Could there be private repos that we don't have access to?

packages/inline-definition/src/InlineDefinition.stories.tsx
```diff
@@ -81,7 +82,11 @@ export const LiveExample: StoryFn<InlineDefinitionProps> = ({
     `}
   >
     <H3 darkMode={darkMode}>
-      <InlineDefinition darkMode={darkMode} {...args}>
+      <InlineDefinition
+        darkMode={darkMode}
+        baseFontSize={BaseFontSize.Body1}
```
shaneeza, line 87: nit: why do we need to hardcode this?

packages/guide-cue/src/GuideCue.tsx
shaneeza, line 1: I noticed that when using the keyboard, the focus does not move to the first focusable item in the tooltip and tabbing inside the tooltip does not keep the focus trapped in it.

packages/side-nav/src/CollapseToggle/CollapseToggle.tsx
```diff
@@ -33,10 +33,10 @@ export function CollapseToggle({
 
   return (
     <Tooltip
-      usePortal={false}
-      align="right"
-      justify="middle"
+      align={Align.Right}
+      justify={Justify.Middle}
       open={typeof hideTooltip === 'boolean' ? !hideTooltip : undefined}
+      renderMode={RenderMode.TopLayer}
```
shaneeza, line 39: In the `In Layout` story, when toggling the toggle with the enter key, the tooltip sometimes appears in the wrong position and then returns to the correct spot after a second.

shaneeza:
I think that the tooltips are opening really fast compared to the current implementation

packages/tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -143,11 +146,17 @@ function Tooltip({
               // Without this the tooltip sometimes opens without a transition. flushSync prevents this state update from automatically batching. Instead updates are made synchronously.
               // https://react.dev/reference/react-dom/flushSync#flushing-updates-for-third-party-integrations
               flushSync(() => {
```
shaneeza, line 148: Do we still need `flushSync`?

packages/tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -111,6 +113,7 @@ function Tooltip({
   const setOpen =
     isControlled && controlledSetOpen ? controlledSetOpen : uncontrolledSetOpen;
 
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
```
shaneeza, line 116: Should we also clear this in a useEffect?

packages/tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -88,7 +90,7 @@ function Tooltip({
   align = 'top',
   justify = 'start',
   spacing = 12,
-  usePortal = true,
+  renderMode = RenderMode.TopLayer,
```
shaneeza, line 93: When using `topLayer` or `inline`, the tooltip inherits the styles of the parents. For example, the tooltip is in uppercase in the story file because the button is uppercase, but it probably should not be uppercase.

shaneeza: I wonder if this is more of a new consumer responsibility since the tooltip is now a child of the trigger.

packages/side-nav/src/CollapseToggle/CollapseToggle.tsx
shaneeza: I can't replicate this anymore. I'll mark it as resolved!

packages/tooltip/src/Tooltip/Tooltip.tsx
shaneeza: flushSync is to support react18. Sometimes, tooltip transitions would not appear, so I was curious if the new popover changes fix that issue. But it's understandable if you don't want to touch it right now.

packages/tooltip/src/Tooltip/Tooltip.tsx
shaneeza: I think we should ask design because i'm not sure.

packages/tooltip/src/Tooltip/Tooltip.tsx
shaneeza: I guess in rare cases where a user hovers over the trigger and the component unmounts before the user can move the mouse away from the trigger.

packages/tooltip/src/Tooltip/Tooltip.tsx
shaneeza: makes sense

packages/menu/src/Menu.spec.tsx
```diff
@@ -234,9 +235,9 @@ describe('packages/menu', () => {
         expect(menuEl).not.toBeInTheDocument();
       });
 
-      test('Returns focus to trigger {usePortal: true}', async () => {
+      test(`Returns focus to trigger when renderMode=${RenderMode.TopLayer}`, async () => {
         const { openMenu, triggerEl } = renderMenu({
-          usePortal: true,
+          renderMode: RenderMode.TopLayer,
```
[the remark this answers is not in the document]
tsck: Yup, probably just a build issue on my end then!

packages/menu/src/Menu.spec.tsx
```diff
@@ -291,7 +307,7 @@ describe('packages/menu', () => {
 
             const { menuItemElements } = await openMenu();
             expect(menuItemElements).toHaveLength(3);
-            expect(queryByTestId('submenu')).toHaveFocus();
+            await waitFor(() => expect(queryByTestId('submenu')).toHaveFocus());
```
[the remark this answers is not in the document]
tsck: All good, just wanted to clarify!

packages/select/src/Select.stories.tsx
```diff
@@ -96,7 +97,7 @@ const meta: StoryMetaType<typeof Select> = {
     allowDeselect: false,
     darkMode: false,
     children: childrenArray,
-    usePortal: true,
+    renderMode: RenderMode.TopLayer,
```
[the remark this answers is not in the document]
tsck: Is the limitation that inline will be cut off by its container? I wonder if users will actually understand that that's what we're trying to communicate here and not just think that we broke storybook. Perhaps there's a way we can communicate this more explicitly? Not sure if there's a good way to do this.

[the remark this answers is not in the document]
tsck: Makes sense, but agree that we should probably fix it, since I don't think it leaving it broken quite communicates what we want it to here.

tsck (approved):

shaneeza (approved):

packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenu.types.ts
TheSonOfThomp: I feel like the _ideal_ case would be for everything to expose the same props (for API consistency). For now while we're going for parity, this is fine

TheSonOfThomp (approved):
There still seems to be a delay in opening the Menu/Select etc. but I don't think that should hold up this PR

PR #2510, opened 2024-10-17
LG-4574: Add base chart components

tsck:
## ✍️ Proposed changes

Adds the `core` packages to `@lg-charts`. These export core chart components, `Chart`, `Line`, and `Grid` to provide basic support of rendering lines onto a chart. All other functionality (axis, tooltips, etc.) will be added iteratively in separate PR's.

🎟 _Jira ticket:_ [LG-4574](https://jira.mongodb.org/browse/LG-4574)

### For new components

## 🧪 How to test changes
- `LineChart` story added to Storybook


charts/core/src/Chart/Chart.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+    addSeries: jest.fn(),
+    removeSeries: jest.fn(),
+  })),
+}));
+
+jest.mock('@leafygreen-ui/leafygreen-provider', () => ({
+  useDarkMode: jest.fn(() => ({ theme: 'light' })),
+}));
+
+describe('lg-charts/core/Chart', () => {
```
tsck, line 27: Wasn't sure what else could/should be tested for this component without falling too heavily into implementation detail testing. I leaned towards relying more on visual regression and chromatic tests here, but open to suggestions.

charts/core/src/Chart/Chart.types.ts
```diff
@@ -0,0 +1,65 @@
...
+export type ChartOptions = ComposeOption<
+  | TooltipComponentOption
+  | GridComponentOption
+  | DatasetComponentOption
+  | TitleComponentOption
+  | LegendComponentOption
+  | ToolboxComponentOption
+  | XAXisComponentOption
+  | YAXisComponentOption
+> & { series?: Array<SeriesOption> };
```
tsck, line 34: Essentially, I think this might need to be made more granular or custom depending on specific what we want to support. But this is essentially how they recommend producing an options object type

charts/core/src/Line/Line.tsx
```diff
@@ -0,0 +1,22 @@
...
+      ...defaultLineOptions,
+      name,
+      data,
+    });
+  }, [name, data]);
+
+  return null;
+}
+
+Line.displayName = 'Line';
```
tsck, line 22: I didn't implement any unit tests for this component. There's no actual logic or rendering being done so beyond checking implementation details, that seemed correct. I'm very open to other ideas here though since this is probably the pattern we'll see in most of the core components.

charts/line-chart/src/LineChart.stories.tsx
```diff
@@ -2,16 +2,29 @@ import React from 'react';
...
   title: 'Charts/LineChart',
   component: LineChart,
   argTypes: {
     darkMode: storybookArgTypes.darkMode,
   },
+  decorator: (Instance, context) => {
+    return (
+      <LeafyGreenProvider darkMode={context?.args.darkMode}>
+        <Instance darkMode={context?.args.darkMode} />
+      </LeafyGreenProvider>
```
tsck, line 20: This seemed possibly incorrect, using the provider and passing `darkMode` directly. Saw various patterns here so added the provider, but would love some feedback on this

charts/line-chart/src/LineChart.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Chart, Line } from '@lg-charts/core';
+
+import { LineChartProps } from './LineChart.types';
+
+export function LineChart({ data }: LineChartProps) {
+  return (
+    <Chart>
+      {data.map(({ name, data }) => (
+        <Line name={name} data={data} key={name} />
```
tsck, line 10: Names are currently not technically guaranteed to be unique. Could concat with the index, but seeing that index usage could also be problematic, just went with name since they probably _should_ be unique, given that they ID a given line on a chart. Curious if there's a specific pattern we typically follow

tsck:
Note to reviewer: Would love some special attention on non-logical patterns as well - comments, typing, etc. Tried to match what I saw in places, but definitely imagine some changes will be needed in this domain.

charts/line-chart/src/LineChart.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Chart, Line } from '@lg-charts/core';
+
+import { LineChartProps } from './LineChart.types';
+
+export function LineChart({ data }: LineChartProps) {
+  return (
+    <Chart>
```
tsck, line 8: Forgot to allow this to be re-styled/sized by the consumer. Will update this shortly

tsck: Done

tsck:
^ Noticed I messed up some nesting on the README's. Updated those. Sorry for the post-request changes!

tsck:
Additional note: Going to be updating the README files significantly in forthcoming PR's

charts/line-chart/src/LineChart.types.tsx
```diff
@@ -0,0 +1,12 @@
+import { SeriesOption } from '@lg-charts/core';
+
+import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
+
+export interface Series extends SeriesOption {
+  name: string;
+  data: Array<[Date, number]>;
+}
```
tsck, line 8: Just use SeriesOption

charts/core/src/Chart/Chart.types.ts
```diff
@@ -0,0 +1,65 @@
...
+  TitleComponentOption,
+  ToolboxComponentOption,
+  TooltipComponentOption,
+} from 'echarts/components';
+import type { ComposeOption } from 'echarts/core';
+
+import { DarkModeProps, type HTMLElementProps } from '@leafygreen-ui/lib';
+
+type RequiredSeriesProps = 'type' | 'name' | 'data';
+export type SeriesOption = Pick<LineSeriesOption, RequiredSeriesProps> &
```
tsck, line 16: Use PartialRequire

charts/line-chart/src/LineChart.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Chart, Line } from '@lg-charts/core';
+
+import { LineChartProps } from './LineChart.types';
+
+export function LineChart({ data, ...rest }: LineChartProps) {
```
tsck, line 6: This probably shouldn't exist yet. It make more sense to just expose the composable pieces. This may be something to do at a later date.

charts/core/src/Chart/Chart.tsx
```diff
@@ -0,0 +1,57 @@
...
+    removeChartSeries,
+    chartRef,
+  } = useChart({
+    darkMode: darkModeProp,
+    onChartReady,
+  });
+  const { theme } = useDarkMode(darkModeProp);
+
+  return (
+    <ChartProvider
```
tsck, line 38: Wrap in darkModeProvider

charts/core/src/Chart/Chart.tsx
```diff
@@ -0,0 +1,57 @@
...
+      chartOptions={chartOptions}
+      updateChartOptions={updateChartOptions}
+      addChartSeries={addChartSeries}
+      removeChartSeries={removeChartSeries}
+      darkMode={darkModeProp}
```
tsck, line 43: Remove after provider

charts/core/src/ChartContext/ChartContext.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { ChartOptions, SeriesOption } from '../Chart/Chart.types';
+
+export interface ChartContextType extends DarkModeProps {
+  chartOptions: any;
+  updateChartOptions: (newOptions: Partial<ChartOptions>) => void;
+  addChartSeries: (series: SeriesOption) => void;
+}
+
+export const ChartContext = createContext<ChartContextType | undefined>(
```
tsck, line 13: Include provider and hook in this file

tsck (commented):
Feedback from group CR

charts/core/src/Chart/Chart.types.ts
tsck: Actually, leaving this for now. `PartialRequire<LineSeriesOption, RequiredSeriesProps>` doesn't seem to be typing this quite correctly, and I'm receiving errors when trying to access anything other than `type` of a `SeriesOption`.

bruugey (approved):

charts/core/package.json
```diff
@@ -16,13 +16,21 @@
...
     "@leafygreen-ui/emotion": "^4.0.7",
-    "@leafygreen-ui/lib": "^12.0.0",
-    "@leafygreen-ui/palette": "^4.1.1",
+    "@leafygreen-ui/lib": "13.7.0",
     "@leafygreen-ui/tokens": "^2.11.0",
     "@lg-tools/storybook-utils": "^0.1.1",
-    "echarts": "^5.5.1"
+    "echarts": "^5.5.1",
+    "lodash.debounce": "^4.0.8",
+    "lodash.merge": "^4.6.2"
```
TheSonOfThomp, line 24: Oh I like this a lot more than a blanket lodash import

charts/core/src/Chart/hooks/useChart.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+  }, []);
+
+  const updateChartRef = debounce((chartOptions: Partial<ChartOptions>) => {
+    chartInstanceRef.current?.setOption(chartOptions);
+  }, 50);
+
+  const addChartSeries = (data: SeriesOption) => {
+    setChartOptions(currentOptions => {
+      const updatedOptions = addSeries(currentOptions, data);
```
TheSonOfThomp, line 84: I still think using a reducer would be a bit better, but not significant enough to change this

charts/core/src/Chart/hooks/useChart.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    return () => {
+      window.removeEventListener('resize', resizeHandler);
+      chartInstance.dispose();
+    };
+  }, []);
+
+  const updateChartRef = debounce((chartOptions: Partial<ChartOptions>) => {
+    chartInstanceRef.current?.setOption(chartOptions);
+  }, 50);
```
TheSonOfThomp, line 80: Still not convinced we need the debounce. I feel like there's a way around this

charts/core/src/ChartContext/ChartContext.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const ChartContext = createContext<ChartContextType | undefined>(
+  undefined,
+);
+
+export const ChartProvider = ({
+  children,
+  chartOptions,
+  updateChartOptions,
+  addChartSeries,
+}: PropsWithChildren<ChartProviderProps>) => {
```
TheSonOfThomp, line 14: The type file includes `removeChartSeries` but we don't use that here

charts/core/src/Grid/Grid.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { useGridOptions } from './config';
+import { GridProps } from './Grid.types';
+
+export function Grid({ horizontal = true, vertical = true }: GridProps) {
+  const { updateChartOptions } = useChartContext();
+  const gridOptions = useGridOptions({ horizontal, vertical });
+
+  useEffect(() => {
+    updateChartOptions(gridOptions);
+  }, [gridOptions]);
```
TheSonOfThomp, line 14: We're listening to a state that we control here. That's a signal that we're using `useEffect` wrong.

I'd recommend either updating the chart options from within `useGridOptions`, or just move the content of `useGridOptions` into this component

charts/core/src/Grid/config/useGridOptions.ts
```diff
@@ -0,0 +1,32 @@
+import { useEffect, useState } from 'react';
+
+import { useDarkMode } from '@leafygreen-ui/leafygreen-provider';
+import { color, InteractionState, Variant } from '@leafygreen-ui/tokens';
+
+import { ChartOptions } from '../../Chart';
+import { GridProps } from '../Grid.types';
+
+export const useGridOptions = ({ horizontal, vertical }: GridProps) => {
+  const [options, setOptions] = useState<Partial<ChartOptions>>({});
```
TheSonOfThomp, line 10: If we moved the `updateChartOptions` call into this hook (or the inverse) I don't think we'd need this state object.

charts/core/src/Line/Line.types.ts
```diff
@@ -0,0 +1,25 @@
...
+   *     //  X      Y
+   *     [  3.4,   4.5 ],
+   *     [  4.2,   2.3 ],
+   *     [  10.8,  9.5 ],
+   *     [  7.2,   8.8 ]
+   *   ]
+   * }]
+   * ```
+   */
+  data: Array<[string | number | Date, string | number | Date]>;
```
TheSonOfThomp, line 24: Could we extract this tuple into it's own type?

TheSonOfThomp: Also, X and Y don't need to be the same type, right?

charts/core/src/Line/Line.tsx
TheSonOfThomp: I feel like testing that `useChartContext` works as expected, and then a solid suite of Chromatic tests will be good for this package

charts/core/src/utils/makeLineData.testUtils.ts
TheSonOfThomp, line 1: I think it's helpful to separate the `utils` directory from `testUtils`. That way it's easier to catch (IMO) if you accidentally import test utils in a source file

charts/core/src/Chart/Chart.styles.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Theme } from '@leafygreen-ui/lib';
+import {
+  borderRadius,
+  color,
+  InteractionState,
+  Variant,
+} from '@leafygreen-ui/tokens';
+
+export const chartStyles = css`
+  grid-row: 2;
```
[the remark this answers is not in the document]
tsck: Completely agree! Right now this isn't really doing much since the header doesn't exist yet. But I'll switch this to `grid-area` when we add in the header.

charts/core/src/Chart/Chart.styles.ts
```diff
@@ -0,0 +1,22 @@
...
+`;
+
+export const getWrapperStyles = (theme: Theme) => css`
+  height: 280px;
+  width: 100%;
+  border: 1px solid
+    ${color[theme].border[Variant.Disabled][InteractionState.Default]};
```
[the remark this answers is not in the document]
tsck: Good catch, this should be `Variant.Secondary` will update

charts/core/src/Chart/Chart.types.ts
```diff
@@ -0,0 +1,65 @@
...
+  type: typeof ChartActionType.removeChartSeries;
+  name: string;
+}
+
+interface UpdateOptionsAction {
+  type: typeof ChartActionType.updateOptions;
+  options: Partial<Omit<ChartOptions, 'series'>>;
+}
+
+export type ChartAction =
```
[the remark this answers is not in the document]
tsck: Oh this is leftover from when I made this a reducer previously. I'll remove these for now

charts/core/src/Chart/config/getDefaultChartOptions.ts
```diff
@@ -0,0 +1,75 @@
...
+  spacing,
+  Variant,
+} from '@leafygreen-ui/tokens';
+
+import { ChartOptions } from '../Chart.types';
+import { colors } from '../colors';
+
+const commonAxisOptions = {
+  /**
+   * Axis need to be hidden explicitly to prevent error when rendering chart without Axis
```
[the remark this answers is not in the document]
tsck: Axes need 😅 fixing!

charts/core/src/ChartContext/ChartContext.tsx
tsck: Good catch! This isn't really doing anything yet but might as well pass it down.

I've also updated types to just use `ChartContextType` since really the provider and context should have the same types.

charts/core/src/Grid/Grid.tsx
tsck: I like the idea of just getting rid of the `useGridOptions` hook entirely. With this component's sole responsibility being to update options, I don't think it will bloat the component in any way, and brings the logic closer to the component. Updating!

charts/core/src/Line/Line.types.ts
tsck: Nope they don't. I've created `XValue` and `YValue`, let me know if this works better? Could also create a 3rd type for the tuple if that makes more sense.

charts/core/src/Chart/hooks/useChart.ts
tsck: Definitely open to ideas here. Just running a small test, I added a log in this method to see how many updates run on a chart with 100 lines, with and without debouncing the update call. 

With `debounce` it just runs once and there is 1 `onChartReady` action triggered.
Screenshot 2024-10-29 at 10 07 09 AM
Screenshot 2024-10-29 at 10 07 15 AM

Without `debounce` however, due to re-renders, the chart is update 400+ times and triggering 400+ `onChartReady` calls as well:
Screenshot 2024-10-29 at 10 06 26 AM
Screenshot 2024-10-29 at 10 06 40 AM

So if we didn't want to debounce the call, I still think something is required here to fix this.

charts/line-chart/src/LineChart.stories.tsx
TheSonOfThomp: Generally you shouldn't need to do this if whatever `Instance` component reads from context correctly

TheSonOfThomp: (There is a case for doing something similar if you want to test how the `Instance` responds to the provider _and_ to an override prop. You'd pass different variables in this case then)

charts/line-chart/src/LineChart.tsx
TheSonOfThomp: If echarts expects to identify a series based on the `name` then I'd say that'd good enough for our React keys. Maybe add a comment here explaining that? 
I agree using `index` isn't great, because that could lead to bugs if lines get reordered.

TheSonOfThomp: Probably not worth the computation, but we could use some checksum of the `data` value?

charts/core/src/Chart/Chart.types.ts
TheSonOfThomp: This _should_ work. Not sure why it's not

charts/line-chart/src/LineChart.tsx
tsck: This actually no longer exists since we're no longer exposing the higher level component!

charts/core/src/core.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+  },
+};
+
+interface StoryChartProps {
+  data: Array<LineProps>;
+  verticalGridLines: boolean;
+  horizontalGridLines: boolean;
+}
+
+const Template: React.FC<StoryChartProps> = props => {
```
TheSonOfThomp, line 41: Was `StoryFn` not working here?

charts/core/src/core.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+    </Chart>
+  );
+};
+
+export const LiveExample: StoryFn<StoryChartProps> = Template.bind({});
+LiveExample.args = {
+  data: makeLineData(10),
+  horizontalGridLines: true,
+  verticalGridLines: false,
+};
```
TheSonOfThomp, line 58: If you only have one story, you don't necessarily need to declare template and to all this

TheSonOfThomp (approved):

charts/core/src/core.stories.tsx
tsck: `StoryFn<StoryChartProps>` wasn't working for some reason, it was leading to the following linting error:
```
/Users/terrence/Code/leafygreen-ui/charts/core/src/core.stories.tsx
  42:11  error  'data' is missing in props validation                 react/prop-types
  42:17  error  'verticalGridLines' is missing in props validation    react/prop-types
  42:36  error  'horizontalGridLines' is missing in props validation  react/prop-types
  46:13  error  'data.map' is missing in props validation             react/prop-types
```

PR #2513, opened 2024-10-21
Fix escape key close in `Modal`

stephl3:
## ✍️ Proposed changes

Fixes a bug introduced after merging PR https://github.com/mongodb/leafygreen-ui/pull/2456. Consolidating the `PopoverContext` and `PortalContext` made it so that when a `PopoverProvider` was added to replace prop-drilling `Popover` component props, the `isPopoverOpen` was getting reset to `false` when opened within a `Modal`. This PR fixes this by breaking out the `isPopoverOpen`/`setIsPopoverOpen` values into a separate `ModalPopoverProvider`.

Also, moves the `ModalPopoverContext` from `@leafygreen-ui/leafygreen-provider` to `@leafygreen-ui/modal` and updates additional dev dependency relationships to prevent circular dependencies.

### For new components

## 🧪 How to test changes

- [ ] Review `DatePicker/DatePicker/InModal` story

### DIFF
--- a/packages/leafygreen-provider/src/LeafyGreenContext.tsx
+++ b/packages/leafygreen-provider/src/LeafyGreenContext.tsx
@@ -6,6 +6,11 @@ import { DarkModeProps } from '@leafygreen-ui/lib';
 import { RenderMode } from './PopoverContext/PopoverContext.types';
 import DarkModeProvider, { useDarkModeContext } from './DarkModeContext';
 import {
+  MigrationContextType,
+  MigrationProvider,
+  useMigrationContext,
+} from './MigrationContext';
+import {
   PopoverProvider,
   PopoverProviderProps,
   usePopoverContext,
@@ -27,13 +32,15 @@ export type LeafyGreenProviderProps = {
    */
   popoverPortalContainer?: PopoverPortalContainerType;
 } & TypographyProviderProps &
-  DarkModeProps;
+  DarkModeProps &
+  MigrationContextType;
 
 function LeafyGreenProvider({
   children,
   baseFontSize: fontSizeProp,
   popoverPortalContainer: popoverPortalContainerProp,
   darkMode: darkModeProp,
+  forceUseTopLayer: forceUseTopLayerProp = false,
 }: PropsWithChildren<LeafyGreenProviderProps>) {
   /**
    * If `darkMode` prop is provided, use that. Otherwise, use context value
@@ -52,6 +59,13 @@ function LeafyGreenProvider({
    */
   const inheritedFontSize = useBaseFontSize();
   const baseFontSize = fontSizeProp ?? inheritedFontSize;
+
+  /**
+   * If `forceUseTopLayerProp` is true, it will globally apply to all children
+   */
+  const migrationContext = useMigrationContext();
+  const forceUseTopLayer =
+    forceUseTopLayerProp || migrationContext.forceUseTopLayer;
 
   /**
    * If `popoverPortalContainer` prop is provided, use that. Otherwise, use context value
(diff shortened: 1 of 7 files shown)


packages/leafygreen-provider/src/ModalPopoverContext/ModalPopoverContext.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  test('when passed true, `setIsPopoverOpen` sets `isPopoverOpen` to true', () => {
+    const { testChild, getByTestId } = renderProvider();
+
+    // The button's click handler fires setIsPopoverOpen(true)
+    userEvent.click(getByTestId(buttonTestId));
+
+    expect(testChild.textContent).toBe('true');
+  });
```
TheSonOfThomp, line 60: Technically we don't need these tests, since we test `useModalPopoverContext` explicitly below

packages/leafygreen-provider/src/ModalPopoverContext/ModalPopoverContext.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+  test('`setIsPopoverOpen` updates the value of `isPopoverOpen`', async () => {
+    const { result, rerender } = renderHook(useModalPopoverContext, {
+      wrapper: ({ children }) => (
+        <ModalPopoverProvider>{children}</ModalPopoverProvider>
+      ),
+    });
+
+    act(() => result.current.setIsPopoverOpen(true));
+    rerender();
+    await waitFor(() => {
```
TheSonOfThomp, line 78: does this need to be awaited?

packages/leafygreen-provider/src/ModalPopoverContext/ModalPopoverContext.types.ts
```diff
@@ -0,0 +1,16 @@
...
+  isPopoverOpen: boolean;
+
+  /**
+   * Sets the internal state
+   * @internal
+   */
+  setIsPopoverOpen: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+export interface ModalPopoverProviderProps {
```
TheSonOfThomp, line 14: `PropsWithChildren<{}>` ? Could also just inline this type in the Provider component

TheSonOfThomp (commented):
Can we add a changeset here as well? Also note that bumping a `peerDependency` (`leafygreen-provider`) require a major change in the consuming component

packages/date-picker/src/shared/components/MenuWrapper/MenuWrapper.tsx
```diff
@@ -44,11 +42,9 @@ export const MenuWrapper = forwardRef<HTMLDivElement, MenuWrapperProps>(
       >
         {/*
          * Prevents the opening and closing state of a select dropdown from propagating up
-         * to other PopoverProviders in parent components. E.g. Modal
-         *
-         * TODO: https://jira.mongodb.org/browse/LG-4474
+         * to other ModalPopoverProviders in parent components. E.g. Modal
          */}
-        <PopoverProvider>{children}</PopoverProvider>
+        <ModalPopoverProvider>{children}</ModalPopoverProvider>
```
TheSonOfThomp, line 47: should this one be `ModalPopoverProvider`?

packages/leafygreen-provider/src/ModalPopoverContext/ModalPopoverContext.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { createContext, useContext, useMemo, useState } from 'react';
+import PropTypes from 'prop-types';
+
+import {
+  ModalPopoverContextType,
+  ModalPopoverProviderProps,
+} from './ModalPopoverContext.types';
+
+export const ModalPopoverContext = createContext<ModalPopoverContextType>({
```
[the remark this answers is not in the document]
TheSonOfThomp: Ah shoot. I forgot about that. Forcing `Popover` to depend on `modal` isn't great. Let's sync on that

packages/date-picker/src/shared/components/MenuWrapper/MenuWrapper.tsx
TheSonOfThomp: Got it, so this is here in case the date picker is rendered within a modal?

TheSonOfThomp (approved):
I think the ModalPopoverProvider could use some more documentation, but otherwise lgtm

PR #2514, opened 2024-10-22
LG-4525: MigrationProvider with forceUseTopLayer flag

stephl3:
## ✍️ Proposed changes

- adds `MigrationContext` / `MigrationProvider` to globally turn on top layer for all popover-consuming components
- `Popover` components check `forceUseTopLayer` global flag

🎟 _Jira ticket:_ [LG-4525](https://jira.mongodb.org/browse/LG-4525)

### DIFF
--- a/packages/popover/src/Popover/Popover.hooks.tsx
+++ b/packages/popover/src/Popover/Popover.hooks.tsx
@@ -4,7 +4,10 @@ import {
   useIsomorphicLayoutEffect,
   useObjectDependency,
 } from '@leafygreen-ui/hooks';
-import { usePopoverContext } from '@leafygreen-ui/leafygreen-provider';
+import {
+  useMigrationContext,
+  usePopoverContext,
+} from '@leafygreen-ui/leafygreen-provider';
 
 import { getElementDocumentPosition } from '../utils/positionUtils';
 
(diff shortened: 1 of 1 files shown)


packages/popover/src/Popover/Popover.hooks.tsx
```diff
@@ -48,9 +52,13 @@ export function usePopoverContextProps({
...
 >) {
+  const { forceUseTopLayer } = useMigrationContext();
   const context = usePopoverContext();
-  const renderMode =
-    renderModeProp || context.renderMode || RenderMode.TopLayer;
+  const renderMode = getRenderMode({
+    forceUseTopLayer,
+    renderModeProp,
+    renderModeCtxVal: context.renderMode,
+  });
```
tsck, line 61: I'm wondering if this helper method is providing much given that we have to pass in 2 of the 3 possible options anyway. It creates the need to have to jump to another file to see what it's doing and I'm not sure it's more readable than just having the logic inline:
```ts
const renderMode = forceUseTopLayer ? 
  RenderMode.TopLayer : 
  renderModeProp || context.renderMode || RenderMode.TopLayer;
```
Mostly nit, but curious to hear your thoughts

tsck (commented):
Looks great! Mostly LGTM, just one small question.

tsck (approved):
Taylor LGTM

PR #2521, opened 2024-10-25
LG-4630: apply focus styles on highlighted menu items when `:focus-visible`

stephl3:
## ✍️ Proposed changes

- apply focus styles on highlighted menu items when `:focus-visible`

🎟 _Jira ticket:_ [LG-4630](https://jira.mongodb.org/browse/LG-4630)

## 🧪 How to test changes

In `Menu` and `SplitButton` live example:
- [ ] open menu with click and focus styles should not be applied
- [ ] open menu with keyboard navigation and focus styles should be applied

### DIFF
--- a/packages/menu/src/Menu/Menu.tsx
+++ b/packages/menu/src/Menu/Menu.tsx
@@ -233,6 +233,12 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
   if (trigger) {
     const triggerClickHandler = (event?: React.MouseEvent) => {
       event?.preventDefault();
+
+      // If enter or space key is pressed, event detail is 0 https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#usage_notes
+      if (event?.detail === 0) {
+        keyboardUsedRef.current = true;
+      }
+
       setOpen((curr: boolean) => !curr);
 
       if (trigger && typeof trigger !== 'function') {
@@ -245,16 +251,9 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
       event?.nativeEvent?.stopPropagation?.();
     };
 
-    const triggerKeyDownHandler = (event?: React.KeyboardEvent) => {
-      if (event?.key === keyMap.Enter || event?.key === keyMap.Space) {
-        keyboardUsedRef.current = true;
-      }
-    };
-
     if (typeof trigger === 'function') {
       return trigger({
         onClick: triggerClickHandler,
-        onKeyDown: triggerKeyDownHandler,
         ref: triggerRef,
         children: popoverContent,
         ['aria-expanded']: open,
@@ -265,7 +264,6 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
     const renderedTrigger = React.cloneElement(trigger, {
       ref: triggerRef,
       onClick: triggerClickHandler,
-      onKeyDown: triggerKeyDownHandler,
       children: (
         <>
           {trigger.props.children}


packages/input-option/src/InputOption/InputOption.style.ts
```diff
@@ -113,8 +113,10 @@ export const getInputOptionWedge = ({
 
     ${highlighted &&
     css`
-      transform: scaleY(1) translateY(-50%);
-      background-color: ${palette.blue.base};
+      &:focus-visible {
+        transform: scaleY(1) translateY(-50%);
+        background-color: ${palette.blue.base};
+      }
```
TheSonOfThomp, line 119: Looks like this is causing the wedge to disappear in Storybook

packages/menu/src/MenuItem/MenuItem.styles.ts
```diff
@@ -77,9 +77,7 @@ export const getMenuItemStyles = ({
 
       // Highlighted
       [css`
-        &,
-        &:hover,
-        &:focus {
+        &:focus-visible {
```
TheSonOfThomp, line 80: This is also breaking some generated stories. We should add some test data-property selector to allow us to test this in SB
i.e. `[data-focus="true"] {...}`

TheSonOfThomp (changes requested):
Realized why we went with the original approach. If I click the button, then press down 
arrow, it looks like it skips the first item (since it's "focused" but not "focus-visible")

https://jmp.sh/Q0ScRgOU

If we want to go the `focus-visible` route, I think we'll have top not focus the first item by default, and keep focus in the trigger

packages/menu/src/Menu/Menu.tsx
```diff
@@ -241,9 +245,16 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
       event?.nativeEvent?.stopPropagation?.();
     };
 
+    const triggerKeyDownHandler = (event?: React.KeyboardEvent) => {
+      if (event?.key === keyMap.Enter || event?.key === keyMap.Space) {
+        keyboardUsedRef.current = true;
+      }
+    };
```
TheSonOfThomp, line 252: Trying to decide if hooking into `onKeyDown` like this is better than checking the `event.detail`  in the click handler.
In this case it's clear that we're looking for keydowns, but we've now added another event handler.

Alternatively we could set `keyboardUsedRef` from within the `triggerClickHandler`, but in this case it's not really clear what's going on

```
const triggerClickHandler = (event) => {
  if (event.detail === 0) {
    keyboardUsedRef.current = true;
  }
  /// ...
}
```

(See https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#usage_notes)

TheSonOfThomp: If you do decide to keep it as-is, we'll also need to explicitly call `trigger.props.onKeyDown` if it exists

TheSonOfThomp (commented):
Just need to add the call to `trigger.props.onKeyDown`

packages/menu/src/Menu/Menu.tsx
```diff
@@ -229,6 +233,12 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
   if (trigger) {
     const triggerClickHandler = (event?: React.MouseEvent) => {
       event?.preventDefault();
+
+      // if enter or space key is pressed, event detail is 0
```
TheSonOfThomp, line 237: Might be worth linking the mdn docs, but that's a super nit

TheSonOfThomp (approved):

PR #2525, opened 2024-10-29
LG-4619: Add axes components

tsck:
## ✍️ Proposed changes
- Adds `XAxis` component for adding an x-axis to a chart.
- Adds `YAxis` component for adding a y-axis to a chart.
- Improves chart options update logic
  - `lodash.merge` was being used originally, but this didn't work quite as expected. The goal is to merge partial options objects in without overwriting previously set options that are unrelated to the update. This was not how it worked, leading to bugs when updating the same base keys with different options. I.e. when adding `splitLine` configurations on `xAxis` in order to add grid lines, then adding more axis specific config in `XAxis` related to the actual axis, the `splitLine` would be totally overwritten.
- Improves `@lg-charts/core` Storybook, by adding better descriptions and organization.

🎟 _Jira ticket:_ [LG-4619: Add axes components](https://jira.mongodb.org/browse/LG-4619)

### For new components

## 🧪 How to test changes
- Run in Storybook
- Confirm previous tests still pass


charts/core/src/Chart/hooks/updateUtils.ts
```diff
@@ -32,9 +30,38 @@ export function removeSeries(
...
+
+  return updatedObj;
+}
+
 export function updateOptions(
   currentOptions: ChartOptions,
   options: Partial<ChartOptions>,
 ): Partial<ChartOptions> {
-  return merge({ ...currentOptions, ...options });
+  return recursiveMerge(currentOptions, options);
```
bruugey, line 66: Curious why we're replacing the lodash merge, but that's a question not a blocker

bruugey (approved):

charts/core/src/Chart/hooks/updateUtils.ts
tsck: The goal is to merge partial options objects in without overwriting previously set options that are unrelated to the update. This wasn't actually how `lodash.merge` worked (which I thought it would), leading to bugs when updating the same base keys with different options. I.e. when adding `splitLine` configurations on `xAxis` in order to add grid lines, then adding more axis specific config in the `XAxis` component related to the actual axis, the `splitLine` would be totally overwritten.

The updated test is a good example of what should pass and wasn't with `lodash.merge`:
```ts
  /**
   * Tests that option updates don't overwrite the entire chart options object.
   */
  test('should merge chart options non-destructively', () => {
    const currentOptions: Partial<ChartOptions> = {
      xAxis: {
        show: true,
        splitLine: {
          show: true,
        },
      },
    };
    const updatedOptions = updateOptions(currentOptions, {
      xAxis: {
        show: false, // This should only update the show property and not other properties
      },
      grid: {
        show: true,
      },
    });
    // @ts-ignore: Property 'show' does not exist on type 'Arrayable<AriaOption>'.
    expect(updatedOptions?.xAxis?.show).toBe(false);
    // @ts-ignore: Property 'show' does not exist on type 'Arrayable<AriaOption>'.
    expect(updatedOptions?.xAxis?.splitLine?.show).toBe(true);
    // @ts-ignore: Property 'show' does not exist on type 'Arrayable<GridOption>'.
    expect(updatedOptions?.grid?.show).toBe(true);
  });
```

PR #2529, opened 2024-10-30
Fix updates

tsck:
## ✍️ Proposed changes
- Removes chart props on unmount to ensure that they are removed when the component is hidden. 

### For new components

## 🧪 How to test changes
- Pulled down package into testing codebase and verified pieces were removed when the components were removed.

### DIFF
--- a/charts/core/src/Chart/hooks/updateUtils.ts
+++ b/charts/core/src/Chart/hooks/updateUtils.ts
@@ -22,9 +22,9 @@ export function removeSeries(
   const updatedOptions = { ...currentOptions };
 
   if (updatedOptions.series) {
-    updatedOptions.series = updatedOptions.series.filter(
-      series => series.name !== name,
-    );
+    updatedOptions.series = [
+      ...updatedOptions.series.filter(series => series.name !== name),
+    ];
   }
 
   return updatedOptions;
--- a/charts/core/src/Grid/Grid.tsx
+++ b/charts/core/src/Grid/Grid.tsx
@@ -26,6 +26,24 @@ export function Grid({ horizontal = true, vertical = true }: GridProps) {
     updatedOptions.xAxis = getUpdatedLineOptions(!!vertical);
     updatedOptions.yAxis = getUpdatedLineOptions(!!horizontal);
     updateChartOptions(updatedOptions);
+
+    return () => {
+      /**
+       * Hides the grid lines when the component is unmounted.
+       */
+      updateChartOptions({
+        xAxis: {
+          splitLine: {
+            show: false,
+          },
+        },
+        yAxis: {
+          splitLine: {
+            show: false,
+          },
+        },
+      });
+    };
   }, [horizontal, vertical, theme]);
 
   return null;
--- a/charts/core/src/Line/Line.tsx
+++ b/charts/core/src/Line/Line.tsx
@@ -6,7 +6,7 @@ import { defaultLineOptions } from './config';
 import { LineProps } from './Line.types';
 
 export function Line({ name, data }: LineProps) {
-  const { addChartSeries } = useChartContext();
+  const { addChartSeries, removeChartSeries } = useChartContext();
 
   useEffect(() => {
     addChartSeries({
@@ -14,6 +14,14 @@ export function Line({ name, data }: LineProps) {
       name,
       data,
     });
+
+    return () => {
+      /**
+       * Remove the series when the component unmounts to make sure the series
+       * is removed when a `Line` is hidden.
+       */
+      removeChartSeries(name);
+    };
   }, [name, data]);
 
   return null;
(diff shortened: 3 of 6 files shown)


charts/core/src/Chart/hooks/updateUtils.ts
```diff
@@ -22,9 +22,9 @@ export function removeSeries(
   const updatedOptions = { ...currentOptions };
 
   if (updatedOptions.series) {
-    updatedOptions.series = updatedOptions.series.filter(
-      series => series.name !== name,
-    );
+    updatedOptions.series = [
+      ...updatedOptions.series.filter(series => series.name !== name),
+    ];
```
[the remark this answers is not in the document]
bruugey: 👍

charts/core/src/Grid/Grid.tsx
```diff
@@ -26,6 +26,24 @@ export function Grid({ horizontal = true, vertical = true }: GridProps) {
     updatedOptions.xAxis = getUpdatedLineOptions(!!vertical);
     updatedOptions.yAxis = getUpdatedLineOptions(!!horizontal);
     updateChartOptions(updatedOptions);
+
+    return () => {
```
bruugey, line 30: This does not get unmounted by default?

charts/core/src/YAxis/YAxis.tsx
```diff
@@ -96,6 +96,22 @@ export function YAxis({ type, label, unit }: YAxisProps) {
 
   useEffect(() => {
     updateChartOptions(getOptions({ type, label, unit, theme }));
+
+    return () => {
+      /**
+       * Hides the axis when the component is unmounted.
```
bruugey, line 102: Is there a cleaner way to handle unmount?

charts/core/src/Grid/Grid.tsx
tsck: This issue is that this isn't a real component, in a sense, given that it doesn't actually render anything to the DOM. Instead, this ends up basically being "config via JSX". So on unmount, essentially nothing happens. The component itself, which renders nothing, IS removed from the DOM. But this does nothing to actually update the chart that's rendered, since the chart is config based.

charts/core/src/YAxis/YAxis.tsx
tsck: I'm not sure. Definitely open to ideas! The problem is as noted in [this comment](https://github.com/mongodb/leafygreen-ui/pull/2529#discussion_r1823225539). So the goal here is to essentially manually trigger a chart config update any time the component unmounts.

I'll actually chuck this out to the team async and see if anyone has any cleaner ideas for this!

charts/core/src/Line/Line.tsx
```diff
@@ -6,14 +6,22 @@ import { defaultLineOptions } from './config';
...
       name,
       data,
     });
+
+    return () => {
+      /**
+       * Remove the series when the component unmounts to make sure the series
+       * is removed when a `Line` is hidden.
+       */
+      removeChartSeries(name);
```
TheSonOfThomp, line 23: The only place I see a potential issue is if the consumer has rendered another line with the same name before this one unmounts, and we then remove the wrong line from the chart. But at that point, I'd almost call that "user error"

TheSonOfThomp: Probably a check within `addChartSeries` that logs something to the console, and exits early?

tsck: Yeah and actually `addChartSeries` is buggy atm. It adds duplicates on re-renders. Was going to fix that in a followup but maybe I should just fix that here. Will push up this fix shortly

tsck: So oddly I'm finding that `addChartSeries` ends up getting called twice on render, even if the `useEffect` in `Line` only runs once. I'm kind of confused as to why 🤔. That said, this means that even legit series are sometimes getting added twice. I've added logic to prevent duplication and updated the docs to say that it needs to be unique and that non-uniquely named series won't be rendered, however I didn't add an error log due to it triggering on legit cases. Do you think that's enough?

charts/core/src/Chart/hooks/updateUtils.spec.ts
```diff
@@ -14,6 +14,17 @@ describe('@lg-charts/core/Chart/hooks/updateUtils', () => {
     expect(updatedOptions.series?.[1].name).toBe(newSeriesName);
   });
 
+  test('should not add a series if a chart with the same name exists', () => {
```
TheSonOfThomp, line 17: nit: add the name of the util we're testing to the test name (i.e. "`addSeries` should ..."

TheSonOfThomp (approved):
LGTM!

PR #2536, opened 2024-11-01
LG-4659: Replace axes `unit` prop with `formatter`

tsck:
## ✍️ Proposed changes
- Replace `unit` prop of `XAxis` and `YAxis` with `formatter` prop.

🎟 _Jira ticket:_ [LG-4659](https://jira.mongodb.org/browse/LG-4659)

### For new components

## 🧪 How to test changes
- Use it in Storybook

### DIFF
--- a/.changeset/silly-jars-cheat.md
+++ b/.changeset/silly-jars-cheat.md
@@ -1,0 +1,5 @@
+---
+'@lg-charts/core': minor
+---
+
+Replaces `unit` prop of `XAxis` and `YAxis` with `formatter` prop
--- a/charts/core/README.md
+++ b/charts/core/README.md
@@ -92,11 +92,11 @@ Renders an x-axis.
 
 #### Props
 
-| Name                     | Description                                                                                                                                                                                                         | Type                                                                              | Default |
-| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------- | ------- |
-| `type`                   | Type of axis.                                                                                                                                                                                                       | 'category' \| 'value' \| 'time' \| 'log'                                          |         |
-| `label` _(optional)_     | Label name to be rendered on the axis.                                                                                                                                                                              | string                                                                            |         |
-| `formatter` _(optional)_ | Formatter of axis label, which supports string template and callback function. (See [ECharts xAxis.axisLabel.formatter docs](https://echarts.apache.org/en/option.html#xAxis.axisLabel.formatter) for more details) | string \| ((value: string, index: number) => string) \| { [key: string]: string } |         |
+| Name                     | Description                                   | Type                                     | Default |
+| ------------------------ | --------------------------------------------- | ---------------------------------------- | ------- |
+| `type`                   | Type of axis.                                 | 'category' \| 'value' \| 'time' \| 'log' |         |
+| `label` _(optional)_     | Label name to be rendered on the axis.        | string                                   |         |
+| `formatter` _(optional)_ | Callback function for formatting tick values. | (value: string, index: number) => string |         |
 
 ### `YAxis`
 
@@ -104,8 +104,8 @@ Renders a y-axis.
 
 #### Props
 
-| Name                     | Description                                                                                                                                                                                                         | Type                                                                              | Default |
-| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------- | ------- |
-| `type`                   | Type of axis.                                                                                                                                                                                                       | 'category' \| 'value' \| 'time' \| 'log'                                          |         |
-| `label` _(optional)_     | Label name to be rendered on the axis.                                                                                                                                                                              | string                                                                            |         |
-| `formatter` _(optional)_ | Formatter of axis label, which supports string template and callback function. (See [ECharts yAxis.axisLabel.formatter docs](https://echarts.apache.org/en/option.html#yAxis.axisLabel.formatter) for more details) | string \| ((value: string, index: number) => string) \| { [key: string]: string } |         |
+| Name                     | Description                                   | Type                                     | Default |
+| ------------------------ | --------------------------------------------- | ---------------------------------------- | ------- |
+| `type`                   | Type of axis.                                 | 'category' \| 'value' \| 'time' \| 'log' |         |
+| `label` _(optional)_     | Label name to be rendered on the axis.        | string                                   |         |
+| `formatter` _(optional)_ | Callback function for formatting tick values. | (value: string, index: number) => string |         |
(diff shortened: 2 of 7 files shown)


.changeset/silly-jars-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-charts/core': minor
```
tsck, line 2: I guess this is technically major since it's not backwards compatible, but since it's still in v0, I made it minor. Let me know if this is incorrect, though!

TheSonOfThomp: Yeah, pre-v1 changesets can be a bit loose. Rule of thumb for me is everything shifts down a level

charts/core/README.md
```diff
@@ -92,20 +92,20 @@ Renders an x-axis.
...
+| `type`                   | Type of axis.                                                                                                                                                                                                       | 'category' \| 'value' \| 'time' \| 'log'                                          |         |
+| `label` _(optional)_     | Label name to be rendered on the axis.                                                                                                                                                                              | string                                                                            |         |
+| `formatter` _(optional)_ | Formatter of axis label, which supports string template and callback function. (See [ECharts xAxis.axisLabel.formatter docs](https://echarts.apache.org/en/option.html#xAxis.axisLabel.formatter) for more details) | string \| ((value: string, index: number) => string) \| { [key: string]: string } |         |
```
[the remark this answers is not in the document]
TheSonOfThomp: Do we want to commit ourselves to sticking with the ECharts API, or do we want to give ourselves more feedom?

charts/core/src/XAxis/XAxis.tsx
```diff
@@ -87,17 +84,17 @@ const getOptions = ({
  *   <XAxis
  *     type="time",
  *     label="My X-Axis Data",
- *     unit="GB"
+ *     formatter="{value}GB"
```
TheSonOfThomp, line 87: is this correct?

charts/core/src/XAxis/XAxis.types.ts
```diff
@@ -18,7 +18,21 @@ export interface XAxisProps {
...
+   * formatter: function (value, index) {
+   *   return value + 'kg';
+   *}
+   * ```
    */
-  unit?: string;
+  formatter?:
+    | string
+    | ((value: string, index: number) => string)
+    | { [key: string]: string };
```
TheSonOfThomp, line 37: What would an object formatter look like?

charts/core/src/XAxis/XAxis.tsx
tsck: Yup, exactly. Specifically for rendering the value. There are other formatters such as `'{yyyy}-{MM}-{dd}'` as well.

charts/core/README.md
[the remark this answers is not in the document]
bruugey: Good call out - agreed

charts/core/README.md
```diff
@@ -26,7 +26,7 @@ import { Chart, Line, Grid, XAxis, YAxis } from '@lg-charts/core';
 <Chart>
   <Grid vertical={false}>
   <XAxis type="time" />
-  <YAxis type="value" unit="GB" />
+  <YAxis type="value" formatter="{value}GB" />
```
bruugey, line 29: this api does feel a little tied to echarts - should we leverage a more expected API and then normalize it on our side before sending thru

bruugey: Certainly open to discussion here, just wanted to call it out

charts/core/src/XAxis/XAxis.types.ts
tsck: The object formatter is what echarts calls a "cascading template" and only works with time. From their site:
> Sometimes, we wish to use different formats for different time granularity. For example, in a quarter-year chart, we may wish to see the month name with the first date of the month, while see the date name with others. This can be made with:
> ```ts
> formatter: {
>    month: '{MMMM}', // Jan, Feb, ...
>    day: '{d}' // 1, 2, ...
>}
>```

I guess that means this could actually be typed better:
```suggestion
    | { [key: CascadingTemplateLevels]: string };
```
where `CascadingTemplateLevels` is defined as:
```ts
type CascadingTemplateLevels =
  | 'year'
  | 'month'
  | 'day'
  | 'hour'
  | 'minute'
  | 'second'
  | 'millisecond'
  | 'none';
```

However, I think given your point in [this comment](https://github.com/mongodb/leafygreen-ui/pull/2536/files#r1826051813), I'm wondering if we should publicly support this. Given that EChart's string interpolation isn't standard JS either, I'm almost wondering if we should just publicly support the callback function. Consumers can still get hacky, override the type and use either of these, but by not publicly supporting them it probably helps save some headaches in the future if we ever need to switch the underlying library. What do you think?

charts/core/src/YAxis/YAxis.tsx
```diff
@@ -86,17 +83,17 @@ const getOptions = ({
  *   <YAxis
  *     type="value",
  *     label="My Y-Axis Data",
- *     unit="GB"
+ *     formatter="{value}GB"
  *   />
  * </Chart>
  */
-export function YAxis({ type, label, unit }: YAxisProps) {
+export function YAxis({ type, label, formatter }: YAxisProps) {
```
TheSonOfThomp, line 90: Looks like there's a lot of similarity between XAxis & YAxis. Not necessary in this PR, but could be worth DRY-ing this up a bit

TheSonOfThomp (approved):

charts/core/README.md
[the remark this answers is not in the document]
tsck: Addressing this more in [thread here](https://github.com/mongodb/leafygreen-ui/pull/2536#discussion_r1826138443) (TL;DR - I think this probably should be removed, and we should only publicly support a subset for this)

charts/core/README.md
tsck: Agreed! I'm thinking that we just publicly expose the callback. Addressing this more in [this thread](https://github.com/mongodb/leafygreen-ui/pull/2536#discussion_r1826138443)

charts/core/src/YAxis/YAxis.tsx
tsck: Yup, they're almost identical atm with just a few different props. I considered having a more common component that these two work off of but decided not to since there are only 2, but I agree, it probably makes sense. I'll move in this direction in a future PR.

charts/core/README.md
bruugey: awesome

charts/core/src/XAxis/XAxis.types.ts
tsck: Decided to make the decision to only support the callback function and have pushed changes to reflect that. Let me know if there's any disagreement here.

PR #2506, opened 2024-10-16
Table Updates

shaneeza:
## This PR includes:

- updated`useLeafyGreenTable` hook
- adds `useLeafyGreenVirtualTable` hook
- remove mapping of children from inside the [InternalRowWithRT](packages/table/src/Row/InternalRowWithRT.tsx) component
- memorizes rows inside [InternalRowWithRT](packages/table/src/Row/InternalRowWithRT.tsx)
- supports styled components
- multiline row support
- removal of table v10 and it's adapter

## Not include:

- ForwardRefs. #2539 
- Documentation and examples - PR coming soon
- Updated specs and test utils - PR coming soon
- sticky header scroll shadow - PR coming soon
- changesets

## Storybook links
_Note_: Some stories with hundreds of rows will freeze storybook on initial render. A workaround while I figure this out is to view the story outside of an iframe. 
_Also_, tables currently expand faster and scroll better in incognito

🎟 _Jira ticket:_ [LG-4414](https://jira.mongodb.org/browse/LG-4414)

### For bug fixes, new features & breaking changes

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] I have run `yarn changeset` and documented my changes

### For new components

- [ ] I have added my new package to the global tsconfig
- [ ] I have added my new package to the Table of Contents on the global README
- [ ] I have verified the Live Example looks as intended on the design website.

PR #2539, opened 2024-11-04
[WIP ]Table forwardRefs

shaneeza:
## ✍️ Proposed changes

### For new components


packages/hooks/src/index.ts
```diff
@@ -9,6 +9,7 @@ export { useForceRerender } from './useForceRerender';
 export { useForwardedRef, useObservedRef } from './useForwardedRef';
 export { default as useIdAllocator } from './useIdAllocator';
 export { default as useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
+export { default as useMergeRefs } from './useMergeRefs';
```
shaneeza, line 12: Update to use the same hook that stephen is using in his popover work

packages/table/src/ExpandedContent/ExpandedContent.spec.tsx
```diff
@@ -39,21 +38,23 @@ const RowWithExpandableContent = args => {
...
+                      <Cell key={cell.id}>
+                        {flexRender(
+                          cell.column.columnDef.cell,
+                          cell.getContext(),
+                        )}
+                      </Cell>
+                    );
+                  })}
+                </Row>
+                {row.original.renderExpandedContent && row.getIsExpanded() && (
```
shaneeza, line 54: This isn't correct, update

packages/table/src/Row/Row.tsx
```diff
@@ -8,51 +8,50 @@ import { LGRowData } from '../useLeafyGreenTable';
...
-          <InternalRowWithoutRT {...rest} />
+          <InternalRowWithoutRT ref={ref} {...rest} />
         </RowContextProvider>
       )}
     </>
   );
 };
 
+// TODO: Can i avoid asserting?
+export const Row = React.forwardRef(RowWithRef) as RowComponentType;
```
tsck, line 53: In the [SO answer you shared in the type comment](https://stackoverflow.com/a/58473012) the 2nd wrapper option would avoid having to assert here. The 3rd option is quite clean, but I get why we wouldn't want to change the API to some custom ref prop.

Regardless of approach, it might be worth adding a small comment here as well! It's definitely a bit odd at a glance why we're doing this without looking at the type itself to see the comment there. Even if it's just one line summary.

packages/table/src/Cell/Cell.tsx
```diff
@@ -1,31 +1,35 @@
...
-        <InternalCellWithRT {...rest} cell={cell}>
+        <InternalCellWithRT ref={ref} {...rest} cell={cell}>
           {children}
         </InternalCellWithRT>
       )}
     </>
   );
 };
 
+export const Cell = React.forwardRef(CellWithForwardRef) as CellComponentType;
```
tsck, line 31: Ditto on maybe there being some value having a comment here explaining this a bit in addition to the type comment.

packages/table/src/Row/Row.spec.tsx
```diff
@@ -58,27 +57,28 @@ const RowWithNestedRows = args => {
...
-                            {flexRender(
-                              cell.column.columnDef.cell,
-                              cell.getContext(),
-                            )}
-                          </Cell>
-                        );
-                      })}
-                    </Row>
-                  ))}
               </Row>
```
tsck, line 60: Q - For my own context: Is this the API we currently have, where sub-rows are siblings and not children?

packages/table/src/Cell/InternalCellWithRT.tsx
```diff
@@ -43,10 +53,14 @@ const InternalCellWithRT = <T extends LGRowData>({
...
       )}
       <div className={getCellEllipsisStyles(shouldTruncate)}>{children}</div>
-    </InternalCell>
+    </InternalCellBase>
   );
 };
 
+export const InternalCellWithRT = React.forwardRef(
+  InternalCellWithRTForwardRef,
+) as InternalCellWithRTComponentType;
```
tsck, line 62: Ditto, small comment as to why would be helpful

packages/table/src/Cell/HeaderCell/HeaderCell.tsx
```diff
@@ -73,4 +82,8 @@ const HeaderCell = <T extends LGRowData>({
   );
 };
 
+export const HeaderCell = React.forwardRef(
+  HeaderCellWithRef,
+) as HeaderCellComponentType;
```
tsck, line 87: Ditto on comment

packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -77,20 +82,22 @@ const InternalRowWithRT = <T extends LGRowData>({
   );
 };
 
+export const InternalRowWithRT = React.forwardRef(
+  InternalRowWithRTForwardRef,
+) as RowComponentWithRTType;
```
tsck, line 87: Another good place for a comment 🙂

packages/table/src/ExpandedContent/ExpandedContent.tsx
```diff
@@ -47,6 +46,11 @@ const ExpandedContent = <T extends RowData>({
   );
 };
 
+// TODO: Can I avoid asserting
+export const ExpandedContent = React.forwardRef(
+  ExpandedContentWithRef,
+) as ExpandedContentComponentType;
```
tsck, line 52: One more time lol

tsck (commented):
Have a few questions and some small comments, but mostly LGTM!

packages/table/src/TableBody/TableBody.spec.tsx
```diff
@@ -18,53 +15,13 @@ describe('packages/table/TableBody', () => {
...
-          </div>
-        );
-      };
+    test('accepts a ref', () => {
+      const ref = React.createRef<HTMLTableSectionElement>();
+      render(<TableBody ref={ref}>Hello</TableBody>);
 
-      const { container } = render(<DefaultTest />);
-      // would be better to check explicitly if it's a fragment
-      expect(container.querySelector('tbody')).toBeInTheDocument();
```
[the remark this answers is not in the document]
shaneeza: Sorry for the confusion, it's hard to see in this PR but TableBody used to be polymorphic. You can check out the diff here https://github.com/mongodb/leafygreen-ui/pull/2506/files#diff-bb0a0315ae6a1543465b038315ec8f320af48e6383909a9181f3c4f4fcc91175. In this version, it's no longer polymorphic and it always renders as `tbody`

packages/table/src/Row/Row.spec.tsx
shaneeza: Correct, and the new API is not actually reflected in my changes to file. I'm going to create a new PR updating all the specs.

tsck (approved):
Look like there might be some TS issues, but otherwise LGTM!

PR #2550, opened 2024-11-20
LG-4620: `ChartCard` and `Header` components

tsck:
## ✍️ Proposed changes

This adds both `ChartCard` and `Header` components.

🎟 _Jira tickets:_ 
- [LG-4620](https://jira.mongodb.org/browse/LG-4620)
- [LG-4575](https://jira.mongodb.org/browse/LG-4575)

### For new components

## 🧪 How to test changes
- See working in Storybook
- Confirm unit tests are passing


charts/core/README.md
```diff
@@ -58,66 +84,92 @@ Chart container component.
 
 #### Props
 
-| Name           | Description                                             | Type       | Default |
-| -------------- | ------------------------------------------------------- | ---------- | ------- |
-| `onChartReady` | Callback to be called when chart is finished rendering. | () => void |         |
+| Name           | Description                                             | Type         | Default |
+| -------------- | ------------------------------------------------------- | ------------ | ------- |
+| `onChartReady` | Callback to be called when chart is finished rendering. | `() => void` |         |
```
tsck, line 89: Made some drive-by style changes in the README, mostly to use backticks on some code that wasn't previously formatted

charts/core/src/core.stories.tsx
tsck, line 1: I made a bunch of updates to the story to provide more configuration controls. Since there are multiple components shown at once, I thought it made sense to provide controls to turn them all off and on.

charts/core/src/ChartCard/ChartCard.tsx
```diff
@@ -0,0 +1,118 @@
...
+      setHeaderHeight(headerRef.current.offsetHeight);
+    }
+  }, [containerRef, headerRef]);
+
+  return (
+    <div
+      className={cx(
+        getContainerStyles(theme, height, headerHeight),
+        className,
+        isOpen && 'open',
```
shaneeza, line 70: We don't typically use this pattern. An alternative could be to pass `isOpen` to `getContainerStyles(theme, height, headerHeight, isOpen)` and then conditionally render the style:

```
export const getContainerStyles = (
  theme: Theme,
  height?: number,
  headerHeight?: number,
  isOpen,
) => cx(css`
  ... default styles here
  `, 
  {
   [css`
       ... open styles here
   `]: isOpen
  }

File: charts/core/src/ChartCard/ChartCard.styles.ts [hunk 3]
```diff
@@ -0,0 +1,63 @@
...
+  * a wrapper around the children because a bottom border is set on the container and header.
+  * By shrinking the container height we prevent a conflict between the container border 
+  * and the header border when collapsed. By making the container height 1px shorter on collapse
+  * we just cutoff the header border entirely.
+  */
+  /* Accounts for 1px border */
+  max-height: ${headerHeight ? headerHeight + 1 + 'px' : 'auto'};
+  transition: max-height ${transitionDuration.slower}ms ease-in-out;
+  &.open {
+    max-height: ${height ? height + 'px' : 'auto'};
```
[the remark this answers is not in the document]
tsck: Oh very interesting, I love this. Will push a change for this shortly!

charts/core/src/ChartCard/ChartCard.tsx
```diff
@@ -0,0 +1,118 @@
...
+        ref={headerRef}
+      >
+        <div className={leftInnerContainerStyles}>
+          <IconButton
+            className={toggleButtonStyles}
+            id={toggleId}
+            aria-label="Toggle button"
+            aria-controls={childrenId}
+            aria-expanded={isOpen}
+            onClick={(e: MouseEvent<HTMLButtonElement>) => {
```
shaneeza, line 87: nit: you could create a function like `handleIconClick` and pass that to `onClick`

charts/core/src/ChartCard/ChartCard.tsx
```diff
@@ -0,0 +1,118 @@
...
+              className={cx(toggleIconStyles, isOpen && 'open')}
+            />
+          </IconButton>
+          <Body weight="medium" baseFontSize={BaseFontSize.Body2}>
+            {title}
+          </Body>
+        </div>
+        <div>{headerContent}</div>
+      </div>
+      <div style={{ overflow: 'hidden' }}>
```
shaneeza, line 105: nit: classname

charts/core/src/Header/Header.styles.ts
```diff
@@ -0,0 +1,28 @@
+import { css } from '@leafygreen-ui/emotion';
+import { Theme } from '@leafygreen-ui/lib';
+import {
+  color,
+  InteractionState,
+  spacing,
+  Variant,
+} from '@leafygreen-ui/tokens';
+
+export const getContainerStyles = (theme: Theme, showDivider?: boolean) => css`
```
shaneeza, line 10: A tip i got from Adam was to use 
```
cx(css`...`, {[css`...`]:showDivider})
```
instead of 
```
css`
...
${showDivider && ``}
`
```
`cx` is better for conditional classes

charts/core/src/Chart/Chart.styles.ts
```diff
@@ -1,22 +1,16 @@
...
-  InteractionState,
-  Variant,
-} from '@leafygreen-ui/tokens';
 
-export const chartStyles = css`
-  grid-row: 2;
+export const chartContainerStyles = css`
+  display: grid;
+  grid-template-areas:
+    'chartHeader'
```
shaneeza, line 6: nit: Do you explicitly state anywhere that the grid-area for the `headerContent` is `chartHeader`?

charts/core/src/Chart/Chart.tsx
```diff
@@ -45,12 +45,21 @@ export function Chart({
...
+             * but are used to provide a more declarative API for adding functionality
+             * to the chart canvas. They have access to the ChartContext and can be
+             * used to add components like Line, Grid, etc.
+             */}
+            {children}
+          </div>
           <div
             ref={chartRef}
-            className={`echart ${chartStyles}`}
+            className={cx('echart', chartStyles)}
```
shaneeza, line 60: nit: why do you need `echart`?

charts/core/src/ChartCard/ChartCard.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+  title: 'test',
+  headerContent: <div>{defaultContentText}</div>,
+  children: 'Chartchartchartchartchart',
+};
+
+const renderChartCard = (props?: Partial<ChartCardProps>) =>
+  render(<ChartCard {...defaultProps} {...props} />);
+
+const clickToggleButton = () => {
+  userEvent.click(screen.getByLabelText('Toggle button'));
```
shaneeza, line 21: nit: Thoughts on using a testid here? Also, will we be providing test utils for this component? If so, we have a pattern of creating a constant file to store testids. For example: https://github.com/mongodb/leafygreen-ui/blob/main/packages/select/src/constants.ts

charts/core/README.md
```diff
@@ -58,66 +84,92 @@ Chart container component.
...
+
+Component for rendering a header in a chart.
+
+#### Props
+
+| Name                         | Description                                                                                                                                           | Type              | Default |
+| ---------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------- | ------- |
+| `title`                      | The title to display in the chart header.                                                                                                             | `string`          |         |
+| `showDivider` _(optional)_   | When true, renders a dividing line on top of header. Useful when stacking charts, such as in a `ChartGroup`.                                          | `boolean`         |         |
+| `headerContent` _(optional)_ | Content that will be rendered to the right of the title. Useful for adding components such as `IconButton`'s that control functionality in the chart. | `React.ReactNode` |         |
```
shaneeza, line 128: nit: I didn't see a story that showcases this feature. Would you consider adding one to ensure that it works?

tsck: Absolutely. I'm curious how we demo something like this since it takes in a `ReactNode`. Any suggestions?

charts/core/src/Chart/Chart.styles.ts
tsck: I'm not sure that's correct. `chartHeader` is the grid-area for `Header`, `headerContent` is a node that gets ultimately inside of the `Header`. Am I misunderstanding what's being asked here?

shaneeza: My bad! I confused myself. Where did you add `grid-area: chartHeader`?

charts/core/src/ChartCard/ChartCard.spec.tsx
shaneeza: Thanks for clarifying!

charts/core/README.md
shaneeza: Code might be a good example: https://storybook.mongodb.design/?path=/story/components-code--with-custom-actions.

I think you would need to have individual stories for standalone components.

charts/core/src/Chart/Chart.styles.ts
tsck: Great catch! Totally missed it. Since I happened to order it correctly in storybook I didn't realize I didn't actually add the area to the component. Will update!

charts/core/src/ChartCard/ChartCard.tsx
```diff
@@ -0,0 +1,118 @@
...
+  leftInnerContainerStyles,
+  toggleButtonStyles,
+  toggleIconStyles,
+} from './ChartCard.styles';
+import { ChartCardProps } from './ChartCard.types';
+
+/**
+ * Card component that contains charts and can expand and collapse.
+ */
+export function ChartCard({
```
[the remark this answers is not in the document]
tsck: I'm not sure but I can't think of a good reason for it to not? I can add it in the next push

charts/core/src/ChartCard/ChartCard.tsx
tsck: Gotcha, will change!

charts/core/src/ChartCard/ChartCard.tsx
tsck: This is actually an inline style and not a class, BUT it's leftover for something I was trying quickly inline that didn't work 😅 I'll clean this up

charts/core/src/Header/Header.styles.ts
tsck: Cool! Will do!

charts/core/README.md
tsck: I ended up doing a pretty big refactor here. I separated things into 2 stories "Chart" and "ChartCard", with the following stories:
- Chart
  - Basic
  - With Header Content 

- ChartCard
  - Basic
  - With Header Content
  - Containing Chart
  - Containing Multi Charts

I'd rather not do it in this PR (because folks are waiting to use this), but I actually think this is how the packages should ultimately be split as well. `@lg-charts/core` should probably really be `@lg-charts/chart` and only export what's needed for a chart itself. `@lg-charts/chart-card` should be separate and only export the card.

charts/core/src/ChartCard/ChartCard.styles.ts
```diff
@@ -0,0 +1,54 @@
...
+  border: 1px solid
+    ${color[theme].border[Variant.Disabled][InteractionState.Default]};
+  border-radius: ${borderRadius[200]}px;
+  overflow: hidden;
+  width: 100%;
+  display: grid;
+  grid-template-rows: 40px 0fr;
+  transition: grid-template-rows ${transitionDuration.slower}ms ease-in-out;
+
+  &.open {
```
shaneeza, line 22: Curious: why do you prefer using `.open` instead of conditionally rendering the styles with `cx`?

charts/core/README.md
shaneeza: sounds good to me

charts/core/src/ChartCard/ChartCard.styles.ts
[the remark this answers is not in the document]
shaneeza: i'm glad this worked!!!

charts/core/src/ChartCard/ChartCard.tsx
```diff
@@ -0,0 +1,107 @@
...
+              className={toggleButtonStyles}
+              id={toggleId}
+              aria-label="Toggle button"
+              aria-controls={childrenId}
+              aria-expanded={isOpen}
+              onClick={handleToggleButtonClick}
+            >
+              <Icon
+                glyph="ChevronDown"
+                className={cx(toggleIconStyles, isOpen && 'open')}
```
shaneeza, line 82: nit: 
```
[css`...`]:open
```

shaneeza (approved):

charts/core/src/ChartCard/ChartCard.styles.ts
tsck: Actually not a preference, just not used to having `cx` as an option! Will update this.

PR #2555, opened 2024-11-21
`Table` updated tests

shaneeza:
## ✍️ Proposed changes

### For new components


packages/table/src/Row/Row.spec.tsx
shaneeza, line 1: The tests from this file were moved to `RowWithRT.spec.tsx`, and the tests from `RowWithoutRT.spec.tsx` were moved to this file.

packages/table/src/Cell/HeaderCell/HeaderCell.spec.tsx
```diff
@@ -198,4 +199,56 @@ describe('packages/table/HeaderCell', () => {
...
+    test('works with `styled`', () => {
+      const StyledHeaderCell = styled(HeaderCell)`
+        color: #69ffc6;
+      `;
+
+      const { getByTestId } = render(
+        <StyledHeaderCell data-testid="styled">Some text</StyledHeaderCell>,
+      );
+
+      expect(getByTestId('styled')).toBeInTheDocument();
```
tsck, line 213: Nit: Wonder if it's necessary to check if this is in the document? It's not actually an implementation detail since we're rendering it directly in the test. Won't comment this on each one to bloat the review, but have the same thinking for each of these styled tests.

packages/table/src/Row/RowWithoutRT.spec.tsx
tsck, line 1: As we discussed yesterday, I wanted this think about this some more but I think this is the correct call. We don't specify `WithoutRT` anywhere else, so I think this makes sense.

One question - it doesn't seem like all of these tests were moved over. Was that intentional?

packages/table/src/Row/Row.spec.tsx
```diff
@@ -1,131 +1,154 @@
...
+    </table>,
   );
-};
+}
 
-// TODO: this should not be called NestedRows
-describe('packages/table/Row/NestedRows', () => {
-  test('renders the correct number of children', () => {
-    render(<RowWithNestedRows />);
+describe('packages/table/RowWithoutRT', () => {
```
tsck, line 35:
```suggestion
describe('packages/table/Row', () => {
```

packages/table/src/Row/Row.spec.tsx
```diff
@@ -1,131 +1,154 @@
...
-    test('regular cell', () => {
-      const ref = React.createRef<HTMLTableRowElement>();
-      render(<Row ref={ref}>Hello</Row>);
+  describe('onClick prop applies correct styles and tabIndex', () => {
+    test('onClick prop is called correctly on click', async () => {
+      const onClick = jest.fn();
+      renderRow({ ...defaultProps, onClick });
+      const { getRowByIndex } = getTestUtils();
+      const firstRow = getRowByIndex(0)?.getElement();
+      fireEvent.click(firstRow!);
```
tsck, line 84: Any reason for using `fireEvent` instead of `userEvent` here?

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.spec.tsx
```diff
@@ -1,44 +1,73 @@
...
+        data: getDefaultTestData({
+          // eslint-disable-next-line react/display-name
+          renderExpandedContent: (_: LeafyGreenTableRow<Person>) => {
+            return <>Expandable content test</>;
+          },
+        }),
+        columns: getDefaultTestColumns({}),
+      }),
+    );
+    expect(result.current.getRowModel().rows.length).toEqual(3);
```
tsck, line 47: If I'm correctly understanding why we're using 3 here, we could map this number more directly to the source here.

```suggestion
    const data = getDefaultTestData({
      // eslint-disable-next-line react/display-name
      renderExpandedContent: (_: LeafyGreenTableRow<Person>) => {
        return <>Expandable content test</>;
      },
    });
    const { result } = renderHook(() =>
      useLeafyGreenTable({
        data,
        columns: getDefaultTestColumns({}),
      }),
    );
    expect(result.current.getRowModel().rows.length).toEqual(data.length);
```

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.spec.tsx
```diff
@@ -1,44 +1,73 @@
...
+  test('returns the correct number of rows with subRows', () => {
+    const { result } = renderHook(() =>
+      useLeafyGreenTable({
+        data: getDefaultTestData({
+          subRows: getDefaultTestData({}),
+        }),
+        columns: getDefaultTestColumns({}),
+      }),
+    );
+    expect(result.current.getRowModel().rows.length).toEqual(3);
```
tsck, line 59: Ditto 

```suggestion
    const data = getDefaultTestData({
      subRows: getDefaultTestData({}),
    });
    const { result } = renderHook(() =>
      useLeafyGreenTable({
        data,
        columns: getDefaultTestColumns({}),
      }),
    );
    expect(result.current.getRowModel().rows.length).toEqual(data.length);
```

packages/table/src/useLeafyGreenVirtualTable/useLeafyGreenVirtualTable.spec.tsx
```diff
@@ -1 +1,44 @@
...
+        columns: [],
+        data: [],
+      }),
+    );
+    expect(typeof result.current).toBe('object');
+  });
+
+  // eslint-disable-next-line jest/no-disabled-tests
+  test.skip('Typescript', () => {
+    // @ts-expect-error - requires columns, data, containerReff
```
tsck, line 21: Very nit =)

```suggestion
    // @ts-expect-error - requires columns, data, containerRef
```

packages/table/src/Row/RowWithoutRT.spec.tsx
shaneeza: Yes. I removed the tests for the zebra stripping because that seems more of a chromatic test. I thought we had a generated story to test zebra stripes, but it looks like we don't, so I'll update the generated row story to include it.

tsck: That makes sense to me!

tsck (approved):

PR #2558, opened 2024-11-26
`Table` types r17 

shaneeza:
## ✍️ Proposed changes

My initial solution worked in R18 but failed in R17 when a generic forward and generic memo were combined. This is a new reusable implementation.

In this PR

### For new components


packages/lib/src/types/ForwardRefWithGenerics.ts
```diff
@@ -0,0 +1,32 @@
...
+ * E.g.
+ * 
+ * ```
+ * const forwardRefWithGenerics: ForwardRefWithGenerics = React.forwardRef;
+ * 
+ * export const Row = forwardRefWithGenerics(
+  function Row<T extends LGRowData>(){})
+ *
+ * // @ts-expect-error
+ * Row.propTypes = {};
```
TheSonOfThomp, line 23: I think this might fail if someone tries to use `styled(Row)`

PR #2567, opened 2024-12-03
LG-4577: Implement syncing tooltips across charts

tsck:
## ✍️ Proposed changes

Adds the ability to synchronize tooltips across charts

🎟 _Jira ticket:_ [LG-4577](https://jira.mongodb.org/browse/LG-4577)

### For new components

## 🧪 How to test changes
Check 'With Same Group Ids' story in Storybook

### DIFF
--- a/charts/core/README.md
+++ b/charts/core/README.md
@@ -57,7 +57,7 @@ import { Chart, Line, Grid, XAxis, YAxis } from '@lg-charts/core';
 import { ChartCard, Chart, Line, XAxis, YAxis } from '@lg-charts/core';
 
 <ChartCard title="My Group of Charts">
-  <Chart>
+  <Chart groupId="group1">
     <XAxis type="time" />
     <YAxis type="value" formatter={(value) => `${value}GB`} />
     <Line
@@ -65,7 +65,7 @@ import { ChartCard, Chart, Line, XAxis, YAxis } from '@lg-charts/core';
       data={seriesData}
     />
   </Chart>
-  <Chart>
+  <Chart groupId="group1">
     <XAxis type="time" />
     <YAxis type="value" formatter={(value) => `${value}GB`} />
     <Line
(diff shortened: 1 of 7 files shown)


charts/core/README.md
```diff
@@ -88,6 +88,7 @@ Chart container component.
 | --------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------ | ------- |
 | `onChartReady`              | Callback to be called when chart is finished rendering.                                                                                       | `() => void`                   |         |
 | `onZoomSelect` _(optional)_ | Callback to be called when a user clicks and drags on a chart to zoom. Click and drag action will only be enabled if this handler is present. | `(ZoomSelectionEvent) => void` |         |
+| `groupId` _(optional)_      | Charts with the same `groupId` will have their tooltips synced across charts.                                                                 | `string`                       |         |
```
tsck, line 91: `groupId` is what it's called in ECharts. We could in theory call it whatever we want though. I was thinking like `syncId` but wasn't sure if that was clearer. Open to ideas here!

shaneeza: Since ECharts calls it `groupId` i think the name makes sense. Could groupId be used for anything else?

charts/core/src/Chart.stories.tsx
```diff
@@ -209,6 +209,14 @@ export default {
         disable: true,
       },
     },
+    groupId: {
+      description: 'Group ID for synced tooltips',
+      name: 'groupId',
+      table: {
+        category: 'Chart',
+        disable: true,
```
shaneeza, line 217: curious: I don't see this in the controls panel. Does `disable: true` hide it?

charts/core/src/Chart/hooks/useChart.ts
```diff
@@ -58,6 +59,12 @@ export function useChart({
     chartInstanceRef.current = chartInstance;
     chartInstance.setOption(chartOptions);
 
+    // Connects a chart to a group which allows for synchronized tooltips
+    if (groupId) {
+      chartInstance.group = groupId;
```
shaneeza, line 64: Can there only be one `groupId` per chart?

charts/core/src/Chart.stories.tsx
tsck: Yup, that's exactly it. If I don't add this it gives user the option to set the `groupId` which felt like it perhaps didn't demonstrate anything? Instead I just set them manually to the same thing in the `WithSameGroupIds` story so that they sync

charts/core/README.md
tsck: I don't believe so. At least at the moment that's it. Perhaps in the future we could sync more in a group, but nothing is needed with it atm that I'm aware of

charts/core/src/Chart/hooks/useChart.ts
tsck: Yeah, there should be just one `groupId` per chart

shaneeza (approved):

PR #2578, opened 2024-12-05
LG-4698: Allow choosing of axis for zoom selection 

tsck:
## ✍️ Proposed changes

It was an ask from consumers that they be able to optionally turn zoom selection on and off for axis. This change primarily adds that.

Note, this also updates the component's props. Previously we decided to just allow the handler to turn off and on the select-ability. Since now there is an option to enable it by axis, it made sense to just create a `zoomSelect` prop, that configures zoom selection. I like this approach better as a) the name `zoomSelect` clearer communicates that we're not actually zooming on data, and b) it removes the implied enablement of the previous design that none of us were crazy about.

🎟 _Jira ticket:_ [LG-4698](https://jira.mongodb.org/browse/LG-4698)

### For new components

## 🧪 How to test changes
Added to storybook so it can now be tested there.

### DIFF
--- a/charts/core/README.md
+++ b/charts/core/README.md
@@ -86,10 +86,10 @@ Chart container component.
 
 | Name                        | Description                                                                   | Type                                   | Default                          |
 | --------------------------- | ----------------------------------------------------------------------------- | -------------------------------------- | -------------------------------- |
-| `onChartReady`              | Callback to be called when chart is finished rendering.                       | `() => void`                           |                                  |
-| `zoomSelect` _(optional)_   | Enable zoom select (click and drag area selection) for either axis.           | `{ xAxis?: boolean; yAxis?: boolean }` | `{ xAxis: false, yAxis: false }` |
-| `onZoomSelect` _(optional)_ | Callback to be called when a zoom selection is made if enabled.               | `(e: ZoomSelectionEvent) => void`      |                                  |
 | `groupId` _(optional)_      | Charts with the same `groupId` will have their tooltips synced across charts. | `string`                               |                                  |
+| `onChartReady`              | Callback to be called when chart is finished rendering.                       | `() => void`                           |                                  |
+| `onZoomSelect` _(optional)_ | Callback to be called when a zoom selection is made if enabled.               | `(e: ZoomSelectionEvent) => void`      |                                  |
+| `zoomSelect` _(optional)_   | Enable zoom select (click and drag area selection) for either axis.           | `{ xAxis?: boolean; yAxis?: boolean }` | `{ xAxis: false, yAxis: false }` |
 
 **Note**: Callback passed to `onZoomSelect` callback receives the following `ZoomSelectionEvent` as an argument:
 
--- a/charts/core/README.md
+++ b/charts/core/README.md
@@ -84,12 +84,12 @@ Chart container component.
 
 #### Props
 
-| Name                        | Description                                                                   | Type                                   | Default                          |
-| --------------------------- | ----------------------------------------------------------------------------- | -------------------------------------- | -------------------------------- |
-| `groupId` _(optional)_      | Charts with the same `groupId` will have their tooltips synced across charts. | `string`                               |                                  |
-| `onChartReady`              | Callback to be called when chart is finished rendering.                       | `() => void`                           |                                  |
-| `onZoomSelect` _(optional)_ | Callback to be called when a zoom selection is made if enabled.               | `(e: ZoomSelectionEvent) => void`      |                                  |
-| `zoomSelect` _(optional)_   | Enable zoom select (click and drag area selection) for either axis.           | `{ xAxis?: boolean; yAxis?: boolean }` | `{ xAxis: false, yAxis: false }` |
+| Name                        | Description                                                                     | Type                                   | Default                          |
+| --------------------------- | ------------------------------------------------------------------------------- | -------------------------------------- | -------------------------------- |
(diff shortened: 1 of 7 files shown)


charts/core/src/Chart/hooks/useChart.ts
```diff
@@ -64,78 +146,39 @@ export function useChart({
     };
     window.addEventListener('resize', resizeHandler);
 
-    function onInitialRender() {
```
tsck, line 67: This was moved outside of the useEffect for readability

charts/core/src/Chart/hooks/useChart.ts
```diff
@@ -64,78 +146,39 @@ export function useChart({
...
-         * this button and want it turned on by default. This is done by dispatching
-         * an action to enable the "dataZoomSelect" feature, as if it were clicked.
-         */
-        chartInstance.dispatchAction({
-          type: 'takeGlobalCursor',
-          key: 'dataZoomSelect',
-          dataZoomSelectActive: true,
-        });
-
-        chartInstance.on('dataZoom', (params: any) => {
```
tsck, line 90: This callback was moved into a function `handleDataZoom` outside of the `useEffect` for readability.

charts/core/src/Chart/hooks/useChart.ts
tsck, line 1: This whole file should probably be refactored as it's become a bit hard to follow. For time-sake I didn't do that here, but would love to do it before wrapping up the epic if time permits.

greggb: 👍 It would be nice, at some point, to have an easy to reference import of types without digging into the `dist` dir.

charts/core/src/Chart/hooks/useChart.types.ts
```diff
@@ -1,12 +1,29 @@
 import { Theme } from '@leafygreen-ui/lib';
 
 export interface ZoomSelectionEvent {
-  xAxis: { startValue: number; endValue: number };
-  yAxis: { startValue: number; endValue: number };
+  xAxis?: { startValue: number; endValue: number };
+  yAxis?: { startValue: number; endValue: number };
+}
+
+export interface zoomSelect {
```
shaneeza, line 8:
```suggestion
export interface ZoomSelect {
```

charts/core/src/Chart/hooks/useChart.ts
```diff
@@ -41,17 +41,99 @@ echarts.use([
...
+       * this button and want it turned on by default. This is done by dispatching
+       * an action to enable the "dataZoomSelect" feature, as if it were clicked.
+       */
+      chartInstance.dispatchAction({
+        type: 'takeGlobalCursor',
+        key: 'dataZoomSelect',
+        dataZoomSelectActive: true,
+      });
+
+      chartInstance.on('dataZoom', handleDataZoom);
```
shaneeza, line 123: If I understand correctly, even though `onInitialRender` is only called once on initial render, `handleDataZoom` will always be called when someone zooms?

charts/core/src/Chart/hooks/useChart.ts
tsck: Agreed! I've noticed this a few times where I've neglected to export types. I'm going to throw a ticket on the epic to make sure I remember.

charts/core/src/Chart/hooks/useChart.ts
tsck: Correct. We remove `onInitialRender` as a render callback, but we never remove `handleDataZoom` as a dataZoom callback.

shaneeza (approved):

charts/core/README.md
```diff
@@ -86,10 +86,10 @@ Chart container component.
...
+| `groupId` _(optional)_      | Charts with the same `groupId` will have their tooltips synced across charts. | `string`                               |                                  |
 | `onChartReady`              | Callback to be called when chart is finished rendering.                       | `() => void`                           |                                  |
-| `zoomSelect` _(optional)_   | Enable zoom select (click and drag area selection) for either axis.           | `{ xAxis?: boolean; yAxis?: boolean }` | `{ xAxis: false, yAxis: false }` |
 | `onZoomSelect` _(optional)_ | Callback to be called when a zoom selection is made if enabled.               | `(e: ZoomSelectionEvent) => void`      |                                  |
-| `groupId` _(optional)_      | Charts with the same `groupId` will have their tooltips synced across charts. | `string`                               |                                  |
+| `zoomSelect` _(optional)_   | Enable zoom select (click and drag area selection) for either axis.           | `{ xAxis?: boolean; yAxis?: boolean }` | `{ xAxis: false, yAxis: false }` |
```
shaneeza, line 92: Should we be more explicit that`onZoomSelect` will do nothing if nothing is passed to this prop?

tsck: That was my intention when I noted "if enabled" in the `onZoomSelect` description. Does that seem like not enough?

I see this like `Code`'s `copyable` and `onCopy`. Someone could add an `onCopy` handler, and if they never enable `copyable` it won't do anything https://github.com/mongodb/leafygreen-ui/blob/main/packages/code/README.md

shaneeza: I think if I didn't have any context, I would not know right away what "if enabled" meant. Maybe something like:

> "Callback to be called when a zoom selection is made if `zoomSelect` is enabled"

tsck: Updated! That does seem clearer. I think my lack of concern was that like in the `Code` docs, we don't mention anything about it. To me it seems inferable, but I might be too close to it so this is a good point.

PR #2582, opened 2024-12-09
`Table` documentation

shaneeza:
## ✍️ Proposed changes

### For new components


packages/table/README.md
bruugey, line 1: You can delete the contents of this file and replace with DOCUMENTATION.md - I know that it breaks our pattern with read mes but I don't think that we need multiple markdown files and you've accomplished what's necessary in the other file (so just move the contents here)

packages/table/OLD_README.md
bruugey, line 1: Let's rename this v11ReadMe or something similar

bruugey: we now have two of these files, we can rm one

bruugey (approved):
Once you remove the dup'd file, this looks good to me.
Again, we can add examples slowly over time but I do not want to to block on this.

shaneeza:
> Once you remove the dup'd file, this looks good to me. Again, we can add examples slowly over time but I do not want to to block on this.

Once the migration guide is done, I'll merge this in

bruugey (approved):

PR #2520, opened 2024-10-25
LG-4121: popover refactor

stephl3:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4121](https://jira.mongodb.org/browse/LG-4121)

### For bug fixes, new features & breaking changes

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `yarn changeset` and documented my changes

## 🧪 How to test changes

- [x] most testing in PR https://github.com/mongodb/leafygreen-ui/pull/2501
- [x] [bug bash](https://docs.google.com/spreadsheets/d/1lcbERGltSfJzK9ERK_2ikqUUaO8PBR5TuJGLIhdMnNQ/edit?gid=0#gid=0)

PR #2596, opened 2024-12-13
LG-4714: Fix using `SplitButton` as a managed component

kraenhansen:
## ✍️ Proposed changes

I propose removing a check for the `open` prop (`controlledOpen` internally) of the `SplitButton`'s internal `Menu` component to unconditionally call any `setOpen` function passed via props.

I've also included a few (hopefully not too controversial) improvements to the tests, now I was there. Let me know if I should split those into a separate PR 👍

🎟 _Jira ticket:_ [LG-4714](https://jira.mongodb.org/browse/LG-4714)

## 🧪 How to test changes

I've added a failing test which pass after applying the fix.

### DIFF
--- a/packages/split-button/src/Menu/Menu.tsx
+++ b/packages/split-button/src/Menu/Menu.tsx
@@ -58,9 +58,7 @@ export const Menu = ({
 
   const handleTriggerClick: MouseEventHandler = e => {
     onTriggerClick?.(e);
-    if (typeof controlledOpen !== 'boolean') {
-      setOpen(o => !o);
-    }
+    setOpen(o => !o);
   };
 
   const handleClose = useCallback(() => {
--- a/packages/split-button/src/SplitButton/SplitButton.spec.tsx
+++ b/packages/split-button/src/SplitButton/SplitButton.spec.tsx
@@ -209,7 +209,7 @@ describe('packages/split-button', () => {
       document.body.appendChild(portalContainer);
       const portalRef = createRef<HTMLElement>();
       renderSplitButton({
-        open,
+        open: true,
         portalContainer,
         portalRef,
         renderMode: RenderMode.Portal,
@@ -255,9 +255,9 @@ describe('packages/split-button', () => {
         await waitForElementToBeRemoved(menuEl);
         expect(menuEl).not.toBeInTheDocument();
       });
-      test('Returns focus to trigger {usePortal: true}', async () => {
+      test('Returns focus to trigger {renderMode: "portal"}', async () => {
         const { openMenu, menuTrigger } = renderSplitButton({
-          usePortal: true,
+          renderMode: "portal"
         });
         const { menuEl } = await openMenu();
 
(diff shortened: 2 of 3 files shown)


packages/split-button/src/Menu/Menu.tsx
```diff
@@ -58,9 +58,7 @@ export const Menu = ({
 
   const handleTriggerClick: MouseEventHandler = e => {
     onTriggerClick?.(e);
-    if (typeof controlledOpen !== 'boolean') {
-      setOpen(o => !o);
-    }
+    setOpen(o => !o);
```
kraenhansen, line 61: @shaneeza do you remember the context for why this was conditional on `controlledOpen` when adding it in https://github.com/mongodb/leafygreen-ui/commit/586703fc4f3ed05b400b29ccf53c99a24ce6dd93?

packages/split-button/src/SplitButton/SplitButton.spec.tsx
```diff
@@ -205,22 +209,32 @@ describe('packages/split-button', () => {
       document.body.appendChild(portalContainer);
       const portalRef = createRef<HTMLElement>();
       renderSplitButton({
-        open,
+        open: true,
```
kraenhansen, line 212: This was passing a global `open` function as prop, which was only revealed when restricting type of the argument passed to `renderSplitButton`.

packages/split-button/src/SplitButton/SplitButton.spec.tsx
```diff
@@ -251,9 +265,9 @@ describe('packages/split-button', () => {
         await waitForElementToBeRemoved(menuEl);
         expect(menuEl).not.toBeInTheDocument();
       });
-      test('Returns focus to trigger {usePortal: true}', async () => {
+      test('Returns focus to trigger {renderMode: "portal"}', async () => {
         const { openMenu, menuTrigger } = renderSplitButton({
-          usePortal: true,
+          renderMode: "portal"
```
kraenhansen, line 270: Fix to account for the breaking change introduced in https://github.com/mongodb/leafygreen-ui/pull/2520

kraenhansen:
I've just rebased the branch. The previous failure was from a missing token, most likely because I'm considered an outside contributor and my runs on GHA aren't allowed to run with secrets injected.

packages/split-button/src/Menu/Menu.tsx
shaneeza: Hey, I am trying to remember why this was conditional. For clarity, because of this check, your controlled `setOpen` was not being called. Is that correct?

Taking a look at [Menu](https://github.com/mongodb/leafygreen-ui/blob/main/packages/menu/src/Menu/Menu.tsx#L248), we don't include this check, so there is a chance we probably don't need this check.

kraenhansen: > because of this check, your controlled setOpen was not being called. Is that correct?

Correct.

shaneeza:
> I've just rebased the branch. The CI failure is from a missing token, since I'm considered an outside contributor and my runs on GHA aren't allowed to run with secrets injected.

That's correct. I'll make a new branch and pull in your changes there.

PR #2600, opened 2024-12-17
Allow defaultChecked to be used in Checkbox

ilai2:
## ✍️ Proposed changes

Allows Checkbox to default to true when uncontrolled. Currently, if the checkbox is uncontrolled, it only supports starting in a false state. However, there are cases where we want to have the checkbox be uncontrolled, but start in a true state.

In essence, this would support usage of the [defaultChecked](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/defaultChecked) property!

Let me know if there's anything I can do!! Also, feel free to close if it's not something that's right for leafy green :D it would just be a really useful feature for us on the MANA team!


.changeset/eleven-monkeys-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/checkbox': patch
```
TheSonOfThomp, line 2: I think this would be a `minor` change, since it adds a feature

packages/checkbox/src/Checkbox/Checkbox.tsx
```diff
@@ -50,6 +50,7 @@ function Checkbox({
   id: idProp,
   indeterminate: indeterminateProp,
   label = '',
+  defaultChecked = false,
```
TheSonOfThomp, line 53: I believe other components use `initialValue` (e.g. Combobox). Unsure whether consistency with other LG components or with HTML checkbox api is more valuable

.changeset/eleven-monkeys-teach.md
ilai2: Okay!! Should I change it manually to `minor` in this file or should I run the script again to change it?

packages/checkbox/src/Checkbox/Checkbox.tsx
zhaodext: The `Text Area` uses [defaultValue](https://github.com/mongodb/leafygreen-ui/blob/main/packages/text-area/src/TextArea/TextArea.tsx#L64). I would vote to use `defaultValue` across all components to meet HTML naming rule.

ilai2: I don't have a strong opinion; I'll defer to @TheSonOfThomp since you regularly work in this repo.

.changeset/eleven-monkeys-teach.md
TheSonOfThomp: @ilai2 You can update this file manually

packages/checkbox/src/Checkbox/Checkbox.tsx
TheSonOfThomp: Since Checkbox takes the prop `checked`, I'd say we can stick with `defaultChecked` 👍

TheSonOfThomp (approved):

PR #2601, opened 2024-12-17
LG-4721: Fix `SplitButton`'s menu is rendered in dark-mode regardless of value passed through `renderDarkMenu`

kraenhansen:
## ✍️ Proposed changes

The component's types declares that it takes a  `renderDarkMenu` prop, which isn't passed to the underlying `Menu` component. This makes it impossible for the consumer to force rendering a light menu in light-mode.

This PR suggests:
- Passing the prop to the `Menu`.
- Updating the storybook to reflect this.

🎟 _Jira ticket:_ [LG-4721](https://jira.mongodb.org/browse/LG-4721)

### NOTE: Avoiding future regressions

To avoid future regressions like this, I suggest using a `Pick` instead of an `Omit` (i.e. allowlist instead of denylist) in the declaration of the `SelectedMenuProps` type:

https://github.com/mongodb/leafygreen-ui/blob/3e821156e82f2ee6a486bb96174208501d3df767/packages/split-button/src/SplitButton/SplitButton.types.ts#L56-L68

I've provided a PR for that with #2602 and I didn't include it here since it's a breaking change in the API of the `SplitButton`.

### For bug fixes, new features & breaking changes

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] I have run `yarn changeset` and documented my changes

### For new components

- [ ] I have added my new package to the global tsconfig
- [ ] I have added my new package to the Table of Contents on the global README
- [ ] I have verified the Live Example looks as intended on the design website.

## 🧪 How to test changes

- Open the storybook UI and navigate to the `SplitButton` page.
- Toggle off the `darkMode` prop.
- Toggle off the `renderDarkMenu` prop.
- Click the right glyph on the button to have the menu drop down.
- See how the menu is rendered in light mode as expected.

PR #2602, opened 2024-12-17
Refactored `SplitButton` (breaking)

kraenhansen:
## ✍️ Proposed changes

As per my suggestion in https://github.com/mongodb/leafygreen-ui/pull/2601, future regressions could be avoided if the types derived from `MenuProps` intended for pass-through is allow-listed via a `Pick` instead of deny-listed via `Omit`.

This change is breaking, since it's removing a lot of props from the `SplitButton` which were never actually passed through to the `Menu`.

~This also optimize the internals by passing `darkMode` directly to child components instead of wrapping them in a new `LeafyGreenProvider` - this is not breaking, just a "drive-by" optimization.~

NOTE: If https://github.com/mongodb/leafygreen-ui/pull/2601 merge before this PR, it needs to be updated to add `renderDarkMenu` in the allow-list.

### For new components

## 🧪 How to test changes

- Add the `@leafygreen-ui/split-button` to a project
- Observe how the `onEnter` or `renderDarkMenu` props are available on the `SplitButton`.
- Apply this patch.
- Observe how the `onEnter` or `renderDarkMenu` props are no longer available on the `SplitButton`.

### DIFF
--- a/packages/split-button/src/SplitButton/SplitButton.types.ts
+++ b/packages/split-button/src/SplitButton/SplitButton.types.ts
@@ -53,18 +53,22 @@ type ButtonProps = Omit<
   'rightGlyph' | 'href' | 'as' | 'variant'
 >;
 
-export type SelectedMenuProps = Omit<
+export type SelectedMenuProps = Pick<
   ImportedMenuProps,
-  | 'children'
-  | 'trigger'
-  | 'shouldClose'
   | 'darkMode'
-  | 'onClick'
+  | 'maxHeight'
+  | 'adjustOnMutation'
+  | 'popoverZIndex'
+  | 'renderMode'
+  | 'portalClassName'
+  | 'portalContainer'
+  | 'portalRef'
+  | 'scrollContainer'
   | 'align'
   | 'justify'
-  | 'className'
-  | 'refEl'
-  | 'spacing'
+  | 'id'
+  | 'open'
+  | 'setOpen'
 >;
 
 export interface MenuProps extends SelectedMenuProps {
--- a/.changeset/chilled-hounds-laugh.md
+++ b/.changeset/chilled-hounds-laugh.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/split-button': major
+---
+
+Update types to reflect what is actually being passed through to the underlying `Menu`


packages/split-button/src/SplitButton/SplitButton.tsx
```diff
@@ -56,7 +54,7 @@ export const SplitButton = InferredPolymorphic<SplitButtonProps, 'button'>(
     ref: React.Ref<any>,
   ) => {
     const { Component } = useInferredPolymorphic(as, rest, 'button');
-    const { darkMode, theme } = useDarkMode(darkModeProp);
+    const { theme } = useDarkMode(darkMode);
```
shaneeza, line 57: We still want to use the `darkMode` value returned from `useDarkMode` because it reads and returns the `darkMode` value from the parent [LeafyGreenProvider](https://github.com/mongodb/leafygreen-ui/blob/main/packages/leafygreen-provider/src/DarkModeContext.tsx#L23-L28). If `darkMode` is passed into the component, it will take precedence over the `LeafyGreenProvider` `darkMode` value.

kraenhansen:
I've drafted this again, since there are more changes that I'd like to include: It isn't possible to pass menu separators through the `menuItems` prop.

packages/split-button/src/SplitButton/SplitButton.tsx
kraenhansen: > If `darkMode` is passed into the component, it will take precedence over the `LeafyGreenProvider` `darkMode` value.

That's the exact the behaviour I'd expect though?

I expect the `Button` and `Menu` (rendered internally in `SplitButton`) to call `useDarkMode` by them selves to apply the right theme from the provider iff a `darkMode` prop isn't passed to the `SplitButton`.

shaneeza: Thank you for pointing this out! You're absolutely correct.

However, we’ll stick with the current implementation, as we intentionally pass `darkMode` directly to the `LeafyGreenProvider`. This approach is deliberate and aligns with how we typically handle `darkMode` across other LeafyGreen components.

packages/split-button/src/SplitButton/SplitButton.types.ts
```diff
@@ -53,18 +53,22 @@ type ButtonProps = Omit<
   'rightGlyph' | 'href' | 'as' | 'variant'
 >;
 
-export type SelectedMenuProps = Omit<
+export type SelectedMenuProps = Pick<
```
shaneeza, line 56: I think this makes sense, this way we are explicit with what props menu eccepts

shaneeza (changes requested):
Thanks for the PR! The type updates make sense. However, we wouldn't want to remove the `LeafyGreenProvider` since passing `darkMode` to the `LeafyGreenProvider` is an explicit pattern we've adopted in other LG components.

kraenhansen:
I've created one more PR with suggestioned breaking changes to the `SplitButton` API: https://github.com/mongodb/leafygreen-ui/pull/2604 - if those are accepted it would be ideal if they get released together, to avoid two major version bumps right after each other.

shaneeza (approved):

kraenhansen:
Closing this PR in favor of https://github.com/mongodb/leafygreen-ui/pull/2618 (based on a branch in the original repo instead of my fork) + I added the `renderDarkMenu` prop from https://github.com/mongodb/leafygreen-ui/pull/2614.

PR #2604, opened 2024-12-17
Refactored SplitButton (breaking - more)

kraenhansen:
I'm working on a feature in Compass where I'd like to pass arbitrary elements to be rendered in the `SplitButton` menu, ex:
- `MenuSeparator` elements.
- `MenuItem` elements with controls that shouldn't cause the menu to close when clicked.

## ✍️ Proposed changes

1. The types of the `SplitButton` currently declare taking `children` (from the `ButtonProps`) but these are never used internally. I suggest that these are instead passed into the `Menu`. This will make the `menuItems` prop obsolete.
2. To make the behavior of clicking a `MenuItem` configurable, I suggest adding an `onItemClick` prop taking a callback which will be called (just like `onChange` is called now) but also passed a second options object with a reference to the React `element` being clicked (to allow inspection of props) as well as a `close` function, which can be called to close the menu.
3. To avoid breaking the existing behavior of the menu closing when an item is clicked, I suggest a "default" implementation of the `onItemClick` used when no function is explicitly passed. Since the signature and semantics of this new `onItemClick` prop is so similar to `onChange`, I suggest removing the latter. The main caveat with this is that someone renaming their `onChange` callback to `onItemClick` now also has to explicitly call the `close` function in the callback.

NOTE: This builds upon https://github.com/mongodb/leafygreen-ui/pull/2602 and needs to be rebased once that gets merged. Once rebased, I'll update the changeset description, including examples of how to migrate to the two new props (`menuItems` → `children` and `onChange` → `onItemClick`).

### For new components

### DIFF
--- a/packages/split-button/src/SplitButton/SplitButton.tsx
+++ b/packages/split-button/src/SplitButton/SplitButton.tsx
@@ -14,7 +14,7 @@ import {
 import { RenderMode } from '@leafygreen-ui/popover';
 import { BaseFontSize } from '@leafygreen-ui/tokens';
 
-import { Menu } from '../Menu';
+import { Menu, defaultItemClick } from '../Menu';
 
 import {
   buttonBaseStyles,
@@ -26,6 +26,7 @@ import { Align, Justify, SplitButtonProps, Variant } from './SplitButton.types';
 export const SplitButton = InferredPolymorphic<SplitButtonProps, 'button'>(
   (
     {
+      children,
       darkMode: darkModeProp,
       variant = Variant.Default,
       type = 'button',
@@ -33,7 +34,6 @@ export const SplitButton = InferredPolymorphic<SplitButtonProps, 'button'>(
       justify = Justify.End,
       size = Size.Default,
       disabled = false,
-      menuItems = [],
       as,
       baseFontSize,
       label,
@@ -49,8 +49,8 @@ export const SplitButton = InferredPolymorphic<SplitButtonProps, 'button'>(
       open,
       setOpen,
       onTriggerClick,
+      onItemClick = defaultItemClick,
       triggerAriaLabel,
-      onChange,
       ...rest
     },
     ref: React.Ref<any>,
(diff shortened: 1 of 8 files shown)


packages/split-button/src/SplitButton/SplitButton.tsx
```diff
@@ -26,14 +26,14 @@ import { Align, Justify, SplitButtonProps, Variant } from './SplitButton.types';
 export const SplitButton = InferredPolymorphic<SplitButtonProps, 'button'>(
   (
     {
+      children,
```
TheSonOfThomp, line 29: I believe we decided on the `menuItems` API to avoid confusion on whether the children should be the menu items or the button label

kraenhansen: > avoid confusion on whether the children should be the menu items or the button label

That is easily expressed through a comment on the prop :) Right now the component takes `children` and just throws them away? I feel that's more confusing 🤷 and much less expressive.

kraenhansen: I've updated the description above with more context on the issues I see with of the current `menuItems` prop.

TheSonOfThomp: @kraenhansen  To make this change we'd have to provide a codemod that would be pretty difficult to write.
I'd propose a compromise approach where we eliminate the check for `MenuItem` and allow any `ReactNode` in the `menuItems` prop

kraenhansen: As mentioned in the description, the `children` declared by the component isn't actually being used internally (because [another `children` is provided to the `Button`](https://github.com/mongodb/leafygreen-ui/blob/94e895ae758e11b1fb15296b7cf31df66f7270c4/packages/split-button/src/SplitButton/SplitButton.tsx#L96)), so repurposing that won't realistically break any existing code if:
- `menuItems` could gets annotated as `@deprecated` and declared optional.
- `children` prop defaults to the value of the `menuItems` prop and log a `console.warn`, if provided.
- providing both `menuItems` and `children` would throw.

PR #2608, opened 2024-12-18
adam/pnpm 12 17

TheSonOfThomp:


tools/storybook-addon/src/main/index.ts
```diff
@@ -72,15 +56,36 @@ export const webpackFinal: StorybookConfig['webpackFinal'] = config => {
...
 
   config.module.rules.push({
-    test: '/\.svg$/',
+    test: /\.svg$/,
     enforce: 'pre',
     loader: require.resolve('@svgr/webpack'),
   });
 
+  config.module.rules.push({
+    test: /\.(ts|tsx)$/,
```
TheSonOfThomp, line 69: For whatever reason, the default storybook webpack config (with an explicit `babel` export) doesn't apply to `.ts` files

PR #2609, opened 2024-12-18
LG-4724: Export prop types which are already wrapped in the polymorphic types

kraenhansen:
## ✍️ Proposed changes

I went through the codebase to find uses of the "polymorphic" types and replicated the pattern already established for the `SubMenu` component:
- Prepended the `*Props` type name with "Internal".
- Updated the internal use of ☝️ to match the new name.
- Added a new type in-place of the type which was previously exported, wrapping the internal type in either `PolymorphicProps` or `InferredPolymorphicProps` (depending on which was used by the implementation of the component).

I'm marking this as a patch, since it's extending the existing prop types and therefore shouldn't cause downstream breakage. If a consumer is wrapping these types already this should be a no-op.

🎟 _Jira ticket:_ [LG-4724](https://jira.mongodb.org/browse/LG-4724)

### For new components

### DIFF
--- a/.changeset/grumpy-moose-exercise.md
+++ b/.changeset/grumpy-moose-exercise.md
@@ -1,0 +1,9 @@
+---
+'@leafygreen-ui/input-option': patch
+'@leafygreen-ui/search-input': patch
+'@leafygreen-ui/split-button': patch
+'@leafygreen-ui/card': patch
+'@leafygreen-ui/menu': patch
+---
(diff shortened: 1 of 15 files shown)


kraenhansen:
Moved this to a branch on the repo to run the Chromatic workflow 🤞 https://github.com/mongodb/leafygreen-ui/pull/2613

PR #2611, opened 2024-12-18
LG-4662: Fix bundle size + other charts improvements

tsck:
## ✍️ Proposed changes

- Adds treeshaking of echarts to minimize bundle size.
- Improves echart initialization logic to better handle race conditions. This appears to fix previously seen `offsetWidth` bug.
- Separates echarts logic from our business logic with the creation of `useEchart` hook.
- Passes `ChartInstance` in full via the `ChartProvider` instead of individual methods.

🎟 _Jira tickets:_ 
- [LG-4662: Investigate / fix import issue](https://jira.mongodb.org/browse/LG-4662)
- [LG-4699: Fix `offsetWidth` error](https://jira.mongodb.org/browse/LG-4699)
- [LG-4702: Refactor `useChart`](https://jira.mongodb.org/browse/LG-4702)

### For new components

## 🧪 How to test changes
- All tests are passing
- Test in storybook


charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,389 @@
...
+    fn: T,
+  ) => {
+    return (...args: Parameters<T>) => {
+      if (!echartsInstance) {
+        console.error('Echart instance not initialized');
+        return;
+      }
+      fn(...args);
+    };
+  };
```
tsck, line 106: I went back and forth on whether this should actually wait for an instance. In theory, we could store a promise in a ref, and make all of the methods asynchronous. I ultimately decided not to and instead essentially do nothing if the instance doesn't exist yet. But would be curious to here other thoughts.

charts/core/src/Grid/Grid.tsx
```diff
@@ -15,10 +15,12 @@ const unsetGridOptions = {
 };
 
 export function Grid({ horizontal = true, vertical = true }: GridProps) {
-  const { updateChartOptions } = useChartContext();
+  const { chart } = useChartContext();
   const { theme } = useDarkMode();
 
   useEffect(() => {
+    if (!chart.ready) return;
```
TheSonOfThomp, line 22: I might be misunderstanding how echarts works, but why can't we check if `chart` is defined?

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+  EChartsInstance,
+  ZoomSelectionEvent,
+} from './Echart.types';
+
+// Singleton promise for initialization to prevent duplication
+let initPromise: Promise<void> | null = null;
+let echartsCore: any;
+let echartsCharts: any;
+let echartsRenders: any;
+let echartsComponents: any;
```
TheSonOfThomp, line 19: I believe we can type these as `Module`

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+      }
+    })();
+
+    return () => {
+      if (echartsInstance) {
+        window.removeEventListener('resize', resizeHandler);
+        echartsInstance.dispose();
+      }
+    };
+  }, [container]);
```
TheSonOfThomp, line 340: Can we add some comments on what these `useEffects` are doing? An IIFE inside an effect callback is definitely an interesting pattern 😅

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+} from './Echart.types';
+
+// Singleton promise for initialization to prevent duplication
+let initPromise: Promise<void> | null = null;
+let echartsCore: any;
+let echartsCharts: any;
+let echartsRenders: any;
+let echartsComponents: any;
+
+async function initializeEcharts() {
```
TheSonOfThomp, line 21: Can we add TSDoc for this func?

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+ * Wrapper around the ECharts library. Instantiates an ECharts instance.
+ * Provides helper methods to hide ECharts specific logic and give a cleaner API
+ * for interacting with a chart.
+ */
+export function useEchart({
+  container,
+  initialOptions,
+  theme,
+}: EChartHookProps): EChartsInstance {
+  const [echartsInstance, setEchartsInstance] = useState<any>(null); // has to be any since no types exist until import
```
TheSonOfThomp, line 83: might be worth maintaining our own copy of the correct type

charts/core/src/Echart/useEchart.ts
TheSonOfThomp: I'm a little confused why this is necessary. Is it just for the error logging? 
Unless I'm missing something, the optional chain check (e.g. L114) should catch if the `echartsInstance` singleton is undefined/null

TheSonOfThomp: If this were to wait for the instance, that would mean all the instance methods (like `addSeries`) would all be async?

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+        });
+        activeHandlers.current.clear();
+      }
+    };
+  }, [echartsInstance]);
+
+  return {
+    _echartsInstance: echartsInstance,
+    ready,
+    options,
```
TheSonOfThomp, line 375: remind me, why do we need to keep track of `options` in state if we can access them via `echartsInstance.options`?

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+ */
+export function useEchart({
+  container,
+  initialOptions,
+  theme,
+}: EChartHookProps): EChartsInstance {
+  const [echartsInstance, setEchartsInstance] = useState<any>(null); // has to be any since no types exist until import
+  const [error, setError] = useState<EChartsInstance['error']>(null);
+  const [ready, setReady] = useState<EChartsInstance['ready']>(false);
+  const [options, setOptions] = useState<EChartsInstance['options']>(
```
TheSonOfThomp, line 86: remind me, why do we need to keep track of options in state if we can access them via echartsInstance.options?

TheSonOfThomp: If we want it more easily accessible from this hook's return object, could we just have
```ts
const options = useMemo(() => echartsInstance.options, [...]) // or whatever the correct property is?
```

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+
+  // ECharts does not automatically resize when the window resizes.
+  const resizeHandler = withInstanceCheck(() => {
+    echartsInstance.resize();
+  });
+
+  const addToGroup: EChartsInstance['addToGroup'] = useCallback(
+    withInstanceCheck(groupId => {
+      echartsInstance.group = groupId;
+      echartsCore.connect(groupId);
```
TheSonOfThomp, line 165: should this check for existence of `echartsCore`?

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+        echartsComponents.DataZoomInsideComponent,
+      ]);
+    } catch (error) {
+      // Ensure we clear the promise if initialization fails
+      initPromise = null;
+      throw error;
+    }
+  })();
+
+  return initPromise;
```
TheSonOfThomp, line 70: Should this return the modules themselves?

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+
+        await initializeEcharts();
+
+        if (container) {
+          const newChart = echartsCore.init(container);
+          newChart.setOption(options);
+          window.addEventListener('resize', resizeHandler);
+          setEchartsInstance(newChart);
+          setReady(true);
+        }
```
TheSonOfThomp, line 322: The functional programmer in me feels like `initializeEcharts` should return the modules, and then instead of module-level singletons, we store the modules as `refs` within the hook itself:

```ts
initializeEcharts.then(({ echartsCore }) => {
  echartsCoreModuleRef.current = echartsCore;
  ...

}).catch((err) => {
  // secondary benefit is that we can avoid the try/catch
})
```

charts/core/src/Echart/useEchart.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+// Mock echarts instance creation with all required methods
+const mockEchartsInstance = {
+  setOption: jest.fn(),
+  dispose: jest.fn(),
+  resize: jest.fn(),
+  on: jest.fn(),
+  off: jest.fn(),
+  dispatchAction: jest.fn(),
+  group: null,
+};
```
TheSonOfThomp, line 18: 🔥

charts/core/src/Echart/Echart.types.ts
```diff
@@ -0,0 +1,99 @@
+import { Theme } from '@leafygreen-ui/lib';
+
+import type { ChartOptions, SeriesOption } from '../Chart';
+
+type EChartsEvents =
```
TheSonOfThomp, line 5: Should this be a enum? Not necessary, just wanted to call it out

charts/core/src/Grid/Grid.tsx
tsck: `chart` in this case is an instance of `ChartInstance`. In theory, that's actually always defined even when the `echart` hasn't been initialized. We could check if `chart._echartsInstance` is defined, but I just added `chart.ready` as a helper boolean to avoid having to interact directly with the instance outside of `useEchart`. Does that make sense?

charts/core/src/ChartContext/ChartContext.types.ts
```diff
@@ -1,8 +1,5 @@
-import { ChartOptions, SeriesOption } from '../Chart/Chart.types';
+import { ChartInstance } from '../Chart/hooks/useChart.types';
 
 export interface ChartContextType {
-  chartOptions: any;
-  updateChartOptions: (newOptions: Partial<ChartOptions>) => void;
-  addChartSeries: (series: SeriesOption) => void;
-  removeChartSeries: (name: string) => void;
+  chart: ChartInstance;
```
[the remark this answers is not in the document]
tsck: `ChartInstance`, the object with the instance methods, I don't believe should be, but props like `ChartInstance._echartsInstance` could be `null`.

charts/core/src/Echart/Echart.types.ts
tsck: Yeah, why not. Will update (probably with the help of my copilot 😄)

charts/core/src/Echart/useEchart.ts
tsck: I did look into the echarts code for the type, but it was very composed. I do think this makes sense though. I'll revisit and see if I can come up with something logical.

charts/core/src/Echart/useEchart.ts
tsck: I _think_ at one point early on I was tying into changes in order to trigger updates. Since this is no longer the case, I think this is a really good point. Will look into removing this.

charts/core/src/Echart/useEchart.ts
tsck: > Edit: I guess it's useful to fully abort if the instance isn't defined
That's correct. This is used as a higher order function to wrap any function that relies on the existence of an instance. This is an extraction of duplicate logic that was being used inline in each of those methods, attempting to handle the case where it doesn't exist yet gracefully. 

> If this were to wait for the instance, that would mean all the instance methods (like addSeries) would all be async?
That's correct. It would basically wrap the given function in a promise, converting it to an async function that's capable of being awaited, so instead of just (almost) silently ignoring this case, the execution wouldn't occur until the instance did exist.

charts/core/src/Echart/useEchart.ts
tsck: Okay I dug into this more and remember why we were holding a copy of the options. This is what made it possible to batch updates. Since we debounce calls to `echartsInstance.setOption()` via `setEchartOptions`, when additional updates occur in short order, they don't have an accurate copy of the current options since `echartsInstance.getOption()` is not necessarily up to date yet.

I know there was some initial disagreement regarding whether this batching was necessary. I tried to run some tests rendering a chart with 100 lines. With batching, the chart appeared after about 4000ms. Without batching, it didn't appear until after the profiler stopped recording which seemed to be around 6000ms. I'm not 100% sure what caused this but I am under the impression that each additional update is supposed to cause a repaint of the canvas, so it would be 1 repaint vs 100+ repaints. Though, I have no idea if that's what caused this, it could be something totally unrelated, at least withe current implementation it caused an inherent slowdown.

Happy to discuss further of course if there are other ideas, but just going to leave as is for now.

charts/core/src/Echart/useEchart.ts
```diff
@@ -0,0 +1,386 @@
...
+          if (action === 'zoomselect') {
+            echartsInstance.off('datazoom', handler);
+          } else {
+            echartsInstance.off(action, handler);
+          }
+        });
+        activeHandlers.current.clear();
+      }
+    };
+  }, [echartsInstance]);
```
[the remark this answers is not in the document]
tsck: Good call. Because all of the disposal stuff happens in that useEffect, this should all be there as well. I'll move this.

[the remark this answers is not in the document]
tsck: On second thought, since the instance is already being disposed, I'll get rid of this and just add `activeHandlers.current.clear();` to that cleanup callback.

charts/core/src/Echart/useEchart.ts
tsck: Fwiw, I don't think I like this. I think the interface of just exposing some way to determine readiness is fine.

charts/core/src/Echart/useEchart.ts
tsck: We could avoid the try/catch by just not using async/await here as well. I tend to be a fan of async/await over nesting as I just find it more readable, but it's just personal preference. 

I do agree about returning modules here though! However, really only the core library needs to be returned, as the remaining modules are only used internally. I'll refactor this and update it use chaining instead if that's our preference.

charts/core/src/Echart/useEchart.ts
tsck: Updated to return the core module since it's the only one used outside of this method.

charts/core/src/Echart/useEchart.ts
tsck: This didn't seem to work, but I've updated them to specifically be of type: `typeof import(...)` accordingly.

charts/core/src/Echart/useEchart.ts
tsck: Added the type. I wasn't thinking that importing it as a type should impact the bundle so we should be fine.

charts/core/src/Echart/useEchart.ts
TheSonOfThomp: async/await is fine. I was more focused on what's returned

TheSonOfThomp (approved):

PR #2622, opened 2025-01-07
Add missing eslint deps

TheSonOfThomp:


tools/lint/package.json
```diff
@@ -39,9 +43,5 @@
     "typescript": "~4.9.0"
   },
   "devDependencies": {
```
bruugey, line 45: should we rm this entirely then?

PR #2632, opened 2025-01-15
LG-4743: Fix resize bug in charts

tsck:
### Before
Resizing didn't trigger a resize of the chart because the echart instance was null from the callback.

### After
Creates the callback inside of the initialization hook to make sure it is accessible from the callback.

🎟 _Jira ticket:_ [LG-4743](https://jira.mongodb.org/browse/LG-4743)

### For new components

## 🧪 How to test changes
- Test in storybook by resizing the browser and confirming that the chart resizes correctly and that there are no errors in the console.

### DIFF
--- a/charts/core/src/Echart/useEchart.ts
+++ b/charts/core/src/Echart/useEchart.ts
@@ -100,10 +100,6 @@ export function useEchart({
     [setEchartOptions],
   );
 
-  const resizeEchartInstance = withInstanceCheck(() => {
-    echartsInstance?.resize();
-  });
-
   const addToGroup: EChartsInstance['addToGroup'] = useCallback(
     withInstanceCheck((groupId: string) => {
       // echartsCoreRef.current should exist if instance does, but checking for extra safety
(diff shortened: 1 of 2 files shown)


charts/core/src/Echart/useEchart.ts
```diff
@@ -260,23 +256,31 @@ export function useEchart({
...
         if (container) {
           // Init an echart instance
           const newChart = echartsCoreRef.current.init(container);
+
           // Set the initial options on the instance
           newChart.setOption(options);
+
           // Resize chart when window resizes because echarts don't be default
-          window.addEventListener('resize', resizeEchartInstance);
+          resizeCallback = () => {
```
TheSonOfThomp, line 280: do we need to do anything with `e: UIEvent` here?

TheSonOfThomp (approved):

charts/core/src/Echart/useEchart.ts
tsck: Nope, good call, I'll remove it from the type.

PR #2650, opened 2025-01-22
LG-4764: De-emphasize other series lines when one is hovered

tsck:
## ✍️ Proposed changes
- Adds config that causes other series lines to become opaque when a given series line is hovered over.
  - This differs from designs slightly. These were discussed and agreed upon with design. 
    - The given line doesn't widen on hover. In order to always show the tooltip, we have the `tooltip.trigger` set to `axis`. Due to this, the `emphasis` event on a series occurs when a user hovers anywhere on the chart. So setting the line to a larger width on emphasis makes it so that all lines become wider when the cursor enters the chart.
    - When a point marker is hovered, it doesn't emphasize the line below it. Currently marker points are not associated with a series. I added the option to associate them with a series, but this didn't make it so that the associated series was deemed emphasized on hover of the mark point either. This just appears to be an echarts limitation. Ultimately the plan is to highlight the point and associated line in the custom tooltip when that is built out. I've reverted the ability to associate a point with a line for now until that functionality is necessary.
- Drive-bys:
  - Improves the types of `EventMarkerPoint` and `EventMarkerLine` by distinguishing the position prop depending on which component it is.
  - Removes `actions` from storybook. This was a nice to have, but is causing errors since pnpm, I believe it might be a version mismatch with storybook. It didn't add a ton of value so I just removed it.

🎟 _Jira ticket:_ [LG-4764](https://jira.mongodb.org/browse/LG-4764)

### For new components

## 🧪 How to test changes
- Test in Storybook

## Screenshot
Screenshot 2025-01-22 at 9 57 18 AM

### DIFF
--- a/.changeset/rich-melons-destroy.md
+++ b/.changeset/rich-melons-destroy.md
@@ -2,4 +2,5 @@
 '@lg-charts/core': patch
 ---
 
-Fixes series emphasis on line hover by de-emphasizing other lines
+- Fixes series emphasis on line hover by de-emphasizing other lines.
+- Cleans up mark base component types internally so position type can be determined by type prop.
--- a/.changeset/rich-melons-destroy.md
+++ b/.changeset/rich-melons-destroy.md
@@ -1,0 +1,5 @@
+---
+'@lg-charts/core': patch
+---
+
+Fixes series emphasis on line hover by de-emphasizing other lines
--- a/charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
+++ b/charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
@@ -14,14 +14,12 @@ interface BaseEventMarkerProps {
   type: 'line' | 'point';
 }
 
-export interface BaseEventMarkerLineProps
-  extends Omit<BaseEventMarkerProps, 'name' | 'theme'> {
+export interface BaseEventMarkerLineProps extends BaseEventMarkerProps {
   type: 'line';
   position: string | number;
 }
 
-export interface BaseEventMarkerPointProps
-  extends Omit<BaseEventMarkerProps, 'name' | 'theme'> {
+export interface BaseEventMarkerPointProps extends BaseEventMarkerProps {
   type: 'point';
   position: [string | number, string | number];
 }
--- a/charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
+++ b/charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
@@ -5,7 +5,8 @@ export const EventLevel = {
   Info: 'info',
 } as const;
 export type EventLevel = (typeof EventLevel)[keyof typeof EventLevel];
-export interface BaseEventMarkerProps {
+
+interface BaseEventMarkerProps {
   label?: string;
   message: string;
   level: EventLevel;
--- a/charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
+++ b/charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
@@ -26,12 +26,6 @@ export interface BaseEventMarkerPointProps
   position: [string | number, string | number];
 }
 
-export interface EventMarkerLineProps
-  extends Omit<BaseEventMarkerLineProps, 'type'> {}
-
-export interface EventMarkerPointProps
-  extends Omit<BaseEventMarkerPointProps, 'type'> {}
-
 export interface GetMarkConfigProps extends BaseEventMarkerProps {
   name: string;
   theme: Theme;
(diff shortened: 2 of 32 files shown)


charts/core/src/EventMarkers/EventMarkerLine/EventMarkerLine.tsx
```diff
@@ -1,15 +1,17 @@
...
+  EventMarkerLineProps,
+} from '../BaseEventMarker';
 
 export function EventMarkerLine({
   position,
   label,
   message,
   level = EventLevel.Warning,
-}: EventMarkerLineProps) {
+}: Omit<EventMarkerLineProps, 'type'>) {
```
shaneeza, line 14: Why does `EventMarkerLineProps` include `type` if it needs to be omitted? Is it only needed for `BaseEventMarker`?

.changeset/rich-melons-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-charts/core': patch
+---
+
+Fixes series emphasis on line hover by de-emphasizing other lines
```
shaneeza, line 5: Can you also add the prop changes?

charts/core/src/Line/config/defaultLineOptions.ts
```diff
@@ -7,7 +7,12 @@ export const defaultLineOptions: Omit<SeriesOption, 'name' | 'data'> = {
   clip: false,
   symbolSize: 7,
   emphasis: {
-    disabled: true,
+    focus: 'series',
+  },
+  blur: {
```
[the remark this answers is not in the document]
tsck: I'm not sure there's a way to transition, but I actually don't know if a transition would actually help the blinking so much. I do think this is a good callout though and would be curious to get @sooachung's thoughts on this.

@sooachung Shaneeza raises a good point about how having this emphasis on the hovered line causes a great deal of blinking when the user hovers over the chart. Wanted to check in with you about this. (Example below)

https://github.com/user-attachments/assets/db66aa2f-a305-41c5-a6eb-e1b163da25ef

charts/core/src/EventMarkers/EventMarkerLine/EventMarkerLine.tsx
tsck: Actually, I think being more explicit like the example I gave makes more sense since we do ultimately export `EventMarkerLineProps` and `EventMarkerPointProps`. So without this, it's actually currently exporting the incorrect type. Will push a fix for this

.changeset/rich-melons-destroy.md
tsck: I didn't add that because it doesn't actually change what's exposed, it just cleans up the type internally for the base component so that `position` can be determined by `type`. But the exposed types should be the same. Is this something that we'd normally note here?

charts/core/src/Line/config/defaultLineOptions.ts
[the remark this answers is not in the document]
tsck: Sooa is checking with Visudha on this

charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
```diff
@@ -6,14 +6,26 @@ export const EventLevel = {
...
+interface BaseEventMarkerProps {
   label?: string;
   message: string;
   level: EventLevel;
   position: [string | number, string | number] | string | number;
   type: 'line' | 'point';
 }
 
+export interface BaseEventMarkerLineProps
+  extends Omit<BaseEventMarkerProps, 'name' | 'theme'> {
```
shaneeza, line 18: It doesn't look like `name` and `theme` are included in `BaseEventMarkerProps`

charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
```diff
@@ -6,14 +6,24 @@ export const EventLevel = {
...
-export interface BaseEventMarkerProps {
+interface BaseEventMarkerProps {
   label?: string;
   message: string;
   level: EventLevel;
   position: [string | number, string | number] | string | number;
   type: 'line' | 'point';
 }
 
+export interface BaseEventMarkerLineProps extends BaseEventMarkerProps {
```
shaneeza, line 17: In cases where you are extending another type but overriding some of the props from the extended type, do you need to omit those props? For example, sometimes I run into errors with a custom `onChange` prop if I don't omit it from the extended div props.
```
Omit<BaseEventMarkerProps, 'type' | 'position'>
```

By the way, this isn't blocking; I'm just curious.

shaneeza (approved):

charts/core/src/EventMarkers/BaseEventMarker/BaseEventMarker.types.ts
tsck: Good point! I believe it's not required here because we're narrowing down the extended type instead of completely overriding it. For example, if I wanted to set `label: number` on `BaseEventMarkerLineProps`, this wouldn't be allowed without omitting `label` first, since `number` wouldn't be compatible with `string` on the extended type.

PR #2654, opened 2025-01-22
Fix eslint error propagation

stephl3:
## ✍️ Proposed changes

The eslint portion of the lint script was updated during PNPM migration. Instead of returning a `Promise<boolean>` as it previously was doing, it was now returning a `Promise<Array<ESLint.LintResult>>`. The lint script was properly logging errors but silently failing and allowing errors to pass the lint step. This now correctly fails the lint step when there are any errors.

### For new components

### DIFF
--- a/.changeset/quiet-rules-obey.md
+++ b/.changeset/quiet-rules-obey.md
@@ -1,5 +1,5 @@
 ---
-'@lg-tools/lint': minor
+'@lg-tools/lint': patch
 ---
 
 Fix eslint so errors properly propagate during CI lint check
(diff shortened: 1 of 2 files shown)


.changeset/quiet-rules-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-tools/lint': minor
```
TheSonOfThomp, line 2: probably just a patch, since this isn't a user-facing api

tools/lint/src/eslint.ts
```diff
@@ -41,8 +43,12 @@ export async function eslint(
...
-  // eslint-disable-next-line no-console
   console.log(resultText);
 
-  return results;
+  const totalErrors = results.reduce(
+    (acc, result) => acc + result.errorCount,
+    0,
+  );
+
+  return totalErrors > 0;
```
TheSonOfThomp, line 53: nit: I'd turn this into a variable, and return that var

TheSonOfThomp (approved):

PR #2660, opened 2025-01-24
LG-4797: Fix charts not resizing when container dynamically changes

tsck:
## ✍️ Proposed changes
- Currently, resizing only works if the browser resizes. However, if the container gets dynamically resized, such as when a sidebar opens, it doesn't currently resize. This PR switches to instead using a `ResizeObserver` on the chart container to trigger resize.

🎟 _Jira ticket:_ [LG-4797](https://jira.mongodb.org/browse/LG-4797)

### For new components

## 🧪 How to test changes
- New story added for "Resizing with container" where you can dynamically resize the container and make sure the chart shrinks and expands correctly
- Can also test that browser resizing also still works on the live example

### DIFF
--- a/charts/core/src/Chart/hooks/useChart.ts
+++ b/charts/core/src/Chart/hooks/useChart.ts
@@ -117,22 +117,31 @@ export function useChart({
         return;
       }
 
-      echart.resize();
-
-      /**
-       * If the chart has been resized, the chart appears to reset zoom, which
-       * disables it. We need to re-enable it after the resize however, doing so
-       * immediately doesn't work. Tying into the 'rendered' and 'finished'
-       * events was attempted but doesn't work either. Manually waiting a bit
-       * does work however... I know, I know... I hate it too.
-       */
       if (zoomSelect?.xAxis || zoomSelect?.yAxis) {
-        setTimeout(() => {
-          echart.enableZoom();
-        }, 100);
+        /**
+         * If the chart has been resized, the chart appears to reset zoom, which
+         * disables it. We need to re-enable it after the resize however, doing so
+         * immediately doesn't work. To work around this, we listen for the `finished`
+         * event, which is triggered after the chart has been rendered, and then
+         * execute the re-enable zoom logic after all tasks on the queue have been
+         * processed.
+         *
+         * TODO(LG-4818): Investigate why this is necessary
+         */
+        function reEnableZoom() {
+          function reEnableZoomCallback() {
+            echart.enableZoom();
+            echart.off('finished', reEnableZoom);
+          }
+          setTimeout(reEnableZoomCallback, 0);
+        }
+
+        echart.on('finished', reEnableZoom);
       }
+
+      echart.resize();
     }
-  }, [echart.ready, initialRenderRef]);
+  }, [echart.ready, initialRenderRef, zoomSelect]);
 
   useEffect(() => {
     if (echart.ready && container) {
--- a/charts/core/src/Chart/hooks/useChart.ts
+++ b/charts/core/src/Chart/hooks/useChart.ts
@@ -1,10 +1,13 @@
-import { RefCallback, useCallback, useEffect, useState } from 'react';
+import { RefCallback, useCallback, useEffect, useRef, useState } from 'react';
 
 import { useEchart } from '../../Echart';
 import { EChartEvents } from '../../Echart';
 import { getDefaultChartOptions } from '../config';
 
 import type { ChartHookProps, ChartInstance } from './useChart.types';
+
+// TODO(LG-4803): Fix linting issues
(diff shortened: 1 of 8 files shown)


charts/core/src/Chart/hooks/useChart.ts
```diff
@@ -95,9 +101,45 @@ export function useChart({
...
+       * If the chart has been resized, the chart appears to reset zoom, which
+       * disables it. We need to re-enable it after the resize however, doing so
+       * immediately doesn't work. Tying into the 'rendered' and 'finished'
+       * events was attempted but doesn't work either. Manually waiting a bit
+       * does work however... I know, I know... I hate it too.
+       */
+      if (zoomSelect?.xAxis || zoomSelect?.yAxis) {
+        setTimeout(() => {
+          echart.enableZoom();
+        }, 100);
```
tsck, line 128: I hate everything about this. 100% open to suggestions.

download (1)

charts/core/src/Chart/hooks/useChart.ts
```diff
@@ -49,19 +50,26 @@ export function useChart({
...
     if (echart.ready) {
       echart.setupZoomSelect({
         xAxis: zoomSelect?.xAxis,
         yAxis: zoomSelect?.yAxis,
       });
+
+      if (zoomSelect?.xAxis || zoomSelect?.yAxis) {
+        function enableZoomOnRender() {
+          echart.enableZoom();
+          // Zooming triggers a render itself. This prevents an infinite loop.
```
shaneeza, line 66: You're saying that calling `echart.enableZoom();` on every render causes an infinite loop unless you call `echart?.off`?

charts/core/src/Chart/hooks/useChart.ts
shaneeza: lol. 

Is debounce a possibility? This is rough but maybe something like this?

```
const handleResize = (() => {
  const debounced = debounce(() => {
    console.log('Resize finished!');
    
    // Perform additional actions if needed
  }, 100); // Adjust debounce delay as needed

  return function () {
    console.log('Resizing...');
    myChart.resize(); // Adjust chart size
    debounced();
  };
})();
```

charts/core/src/Chart/hooks/useChart.ts
tsck: I'm saying that enabling zoom triggers a render, so once we enable it, we want to remove the handler or else there will be an infinite loop of render -> enable -> render -> etc.

I've updated the comment to try to make it more clear. Let me know if it isn't!

charts/core/src/Chart/hooks/useChart.ts
shaneeza: That makes sense. What are the 'rendered' and 'finished' events you mentioned in the comment?

shaneeza: I found it: https://echarts.apache.org/en/api.html#events.finished; where were you calling this?

tsck: Right here. Instead of using the timeout, I attempted to add a callback to those events that would turn on zoom. Then like on initial render, the callback needed to be removed after so that it wouldn't create an infinite loop. So it looked like:
```ts
if (zoomSelect?.xAxis || zoomSelect?.yAxis) {
      function enableZoomOnRender() {
        echart.enableZoom();
        echart?.off('finished', enableZoomOnRender);
      }

      echart?.on('rendered', enableZoomOnRender);
    }
```

I put this both before and after the call to `echart.resize()`, and with `finished` and `rendered`, all with no luck.

`setTimeout` with 100ms, seems to render 30-40 ms after the `finished` event does, which is apparently enough. I also tried to just use `setTimeout` with 0 ms, and the microtask queue, but both seem to get called too soon ☹️.

shaneeza: Hmm, tricky. Does `finished` need to be called outside of `handleResize`?

I also noticed that resize has an animation option. Is it possible to utilize `transitionend` or `animationend` if the animation uses CSS?

tsck: > Does finished need to be called outside of handleResize?

I don't believe so. It's successfully registered as a handler and the callback does get called, it just appears to have no impact on the chart. I assume it's a timing thing since setting the timeout `ms` arg to something like `0` also makes the call have no impact.

> Is it possible to utilize transitionend or animationend if the animation uses CSS?

We actually have animation turned completely off in echarts in an attempt to optimize performance. So there should be no transitions happening. Also, I believe that's the core difference between `rendered` and `finished` in that `finished` should be called after any animations, if animations were turned on.

Stupid echarts lol.

shaneeza (approved):

PR #2681, opened 2025-02-03
LG-4795: fix cell clipping hover and focus styles

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4795](https://jira.mongodb.org/browse/LG-4795)

### For new components

### DIFF
--- a/.changeset/olive-hats-invent.md
+++ b/.changeset/olive-hats-invent.md
@@ -1,0 +1,11 @@
+---
+'@leafygreen-ui/table': minor
+---
+
+Introduces a new prop, `overrideTruncation`, on `<Cell />` to allow overriding truncation for individual cells. When `shouldTruncate` is set to true on `<Table />`, `overflow: hidden` is applied to all cells, which can unintentionally clip hover and focus states on LeafyGreen components like `<Button />`. Setting `overrideTruncation` to true removes the overflow styles for that specific cell, ensuring interactive elements display correctly.
+
+
+e.g. 
+```js
+<Cell overrrideTruncation>{content}</Cell>
+```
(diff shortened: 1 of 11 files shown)


.changeset/olive-hats-invent.md
```diff
@@ -0,0 +1,11 @@
+---
+'@leafygreen-ui/table': minor
+---
+
+Introduces a new prop, `overrideTruncation`, on `<Cell />` to allow overriding truncation for individual cells. When `shouldTruncate` is set to true on `<Table />`, `overflow: hidden` is applied to all cells, which can unintentionally clip hover and focus states on LeafyGreen components like `<Button />`. Setting `overrideTruncation` to true removes the overflow styles for that specific cell, ensuring interactive elements display correctly.
```
shaneeza, line 5: I went with a prop that hopefully conveys that it will only turn off truncation for a cell if `shouldTruncate` is `true`.

I considered `overrideShouldTruncate` but i felt that it conveyed that prop this would always be the invert of `shouldTruncate`.

I'm open to other prop names or solutions.

shaneeza: This is so much better than my solution. Thank you 🙏🏽 And I also agree with your first comment; It's a code smell.

packages/table/src/Table.stories.tsx
```diff
@@ -95,7 +96,23 @@ const Template: StoryFn<StoryTableProps> = args => {
...
+                          {row[cellKey]}
+                        </Button>
+                      }
+                    >
+                      {"I'm leafy, you're leafy"}
+                    </Tooltip>
+                  ) : (
+                    row[cellKey]
+                  )}
+                </Cell>
```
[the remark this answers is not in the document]
shaneeza: Since there are three types of tables: basic, regular react-table, and virtual react-table, I wanted to make sure that it works for each of those.

PR #2695, opened 2025-02-05
LG-2719 | LG-3260: refactor Button to use InferredPolymorphic instead of Box

stephl3:
## ✍️ Proposed changes

- `Button` now uses `InferredPolymorphic` instead of `Box`
- disabled button no longer defaults to a `<button>`

🎟 _Jira ticket:_ [LG-2719](https://jira.mongodb.org/browse/LG-2719) | [LG-3260](https://jira.mongodb.org/browse/LG-3260)

## 🧪 How to test changes

- [ ] in `Button` live example, add `href` text and toggle on `disabled`. No browser error relating to `disabled` element incompatibility should appear

### DIFF
--- a/packages/polymorphic/src/InferredPolymorphic/InferredPolymorphic.types.ts
+++ b/packages/polymorphic/src/InferredPolymorphic/InferredPolymorphic.types.ts
@@ -61,7 +61,7 @@ export type InferredAnchorProps = {
  */
 export type InheritedExplicitAnchorLikeProps<TAsProp extends AnchorLike> = {
   /**
-   *
+   * The element or component to render as
    */
   as?: TAsProp;
 } & Omit<PartialRequired<ComponentPropsWithRef<TAsProp>, 'href'>, 'as'>;


.changeset/sweet-news-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/button': major
+---
+
+[LG-2719](https://jira.mongodb.org/browse/LG-2719): Replace `@leafygreen-ui/box` with `@leafygreen-ui/polymorphic` and refactor `Button` internals. This is a major change because previously, `Button` instances intended to be rendered as an `<a>` that were also `disabled` were forcibly being rendered as a `<button>`. They will now render as an `<a>`.
```
shaneeza, line 5: Do you know why we were forcibly rendering a `button`?

packages/button/src/Button/Button.tsx
```diff
@@ -15,74 +18,74 @@ import { getClassName } from './Button.styles';
...
-    type,
-    isLoading = false,
-    loadingIndicator,
-    loadingText,
-    ...rest
-  }: BoxProps<'button', ButtonProps>,
-  forwardRef,
-) {
-  const { darkMode } = useDarkMode(darkModeProp);
+export const Button = InferredPolymorphic<ButtonProps, 'button'>(
```
shaneeza, line 21: We should export a prop type that is also wrapped in InferredPolymorphic. Similar to [splitButton](https://github.com/mongodb/leafygreen-ui/blob/main/packages/split-button/src/SplitButton/SplitButton.types.ts#L146). This way, if consumers need to extend the type, it will include the inferred polymorphic props.

packages/button/src/Button/Button.tsx
```diff
@@ -15,74 +18,74 @@ import { getClassName } from './Button.styles';
...
+    const buttonProps = {
+      'data-lgid': dataLgId,
+      type: isAnchor ? undefined : type || 'button',
+      className: cx(ButtonClassName, buttonStyles, className),
+      ref: forwardRef,
+      'aria-disabled': !isInteractive,
+      onClick: isInteractive
+        ? onClick
+        : (e: React.MouseEvent) => e.preventDefault(),
+      href: isInteractive ? rest.href : undefined,
```
shaneeza, line 67: I think this href is being overwritten by the `...rest` below it. Do we need this line?

packages/code/src/CustomSelectMenuButton/CustomSelectMenuButton.tsx
```diff
@@ -8,9 +8,9 @@ import Button, { ButtonProps } from '@leafygreen-ui/button';
  * Passing down just the function which will be instantiated inside `Select`
  * @internal
  */
-export const CustomSelectMenuButton = React.forwardRef(
+export const CustomSelectMenuButton = React.forwardRef<HTMLButtonElement>(
   ({ children, ...props }: ButtonProps, ref) => (
-    <Button {...props} ref={ref}>
+    <Button {...props} as="button" ref={ref}>
```
shaneeza, line 13: curious: why do we need to pass the `as` prop if we are passing `button` here: `const { Component } = useInferredPolymorphic(as, rest, 'button');`

packages/polymorphic/src/InferredPolymorphic/InferredPolymorphic.types.ts
```diff
@@ -16,28 +16,53 @@ export type HrefLike = string | NodeUrlLike;
...
+   */
   as?: 'a';
 } & ComponentPropsWithRef<'a'>;
 
 /**
  * Union of {@link AnchorLikeProps} and {@link InheritedProps}
  */
 export type InheritedExplicitAnchorLikeProps<TAsProp extends AnchorLike> = {
+  /**
+   *
```
shaneeza, line 64: This is blank

packages/code/src/CustomSelectMenuButton/CustomSelectMenuButton.tsx
shaneeza: what was the issue?

shaneeza (approved):

PR #2720, opened 2025-02-12
LG-4830: add DrawerTabs component

stephl3:
## ✍️ Proposed changes

- adds `tabListContainerRef` and `tabPanelsContainerRef` props to `Tabs` component
- adds `DrawerTabs` component

🎟 _Jira ticket:_ [LG-4830](https://jira.mongodb.org/browse/LG-4830)

## 🧪 How to test changes

- [ ] review `Drawer/Scroll` story
- [ ] review `Drawer/WithTabs` story

### DIFF
--- a/packages/drawer/src/Drawer/Drawer.constants.ts
+++ b/packages/drawer/src/Drawer/Drawer.constants.ts
@@ -4,4 +4,5 @@ export const LGIDs = {
   root: LGID_ROOT,
 } as const;
 
+export const HEADER_HEIGHT = 48;
 export const PANEL_WIDTH = 432;
--- a/packages/drawer/src/Drawer/Drawer.styles.ts
+++ b/packages/drawer/src/Drawer/Drawer.styles.ts
@@ -5,7 +5,7 @@ import { Theme } from '@leafygreen-ui/lib';
 import { palette } from '@leafygreen-ui/palette';
 import { color, spacing, transitionDuration } from '@leafygreen-ui/tokens';
 
-import { PANEL_WIDTH } from './Drawer.constants';
+import { HEADER_HEIGHT, PANEL_WIDTH } from './Drawer.constants';
 
 const getBaseStyles = ({ open, theme }: { open: boolean; theme: Theme }) => css`
   height: 100%;
@@ -54,7 +54,7 @@ const getBaseHeaderStyles = ({
   hasTabs: boolean;
   theme: Theme;
 }) => css`
-  height: 48px;
+  height: ${HEADER_HEIGHT}px;
   padding: ${spacing[400]}px;
   display: flex;
   justify-content: space-between;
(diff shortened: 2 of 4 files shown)


packages/drawer/src/Drawer/Drawer.hooks.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  useEffect(() => {
+    const scrollContainerEl = scrollContainerRef.current;
+    if (!scrollContainerEl) return;
+
+    const handleScroll = () => {
+      setHasShadowTop(scrollContainerEl.scrollTop > 0);
+    };
+
+    scrollContainerEl.addEventListener('scroll', handleScroll);
```
shaneeza, line 34: nit: while working on table, I found that IntersectionObserver was helpful in avoiding frequent scroll event handling. Maybe it could be beneficial here?

packages/drawer/src/Drawer/Drawer.hooks.ts
```diff
@@ -0,0 +1,42 @@
...
+  const scrollContainerRef = useRef<HTMLDivElement>(null);
+  const [hasShadowTop, setHasShadowTop] = useState(false);
+
+  useEffect(() => {
+    const scrollContainerEl = scrollContainerRef.current;
+    if (!scrollContainerEl) return;
+
+    const handleScroll = () => {
+      setHasShadowTop(scrollContainerEl.scrollTop > 0);
+    };
```
tsck, line 32: Nit 1: Since `scrollContainerRef` is accessible outside of the `useEffect` callback, this function doesn't need to be recreated on every call and can be moved to the outer scope and only initialized once on mount using `useCallback`. It's so small I'm not worried about it so definitely feel free to leave as is if you think it improves readability.

Nit 2: (I think @shaneeza's [suggestion below](https://github.com/mongodb/leafygreen-ui/pull/2720/files#r1953527709) would solve this in a better way, but will still leave this in case you don't go in that direction) Currently this function is going to fire quite a bit on scroll. It's sometimes advisable to debounce scroll handlers to minimize this a bit.

packages/drawer/src/DrawerTabs/DrawerTabs.tsx
```diff
@@ -0,0 +1,40 @@
...
+    registerTabs();
+  }, [registerTabs]);
+
+  return (
+    <Tabs
+      aria-label="Drawer tabs"
+      {...rest}
+      className={getDrawerTabsStyles({ className, hasShadowTop, theme })}
+      size={Size.Small}
+      tabPanelsContainerRef={scrollContainerRef}
```
tsck, line 33: Nit: There might be some value adding a small comment here that this prop order is intentional and we're purposely overriding `size` and `tabPanelsContainerRef` if they're in `rest` so that future us don't come in and accidentally move `{...rest}` and break things.

tsck (approved):

shaneeza (commented):
Are you adding test utils for tabs?

PR #2726, opened 2025-02-14
Update button package test harnesses

stephl3:
## ✍️ Proposed changes

This PR is an initial stab at addressing 2 issues with our current test harnesses approach demonstrated in `@leafygreen-ui/button`.
1. Test utils cannot be used to [wait for appearance](https://testing-library.com/docs/guide-disappearance/#waiting-for-appearance) or [assert elements are not present](https://testing-library.com/docs/guide-disappearance/#asserting-elements-are-not-present)
  - Addressed by exposing `findButton` and `queryButton` test utils
2. `data-lgid` prop does not propagate to nested elements in the root component
  - Addressed by replacing the static `LGIDS_BUTTON` object (in other packages, this may be named `LGIDs`, `LGIDS`, or `LGIDS_*`) with a `getLgIds` util which can be used to propagate/interpolate the `data-lgid` prop into all the other keys
  - This package isn't the best example of this since we only use a `data-lgid` on the root. If you consider [select test utils](https://github.com/mongodb/leafygreen-ui/blob/main/packages/select/src/utils/getTestUtils/getTestUtils.ts), you can think about: how do you test for asserting a specific label or description is present when multiple `Select` instances are rendered?

## 🧪 How to test changes

- units tests

### DIFF
--- a/.changeset/late-chefs-build.md
+++ b/.changeset/late-chefs-build.md
@@ -2,4 +2,4 @@
 '@lg-tools/test-harnesses': minor
 ---
 
-Export `queryByLgId` util
+Export `queryByLgId` util in addition to the already exported `findByLgId` and `getByLgId` utils


.changeset/late-chefs-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-tools/test-harnesses': minor
+---
+
+Export `queryByLgId` util
```
TheSonOfThomp, line 5: Export `findByLgId`, `getByLgId`, `queryByLgId`

TheSonOfThomp (approved):

PR #2727, opened 2025-02-18
LG-4901: table column visibility

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4901](https://jira.mongodb.org/browse/LG-4901)

### For new components


packages/table/package.json
shaneeza, line 1: changeset

packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -80,7 +80,13 @@ export const MemoizedInternalRowWithRT = React.memo(
...
     const { children: nextChildren, ...restNextProps } = nextProps;
 
-    const propsAreEqual = isEqual(restPrevProps, restNextProps);
+    // This allows us to rerender if the child(cell) count changes. E.g. column visibility changes
+    const prevChildrenCount = React.Children.count(prevChildren);
+    const nextChildrenCount = React.Children.count(nextChildren);
+
+    const propsAreEqual =
+      isEqual(restPrevProps, restNextProps) &&
+      prevChildrenCount === nextChildrenCount;
```
TheSonOfThomp, line 89: Would checking `isEqual` on _all_ props work?

packages/table/src/Table/Table.spec.tsx
```diff
@@ -206,6 +224,37 @@ describe('packages/table/Table', () => {
     });
   });
 
+  describe('column visibility', () => {
+    test('renders the correct cells in a row by default', () => {
+      render(<TableWithHook hasToggleCheckboxes />);
```
TheSonOfThomp, line 229: I feel like adding these checkboxes to the test components overcomplicates the tests... But the alternative might be just as complex:

```ts
test('', () => {
  // Render the table as a test hook
  const tableResult = renderHook(useLeafyGreenTable);
  const table = tableResult.result.table;

  // render the table
  const renderResult = render(<Table table={table} .../>);

  // Manually trigger the visibility toggle
  const columnToHide = tableResult.result.getColumn('...');
  const columnToggleFn = columnToHide.getToggleVisibilityHandler();  
  act(() => {
    columnToggleFn();
  })

  // rerender the table
  renderResult.rerender(<Table table={table} ... />);

  expect(...) // same expect
})
```

packages/table/package.json
```diff
@@ -46,7 +46,8 @@
     "@leafygreen-ui/button": "workspace:^",
     "@leafygreen-ui/badge": "workspace:^",
     "@leafygreen-ui/pagination": "workspace:^",
-    "@leafygreen-ui/tooltip": "workspace:^"
+    "@leafygreen-ui/tooltip": "workspace:^",
+    "@emotion/styled": "^11.14.0"
```
TheSonOfThomp, line 50: What are we using styled for?

packages/table/src/Row/InternalRowWithRT.tsx
shaneeza: No, it doesn't.
```js
console.log({
  prevChildren,
  nextChildren,
  childrenAreEqual: isEqual(prevChildren, nextChildren), // false
  propsAreEqual: isEqual(restPrevProps, restNextProps), //true
  everythingIsEqual: isEqual(prevProps, nextProps), //false
});

const propsAreEqual = isEqual(prevProps, nextProps);
```
 [isEqual](https://commerce.nearform.com/blog/2018/introducing-react-fast-compare) should be doing a deep comparison check, so something must differ between the children.

packages/table/package.json
shaneeza: I'm using it in the spec file. I was getting an error locally about it not being included. However, I noticed that polymorphic also uses styled in the spec file, but it's not included in package.json. Maybe I don't need it.

packages/table/src/Table/Table.spec.tsx
shaneeza: Can you elaborate more on "overcomplicates the tests"?

TheSonOfThomp (approved):

PR #2728, opened 2025-02-18
LG-4798: code custom snips obj

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-4798](https://jira.mongodb.org/browse/LG-4798)

We had a request to change the color of certain words in the code editor. In particular, [this ticket](https://jira.mongodb.org/browse/CLOUDP-259897) wanted to highlight `password` and `username`. 

Some examples of this include:
image (3)

image (2)

I initially tried an approach of wrapping the word to highlight in a sequence of characters:
```js
In this line change the color of {{ password}} and {{ username }}
```
However, this approach was challenging because we support a wide range of languages, making it difficult to find a character sequence that wouldn’t interfere with language-specific special characters.

I decided to use an object-based approach. Consumers can pass in an object containing multiple keywords they want to customize. I'll search for each keyword and wrap it in a div with the specified className. However, if the keyword is a reserved word in that language, it will be ignored. The only drawback to this approach is that it will target every occurrence of the keyword.

```js
customKeywordObject: {{ 'keyword': 'className' }}
```

### For new components


packages/code/src/renderingPlugin/renderingPlugin.tsx
```diff
@@ -151,6 +154,87 @@ function getHighlightedRowStyle(darkMode: boolean) {
...
+
+      if (typeof child === 'string') {
+        if (typeof lastItem === 'string') {
+          acc[acc.length - 1] = lastItem + child;
+        } else {
+          acc.push(child);
+        }
+      } else {
+        acc.push(child);
+      }
```
tsck, line 180: I think this can be simplified slightly:


```suggestion
      if (typeof child === 'string' && typeof lastItem === 'string') {
        acc[acc.length - 1] = lastItem + child;
      } else {
        acc.push(child);
      }
```

packages/code/src/renderingPlugin/renderingPlugin.tsx
```diff
@@ -228,9 +312,11 @@ export function flattenNestedTree(
     parentKinds = parentKinds.filter(
       (str): str is string => isString(str) && str.length > 0,
     );
+
     return function (
       entity: string | TokenObject,
     ): string | FlatTokenObject | Array<string | FlatTokenObject> {
+      // console.log({ entity });
```
tsck, line 319: Accidentally left behind?
```suggestion
```

packages/code/src/renderingPlugin/renderingPlugin.tsx
```diff
@@ -151,6 +154,87 @@ function getHighlightedRowStyle(darkMode: boolean) {
...
+ * ```js
+ * keywords: { '_hello': 'custom' }
+ * ```
+ *
+ * ```js
+ * ['_hello world hi', {}, 'bye _hello'] => [{ kind: 'lg-highlight-custom', children: ['_hello'] }, 'world ', 'hi', {}, 'bye', { kind: 'lg-highlight-custom', children: ['_hello'] }]
+ * ```
+ */
+const lineWithKeywords = (line: Array<string | FlatTokenObject>) => {
+  const mergedLines = mergeStringsIntoString(line);
```
tsck, line 202: Q: What benefit do we get from merging the words together? Is there a functional reason this is required?

packages/code/src/renderingPlugin/renderingPlugin.tsx
```diff
@@ -151,6 +154,87 @@ function getHighlightedRowStyle(darkMode: boolean) {
...
+ *
+ * ```js
+ * keywords: { '_hello': 'custom' }
+ * ```
+ *
+ * ```js
+ * ['_hello world hi', {}, 'bye _hello'] => [{ kind: 'lg-highlight-custom', children: ['_hello'] }, 'world ', 'hi', {}, 'bye', { kind: 'lg-highlight-custom', children: ['_hello'] }]
+ * ```
+ */
+const lineWithKeywords = (line: Array<string | FlatTokenObject>) => {
```
tsck, line 201: Q: Do we / could we add some tests around this logic?

packages/code/src/Syntax/Syntax.types.ts
```diff
@@ -32,4 +32,11 @@ export interface SyntaxProps extends ComponentPropsWithoutRef<'code'> {
    * An array of lines to highlight. The array can only contain numbers corresponding to the line numbers to highlight, and / or tuples representing a range (e.g. `[6, 10]`);
    */
   highlightLines?: LineHighlightingDefinition;
+
+  /**
+   * Custom keywords to be highlighted in the code block. The key is the keyword to be highlighted, and the value is the classname to be applied to the keyword.
+   *
+   * E.g. `customKeywordObject: {{ 'keyword': 'className' }}`
+   */
+  customKeywordObject?: Record<string, string>;
```
tsck, line 41: Nit: I'm on the fence about having `Object` in this prop name. I think I'd prefer just `customKeywords` or `customKeywordConfig`. Not a hard stance and definitely not blocking, but curious to hear your thoughts.

tsck:
Should this be a minor bump version bump? Or are we only bumping the version on the integration branch?

packages/code/src/Syntax/Syntax.types.ts
shaneeza: I don't mind removing `Object`.

packages/code/src/renderingPlugin/renderingPlugin.tsx
shaneeza: Yeah, say, for example, you want to highlight `_hello`, I'm accounting for cases where the plugin might return this as: 
```js
['_', 'hello ', 'world ', 'hi', {}, 'bye ', '_', 'hello ']
```
Merging this into one string makes it easer to find `_hello`. The output of merging would be:
```js
['_hello world hi', {}, 'bye _hello']
```
So this is a just-in-case thing.

shaneeza:
> Should this be a minor bump version bump? Or are we only bumping the version on the integration branch?

I think I'll keep this independent from the integration branch and merge it after the integration branch merges.

packages/code/src/renderingPlugin/renderingPlugin.tsx
tsck: Gotcha, that makes sense!

tsck (approved):
Looks like there's a lint issue but otherwise LGTM!

tsck (approved):
LGTM!

PR #2732, opened 2025-02-21
Code test harness updates

shaneeza:
## ✍️ Proposed changes

Updates code test harnesses to utilize the updated button test harnesses #2726

### For new components


packages/code/README.md
```diff
@@ -230,28 +235,30 @@ test('code', () => {
...
 | `getLanguageSwitcherUtils()` | Returns utils for interacting with the language switcher | `LanguageSwitcherUtils`                                                                                                  |
 | `getIsLoading()`             | Returns whether the code block is in loading state       | `boolean`                                                                                                                |
 | `getCopyButtonUtils()`       | Returns utils for interacting with the copy button       | [Button test utils return type](https://github.com/mongodb/leafygreen-ui/blob/main/packages/button/README.md#test-utils) |
-| `getExpandButton()`          | Returns the expand button                                | `HTMLButtonElement`                                                                                                      |
+| `getExpandButton()`          | Returns utils for interacting with the expand button     | [Button test utils return type](https://github.com/mongodb/leafygreen-ui/blob/main/packages/button/README.md#test-utils) |
```
tsck, line 258:
```suggestion
| `getExpandButtonUtils()`          | Returns utils for interacting with the expand button     | [Button test utils return type](https://github.com/mongodb/leafygreen-ui/blob/main/packages/button/README.md#test-utils) |
```

packages/code/src/utils/getTestUtils/getTestUtils.ts
```diff
@@ -55,9 +55,14 @@ export const getTestUtils = (
...
-    const { getButton } = getButtonTestUtils(lgIds.expandButton);
-    return getButton() as HTMLButtonElement;
+  const getExpandButtonUtils = () => {
+    const { queryButton, getButton, findButton } =
+      getButtonTestUtils<HTMLButtonElement>(lgIds.expandButton);
+    return {
+      getButton: () => getButton(),
+      queryButton: () => queryButton(),
+      findButton: () => findButton(),
+    };
```
tsck, line 65: In theory, I don't think new functions need to be created here and the instances returned from `getButtonTestUtils` can just be returned directly
```
    return {
      getButton,
      queryButton,
      findButton,
    };
```
Or even
```
const getExpandButtonUtils = () => getButtonTestUtils<HTMLButtonElement>(lgIds.expandButton);
```
But the latter would currently also return `isDisabled` (and eventually all that are possible), so if those are just irrelevant, the former makes sense to me.

tsck (commented):
I assume the changeset will be added in the integration branch for this as well, but just want to double check again. Also it might still be valuable throwing a small description onto this PR for future context even though it's not going into `main`.

packages/code/src/utils/getTestUtils/getTestUtils.ts
shaneeza: Yeah, I purposely did this because this button will never be disabled

shaneeza:
> I assume the changeset will be added in the integration branch for this as well, but just want to double check again. Also it might still be valuable throwing a small description onto this PR for future context even though it's not going into `main`.

yeah its in the integration branch but I'll add a description in here.

packages/code/src/utils/getTestUtils/getTestUtils.ts
tsck: Makes sense! The new functions still don't need to be created though, as per the first suggestion. Not a blocker, so will leave it up to your discretion, just a bit cleaner imo.
```suggestion
    return {
      getButton,
      queryButton,
      findButton,
    };
```

tsck (approved):

packages/code/src/utils/getTestUtils/getTestUtils.ts
shaneeza: oh that makes sense!

PR #2733, opened 2025-02-25
LG-4905: displayMode prop

stephl3:
## ✍️ Proposed changes

### For new components


packages/drawer/src/Drawer/Drawer.types.ts
```diff
@@ -2,10 +2,29 @@ import React from 'react';
 
 import { DarkModeProps, HTMLElementProps, LgIdProps } from '@leafygreen-ui/lib';
 
+/**
+ * Options to control how the drawer element is displayed
+ * @param Overlay will display a drawer that takes up the full viewport height and elevated above main page content
+ * @param Persistent will display a drawer that takes up the full parent container height and on the same elevation as main page content. It is recommended to wrap a persistent drawer within the `PersistentDrawerLayout` container
```
TheSonOfThomp, line 8: I still think `inline` might be a better descriptor for this behavior

packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -10,66 +10,122 @@ import {
...
 `;
 
+const getOverlayStyles = ({ open, theme }: { open: boolean; theme: Theme }) =>
+  cx(
+    css`
+      position: fixed;
+      top: 0;
+      bottom: 0;
+      right: 0;
+      box-shadow: ${open && theme === Theme.Light
```
TheSonOfThomp, line 54: The shadow styles are disappearing when the drawer transitions out

packages/drawer/src/Drawer/Drawer.types.ts
TheSonOfThomp: > opened alongside the trigger element

That's super fair.

My concern with the name "persistent" is that whether the drawer actually "persists" is not on us, but rather controlled by the consumer.  (Though I may be over-thinking this 😅)

The behavior we control is that the drawer is _beside_ the content, and not above. I'm not opposed to `embedded`

packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -51,21 +62,19 @@ const getOverlayStyles = ({ open, theme }: { open: boolean; theme: Theme }) =>
...
       right: 0;
-      box-shadow: ${open && theme === Theme.Light
-        ? `-10px 0 10px -10px rgba(0,0,0,0.3)`
-        : 'initial'};
-      transition: transform ${openTransitionDuration}ms ease-in-out;
+      transition: transform ${drawerTransitionDuration}ms ease-in-out,
+        box-shadow ${drawerTransitionDuration}ms ease-in-out
+          ${open ? '0ms' : `${drawerTransitionDuration}ms`},
+        opacity ${drawerTransitionDuration}ms ease-in-out
+          ${open ? '0ms' : `${drawerTransitionDuration}ms`};
```
TheSonOfThomp, line 69: Should we be checking for  `prefers-reduced-motion` here?

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -44,7 +44,10 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
     const titleId = useIdAllocator({ prefix: 'drawer' });
 
     // Track when element is no longer visible to add shadow below drawer header
```
TheSonOfThomp, line 46: nit: what element

packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -28,18 +28,29 @@ const getBaseStyles = ({ theme }: { theme: Theme }) => css`
   }
 `;
 
-const overlayOpenStyles = css`
+const getOverlayOpenStyles = (theme: Theme) => css`
+  box-shadow: ${theme === Theme.Light
+    ? `-10px 0 10px -10px rgba(0, 0, 0, 0.3)`
```
TheSonOfThomp, line 33: Do we have tokens for this?

PR #2734, opened 2025-03-02
LG-4706: Improve Chromatic regression testing for charts

tsck:
## ✍️ Proposed changes
- Adds more granular test cases

🎟 _Jira ticket:_ [LG-4706](https://jira.mongodb.org/browse/LG-4706)

## 🧪 How to test changes
- View in storybook


charts/core/src/testUtils/makeLineData.testUtils.ts
tsck, line 1: This was updated by copilot to create consistent data for regression purposes

charts/core/src/Chart.stories.tsx
```diff
@@ -544,452 +572,494 @@ export const LiveExample: StoryObj<StorybookProps> = {
...
-    eventMarkerPointLevel,
-    eventMarkerPointXPosition,
-    eventMarkerPointYPosition,
-    renderThresholdLine,
-    thresholdLineLabel,
-    thresholdLineValue,
-    thresholdLinePosition,
-  }) => {
+/**
+ * Explicitly testing dark mode regression since generated isn't possible given composibility
```
shaneeza, line 576: I don't fully understand what composability has to do with darkMode. Can you clarify?

charts/core/src/testUtils/makeLineData.testUtils.ts
shaneeza: The entire file?

charts/core/src/Chart.stories.tsx
tsck: Yup. What I meant here is that in order to test dark mode generally we'd use a `Generate` story to test the variants of the component. But in the case of `Chart`, the `Chart` component without any children renders nothing, so it wouldn't be testing anything. I can set the `children` arg to the child components, but it just seemed easier to make an explicit `darkMode` story to make sure the dark mode styles are being tested properly.

charts/core/src/Chart.stories.tsx
```diff
@@ -544,452 +572,494 @@ export const LiveExample: StoryObj<StorybookProps> = {
...
-    eventMarkerPointLevel: { table: { disable: true } },
-    eventMarkerPointXPosition: { table: { disable: true } },
-    eventMarkerPointYPosition: { table: { disable: true } },
-    renderThresholdLine: { table: { disable: true } },
-    thresholdLineLabel: { table: { disable: true } },
-    thresholdLineValue: { table: { disable: true } },
-    thresholdLinePosition: { table: { disable: true } },
+};
+
+export const WithTooltip: StoryObj<{}> = {
```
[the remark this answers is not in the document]
tsck: This is a good point. The original intention was for `Tooltip` to utilize a `play` function to get the tooltip to always render. For some reason that wasn't working well though. I'm going to update `Basic` so that there's no tooltip rendered, since I think that is at least a semi unique.

charts/core/src/testUtils/makeLineData.testUtils.ts
tsck: Yup, all the updates

charts/core/src/Chart.stories.tsx
tsck: Oh smart, should be! Will update shortly

tsck: Oh, actually, I'm not sure we would want this. `LiveExample` intentionally has a lot of controls to hide and show all the things. `DarkMode` intentionally doesn't.

shaneeza: i see, thats fine!

charts/core/src/Chart.stories.tsx
```diff
@@ -544,452 +572,493 @@ export const LiveExample: StoryObj<StorybookProps> = {
...
+        />
+        {lineData.map(({ name, data }) => (
+          <Line name={name} data={data} key={name} />
+        ))}
+      </Chart>
+    );
+  },
+};
+
+export const WithZoom: StoryObj<{}> = {
```
shaneeza, line 996: what should this do in storybook? This story, WithXAxisZoom, and WithYAxisZoom all look the same

tsck: Yeah, this was another scenario where I was going to add a play function, but it wasn't interacting correctly with the chart. I left it because it is a "variant" in that zoom can be turned off and on, so I just left it. To your point, it adds nothing in terms of regression testing.

charts/core/src/Chart.stories.tsx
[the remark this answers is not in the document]
shaneeza: i would double check with brooke because i think we pay for a certain amount of stories, so if this isn't a helpful test then maybe we should remove it.

PR #2735, opened 2025-03-02
LG-4578: Add drag and drop functionality to charts

tsck:
## ✍️ Proposed changes
Adds the `@lg-charts/drag-provider` package to provide draggable functionality for `Chart` and `ChartCard` components. 

🎟 _Jira ticket:_ [LG-4578](https://jira.mongodb.org/browse/LG-4578)

### External Tooling
`@dnd-kit` - Used to provide the underlying drag functionality.

### Exposed
`<DragProvider>`
- Enables drag on any first descendant `Chart` or `ChartCard` with a `dragId` prop within the provider.
- Has the ability to be nested to provide nested draggable containers.
- Props:
  - `onDragEnd: ({ over: string; active: string }) => void`
    - `active` - `dragId` of the drag component being dragged.
    - `over` - `dragId` of the drag component the active component is over on `dragEnd`.

~~`<ChildStateUpdateProvider>`~~
- ~~Used inside of `<DragProvider>` and exposes `updateChildState()` to children~~
- ~~Provides:~~
  - ~~`updateChildState(dragId: string, stateUpdate: Object)`~~
    - ~~When the user drags a child of `<DragProvider>` , it creates a new component based on the child it receives and uses that as the “drag overlay” (component that visually moves while the mouse moves). If a child’s internal state controls a visual element (such as uncontrolled open/close), the provider naturally has no idea this occurs. This is an issue because it will render the handle incorrectly when the user attempts to drag it. This method allows for internal state updates to register upwards to correct for this issue.~~
    - ~~This will be specifically used inside of `ChartCard `to propagate uncontrolled changes to the open/close state.~~
 
## Example
```js
function ChartPage() {
  // Chart state and dragEnd handlers configured here

  return (
    <DragProvider onDragEnd={handleChartCardDrag}>
      {/** Controls drag events for cards */}
      <ChartCard dragId="chart-card-1">
        <DragProvider onDragEnd={handleChartDrag}>
          {/** Controls drag events for charts inside chart-card-1 */}
          <Chart dragId="chart-1" />
          <Chart dragId="chart-2" />
        </DragProvider>
      </ChartCard>
      <ChartCard dragId="chart-card-2">
        <DragProvider onDragEnd={handleChartDrag}>
          {/** Controls drag events for charts inside chart-card-2 */}
          <Chart dragId="chart-3" />
          <Chart dragId="chart-4" />
        </DragProvider>
      </ChartCard>
    </DragProvider>
  }
}
```

## Screenshot
https://github.com/user-attachments/assets/621c56c4-ed7b-4259-a01e-20f61d8827fd

### For new components

## 🧪 How to test changes
- All newly added tests are passing
- All previous tests are passing
- Chromatic regression tests are passing
- Manual QA in Storybook


charts/child-state-update-provider/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "module": "./dist/esm/index.js",
+  "types": "./dist/index.d.ts",
+  "license": "Apache-2.0",
+  "scripts": {
+    "build": "lg build-package",
+    "tsc": "lg build-ts",
+    "docs": "lg build-tsdoc"
+  },
+  "publishConfig": {
+    "access": "restricted"
```
tsck, line 15: I set this as `restricted` with the goal of this not being a public package. Is this how we would typically do this? It doesn't look like it's really done elsewhere in the codebase.

For context, I made this it's own package so that it could be a dependency of `core` and `drag-provider` without causing a dependency issue. This doesn't seem like something that's typically done so happy to do this differently if that makes more sense.

charts/drag-provider/src/DndKitProvider/DndKitProvider.tsx
tsck, line 1: I don't love this naming since it ties it directly to Dndkit. This was done simply to separate it from `DragProvider` since the child state logic needed to be added at that level. Open to suggestions here.

charts/child-state-update-provider/src/ChildStateUpdateProvider/ChildStateUpdateProvider.types.ts
```diff
@@ -0,0 +1,5 @@
+import { PropsWithChildren } from 'react';
+
+export type DragProviderProps = PropsWithChildren<{
+  onDragEnd?(event: { active: string; over: string }): void;
+}>;
```
TheSonOfThomp, line 5: Why is this exported from `ChildStateUpdateProvider`?

charts/child-state-update-provider/src/ChildStateUpdateProvider/ChildStateUpdateProvider.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+    expect(result?.current?.childStateUpdates).toEqual({});
+
+    act(() => {
+      result?.current?.updateChildState('dragId', { isOpen: true });
+    });
+
+    expect(result?.current?.childStateUpdates).toEqual({
+      dragId: { isOpen: true },
+    });
+  });
```
TheSonOfThomp, line 23: This isn't testing that the "child state" is updated, only that the context object is changed

charts/drag-provider/src/DndKitProvider/DndKitProvider.tsx
TheSonOfThomp: I think this is fine, since it pretty directly interacts with dndKit

charts/core/src/ChartCard/ChartCard.styles.ts
```diff
@@ -20,16 +27,83 @@ export const getContainerStyles = (theme: Theme) => css`
...
+  transform: ${CSS.Transform.toString(transform)};
+  transition: ${transition};
+`;
+
+const getDraggingContainerStyles = () => css`
+  opacity: 0.3;
+`;
+
+const getOverlayContainerStyles = () => css`
+  box-shadow: 0 18px 18px -15px rgba(0, 30, 43, 0.2);
```
TheSonOfThomp, line 43: should this use a token?

charts/core/src/ChartCard/ChartCard.styles.ts
```diff
@@ -20,16 +27,83 @@ export const getContainerStyles = (theme: Theme) => css`
...
+    css`
+      width: 100%;
+      height: 100%;
+      padding: ${spacing[150]}px ${spacing[300]}px;
+      display: grid;
+      grid-template-columns: 1fr auto;
+      background: ${state === ChartCardStates.Overlay
+        ? color[theme].background[Variant.Primary][InteractionState.Hover]
+        : 'none'};
+    `,
```
TheSonOfThomp, line 99: nit: I might break this out with `cx`
```
cx(css`
  background: 'none';
`,
{
  [css`
    background: '...'
  `]: state === ChartCardStates.Overlay
})
```

charts/core/src/ChartCard/ChartCard.tsx
```diff
@@ -1,24 +1,26 @@
 import React, { forwardRef, MouseEvent, useEffect, useState } from 'react';
+import { useSortable } from '@dnd-kit/sortable';
+import { useChildStateUpdateContext } from '@lg-charts/child-state-update-provider';
```
TheSonOfThomp, line 3: 1. Is there any reason we can't export `useChildStateUpdateContext` from `@leafygreen-ui/draggable`?
2. I feel like we should also re-export `useSortable` from our `/draggable` package

charts/core/src/ChartCard/ChartCard.tsx
```diff
@@ -47,6 +51,29 @@ export const ChartCard = forwardRef<HTMLDivElement, ChartCardProps>(
...
+    const isDraggable = !!(items.length && dragId);
+    const { updateChildState } = useChildStateUpdateContext();
+
+    useEffect(() => {
+      if (isDraggable && updateChildState) {
+        setTimeout(() => {
+          updateChildState(dragId, { isOpen });
+        }, transitionDuration.slower); // Update after transition to prevent re-render before animation
+      }
+    }, [updateChildState, isOpen, isDraggable, dragId]);
```
TheSonOfThomp, line 75: This feels wrong

charts/core/src/Tooltip/Tooltip.tsx
TheSonOfThomp, line 1: nit: can we name this something more specific?

charts/core/src/Tooltip/Tooltip.tsx
```diff
@@ -53,7 +56,12 @@ export function Tooltip({
           : undefined,
         order: getSortOrder(sortDirection, sortKey),
         padding: spacing[200],
-        show: true,
+        // Showing the tooltip in ChartStates.Dragging or ChartStates.Overlay causes issues with drag and drop
+        show:
+          chart.state !== ChartStates.Dragging &&
+          chart.state !== ChartStates.Overlay &&
+          chartCardContext?.state !== ChartCardStates.Dragging &&
+          chartCardContext?.state !== ChartCardStates.Overlay,
```
TheSonOfThomp, line 64: nit: consider extracting this to a `shouldShow` util function

charts/core/src/Chart/Chart.styles.ts
```diff
@@ -12,10 +19,67 @@ export const chartContainerStyles = css`
   width: 100%;
 `;
 
-export const chartHeaderContainerStyles = css`
-  grid-area: chartHeader;
+const getDraggableContainerStyles = (
+  transform: Transform | null,
+  transition?: string,
+) => css`
+  transform: ${CSS.Transform.toString(transform)};
```
TheSonOfThomp, line 26: I assume this is because of how dndKit returns a transform value?  Can we just add a comment about this?

charts/core/src/Chart/Chart.styles.ts
```diff
@@ -12,10 +19,67 @@ export const chartContainerStyles = css`
...
+  transition: ${transition};
 `;
 
+const getDraggingContainerStyles = () => css`
+  opacity: 0.3;
+`;
+
+const getOverlayContainerStyles = (theme: Theme) => css`
+  box-shadow: 0 18px 18px -15px rgba(0, 30, 43, 0.2);
```
TheSonOfThomp, line 35: same here. Should this be a token?

charts/core/package.json
```diff
@@ -15,6 +15,8 @@
     "access": "public"
   },
   "dependencies": {
+    "@dnd-kit/sortable": "^10.0.0",
+    "@dnd-kit/utilities": "^3.2.2",
```
TheSonOfThomp, line 19: If `@lg-charts/core` depends on `@dnd-kit`, I don't see a reason why we can't bundle all `dnd-kit` `@lg/draggable` and  the child-state-update-provider into one dependency.

TheSonOfThomp: This also pretty tightly ties our drag-n-drop to `dnd-kit`, since it creates implicit peer dependencies for `@lg/draggable`

charts/drag-provider/src/DragProvider.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+    </Chart>
+  );
+}
+
+async function clickAndDragByTestId(id: Matcher) {
+  const header = screen.getByTestId(id);
+  fireEvent.mouseDown(header, { clientX: 0, clientY: 0 });
+  fireEvent.mouseMove(header, { clientX: 100, clientY: 100 });
+  fireEvent.mouseUp(header);
+}
```
TheSonOfThomp, line 44: Future: might be worth exporting some sort of drag-n-drop event along with test harnesses (since `userEvent` doesn't support `.drag`)

charts/child-state-update-provider/src/ChildStateUpdateProvider/ChildStateUpdateProvider.spec.tsx
tsck: Right, that's actually the intention. "child state object" is misleading here. Are you suggesting that I just update the description to be less confusing or that we actually check the child state?

charts/core/src/ChartCard/ChartCard.styles.ts
tsck: So this didn't match a token but I think this is a good point that maybe it should. I'll check in with Sooa

charts/child-state-update-provider/package.json
tsck: I can set `"private": true` instead which would prevent it from being published altogether, but that seemed maybe incorrect too?

charts/core/src/ChartCard/ChartCard.tsx
tsck: Yes it does lol. Unfortunately without this the state update propagates upwards triggering a re-render before the animation ☹️

charts/core/src/ChartCard/ChartCard.tsx
tsck: Is the suggestion here more so then to create a `@leafygreen-ui/draggable` and have everything - `DragProvider`, `useChildStateUpdateContext`, and `ChildStateUpdateProvider`, as well as the dndkit utils, be exported from it? I'm not sure the current solution is generic enough to be a universal drag solution since it's so manually tied to the two draggable chart components. 

`child-state-update-provider` was moved into its own package because `chart-card` depended on it, and `drag-provider` had a dev dependency on `chart-card`. So exporting it from `drag-provider` was creating a circular dependency.

charts/core/src/Tooltip/Tooltip.tsx
tsck: I'm wondering if this should be apply to all components in the package then? Should everything be prefixed with `Chart` for example?

charts/child-state-update-provider/package.json
tsck: Package removed so no longer relevant

charts/core/package.json
tsck: Just wanted to clarify the thought here. `child-state-update-provider` in particular was created as a separate dependency to avoid a circular dependency, but that doesn't exist anymore so it's no longer relevant anyway 🙂

Is the ask here then to add `@dnd-kit/utilities` as a dependency of `@lg-charts/drag-provider` and then make `@lg-charts/drag-provider` a dependency of `@lg-charts/core`?

charts/core/src/Chart/Chart.styles.ts
tsck: For future context: In offline chat we decided to leave a TODO about this instead, noting that we will do this once we audit our tokens

charts/core/src/ChartCard/ChartCard.styles.ts
tsck: For future context: In offline chat we decided to leave a TODO about this instead, noting that we will do this once we audit our tokens

tsck:
For future context: Review done in dev sync led to a discussion around handling passing open/close data upwards to the `DragProvider`, in order to properly render the overlay component. A simpler solution was agreed upon using data attributes on the `ChartCard` elements to determine it's state from the the `DragProvider`.

Changes pushed to make this change.

charts/core/src/Tooltip/Tooltip.tsx
TheSonOfThomp: If it's charts specific, then probably?

TheSonOfThomp: at least the things that have a naming collision with core components

charts/drag-provider/src/DragProvider.tsx
```diff
@@ -1,13 +1,174 @@
...
+/**
+ * ChartCards in particular have open and closed states. Since these states
+ * are stored internal to those components, we need some way to determine
+ * if a particular card is open or closed when it is being dragged and we need
+ * to render it's overlay.
+ *
+ * This function finds the open state of a drag element based on data attributes
+ * added to the element.
+ */
+function findDragElementOpenState(
```
TheSonOfThomp, line 35: - why not return just a boolean?
- should this just be `get...`?

charts/drag-provider/src/DragProvider.tsx
```diff
@@ -1,13 +1,174 @@
...
+      return null;
+    }
+
+    const elem = childrenArray.find(
+      child => (child as React.ReactElement).props.dragId === activeId,
+    ) as React.ReactElement;
+
+    const overlayProps: { state: string; isOpen?: boolean } = {
+      state: 'overlay',
+    };
```
TheSonOfThomp, line 140: Can we just set `isOpen` here?
```ts
const overlayProps: { state: string; isOpen?: boolean } = {
  state: 'overlay',
  isOpen: findDragElementOpenState(childrenRef.current!, activeId);
};
```

charts/drag-provider/src/DragProvider.tsx
TheSonOfThomp, line 1: Might be tricky to do, but I think this might warrant a Chromatic `play` test to ensure the drag is visually correct

charts/drag-provider/src/DragProvider.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Should this `overlayProps` type be more explicitly defined?

charts/core/src/Tooltip/Tooltip.tsx
tsck: Ok that's probably just Tooltip then. If it's alright, I'll change the name in the next piece of work since it will change the Tooltip - https://jira.mongodb.org/browse/LG-4661

charts/drag-provider/src/DragProvider.tsx
tsck: Because a `Chart` doesn't have an `isOpen` prop. So just returning the boolean, we'd be setting it every time, even on components where it doesn't exist. I think this is more clear and explicit that we don't expect this to always exist.

charts/drag-provider/src/DragProvider.tsx
```diff
@@ -1,13 +1,174 @@
...
+MouseSensor.activators = [
+  {
+    eventName: 'onMouseDown',
+    handler: ({ nativeEvent: event }: MouseEvent) => {
+      const element = event.target as HTMLElement;
+      let cur: HTMLElement | null = element;
+
+      while (cur) {
+        /**
+         * Block Dnd if element has data-no-dnd attribute. Used so open button and
```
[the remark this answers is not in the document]
tsck: Ours. We can call it whatever we want, it's just an escape hatch to tie into to easily prevent the drag trigger when specific elements are clicked.

charts/drag-provider/src/DragProvider.tsx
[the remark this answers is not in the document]
tsck: Ngl, `core` just isn't a dependency of `drag-provider` so I was inlining this as such to avoid adding just that.

We could be more specific by using `ChartProps` and `ChartCardProps` types, but we'll have to add it as a dependency.

I can add it and update this. Since we'll just be importing types it shouldn't matter I guess. Also I guess you're never going to use `drag-provider` without `core` anyway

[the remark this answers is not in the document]
tsck: Updated the prop types to rely on those imported from `core` here.

`ChartStates.Overlay` is tricky because currently there are `ChartStates` and `ChartCardStates`, where essentially `ChartCardStates === Omit<ChartStates, 'loading'>` at the moment but they aren't tied together like that. At this point, we don't actually know if the child is a `Chart` or a `ChartCard`. We could just rely on one or the other, but that's essentially the equivalent of writing them in. Happy to consider something else, but right now I've just left them.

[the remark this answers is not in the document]
tsck: We came up with a possible solution offline. Adding a `DragStates` enum/type to the `drag-provider` package and exporting that. Then having `drag-provider` be a dep of `core`, and having `ChartStates` and `ChartCardStates` extend `DragStates`. 

However, `drag-provider` currently has a dev dep on `core`. This I believe is important because it's needed for proper unit and visual regression testing. Without it tests would need to rely on mock components with fake drag targets. We could add the stories to the `core` library then instead, but that also seems odd to me. 

Since this is just an effort to create better linked types, I'd rather we just leave things as is. This can always be fixed at a later time if it's needed.

charts/drag-provider/package.json
```diff
@@ -23,9 +23,13 @@
   },
   "dependencies": {
     "@dnd-kit/core": "^6.3.1",
-    "@dnd-kit/sortable": "^10.0.0"
+    "@dnd-kit/sortable": "^10.0.0",
+    "@lg-charts/core": "workspace:^"
```
TheSonOfThomp, line 27: we could get around this by moving the chart drag tests to charts-core. That may be more "technically correct" since the weirdness is in Charts

charts/drag-provider/src/DragProvider.stories.tsx
```diff
@@ -13,16 +13,89 @@ import {
...
+  target,
+  x = 310,
+  y = 50,
+}: {
+  target: HTMLElement;
+  x?: number;
+  y?: number;
+}) {
+  // FIXME:
+  // @ts-expect-error Types & Transpiled code for userEvent resolved from different packages
```
TheSonOfThomp, line 43: 🙃

TheSonOfThomp (approved):

PR #2742, opened 2025-03-04
Update prebuild script

bruugey:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/packages/icon/scripts/prebuild.ts
+++ b/packages/icon/scripts/prebuild.ts
@@ -62,6 +62,8 @@ async function buildSvgFiles(
   const outputDir = await createOutputDirectory(flags);
 
   await Promise.all(svgFiles.map(processFile(outputDir)));
+
+  await createIndexFile(svgFiles);
 }
 
 function filterSvgFiles(str: string): boolean {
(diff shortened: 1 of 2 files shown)


packages/icon/scripts/prebuild.ts
```diff
@@ -84,6 +86,44 @@ async function createOutputDirectory(flags: Flags) {
   return outputDir;
 }
 
+async function indexTemplate(svgFiles: Array<FileObject>) {
```
TheSonOfThomp, line 89: nit: something like `getIndexContent` would be a more apt name

packages/icon/src/glyphs/index.ts
```diff
@@ -332,7 +354,7 @@ const _glyphs = {
   Write,
   X,
   XWithCircle,
-} as const;
+};
```
TheSonOfThomp, line 357: missing `as const` in gen script

TheSonOfThomp (changes requested):
Just missing `as const` from the build script

packages/icon/src/glyphs/index.ts
bruugey: added

TheSonOfThomp (approved):

PR #2767, opened 2025-03-18
Updates Menu Backdrop behavior

TheSonOfThomp:
### Hooks:
- Updates 3rd argument in `useBackdropClick` to accept an options object. Retains (but deprecates) boolean-only functionality.
  - Adds `options.allowPropagation` to allow or disallow the click event to bubble up to other elements.

### Lib
Adds `getClosestFocusableElement`. 
This function crawls up the DOM tree to find the closest focusable element to the provided element
Returns the element itself if it is focusable or if no focusable element is found, it returns the body element.

### Menu:
- Re-enables native event propagation on trigger clicks. (Previously this was done to prevent errors caused by different event handling in React vs Backbone)
- Adds logic when the menu closes to check if the click occurred on an element that is focusable. If it is then we want to focus that element, otherwise we want to focus the menu trigger.


packages/lib/src/getClosestFocusableElement/getClosestFocusableElement.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * Crawls up the DOM tree to find the closest focusable element.
+ * Returns the element itself if it is focusable.
+ * If no focusable element is found, it returns the body element.
+ *
+ * This is useful for ensuring that focus is returned to a valid element
+ * after an event, such as a click or key press.
+ */
+export const getClosestFocusableElement = (el: HTMLElement): HTMLElement => {
+  // check if the element is focusable
+  if (el.matches(focusableSelectors.join(','))) {
```
shaneeza, line 22: Would [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) do the same thing? It would return the nearest ancestor, including itself? And then you can prob avoid recursion.

shaneeza, line 22: nit: If you move `focusableSelectors.join(',')` outside, you can avoid invoking it each time.

packages/menu/src/Menu.spec.tsx
```diff
@@ -183,6 +184,34 @@ describe('packages/menu', () => {
...
+
+    test("Clicking a button outside the menu fires that button's handlers, and sets focus to the button", async () => {
+      const otherButtonHandler = jest.fn();
+      const { getByTestId, findByTestId } = render(
+        <>
+          <Menu trigger={defaultTrigger} data-testid={menuTestId}>
+            <MenuItem>Item A</MenuItem>
+            <MenuItem>Item B</MenuItem>
+          </Menu>
+          <Button data-testid="outside-button" onClick={otherButtonHandler}>
```
shaneeza, line 198: `data-testid='outie-button'`

packages/hooks/src/useBackdropClick.ts
```diff
@@ -42,12 +73,28 @@ export function useBackdropClick(
    * 5. Then we call the callback (typically fires `closeMenu`, setting `isOpen = false`, and rerender the component)
    */
 
+  // TODO: Remove this in a major version
```
shaneeza, line 76: make a ticket?

packages/menu/src/Menu/Menu.tsx
```diff
@@ -89,20 +94,44 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
...
-  const handleClose = useCallback(() => {
-    if (shouldClose()) {
-      keyboardUsedRef.current = false;
-      setOpen(false);
-    }
-  }, [setOpen, shouldClose]);
+
+  const handleClose = useCallback(
+    (event?: MouseEvent | React.MouseEvent) => {
+      // TODO: move this logic to useBackdropClick
```
shaneeza, line 100: ticket?

packages/menu/src/Menu/Menu.tsx
```diff
@@ -199,6 +228,7 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
           onEntered={handlePopoverOpen}
           data-testid={LGIDs.root}
           data-lgid={LGIDs.root}
+          ref={popoverRef}
```
shaneeza, line 231: why the switch?

packages/lib/src/getClosestFocusableElement/getClosestFocusableElement.tsx
TheSonOfThomp: Oh, yeah. Last time I looked at that it wasn't well supported

packages/menu/src/Menu/Menu.tsx
TheSonOfThomp: b/c the popover ref should be on the popover element
I'll add to the changeset

shaneeza (approved):

PR #2777, opened 2025-03-26
LG-4712 | LG-4992: move ChartCard into separate package and fix passing props to 2 DOM nodes

stephl3:
## ✍️ Proposed changes

- move `ChartCard` into separate `@lg-charts/chart-card` package
- fix passing props to 2 DOM nodes

🎟 _Jira ticket:_ [LG-4712](https://jira.mongodb.org/browse/LG-4712) | [LG-4992](https://jira.mongodb.org/browse/LG-4992)

### DIFF
--- a/.changeset/chart-card-fix.md
+++ b/.changeset/chart-card-fix.md
@@ -3,4 +3,3 @@
 ---
 
 Fixes passing props to 2 separate DOM nodes in `ChartCard`
-g
--- a/charts/chart-card/README.md
+++ b/charts/chart-card/README.md
@@ -26,25 +26,11 @@ npm install @lg-charts/legend
 
 ```js
 import { ChartCard } from '@lg-charts/chart-card';
-import { Chart, Line, XAxis, YAxis } from '@lg-charts/core';
+import { Chart } from '@lg-charts/core';
 
 <ChartCard title="My Group of Charts">
-  <Chart groupId="group1">
-    <XAxis type="time" />
-    <YAxis type="value" formatter={(value) => `${value}GB`} />
-    <Line
-      name="Series 1"
-      data={seriesData}
-    />
-  </Chart>
-  <Chart groupId="group1">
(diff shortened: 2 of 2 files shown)


charts/chart-card/README.md
```diff
@@ -0,0 +1,59 @@
...
+  </Chart>
+  <Chart groupId="group1">
+    <XAxis type="time" />
+    <YAxis type="value" formatter={(value) => `${value}GB`} />
+    <Line
+      name="Series 1"
+      data={seriesData}
+    />
+  </Chart>
+<ChartCard>;
```
tsck, line 48: Nit: I wonder if we need to have charts as populated as they are here. I'm thinking it may distract from the card which is what we're actually demoing now that this is a separate package. I'm wondering if instead we should just do:

```typescript
import { ChartCard } from '@lg-charts/chart-card';
import { Chart } from '@lg-charts/core';

<ChartCard title="My Group of Charts">
  <Chart />
  <Chart />
<ChartCard>;
```

Not a strong opinion here.

charts/core/src/ChartCard/ChartCard.tsx
```diff
@@ -109,7 +109,6 @@ export const ChartCard = forwardRef<HTMLDivElement, ChartCardProps>(
             data-testid="lg-charts-core-chart_card-header"
             {...attributes}
             {...listeners}
-            {...rest}
```
tsck, line 112: Good catch!

.changeset/chart-card-fix.md
```diff
@@ -0,0 +1,6 @@
+---
+'@lg-charts/chart-card': minor
+---
+
+Fixes passing props to 2 separate DOM nodes in `ChartCard`
+g
```
tsck, line 6:
```suggestion

```

tsck (approved):

PR #2780, opened 2025-03-28
Deprecates Code props

bruugey:
## ✍️ Proposed changes

### For new components


.changeset/modern-buttons-bow.md
TheSonOfThomp, line 1: Should we add Panel usage guidelines here?

.changeset/modern-buttons-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/code': major
+---
+
+The following props are deprecated and have been removed: copyable, showCustomActionButtons, languageOptions, customActionButtons, chromeTitle, onChange. Please use the Panel component instead.
```
TheSonOfThomp, line 5: nit, these var names should use `\``

packages/code/src/Code/Code.spec.tsx
```diff
@@ -163,6 +151,7 @@ describe('packages/Code', () => {
                 spy.mockReturnValue(true);
                 return renderCode({
                   isLoading: true,
+                  // @ts-expect-error
                   copyable: false,
```
TheSonOfThomp, line 155: why are we passing in copyable here if it's removed?

.changeset/modern-buttons-bow.md
bruugey: They're already below when they were introduced. I can resurface if you think that makes sense?

packages/code/src/Code/Code.spec.tsx
bruugey: i think the point is to test that it would error, but i'll rm.

packages/code/src/Code/Code.tsx
```diff
@@ -71,6 +62,27 @@ function Code({
...
+    // @ts-expect-error detecting deprecated props
+    rest.languageOptions ||
+    // @ts-expect-error detecting deprecated props
+    rest.customActionButtons ||
+    // @ts-expect-error detecting deprecated props
+    rest.chromeTitle ||
+    // @ts-expect-error detecting deprecated props
+    rest.onChange
+  ) {
+    console.warn(
```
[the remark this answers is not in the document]
bruugey: i like it, but i can rm if you want me to

TheSonOfThomp (approved):

packages/code/src/Code/Code.spec.tsx
```diff
@@ -211,7 +198,6 @@ describe('packages/Code', () => {
               return renderCode({
                 isLoading: false,
                 language: languageOptions[0].displayName,
-                panel: <Panel />,
```
shaneeza, line 214: I think the panel prop should be here since we're testing `with panel slot` on line 193. Was this the test that was failing because two copy buttons were rendering?

bruugey: No no all of the tests were passing

bruugey: I can add this back

packages/code/src/Code/Code.spec.tsx
```diff
@@ -75,26 +75,14 @@ describe('packages/Code', () => {
       });
 
       test('announces copied to screenreaders when content is copied in the panel', () => {
-        renderCode({ panel: <Panel /> });
```
shaneeza, line 78: same here

packages/code/src/LanguageSwitcher/LanguageSwitcherExample.tsx
```diff
@@ -98,10 +98,14 @@ export function LanguageSwitcherWithDeprecatedPropsExample({
     <Code
       {...rest}
       language={language.displayName}
-      languageOptions={languageOptions}
-      customActionButtons={customActionButtons}
-      onChange={handleChange}
-      chromeTitle="Title"
+      panel={
```
shaneeza, line 101: You can remove this example since we don't need it anymore. The example above, `LanguageSwitcherWithPanelExample`, already uses the panel prop.

packages/code/src/Code/Code.spec.tsx
bruugey: Added!

packages/code/src/Code/Code.spec.tsx
bruugey: Added!

bruugey:
@shaneeza ready for a re-r

shaneeza (approved):

PR #2798, opened 2025-04-03
ASDF

TheSonOfThomp:
## ✍️ Proposed changes

Replaces our implicit node & pnpm version management with a more explicit tool manager; [`asdf`](https://asdf-vm.com/guide/introduction.html)

Running `asdf install` ensures all contributors are using the same version of Node & pnpm, which can avoid versioning conflicts.

 ```bash
   brew install asdf;
   asdf install;
 ```

Note: the MMS codebase also uses `asdf`, which would make moving between codebases easier

### Don't we already list these versions?

We list Node version in `engines`, and the `pnpm` version is already identified twice, in `engines` and `packageManager`, so this may seem like a third source of truth.
But `asdf install` will actually **install** the correct version, whereas  the "engines" attribute will only tell you if you're out of date

### Alternatives
An alternative to `asdf` would be `nvm` (which many of us already use).
We could set the required node version in a `.nvmrc` file, which would solve the Node versioning issue.

Ensure you have the correct Node version with `nvm install`.,

However, installing the correct `pnpm`  version requires another tool; `corepack`. 
This would require contributors to manually enable the tool (`corepack enable`) and install the correct version `corepack install pnpm@9.15.6`.
As far as I can tell, there is no version file we can write to that will automatically install `corepack` dependencies


TheSonOfThomp:
not a priority atm

PR #2800, opened 2025-04-04
Typescript 5.8

TheSonOfThomp:
## ✍️ Proposed changes

---
'@lg-tools/build': minor
---

- Upgrades to TS 5.8. 
- Adds `--downlevel` option for `build-ts`. This option exports downleveled `*.d.ts` files for a defined set of targets. 
Updates a package's `package.json` if necessary `--update` flag is provided.
This uses [downlevel-dts](https://github.com/sandersn/downlevel-dts) under the hood.

🎟 _Jira ticket:_ https://jira.mongodb.org/browse/LG-5006


TheSonOfThomp:
@bruugey Thinking again on the downleveling. If we enable downleveling to ts4.9, would this be a patch instead of major?

PR #2808, opened 2025-04-14
Storybook export `findStories`

TheSonOfThomp:
### `@lg-tools/storybook-addon`: minor

Updates webpack config to build internal node apis

### `@lg-tools/storybook-utils`: patch

Exports `findStories` from utils


tools/storybook-utils/src/findStories.ts
```diff
@@ -0,0 +1,27 @@
...
+  includePattern: Array<string>;
+  excludePattern: Array<string>;
+}
+/**
+ * Finds story files to include based on the provided include/exclude glob patterns
+ * @param includePattern
+ * @param excludePattern
+ * @returns Array of all included files
+ */
+export function findStories({
```
TheSonOfThomp, line 11: This is all I wanted to do! 😢

tools/storybook-addon/src/main/index.ts
```diff
@@ -2,25 +2,28 @@
...
+} from 'webpack';
 
-import { findStories } from './findStories';
 import { isRule } from './utils';
 
 export { managerHead } from './manager-head';
 export { previewHead } from './preview-head';
 
-import { getLGConfig } from '@lg-tools/meta';
+const directories = ['packages', 'charts', 'chat', 'tools'];
```
TheSonOfThomp, line 19: hard-coding this since it was causing some issues

tsck: Q: Just curious what the issues were? `getLGConfig()` seems fairly straight forward so just wondering what was breaking there since it would still be nice to have this dynamically generated for us.

tools/storybook-utils/src/findStories.ts
tsck: Pain

image

tsck (commented):
Pretty much LGTM, just curious about one thing.

tools/storybook-addon/src/main/index.ts
TheSonOfThomp: Actually, I could probably go back to this now. At first I thought the issues were separate from the webpack ones

tsck (approved):
LGTM!

PR #2813, opened 2025-04-18
Remove `LegacyCheck`

TheSonOfThomp:
## ✍️ Proposed changes

This is still technically a major change since you could still access it at `@leafygreen-ui/checkbox/src/LegacyCheck`

### '@leafygreen-ui/checkbox': major

Removes `LegacyCheck`, previously deprecated an inaccessible since v7.0


.changeset/checkbox-remove-legacy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/checkbox': major
+---
+
+Removes `LegacyCheck`, previously deprecated an inaccessible since v7.0
```
shaneeza, line 5: `deprecated and inaccessible`?

shaneeza (approved):

.changeset/checkbox-remove-legacy.md
TheSonOfThomp: Not only was it deprecated, but it was difficult to even import (maybe "inaccessible is the wrong word")

shaneeza: Yeah, I was pointing out the `and`. Next time, I'll add a suggestion for stuff like this.

PR #2823, opened 2025-04-23
Validate builds based on `package.json`

TheSonOfThomp:
## ✍️ Proposed changes

`@lg-tools/validate`: patch
---

Updates build validation to use the entry files defined in the package.json

### Why?

Changing the entry point in the package.json  (e.g. `main`, `module` or `types`) can break the bundle.
This change ensures that the bundle files exist in the directory we say they'll be in

### DIFF
--- a/tools/validate/src/builds/index.ts
+++ b/tools/validate/src/builds/index.ts
@@ -11,10 +11,6 @@ import path from 'path';
 import { ValidateCommandOptions } from '../validate.types';
 
 import { getModuleTypes } from './getModuleTypes';
-
-// A list of packages to ignore
-const ignorePackages = ['@lg-tools/storybook'];
-
 /**
  * Validates `umd`, `esm` and TS build integrity for all packages in the repository.
  */
@@ -52,12 +48,6 @@ export const validateBuilds = ({
         isESMValid: true,
         tsExists: true,
       };
-
-      // Skip some packages
-      if (ignorePackages.includes(pkgName)) {
-        verbose && console.log(chalk.gray(`Skipping package ${pkgName}`));
-        continue;
-      }
 
       const distDir = path.resolve(pkgPath, 'dist');
       const buildExists = fse.existsSync(distDir);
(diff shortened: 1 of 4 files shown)


tools/validate/src/builds/index.ts
```diff
@@ -3,17 +3,16 @@
...
 import path from 'path';
-import vm from 'vm';
 
 import { ValidateCommandOptions } from '../validate.types';
 
-import { ModuleType } from './modules.types';
+import { getModuleTypes } from './getModuleTypes';
 
-// A list of
+// A list of packages to ignore
```
shaneeza, line 15: Can you add a note on why we ignore some packages?

tools/validate/src/builds/index.ts
```diff
@@ -31,15 +30,32 @@ export const validateBuilds = ({
...
-      // Skip packages
+      // extract the name and output paths from the package.json
+      const {
+        name: pkgName,
+        main: umdPath,
+        module: esmPath,
+        types: typesPath,
+      } = pkgJson;
+
+      const checks = {
```
shaneeza, line 48: Are these defaults?

TheSonOfThomp: we default to assuming everything passes, and then set it to false if it should exist but doesn't (L70-93)

tools/validate/src/builds/index.ts
TheSonOfThomp: that package doesn't actually exist. rm'ing the ignore array

tools/validate/src/index.ts
```diff
@@ -6,14 +7,22 @@ import { ValidateCommandOptions } from './validate.types';
...
+  }
 
   Promise.all(validators)
     .then(() => {
+      options.verbose && console.log('All validations passed. ✅');
       process.exit(0);
     })
-    .catch(() => {
+    .catch(err => {
+      console.log('Some validation failed.');
```
shaneeza, line 24: itty bitty teeny tiny nit: no ❌? To match the code above?

shaneeza (approved):

tools/validate/src/index.ts
shaneeza:
```suggestion
      console.log('Some validations failed.');
```

PR #2826, opened 2025-04-23
Adds `getPackageRoot` utility

TheSonOfThomp:
## ✍️ Proposed changes

Adds `getPackageRoot` utility that returns the full path to a given package dir


.changeset/meta-get-package-root.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-tools/meta': patch
```
TheSonOfThomp, line 2: Only `patch` b/c it's pre v1

PR #2849, opened 2025-05-08
Ingest with recursive crawler

TheSonOfThomp:
Update ingest to use `recursiveCrawlFromBaseURL` from `@lg-tools/crawler`

Update `recursiveCrawlFromBaseURL` to accept callback instead of baking in the document processing


apps/chatbot-server/tsconfig.json
tsck, line 1: Got an unresolved merge conflict here.

TheSonOfThomp: oh weird. how'd it let me commit that

TheSonOfThomp: good now

tsck (approved):

PR #2859, opened 2025-05-19
LG-5043: drawer layout

shaneeza:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/packages/drawer/package.json
+++ b/packages/drawer/package.json
@@ -25,6 +25,7 @@
     "@leafygreen-ui/polymorphic": "workspace:^",
     "@leafygreen-ui/tabs": "workspace:^",
     "@leafygreen-ui/tokens": "workspace:^",
+    "@leafygreen-ui/toolbar": "workspace:^",
     "@leafygreen-ui/typography": "workspace:^",
     "@lg-tools/test-harnesses": "workspace:^",
     "polished": "^4.2.2",
--- a/packages/drawer/src/Drawer.stories.tsx
+++ b/packages/drawer/src/Drawer.stories.tsx
@@ -8,13 +8,15 @@ import {
 import { StoryFn, StoryObj } from '@storybook/react';
 
 import Button from '@leafygreen-ui/button';
-import { css } from '@leafygreen-ui/emotion';
+import { css, cx } from '@leafygreen-ui/emotion';
 import { spacing } from '@leafygreen-ui/tokens';
 import { Body } from '@leafygreen-ui/typography';
 
 import { DisplayMode, Drawer, DrawerProps } from './Drawer';
 import { DrawerStackProvider } from './DrawerStackContext';
 import { EmbeddedDrawerLayout } from './EmbeddedDrawerLayout';
+import { Toolbar, ToolbarIconButton } from '@leafygreen-ui/toolbar';
+import { drawerTransitionDuration } from './Drawer/Drawer.styles';
 
 const SEED = 0;
 faker.seed(SEED);
@@ -33,9 +35,11 @@ export default {
       <div
         className={css`
           height: 100%;
-          width: 100%;
+          /* width: 100%; */
           display: flex;
           align-items: center;
+          margin: -100px;
+          width: 100vw;
         `}
       >
         <StoryFn />
@@ -69,7 +73,7 @@ export default {
 const LongContent = () => {
   const paragraphs = useMemo(() => {
     return faker.lorem
-      .paragraphs(20, '\n')
+      .paragraphs(30, '\n')
       .split('\n')
       .map((p, i) => <Body key={i}>{p}</Body>);
   }, []);
@@ -149,7 +153,6 @@ export const LiveExample: StoryObj<DrawerProps> = {
   render: TemplateComponent,
   args: {
     children: <LongContent />,
-    initialOpen: false,
   },
   parameters: {
     chromatic: {
@@ -159,6 +162,131 @@ export const LiveExample: StoryObj<DrawerProps> = {
       exclude: [...defaultExcludedControls, 'initialOpen'],
     },
   },
+};
+
+export const WithToolbarEmbedded: StoryFn<DrawerProps> = (
+  args: DrawerProps,
+) => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <DrawerStackProvider>
+      {/*  This logic will be handled internally inside ToolbarDrawerLayout */}
+      <EmbeddedDrawerLayout
+        className={css`
+          height: 100vh;
+        `}
+        isDrawerOpen={open}
+        hasToolbar
+      >
+        <main
+          className={css`
+            padding: ${spacing[400]}px;
+            overflow: auto;
+            display: flex;
+            flex-direction: column;
+            align-items: flex-start;
+            gap: ${spacing[200]}px;
+            background: aliceblue;
+          `}
+        >
+          <Button onClick={() => setOpen(prevOpen => !prevOpen)}>
+            Toggle Drawer
+          </Button>
+          <LongContent />
+        </main>
+        <div
+          className={css`
+            display: grid;
+            grid-template-columns: 48px 432px;
+            overflow: hidden;
+          `}
+        >
+          <Toolbar>
+            <ToolbarIconButton glyph="Code" label="Code" onClick={() => {}} />
+            <ToolbarIconButton glyph="Plus" label="Plus" onClick={() => {}} />
+          </Toolbar>
+          <Drawer
+            displayMode="embedded"
+            open={open}
+            onClose={() => setOpen(false)}
+            title="Drawer Title"
+          />
+        </div>
+        {/*  This logic will be handled internally inside ToolbarDrawerLayout */}
+      </EmbeddedDrawerLayout>
+    </DrawerStackProvider>
+  );
+};
+
+export const WithToolbarOverlay: StoryFn<DrawerProps> = (args: DrawerProps) => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <DrawerStackProvider>
+      {/*  This logic will be handled internally inside ToolbarDrawerLayout */}
+      <EmbeddedDrawerLayout
+        className={css`
+          height: 100vh;
+        `}
+        isDrawerOpen={open}
+        hasToolbar
+        displayMode={DisplayMode.Overlay}
+      >
+        <main
+          className={css`
+            padding: ${spacing[400]}px;
+            overflow: auto;
+            display: flex;
+            flex-direction: column;
+            align-items: flex-start;
+            gap: ${spacing[200]}px;
+            background: aliceblue;
+          `}
+        >
+          <Button onClick={() => setOpen(prevOpen => !prevOpen)}>
+            Toggle Drawer
+          </Button>
+          <LongContent />
+        </main>
+        <div
+          className={css`
+            display: grid;
+            grid-template-columns: 48px 432px;
+            overflow: hidden;
+          `}
+        >
+          <Toolbar
+            className={cx(
+              css`
+                position: absolute;
+                top: 0;
+                right: 0;
+                height: 100%;
+                transition: all ${drawerTransitionDuration}ms linear;
+                transform: translateX(0);
+              `,
+              {
+                [css`
+                  transform: translateX(-432px);
+                `]: open,
+              },
+            )}
+          >
+            <ToolbarIconButton glyph="Code" label="Code" onClick={() => {}} />
+            <ToolbarIconButton glyph="Plus" label="Plus" onClick={() => {}} />
+          </Toolbar>
+          <Drawer
+            displayMode="overlay"
+            open={open}
+            onClose={() => setOpen(false)}
+            title="Drawer Title"
+          />
+        </div>
+        {/*  This logic will be handled internally inside ToolbarDrawerLayout */}
+      </EmbeddedDrawerLayout>
+    </DrawerStackProvider>
+  );
 };
 
 const MultipleDrawersComponent: StoryFn<DrawerProps> = (args: DrawerProps) => {
--- a/packages/drawer/src/Drawer.stories.tsx
+++ b/packages/drawer/src/Drawer.stories.tsx
@@ -250,22 +250,41 @@ export const WithToolbarOverlay: StoryFn<DrawerProps> = (args: DrawerProps) => {
           <LongContent />
         </main>
         <div
-          className={css`
-            display: grid;
-            grid-template-columns: 48px 432px;
-            grid-area: drawer;
-            overflow: hidden;
-          `}
+          className={cx(
+            css`
+              display: grid;
+              grid-template-columns: 48px 100px;
+              grid-template-areas: 'toolbar2 drawer2';
+              grid-area: drawer;
+              /* overflow: hidden; */
+              position: relative;
+
+              /* width: 432px; */
+              justify-self: end;
+              transition: all ${drawerTransitionDuration}ms ease-in-out;
+            `,
+            {
+              [css`
+                grid-template-columns: 48px 432px;
+
+                @starting-style {
+                  grid-template-columns: 48px 100px;
+                }
+              `]: open,
+            },
+          )}
         >
           <Toolbar
             className={cx(
               css`
-                position: absolute;
-                top: 0;
-                right: 0;
+                /* position: absolute; */
+                /* top: 0;
+                right: 0; */
                 height: 100%;
-                transition: all ${drawerTransitionDuration}ms linear;
-                transform: translateX(0);
+                /* transition: all ${drawerTransitionDuration}ms linear; */
+                /* transform: translateX(0); */
+
+                grid-area: toolbar2;
               `,
               // {
               //   [css`
@@ -290,9 +309,13 @@ export const WithToolbarOverlay: StoryFn<DrawerProps> = (args: DrawerProps) => {
             open={open}
             onClose={() => setOpen(false)}
             title="Drawer Title"
-            // className={css`
-            //   position: absolute;
-            // `}
+            className={css`
+              /* position: absolute;
+               */
+              width: 432px;
+              grid-area: drawer2;
+              /* position: unset; */
+            `}
           />
         </div>
         {/*  This logic will be handled internally inside ToolbarDrawerLayout */}
--- a/packages/drawer/src/Drawer.stories.tsx
+++ b/packages/drawer/src/Drawer.stories.tsx
@@ -227,7 +227,7 @@ export const WithToolbarOverlay: StoryFn<DrawerProps> = (args: DrawerProps) => {
       {/*  This logic will be handled internally inside ToolbarDrawerLayout */}
       <EmbeddedDrawerLayout
         className={css`
-          height: 100vh;
+          height: 80vh;
         `}
         isDrawerOpen={open}
         hasToolbar
@@ -253,6 +253,7 @@ export const WithToolbarOverlay: StoryFn<DrawerProps> = (args: DrawerProps) => {
           className={css`
             display: grid;
             grid-template-columns: 48px 432px;
+            grid-area: drawer;
             overflow: hidden;
           `}
         >
@@ -266,21 +267,32 @@ export const WithToolbarOverlay: StoryFn<DrawerProps> = (args: DrawerProps) => {
                 transition: all ${drawerTransitionDuration}ms linear;
                 transform: translateX(0);
               `,
-              {
-                [css`
-                  transform: translateX(-432px);
-                `]: open,
-              },
+              // {
+              //   [css`
+              //     transform: translate3d(-432px, 0, 0); //3d
+              //   `]: open,
+              // },
             )}
           >
-            <ToolbarIconButton glyph="Code" label="Code" onClick={() => {}} />
-            <ToolbarIconButton glyph="Plus" label="Plus" onClick={() => {}} />
+            <ToolbarIconButton
+              glyph="Code"
+              label="Code"
+              onClick={() => setOpen(prevOpen => !prevOpen)}
+            />
+            <ToolbarIconButton
+              glyph="Plus"
+              label="Plus"
+              onClick={() => setOpen(prevOpen => !prevOpen)}
+            />
           </Toolbar>
           <Drawer
             displayMode="overlay"
             open={open}
             onClose={() => setOpen(false)}
             title="Drawer Title"
+            // className={css`
+            //   position: absolute;
+            // `}
           />
         </div>
         {/*  This logic will be handled internally inside ToolbarDrawerLayout */}
--- a/packages/drawer/src/Drawer.stories.tsx
+++ b/packages/drawer/src/Drawer.stories.tsx
@@ -270,6 +270,7 @@ export const WithoutToolbarEmbedded: StoryFn<DrawerProps> = (
   );
 };
 
+// FIXME: borders are making the page scroll horizontally:
 export const WithToolbarOverlay: StoryFn<DrawerProps> = (args: DrawerProps) => {
   const [open, setOpen] = useState(false);
 
--- a/packages/drawer/src/Drawer.stories.tsx
+++ b/packages/drawer/src/Drawer.stories.tsx
@@ -196,7 +196,7 @@ export const WithToolbarEmbedded: StoryFn<DrawerProps> = (
           </Button>
           <LongContent />
         </main>
-        <DrawerWithToolbarWrapper isDrawerOpen={open}>
+        <DrawerWithToolbarWrapper displayMode="embedded" isDrawerOpen={open}>
           <Toolbar
             className={cx(
               css`
@@ -300,7 +300,7 @@ export const WithToolbarOverlay: StoryFn<DrawerProps> = (args: DrawerProps) => {
           <LongContent />
           <LongContent />
         </main>
-        <DrawerWithToolbarWrapper isDrawerOpen={open}>
+        <DrawerWithToolbarWrapper displayMode="overlay" isDrawerOpen={open}>
           <Toolbar
             className={cx(
               css`
--- a/packages/drawer/src/Drawer.stories.tsx
+++ b/packages/drawer/src/Drawer.stories.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { faker } from '@faker-js/faker';
 import {
   storybookArgTypes,
@@ -8,15 +8,17 @@ import {
 import { StoryFn, StoryObj } from '@storybook/react';
 
 import Button from '@leafygreen-ui/button';
-import { css, cx } from '@leafygreen-ui/emotion';
+import { css, cx, keyframes } from '@leafygreen-ui/emotion';
 import { spacing } from '@leafygreen-ui/tokens';
+import { Toolbar, ToolbarIconButton } from '@leafygreen-ui/toolbar';
 import { Body } from '@leafygreen-ui/typography';
 
+import { drawerTransitionDuration } from './Drawer/Drawer.styles';
 import { DisplayMode, Drawer, DrawerProps } from './Drawer';
 import { DrawerStackProvider } from './DrawerStackContext';
+import { DrawerWithToolbarWrapper } from './DrawerWithToolbarWrapper';
 import { EmbeddedDrawerLayout } from './EmbeddedDrawerLayout';
-import { Toolbar, ToolbarIconButton } from '@leafygreen-ui/toolbar';
-import { drawerTransitionDuration } from './Drawer/Drawer.styles';
+import { OverlayDrawerLayout } from './OverlayDrawerLayout';
 
 const SEED = 0;
 faker.seed(SEED);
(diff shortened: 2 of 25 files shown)


packages/drawer/README.md
```diff
@@ -142,6 +176,10 @@ function ExampleComponent() {
 }
 ```
 
+### Embedded Drawer with Toolbar
+
+// TODO: in separate PR
```
shaneeza, line 181: #2865

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { Drawer } from '../Drawer/Drawer';
+import { DisplayMode } from '../Drawer/Drawer.types';
+import { DrawerStackProvider } from '../DrawerStackContext';
+import { DrawerWithToolbarWrapper } from '../DrawerWithToolbarWrapper';
+import { EmbeddedDrawerLayout } from '../EmbeddedDrawerLayout';
+import { OverlayDrawerLayout } from '../OverlayDrawerLayout';
+
+import { DrawerToolbarLayoutContainerProps } from './DrawerToolbarLayout.types';
+
+// Dummy content for now
```
shaneeza, line 21: #2865

packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -17,15 +17,50 @@ import { DisplayMode } from './Drawer.types';
...
-const getBaseStyles = ({ open, theme }: { open: boolean; theme: Theme }) => css`
+export const drawerClassName = createUniqueClassName('lg-drawer');
+
+// Because of .show() and .close() in the drawer component, transitioning from 0px to (x)px does not transition correctly. Having the drawer start at the open position while hidden, moving to the closed position, and then animating to the open position is a workaround to get the animation to work.
+// These styles are used for a standalone drawer in overlay mode since it is not part of a grid layout.
+const drawerIn = keyframes`
+  0% {
+    transform: translate3d(0%, 0, 0);
+    opacity: 0;
+    visibility: visible; // Adding visibility: hidden break autoFocus of the close button when using the dialog component
```
shaneeza, line 28: FIXME: This is buggy in safari 😭

shaneeza: I'm going to work on this in another PR. https://642700aa3e49e32bdbf0b0fc-raetxdvjcx.chromatic.com/?path=/story/components-drawer--light-mode-overlay

try refreshing a few times and you'll notice that the drawer is opening in add positions

packages/drawer/src/Drawer.stories.tsx
```diff
@@ -33,9 +42,11 @@ export default {
       <div
         className={css`
           height: 100%;
-          width: 100%;
           display: flex;
           align-items: center;
+          margin: -100px;
+          width: 100vw;
+          border-bottom: 3px solid ${palette.green.base}; // Used to visualize the height of the parent container
```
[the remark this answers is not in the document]
shaneeza: I'm open to suggestions, but I wanted a way to showcase that the drawer is constricted to the height of the container. It's really just a random border color so I can remove it if its awkward.

packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -17,15 +17,50 @@ import { DisplayMode } from './Drawer.types';
...
+  }
+  100% {
+    opacity: 0;
+    visibility: hidden;
+  }
+`;
+
+const getBaseStyles = ({ theme }: { theme: Theme }) => css`
   all: unset;
+  padding: 0;
```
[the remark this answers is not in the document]
shaneeza: very true! i'll update

packages/drawer/src/Drawer/Drawer.constants.ts
```diff
@@ -1,3 +1,4 @@
 export const HEADER_HEIGHT = 48;
 export const MOBILE_BREAKPOINT = 390;
 export const PANEL_WIDTH = 432;
+export const TOOLBAR_WIDTH = 48;
```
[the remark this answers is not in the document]
shaneeza: ~Are you saying that I should create a root `constants.ts` file?~ nvm I get it!  I could also export this constant from `Toolbar`.

packages/drawer/src/Drawer/Drawer.styles.ts
shaneeza: Update: I'm going to continue using `visibility: hidden`. This breaks the auto-focus on a dialog element, but I was able to add the focus back with JavaScript. I also tested in VoiceOver, and it was able to pick up the dialog content despite `visibility: hidden`.

From slack:

> I haven't found documents that flat-out say not to use display: none' or visibility: hidden; most of the documents only mention that the [dialog should autofocus the first focusable item it finds](https://html.spec.whatwg.org/multipage/interactive-elements.html#dialog-focusing-steps). Here's [another doc on focus](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element). And [w3](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/), although this one is about a dialog modal.

> The most significant diff I see is that if a dialog is open on initial load, VoiceOver will automatically read the content below the title without visibility hidden. With visibility hidden, you have to manually press the down arrow, and then it will begin reading the content below the title. Besides that, everything else is the same. When you click toggle, the drawer opens, the focus is on the close button and VoiceOver announces that your focus is on a close button inside a dialog. You then have to manually press the down arrow to have VoiceOver read the content.

packages/drawer/src/Drawer.stories.tsx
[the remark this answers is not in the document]
shaneeza: Very true. I'll remove from `liveExample`.

packages/drawer/src/utils/getTestUtils.tsx
```diff
@@ -42,14 +42,10 @@ export const getTestUtils = <T extends HTMLElement = HTMLElement>(
...
-    const { display, opacity, transform, visibility } =
-      window.getComputedStyle(element);
+    const { display, opacity, visibility } = window.getComputedStyle(element);
 
     const isCSSVisible =
-      display !== 'none' &&
-      opacity === '1' &&
-      transform === 'none' &&
-      visibility !== 'hidden';
+      display !== 'none' && opacity === '1' && visibility !== 'hidden';
```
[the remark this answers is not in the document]
shaneeza: good point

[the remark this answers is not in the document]
shaneeza: I realize checking the transform value is unreliable since I use a keyframe and getComputedStyle does not pick up the end styles, plus in the `DrawerToolbarLayout`, the Drawer itself is not transitioning. `toBeVisible` from jest only checks `display`, `visibility`, and `opacity`.  Since the transform values are fragile, do you see a reason we should continue to check the transform value in this utility? 

One thought is to maybe utilize `onAnimationEnd` to update a boolean data attribute on Drawer, like `data-open`.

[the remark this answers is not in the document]
shaneeza: Side note but this also made me realize that I wasn't testing the [DrawerToolbarLayout with all the test utils](https://github.com/mongodb/leafygreen-ui/pull/2865/files#diff-4d35bd4cedea2f041cce5e80f33db1ed2b7692c320cb5b4e2dc610a5327f9629) correctly in my other PR, so I'll update that

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -93,43 +92,40 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
                 })}
               >
                 <Body
                   as={typeof title === 'string' ? 'h2' : 'div'}
                   baseFontSize={BaseFontSize.Body2}
                   id={titleId}
-                  weight="medium"
                 >
-                  {title}
+                  <strong>{title}</strong>
```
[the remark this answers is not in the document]
shaneeza: Since this is a pattern used in other components, and [we currently support](https://github.com/mongodb/leafygreen-ui/blob/main/packages/typography/src/Body/Body.tsx#L35-L41), I'd rather keep using it and create a ticket to address all instances and tackle it in another PR once we make a decision.

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const LayoutComponent =
+    displayMode === DisplayMode.Overlay
+      ? OverlayDrawerLayout
+      : EmbeddedDrawerLayout;
+
+  const layoutProps = {
+    hasToolbar: true,
+    isDrawerOpen: displayMode === DisplayMode.Embedded ? open : false,
+  };
```
stephl3: Should this be a separate `DrawerLayout` component for consumers who may want to use it without a `Toolbar`? My thought is consumers would only be using `DrawerLayout` rather than the separate `OverlayDrawerLayout` or `EmbeddedDrawerLayout`. Does that sound right to you?

packages/drawer/src/EmbeddedDrawerLayout/EmbeddedDrawerLayout.types.ts
```diff
@@ -5,4 +5,9 @@ export interface EmbeddedDrawerLayoutProps extends HTMLElementProps<'div'> {
    * Determines if the Drawer instance is open or closed
    */
   isDrawerOpen: boolean;
+
+  /**
+   * Determines if the Toolbar is present in the layout
+   */
+  hasToolbar?: boolean;
```
stephl3: since `hasToolbar` is used by both `EmbeddedDrawerLayout` and `OverlayDrawerLayout`, can we move this into a separate `BaseDrawerLayout` interface and extend on that?

packages/drawer/src/EmbeddedDrawerLayout/EmbeddedDrawerLayout.styles.ts
```diff
@@ -1,38 +1,67 @@
...
+import { GRID_AREA } from '../constants';
+import { PANEL_WIDTH, TOOLBAR_WIDTH } from '../constants';
+import { MOBILE_BREAKPOINT } from '../Drawer';
+import { drawerTransitionDuration } from '../Drawer/Drawer.styles';
 
 const baseStyles = css`
   width: 100%;
   display: grid;
   grid-template-columns: auto 0;
+  transition: all ${drawerTransitionDuration}ms ease-in-out;
```
stephl3: can we specify the transition properties to avoid negative perf impact and make this more explicit about what is transitioning?

packages/drawer/src/EmbeddedDrawerLayout/EmbeddedDrawerLayout.styles.ts
```diff
@@ -1,38 +1,67 @@
...
   @media only screen and (max-width: ${MOBILE_BREAKPOINT}px) {
-    grid-template-columns: unset;
     grid-template-rows: 50% 50%;
   }
 `;
 
+const withToolbarBaseStyles = css`
+  grid-template-columns: auto ${TOOLBAR_WIDTH}px;
+  grid-template-areas: '${GRID_AREA.content} ${GRID_AREA.drawer}';
+`;
```
stephl3, line 40: some of these styles are identical with styles defined in `OverlayDrawerLayout.styles.ts`. If we are making a main `DrawerLayout` component, what do you think about having a DrawerLayout.styles.ts that has shared styles. This will be more DRY and could be more easily identifiable how the 2 layouts are similar or different

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+export const DrawerToolbarLayoutContainer = ({
+  children,
+  displayMode = DisplayMode.Embedded,
+  data,
+  onClose,
+  darkMode: darkModeProp,
+}: DrawerToolbarLayoutContainerProps) => {
+  const { darkMode } = useDarkMode(darkModeProp);
+  const [open, setOpen] = useState(false);
```
stephl3: Do we need to expose the `open` prop in this DrawerToolbarLayoutContainer`? Wondering about a case where there's a hyperlink that navigates to drawer content of a specific active toolbar icon

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const handleOnClose = (event: React.MouseEvent<HTMLButtonElement>) => {
+    onClose?.(event);
+    setOpen(false);
+
+    // TODO: add call to context to update the drawer content
```
stephl3: is this addressed in a follow-up PR?

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+  return (
+    <LeafyGreenProvider darkMode={darkMode}>
+      <DrawerStackProvider>
+        <LayoutComponent {...layoutProps}>
+          <div
+            className={css`
+              grid-area: ${GRID_AREA.content};
+              overflow: scroll;
+              height: inherit;
+            `}
```
stephl3: nit: break out into separate var

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const DrawerToolbarLayout = ({
+  children,
+  ...rest
+}: DrawerToolbarLayoutProps) => {
+  return (
+    <DrawerToolbarLayoutContainer {...rest}>
+      {children}
+    </DrawerToolbarLayoutContainer>
+  );
+};
```
stephl3: could you share the purpose of having `DrawerToolbarLayoutContainer`? looks like there's little going on in this file that it could be combined into `DrawerToolbarLayout` but curious if there's something I'm missing

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+          >
+            {children}
+          </div>
+          <DrawerWithToolbarWrapper
+            displayMode={displayMode}
+            isDrawerOpen={open}
+          >
+            <Toolbar>
+              {data?.map(toolbar => (
```
stephl3: is `toolbar` the right naming here? it looks like it has `content` and `title` as well. alternatively, maybe it could be 2 separate root keys that then have nested keys relating to what they're used for

```
data: {
  drawer: { content, title },
  toolbar: { ... },
}
```

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: With the current api, the only way to open a drawer with a toolbar [is to call the `openDrawer` hook](https://github.com/mongodb/leafygreen-ui/pull/2865/files#diff-733f2a760f041bf05129bf53f9f893ce03179ce087309e4b15034d8cf98362d3R407).  If `open` is exposed, we wouldn't know what content to render.

`openDrawer('Code')` instead of `setOpen(true)`

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: [Yes!](https://github.com/mongodb/leafygreen-ui/pull/2865/files#diff-3e0c621f914d3e37e605c973d22fb713ce77afacae2c88d08f5094049c9d2387)

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout.tsx
shaneeza: [It makes more sense in this PR](https://github.com/mongodb/leafygreen-ui/pull/2865/files#diff-3e0c621f914d3e37e605c973d22fb713ce77afacae2c88d08f5094049c9d2387), but it allows me to use `useDrawerToolbarContext()`. I'm not able to use the hook in the same component as the provider.

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: `ToolbarItem` might be a better name. 

Also, I'm not sure I understand the separate key approach. This is what the current data looks like:
```
const DRAWER_TOOLBAR_DATA: DrawerToolbarLayoutProps['data'] = [
  {
    id: 'Code',
    label: 'Code',
    content: <DrawerContent />,
    title: 'Code Title',
    glyph: 'Code',
    onClick: () => {
      console.log('Code clicked');
    },
  },
  {
    id: 'Dashboard',
    label: 'Dashboard',
    content: <DrawerContent />,
    title: 'Dashboard Title',
    glyph: 'Dashboard',
    onClick: () => {
      console.log('Dashboard clicked');
    },
  },
  {
    id: 'Plus',
    label: "Perform some action, doesn't open a drawer",
    glyph: 'Plus',
    onClick: () => {
      console.log('Plus clicked, does not update drawer');
    },
  },
];
```

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: I decided to keep them separate for a few reasons:

1.  Since `EmbeddedDrawerLayout` is already a published layout, I wanted to keep it as-is.
2. The use of a standalone drawer is now discouraged. I think making `DrawerLayout` would encourage a standalone drawer because it's easier to switch between the two layouts. Does that make sense?

packages/drawer/src/EmbeddedDrawerLayout/EmbeddedDrawerLayout.styles.ts
shaneeza: I like that. Even if I don't end up making the `DrawerLayout`, I can still have shared styles.

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: If a hyperlink is supposed to open a drawer, that can still be done by reading the parameters and calling the hook with the correct ID.

packages/drawer/src/Drawer.stories.tsx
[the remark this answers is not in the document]
stephl3: not really sure on an alternative for that, but I think it might be covered in that it will be restricted to the live example section on .design

packages/drawer/src/utils/getTestUtils.tsx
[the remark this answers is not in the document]
stephl3: I believe `transform` was included because a drawer instance could be "visible" if those 3 other properties were correctly set, but the translateX value could position it off the screen. I don't think a data attribute would solve that since you could still have a `transform` value that positions it off the screen even if `data-open={true}`

What value does `getComputedStyle` return? Can we check that it's _not_ the starting style rather than looking for a specific ending style? Alternatively, should we make this async?

packages/drawer/package.json
```diff
@@ -25,6 +25,7 @@
     "@leafygreen-ui/polymorphic": "workspace:^",
     "@leafygreen-ui/tabs": "workspace:^",
     "@leafygreen-ui/tokens": "workspace:^",
+    "@leafygreen-ui/toolbar": "workspace:^",
```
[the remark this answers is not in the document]
stephl3: Consumers that use a standalone drawer will have the toolbar dep bundled into their code even if it's not used. It seems like an odd pattern to me, but not something I'll block on

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
stephl3: maybe my question was posed incorrectly. I'm not too concerned about toolbar or no toolbar. I think we previously discussed that `DrawerLayout` is a singleton pattern. shouldn't it be exported as a combined `DrawerLayout` rather than as the separate `OverlayDrawerLayout` and `EmbeddedDrawerLayout`? are consumers allowed to use 1 or multiple of both?

re: `EmbeddedDrawerLayout`, it wasn't exported from the root index file so it's not available to consumers yet

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout.tsx
stephl3: gotcha, that does make more sense. as feedback for the future, it'd be helpful to break up the problem in a different way to more contextualize code decisions without needing to jump between PRs

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
stephl3: oops yeah `toolbarItem`. something like this:
```
data?.map(({ id, drawer, toolbarItem }) => ...);
```

and then maybe something like:
- `drawer` would have the keys of `title` and `content`
- `toolbarItem` would have the keys of `label`, `glyph`, `onClick`

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: > re: EmbeddedDrawerLayout, it wasn't exported from the root index file so it's not available to consumers yet


 I missed that, thanks for pointing it out.

shaneeza: I'm not fully remembering that conversation, but if we discussed it, then I don't mind updating it. So the pattern would be:

- `DrawerToolbarLayout` - Drawer + Toolbar, can choose between `overlay` or `embedded`
- `DrawerLayout` - Drawer without a Toolbar, can choose between `overlay` or `embedded`

Does that happen to sound right to you?

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: I see. This is clearer. I would prefer to make these changes in a new PR since they will affect the other work in the Provider PR.

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
stephl3: yes, and I'll also say that I'm not necessarily saying we _need_ to have a `DrawerLayout`. I'm only saying that I think it makes more sense to export `DrawerLayout` rather than export the separate `OverlayDrawerLayout` + `EmbeddedDrawerLayout`

however, I guess it's one thing to discourage usage without a toolbar vs not exporting `DrawerLayout` altogether which means the only option is with a toolbar

packages/drawer/src/index.ts
```diff
@@ -3,4 +3,16 @@ export {
...
+  type DrawerToolbarLayoutProps,
+} from './DrawerToolbarLayout';
+export {
+  EmbeddedDrawerLayout,
+  type EmbeddedDrawerLayoutProps,
+} from './EmbeddedDrawerLayout';
+export {
+  OverlayDrawerLayout,
+  type OverlayDrawerLayoutProps,
+} from './OverlayDrawerLayout';
```
stephl3: proactively approving but let's get the final exports right!

packages/drawer/src/utils/getTestUtils.tsx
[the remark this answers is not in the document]
shaneeza: I was thinking on animationEnd, check that the end transition has the correct transform value, however this might not register immediately when the drawer opens. 

> What value does getComputedStyle return? Can we check that it's not the starting style rather than looking for a specific ending style? Alternatively, should we make this async?

getComputedStyle keeps returning the starting style. I've tried making it async, along with adding the animationend event listener with a timeout in the test but it's not registering in the test.

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: As mentioned in dev sync, I will create a `DrawerLayout` wrapper that can be used in both overlay and embedded modes, with or without a toolbar

packages/drawer/src/EmbeddedDrawerLayout/EmbeddedDrawerLayout.styles.ts
```diff
@@ -10,7 +10,10 @@ const baseStyles = css`
   width: 100%;
   display: grid;
   grid-template-columns: auto 0;
-  transition: all ${drawerTransitionDuration}ms ease-in-out;
+  transition-property: grid-template-columns, grid-template-rows;
+  transition-timing-function: ease-in-out, ease-in-out;
+  transition-duration: ${drawerTransitionDuration}ms,
+    ${drawerTransitionDuration}ms;
```
stephl3: since they're using the same timing-function and duration, you only need to list them once. you can list them comma-separated if they're all blocked together with the `transition` property

packages/drawer/src/DrawerLayout/DrawerLayout.tsx
```diff
@@ -0,0 +1,51 @@
...
+      onClose,
+      children,
+      displayMode = DisplayMode.Overlay,
+      darkMode,
+      ...rest
+    }: DrawerLayoutProps,
+    forwardedRef,
+  ) => {
+    // If there is data, we render the DrawerToolbarLayout.
+    if (data) {
```
stephl3: now that this covers both cases (with a toolbar and without), is there a more explicit name than `data` that can be used? it isn't clear that `data` is only intended to be used when looking to render the toolbar with the layout

packages/drawer/src/DrawerContext/DrawerContext.tsx
```diff
@@ -1,21 +1,47 @@
...
+   * Boolean indicating if the drawer is currently open.
+   */
+  isDrawerOpen: boolean;
 }
 
-export const DrawerContext = createContext<DrawerContextProps | null>(null);
+export const DrawerContext = createContext<DrawerContextProps>({
+  setIsDrawerOpen: () => {},
+  isDrawerOpen: false,
+});
```
stephl3: This `DrawerContext` was separate from `DrawerStackContext` because it was intended to have a 1:1 relationship with each drawer instance. If you only need to keep track of `isDrawerOpen` within a layout of multiple drawer instances, we should remove this context and add this logic to `DrawerStackContext`

packages/drawer/src/DrawerLayout/DrawerLayout.tsx
```diff
@@ -0,0 +1,51 @@
...
+      );
+    }
+
+    // If there is no data, we render the LayoutComponent.
+    // The LayoutComponent will handle the displayMode and render the appropriate layout.
+    return (
+      <LayoutComponent ref={forwardedRef} {...rest} displayMode={displayMode}>
+        {children}
+      </LayoutComponent>
+    );
```
stephl3: should there be a `DrawerStackProvider` and `LeafyGreenProvider` wrapping this?

stephl3: If so, it also makes me wonder if the returned JSX of `DrawerLayout` can be organized differently for maintainability. rather than forking based on existence of `data`, what if there was a boolean var tracking that and we forked with rendering the `Toolbar` instance. In other words, `DrawerLayout` would replace `DrawerToolbarLayout`

Another issue with the current organization is that it's not clear that changes to `DrawerToolbarLayout` might also require changes to `LayoutComponent` and vice-versa, especially since `LayoutComponent` is used in `DrawerToolbarLayout`

packages/drawer/src/DrawerContext/DrawerContext.tsx
shaneeza: Spoke in person, i'm going to remove the DrawerContext and use a prop, `isDrawerOpen`

packages/drawer/src/DrawerLayout/DrawerLayout.tsx
shaneeza: I was thinking about our conversation, and I like the idea of making this component more generic by rendering only one `LayoutComponent` and conditionally rendering the children. However, the `LayoutComponent` inside the `DrawerToolbarLayout` depends on data from the toolbar context, and because of that, I'd prefer to keep the `LayoutContainer` inside the `DrawerToolbarLayout`. This way, only the `DrawerToolbarLayout` is aware of the toolbar context.

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -93,43 +92,40 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+        <Component
+          aria-hidden={!open}
+          aria-labelledby={titleId}
+          className={getDrawerStyles({
+            theme,
+            open,
+            shouldAnimate,
+            className,
+            displayMode,
+            zIndex: 1000 + drawerIndex,
```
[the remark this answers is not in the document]
shaneeza: I believe we try to stay away from style-specific props. If consumers want, they can update this with CSS.

PR #2865, opened 2025-05-22
LG-5046: Drawer provider

shaneeza:
### What's included:
- [LG-5045](https://jira.mongodb.org/browse/LG-5045) - Updating test harnesses to include `Toolbar` test utils.
- [LG-5046](https://jira.mongodb.org/browse/LG-5046) - Creating `DrawerToolbarProvider`. This includes the `useDrawerToolbarContext` hook that consumers can utilize to update the drawer content.
- Reorganizes folder structure so that test utils are in the `testing` directory.
- Updates `Toolbar`'s `getLgIds` util to use the correct root name

### What's not included:
- Changesets - will add in a separate PR

### For new components


packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -1,88 +1,68 @@
...
+  displayMode = DisplayMode.Embedded,
+  'data-lgid': dataLgId = DEFAULT_LGID_ROOT,
   darkMode: darkModeProp,
 }: DrawerToolbarLayoutContainerProps) => {
   const { darkMode } = useDarkMode(darkModeProp);
-  const [open, setOpen] = useState(false);
+  const { openDrawer, closeDrawer, getActiveDrawerContent, shouldCloseDrawer } =
+    useDrawerToolbarContext();
+  const isDrawerOpen = shouldCloseDrawer;
+  const { id, title, content } = getActiveDrawerContent() || {};
```
tsck, line 36: Nit: I'm wondering if `getActiveDrawerContent()` should return better defaults. In the current case when `content` is `undefined` in `DrawerToolbarContext`, this pattern seems a bit odd to me `const { id, title, content } = {};`. This isn't critical and what you have totally works, I'm just wondering if there's a clearer way to structure this.

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -1,88 +1,68 @@
...
+  const lgIds = getLgIds(dataLgId);
+
+  // If there is no data, we don't want to render anything
+  if (!data || data.length === 0) return null;
```
tsck, line 40: Q: If there's no data, should we even run `getActiveDrawerContent()` and `getLgIds()` above?

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -1,88 +1,68 @@
...
   const LayoutComponent =
     displayMode === DisplayMode.Overlay
       ? OverlayDrawerLayout
       : EmbeddedDrawerLayout;
 
+  // We want to pass the isDrawerOpen prop to the LayoutComponent only if the display mode is embedded
   const layoutProps = {
     hasToolbar: true,
-    isDrawerOpen: displayMode === DisplayMode.Embedded ? open : false,
+    isDrawerOpen: displayMode === DisplayMode.Embedded ? isDrawerOpen : false,
```
tsck, line 56: Nit: I believe this is logically equivalent to just

```suggestion
    isDrawerOpen: displayMode === DisplayMode.Embedded && isDrawerOpen,
```

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: Thinking about this, since consumers are passing in all their content to this component, I shouldn't return null if there isn't data. Instead, I'll prevent the Drawer and Toolbar from rendering while still allowing the main content to render.

tsck (approved):
LGTM! Just the one nit [here](https://github.com/mongodb/leafygreen-ui/pull/2865/files#r2114275849), but definitely not a blocker so going to approve.

PR #2874, opened 2025-05-28
LG-5073: Add tooltip support to CodeEditor

tsck:
## ✍️ Proposed changes
- Adds tooltip support to `CodeEditor`.
- Adds a number of additional test utils.
- Exports some missed props from previous branch.
- Updates README to define additional props, types, and utils.

🎟 _Jira ticket:_ [LG-5073](https://jira.mongodb.org/browse/LG-5073)

### For new components

## 🧪 How to test changes
- New Tooltip on Hover story in Storybook

### DIFF
--- a/packages/code-editor/src/CodeEditor.stories.tsx
+++ b/packages/code-editor/src/CodeEditor.stories.tsx
@@ -10,6 +10,39 @@ import { css } from '@leafygreen-ui/emotion';
 import { IndentUnits } from './CodeEditor/CodeEditor.types';
 import { CodeEditor } from '.';
 
+const MyTooltip = ({ line, column }: { line: number; column: number }) => {
+  return (
+    <div
+      className={css`
+        padding: 8px;
+      `}
+    >
+      <div
+        className={css`
+          font-weight: bold;
+          margin-bottom: 4px;
+        `}
+      >
+        My Tooltip
+      </div>
+      <div
+        className={css`
+          margin-bottom: 4px;
+        `}
+      >
+        Line: {line}
+      </div>
+      <div
+        className={css`
+          margin-bottom: 4px;
+        `}
+      >
+        Column: {column}
+      </div>
+    </div>
+  );
+};
+
 const meta: StoryMetaType<typeof CodeEditor> = {
   title: 'Components/CodeEditor',
   component: CodeEditor,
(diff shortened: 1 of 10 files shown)


packages/code-editor/src/CodeEditor.stories.tsx
```diff
@@ -44,6 +78,14 @@ const meta: StoryMetaType<typeof CodeEditor> = {
     readOnly: false,
     indentSize: 2,
     indentUnit: IndentUnits.Space,
+    tooltips: [
```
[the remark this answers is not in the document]
tsck: This is a bit of a problem. It appears that [hover tooltips in multiple locations is not easily achievable](https://discuss.codemirror.net/t/multiple-hovertooltips/7201). I don't actually recall why I had this in scope and what the use case was. I'm reaching out to Sandy to see if she recalls, and will reach out to stakeholders today to see if either a) just having one at a time covers the use case, or b) if this isn't even actually required. Will update shortly.

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -102,6 +104,7 @@ export const CodeEditor = forwardRef<CodeMirrorRef, CodeEditorProps>(
       const hyperLinkCompartment = new Compartment();
       const lineWrappingCompartment = new Compartment();
       const indentExtensionCompartment = new Compartment();
+      const tooltipCompartment = new Compartment();
```
TheSonOfThomp, line 107: Super nit for later, but we may want to consider a data structure to contain all these compartments?

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
```diff
@@ -123,4 +163,9 @@ export interface CodeEditorProps extends DarkModeProps {
    * Enables read only mode, making the contents uneditable.
    */
   readOnly?: boolean;
+
+  /**
+   * Add tooltips to the editor content.
+   */
+  tooltips?: Array<CodeEditorTooltip>;
```
TheSonOfThomp, line 170: If we're only officially supporting one tooltip, we could potentially type this as a 1D tuple: `[CodeEditorTooltip]`

packages/code-editor/src/CodeEditor/utils/createTooltipExtension.ts
```diff
@@ -0,0 +1,61 @@
...
+        const to = from + length;
+
+        let message = '';
+        let renderMessage: (() => HTMLElement) | undefined;
+
+        if (typeof content === 'string') {
+          message = content;
+        } else if (React.isValidElement(content) || Array.isArray(content)) {
+          renderMessage = () => {
+            const dom = document.createElement('div');
```
TheSonOfThomp, line 39: Would `React.createElement(div)` work here instead?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: I like that. Will definitely update this later on!

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
tsck: That's fair. I'll drive-by this in a later PR as well

packages/code-editor/src/CodeEditor/utils/createTooltipExtension.ts
tsck: Ultimately this does have to return an `HTMLElement`, so I don't believe so.

PR #2885, opened 2025-06-02
[LG-5143] fix(chat input-bar): control state with textareaProps

stephl3:
## ✍️ Proposed changes

- `textareaProps.value` was previously initialized in internal state as `messageBody` but changes through `onChange` could not call `setMessageBody`

🎟 _Jira ticket:_ [LG-5143](https://jira.mongodb.org/browse/LG-5143)

## 🧪 How to test changes

- unit tests

### DIFF
--- a/chat/input-bar/src/InputBar/InputBar.spec.tsx
+++ b/chat/input-bar/src/InputBar/InputBar.spec.tsx
@@ -9,7 +9,13 @@ import { InputBar } from '.';
 const testText = 'test';
 
 describe('packages/input-bar', () => {
-  test('onChange is fired when user types', () => {
+  test('renders `badgeText` when the prop is set', () => {
+    render(<InputBar badgeText="beta" />);
+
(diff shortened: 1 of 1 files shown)


chat/input-bar/src/InputBar/InputBar.spec.tsx
```diff
@@ -130,4 +130,50 @@ describe('packages/input-bar', () => {
...
+      />,
+    );
+
+    const textarea = screen.getByRole('textbox');
+    const sendButton = screen.getByRole('button');
+
+    expect(textarea).toHaveValue(controlledValue);
+    userEvent.click(sendButton);
+
+    // The value should remain unchanged after submission in controlled mode
```
TheSonOfThomp: Do we have a test that asserts that the value _does_ change when uncontrolled?

stephl3: added!

PR #2887, opened 2025-06-04
LG-4217: retain className styles for tooltip triggers passed as functions

charcoalyy:
## ✍️ Proposed changes

Originally, tooltip triggers passed as functions may have styles overwritten if they use class names. Tooltip triggers passed as functions are now called to get a reference to the component's original class names, which are then applied as a combination rather than an overwrite.

🎟 _Jira ticket:_ [Tooltip trigger requires implicit pass-through className](https://jira.mongodb.org/browse/LG-4217])

### For new components

### DIFF
--- a/packages/tooltip/src/Tooltip/Tooltip.spec.tsx
+++ b/packages/tooltip/src/Tooltip/Tooltip.spec.tsx
@@ -429,13 +429,13 @@ describe('packages/tooltip', () => {
   });
 
   describe('when trigger is an inline function', () => {
-    function renderInlineTrigger(props = {}, trigger: any = undefined) {
+    function renderInlineTrigger(props = {}, customTrigger: any = undefined) {
       const utils = render(
         <>
           <div data-testid="backdrop" />
           <Tooltip
             trigger={
-              trigger ??
+              customTrigger ??
               (({ children, ...rest }: ButtonTestProps) => (
                 <button {...rest} data-testid="inline-trigger">
                   {buttonText}
@@ -472,22 +472,24 @@ describe('packages/tooltip', () => {
       await waitForElementToBeRemoved(tooltip);
     });
 
-    test(`retains class names of trigger component`, () => {
-      const styledTrigger = ({ children, ...rest }: ButtonTestProps) => (
+    test(`retains existing class names of trigger component`, () => {
+      const TEST_CLASS_NAME = 'test-class-name';
+
+      const customTrigger = ({ children, ...rest }: ButtonTestProps) => (
         <button
-          {...rest}
-          className="styled-trigger"
+          className={TEST_CLASS_NAME}
           data-testid="inline-trigger"
+          {...rest}
         >
           {buttonText}
           {children}
         </button>
       );
 
-      const { button } = renderInlineTrigger({}, styledTrigger);
+      const { button } = renderInlineTrigger({}, customTrigger);
 
       expect(button).toBeInTheDocument();
-      expect(button).toHaveClass('styled-trigger');
+      expect(button).toHaveClass(TEST_CLASS_NAME);
     });
   });
 
--- a/packages/tooltip/src/Tooltip/Tooltip.spec.tsx
+++ b/packages/tooltip/src/Tooltip/Tooltip.spec.tsx
@@ -429,7 +429,10 @@ describe('packages/tooltip', () => {
   });
 
   describe('when trigger is an inline function', () => {
-    function renderInlineTrigger(props = {}, customTrigger: any = undefined) {
+    function renderInlineTrigger(
+      props = {},
+      customTrigger?: (props: ButtonTestProps) => ReactElement,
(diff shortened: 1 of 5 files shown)


packages/tooltip/src/Tooltip/Tooltip.spec.tsx
```diff
@@ -429,17 +429,23 @@ describe('packages/tooltip', () => {
   });
 
   describe('when trigger is an inline function', () => {
-    function renderInlineTrigger(props = {}) {
+    function renderInlineTrigger(
+      props = {},
+      customTrigger?: (props: ButtonTestProps) => ReactElement,
```
charcoalyy: wanted to increase the usability of this utility, since this bug requires modifications to the trigger for testing (see L483) - would love to hear if there are better options to accomplish this? :)

packages/tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -292,25 +292,21 @@ function Tooltip({
...
-      return trigger({
-        ...createTriggerProps(triggerEvent),
-        className: positionRelative,
-        'aria-describedby': active ? tooltipId : undefined,
-        children: tooltip,
-      });
-    }
+    const originalTrigger =
+      typeof trigger === 'function'
+        ? (trigger({}) as React.ReactElement<unknown>)
```
tsck: One thing I'm wondering here, is if we could update the type to be more explicit in the types file so a type assertion isn't needed here. This only seems to be called as a function here, so I believe we could type this as: `(props: Record<string, unknown>) => ReactElement` or if we're scared that it will break consumer code downstream, maybe `(props: any) => ReactElement` would suffice. 

No need to make this change now in this PR because this isn't a blocker and I don't have enough context as to why `trigger` was typed as a generic `Function` in the first place. I'm going to check in with the team and we could always do a drive-by fix later to tighten up the type if it makes sense.

tsck (approved):
One small comment we can check in on later, otherwise LGTM!

Rocket Thumb

packages/tooltip/src/Tooltip/Tooltip.spec.tsx
tsck: This definitely works and I think it's the correct thinking! However, since this already takes a `props` arg which is essentially of type `TooltipProps`, did you consider just passing the `trigger` in via the `props` object? This way we wouldn't have to add an additional argument to this util and can still customize the `trigger`. We would just want to make sure our `trigger` prop override comes after where were spread the props object.

```tsx
function renderInlineTrigger(
  props: TooltipProps = {} // Added type of TooltipProps
) {
  const utils = render(
    <>
      <div data-testid="backdrop" />
      <Tooltip
        {...props} // Props spread before overriding trigger
        trigger={
          props.trigger ??
          (({ children, ...rest }: ButtonTestProps) => (
            <button {...rest} data-testid="inline-trigger">
              {buttonText}
              {children}
            </button>
          ))
        }
      >
        <div data-testid={tooltipTestId}>Tooltip Contents!</div>
      </Tooltip>
    </>,
  );
  const button = utils.getByTestId('inline-trigger');
  const backdrop = utils.getByTestId('backdrop');
  return { ...utils, button, backdrop };
}
```

An alternative could be to just do process props before the render, to make it a bit more readable:
```tsx
function renderInlineTrigger(
  props: TooltipProps = {} // Added type of TooltipProps
) {
  props.trigger =
    props.trigger ??
    (({ children, ...rest }: ButtonTestProps) => (
      <button {...rest} data-testid="inline-trigger">
        {buttonText}
        {children}
      </button>
    ));

  const utils = render(
    <>
      <div data-testid="backdrop" />
      <Tooltip {...props}>
        <div data-testid={tooltipTestId}>Tooltip Contents!</div>
      </Tooltip>
    </>,
  );
  const button = utils.getByTestId('inline-trigger');
  const backdrop = utils.getByTestId('backdrop');
  return { ...utils, button, backdrop };
}
```

packages/tooltip/src/Tooltip/Tooltip.tsx
tsck: OK, I checked in with the team and there doesn't appear to be any reason we made the type a generic `Function`. I think we should give updating this a shot since we haven't merged this yet. 

So instead of doing the type assertion here, I think we should update the type in the types file. `React.FC<any>` is almost what we want, but that returns a `ReactNode` which can also be a string, null, etc., where we expect this function to explicitly return a `ReactElement`. Due to this, I think we'll need to make our own type definition. I'm think `(props: any) => ReactElement` is open enough while still narrowing down the return type. How do you feel about that?

charcoalyy: that makes sense!! awesome, i'll get on that

tsck (approved):
LGTM! You just need to pull down `main` before you merge, but looks great!

PR #2890, opened 2025-06-04
LG-5064: resolve outstanding TS TODOs

TheSonOfThomp:
## ✍️ Proposed changes

[LG-5064](https://jira.mongodb.org/browse/LG-5064)

`@lg-charts/core`: patch
---

Updates axis `formatter` prop signature to match echarts.
```ts
{
  formatter?: (value: number, index?: number) => string | string;
}
```

`@leafygreen-ui/polymorphic`: patch
---

Replace `ReactElement` with `ReactNode` in type definitions

### DIFF
--- a/packages/code/src/CopyButton/CopyButton.spec.tsx
+++ b/packages/code/src/CopyButton/CopyButton.spec.tsx
@@ -192,6 +192,23 @@ describe('CopyButton', () => {
     );
   });
 
-  // TODO: get this to work https://jira.mongodb.org/browse/LG-4760
-  test.todo('copies the correct text when copy button is clicked');
+  test('copies the correct text when copy button is clicked', async () => {
+    const onCopy = jest.fn();
+    const clipboardTextFn = jest.fn().mockReturnValue(contents);
+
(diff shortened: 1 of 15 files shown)


packages/code/src/CopyButton/CopyButton.spec.tsx
```diff
@@ -192,6 +192,23 @@ describe('CopyButton', () => {
...
+
+    const { getByTestId, getByRole } = renderCopyButton({ onCopy });
+    const copyButton = getByTestId(testIds.copyButton);
+
+    userEvent.click(copyButton);
+
+    await waitFor(() => {
+      expect(onCopy).toHaveBeenCalledTimes(1);
+      expect(ClipboardJS).toHaveBeenCalled();
+      expect(clipboardTextFn()).toBe(contents);
```
shaneeza: Would this still return the contents even if the copy button was not clicked?

TheSonOfThomp: that's a good point. Copilot wrote this test, and it passed so I didn't look too hard at it

shaneeza (approved):

PR #2891, opened 2025-06-05
LG-5035: prevent consumer initial state from being overwritten

charcoalyy:
## ✍️ Proposed changes

Checks if consumer has set initialState, and if so, make that the default value of the managed expanded state.

🎟 _Jira ticket:_ [LG-5035](https://jira.mongodb.org/browse/LG-5035)

### For new components

### DIFF
--- a/packages/table/src/useLeafyGreenTable/useLeafyGreenTable.spec.tsx
+++ b/packages/table/src/useLeafyGreenTable/useLeafyGreenTable.spec.tsx
@@ -93,7 +93,6 @@ describe('packages/table/useLeafyGreenTable', () => {
     expect(result.current.getRowCount()).toEqual(data.length + 1);
   });
 
-  // FIX THIS
   test('returns the correct number of rows with initialState set by consumer', () => {
     const data = getDefaultTestData({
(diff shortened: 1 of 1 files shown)


packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
```diff
@@ -30,7 +30,10 @@ function useLeafyGreenTable<T extends LGRowData, V extends unknown = unknown>({
   allowSelectAll = true,
   ...rest
 }: LeafyGreenTableOptions<T, V>): LeafyGreenTable<T> {
-  const [expanded, setExpanded] = useState<ExpandedState>({});
+  const [expanded, setExpanded] = useState<ExpandedState>(
+    (rest.initialState?.expanded as ExpandedState) ?? {},
+  );
```
stephl3, line 35: appreciate the story but wondering if we can add a test spec for this?

charcoalyy: ahh. that makes sense - let me get on that swap!

PR #2895, opened 2025-06-06
LG-5078: Add language support to `CodeEditor`

tsck:
## ✍️ Proposed changes

- Adds language support to the `CodeEditor` component

### For new components

## 🧪 How to test changes


packages/code-editor/src/CodeEditor/utils/codeSnippets.ts
```diff
@@ -0,0 +1,190 @@
...
+    "origin": {
+      "starSystem": "Zeta Reticuli",
+      "planet": "Apex IV",
+      "distanceLy": 39.17
+    },
+    "crewCount": 5
+  },
+  "hostile": false,
+  "initialMessage": "We have triangulated this location as your planet's cultural epicenter. We come in peace.",
+  "culturalRequests": [
```
stephl3, line 129: 😂 😂 😂

packages/code-editor/src/CodeEditor/utils/codeSnippets.ts
```diff
@@ -0,0 +1,190 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Greetings Earthlings</title>
+    <link rel="stylesheet" href="styles.css">
+</head>
+<body>
+    <main>
+        <article>
+            <h1>Greetings, Earthlings!</h1>
+            <p>We have traveled 39 light-years from the Zeta Reticuli system to deliver this message. We come in peace.</p>
+            <p>Take us to your leader... or at least to your best taco stand.</p>
```
stephl3, line 86: I'm dead :shipit:

packages/code-editor/src/CodeEditor/utils/createCodeMirrorLanguageExtension.ts
```diff
@@ -0,0 +1,81 @@
...
+} as const;
+
+export type LanguageName = keyof typeof codeMirrorLanguageExtensions;
+export const LanguageName = Object.keys(codeMirrorLanguageExtensions).reduce(
+  (acc, key) => {
+    acc[key as LanguageName] = key as LanguageName;
+    return acc;
+  },
+  {} as Record<LanguageName, LanguageName>,
+);
```
stephl3, line 77: We also have `SupportedLanguages` from `@leafygreen-ui/code` which is coupled with highlight.js. We now have `LanguageName` which is coupled with codemirror. It's not 100% overlap, but there is enough that I wonder if it might be worth maintaining them together. It may help consumers in understanding what is / is not "supported" for `Code` vs for `CodeEditor`. What do you think?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -40,19 +41,23 @@ export const CodeEditor = forwardRef<CodeMirrorRef, CodeEditorProps>(
...
+      extensions: consumerExtensions = [],
       ...rest
     },
     forwardedRef,
   ) => {
     const [value, setValue] = useState(defaultValue || '');
+    const [languageExtension, setLanguageExtension] =
+      useState<CodeMirrorExtension>([]);
     const editorRef = useRef<CodeMirrorRef>(null);
+    const refs = useMergeRefs([editorRef, forwardedRef]);
```
stephl3: nit: singular `ref` may still be appropriate here because it returns a single callback that loops through the array of refs

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -134,18 +139,34 @@ export const CodeEditor = forwardRef<CodeMirrorRef, CodeEditorProps>(
...
+      async function setupLanguageExtension() {
+        const languageCompartment = new Compartment();
+        setLanguageExtension(
+          languageCompartment.of(
+            language ? await createCodeMirrorLanguageExtension(language) : [],
+          ),
+        );
+      }
+      setupLanguageExtension();
+    }, [language]);
```
stephl3, line 156: does the "language support" described in the accompanying comment also refer to syntax highlighting? or where does that happen?

when I type in some code in the live example story and use the language dropdown, I'm expecting the syntax to match the highlighting of the language, but that doesn't seem to be happening. should it be and if so, is there another control I should be updating?

tsck: This is syntax highlighting as well! There was a bug in storybook where I misspelled "language" as "langauge" for the arg type 🙃. Will push a fix shortly

packages/code-editor/src/CodeEditor/utils/createCodeMirrorLanguageExtension.ts
tsck: I'm not sure. I'm wondering if that would too strongly couple them? They really are two distinct components with different language support.

stephl3: what about a separate package to have a list of all possible and both `@leafygreen-ui/code` and `@leafygreen-ui/code-editor` could refer to this list? I worry about usage/maintainability if identical languages are represented by different strings in either component unless `CodeEditor` will eventually outright replace `Code`

PR #2899, opened 2025-06-13
LG-5139: optional icon in chart titles

charcoalyy:
## ✍️ Proposed changes

Additional optional prop added to render content (e.g., tooltip) directly right of the chart header title. 

🎟 _Jira ticket:_ [LG-5139](https://jira.mongodb.org/browse/LG-5139)

### For new components


charts/core/src/ChartHeader/ChartHeader.tsx
```diff
@@ -25,6 +26,7 @@ export function ChartHeader({
         <Body weight="regular" baseFontSize={BaseFontSize.Body1}>
           {title}
         </Body>
+        <div>{titleIcon}</div>
```
tsck: Nit: Since this could be `undefined` we might want to only render the container `div` if `titleIcon` is defined. This just avoids rendering an empty `div` when it's `undefined`.
```suggestion
        {titleIcon && <div>{titleIcon}</div>}
```

PR #2906, opened 2025-06-18
LG-5314: base progress bar component

charcoalyy:
## ✍️ Proposed changes

Base component for progress bar (variants, sizes, formatting, and base case tests). This PR does NOT include animations, accessibility features, screenreader, or edge case testing.

🎟 _Jira ticket:_ [LG-5313](https://jira.mongodb.org/browse/LG-5313), [LG-5314](https://jira.mongodb.org/browse/LG-5314)

### For new components

### DIFF
--- a/packages/progress-bar/src/ProgressBar/index.ts
+++ b/packages/progress-bar/src/ProgressBar/index.ts
@@ -4,7 +4,7 @@ export {
   FormatValueType,
   LoaderVariant,
   MeterStatus,
-  ProgressBarProps,
+  type ProgressBarProps,
   Size,
   Type,
 } from './ProgressBar.types';


packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+      setValue(200);
+    }, 1500);
+
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [props.value]);
+
+  return <ProgressBar {...props} value={value} />;
+};
```
stephl3: Using this to capture the simulated progression of the `ProgressBar` instance is helpful for visually testing locally, but it may be flaky with snapshot testing depending on this timeout. We may want to consider using a playtest when simulating progression

Here's an example (pulled from gemini so it may need additional work to get it working in our stories but gives the essence of what we might want):
```
// YourProgressBar.stories.tsx
import type { Meta, StoryObj } from '@storybook/react';
import { userEvent, within } from '@storybook/testing-library';
import { ProgressBar } from './ProgressBar';

const meta: Meta<typeof ProgressBar> = {
  component: ProgressBar,
};

export default meta;
type Story = StoryObj<typeof meta>;

export const LoadingSequence: Story = {
  play: async ({ canvasElement }) => {
    const canvas = within(canvasElement);
    const startButton = await canvas.getByRole('button', { name: /Start/i });

    // Initial state
    await window.takeSnapshot('0-percent');

    await userEvent.click(startButton);

    // Intermediate states
    await new Promise(resolve => setTimeout(resolve, 500)); // Wait for animation
    await window.takeSnapshot('25-percent');

    await new Promise(resolve => setTimeout(resolve, 500));
    await window.takeSnapshot('50-percent');

    await new Promise(resolve => setTimeout(resolve, 500));
    await window.takeSnapshot('75-percent');

    await new Promise(resolve => setTimeout(resolve, 500));
    await window.takeSnapshot('100-percent');
  },
};
```

Searching for "play:" in `*.stories.*` files will show more examples of how we use play tests

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+  render: SimulatedProgressBar,
+};
+
+export const ShimmerDeterminate: StoryObj<typeof ProgressBar> = {
+  args: {
+    ...BasicDeterminate.args,
+    enableAnimation: true,
+  },
+  render: SimulatedProgressBar,
+};
```
stephl3: Can we hold on adding stories testing things that aren't implemented yet? In my experience, it's helpful to colocate with the actual changes, so the reviewer can give attention to the diff when reviewing the corresponding changes

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -0,0 +1,124 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
```
stephl3, line 2: TIL that `screen` is the recommended way to use RTL queries! https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

I was originally going to make a comment about consistently destructuring queries from `render` vs using `screen` queries, but it sounds like the latter is the best practice :)

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+import { ProgressBar } from '.';
+
```
stephl3, line 10: Something I like doing with specs in which props may use the same values across multiple test blocks is moving them to the highest appropriate context to reuse the same values. While not necessary for `TEST_LABEL` and `TEST_DESCRIPTION`, it might be helpful for `TEST_VALUE` and `TEST_MAX_VALUE` unless they absolutely need to be different in which case they can be defined at the test block level

Usually, I might make it an object with keys like:
```
const defaultProps = {
  label: 'Label',
  description: ...
  ...
};
```

I've also gotten feedback about using `@faker-js/faker` which can be used as a dev dep for mock data used in tests or stories

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+describe('packages/progress-bar', () => {
+  describe('basic rendering', () => {
+    it('renders with a label', () => {
```
stephl3: these should be using `test()` instead of `it()`

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+      const TEST_LABEL = 'placeholder label';
+      const { getByText } = render(
+        <ProgressBar isIndeterminate={true} label={TEST_LABEL} />,
+      );
+      expect(getByText(TEST_LABEL)).toBeInTheDocument();
```
stephl3: This is one of those cases where we may want to be more strict about the assertion and use `toBeVisible` to ensure this text is visible to the user

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <div
+          data-testid="progress-bar-track"
+          className={cx(getBarTrackStyles({ theme, size }))}
+        >
+          <div
+            data-testid="progress-bar-fill"
+            className={cx(
+              getBarFillStyles({ theme, variant, disabled }),
+              getTypedProgressBarFillStyles(),
+            )}
```
stephl3: 1. This looks like the only case you'd need to use the `cx` util. [See cx docs](https://emotion.sh/docs/@emotion/css#cx). In other cases where `cx` is only wrapping 1 classname, you should be fine to drop `cx`
2. nit: generally, I try to only import from `@leafygreen-ui/emotion` in the `*.styles.ts` file to keep code cleaner. This isn't perfectly followed in the codebase but could do similar by having ` getBarFillStyles` wrap `getBaseBarFillStyles` and `getTypedBarFilledStyles`

```
const getBarFillStyles = ({ ... }) => cx(
  getBaseBarFillStyles({ ... }),
  getTypedBarFillStyles({ ... }),
);
```

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,122 @@
...
+      >
+        <div
+          data-testid="progress-bar-track"
+          className={cx(getBarTrackStyles({ theme, size }))}
+        >
+          <div
+            data-testid="progress-bar-fill"
+            className={cx(
+              getBarFillStyles({ theme, variant, disabled }),
+              getTypedProgressBarFillStyles(),
```
stephl3: Can `getTypedProgressBarFillStyles` similarly accept an object as done with `getBarFillStyles` and use `isDeterminate`,  `value`, and `maxValue` as parameters to determine the correct styles?

packages/progress-bar/src/ProgressBar/ProgressBar.utils.tsx
```diff
@@ -0,0 +1,68 @@
...
+): string => {
+  if (typeof formatValue === 'function') {
+    return formatValue(value, maxValue);
+  }
+
+  switch (formatValue) {
+    case 'fraction':
+      return `${value}/${maxValue}`;
+    case 'percentage':
+      return `${getPercentage(value, maxValue)}%`;
```
stephl3, line 108: Is this not a `ReferenceError`? Not sure why it's not flagged

packages/progress-bar/src/ProgressBar/ProgressBar.utils.tsx
```diff
@@ -0,0 +1,68 @@
...
+import InfoWithCircleIcon from '@leafygreen-ui/icon/dist/InfoWithCircle';
+import WarningIcon from '@leafygreen-ui/icon/dist/Warning';
+
+import { ProgressBarValueType, ProgressBarVariant } from './ProgressBar.types';
+
+export const DEFAULT_MAX_VALUE = 1;
+
+export const iconsVisibleOnComplete = ['success'];
+
+export const getFormattedValue = (
```
stephl3, line 95: can we add test specs for this util?

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -0,0 +1,58 @@
...
+  | typeof Variant.Info
+  | typeof Variant.Success
+  | typeof Variant.Warning
+  | typeof Variant.Error;
+
+export type ProgressBarValueType =
+  | 'fraction'
+  | 'percentage'
+  | 'number'
+  | ((value: number, maxValue?: number) => string);
```
stephl3: We [prefer constants and types over enums](https://github.com/mongodb/leafygreen-ui/blob/main/STYLEGUIDE.md#prefer-constants-and-types-over-enums)

We then export `Size`, `Variant`, and `ValueType` from the package which can be used by consumers

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -0,0 +1,58 @@
...
+
+interface BaseProgressBarProps {
+  variant?: ProgressBarVariant;
+  label?: React.ReactNode;
+  size?: ProgressBarSize;
+  description?: React.ReactNode;
+
+  darkMode?: boolean;
```
stephl3: you can import `DarkModeProps` from `@leafygreen-ui/lib` instead of needing to define here

`interface BaseProgressBarProps extends DarkModeProps { ... }`

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -0,0 +1,58 @@
...
+  | 'fraction'
+  | 'percentage'
+  | 'number'
+  | ((value: number, maxValue?: number) => string);
+
+interface BaseProgressBarProps {
+  variant?: ProgressBarVariant;
+  label?: React.ReactNode;
+  size?: ProgressBarSize;
+  description?: React.ReactNode;
```
stephl3: Add tsdocs to describe props. These can be the same as what is used in the `README`

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  darkMode?: boolean;
+  disabled?: boolean;
```
stephl3: Add tsdoc

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Shimmer animation recommended for long-running tasks to provide additional visual feedback.
+   */
+  enableAnimation?: boolean;
+}
+interface IndeterminateProgressBarProps {
+  isIndeterminate: true;
+  value?: number;
+}
```
stephl3: For discriminated unions, you'll need to add the same tsdocs for both because only the first type definition may surface for consumers

stephl3: You can reference `Popover.types.ts` for a similar case

packages/progress-bar/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+```shell
+yarn add @leafygreen-ui/progress-bar
+```
+
+### NPM
+
+```shell
+npm install @leafygreen-ui/progress-bar
+```
```
stephl3, line 34: Adding an example + basic props table in this PR would make sense to colocate in the same PR with the newly introduced code

PR #2908, opened 2025-06-20
LG-5136: render named React components to get node text content

charcoalyy:
## ✍️ Proposed changes

Previously, passing functional components into `getNodeTextContent` returned an empty string. Changed to render, then continue recursion.

🎟 _Jira ticket:_ [LG-5136](https://jira.mongodb.org/browse/LG-5136)

### For new components

### DIFF
--- a/packages/lib/src/getNodeTextContent/index.ts
+++ b/packages/lib/src/getNodeTextContent/index.ts
@@ -1,4 +1,4 @@
-import { ReactElement, ReactNode, ReactText } from 'react';
+import { FunctionComponent, ReactElement, ReactNode, ReactText } from 'react';
 
 /**
  * Returns the text string of a React node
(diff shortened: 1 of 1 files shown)


packages/lib/src/getNodeTextContent/index.ts
```diff
@@ -12,14 +12,23 @@ export default function getNodeTextContent(node?: ReactNode): string {
     return node.map(getNodeTextContent).join(' ').trim();
   }
 
-  if (hasChildren(node)) {
+  if (isReactElement(node)) {
+    if (isFunctionalComponent(node)) {
+      const Component = node.type as FunctionComponent<any>;
+      return getNodeTextContent(Component(node.props));
+    }
+
```
TheSonOfThomp, line 20: can we add some test cases for this?

.changeset/four-bikes-relate.md
TheSonOfThomp, line 1: looks like there's 2 changesets. was that intentional?

charcoalyy: oh yikes! totally forgot i did that friday. thanks for catching :)

packages/lib/src/getNodeTextContent/index.ts
charcoalyy: good shout, will create the spec file :)

PR #2910, opened 2025-06-23
Exports Tooltip event handler utils

TheSonOfThomp:
## ✍️ Proposed changes

Exports utilities (`createTooltipTriggerEventHandlers` & `useTooltipTriggerEventHandlers`) to enable handling external tooltip triggers.

```tsx
const tooltipEventHandlers = useTooltipTriggerEventHandlers({
  triggerEvent: TriggerEvent.Hover,
  setState: setOpen,
  onFocus: (e) => { console.log(e) } // side effects called on focus of the trigger
});

return (
<>
  <Button ref={triggerRef} {...tooltipEventHandlers}>
    Button
  </Button>
  <Tooltip
    refEl={triggerRef}
    open={open}
    setOpen={setOpen}
  >Content</Tooltip>
</>
)
```


packages/tooltip/src/Tooltip/utils/createTooltipTriggerEventHandlers.ts
```diff
@@ -0,0 +1,242 @@
...
+  onClick: MouseEventHandler<HTMLElement>;
+}
+
+/**
+ * Creates the appropriate event handlers
+ * for a given Tooltip triggerEvent.
+ *
+ * Spread the returned object onto the tooltip trigger element
+ */
+export function createTooltipTriggerEventHandlers(
```
shaneeza: Is this going to replace `createTriggerProps`?

packages/tooltip/src/Tooltip/utils/createTooltipTriggerEventHandlers.ts
```diff
@@ -0,0 +1,242 @@
...
+  // of checking if the tooltip is enabled from within the callbacks
+  // without creating a new callback on each render
+  const isEnabledRef = useRef(isEnabled);
+  useEffect(() => {
+    isEnabledRef.current = isEnabled;
+  }, [isEnabled]);
+  const getIsEnabled = useCallback(() => isEnabledRef.current, []);
+
+  return {
+    onMouseEnter: useMemo(
```
shaneeza: Could we extract all of these triggers and use them in both `createTooltipTriggerEventHandlers` and `useTooltipTriggerEventHandlers`?

packages/tooltip/src/Tooltip/utils/createTooltipTriggerEventHandlers.ts
```diff
@@ -0,0 +1,242 @@
...
+              args.onClick?.(e);
+              setState(curr => !curr);
+            }
+          }
+        },
+      };
+  }
+}
+
+export function useTooltipTriggerEventHandlers(
```
shaneeza: Could you add TSdocs and a section in the README on how to use this?

packages/tooltip/src/Tooltip/utils/createTooltipTriggerEventHandlers.ts
TheSonOfThomp: Yeah it's meant to be a reusable version of that. I'll update the internal TT implementation to use this fn

.changeset/tooltip-trigger-handlers.md
```diff
@@ -0,0 +1,27 @@
+---
+'@leafygreen-ui/tooltip': minor
```
shaneeza, line 2: Can these changesets be combined?

.changeset/tooltip-trigger-handlers.md
```diff
@@ -0,0 +1,27 @@
...
+---
+
+Exports utilities (`createTooltipTriggerEventHandlers` & `useTooltipTriggerEventHandlers`) to enable handling external tooltip triggers.
```
shaneeza: rm `createTooltipTriggerEventHandlers`?

packages/tooltip/src/Tooltip/utils/tooltipHandlers.types.ts
```diff
@@ -0,0 +1,95 @@
+import { FocusEventHandler, MouseEventHandler, RefObject } from 'react';
+
+import type { TriggerEvent } from '../Tooltip.types';
+
+export interface CreateTooltipEventsBaseArgs {
```
shaneeza: Should this be updated to `UseTooltipEventsBaseArgs` or just `TooltipEventsBaseArgs`?

packages/tooltip/src/Tooltip/utils/tooltipHandlers.types.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Whether the tooltip event handlers should be enabled
+   */
+  isEnabled?: boolean;
+
+  /**
+   * The delay in milliseconds before the tooltip opens on hover.
+   */
+  delay?: number;
```
shaneeza: If consumers use `useTooltipTriggerEventHandlers`, are we giving them control of the delay?

packages/tooltip/src/Tooltip/utils/useTooltipTriggerEventHandlers.spec.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+    // Update props with isEnabled=false
+    setState.mockClear();
+    isEnabled = false;
+    const updatedArgs = { ...hoverArgs, isEnabled };
+    rerender(updatedArgs);
+
+    // After rerender with isEnabled=false, handlers should be disabled
+    result.current.onMouseEnter(mockMouseEvent);
+    jest.advanceTimersByTime(CALLBACK_DEBOUNCE);
```
shaneeza: Why are you not including `DEFAULT_HOVER_DELAY` here?

packages/tooltip/src/Tooltip/utils/useTooltipTriggerEventHandlers.spec.tsx
```diff
@@ -0,0 +1,138 @@
...
+      {
+        initialProps: hoverArgs,
+      },
+    );
+    const initialHandlers = result.current;
+
+    // Rerender with the same props
+    rerender({ ...hoverArgs, delay: 0 });
+
+    // Handlers should be the same objects (memoized)
```
shaneeza:
```suggestion
    // Handlers should not be the same objects 
```

.changeset/tooltip-trigger-handlers.md
```diff
@@ -0,0 +1,27 @@
...
+    Button
+  </Button>
+  <Tooltip
+    refEl={triggerRef}
+    open={open}
+    setOpen={setOpen}
+  >Content</Tooltip>
+</>
+)
+```
```
shaneeza, line 27: Do we want to include this in the README?

packages/tooltip/src/Tooltip/utils/tooltipHandlers.types.ts
TheSonOfThomp: good callout 👍

packages/tooltip/src/Tooltip/utils/tooltipHandlers.types.ts
TheSonOfThomp: Yeah, I feel like this would be a nice escape hatch in cases we need it. Open to removing it though

shaneeza: feels like a q for design since, I believe, we picked the delay for accessibility reasons.

packages/tooltip/src/Tooltip/utils/tooltipHandlers.types.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export interface CreateTooltipEventsArgsHover {
```
shaneeza: We should also update these other types that start with `Create`

PR #2911, opened 2025-06-23
Moves Utils & Hooks from /private

TheSonOfThomp:
## ✍️ Proposed changes

`@leafygreen-ui/hooks`: minor
---

Exports `useFirstRender`: Returns whether this is the first render of a component.

`@leafygreen-ui/lib`: minor
---

Exports `deepMapValues`: Recursively maps the values of an object using a callback function


.changeset/lib-from-private.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/lib': minor
+---
+
+Exports `deepMapValues`: Recursively maps the values of an object using a callback function
```
stephl3: looks like we're also adding `getIndefiniteArticle` and `queryFocusableElements`?

packages/lib/src/getIndefiniteArticle/getIndefiniteArticle.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Returns the appropriate indefinite article ('a' or 'an') for a given word
+ * based on whether it starts with a vowel sound.
+ *
+ * @param word - The word to get the indefinite article for
+ * @returns 'a' or 'an'
+ */
+export function getIndefiniteArticle(word: string): string {
+  return /^[aeiou]/i.test(word) ? 'an' : 'a';
```
stephl3: I'm not sure what this is specifically for, but we may want to rely on a library like `indefinite` for this because of words that don't follow this pattern. For example, "an hour" or "a user"

https://www.npmjs.com/package/indefinite

packages/lib/src/queryFocusableElements/queryFocusableElements.ts
```diff
@@ -0,0 +1,28 @@
+import last from 'lodash/last';
+
+export const focusableElementSelector =
```
stephl3: Can we update the tsdocs for each of these to call out what is specifically being looked for in the `focusableElementSelector`? This might be helpful for debugging if we see unexpected behavior

packages/lib/src/getIndefiniteArticle/getIndefiniteArticle.ts
TheSonOfThomp: I like that. This is really only used in cloud-nav for now. I like using that package instead of rolling our own

PR #2916, opened 2025-06-24
LG-5101/5102: Updates editor and syntax highlight styling

tsck:
## ✍️ Proposed changes
- Updates editor and syntax highlight styling to match currently known designs. These aren't finalized and there's no eng spec yet, but this is meant to get things closer. Changes are based off of [this Figma presentation](https://www.figma.com/design/QsOruUtjXisdkdZHJh9CsG/Code-Editor?node-id=1-8). (Note: this only styles what's already been implemented.)
- Fixes the ability for consumers to style things themselves by passing their own styles/classNames.
- Removes `enableActiveLineHighlighting` and the ability to do so. This wasn't in the designs, I spoke with Sandy and we decided to just leave this out for now.

🎟 _Jira ticket:_ 
- [LG-5101](https://jira.mongodb.org/browse/LG-5101)
- [LG-5102](https://jira.mongodb.org/browse/LG-5102)

### For new components

## 🧪 How to test changes
- Visual QA in Storybook

### Example Before
Screenshot 2025-06-24 at 4 29 51 PM

### Example After
Screenshot 2025-06-24 at 4 29 27 PM


packages/code-editor/src/CodeEditor.stories.tsx
```diff
@@ -131,6 +128,9 @@ const meta: StoryMetaType<typeof CodeEditor> = {
       control: { type: 'select' },
       options: Object.values(LanguageName),
     },
+    darkMode: {
+      control: { type: 'boolean' },
+    },
```
stephl3: nit: can use `storybookArgTypes` from `'@lg-tools/storybook-utils` for this

```suggestion
    darkMode: storybookArgTypes.darkMode,,
```

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
```diff
@@ -65,35 +65,15 @@ describe('packages/code-editor', () => {
...
-    });
-    expect(
-      editor.queryBySelector(CodeEditorSelectors.ActiveLine),
-    ).not.toBeInTheDocument();
-    expect(
-      editor.queryBySelector(CodeEditorSelectors.ActiveLineGutter, {
-        text: '1',
-      }),
-    ).not.toBeInTheDocument();
+    ).toHaveStyle('visibility: hidden');
```
stephl3: should we instead check that `CodeEditorSelectors.LineNumbers` is not in the document?

packages/code-editor/src/CodeEditor/CodeEditor.styles.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * The CM Editor by default grows as lines are added. There are ways using
+ * specific classes to set things like heights on the editor, but this is
+ * inconvenient for consumers because they couldn't just use emotion's css prop
+ * or className prop to set a height. By wrapping the editor in 2 divs, the
+ * inner set to 100% height of the outer with the appropriate overflow, consumers
+ * can set the height of the outer wrapper and have the editor behave as expected.
+ */
+export const editorInnerWrapperStyles = css`
```
stephl3: do we want to add overflow shadows when scrolling similar to `Code`? we can use `addOverflowShadows` from `@leafygreen-ui/tokens for this

packages/code-editor/src/CodeEditor/codeMirrorExtensions/createHighlightExtension.ts
```diff
@@ -0,0 +1,90 @@
...
+      tag: [
+        tags.propertyName,
+        tags.link,
+        tags.typeName,
+        tags.list,
+        tags.atom,
+        tags.self,
+        tags.deleted,
+      ],
+      color: theme === Theme.Light ? '#D83713' : '#FF6F44',
```
stephl3, line 68: It looks like these are the same colors used for highlighting the `Code` component. Does it make sense to tokenize these for usage in both?

packages/code-editor/src/CodeEditor/codeMirrorExtensions/createThemeExtension.ts
```diff
@@ -0,0 +1,69 @@
...
+        fontSize: `${fontSize}px`,
+      },
+
+      [`${CodeEditorSelectors.LineNumbers} ${CodeEditorSelectors.GutterElement}`]:
+        {
+          width: '48px',
+        },
+
+      [CodeEditorSelectors.Line]: {
+        paddingLeft: '12px',
```
stephl3:
```suggestion
        paddingLeft: `${spacing[300]}px`,
```

packages/code-editor/src/CodeEditor/codeMirrorExtensions/createThemeExtension.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+
+      [CodeEditorSelectors.SelectionBackground]: {
+        background: 'rgba(1, 107, 248, 0.25) !important',
```
stephl3, line 76: is it worth adding a comment about why we need to use "!important" here?

packages/code-editor/tsconfig.json
```diff
@@ -17,6 +17,15 @@
     },
     {
       "path": "../lib"
+    },
+    {
+      "path": "../tokens"
+    },
```
stephl3, line 26:
```suggestion
    },
    {
      "path": "../palette"
    },
    {
      "path": "../tokens"
    },
```

tools/install/src/ALL_PACKAGES.ts
```diff
@@ -94,6 +94,7 @@ export const ALL_PACKAGES = [
   '@lg-tools/build',
   '@lg-tools/cli',
   '@lg-tools/codemods',
+  '@lg-tools/crawler',
```
stephl3: rm?

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
tsck: Yes, great call!

packages/code-editor/src/CodeEditor/CodeEditor.styles.ts
tsck: Discussed offline and we'll look into adding this in the later style iteration since it's not yet on the designs and could be a bit of a lift.

packages/code-editor/src/CodeEditor/codeMirrorExtensions/createHighlightExtension.ts
tsck: The only reason I didn't is because I don't know if there's a good communal place to put them. So I just followed the [rule of threes](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) here and accepted the duplication.

tools/install/src/ALL_PACKAGES.ts
tsck: Whoops! This is for some reason stuck in my env and keeps getting randomly added lol. I need to look into it I just haven't found the time. Thanks for catching!

tsck:
@stephl3 This should be ready for another look. Tried to keep the commit history fairly clean, but here's some additional context for the next pass in case it helps.

As discussed offline, I discovered that the container scrolling solution was not working as expected, this was specifically addressed in [5bbd919](https://github.com/mongodb/leafygreen-ui/pull/2916/commits/5bbd9199e7d971749238f7498dd6813988dac9bd) and [78426d2](https://github.com/mongodb/leafygreen-ui/pull/2916/commits/78426d2f6cb6ffde967585313614d33ef079ece6).

Sorry in advance for adding an additional refactor! I saw an unrelated issue with the location of test utils. I was exporting this package from the root, which meant that `testing-library` needed to technically be added as a dependency, which would bloat our package. Previous work was done to setup a `/testing` export if added to a 'testing' directory (I'm told?) so I've just moved it to that directory in this PR, addressed in [e3af3e9](https://github.com/mongodb/leafygreen-ui/pull/2916/commits/e3af3e9c5bcc277f7d316dd2aea3b806e0857b88) and [6412277](https://github.com/mongodb/leafygreen-ui/pull/2916/commits/6412277fd5e33d254fc895725eb6f5ff30472b8e).

Other feed addressed in [77d8185](https://github.com/mongodb/leafygreen-ui/pull/2916/commits/77d81855241b19c58e1e953d7670ddd16e56fd08), [27a8da4](https://github.com/mongodb/leafygreen-ui/pull/2916/commits/27a8da4ae0e0b70ac5cc029cee0220525a832a3b), and [d0f8817](https://github.com/mongodb/leafygreen-ui/pull/2916/commits/d0f881704ffbf7bee7be52c8272fd6d81ef1b8e7)

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
stephl3: bumping because it looks like you intended to update? but no worries if this works too

packages/code-editor/src/CodeEditor/codeMirrorExtensions/createThemeExtension.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      [CodeEditorSelectors.SelectionBackground]: {
+        /**
+         * The background was not getting applied correctly without !important.
+         * I'm not entirely sure why, but there must be a sheet being applied
+         * higher or with greater specificity. It was hard to track down because
+         * for some reason our background color was getting applied correctly
+         * as soon as I'd select text and pull it up in devtools. ¯\_(ツ)_/¯
+         */
+        background: 'rgba(1, 107, 248, 0.25)',
```
stephl3: 😂 I think you added the comment but dropped the `!important`. can we remove the comment or add back the `!important`?

packages/code-editor/package.json
```diff
@@ -37,23 +37,29 @@
...
     "@uiw/codemirror-extensions-hyper-link": "^4.23.12",
     "@uiw/react-codemirror": "^4.23.10"
   },
+  "devDependencies": {
+    "@lg-tools/build": "workspace:^",
+    "@storybook/test": "^8.6.14"
+  },
+  "peerDependencies": {
+    "@leafygreen-ui/leafygreen-provider": "workspace:^"
+  },
```
stephl3, line 61: you need to update the `"exports"` key in this file to export from the testing subdir

```
"exports": {
    ".": {
      "require": "./dist/umd/index.js",
      "import": "./dist/esm/index.js",
      "types": "./dist/types/index.d.ts"
    },
    "./testing": {
      "require": "./dist/umd/testing/index.js",
      "import": "./dist/esm/testing/index.js",
      "types": "./dist/types/testing/index.d.ts"
    }
  },
```

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -158,21 +207,63 @@ export const CodeEditor = forwardRef<CodeMirrorRef, CodeEditorProps>(
...
+              ${CodeEditorSelectors.Editor} {
+                min-width: ${minWidth};
+              }
+            `]: !!minWidth,
+            [css`
+              ${CodeEditorSelectors.Editor} {
+                width: ${width};
+              }
+            `]: !!width,
+          },
```
stephl3: it looks like this is following an alphabetical pattern although I'm wondering if we might want to order them a certain way based on logical flow and what could potentially override what

in terms of logical flow, I like this idea from gemini: "Set the component's `width`, but constrain it by `minWidth` and `maxWidth`."

for overrides, something to consider is: what happens when `width` is smaller than `minWidth` or `width` is larger than `maxWidth`?

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
tsck: Lol I swear I changed this but must have gotten pulled away and stashed it before committing. It's been a busy week lol. Will push

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Unless things have changed, I don't think order matters in regards to overrides. https://stackoverflow.com/a/16063871. If this is incorrect or you'd rather it, I'll just pass through and order them width -> max-width -> min-width -> etc.

> for overrides, something to consider is: what happens when width is smaller than minWidth or width is larger than maxWidth?
This should work exactly as CSS works. I wouldn't want to add logic to change that.

packages/code-editor/src/CodeEditor/codeMirrorExtensions/createThemeExtension.ts
tsck: Lmao oops. I was testing it. Good catch. Will fix

packages/code-editor/src/CodeEditor.stories.tsx
tsck: I updated this too! I've been bouncing around branches this must have gotten lost as well. Will also update this

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Precedence is also described well on MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/width

PR #2920, opened 2025-06-26
LG-5314 (refactor): progress bar meter/loader distinction

charcoalyy:
## ✍️ Proposed changes

Refactors current progress bar implementation to distinguish meter- versus loader-type usage. [Full context available here](https://docs.google.com/document/d/1bE0W5MtCmTM7ElNKPlVu0ioeZktog3lBLdhnbJIBQqk/edit?tab=t.8zbcdnqgk17b).

What did this fix?

- Clearer semantics for meter-type usage ("success" now only has one meaning instead of two)
- Assigning "role" attribute now matches semantic HTML
- Unsupported prop combinations are no longer possible for consumers to input (e.g., cannot have warning/error indeterminate bars)

### For new components


packages/progress-bar/README.md
```diff
@@ -52,19 +53,26 @@ const processUpload = () => {...}
...
-| `maxValue?`        | `number`                                                                                               | `1`        | determinate only | Optional total progress value. Unavailable if `isIndeterminate` is `true`.                                                       |
-| `enableAnimation?` | `boolean`                                                                                              | `false`    | determinate only | If true, enables shimmer animation to indicate progression for longer-running tasks. Unavailable if `isIndeterminate` is `true`. |
-| `disabled?`        | `boolean`                                                                                              | `false`    | determinate only | If true, pauses the progress bar and renders it in a gray color. Unavailable if `isIndeterminate` is `true`.                     |
+The `ProgressBar` component supports two types:
+
+- **Meter**, which displays a static snapshot of progress at some point in time.
```
stephl3: Since there is a note with the `Loader` type that it can be determinate/indeterminate, can we add a note with `Meter` type that it should only ever determinate?

packages/progress-bar/README.md
```diff
@@ -52,19 +53,26 @@ const processUpload = () => {...}
...
+- **Loader**, which supports determinate and indeterminate modes. Determinate loaders dynamically approach a maximum; indeterminate loaders loop infinitely.
+
+Use a meter when you want to show precise progress alongside a health status (e.g., amount of disk space used). Use loaders when you want to show progress towards completion or ongoing activity (e.g., file upload progress).
```
stephl3: This looks like a good description that would be appropriate to include higher up above the example

```
## Description

The `ProgressBar` component supports ...

## Example

...

## Properties

...
```

packages/progress-bar/README.md
```diff
@@ -36,6 +36,7 @@ const total = 100;
 const processUpload = () => {...}
 
 <ProgressBar
+  type="loader"
```
stephl3, line 48: I think it's worth adding examples of the different final options that users will most likely be using

### Determinate loader

### Indeterminate loader

### Meter

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -2,12 +2,24 @@ import React from 'react';
 
 import { DarkModeProps } from '@leafygreen-ui/lib';
 
+export const Type = {
+  Meter: 'meter',
+  Loader: 'loader',
+} as const;
+export type Type = (typeof Type)[keyof typeof Type];
```
stephl3, line 9: I generally try to stray away from naming a prop anything that could be conflated with a native html attribute. In this case, the `type` attribute does not apply for `<div>` element, so it's probably fine.

Alternatively, this could be prefixed as `*Something*Type` or replaced with `Kind`

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -2,12 +2,24 @@ import React from 'react';
...
+
+export const MeterStatus = {
+  Healthy: 'healthy',
+  Warning: 'warning',
+  Error: 'error',
+} as const;
+export type MeterStatus = (typeof MeterStatus)[keyof typeof MeterStatus];
```
stephl3: I'm wondering if `Variant` should instead be a master list of all the possible variants between meter or loader types. Thinking this because:
1. there is duplication with `'warning'` and `'error'`
2. `AnimatedDeterminateLoaderProps.variant` and `IndeterminateLoaderProps.variant` are already a subset of `Variant`s

Wondering if the `MeterProps.variant` could also be a subset and the styling for `Variant.Success` and `Variant.Healthy` are the same

stephl3: Could also break them up as `AnimatedDeterminateLoaderVariant`, `IndeterminateLoaderVariant`, and `MeterVariant` if you wanted to make these distinctions clearer

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -13,44 +12,37 @@ import {
...
-
-  const value = rest.value ?? undefined;
+  const {
+    type,
+    label,
+    size = Size.Default,
+    description,
+    darkMode = false,
+    formatValue,
+    showIcon: showIconProps = false,
```
stephl3: nit
```suggestion
    showIcon: showIconProp = false,
```

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
charcoalyy: hmm. i originally wanted to keep the concept of "variants" for loaders and the concept of "statuses" for meters so that we don't double assign meanings. it is true that the underlying colors (styling) have no difference- this was more so for semantic clarity/user mental model. i was also contemplating renaming meter statuses to closer align with "system health" terminology (e.g., critical or danger instead of error).

stephl3: Gotcha, I think that makes more sense to me then. I still think it's a bit weird from a code maintainer perspective that the meter statuses get mapped to a loader variant though. If they are intended to have different meanings externally, I would think similar should apply internally. What do you think about a separate enum for the coloring of the progress bar and `Variant` and `Status` both map to a `Color`?

charcoalyy: i like that a lot! i briefly considered it and didn't pursue it, but thinking on it more, that makes way more sense maintenance-wise. i'll make that change :)

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -154,11 +187,18 @@ export const DeterminateVariants: StoryObj<typeof ProgressBar> = {
       },
       excludeCombinations: [
         {
-          variant: ['warning', 'error'],
+          variant: Object.values(LoaderVariant).filter(
+            v =>
+              !(
+                Object.values(AnimatedLoaderVariant) as Array<LoaderVariant>
+              ).includes(v),
+          ),
```
charcoalyy, line 198: is this overkill? i could list them literally, but this is most declarative.

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
charcoalyy: this is super fair! i'm wondering in this case if `type` is still the best choice because of its clear meaning and shortness, and as it's not being passed to any native HTML elements?

packages/progress-bar/src/ProgressBar.stories.tsx
stephl3: This looks fine to me. It looks like you use the enums in array form a few times. Those arrays could be defined at the top of the file to make these more readable

packages/progress-bar/src/ProgressBar/ProgressBar.utils.tsx
```diff
@@ -5,11 +5,88 @@ import ImportantWithCircleIcon from '@leafygreen-ui/icon/dist/ImportantWithCircl
...
+      return Color.Green;
+    case LoaderVariant.Warning:
+      return Color.Yellow;
+    case LoaderVariant.Error:
+      return Color.Red;
+    case LoaderVariant.Info:
+    default:
+      return Color.Blue;
+  }
+};
```
stephl3, line 47: nice I like this approach better!

packages/progress-bar/src/ProgressBar/index.ts
```diff
@@ -1,2 +1,2 @@
 export { ProgressBar } from './ProgressBar';
-export { type ProgressBarProps } from './ProgressBar.types';
+export * from './ProgressBar.types';
```
stephl3: Do we want to be more explicit about what is and isn't exported? Will consumers ever need access to `Color` and `ResolvedProgressBarProps`?

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
stephl3: `type` works for me!

PR #2922, opened 2025-06-26
[LG-5293] feat: ContextDrawer part 2

stephl3:
## ✍️ Proposed changes

- fix glyph rotation in `ContextDrawerButton`
- add `ContextDrawer` component

Figma: https://www.figma.com/design/4h2VwjCuJUbeZ7hzD2J1rq/LeafyGreen-Design-System?m=auto&node-id=41064-134808
🎟 _Jira ticket:_ [LG-5293](https://jira.mongodb.org/browse/LG-5293)

### For new components

## 🧪 How to test changes

- unit tests
- [ ] review live example
- [ ] review `Controlled` story
- [ ] review `Generated` stories

### DIFF
--- a/packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
+++ b/packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
@@ -6,6 +6,7 @@ import LeafyGreenProvider, {
   useDarkMode,
 } from '@leafygreen-ui/leafygreen-provider';
 import { queryFirstFocusableElement } from '@leafygreen-ui/lib';
+import { formatCssSize } from '@leafygreen-ui/lib';
 
 import { TRANSITION_DURATION } from '../constants';
 
@@ -116,10 +117,7 @@ export const ContextDrawer = forwardRef<HTMLDivElement, ContextDrawerProps>(
               className={getContentWrapperStyles({
                 hasBottomShadow: !isBottomInView,
                 hasTopShadow: !isTopInView,
-                height:
-                  typeof expandedHeight === 'number'
-                    ? `${expandedHeight}px`
-                    : expandedHeight,
+                height: formatCssSize(expandedHeight),
                 isOpen: !!isOpen,
                 theme,
               })}
--- a/.changeset/lemon-adults-accept.md
+++ b/.changeset/lemon-adults-accept.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/lib': minor
+---
+
+Adds `formatCssSize` util for properly formatting a number or string into a valid css value
--- a/packages/lib/src/formatCssSize/formatCssSize.spec.ts
+++ b/packages/lib/src/formatCssSize/formatCssSize.spec.ts
@@ -1,0 +1,42 @@
+import { formatCssSize } from './formatCssSize';
+
+describe('formatCssSize', () => {
+  test('appends px to number input', () => {
+    expect(formatCssSize(100)).toBe('100px');
+    expect(formatCssSize(0)).toBe('0px');
+    expect(formatCssSize(-5)).toBe('-5px');
+  });
+
+  test('appends px to numeric string input', () => {
+    expect(formatCssSize('100')).toBe('100px');
+    expect(formatCssSize('0')).toBe('0px');
+    expect(formatCssSize('-5')).toBe('-5px');
+    expect(formatCssSize(' 42 ')).toBe('42px');
+  });
+
+  test('returns string with units as is', () => {
+    expect(formatCssSize('100px')).toBe('100px');
+    expect(formatCssSize('5rem')).toBe('5rem');
+    expect(formatCssSize('2em')).toBe('2em');
+    expect(formatCssSize('50%')).toBe('50%');
+    expect(formatCssSize('auto')).toBe('auto');
+    expect(formatCssSize('inherit')).toBe('inherit');
+    expect(formatCssSize('calc(100% - 10px)')).toBe('calc(100% - 10px)');
+  });
+
+  test('handles edge cases for string input', () => {
+    expect(formatCssSize('')).toBe('');
+    expect(formatCssSize(' ')).toBe('');
+    expect(formatCssSize('px')).toBe('px');
+    expect(formatCssSize('100 px')).toBe('100 px');
+    expect(formatCssSize('100.5')).toBe('100.5px');
+    expect(formatCssSize('.5')).toBe('.5px');
+    expect(formatCssSize('-.5')).toBe('-.5px');
+  });
+
+  test('does not append px to non-numeric stripnngs', () => {
+    expect(formatCssSize('foo')).toBe('foo');
+    expect(formatCssSize('barpx')).toBe('barpx');
+    expect(formatCssSize('100abc')).toBe('100abc');
+  });
+});
--- a/packages/lib/src/formatCssSize/formatCssSize.ts
+++ b/packages/lib/src/formatCssSize/formatCssSize.ts
@@ -1,0 +1,42 @@
+/**
+ * Formats a number or string value into a valid CSS size string.
+ * If the value is a number or a numeric string, it appends 'px'.
+ * If the value is a string with units, it returns it as is.
+ *
+ * This utility improves developer experience by making components more
+ * forgiving with size-related props.
+ *
+ * @param value - The value to format, e.g., 100, "100", or "100px".
+ * @returns A valid CSS size string (e.g., "100px") or undefined if the input is invalid.
+ *
+ * @example
+ * formatCssSize(100);       // "100px"
+ * formatCssSize("100");    // "100px"
+ * formatCssSize("100px");   // "100px"
+ * formatCssSize("5rem");    // "5rem"
+ * formatCssSize("auto");    // "auto"
+ * formatCssSize(undefined); // undefined
+ */
+export const formatCssSize = (value: number | string): string => {
+  // If the value is a number, simply append 'px'.
+  if (typeof value === 'number') {
+    return `${value}px`;
+  }
+
+  const trimmedValue = value.trim();
+
+  // If the value is a string, check if it's purely numeric.
+  // `isFinite` is used to correctly handle string-to-number conversion
+  // and reject strings that already have units (e.g., "100px").
+  if (
+    typeof trimmedValue === 'string' &&
+    !isNaN(parseFloat(trimmedValue)) &&
+    isFinite(Number(trimmedValue))
+  ) {
+    return `${trimmedValue}px`;
+  }
+
+  // If it's a string that's not purely numeric (e.g., "5rem", "auto"),
+  // return it directly, assuming it's a valid CSS value.
+  return trimmedValue;
+};
--- a/packages/lib/src/formatCssSize/index.ts
+++ b/packages/lib/src/formatCssSize/index.ts
@@ -1,0 +1,1 @@
+export { formatCssSize } from './formatCssSize';
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -6,6 +6,7 @@ import { type LgIdProps, type LgIdString } from './LgIdProps';
 import * as typeIs from './typeIs';
 export { createSyntheticEvent } from './createSyntheticEvent';
 export { deepMapValues } from './deepMapValues';
+export { formatCssSize } from './formatCssSize';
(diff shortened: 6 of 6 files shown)


packages/context-drawer/src/ContextDrawer/ContextDrawer.spec.tsx
```diff
@@ -0,0 +1,179 @@
...
+    />
+  );
+};
+
+const renderContextDrawer = (props: Partial<ContextDrawerProps> = {}) => {
+  return render(<MockContextDrawer {...props} />);
+};
+
+describe('packages/context-drawer', () => {
+  describe('a11y', () => {
```
TheSonOfThomp, line 24: should we test focus order?

packages/context-drawer/src/ContextDrawer/ContextDrawer.styles.ts
```diff
@@ -0,0 +1,112 @@
...
+  opacity: 0;
+  visibility: hidden;
+
+  :focus-visible {
+    outline: none;
+  }
+`;
+
+const getExpandedContentStyles = (height: number | string) => {
+  const expandedHeight = typeof height === 'number' ? `${height}px` : height;
```
TheSonOfThomp: on one hand I like this api, on the other it'll break silently if someone passes `"100"` as `expandedHeight`. Not sure how to remedy that, and probably a fine caveat. Just wanted to call it out in case you had any ideas for a fix

packages/context-drawer/src/ContextDrawer/ContextDrawer.styles.ts
```diff
@@ -0,0 +1,112 @@
...
+}: {
+  hasBottomShadow: boolean;
+  hasTopShadow: boolean;
+  height: number | string;
+  isOpen: boolean;
+  theme: Theme;
+}) =>
+  cx(contentWrapperStyles, {
+    [getExpandedContentStyles(height)]: isOpen,
+    [addOverflowShadow({ side: Side.Top, theme, isInside: true })]:
```
TheSonOfThomp, line 88: 🤘

packages/context-drawer/src/ContextDrawer/ContextDrawer.styles.ts
```diff
@@ -0,0 +1,112 @@
...
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  overflow-y: auto;
+  overscroll-behavior: contain;
+`;
+
+export const bottomInterceptStyles = css`
+  // Ensures bottom intercept is visible when end of scroll container is reached.
+  padding-bottom: ${spacing[25]}px;
```
TheSonOfThomp: nit: honestly, this is a case where I'd probably prefer a hard-coded `1px`, since it doesn't really have anything to do with spacing.

packages/context-drawer/src/ContextDrawer/ContextDrawer.types.ts
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * The element that the drawer is positioned relative to.
+   */
+  reference: ReactElement;
+
+  /**
+   * The element that triggers the opening and closing of the drawer.
+   * Can be a React element or a function that receives the `isOpen` state.
+   */
+  trigger: ReactElement | ((props: { isOpen: boolean }) => ReactElement);
```
TheSonOfThomp, line 46: can a trigger function take in additional props other than `isOpen`? Might need to extend that props object here if yes

packages/context-drawer/src/ContextDrawer/ContextDrawer.spec.tsx
stephl3: have a spec further down testing `'content receives focus when opened'`. were you thinking something in addition to this?

packages/context-drawer/src/ContextDrawer/ContextDrawer.styles.ts
stephl3: I can add a `formatCssSize` util to lib to be more forgiving about what gets passed. I'll do that in a follow-up PR and link here

packages/context-drawer/src/ContextDrawer/ContextDrawer.types.ts
stephl3: `isOpen` is the only one exposed in the function because it will determine which default button text to use when the `ContextDrawer` is uncontrolled. However, if a user _does_ need to add additional props, they can do this:
```
({ isOpen }) => <ContextDrawerButton onClick={...} disabled={...} />
```

packages/context-drawer/src/ContextDrawer/ContextDrawer.styles.ts
stephl3: https://github.com/mongodb/leafygreen-ui/pull/2930

packages/context-drawer/src/ContextDrawer/ContextDrawer.styles.ts
```diff
@@ -0,0 +1,113 @@
...
+ * that the reference element and content element appear to be flush with the border.
+ */
+export const referenceWrapperStyles = cx(
+  css`
+    margin-top: -${BORDER_WIDTH}px;
+    margin-left: -${BORDER_WIDTH}px;
+    margin-right: -${BORDER_WIDTH}px;
+  `,
+  referenceWrapperClassName,
+);
```
stephl3: I wasn't able to find a better way to approach collapsing the border used for the "inner container" element vs the element passed in the `reference` prop. This makes an assumption that the `reference` always has a 1px border which it does in the case of `CodeEditor`

For cases where the `reference` does not have a 1px border, I added/exported a `referenceWrapperClassName` to undo these built-in margins

packages/context-drawer/src/ContextDrawer/ContextDrawer.spec.tsx
stephl3: Added an additional spec for tab navigation. JSDOM doesn't fully support skipping focus on `visibility: hidden;` elements, so for now, I added a test.skip for what the real test should be along with a stopgap test + comment

TheSonOfThomp: I think we should also test how focus flows 
1. if the drawer is not opened (focus bypasses content?)
2. after the focus moves to content, where does it flow after?

stephl3: 1 has to be covered by the skipped test spec

for 2, I updated a spec to show that it flows back to the trigger

packages/context-drawer/src/ContextDrawer/ContextDrawer.styles.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Negative margins are used to offset the border width of the inner container so
+ * that the reference element and content element appear to be flush with the border.
+ */
+export const referenceWrapperStyles = cx(
+  css`
+    margin: -${BORDER_WIDTH}px;
+  `,
+  referenceWrapperClassName,
+);
```
TheSonOfThomp, line 50: 👍 this is prob the cleanest way to do it

packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
```diff
@@ -0,0 +1,152 @@
...
+            <div className={referenceWrapperStyles}>{reference}</div>
+            <div
+              aria-labelledby={triggerId}
+              className={getContentWrapperStyles({
+                hasBottomShadow: !isBottomInView,
+                hasTopShadow: !isTopInView,
+                height:
+                  typeof expandedHeight === 'number'
+                    ? `${expandedHeight}px`
+                    : expandedHeight,
```
TheSonOfThomp: should this use the new utility?

packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
```diff
@@ -0,0 +1,152 @@
...
+      updateValue(!isOpen, contentRef);
+
+      if (typeof trigger !== 'function' && trigger.props.onClick) {
+        trigger.props.onClick(e);
+      }
+    };
+
+    const renderTrigger = () => {
+      const triggerEl =
+        typeof trigger === 'function' ? trigger({ isOpen: !!isOpen }) : trigger;
```
TheSonOfThomp, line 96: `isOpen` is already boolean, isn't it?

packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
```diff
@@ -0,0 +1,152 @@
...
+    });
+    const { ref: bottomInterceptRef, inView: isBottomInView } = useInView({
+      root: scrollContainerRef.current,
+      threshold: 0,
+    });
+
+    const contentId = useIdAllocator({ prefix: 'lg-context_drawer-content' });
+    const triggerId = useIdAllocator({
+      prefix: 'lg-context_drawer-trigger',
+      id: typeof trigger !== 'function' ? trigger.props.id : undefined,
```
TheSonOfThomp, line 62: would `trigger?.props?.id ?? undefined` work here (and in other places where you check this ternary)?

packages/context-drawer/src/ContextDrawerButton/ContextDrawerButton.styles.ts
```diff
@@ -1,18 +1,33 @@
...
+  borderRadius,
+  color,
+  fontWeights,
+  spacing,
+  typeScales,
+} from '@leafygreen-ui/tokens';
+
+import { TRANSITION_DURATION } from '../constants';
+
+const BUTTON_HEIGHT = 20;
```
TheSonOfThomp, line 14: nit: use `spacing`?

packages/context-drawer/src/ContextDrawerButton/ContextDrawerButton.styles.ts
```diff
@@ -29,16 +44,30 @@ const getTriggerStyles = (theme: Theme) => {
...
 
-const getGlyphStyles = (theme: Theme) => css`
-  & svg[role='presentation'] {
-    color: ${palette.blue[theme === Theme.Dark ? 'light2' : 'dark2']};
-  }
-`;
+const getGlyphStyles = ({ isOpen, theme }: { isOpen: boolean; theme: Theme }) =>
+  cx(
+    css`
+      & svg[role='presentation'] {
```
TheSonOfThomp: does selecting by `[role="presentation"]` add anything here? It's just the one icon, so just selecting `svg` should be fine

TheSonOfThomp: (adding the styles directly to the svg would be ideal though )

TheSonOfThomp (approved):

packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
stephl3: it will once I merge in the PR https://github.com/mongodb/leafygreen-ui/pull/2930 into this branch

packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
stephl3: Assuming this is related to the `!!isOpen`? I needed to add this because the type from `useControlledValue` can be undefined

packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
stephl3: no, typescript complains about a type mismatch if `trigger` is a function

packages/context-drawer/src/ContextDrawerButton/ContextDrawerButton.styles.ts
stephl3: Hmm not sure I agree with using `spacing` tokens for anything other than margin/padding/gap. This has come up before so maybe we can discuss further during the next dev sync

packages/context-drawer/src/ContextDrawerButton/ContextDrawerButton.styles.ts
stephl3: dropped!

yes I initially added directly to the `ChevronDown` but this was overridden by styles defined in `ButtonIcon`, so I targeted it this way instead

PR #2926, opened 2025-06-27
fix(Chart): Fix CSS so consumers can adjust height

tsck:
## ✍️ Proposed changes

A charts height was accidentally set to fixed. This instead sets the height of the container which the consumer does have control over, with the chart filling the remainder of the container.

### For new components

### DIFF
--- a/charts/core/src/Chart/Chart.tsx
+++ b/charts/core/src/Chart/Chart.tsx
@@ -8,7 +8,7 @@
 import React from 'react';
 import { useSortable } from '@dnd-kit/sortable';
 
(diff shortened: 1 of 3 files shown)


charts/core/src/Chart/Chart.tsx
```diff
@@ -8,7 +8,7 @@
 import React from 'react';
 import { useSortable } from '@dnd-kit/sortable';
 
-import { cx } from '@leafygreen-ui/emotion';
```
tsck, line 11: Have to move this. Pushing shortly

PR #2933, opened 2025-06-30
LG-5146: SectionNav

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-5146](https://jira.mongodb.org/browse/LG-5146)

### Included:
- The new component, `SectionNav`.

### Not Included
- The utility to transform flat data. That is in a separate PR.

### For new components


packages/section-nav/src/Context/SectionNavContext.types.ts
```diff
@@ -0,0 +1,15 @@
...
+export type SectionNavProviderProps = DarkModeProps & {
+  /**
+   * LGIDs for SectionNav components.
+   */
+  lgIds: GetLgIdsReturnType;
+
+  /**
+   * Whether or not the component is within a SectionNav component.
+   */
+  hasContext: boolean;
```
TheSonOfThomp, line 14: I believe we should be able to determine this if `useContext` returns undefined/empty object. Nothing wrong with this approach though—it's more explicit

packages/section-nav/src/Context/SectionNavContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+  children,
+  lgIds,
+  hasContext,
+  darkMode = false,
+}: PropsWithChildren<SectionNavProviderProps>) => {
+  const SectionNavProvider = (
+    SectionNavContext as React.Context<SectionNavProviderProps>
+  ).Provider;
+
+  const SectionNavProviderData = useMemo(() => {
```
TheSonOfThomp: nit: lowercase `sectionNav...`

packages/section-nav/src/Context/SectionNavContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+  );
+
+export const SectionNavContextProvider = ({
+  children,
+  lgIds,
+  hasContext,
+  darkMode = false,
+}: PropsWithChildren<SectionNavProviderProps>) => {
+  const SectionNavProvider = (
+    SectionNavContext as React.Context<SectionNavProviderProps>
```
TheSonOfThomp: do we need the type assertion here?

packages/section-nav/src/Context/SectionNavContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { SectionNavProviderProps } from './SectionNavContext.types';
+
+export const SectionNavContext = createContext<SectionNavProviderProps>({
+  lgIds: getLgIds(),
+  hasContext: false,
+});
+
+export const useSectionNavContext = () =>
+  useContext<SectionNavProviderProps>(
+    SectionNavContext as React.Context<SectionNavProviderProps>,
```
TheSonOfThomp: do we need the type assertion here?

packages/section-nav/src/Context/SectionNavContext.tsx
```diff
@@ -0,0 +1,48 @@
...
+}: PropsWithChildren<SectionNavProviderProps>) => {
+  const SectionNavProvider = (
+    SectionNavContext as React.Context<SectionNavProviderProps>
+  ).Provider;
+
+  const SectionNavProviderData = useMemo(() => {
+    return {
+      lgIds,
+      hasContext,
+    };
```
TheSonOfThomp: I think just passing `value={{ lgIds, hasContext }}` into the provider should behave the same, since it'll trigger a re-render when those props change

packages/section-nav/src/Context/SectionNavNestedContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+}: PropsWithChildren<SectionNavNestedProviderProps>) => {
+  const SectionNavNestedProvider = (
+    SectionNavNestedContext as React.Context<SectionNavNestedProviderProps>
+  ).Provider;
+
+  const sectionNavNestedData = useMemo(() => {
+    return {
+      level,
+    };
+  }, [level]);
```
TheSonOfThomp: same here, I don't think the useMemo is necessary

packages/section-nav/src/SectionNav/SectionNav.tsx
```diff
@@ -0,0 +1,51 @@
...
+          data-testid={lgIds.root}
+          {...rest}
+          ref={forwardedRef}
+        >
+          {title && (
+            <Body
+              data-lgid={lgIds.title}
+              data-testid={lgIds.title}
+              className={getTitleStyles({ theme })}
+              as="h2"
```
TheSonOfThomp, line 39: is this always semantically correct?

packages/section-nav/src/SectionNav/SectionNav.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+    renderSectionNav({ title: 'On this page' });
+    const { getTitle } = getTestUtils();
+    expect(getTitle()?.textContent).toBe('On this page');
+  });
+
+  test('renders the correct number of items', () => {
+    renderSectionNav({});
+    const { getAllSectionNavItems } = getTestUtils();
+    const items = getAllSectionNavItems();
+    expect(items.length).toBe(7);
```
TheSonOfThomp, line 28: can we have `renderSectionNav` take in an `itemCount` prop (or similar)? It would make the `expect(items.length).toBe(7)` feel less arbitrary and like a "magic number"

packages/section-nav/src/Context/SectionNavContext.types.ts
```diff
@@ -0,0 +1,15 @@
+import { DarkModeProps } from '@leafygreen-ui/lib';
+
+import { GetLgIdsReturnType } from '../utils/getLgIds';
+
+export type SectionNavProviderProps = DarkModeProps & {
+  /**
+   * LGIDs for SectionNav components.
+   */
+  lgIds: GetLgIdsReturnType;
```
TheSonOfThomp, line 9: I like having the ids in context. Nice!

packages/section-nav/src/SectionNavItem/SectionNavItem.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+import { color, fontWeights, spacing, typeScales } from '@leafygreen-ui/tokens';
+
+import {
+  INDENTATION,
+  TRANSITION_DURATION,
+  WEDGE_BORDER_RADIUS,
+  WEDGE_WIDTH,
+} from './SectionNavItem.constants';
+
+// This component currenly only supports two levels of indentation.
```
TheSonOfThomp:
```suggestion
// This component currently only supports two levels of indentation.
```

packages/section-nav/src/SectionNavItem/SectionNavItem.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+export const getActiveStyles = ({ theme }: { theme: Theme }) => css`
+  color: ${color[theme].text.primary.default};
+  font-weight: ${fontWeights.bold};
+
+  &::before {
+    background-color: ${activeWedgeColor[theme]};
+  }
+`;
+
+export const listSyles = css`
```
TheSonOfThomp:
```suggestion
export const listStyles = css`
```

packages/section-nav/src/SectionNavItem/SectionNavItem.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+  TRANSITION_DURATION,
+  WEDGE_BORDER_RADIUS,
+  WEDGE_WIDTH,
+} from './SectionNavItem.constants';
+
+// This component currenly only supports two levels of indentation.
+// If more levels are needed, this function can be extended.
+const getIndentation = (level: number) => {
+  if (level === 1) return INDENTATION[1];
+  return INDENTATION[2];
```
TheSonOfThomp: are there really only 3 levels of indentation?

packages/section-nav/src/SectionNavItem/SectionNavItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+          data-testid={`${lgIds.item}`}
+          data-lgid={`${lgIds.item}`}
+          data-level={level}
+          {...rest}
+        >
+          {label}
+        </a>
+        {children && (
+          <SectionNavNestedContextProvider level={level + 1}>
+            <ol className={listSyles}>{children}</ol>
```
TheSonOfThomp:
```suggestion
            <ol className={listStyles}>{children}</ol>
```

packages/section-nav/src/SectionNavItem/SectionNavItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    if (!hasContext) {
+      console.error(
+        'SectionNavItem must be used within a SectionNav component',
+      );
+    }
+
+    // This is a warning for consumers to know that the component currently only supports 2 levels of nesting.
+    if (level > 2) {
+      console.warn(
```
TheSonOfThomp: `consoleOnce`

packages/section-nav/src/SectionNavItem/SectionNavItem.types.ts
```diff
@@ -0,0 +1,19 @@
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+
+export interface SectionNavItemProps
+  extends Omit<HTMLElementProps<'a'>, 'href'> {
+  /**
+   *  Whether the item is active.
+   */
+  active?: boolean;
```
TheSonOfThomp, line 8: should this be `isActive`?

packages/section-nav/src/SectionNavItem/SectionNavItem.constants.ts
```diff
@@ -0,0 +1,10 @@
+import { spacing, transitionDuration } from '@leafygreen-ui/tokens';
+
+export const WEDGE_WIDTH = 2;
+export const WEDGE_BORDER_RADIUS = 100;
+export const TRANSITION_DURATION = transitionDuration.default;
+
+export const INDENTATION = {
+  1: spacing[400],
+  2: spacing[900],
+};
```
TheSonOfThomp: I don't love having an object with numbered keys starting at `1`... Can we change this, either to an array, or more semantic keys?

packages/section-nav/src/SectionNavItem/SectionNavItem.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      expect(activeItem?.isActive()).toBeFalsy();
+    });
+  });
+
+  test('calls onClick when clicked', () => {
+    const onClick = jest.fn();
+    render(
+      <SectionNav>
+        <SectionNavItem onClick={onClick} href="#section-1" label="Section 1" />
+        <SectionNavItem href="#section-2" label="Section 2" />
```
TheSonOfThomp, line 55: should we test how items with just an `href` behave?

packages/section-nav/src/SectionNavItem/SectionNavItem.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+    label: 'Section Nav Item 1',
+    active: false,
+    href: '#section-1',
+  },
+  argTypes: {
+    darkMode: storybookArgTypes.darkMode,
+  },
+} satisfies StoryMetaType<typeof SectionNavItem>;
+
+export const LiveExample: StoryObj<typeof SectionNavItem> = {
```
TheSonOfThomp, line 52: should we disable the liveExample on chromatic?

packages/section-nav/src/testing/getTestUtils.tsx
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Returns the element using the `data-lgid` data attribute or `null` if no elements match.
+   * Will throw if more than one match is found.
+   */
+  const querySectionNav = () => queryByLgId!<HTMLElement>(lgIds.root);
+
+  /**
+   * Returns the title element of the SectionNav.
+   */
+  const getTitle = () => {
```
TheSonOfThomp: Not 100% sure what our pattern is, but I think the names should be a bit more specific? 

```suggestion
  const getSectionTitle = () => {
```

packages/section-nav/src/utils/getLgIds.ts
```diff
@@ -0,0 +1,10 @@
+export const DEFAULT_LGID_ROOT = 'lg-section-nav';
+
+export const getLgIds = (root: `lg-${string}` = DEFAULT_LGID_ROOT) =>
```
TheSonOfThomp: Do we have a shared `LGID` type?

packages/section-nav/src/utils/getLgIds.ts
```diff
@@ -0,0 +1,10 @@
+export const DEFAULT_LGID_ROOT = 'lg-section-nav';
```
TheSonOfThomp:
```suggestion
export const DEFAULT_LGID_ROOT = 'lg-section_nav';
```

packages/section-nav/src/SectionNavItem/SectionNavItem.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+    background-color: transparent;
+    border-top-right-radius: ${WEDGE_BORDER_RADIUS}px;
+    border-bottom-right-radius: ${WEDGE_BORDER_RADIUS}px;
+  }
+`;
+
+const activeWedgeColor: Record<Theme, string> = {
+  [Theme.Light]: palette.black,
+  [Theme.Dark]: palette.gray.light2,
+};
```
TheSonOfThomp, line 88: should these be green?

TheSonOfThomp (commented):
Left a few nits. Overall looks good though!

packages/section-nav/src/SectionNavItem/SectionNavItem.styles.ts
shaneeza: This goes back to the discussion on the depth prop. Sooa is adamant that this component should only support two levels of nesting, but we wanted to build this in a way that if that changes in the future, we could easily support it. 

I could have made so if someone adds more levels of nesting, the padding would reflect that, but I'm unsure of what the padding should be since there isn't really a pattern. The first level is 16px, and the second level is 36px.

packages/section-nav/src/SectionNavItem/SectionNavItem.spec.tsx
shaneeza: test what happens on click with only an href?

packages/section-nav/src/SectionNavItem/SectionNavItem.stories.tsx
shaneeza: Yes, will do

packages/section-nav/src/Context/SectionNavContext.tsx
shaneeza: we don't

packages/section-nav/src/SectionNavItem/SectionNavItem.types.ts
shaneeza: We use `active` for similar components, like `SideNav`, `menuItem`, `IconButton`, etc.

sooachung (approved):
LGTM!

packages/section-nav/src/SectionNavItem/SectionNavItem.styles.ts
shaneeza: [Figma](https://www.figma.com/design/4h2VwjCuJUbeZ7hzD2J1rq/LeafyGreen-Design-System?m=auto&node-id=42539-19379&t=I0juiXoWPIDehh10-1)

Screenshot 2025-07-09 at 11 12 37 AM

packages/section-nav/src/SectionNavItem/SectionNavItem.constants.ts
shaneeza: Why don't you love an object with numbered keys starting at `1`?

packages/section-nav/src/SectionNav/SectionNav.spec.tsx
shaneeza: What would the `itemCount` prop do?

packages/section-nav/src/SectionNav/SectionNav.tsx
shaneeza: This component should primarily be used for `h2`s and `h3`s on a page. I think setting this as an `h2` is a good way to portray this as a navigational block. However, as we discussed during dev sync, we're not fully aware of how consumers might utilize this component. Maybe a good compromise is to set the default as `h2` and create a prop, like, `titleAs`, for consumers to control. Thoughts?

packages/section-nav/src/SectionNavItem/SectionNavItem.constants.ts
shaneeza: Maybe something like: 

```js
export const INDENTATION = {
  'level1': spacing[400],
  'level2': spacing[900],
};
```

packages/section-nav/src/SectionNav/SectionNav.tsx
TheSonOfThomp: I think we can hard-code it for now (assuming `h2` is how most other systems render a ToC)

packages/section-nav/src/SectionNavItem/SectionNavItem.spec.tsx
TheSonOfThomp: Yeah, although I don't think jsdom allows testing that, so I'm not sure what we'd `expect`

packages/section-nav/src/SectionNav/SectionNav.spec.tsx
TheSonOfThomp: Would'nt bea "prop" per-se on the component, but just an additional arg that's passed into the test render function to avoid expecting magic numbers. Not necessary, I just was confused why we were expecting `7` items
e.g.
```ts
const ITEMS = 7;
renderSectionNav({itemCount: ITEMS});
...
expect(items.length).toBe(ITEMS);
```

packages/section-nav/src/SectionNavItem/SectionNavItem.constants.ts
TheSonOfThomp: > Why don't you love an object with numbered keys starting at 1?

It looks like it's supposed to be an array, but it's not. When I read `INDENTATION[2]` somewhere, I'd expect to be able to call other `Array` methods on `INDENTATION`, or get the first index as `INDENTATION[0]`

TheSonOfThomp: Also shouldn't the indentation padding just be a some multiple of the base padding  x `level`?

packages/section-nav/src/SectionNav/SectionNav.spec.tsx
shaneeza: I see. `renderSectionNav` internally renders seven `SectionNavItem`s. It's not customizable, so it always renders seven items. It mimics the consumer rendering `SectionNav` and testing to see how many children it has. Does that clear things up?

packages/section-nav/src/SectionNavItem/SectionNavItem.constants.ts
shaneeza: Thanks for clarifying. And for the padding, the first level is 16px, then it jumps to 36px. So that's `16 * 2.25`, not `16 * 2`. I'm not fully sure what the next level should be, but my guess is it would be:

```js
16 // level 1
16*2.25 = 36  // level 2
36*2.25 = 81 // level 3?
```

I'm going to double-check with Sooa.

shaneeza: Since there is no 3rd level in the specs, I could assume that this is the padding:

```js
16 // level 1
16+20 = 36  // level 2
16+40 = 56 // level 3?
```

and get rid of the obj and do this based on the pattern above:
```
const getIndentation = (level: number) => {
  if (level === 1) return spacing[400];

  return spacing[400] + 20 * (level - 1);
};
```

So if someone adds a 3rd level, it will add padding, but we will still log a warning that only two levels are currently recommended

It would look like this:

Screenshot 2025-07-10 at 5 36 01 PM

However since there is no clear pattern, It doesn't feel correct to assume a pattern

PR #2934, opened 2025-07-01
[breaking] feat(Chart): add enableGroupTooltipSync prop for grouped charts

stephl3:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/.changeset/chart-enableGroupTooltipSync.md
+++ b/.changeset/chart-enableGroupTooltipSync.md
@@ -1,0 +1,9 @@
+---
+'@lg-charts/core': major
+---
+
+Add `enableGroupTooltipSync` prop to `Chart` component to turn on/off tooltip syncing for grouped charts.
+
+Previously, all charts only synced axis pointers and did not sync tooltips. This was done to prevent longer tooltips from visually colliding with other long tooltips.
+
+`enableGroupTooltipSync` defaults to `true` and will always sync tooltips. The prop can be explicitly set to `false` to achieve the previous behavior of only syncing axis pointers and not syncing tooltips.


charts/core/src/Chart/Chart.tsx
```diff
@@ -30,26 +30,28 @@ import { useChart } from './hooks';
 
 export function Chart({
   children,
+  className,
   darkMode: darkModeProp,
+  dragId = '',
+  enableGroupTooltipSync = true,
```
tsck, line 36: If the default state is enabled, I'm inclined to think that the prop should be `disableGroupTooltipSync`. 

#### Current
```tsx
<Chart enableGroupTooltipSync={false} />
```

#### Proposed
```tsx
<Chart disableGroupTooltipSync />
```

It's just a bit more concise and to the point, and is in my opinion more inline with standard HTML elements that have the ability to disable things. Any thoughts?

stephl3: I like keeping as-is. Aligning positive phrasing and mapping enabling to true and disabling to false feels more readable to me. `disableGroupTooltipSync={false}` in particular requires more mental overhead when I read that. (This is also endorsed by Gemini 🤣)

I agree with `disabled`/`isDisabled` for components that rely on the native HTML `disabled` attribute but would prefer sticking with this `enable*` pattern for props handling more business logic. Happy to discuss with the team during dev sync if you feel strongly on it. What do you think?

tsck: I'm still not sure I agree with Gemini lol. My issue is that I never have to write or process `disableGroupTooltipSync={false}` from the consumer's standpoint, since that's the default. What I do need to process in the current implementation is `<Chart enableGroupTooltipSync={false} />` which to me is more verbose and less scannable than just `<Chart disableGroupTooltipSync />`. 

I don't feel strongly to the extent that I'll be a hard blocker, but yeah, let's just check in with the team. Happy to go with the consensus on this.

stephl3: I see what you're saying. Yes let's discuss with the team. Happy to swap if the other phrasing makes more sense

stephl3: synced with team and landed on using current prop naming

tsck (approved):
Dinesh Thumb

PR #2935, opened 2025-07-01
Add title attribute to chat input button

greggb:
## ✍️ Proposed changes

The submit button for the lg-chat InputBar often has only an icon or just the text `enter`. By adding a title we can be more descriptive for screen readers and allow more realistic targeting in tests.

Would appreciate some feedback on using the `title` attribute instead of `aria-label`. The issue with `aria-label` is that it should match the contents of the button ([a WCAG 2.1 requirement](https://www.w3.org/WAI/WCAG21/quickref/#label-in-name)), but `enter` is not very descriptive of the action. On the flip-side `title` has the side effect of a system tooltip on hover, so it does change the look and feel of the component.

### For new components

### DIFF
--- a/chat/input-bar/src/InputBar/InputBar.tsx
+++ b/chat/input-bar/src/InputBar/InputBar.tsx
@@ -433,6 +433,7 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
                   className={cx({
                     [sendButtonDisabledStyles]: isSendButtonDisabled(),
                   })}
+                  title="Send message"
                 >
                   {shouldRenderButtonText && 'Enter'}
                 </Button>
--- a/.changeset/spotty-areas-work.md
+++ b/.changeset/spotty-areas-work.md
@@ -1,0 +1,5 @@
+---
+'@lg-chat/input-bar': patch
+---
+
+Add an accessible title to the submit button
--- a/chat/input-bar/src/InputBar/InputBar.spec.tsx
+++ b/chat/input-bar/src/InputBar/InputBar.spec.tsx
@@ -71,7 +71,7 @@ describe('packages/input-bar', () => {
 
     test('fires when enter is clicked', () => {
       const textarea = screen.getByRole('textbox');
-      const sendButton = screen.getByRole('button');
+      const sendButton = screen.getByTitle('Send message');
       userEvent.type(textarea, testText);
       userEvent.click(sendButton);
 


chat/input-bar/src/InputBar/InputBar.tsx
```diff
@@ -433,6 +433,7 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
                   className={cx({
                     [sendButtonDisabledStyles]: isSendButtonDisabled(),
                   })}
+                  title="Send message"
```
shaneeza, line 436: > The issue with aria-label is that it should match the contents of the button ([a WCAG 2.1 requirement](https://www.w3.org/WAI/WCAG21/quickref/#label-in-name)), but enter is not very descriptive of the action.

According to [this page](https://www.w3.org/WAI/WCAG22/Techniques/failures/F96), I think it's saying that the aria-label does not have to match the contents of the button, but it should _contain_ it. 

> When the use of these techniques results in an accessible name in which the exact string of the visible label does not occur in the accessible name, speech users may be unable to activate that control.

So something like the following would be considered accessible, since the visible label `Enter` is present within the `aria-label`:
```html
<button aria-label='Enter message'>Enter</button>
```

Additionally, since the button can sometimes render as an icon-only button, `aria-label` would be beneficial over `title` since `aria-label` provides a text label for elements that might not have a visible one, while `title` is more for additional context. 

Given that the button may include or exclude visible text, relying on `aria-label` seems like the most robust and accessible solution.

greggb: Thanks @shaneeza - Is `Enter Message` the right label for a submit button? This seems like a good compromise, but in isolation it sounds like the label for an input, not submit, action. If it wasn't so wordy I would suggest `aria-label="press enter to submit message"`. Any thoughts before I make this change?

shaneeza: I did a little more research, and it seems that the accessible name doesn't just need to contain the visible text; it needs to match or clearly include what you see on the screen.

>When speech input users interact with a web page, they usually speak a command followed by the reference to some visible label (like the text in a button or a link, or the text labelling some input). If the visible label does not match the accessible name of the control, speech users may be unable to directly activate that control.

For example, speech input users interact with web pages by speaking commands like `click send` or `click enter`. The tools they use rely on the accessible label to match what the user says.

So if a button is labeled `Enter`, but its accessible name is `Press enter to submit`, that mismatch can cause problems. The word `Enter` is present, but the label now sounds like an instruction instead of a button name. The speech tool might not recognize it when the user says `Click Enter`.

In contrast, something like `Enter message` still feels like an action and would likely be recognized correctly.

That said, I think that `Enter Message` makes sense as an accessible label because it's on a button. It's universally understood that after typing a message, you would press `Enter` to submit/send it.

However, maybe this brings up another point that instead of `Enter`, we should use `Send` or `Submit`. I'm not sure why we went with `Enter`.

shaneeza: @greggb, any thoughts?

greggb: Happy to update this to have `aria-label="Enter message"`. I wasn't sure if you wanted to further modify the action text. I think for the scope of this change keeping it as-is feels more appropriate and just adding the label. Agreed?

shaneeza: Agree!

greggb:
I'm back in the office today, will update this PR shortly

PR #2936, opened 2025-07-01
LG-5316: progress bar animations

charcoalyy:
## ✍️ Proposed changes

- shimmer and looping animation for determinate and indeterminate variants, respectively
- smooth opacity transition from indeterminate to determinate variants
- refactor storybook helper function to render dynamic changes to the progress bar after fixed timeouts
- create a constants file purely for styles due to its bulk
- create a general constants file

🎟 _Jira ticket:_ [LG-5316](https://jira.mongodb.org/browse/LG-5316)

### For new components

### DIFF
--- a/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react';
+import { StoryMetaType } from '@lg-tools/storybook-utils';
 import { StoryObj } from '@storybook/react';
 import { expect, waitFor, within } from '@storybook/test';
 
@@ -31,13 +32,14 @@ const DynamicProgressBar = ({
   return <ProgressBar {...currentProps} />;
 };
 
-export default {
+const meta: StoryMetaType<typeof ProgressBar> = {
   title: 'Components/ProgressBar/Interactions',
   component: ProgressBar,
   parameters: {
     default: 'LiveExample',
   },
 };
+export default meta;
 
 export const WithChangingValue: StoryObj<typeof ProgressBar> = {
   args: {
--- a/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
@@ -36,7 +36,7 @@ const meta: StoryMetaType<typeof ProgressBar> = {
   title: 'Components/ProgressBar/Interactions',
   component: ProgressBar,
   parameters: {
-    default: 'LiveExample',
+    default: null,
   },
 };
 export default meta;
--- a/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
@@ -8,6 +8,7 @@ import { ProgressBar, ProgressBarProps, Type } from '.';
 const testValues = {
   value: 53,
   maxValue: 200,
+  timeoutBuffer: 500,
 };
 
 const DynamicProgressBar = ({
@@ -46,7 +47,7 @@ export const WithChangingValue: StoryObj<typeof ProgressBar> = {
     type: Type.Loader,
     value: testValues.value,
     maxValue: testValues.maxValue,
-    formatValue: 'percentage',
+    formatValue: 'fraction',
   },
   render: initialArgs => (
     <DynamicProgressBar
@@ -69,8 +70,7 @@ export const WithChangingValue: StoryObj<typeof ProgressBar> = {
           testValues.maxValue.toString(),
         );
       },
-      // transition timeout with buffer time of 500ms
-      { timeout: 2000 },
+      { timeout: 1500 + testValues.timeoutBuffer },
     );
   },
 };
@@ -103,8 +103,7 @@ export const WithChangingDescriptions: StoryObj<typeof ProgressBar> = {
 
     const finalText = await waitFor(
       () => canvas.getByText('Even newer helper text...!'),
-      // transition timeout with buffer time of 500ms
-      { timeout: 4000 },
+      { timeout: 3500 + testValues.timeoutBuffer },
     );
 
     expect(finalText).toBeInTheDocument();
@@ -144,8 +143,7 @@ export const IndeterminateToDeterminate: StoryObj<typeof ProgressBar> = {
           testValues.maxValue.toString(),
         );
       },
-      // transition timeout with buffer time of 500ms
-      { timeout: 4000 },
+      { timeout: 3500 + testValues.timeoutBuffer },
     );
   },
 };
--- a/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
@@ -1,0 +1,149 @@
+import React, { useEffect, useState } from 'react';
+import { StoryObj } from '@storybook/react';
+import { expect, waitFor, within } from '@storybook/test';
+
+import { ProgressBar, ProgressBarProps, Type } from '.';
+
+const testValues = {
+  value: 53,
+  maxValue: 200,
+};
+
+const DynamicProgressBar = ({
+  transitions,
+  ...initialProps
+}: ProgressBarProps & {
+  transitions?: Array<[number, ProgressBarProps]>;
+}) => {
+  const [currentProps, setCurrentProps] =
+    useState<ProgressBarProps>(initialProps);
+
+  useEffect(() => {
+    const timers = transitions?.map(([delay, newProps]) =>
+      setTimeout(() => {
+        setCurrentProps(newProps);
+      }, delay),
+    );
+
+    return () => timers?.forEach(clearTimeout);
+  }, [transitions]);
+
+  return <ProgressBar {...currentProps} />;
+};
+
+export default {
+  title: 'Components/ProgressBar/Interactions',
+  component: ProgressBar,
+  parameters: {
+    default: 'LiveExample',
+  },
+};
+
+export const WithChangingValue: StoryObj<typeof ProgressBar> = {
+  args: {
+    type: Type.Loader,
+    value: testValues.value,
+    maxValue: testValues.maxValue,
+    formatValue: 'percentage',
+  },
+  render: initialArgs => (
+    <DynamicProgressBar
+      transitions={[[1500, { ...initialArgs, value: testValues.maxValue }]]}
+      {...initialArgs}
+    />
+  ),
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement);
+    const progressBar = canvas.getByRole('progressbar');
+
+    expect(progressBar).toHaveAttribute(
+      'aria-valuenow',
+      testValues.value.toString(),
+    );
+
+    await waitFor(
+      () => {
+        expect(progressBar.getAttribute('aria-valuenow')).toBe(
+          testValues.maxValue.toString(),
+        );
+      },
+      // transition timeout with buffer time of 500ms
+      { timeout: 2000 },
+    );
+  },
+};
+
+export const WithChangingDescriptions: StoryObj<typeof ProgressBar> = {
+  args: {
+    ...WithChangingValue.args,
+    description: <span>Helper text</span>,
+  },
+  render: initialArgs => (
+    <DynamicProgressBar
+      transitions={[
+        [
+          1500,
+          { ...initialArgs, description: <span>New helper text...</span> },
+        ],
+        [
+          3500,
+          {
+            ...initialArgs,
+            description: <span>Even newer helper text...!</span>,
+          },
+        ],
+      ]}
+      {...initialArgs}
+    />
+  ),
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement);
+
+    const finalText = await waitFor(
+      () => canvas.getByText('Even newer helper text...!'),
+      // transition timeout with buffer time of 500ms
+      { timeout: 4000 },
+    );
+
+    expect(finalText).toBeInTheDocument();
+  },
+};
+
+export const IndeterminateToDeterminate: StoryObj<typeof ProgressBar> = {
+  args: {
+    type: Type.Loader,
+    isIndeterminate: true,
+  },
+  render: initialArgs => (
+    <DynamicProgressBar
+      transitions={[
+        [
+          3500,
+          {
+            type: Type.Loader,
+            isIndeterminate: false,
+            value: testValues.maxValue,
+            maxValue: testValues.maxValue,
+          },
+        ],
+      ]}
+      {...initialArgs}
+    />
+  ),
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement);
+    const progressBar = canvas.getByRole('progressbar');
+
+    expect(progressBar).not.toHaveAttribute('aria-valuenow');
+
+    await waitFor(
+      () => {
+        expect(progressBar.getAttribute('aria-valuenow')).toBe(
+          testValues.maxValue.toString(),
+        );
+      },
+      // transition timeout with buffer time of 500ms
+      { timeout: 4000 },
+    );
+  },
+};
--- a/packages/progress-bar/src/ProgressBar.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.stories.tsx
@@ -76,7 +76,10 @@ const meta: StoryMetaType<typeof ProgressBar> = {
     },
     formatValue: {
       control: { type: 'select' },
-      options: [...FORMAT_VALUE_TYPES, undefined],
+      options: [...FORMAT_VALUE_TYPES, 'custom example', undefined],
+      mapping: {
+        'custom example': (value: number) => `Currently at ${value} GB`,
+      },
     },
     isIndeterminate: {
       if: { arg: 'type', eq: Type.Loader },
@@ -86,7 +89,8 @@ const meta: StoryMetaType<typeof ProgressBar> = {
       // if: { arg: 'type', eq: Type.Meter } OR { arg: 'isIndeterminate', eq: false }, // TODO: not supported
       description:
         '**Not available** if both type=loader and isIndeterminate=true',
-      control: { type: 'number' },
+      control: { type: 'none' },
+      defaultValue: testValues.maxValue,
     },
     variant: {
       if: { arg: 'type', eq: Type.Loader },
@@ -102,9 +106,15 @@ const meta: StoryMetaType<typeof ProgressBar> = {
       // if: { arg: 'type', neq: Type.Meter } AND { arg: 'isIndeterminate', neq: true }, // TODO: not supported
       description:
         '**Not available** if either type="meter" or isIndeterminate=true',
-      control: { type: 'boolean' },
+      control: { type: 'none' },
+      defaultValue: false,
     },
   },
+  decorators: [
+    (Story, context) => {
+      return <Story {...context} />;
+    },
+  ],
 };
 export default meta;
 
--- a/packages/progress-bar/src/ProgressBar.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.stories.tsx
@@ -167,10 +167,48 @@ export const WithValueDisplay: StoryObj<typeof ProgressBar> = {
   },
 };
 
-export const WithDescription: StoryObj<typeof ProgressBar> = {
+export const WithDescriptions: StoryObj<typeof ProgressBar> = {
   args: {
     ...DeterminateLoader.args,
     description: <span>Helper text</span>,
+  },
+  render: function TransitioningProgressBar(props: ProgressBarProps) {
+    const [newProps, setNewProps] = useState({});
+
+    useEffect(() => {
+      const timeout1 = setTimeout(() => {
+        setNewProps({
+          ...DeterminateLoader.args,
+          description: <span>New helper text...</span>,
+        });
+
+        timeout2 = setTimeout(() => {
+          setNewProps({
+            ...DeterminateLoader.args,
+            description: <span>Even newer helper text...!</span>,
+          });
+        }, 2000);
+      }, 1500);
+
+      let timeout2: ReturnType<typeof setTimeout>;
+
+      return () => {
+        clearTimeout(timeout1);
+        clearTimeout(timeout2);
+      };
+    }, [props.value]);
+
+    return <ProgressBar {...props} {...newProps} />;
+  },
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement);
+
+    const finalText = await waitFor(
+      () => canvas.getByText('Even newer helper text...!'),
+      { timeout: 4000 },
+    );
+
+    expect(finalText).toBeInTheDocument();
   },
 };
 
--- a/packages/progress-bar/src/ProgressBar.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.stories.tsx
@@ -90,7 +90,6 @@ const meta: StoryMetaType<typeof ProgressBar> = {
       description:
         '**Not available** if both type=loader and isIndeterminate=true',
       control: { type: 'none' },
-      defaultValue: testValues.maxValue,
     },
     variant: {
       if: { arg: 'type', eq: Type.Loader },
@@ -107,7 +106,6 @@ const meta: StoryMetaType<typeof ProgressBar> = {
       description:
         '**Not available** if either type="meter" or isIndeterminate=true',
       control: { type: 'none' },
-      defaultValue: false,
     },
   },
   decorators: [
@@ -126,6 +124,7 @@ export const LiveExample: StoryObj<typeof ProgressBar> = {
     label: 'Label',
     description: 'Helper text',
     isIndeterminate: false,
+    enableAnimation: false,
   },
 };
 
--- a/packages/progress-bar/src/ProgressBar.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.stories.tsx
@@ -83,7 +83,7 @@ const meta: StoryMetaType<typeof ProgressBar> = {
       control: { type: 'boolean' },
     },
     maxValue: {
-      if: { arg: 'isIndeterminate', neq: true },
+      if: { arg: 'isIndeterminate', neq: true }, // TODO: FIX THIS FOR METERS
       control: { type: 'number' },
     },
     variant: {
@@ -99,8 +99,8 @@ const meta: StoryMetaType<typeof ProgressBar> = {
     enableAnimation: {
       if: {
         and: [
-          { arg: 'type', eq: Type.Loader },
           { arg: 'isIndeterminate', eq: false },
+          { arg: 'type', eq: Type.Loader }, // TODO: FIX THIS FOR METERS
         ],
       },
       control: { type: 'boolean' },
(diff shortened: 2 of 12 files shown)


packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -248,3 +248,45 @@ export const MeterVariants: StoryObj<typeof ProgressBar> = {
     },
   },
 };
+
+export const IndeterminateToDeterminate: StoryObj<typeof ProgressBar> = {
+  args: {
+    type: Type.Loader,
+    isIndeterminate: true,
+  },
+  render: function TransitioningProgressBar(props: ProgressBarProps) {
```
charcoalyy: contemplated making `SimulatedProgressBar` more general of a wrapper component, such that it can take any `newProps` and change to those new props after a given timeout. that way it can be re-used even here. however, there are some complications with taking `Partial`'s of `ProgressBarProps` due to discriminated union. thoughts on if it's worth it?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -40,10 +42,39 @@ export function ProgressBar(props: ProgressBarProps) {
...
+  // if progress bar was previously indeterminate and is turning determinate, apply fadeout transition
+  const [animationMode, setAnimationMode] = useState<AnimationMode>(
+    getAnimationMode({
+      type,
+      isIndeterminate,
+      enableAnimation,
+    }),
+  );
+
+  useEffect(() => {
```
charcoalyy, line 60: how do we feel about this from a code maintainer perspective? it's very particular, but it's also to hit a very particular spec.

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -120,63 +137,184 @@ export const getBarTrackStyles = ({
...
+    color,
+    disabled,
+    width: getPercentage(value, maxValue),
+  });
 
-  if (value && maxValue)
-    typedBarFillStyles = getDeterminateBarFillStyles(
-      getPercentage(value, maxValue),
-    );
+  const animatedDeterminate = !disabled
```
charcoalyy: or does it make more sense to pass disabled into `getAnimatedDeterminateBarFillStyles` to match `getDeterminateBarFillStyles`?

packages/progress-bar/src/ProgressBar.stories.tsx
shaneeza: That makes sense. Why do you need to use partial `ProgressBarProps`? 

But also, it's such a specific example. I don't see the harm in leaving as-is.

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -248,3 +248,45 @@ export const MeterVariants: StoryObj<typeof ProgressBar> = {
...
+
+    useEffect(() => {
+      const timeout = setTimeout(() => {
+        setNewProps({
+          type: Type.Loader,
+          isIndeterminate: false,
+          value: 200,
+          maxValue: 200,
+        });
+      }, 3500);
```
shaneeza: Why 3500?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -7,10 +7,23 @@ import {
   spacing as spacingToken,
 } from '@leafygreen-ui/tokens';
 
-import { Color, Size } from './ProgressBar.types';
+import { AnimationMode, Color, Size } from './ProgressBar.types';
 import { getPercentage } from './ProgressBar.utils';
 
-const progressBarSizeStyles = {
+export const FADEOUT_DURATION = 500;
```
shaneeza: We usually keep constants in a `constants.ts` file so that they're all in one place.

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -7,10 +7,23 @@ import {
...
+
+const opacityAlphaChannels = {
```
shaneeza: I'm not that familiar with alpha channels; how did you determine these values?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -7,10 +7,23 @@ import {
...
+  50: '80',
+  75: 'BF',
+  100: 'FF',
+};
+
+const fadePalette = {
+  blue: '#C3E7FE',
```
shaneeza: It looks like we can use tokens here.

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -120,63 +137,190 @@ export const getBarTrackStyles = ({
...
 }: {
   theme: Theme;
   color: Color;
   disabled?: boolean;
+  width: number;
 }) => css`
-  height: 100%;
-  border-radius: inherit;
+  width: ${width}%;
+  transition: width 0.5s ease-in-out;
```
shaneeza: Can we create a constant for the duration?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -120,63 +137,190 @@ export const getBarTrackStyles = ({
...
+        background: linear-gradient(
+          90deg,
+          ${selectedColorStyle.bar} 0%,
+          ${selectedColorStyle.shimmerFade} 50%,
+          ${selectedColorStyle.bar} 100%
+        );
+        background-size: 200% 100%;
+        animation: shimmer 3s linear infinite;
+      }
+      @keyframes shimmer {
```
shaneeza: We can break out these keyframes. [Check out this example from Toast](https://github.com/mongodb/leafygreen-ui/blob/main/packages/toast/src/InternalToast/ProgressBar/ProgressBar.styles.ts#L26)!

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -120,63 +137,190 @@ export const getBarTrackStyles = ({
...
+  const selectedColorStyle = barColorStyles[theme][color];
+
+  return css`
+    width: 100%;
+    background-color: transparent;
+    &::before {
+      content: '';
+      position: absolute;
+      top: 0;
+      left: -33%;
```
shaneeza: Can we create variables for these values?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
shaneeza: I noticed that the `Transition` animation uses `FADEOUT_DURATION`:

```css
transition: opacity ${FADEOUT_DURATION}ms ease-out,
    width 100ms ease-out ${FADEOUT_DURATION - 100}ms;
```
Instead of setting a timeout that matches this duration, we could listen for when the transition ends by using `onTransitionEnd` on the div with the transition.

So on [Ln 126](https://github.com/mongodb/leafygreen-ui/pull/2936/files#diff-d78c22ea1ef46f7305044c0cfa9a5ab55f818204dec05c02f1fd0d6feec8c4faR126), you could do something like this, and we shouldn't need the timeout inside the `useEffect` anymore since we're leaving the timing to the browser:

```js
onTransitionEnd={() => {
  if (animationMode === AnimationMode.Transition) {
    setAnimationMode(
      getAnimationMode({
        type,
        isIndeterminate,
        enableAnimation,
      }),
    );
  }
}}
```

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -164,3 +164,11 @@ export interface ResolvedProgressBarProps {
   /** When `true`, enables shimmer animation for long-running processes. Not available for meters or if `isIndeterminate` is `true` for loaders. */
   enableAnimation: boolean;
 }
+
+export const AnimationMode = {
```
shaneeza, line 168: If I'm understanding correctly, there are 4 animation modes:

1. none - animations are not enabled(currently only applies to determinate)
2. determinate - animates if animations are enabled
3. indeterminate - always animates
4. transition - always animates

Is this correct?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -120,63 +137,190 @@ export const getBarTrackStyles = ({
...
+        90deg,
+        ${palette.transparent} 0%,
+        ${selectedColorStyle.bar}${opacityAlphaChannels[75]} 25%,
+        ${selectedColorStyle.bar}${opacityAlphaChannels[100]} 50%,
+        ${selectedColorStyle.bar}${opacityAlphaChannels[75]} 75%,
+        ${palette.transparent} 100%
+      );
+      animation: cycle 1.5s linear infinite;
+    }
+    @keyframes cycle {
```
shaneeza: This looks really great! I think it's missing a little bit of a pause towards the end. Looking at the prototype, towards the end, the animation almost stops and then continues to the end.

prototype:
https://jmp.sh/ju4w2ZMx

storybook:
https://jmp.sh/rpwgudoi

I wonder if we need to add a little pause somewhere, like this (this is just a rough guess):
```css
50% {
  left: 17%;
  width: 66%;
}
54% {
  left: 17%;
  width: 66%;
}
75% {
  left: 67%;
  width: 33%;
}
```

packages/progress-bar/src/ProgressBar.stories.tsx
charcoalyy: i was figuring to reduce the repetition of repeating other props that stay the same - but that makes sense!

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
charcoalyy: oh!! that's awesome, TIL!! thank you :)

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: i wanted to avoid having to convert to RGBA for opacity since we're given hexes already, so i opted to use the [8-digit hexadecimal](https://davidwalsh.name/hex-opacity). now that you mention it, let me triple check if it's supported in CSS 🤔

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: in this case, these are colors custom to the progress bar component! :)

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: yes! to clarify, are you referring to the -33%, 33%, etc.?

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
charcoalyy: that is correct! i may try to workshop these names, but i didn't want to diverge too far from the props terminology. any thoughts?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: ooh. great eye, let me experiment with this :) !

packages/progress-bar/src/ProgressBar.stories.tsx
charcoalyy: hm. i chose an arbitrary value here just to see enough movement before the transition. what would you recommend to make it less arbitrary?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: awesome! that makes sense especially given i'm using these across multiple files. one Q, i was debating also taking out all the style constants to a separate file (e.g., `barSizeStyles`, `barColorStyles`) to make the main styles file smaller, but wondered if this would be too much separation. any thoughts?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: neato! i'm assuming the main benefit of this is the readability of the larger `getXXXStyles` functions, since the animation is one-off and unlikely to be re-used?

packages/progress-bar/src/ProgressBar.stories.tsx
shaneeza: I think this is fine. Would you consider leaving a comment on why you chose this value for context and creating a variable for it? I'm guessing the 4000 below is the 3500 + a little extra?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
shaneeza: yess

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
shaneeza: Good to know! We usually use the `transparentize` package from `polished` to help with opacity with hexs. It does the converting for us.
```js
${transparentize(0.85, '#C3E7FE')}
```

charcoalyy: amazing! thank you

charcoalyy: would you mind pointing me to a current usage in our codebase? i can't seem to find one 😅

shaneeza: https://github.com/mongodb/leafygreen-ui/blob/2622c191795c40b587c9f61142eb8e9f4c0d2d01/packages/select/src/styleSets.ts#L43

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
shaneeza: These names are great and very clear. I was thinking of a way that we can make this more intuitive, what if we adjusted the naming slightly so that the names start with the type, so:

```js
export const AnimationMode = {
  DeterminateBase: 'determinate-base',
  DeterminateAnimated: 'determinate-animated',
  Indeterminate: 'indeterminate',
  Transition: 'indeterminate-to-determinate-transition',
};
```

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
shaneeza: I think that's fine. You can create `ProgressBar.constants.ts` and add them there if you want.

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
shaneeza: Yeah. I think it's also a little cleaner, and I believe it's the recommended way to use [keyframes with emotion](https://emotion.sh/docs/keyframes).

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -166,11 +167,49 @@ export const WithValueDisplay: StoryObj<typeof ProgressBar> = {
   },
 };
 
-export const WithDescription: StoryObj<typeof ProgressBar> = {
+export const WithDescriptions: StoryObj<typeof ProgressBar> = {
   args: {
     ...DeterminateLoader.args,
     description: <span>Helper text</span>,
   },
+  render: function TransitioningProgressBar(props: ProgressBarProps) {
```
shaneeza: Are consumers responsible for changing the descriptions?

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
charcoalyy: i like that! thank you.

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: just realized i had default filtered my VScode search space is why i couldn't find any 🤦 🤦  thank you!

packages/progress-bar/src/ProgressBar.stories.tsx
charcoalyy: hmm. i imagine for maximum flexibility, yes - there is a world where we could also take in a list of fixed strings and loop them at arbitrary intervals, but i wonder if that might be better to hold off on in this PR?

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -104,10 +111,15 @@ export const LiveExample: StoryObj<typeof ProgressBar> = {
     label: <span>Label</span>,
     description: <span>Helper text</span>,
   },
-  render: SimulatedProgressBar,
+  render: initialArgs => (
```
shaneeza: Sorry, I'm just catching this now - the live example is the example used and interacted with on .design, so we should keep this story straightforward and easy to interact with. For example, with this story, consumers can change props, like `type`, `size`, `value`, etc., and the changes will be reflected in the story.  [Check out Button, for example](https://www.mongodb.design/component/button/live-example).  

You can remove the play function for this story since there won't be anything to test.

And also, lately I've been putting interaction tests in a separate file like so, `ProgressBar.interactions.stories.tsx`, to separate stories and generated stories from interaction tests. [`Drawer` is a good example of this](https://github.com/mongodb/leafygreen-ui/blob/main/packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout.interactions.stories.tsx).

So, any story in this file that uses the `play` function should be moved to the new file.

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -36,14 +39,41 @@ export function ProgressBar(props: ProgressBarProps) {
...
+    if (animationMode === AnimationMode.Indeterminate && !isIndeterminate) {
```
shaneeza: Something else I missed is the `usePrevious` hook, which you can use to store previous prop values. [Check out this example from `Code`](https://github.com/mongodb/leafygreen-ui/blob/main/packages/code/src/LanguageSwitcher/LanguageSwitcher.tsx#L20).  With this, you can store the value of `isIndeterminate` and check if the previous value is different from the new value. This way, you're checking the same prop.

Also, it's probably unlikely that someone will do this, but will this check cause any unexpected behavior if the `type` prop changes?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -36,14 +39,41 @@ export function ProgressBar(props: ProgressBarProps) {
...
+      setAnimationMode(AnimationMode.Transition);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
shaneeza: What are we ignoring here?

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
shaneeza, line 1: This isn't related to this PR, and this can be addressed in another PR. I noticed that this component doesn't extend HTML `div` attributes, meaning consumers can't pass any div attributes, such as `className`, to the parent div container. [Here's an example in Drawer where we extend `div`](https://github.com/mongodb/leafygreen-ui/blob/main/packages/drawer/src/Drawer/Drawer.types.ts#L17). If you add this, you might need to omit `children`, since this component doesn't accept children. You'll also need to [spread `...rest`](https://github.com/mongodb/leafygreen-ui/blob/main/packages/drawer/src/Drawer/Drawer.tsx#L133) inside `ProgressBar.tsx` and ensure that [the `className` prop is included in the styles](https://github.com/mongodb/leafygreen-ui/blob/main/packages/drawer/src/Drawer/Drawer.tsx#L123), as that's how consumers can add custom styles.

packages/progress-bar/src/ProgressBar.stories.tsx
shaneeza: Yeah, if we end up doing that, it can be a separate PR.

packages/progress-bar/src/ProgressBar.stories.tsx
shaneeza: Also, I would make sure that the story controls are correct. For example, I would make the `label` and `description` [controls accept text](https://github.com/mongodb/leafygreen-ui/blob/main/packages/drawer/src/Drawer.stories.tsx#L75).

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
charcoalyy: thank you so much for the links to examples!! am i correct to assume we should by default always do this with our components to accommodate consumer styling?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
charcoalyy: it's ignoring `animationMode` since it relies on previous value of `animationMode` but then resets it immediately after 😳 i will look into your comment about `usePrevious` to see if this clunkiness can be avoided!

packages/progress-bar/src/ProgressBar.stories.tsx
charcoalyy: ah, i see i see. that's super helpful context, thank you!!

charcoalyy: i like the idea of separating those with the play function/dynamic input versus the static ones - however, in this case, no real user "interaction" exists with the progress bar, which makes me hesitant about the file name. do you have any thoughts on this?

shaneeza: I see your point; the end user does not directly interact with the progress bar. However, I think that "interaction" can encompass any action or change that causes the component's state or appearance to update.

For example, in the indeterminate-to-determinate story, we simulate real-world behavior by changing the state, observing a timeout, and asserting that the determinate animation is shown and the indeterminate animation is gone. This interaction reflects how the end user will experience this component.

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
shaneeza: yess

packages/progress-bar/src/ProgressBar.stories.tsx
charcoalyy: awesome! that makes a lot of sense.

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }, [transitions]);
+
+  return <ProgressBar {...currentProps} />;
+};
+
+const meta: StoryMetaType<typeof ProgressBar> = {
+  title: 'Components/ProgressBar/Interactions',
+  component: ProgressBar,
+  parameters: {
+    default: 'LiveExample',
```
shaneeza: I think you can make this `null` since there is no `LiveExample`

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -57,22 +57,31 @@ export function ProgressBar(props: ProgressBarProps) {
   );
 
   useEffect(() => {
-    if (animationMode === AnimationMode.Indeterminate && !isIndeterminate) {
-      setAnimationMode(AnimationMode.Transition);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isIndeterminate]);
+    setAnimationMode(currentMode => {
```
shaneeza, line 61: I like this change! This way, you always have the most up-to-date animation mode.

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
```diff
@@ -0,0 +1,151 @@
...
+      ]}
+      {...initialArgs}
+    />
+  ),
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
+    const canvas = within(canvasElement);
+
+    const finalText = await waitFor(
+      () => canvas.getByText('Even newer helper text...!'),
+      // transition timeout with buffer time of 500ms
```
shaneeza: 500?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -86,16 +125,34 @@ export function ProgressBar(props: ProgressBarProps) {
...
+              }
+            }}
           ></div>
         </div>
       </div>
 
       {description && (
-        <Description darkMode={darkMode} disabled={disabled}>
+        <Description
+          darkMode={darkMode}
```
shaneeza, line 151: Just noticed this but if you wrap the component in the `LeafyGreenProvider`, then you don't have to pass `darkMode` to each child component. The component will read the value passed to the provider. Here's [an example from Callout](https://github.com/mongodb/leafygreen-ui/blob/main/packages/callout/src/Callout/Callout.tsx#L38).

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
charcoalyy: too long? or unclear where it's derived from?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
charcoalyy: just so happened to read that in the style guide now 🤭 thank you! let me address this in a separate clean-up PR rather than this one.

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
shaneeza: sorry, i should have been clearer. It says 500ms here, but the timeout below is 4000

charcoalyy: oh gotcha. yes - i took the length of the play/interaction and added 500ms. i should make that clearer with a variable. in any case, separate question: does it even make sense to have a "buffer"? if not, is using exact times appropriate? if so, is 500ms too long?

shaneeza: Does the expect still pass without a buffer?

charcoalyy: hmmmm. when i try without a buffer locally, it passes. but i was unsure if we're guaranteed a "perfect"(?) test run in every environment, i.e., how `waitFor` works internally.

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -109,7 +105,16 @@ export const getHeaderIconStyles = ({
   margin-bottom: ${spacingToken[50]}px; // align icon with text baseline
   color: ${disabled
     ? colorToken[theme].icon.disabled.default
-    : progressBarColorStyles[theme][color].icon};
+    : barColorStyles[theme][color].icon};
+`;
+
+export const getAnimatedTextStyles = (isNewDescription: boolean) => css`
+  ${isNewDescription &&
```
shaneeza: I won't block on this, but we tend to use `cx` for conditional classes. So, for example, instead of checking `isNewDescription &&` and generating conditional styles inside this style obj, we always define the styles in the obj and conditionally compose the styles with `cx`. This way it is clearer and explicit about what styles are applied.

Here's an example from [ContextDrawer](https://github.com/mongodb/leafygreen-ui/blob/main/packages/context-drawer/src/ContextDrawer/ContextDrawer.styles.ts#L73-L92)

So this would be:
```js
className={cx({[animatedTextStyle]: isNewDescription})}
```

shaneeza (approved):

PR #2941, opened 2025-07-02
Updates `Tooltip` tests

TheSonOfThomp:


packages/tooltip/package.json
```diff
@@ -28,6 +28,7 @@
     "@leafygreen-ui/popover": "workspace:^",
     "@leafygreen-ui/tokens": "workspace:^",
     "@leafygreen-ui/typography": "workspace:^",
+    "@storybook/test": "^8.6.14",
```
stephl3: can we move this to dev deps?

packages/tooltip/src/Tooltip.stories.tsx
```diff
@@ -64,134 +92,181 @@ const meta: StoryMetaType<typeof Tooltip> = {
...
+              {...args}
+              key={a + j}
+              darkMode={darkMode}
+              align={a}
+              justify={j}
+              triggerEvent="click"
+              initialOpen
+              trigger={<Button size={Size.XSmall} />}
+            >
+              {a}, {j}
```
stephl3: `align="right"` doesn't show any variance because the height of the tooltip is the same for the different `align="right"`+`justify` combos
Screenshot 2025-07-02 at 1 14 58 PM

This looks like a good opportunity to make them separate lines to better show that variance. Maybe we can do something like this?
```suggestion
              <Body>{a}</Body>
              <Body>{j}</Body>
```

packages/tooltip/src/Tooltip.stories.tsx
```diff
@@ -64,134 +92,181 @@ const meta: StoryMetaType<typeof Tooltip> = {
...
-  return (
+  render: ({ darkMode, ...args }: TooltipProps) => (
     <div
       className={css`
         padding: 100px;
       `}
     >
-      <Tooltip initialOpen {...args} />
+      <Tooltip
+        trigger={<Button size={Size.XSmall}>{args.triggerEvent} me!</Button>}
```
stephl3, line 111: very nice! 😎

stephl3: thoughts on including a patch changeset to update the example on .design?

TheSonOfThomp: LiveExample should be the one that's imported on .design

stephl3: Curious to confirm this non-blocking question: my current understanding is that this change won't reflect on .design because it will need to be published on a new version and bumped in .design's dep list. Otherwise, it will continue building the latest published version's live example which won't have this change. Is that right?

TheSonOfThomp: Correct, this change will only make it to .design once we bump to the latest version in that project

PR #2942, opened 2025-07-02
[LG-5298] feat(context-drawer): add test utils and refactor test file

stephl3:
## ✍️ Proposed changes

- add test utils for `ContextDrawer`
- refactor `ContextDrawer` test file
- add `Test Harnesses` section to README

🎟 _Jira ticket:_ [LG-5298](https://jira.mongodb.org/browse/LG-5298)

### For bug fixes, new features & breaking changes

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `pnpm changeset` and documented my changes

## 🧪 How to test changes

- unit tests

PR #2938, opened 2025-07-01
[LG-5336] feat(preview-card): add test utils and refactor test file

stephl3:
## ✍️ Proposed changes

- add test utils for `PreviewCard`
- refactor `PreviewCard` test file
- add `Test Harnesses` section to README

🎟 _Jira ticket:_ [LG-5336](https://jira.mongodb.org/browse/LG-5336)

### For bug fixes, new features & breaking changes

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `pnpm changeset` and documented my changes

## 🧪 How to test changes

- unit tests

PR #2943, opened 2025-07-02
LG-5349: progress bar getTestUtils and edge case testing

charcoalyy:
## ✍️ Proposed changes

- introduce test harnesses exposed to consumers
- separate internal utils into their own directory
- add specs for edge cases (truncation, unexpected numerical inputs)

🎟 _Jira ticket:_ [LG-5349](https://jira.mongodb.org/browse/LG-5349)

### For new components


packages/progress-bar/package.json
```diff
@@ -21,7 +21,8 @@
     "@leafygreen-ui/palette": "workspace:^",
     "@leafygreen-ui/tokens": "workspace:^",
     "@leafygreen-ui/typography": "workspace:^",
-    "@leafygreen-ui/icon": "workspace:^"
+    "@leafygreen-ui/icon": "workspace:^",
+    "@lg-tools/test-harnesses": "workspace:^"
```
stephl3: can also list this under `references` in your tsconfig.json

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -44,10 +49,21 @@ export function ProgressBar(props: ProgressBarProps) {
...
+      aria-disabled={disabled}
+      data-lgid={lgIds.root}
+    >
       <div className={headerStyles}>
-        <Label htmlFor={progressBarId} darkMode={darkMode} disabled={disabled}>
+        <Label
+          htmlFor={progressBarId}
+          darkMode={darkMode}
+          disabled={disabled}
+          className={truncatedTextStyles}
```
stephl3, line 117: left a couple comments with the chromatic snapshots but I think these styles are leading to premature text wrapping even when the text is not long

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -44,10 +49,21 @@ export function ProgressBar(props: ProgressBarProps) {
...
-    <div className={containerStyles} aria-disabled={disabled}>
+    <div
+      className={containerStyles}
+      aria-disabled={disabled}
+      data-lgid={lgIds.root}
+    >
       <div className={headerStyles}>
-        <Label htmlFor={progressBarId} darkMode={darkMode} disabled={disabled}>
+        <Label
+          htmlFor={progressBarId}
```
stephl3: I missed this in a previous review, but can you take another look at the `progressBarId`? If you look at the  `WithLongTextLabel` story in dev tools, you'll see the id has the full text in it, and I think there are better ways to approach that

1. We have a `useIdAllocator` hook from `@leafygreen-ui/hooks` that can be used to generate a unique id for the progress bar
2. `aria-label` should be a human readable text, while `aria-labelledby` can be an id of a node that contains the label text for a given element. In this implementation, you may want to specify a separate `id` for label (maybe `${progressBarId}-label`) and then use that for the `aria-labelledby` attribute instead of `aria-label`. Alternatively, you could use `aria-label={label}`, but it's possible that the value of `label` is actually a node, so you would need to account for that case

stephl3: ^could be handled in a separate PR as well if you want to avoid PR bloat

packages/progress-bar/src/utils/getTestUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+  };
+};
+
+interface GetTestUtilsReturnType<T extends HTMLDivElement = HTMLDivElement> {
+  queryLoaderElement: () => T | null;
+  queryMeterElement: () => T | null;
+  getBarFillElement: () => T | null;
+  getBarTrackElement: () => T | null;
+  getIconElement: () => T | null;
+}
```
stephl3: can we add tsdocs for these so consumers can easily decipher what these specifically do? also we'll usually set up a separate `getTestUtils.types.ts` file for these

packages/progress-bar/src/utils/getTestUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+    track: `${root}-track`,
+    fill: `${root}-fill`,
+  } as const;
+};
+
+export const getTestUtils = <T extends HTMLDivElement = HTMLDivElement>(
+  lgId: `lg-${string}` = DEFAULT_LGID_ROOT,
+): GetTestUtilsReturnType<T> => {
+  const lgIds = getLgIds(lgId);
+  const parent = getByLgId!<T>(lgIds.root);
```
stephl3: It's ideal to avoid this pattern of calling a `get*` util in this way because if the root element is not found, `getTestUtils` will throw an error. Instead, similar to how you do `getBarFillElement` and `getBarTrackElement` below, we can return a `getProgressBar` util. In the below utils, you can call `getProgressBar()` when you need the root element

packages/progress-bar/src/utils/getTestUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const queryLoaderElement = () =>
+    within(parent).queryByRole('progressbar') as T | null;
+  const queryMeterElement = () =>
+    within(parent).queryByRole('meter') as T | null;
```
stephl3: in this implementation, it's possible that these elements make more sense to be considered the `root` element. what do you think about that?

packages/progress-bar/src/utils/getTestUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const getBarFillElement = () => getByLgId!<T>(lgIds.fill) as T | null;
+  const getBarTrackElement = () => getByLgId!<T>(lgIds.track) as T | null;
+
+  const getIconElement = () => within(parent).queryByRole('img') as T | null;
```
stephl3: Anothe one I missed previously: if these icon elements are `role="img"` and are intended to provide semantic meaning, we probably should pass an `aria-label` or `aria-labelledby` to the svg

packages/progress-bar/src/utils/getTestUtils.ts
stephl3: also wondering if we want to include test utils for the label + description?

packages/progress-bar/package.json
stephl3: (Not the right line to leave this comment but wanted to specify in this file)

We should keep consumer test utils separate from internally used utils and consumer test utils should be housed in a `testing` subdir that is exported via the `exports` key in `package.json`. Here are a couple recent examples:
- https://github.com/mongodb/leafygreen-ui/pull/2942/files
- https://github.com/mongodb/leafygreen-ui/pull/2938/files

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
charcoalyy: yes- thank you! i flagged this for addressing in an (upcoming) PR dedicated specifically to aria attribute details and screen reader behaviour.

packages/progress-bar/src/utils/getTestUtils.ts
charcoalyy: these come from the leafy-green icons package, which i believe auto-applies an aria-label for each unique icon supported!

packages/progress-bar/src/utils/getTestUtils.ts
charcoalyy: this is a great point, and i originally had it implemented as so. however, some elements of the component exist outside of the actual bar itself, i.e., the icon exists next to, not within, the div with `role='progressbar'`. same with the label and description. thus here, the root now refers to the entire container. i do see how it makes less sense with the explanation, however. any thoughts on how it could be done clearer?

packages/progress-bar/src/utils/utils.spec.ts
charcoalyy, line 1: aside: these specs are not new, they were moved from the original ProgressBar.spec.tsx file

packages/progress-bar/src/utils/getTestUtils.ts
stephl3: hmm I'm not sure myself but maybe the id for the root node can be used for the progress bar container and then a separate id exists for the progress bar itself? could be worth asking about this in the dev channel

packages/progress-bar/src/utils/getTestUtils.ts
stephl3: TIL!

packages/progress-bar/package.json
charcoalyy: OHHH okay the brain is now braining 🧘🏻 internal utils still belong in our ProgressBar package, while external utils will be in this new `testing` subdir. would it be appropriate to have a sub-subdir `utils` inside subdir `ProgressBar`? with expansion of the specs, i figure it's cleaner to separate utility testing versus anything using the component directly.

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
charcoalyy: thank you for this catch! this was totally an oversight after trying to re-add some accidentally deleted styling 😅

packages/progress-bar/package.json
stephl3: Sounds great! For added context, the main reason we use the `/testing` subdir is so that consumers who use test utils in their tests don't pull testing dependencies into their production builds. In test files, they can do something like this:
```
// Component.spec.tsx
import { getTestUtils } from '@leafygreen-ui/progress-bar/testing';
...
```

charcoalyy: TIL!!

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -92,90 +97,78 @@ describe('packages/progress-bar', () => {
...
-      expect(resolvedProps).toEqual({
-        value: 50,
-        maxValue: 100,
-        disabled: false,
-        color: Color.Yellow,
-        isIndeterminate: false,
-        enableAnimation: false,
-      });
-    });
+      describe('with unexpected input', () => {
```
charcoalyy: git history is less intuitive here - the specs for `resolveProgressBarProps` and `getFormattedValue` are unchanged, simply moved to the utils subdirectory. new specs were added for edge cases in this main spec file.

packages/progress-bar/README.md
```diff
@@ -116,3 +116,73 @@ const totalSpaceAvailable = 128
 | `variant?`         | `'info'` \| `'success'` \| `'warning'` \| `'error'`                                            | `'info'`    | loader only                    | Variant for loader type. Animation is only available for `'info'` or `'success'` variants.                                                 |
 | `isIndeterminate?` | `boolean`                                                                                      | `false`     | loader only                    | When `true`, shows an infinite looping animation along the bar.                                                                            |
 | `enableAnimation?` | `boolean`                                                                                      | `false`     | determinate loader only        | When `true`, enables shimmer animation for long-running processes. Not available for meters or if `isIndeterminate` is `true` for loaders. |
+
+## Test Harnesses
+
+### `getTestUtils`
+
+`getTestUtils()` allows you to interact with the `ProgressBar` component in test suites. If the `ProgressBar` instance cannot be found, an error will be thrown.
```
stephl3: I know this is a copy and paste of the examples I sent over, but I'm now realizing that this 2nd sentence is inaccurate since we did away with that pattern 🤦 mind dropping this line before merge?

packages/progress-bar/README.md
```diff
@@ -116,3 +116,73 @@ const totalSpaceAvailable = 128
...
+#### Single `ProgressBar`
+
+```tsx
+import {
+  getTestUtils,
+  renderProgressBar,
+} from '@leafygreen-ui/progress-bar/testing';
+
+test('renders single progress bar', () => {
+  // ...code to render a single progress bar
```
stephl3:
```suggestion
  renderProgressBar();
```

packages/progress-bar/README.md
```diff
@@ -116,3 +116,73 @@ const totalSpaceAvailable = 128
...
+#### Multiple `ProgressBar` components
+
+```tsx
+import {
+  getTestUtils,
+  renderMultipleProgressBars,
+} from '@leafygreen-ui/progress-bar/testing';
+
+test('renders multiple progress bars', () => {
+  // ...code to render multiple progress bars
```
stephl3, line 146:
```suggestion
  renderMultipleProgressBars();
```

packages/progress-bar/src/testing/getTestUtils.types.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  /** Returns the label element, if present. */
+  getLabelElement: () => T | null;
+
+  /** Returns the description element, if present. */
+  getDescriptionElement: () => T | null;
+
+  /** Returns the value text element, if present. */
+  getValueTextElement: () => T | null;
+}
```
stephl3: Can we update these util names and types to match the patterns that testing library uses? https://testing-library.com/docs/queries/about/#types-of-queries

Since it's a commonly used library, certain verbs suggest certain types. It can be confusing/unexpected if our test utils don't follow the same pattern

Usually...
- `get*` means find the element or throw if no elements or multiple elements are found
- `query*` means find the element or return null and throw if multiple elements are found
- `find*` is used for promises which might be useful in this context if the loader or meter has an async reaction due to a user interaction

packages/progress-bar/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,56 @@
+import { getByLgId, queryByLgId } from '@lg-tools/test-harnesses';
+import { within } from '@testing-library/react';
+
+import { type GetTestUtilsReturnType } from './getTestUtils.types';
+
+export const DEFAULT_LGID_ROOT = 'lg-progress-bar';
```
stephl3:
```suggestion
export const DEFAULT_LGID_ROOT = 'lg-progress_bar';
```
https://github.com/mongodb/leafygreen-ui/blob/main/STYLEGUIDE.md#follow-bem-ish-patterns-when-hard-coding-a-data-testid-or-data-lgid

packages/progress-bar/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const getLgIds = (root: `lg-${string}` = DEFAULT_LGID_ROOT) => {
+  return {
+    root,
+    track: `${root}-track`,
+    fill: `${root}-fill`,
+    icon: `${root}-icon`,
+    label: `${root}-label`,
+    description: `${root}-description`,
+    valueText: `${root}-value-text`,
```
stephl3:
```suggestion
    valueText: `${root}-value_text`,
```
https://github.com/mongodb/leafygreen-ui/blob/main/STYLEGUIDE.md#follow-bem-ish-patterns-when-hard-coding-a-data-testid-or-data-lgid

packages/progress-bar/README.md
charcoalyy: is there a reason this is preferred? should i implement a test function `renderMultipleProgressBars` - i didn't initially because i was on the fence about it being used at all.

stephl3: oh gotcha, I only suggested this change because I saw it in the import declaration! If it's not actually implemented, then probably just remove from the import

charcoalyy: OOPS. great catch, thank you, i hauled that over from the other component examples! 😅

packages/progress-bar/README.md
```diff
@@ -116,3 +116,68 @@ const totalSpaceAvailable = 128
...
+  expect(getBar()).toBeInTheDocument();
+});
+```
+
+#### Multiple `ProgressBar` components
+
+```tsx
+import {
+  getTestUtils,
+  renderMultipleProgressBars,
```
stephl3: remove?

packages/progress-bar/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+): GetTestUtilsReturnType<T> => {
+  const lgIds = getLgIds(lgId);
+
+  const getContainer = () => getByLgId!<T>(lgIds.root);
+
+  const getBar = () => {
+    const container = getContainer();
+    return (within(container).queryByRole('progressbar') ??
+      within(container).queryByRole('meter')) as T;
+  };
```
stephl3, line 30: can we not use `getByRole` here? if not, can we rename this util `queryBar`?

charcoalyy: the bar is guaranteed (either progressbar or meter) but internally individually neither is guaranteed. i was thinking `getByRole` would throw an error at not finding one or the other.

stephl3: @charcoalyy I wonder if instead this is better to split into 2 separate queries of `getLoader` and `getMeter`. Thoughts on that?

charcoalyy: hmm. bit of a longer explanation, but there's another refactor in prop naming/structure that i need to run through after this merge to better align with design's mental model. one of those changes involves the loader/meter distinguisher no longer being the "top" level fork, if that makes sense. so testing-wise (as in getting the element), referring to the bar as one entity will better reflect design.

PR #2947, opened 2025-07-08
LG-5318: progress bar hidden live region and aria attributes

charcoalyy:
## ✍️ Proposed changes

- hook for live screen reader announcements (only initially, once past halfway, and once at completion)*
- utils for extracting aria attributes and necessary ids

*Note that the progress value is _always_ available to users via virtual cursor due to the aria-value attributes. Live announcements (which are unprompted and automatic no matter what content the user is on) are intended to be minimally disturbing, and therefore only at significant changes.

🎟 _Jira ticket:_ [LG-5318](https://jira.mongodb.org/browse/LG-5318)

### For new components

### DIFF
--- a/packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
+++ b/packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
@@ -8,6 +8,10 @@ import {
   resolveProgressBarProps,
 } from './ProgressBar.utils';
 
+const getAnnouncementMessage = (value: number, maxValue: number): string => {
+  return `Update: current progress is ${value}% (${value} out of ${maxValue}).`;
+};
+
 describe('packages/progress-bar', () => {
   describe('basic rendering', () => {
     test('renders with a label', () => {
@@ -113,8 +117,8 @@ describe('packages/progress-bar', () => {
       expect(screen.queryByRole('status')).toBeNull();
     });
 
-    test('does not announce if progress under 50%', () => {
-      render(
+    test('announces initial progress, but not any further changes if next threshold is not met', () => {
+      const { rerender } = render(
         <ProgressBar
           type={Type.Loader}
           isIndeterminate={false}
@@ -122,11 +126,37 @@ describe('packages/progress-bar', () => {
           maxValue={TEST_MAX_VALUE}
         />,
       );
-      expect(screen.queryByRole('status')).toHaveTextContent('');
+      expect(screen.queryByRole('status')).toHaveTextContent(
+        getAnnouncementMessage(TEST_VALUE_UNDER_50, TEST_MAX_VALUE),
+      );
+
+      rerender(
+        <ProgressBar
+          type={Type.Loader}
+          isIndeterminate={false}
+          value={TEST_VALUE_UNDER_50 + 1}
+          maxValue={TEST_MAX_VALUE}
+        />,
+      );
+      expect(screen.queryByRole('status')).toHaveTextContent(
+        getAnnouncementMessage(TEST_VALUE_UNDER_50, TEST_MAX_VALUE),
+      );
     });
 
     test('announces if 50% threshold passed', () => {
-      render(
+      const { rerender } = render(
+        <ProgressBar
+          type={Type.Loader}
+          isIndeterminate={false}
+          value={TEST_VALUE_UNDER_50}
+          maxValue={TEST_MAX_VALUE}
+        />,
+      );
+      expect(screen.queryByRole('status')).toHaveTextContent(
+        getAnnouncementMessage(TEST_VALUE_UNDER_50, TEST_MAX_VALUE),
+      );
+
+      rerender(
         <ProgressBar
           type={Type.Loader}
           isIndeterminate={false}
@@ -135,12 +165,24 @@ describe('packages/progress-bar', () => {
         />,
       );
       expect(screen.queryByRole('status')).toHaveTextContent(
-        `Update: current progress is ${TEST_VALUE_OVER_50}% (${TEST_VALUE_OVER_50} out of 100).`,
+        getAnnouncementMessage(TEST_VALUE_OVER_50, TEST_MAX_VALUE),
       );
     });
 
     test('announces if 100% threshold passed', () => {
-      render(
+      const { rerender } = render(
+        <ProgressBar
+          type={Type.Loader}
+          isIndeterminate={false}
+          value={TEST_VALUE_UNDER_50}
+          maxValue={TEST_MAX_VALUE}
+        />,
+      );
+      expect(screen.queryByRole('status')).toHaveTextContent(
+        getAnnouncementMessage(TEST_VALUE_UNDER_50, TEST_MAX_VALUE),
+      );
+
+      rerender(
         <ProgressBar
           type={Type.Loader}
           isIndeterminate={false}
@@ -149,11 +191,11 @@ describe('packages/progress-bar', () => {
         />,
       );
       expect(screen.queryByRole('status')).toHaveTextContent(
-        'Update: current progress is 100% (100 out of 100).',
+        getAnnouncementMessage(TEST_MAX_VALUE, TEST_MAX_VALUE),
       );
     });
 
-    test('additionally announces color if yellow or red', () => {
+    test('additionally announces variant if yellow or red', () => {
       render(
         <ProgressBar
           type={Type.Loader}
@@ -164,7 +206,10 @@ describe('packages/progress-bar', () => {
         />,
       );
       expect(screen.queryByRole('status')).toHaveTextContent(
-        `Update: current progress is ${TEST_VALUE_OVER_50}% (${TEST_VALUE_OVER_50} out of 100). Status is yellow.`,
+        `${getAnnouncementMessage(
+          TEST_VALUE_OVER_50,
+          TEST_MAX_VALUE,
+        )} Status is ${Color.Yellow}.`,
       );
     });
   });
--- a/packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
+++ b/packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
@@ -135,7 +135,7 @@ describe('packages/progress-bar', () => {
         />,
       );
       expect(screen.queryByRole('status')).toHaveTextContent(
-        `Current progress is ${TEST_VALUE_OVER_50}% (${TEST_VALUE_OVER_50} out of 100).`,
+        `Update: current progress is ${TEST_VALUE_OVER_50}% (${TEST_VALUE_OVER_50} out of 100).`,
       );
     });
 
@@ -149,7 +149,7 @@ describe('packages/progress-bar', () => {
         />,
       );
       expect(screen.queryByRole('status')).toHaveTextContent(
-        'Current progress is 100% (100 out of 100).',
+        'Update: current progress is 100% (100 out of 100).',
       );
     });
 
@@ -164,8 +164,72 @@ describe('packages/progress-bar', () => {
         />,
       );
       expect(screen.queryByRole('status')).toHaveTextContent(
-        `Current progress is ${TEST_VALUE_OVER_50}% (${TEST_VALUE_OVER_50} out of 100). Status is yellow.`,
+        `Update: current progress is ${TEST_VALUE_OVER_50}% (${TEST_VALUE_OVER_50} out of 100). Status is yellow.`,
       );
+    });
+  });
+
+  describe('aria attributes', () => {
+    test('renders with aria-labelledby and aria-describedby', () => {
+      const LABEL_TEXT = 'Test Progress Bar';
+      const DESCRIPTION_TEXT = 'Test description for the progress bar.';
+
+      render(
+        <ProgressBar
+          type={Type.Loader}
+          isIndeterminate={true}
+          label={LABEL_TEXT}
+          description={DESCRIPTION_TEXT}
+        />,
+      );
+
+      const progressBar = screen.getByRole('progressbar');
+      expect(progressBar).toHaveAttribute('aria-labelledby');
+      expect(progressBar).toHaveAttribute('aria-describedby');
+    });
+
+    test('renders with aria-label when no label is provided', () => {
+      render(<ProgressBar type={Type.Loader} isIndeterminate={true} />);
+
+      const progressBar = screen.getByRole('progressbar');
+      expect(progressBar).toHaveAttribute('aria-label', 'progressbar');
+    });
+
+    test('renders with aria-value attributes if value and maxValue are provided', () => {
+      const TEST_VALUE = 75;
+      const TEST_MAX_VALUE = 100;
+
+      render(
+        <ProgressBar
+          type={Type.Loader}
+          isIndeterminate={false}
+          value={TEST_VALUE}
+          maxValue={TEST_MAX_VALUE}
+        />,
+      );
+
+      const progressBar = screen.getByRole('progressbar');
+      expect(progressBar).toHaveAttribute('aria-valuemin', '0');
+      expect(progressBar).toHaveAttribute(
+        'aria-valuemax',
+        String(TEST_MAX_VALUE),
+      );
+      expect(progressBar).toHaveAttribute('aria-valuenow', String(TEST_VALUE));
+    });
+
+    test('renders with aria-valuetext if maxValue is not provided', () => {
+      const TEST_VALUE = 50;
+
+      render(
+        <ProgressBar
+          type={Type.Loader}
+          isIndeterminate={true}
+          value={TEST_VALUE}
+        />,
+      );
+
+      const progressBar = screen.getByRole('progressbar');
+      expect(progressBar).toHaveAttribute('aria-valuetext', String(TEST_VALUE));
     });
   });
 
--- a/packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
+++ b/packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
@@ -97,6 +97,78 @@ describe('packages/progress-bar', () => {
     });
   });
 
+  describe('screen reader behavior', () => {
+    const TEST_VALUE_UNDER_50 = 43;
+    const TEST_VALUE_OVER_50 = 57;
+    const TEST_MAX_VALUE = 100;
+
+    test('does not have a live region for meter types', () => {
+      render(
+        <ProgressBar
+          type={Type.Meter}
+          value={TEST_VALUE_OVER_50}
+          maxValue={TEST_MAX_VALUE}
+        />,
+      );
+      expect(screen.queryByRole('status')).toBeNull();
+    });
+
+    test('does not announce if progress under 50%', () => {
+      render(
+        <ProgressBar
+          type={Type.Loader}
+          isIndeterminate={false}
+          value={TEST_VALUE_UNDER_50}
+          maxValue={TEST_MAX_VALUE}
+        />,
+      );
+      expect(screen.queryByRole('status')).toHaveTextContent('');
+    });
+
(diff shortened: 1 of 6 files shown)


packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
```diff
@@ -0,0 +1,63 @@
+import { useEffect, useRef, useState } from 'react';
+
+import { Color, Type } from '../ProgressBar.types';
+import { getPercentage } from '../ProgressBar.utils';
+
+const announcementThresholds = [0, 50, 100];
+const variantsAnnounced = [Color.Yellow, Color.Red] as Array<Color>;
```
charcoalyy: NOTE: a refactor coming afterwards will unify variants. what's announced will NOT be colors, but actual semantic meanings (i.e., "warning" or "error"). Please regard this as placeholder for now.

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
```diff
@@ -0,0 +1,63 @@
+import { useEffect, useRef, useState } from 'react';
+
+import { Color, Type } from '../ProgressBar.types';
+import { getPercentage } from '../ProgressBar.utils';
+
+const announcementThresholds = [0, 50, 100];
```
charcoalyy, line 8: NOTE: the progress value is _always_ available to users via virtual cursor due to the `aria-value` attributes. Live announcements (which are unprompted and automatic no matter what content the user is on) are intended to be minimally disturbing, and therefore only at significant changes.

TheSonOfThomp: I wonder if we should have a way to disable this? I can imagine a "disk usage" meter suddenly hitting 50% and announcing unnecessarily

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
TheSonOfThomp: can you add an inline `//TODO:` for this?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -36,18 +38,28 @@ export function ProgressBar(props: ProgressBarProps) {
 
   const { theme } = useDarkMode(darkMode);
 
+  const { role, barId, labelId, descId } = getDivAttributes(type, label);
```
TheSonOfThomp: I'd prob. recommend changing the name `getDivAttributes` to something a little more descriptive

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
```diff
@@ -0,0 +1,63 @@
...
+  color,
+}: UseScreenReaderAnnouncerProps): string {
+  const thresholdIndexRef = useRef(-1);
+  const [message, setMessage] = useState('');
+
+  useEffect(() => {
+    // no live region messages for non-loader types or if value is undefined
+    if (type !== Type.Loader || value == undefined) {
+      thresholdIndexRef.current = -1;
+      setMessage('');
```
TheSonOfThomp: should this be `null` or `undefined`? In `ProgressBar.tsx` we check for existence of `screenReaderMessage` before rendering the live region, and an empty string is still truthy

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
```diff
@@ -0,0 +1,63 @@
...
+export function useScreenReaderAnnouncer({
+  type,
+  value,
+  maxValue,
+  color,
+}: UseScreenReaderAnnouncerProps): string {
+  const thresholdIndexRef = useRef(-1);
+  const [message, setMessage] = useState('');
+
+  useEffect(() => {
```
TheSonOfThomp: generally if you're calling `useEffect` to then set a state, that's a good indicator you should consider using `useMemo`. See: [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect)

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
charcoalyy: yes! so by design i moved to remove live announcements for meter types entirely, by default. what do you think of this?

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const percentage = maxValue ? getPercentage(value, maxValue) : value * 100;
+
+    // determinate largest threshold passed by current percentage
+    let newThresholdIndex = -1;
+
+    for (let i = 0; i < announcementThresholds.length; i++) {
+      if (percentage >= announcementThresholds[i]) {
+        newThresholdIndex = i;
+      }
+    }
```
TheSonOfThomp, line 42:
```
const newThresholdIndex = announcementThresholds.findIndex((threshold) => percentage >= threshold)
```

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+    if (newThresholdIndex === thresholdIndexRef.current) return;
+
+    // if new threshold was passed, update live region message
+    thresholdIndexRef.current = newThresholdIndex;
```
TheSonOfThomp, line 49: technically this would still announce if the progress went backwards for some reason

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -56,7 +68,8 @@ export function ProgressBar(props: ProgressBarProps) {
             className={getHeaderValueStyles({ theme, disabled })}
             darkMode={darkMode}
           >
-            {value != null && getFormattedValue(value, maxValue, formatValue)}
+            {value != undefined &&
```
TheSonOfThomp: just to catch both null & undefined, use `isDefined` from lib

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -72,8 +85,10 @@ export function ProgressBar(props: ProgressBarProps) {
 
       <div
         role={role}
-        id={progressBarId}
-        aria-label={progressBarId}
+        id={barId}
+        aria-labelledby={label ? labelId : undefined}
```
TheSonOfThomp: I think we should return `labelId` as possibly undefined from `getDivAttributes`. Then we can just write `aria-labelledby={labelId}`

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -72,8 +85,10 @@ export function ProgressBar(props: ProgressBarProps) {
...
+        aria-label={!label ? role : undefined}
```
TheSonOfThomp: We should probably require an `aria-label` prop if a label node isn't passed in.
See [`AriaLabelProps`](/packages/a11y/src/AriaLabelProps.ts)

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -95,10 +110,21 @@ export function ProgressBar(props: ProgressBarProps) {
...
           {description}
         </Description>
       )}
+
+      {screenReaderMessage && (
+        <div
+          role="status"
+          aria-live="polite"
+          aria-atomic="true"
+          className={getInvisibleStyles()}
```
TheSonOfThomp: I believe there should be some aria connection between this live region and the main progress bar. `aria-controls`/`aria-controlledby` maybe?

packages/progress-bar/src/ProgressBar/ProgressBar.utils.tsx
```diff
@@ -112,14 +113,35 @@ export const getFormattedValue = (
   }
 };
 
-export const getValueAriaAttributes = (value?: number, maxValue?: number) => {
+export const getDivAttributes = (type: Type, label?: React.ReactNode) => {
```
TheSonOfThomp: for this (and all other utils) can you add a quick TSDoc comment about what the util does?

packages/progress-bar/src/ProgressBar/ProgressBar.utils.tsx
```diff
@@ -112,14 +113,35 @@ export const getFormattedValue = (
...
+  const role = type === Type.Meter ? 'meter' : 'progressbar';
```
TheSonOfThomp: Also, can we add return type annotations to the utils that return an object?

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -8,6 +8,10 @@ import {
   resolveProgressBarProps,
 } from './ProgressBar.utils';
 
+const getAnnouncementMessage = (value: number, maxValue: number): string => {
+  return `Update: current progress is ${value}% (${value} out of ${maxValue}).`;
+};
```
TheSonOfThomp: nit: I'd recommend defining this within the `describe` block it's used in

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -97,6 +101,183 @@ describe('packages/progress-bar', () => {
...
+          value={TEST_VALUE_UNDER_50 + 1}
+          maxValue={TEST_MAX_VALUE}
+        />,
+      );
+      expect(screen.queryByRole('status')).toHaveTextContent(
+        getAnnouncementMessage(TEST_VALUE_UNDER_50, TEST_MAX_VALUE),
+      );
+    });
+
+    test('announces if 50% threshold passed', () => {
```
TheSonOfThomp: we're technically not testing "announcement" here, just that the text of the `status` changes

TheSonOfThomp: pretty sure we can't accurately test VO in jsdom

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
TheSonOfThomp: Actually, how difficult would it be to add this to this PR. I noticed some tests that test for the color, and I'd rather not commit those tests if possible!

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -180,3 +180,12 @@ export const getBarFillStyles = ({
...
   );
 };
+
+export const getInvisibleStyles = () => css`
+  position: absolute;
+  width: 0;
+  height: 0;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
```
TheSonOfThomp, line 279: can prob. replace these 3 lines with a `visibility: hidden;`

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
charcoalyy: ooh i really like that. will make the change

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
charcoalyy: this is true, but isn't that out of our testing scope? i.e., the stipulate is that if we change the text in a live region, VO announces it automatically. we test that the text changes properly, and trust that VO works correctly? perhaps an additional assertion for the aria-live attribute makes sense here?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: ooh. i was thinking if `visibility: hidden` would still affect the layout of elements around it, but with width and height being set, i suppose you're right.

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
charcoalyy: light bulb moment! does that look a little incongruous with / should the same logic apply for `descId`?

packages/progress-bar/src/ProgressBar/ProgressBar.utils.tsx
charcoalyy: yes!! thank you. i'll put that into a separate PR for less bloat here.

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
charcoalyy: i fear it's pretty big since it's part of that whole mental model shift. if it's any reassurance, it's just a variable name change while the logic persists 😅

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
charcoalyy: thank you! TIL the actual use case for useMemo! after the long journey of premature optimization, then never memoizing.

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
charcoalyy: hm. i thought an empty string is still considered falsy (verified via tests)?

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
charcoalyy: BEAUTIFUL. thank you. was trying to use `findLastIndex` but versioning got in the way.

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
charcoalyy: hmm. could compare with `<=` but is that something we want to avoid per se? or trust that use cases of loading in consumers will not go backwards?

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
charcoalyy: i take it back. `findIndex` does not work like `findLastIndex`, shockingly enough.

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: update: i believe screen readers will ignore an element with `visibility: hidden` (https://ionic.io/docs/accessibility/visibility)

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
TheSonOfThomp: I think just update the test names to something like "updates live area with text..."

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
TheSonOfThomp: Probably, yeah. Then you can just pass `aria-describedby={ descId }`, and it'll be undefined when relevant

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
```diff
@@ -41,7 +41,7 @@ export function useScreenReaderAnnouncer({
       }
     }
 
-    if (newThresholdIndex === thresholdIndexRef.current) return;
+    if (newThresholdIndex === thresholdIndexRef.current) return '';
```
TheSonOfThomp: style nit: even when there's only one line after the `if`, we tend to prefer wrapping it in a block `{ return '' }`

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -118,6 +124,7 @@ export function ProgressBar(props: ProgressBarProps) {
       {screenReaderMessage && (
```
TheSonOfThomp, line 182: I believe this will still render even if message is `''` since that's still a truthy value

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
TheSonOfThomp: Ah forgot that bit about `visibility`

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
charcoalyy: https://www.nfriedly.com/techblog/2009/07/advanced-javascript-operators-and-truthy-falsy/ ❓ although if it's cleaner to return null/undefined, i will make that change regardless.

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -106,24 +107,33 @@ describe('packages/progress-bar', () => {
...
           value={TEST_VALUE_OVER_50}
           maxValue={TEST_MAX_VALUE}
+          aria-label="required label"
         />,
       );
       expect(screen.queryByRole('status')).toBeNull();
     });
 
-    test('announces initial progress, but not any further changes if next threshold is not met', () => {
+    test('updates live region text for initial value, but not for any further changes if next threshold is not met', () => {
```
TheSonOfThomp, line 129: a little verbose, but that's fine 😆

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
TheSonOfThomp: If we wanted to announce going backwards, we'd probably want to update the announcement text. Otherwise it's unclear which direction it's going

I don't want to hold up this PR because of this, so let's just make a note of this/create a ticket to resolve in the future

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
TheSonOfThomp: The signatures of [findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) and [findLastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex) should be the same, no?

charcoalyy: ah- i meant this use case requires the last index, in which case `findIndex` would not work. it's the last index that the progress value exceeds, rather than the first. 

alternatively, i could order the array backwards (decreasing)? but that feels... incorrect.

packages/progress-bar/src/ProgressBar/hooks/useScreenReaderAnnouncer.tsx
charcoalyy: let me add a note in the appendix

PR #2952, opened 2025-07-11
LG-5350: refactor API design to match design mental model

charcoalyy:
## ✍️ Proposed changes

- moves the distinction between "progressbar" and "meter" roles to second-level fork (i.e., separate indeterminate/determinate, then progressbar/meter).
- unifies variant names for both "progressbar" and "meter" roles
- adds tsdocs for all utils

🎟 _Jira ticket:_ [LG-5350](https://jira.mongodb.org/browse/LG-5350)

### For new components

### DIFF
--- a/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.interactions.stories.tsx
@@ -3,7 +3,7 @@ import { StoryMetaType } from '@lg-tools/storybook-utils';
 import { StoryObj } from '@storybook/react';
 import { expect, waitFor, within } from '@storybook/test';
 
-import { ProgressBar, ProgressBarProps, Type } from '.';
+import { ProgressBar, ProgressBarProps } from '.';
 
 const testValues = {
   value: 53,
@@ -44,7 +44,6 @@ export default meta;
 
 export const WithChangingValue: StoryObj<typeof ProgressBar> = {
   args: {
-    type: Type.Loader,
     value: testValues.value,
     maxValue: testValues.maxValue,
     formatValue: 'fraction',
@@ -112,7 +111,6 @@ export const WithChangingDescriptions: StoryObj<typeof ProgressBar> = {
 
 export const IndeterminateToDeterminate: StoryObj<typeof ProgressBar> = {
   args: {
-    type: Type.Loader,
     isIndeterminate: true,
   },
   render: initialArgs => (
@@ -121,7 +119,6 @@ export const IndeterminateToDeterminate: StoryObj<typeof ProgressBar> = {
         [
           3500,
           {
-            type: Type.Loader,
             isIndeterminate: false,
             value: testValues.maxValue,
             maxValue: testValues.maxValue,
--- a/packages/progress-bar/src/ProgressBar.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.stories.tsx
@@ -94,8 +94,8 @@ const meta: StoryMetaType<typeof ProgressBar> = {
     },
     enableAnimation: {
       description:
-        '**Only available** if roleType="progressbar" and isIndeterminate=true (fixed to false in this example).',
-      control: { type: 'none' },
+        '**Only available** if roleType="progressbar" and isIndeterminate=true.',
+      control: { type: 'boolean' },
       // if: { arg: 'roleType', eq: Role.Progress } AND { arg: 'isIndeterminate', neq: true }, // TODO: not supported
     },
   },
--- a/packages/progress-bar/src/ProgressBar.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.stories.tsx
@@ -3,14 +3,13 @@ import { StoryMetaType } from '@lg-tools/storybook-utils';
 import { StoryObj } from '@storybook/react';
 
 import {
-  AnimatedLoaderVariant,
+  AnimatedVariant,
   FormatValueType,
-  LoaderVariant,
-  MeterStatus,
   ProgressBar,
   ProgressBarProps,
+  Role,
   Size,
-  Type,
+  Variant,
 } from '.';
 
 const testValues = {
@@ -19,18 +18,16 @@ const testValues = {
 };
 
 const defaultArgs: ProgressBarProps = {
-  type: Type.Loader,
   value: testValues.value,
   maxValue: testValues.maxValue,
   'aria-label': 'required label',
 };
 
-const TYPES = Object.values(Type);
+const ROLES = Object.values(Role);
 const SIZES = Object.values(Size);
 const FORMAT_VALUE_TYPES = Object.values(FormatValueType);
-const LOADER_VARIANTS = Object.values(LoaderVariant);
-const ANIMATED_LOADER_VARIANTS = Object.values(AnimatedLoaderVariant);
-const METER_STATUSES = Object.values(MeterStatus);
+const VARIANTS = Object.values(Variant);
+const ANIMATED_VARIANTS = Object.values(AnimatedVariant);
 
 const meta: StoryMetaType<typeof ProgressBar> = {
   title: 'Components/ProgressBar',
@@ -39,9 +36,9 @@ const meta: StoryMetaType<typeof ProgressBar> = {
     default: 'LiveExample',
     generate: {
       storyNames: [
-        'DeterminateVariants',
         'IndeterminateVariants',
-        'MeterVariants',
+        'DeterminateProgressVariants',
+        'DeterminateMeterVariants',
       ],
       args: {
         label: <span key="label">Label</span>,
@@ -62,9 +59,8 @@ const meta: StoryMetaType<typeof ProgressBar> = {
     },
   },
   argTypes: {
-    type: {
-      control: { type: 'select' },
-      options: TYPES,
+    isIndeterminate: {
+      control: { type: 'boolean' },
     },
     size: {
       control: { type: 'select' },
@@ -83,31 +79,24 @@ const meta: StoryMetaType<typeof ProgressBar> = {
         'custom example': (value: number) => `Currently at ${value} GB`,
       },
     },
-    isIndeterminate: {
-      control: { type: 'boolean' },
-      if: { arg: 'type', eq: Type.Loader },
-    },
-    maxValue: {
-      description:
-        '**Not available** if both type=loader and isIndeterminate=true (fixed to 200 in this example).',
-      control: { type: 'none' },
-      // if: { arg: 'type', eq: Type.Meter } OR { arg: 'isIndeterminate', eq: false }, // TODO: not supported
-    },
     variant: {
       control: { type: 'select' },
-      options: [...LOADER_VARIANTS, undefined],
-      if: { arg: 'type', eq: Type.Loader },
+      options: [...VARIANTS, undefined],
     },
-    status: {
+    maxValue: {
+      control: { type: 'number' },
+      if: { arg: 'isIndeterminate', eq: false },
+    },
+    roleType: {
       control: { type: 'select' },
-      options: [...METER_STATUSES, undefined],
-      if: { arg: 'type', eq: Type.Meter },
+      options: ROLES,
+      if: { arg: 'isIndeterminate', eq: false },
     },
     enableAnimation: {
       description:
-        '**Not available** if either type="meter" or isIndeterminate=true (fixed to false in this example).',
+        '**Only available** if roleType="progressbar" and isIndeterminate=true (fixed to false in this example).',
       control: { type: 'none' },
-      // if: { arg: 'type', neq: Type.Meter } AND { arg: 'isIndeterminate', neq: true }, // TODO: not supported
+      // if: { arg: 'roleType', eq: Role.Progress } AND { arg: 'isIndeterminate', eq: false }, // TODO: not supported
     },
   },
   decorators: [
@@ -165,46 +154,13 @@ export const WithHeaderTruncation: StoryObj<typeof ProgressBar> = {
   },
 };
 
-export const DeterminateVariants: StoryObj<typeof ProgressBar> = {
-  parameters: {
-    generate: {
-      combineArgs: {
-        variant: LOADER_VARIANTS,
-        enableAnimation: [false, true],
-        disabled: [false, true],
-      },
-      args: {
-        ...defaultArgs,
-        isIndeterminate: false,
-        formatValue: (value: number, maxValue?: number) =>
-          `${value} / ${maxValue} GB`,
-        showIcon: true,
-      },
-      excludeCombinations: [
-        {
-          variant: LOADER_VARIANTS.filter(
-            v =>
-              !(ANIMATED_LOADER_VARIANTS as Array<LoaderVariant>).includes(v),
-          ),
-          enableAnimation: [true],
-        },
-        {
-          variant: LOADER_VARIANTS.filter(v => v !== LoaderVariant.Info),
-          disabled: [true],
-        },
-      ],
-    },
-  },
-};
-
 export const IndeterminateVariants: StoryObj<typeof ProgressBar> = {
   parameters: {
     generate: {
       combineArgs: {
-        variant: ANIMATED_LOADER_VARIANTS,
+        variant: ANIMATED_VARIANTS,
       },
       args: {
-        type: Type.Loader,
         isIndeterminate: true,
         value: testValues.value,
         formatValue: (value: number) => `${value} MBs`,
@@ -214,24 +170,53 @@ export const IndeterminateVariants: StoryObj<typeof ProgressBar> = {
   },
 };
 
-export const MeterVariants: StoryObj<typeof ProgressBar> = {
+export const DeterminateProgressVariants: StoryObj<typeof ProgressBar> = {
   parameters: {
     generate: {
       combineArgs: {
-        status: [undefined, ...METER_STATUSES],
+        variant: VARIANTS,
+        enableAnimation: [false, true],
         disabled: [false, true],
       },
       args: {
-        type: Type.Meter,
-        value: testValues.value,
-        maxValue: testValues.maxValue,
+        ...defaultArgs,
         formatValue: (value: number, maxValue?: number) =>
           `${value} / ${maxValue} GB`,
         showIcon: true,
       },
       excludeCombinations: [
         {
-          status: METER_STATUSES,
+          variant: VARIANTS.filter(
+            v => !(ANIMATED_VARIANTS as Array<Variant>).includes(v),
+          ),
+          enableAnimation: [true],
+        },
+        {
+          variant: VARIANTS.filter(v => v !== Variant.Info),
+          disabled: [true],
+        },
+      ],
+    },
+  },
+};
+
+export const DeterminateMeterVariants: StoryObj<typeof ProgressBar> = {
+  parameters: {
+    generate: {
+      combineArgs: {
+        variant: [undefined, ...VARIANTS],
+        disabled: [false, true],
+      },
+      args: {
+        ...defaultArgs,
+        roleType: Role.Meter,
+        formatValue: (value: number, maxValue?: number) =>
+          `${value} / ${maxValue} GB`,
+        showIcon: true,
+      },
+      excludeCombinations: [
+        {
+          variant: VARIANTS,
           disabled: [true],
         },
       ],
--- a/packages/progress-bar/src/ProgressBar.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.stories.tsx
@@ -85,25 +85,20 @@ const meta: StoryMetaType<typeof ProgressBar> = {
     },
     maxValue: {
       control: { type: 'number' },
-      if: { arg: 'isIndeterminate', eq: false },
+      if: { arg: 'isIndeterminate', neq: true },
     },
     roleType: {
       control: { type: 'select' },
       options: ROLES,
-      if: { arg: 'isIndeterminate', eq: false },
+      if: { arg: 'isIndeterminate', neq: true },
     },
     enableAnimation: {
       description:
         '**Only available** if roleType="progressbar" and isIndeterminate=true (fixed to false in this example).',
       control: { type: 'none' },
-      // if: { arg: 'roleType', eq: Role.Progress } AND { arg: 'isIndeterminate', eq: false }, // TODO: not supported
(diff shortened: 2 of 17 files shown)


packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -21,40 +15,30 @@ export const FormatValueType = {
...
-  Red: 'red',
+export type AnimatedVariant =
+  (typeof AnimatedVariant)[keyof typeof AnimatedVariant];
+
+export const Role = {
+  Meter: 'meter',
+  Progress: 'progressbar',
 } as const;
-export type Color = (typeof Color)[keyof typeof Color];
+export type Role = (typeof Role)[keyof typeof Role];
```
TheSonOfThomp, line 49: can you add TSDoc that notes that the values of `Role` match the expected `aria-role` values

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -21,40 +15,30 @@ export const FormatValueType = {
...
-export type AnimatedLoaderVariant =
-  (typeof AnimatedLoaderVariant)[keyof typeof AnimatedLoaderVariant];
-
-export const Color = {
-  Blue: 'blue',
-  Green: 'green',
-  Yellow: 'yellow',
-  Red: 'red',
+export type AnimatedVariant =
+  (typeof AnimatedVariant)[keyof typeof AnimatedVariant];
```
TheSonOfThomp: can you add docs outlining why we have both `variant` and `animatedVariant`?

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -65,110 +49,107 @@ type BaseProps = DarkModeProps &
...
-  status?: MeterStatus;
+  /** Optional variant of the progress bar. Defaults to "info". */
+  variant?: AnimatedVariant;
 }
 
-export type ProgressBarProps = BaseProps & (MeterProps | LoaderProps);
+export type ProgressBarProps = BaseProps &
+  (DeterminateProps | IndeterminateProps);
 
 export interface ResolvedProgressBarProps {
```
TheSonOfThomp, line 143: Can you document where this is used and why it's different from `ProgressBarProps`

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
```diff
@@ -44,14 +43,14 @@ export default meta;
...
-    maxValue: testValues.maxValue,
+    ...requiredA11yArgs,
+    value: storyValues.value,
+    maxValue: storyValues.maxValue,
     formatValue: 'fraction',
   },
   render: initialArgs => (
     <DynamicProgressBar
-      transitions={[[1500, { ...initialArgs, value: testValues.maxValue }]]}
+      transitions={[[1500, { ...initialArgs, value: storyValues.maxValue }]]}
```
TheSonOfThomp: nit: what's the 1500 for? Can we assign it to a constant?

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -165,73 +143,70 @@ export const WithHeaderTruncation: StoryObj<typeof ProgressBar> = {
   },
 };
 
-export const DeterminateVariants: StoryObj<typeof ProgressBar> = {
+export const IndeterminateVariants: StoryObj<typeof ProgressBar> = {
```
TheSonOfThomp, line 149: unfortunately I expect this will be flaky in chromatic. We might need to force it to render at a specific point in the indeterminate animation

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -165,73 +143,70 @@ export const WithHeaderTruncation: StoryObj<typeof ProgressBar> = {
...
-  },
-};
-
-export const MeterVariants: StoryObj<typeof ProgressBar> = {
+export const DeterminateMeterVariants: StoryObj<typeof ProgressBar> = {
   parameters: {
     generate: {
       combineArgs: {
-        status: [undefined, ...METER_STATUSES],
+        variant: [undefined, ...VARIANTS],
```
TheSonOfThomp: do we need to test `undefined` variant?

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -165,73 +143,70 @@ export const WithHeaderTruncation: StoryObj<typeof ProgressBar> = {
...
       },
       excludeCombinations: [
         {
-          variant: LOADER_VARIANTS.filter(
-            v =>
-              !(ANIMATED_LOADER_VARIANTS as Array<LoaderVariant>).includes(v),
+          variant: VARIANTS.filter(
+            v => !(ANIMATED_VARIANTS as Array<Variant>).includes(v),
           ),
           enableAnimation: [true],
```
TheSonOfThomp, line 185: I also expect these animated tests might be flaky in chromatic (the snapshot will be taken at a slightly different point in the animation each time)

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
TheSonOfThomp, line 1: These aren't really 'interactions', but just unique test cases. Not sure what word makes more sense though

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -65,110 +49,107 @@ type BaseProps = DarkModeProps &
...
+  /** Maximum progress value. */
   maxValue?: number;
 
-  /** Pauses progress and shows a disabled style. Not available if `isIndeterminate` is `true` for loaders. */
+  /** When `true`, shows a disabled style and pauses animation. */
   disabled?: boolean;
 }
+interface DeterminatePlainProgressProps {
+  /** Specify role of the progress bar ("progressbar" or "meter"). Defaults to "progressbar". */
+  roleType?: typeof Role.Progress;
```
TheSonOfThomp, line 88: why `roleType` and no `role`?

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -65,110 +49,107 @@ type BaseProps = DarkModeProps &
...
 }
 
 export const AnimationMode = {
-  DeterminateBase: 'determinate-base',
-  DeterminateAnimated: 'determinate-animated',
   Indeterminate: 'indeterminate',
+  DeterminatePlain: 'determinate-plain',
+  DeterminateAnimated: 'determinate-animated',
   Transition: 'indeterminate-to-determinate-transition',
 };
```
TheSonOfThomp:
```suggestion
} as const;
```

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -2,79 +2,76 @@ import React, { useEffect, useState } from 'react';
...
-    getAnimationMode({
-      type,
-      isIndeterminate,
-      enableAnimation,
-    }),
+    getAnimationMode(isIndeterminate, enableAnimation),
   );
 
   useEffect(() => {
     setAnimationMode(currentMode => {
```
TheSonOfThomp, line 74: I believe this can be a `useMemo` instead

TheSonOfThomp: Oh I see, maybe not, since you're calling `setAnimationMode` on transition end

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -96,115 +93,114 @@ export function ProgressBar(props: ProgressBarProps) {
...
+                animationMode,
+              })}
+              // if on fade-out transition, revert back to base mode
+              onTransitionEnd={() => {
+                if (animationMode === AnimationMode.Transition) {
+                  setAnimationMode(
+                    getAnimationMode(isIndeterminate, enableAnimation),
+                  );
+                }
+              }}
```
TheSonOfThomp: can we pull this into a named function above?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -246,45 +255,38 @@ export const getBarFillStyles = ({
...
     disabled,
   });
 
-  const indeterminate = getIndeterminateBarFillStyles({ theme, color });
+  const indeterminate = getIndeterminateFillStyles({ theme, variant });
 
   switch (animationMode) {
     case AnimationMode.Transition:
-      addOnStyles = cx(indeterminate, getTransitioningBarFillStyles());
+      addOnStyles = cx(indeterminate, transitioningFillStyles);
```
TheSonOfThomp: nit: not sure of the performance implications on calling `cx` twice for the same styles. Might be more efficient to push to an array, and then spread that array into `cx` on L291

packages/progress-bar/src/ProgressBar/hooks/useIdIdentifiers.tsx
```diff
@@ -0,0 +1,25 @@
...
+} => {
+  const baseId = useIdAllocator({
+    prefix: role,
+  });
+
+  return {
+    barId: baseId,
+    labelId: label ? `label-for-${baseId}` : undefined,
+    descId: description ? `desc-for-${baseId}` : undefined,
+    liveId: `live-region-for-${baseId}`,
```
TheSonOfThomp: does the `liveId` always exist?

packages/progress-bar/src/ProgressBar/utils/utils.tsx
```diff
@@ -4,121 +4,122 @@ import CheckmarkWithCircleIcon from '@leafygreen-ui/icon/dist/CheckmarkWithCircl
...
-};
-
+/**
+ * Returns a valid progress value.
+ * If undefined or greater than 0, returns as-is. If under 0, returns 0.
+ */
 const getValidValue = (value?: number) => {
   if (!isDefined(value) || value >= 0) return value;
   return 0;
 };
```
TheSonOfThomp: should this also check that the value is below the max value?

packages/progress-bar/src/ProgressBar/utils/utils.tsx
```diff
@@ -4,121 +4,122 @@ import CheckmarkWithCircleIcon from '@leafygreen-ui/icon/dist/CheckmarkWithCircl
...
-  if (props.type === Type.Meter) {
+  // indeterminate
+  if (props.isIndeterminate) {
     return {
       ...baseProps,
-      maxValue: getValidMaxValue(props.maxValue),
-      disabled: props.disabled ?? false,
-      color: getMeterStatusColor(props.status),
+      role: Role.Progress,
+      isIndeterminate: true,
```
TheSonOfThomp: nit: should we console warn if `props.role` is `meter` here?

packages/progress-bar/src/ProgressBar/utils/utils.tsx
```diff
@@ -143,59 +144,58 @@ export const getFormattedValue = (
...
-  label?: React.ReactNode,
-  description?: React.ReactNode,
-) => {
-  const role = type === Type.Meter ? 'meter' : 'progressbar';
+/**
+ * Returns appropriate ARIA attributes based on value and maxValue.
+ *
+ * - If `value` is undefined: `{ 'aria-busy': true }`
+ * - If `maxValue` is defined: `aria-valuemin`, `aria-valuemax`, `aria-valuenow`
+ * - Otherwise: `aria-valuetext`
```
TheSonOfThomp, line 193: nit: not required, but it'd be nice to link to the relevant WCAG docs here

packages/progress-bar/src/ProgressBar/utils/utils.spec.ts
```diff
@@ -27,61 +28,56 @@ describe('getFormattedValue', () => {
...
       maxValue: 100,
       disabled: false,
-      color: Color.Green,
       isIndeterminate: false,
       enableAnimation: true,
     });
   });
 
-  test('it correctly resolves props for an indeterminate loader type', () => {
+  test('it correctly resolves props for indeterminate', () => {
```
TheSonOfThomp, line 86: We should explicitly test how it handles the invalid combination of `isIndeterminate` + `meter`

packages/progress-bar/src/ProgressBar/constants/animations.constants.ts
```diff
@@ -0,0 +1,43 @@
...
+export const INDETERMINATE_ANIMATION_DURATION_MS = 1500;
+export const TRANSITION_ANIMATION_DURATION = 500;
+export const TEXT_ANIMATION_DURATION = 250;
+
+export const indeterminateBarWidths = {
+  narrow: '33%',
+  wide: '66%',
+};
+
+export const indeterminateBarPositions = {
```
TheSonOfThomp: can you add comments on how this is used?

packages/progress-bar/src/ProgressBar.stories.tsx
charcoalyy: GREAT point. that's a relic of the previous implementation.

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
charcoalyy: had this discussion with shaneeza in a previous PR, extending "interaction" to refer to any change in appearance/state. see discussion here: https://github.com/mongodb/leafygreen-ui/pull/2936#discussion_r2194856745

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
charcoalyy: because `role` is an intrinsic HTML attribute, it raised warnings about clashing. i figured it was less error-prone to separate the consumer input prop from the actual assignment of `role`. but, i suppose if it's mapped 1-1 anyway, perhaps this is fine. any thoughts here?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
charcoalyy: correct. i investigated that but there are some unique timing requirements for this specific case 😅

packages/progress-bar/src/ProgressBar/utils/utils.tsx
charcoalyy: discriminated union makes it impossible to specify the role of an indeterminate bar, i.e., if `isIndeterminate` is `true`, by default we apply the role of `progressbar`.

packages/progress-bar/src/ProgressBar/utils/utils.spec.ts
charcoalyy: see https://github.com/mongodb/leafygreen-ui/pull/2952#discussion_r2210651040 re:discriminated union. this is tested in the `types.spec.tsx` file, verifying that a role cannot be passed by the consumer to an indeterminate bar.

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
charcoalyy: hmm. it's an arbitrary value assigned based on what makes most (visual) sense in the story. i considered moving it to a constant, but because it held little meaning, i wasn't quite sure if that was the right move. do you have any thoughts on what it would be named?

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
charcoalyy: that's a great point!

packages/progress-bar/src/ProgressBar/hooks/useIdIdentifiers.tsx
charcoalyy: that's a good point. technically, it would only exist if there is a `screenReaderMessage` available. however, it reduces the number of props `useIdIdentifiers` requires if we just always make it available. since it's only ever passed to the `id` of the live region, would this matter?

the reason why undefined is important for `labelId` and `descId` is because these are passed into `aria-labelledby` and `aria-describedby` on top of the id fields themselves, so undefined is necessary in the event they do not exist.

this, however, makes me realize the live region perhaps doesn't need an id and this can be omitted entirely?

packages/progress-bar/src/ProgressBar/utils/utils.tsx
charcoalyy: originally, for prop readability purposes, i decided to not have `getValidValue` take in an optional max value, since by virtue of discriminated union, the logic below would require two extra lines. i capped all percentages at 100% in the `getPercentage` util.

BUT this makes me rethink that, because it makes more logical sense to have the validation util do the capping, not the calculation util.

charcoalyy: would actually love a second thought on this: https://github.com/mongodb/leafygreen-ui/blob/f8f9516f782df1f35f5cb47c98b85a46f94758fb/packages/progress-bar/src/ProgressBar/utils/utils.tsx#L117

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
charcoalyy: i debated giving them another name, but it felt inconsistent with the other packages. any thoughts?

packages/progress-bar/src/ProgressBar.stories.tsx
charcoalyy: oh no! i had looked at the loader/spinner stories for reference and totally missed that they disabled snapshots entirely. experimenting with figuring out a way to disable animation for chromatic only.

charcoalyy: https://www.chromatic.com/docs/animations/: "By default, CSS animations are paused at the end of their animation cycle (i.e., the last frame) when your tests run in Chromatic. This behavior is useful, specifically when working with animations that are used to “animate in” visible elements. If you want to override this behavior and pause the animation at the first frame, add the pauseAnimationAtEnd configuration option to your tests."

for infinite animations, we could pause animations at the first frame automatically. this is relevant in this component since all animations are CSS-only. thoughts?

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
TheSonOfThomp: what is it doing? Is it just an arbitrary delay? maybe something like `story_transition_delay`?

packages/progress-bar/src/ProgressBar/constants/animations.constants.ts
```diff
@@ -8,11 +8,19 @@ export const INDETERMINATE_ANIMATION_DURATION_MS = 1500;
...
 export const indeterminateBarWidths = {
   narrow: '33%',
   wide: '66%',
 };
 
+/**
+ * Positions for the indeterminate bar's left offset at keyframes
+ * for smooth horizontal movement in the animation cycle.
+ */
 export const indeterminateBarPositions = {
```
TheSonOfThomp: nit: maybe call this something like `indeterminateLeftOffset` to be explicit about where & how the var is meant to be used?

packages/progress-bar/src/ProgressBar.stories.tsx
TheSonOfThomp: That could work! We disable the loading indicator snapshots since the animations use `TransitionGroup` which we can't pause

packages/progress-bar/src/ProgressBar.interactions.stories.tsx
TheSonOfThomp: Not really sure. We can leave it for now

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
TheSonOfThomp: If we're extending and spreading the HTMLAttributes, this could still clash.
```jsx
// What's the rendered role here?
<ProgressBar role="button" roleType="meter" />
```

I think we should `Omit` the `role` attribute from the HTMLAttributes

packages/progress-bar/src/ProgressBar/hooks/useIdIdentifiers.tsx
TheSonOfThomp: I thought we might need it for `aria-controls` but I don't know if that's even the right pattern

packages/progress-bar/src/ProgressBar/utils/utils.tsx
TheSonOfThomp: Capping it in `getPercentages` makes sense. Just the name `getValidValue` makes me think it'll never return a value greater than the max. 

Someone could still pass in `value=50` and `maxValue=25` into the component

packages/progress-bar/src/ProgressBar/utils/utils.tsx
TheSonOfThomp: So, it's technically not _impossible_. Someone might not be using TypeScript, or even if they are, they could just ignore the warning

packages/progress-bar/src/ProgressBar/hooks/useIdIdentifiers.tsx
charcoalyy: yes! so, i think the intended use case for `aria-controls` is actually a bit different. it makes it easier to move to the controller/controllee for interactions. however, in this case, because the hidden live region is not something we want users to ever "land on", and is only intended for the screenreader, i believe it's not the right pattern.

packages/progress-bar/src/ProgressBar/utils/utils.tsx
charcoalyy: that is a great point. i guess it makes sense to validate that before having to get to this util in the first place!

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
charcoalyy: that will be addressed in a PR following this one! :) see here: https://github.com/mongodb/leafygreen-ui/compare/alina/progress-bar-refactor-v2...alina/progress-bar-add-ons

packages/progress-bar/src/ProgressBar/utils/utils.tsx
charcoalyy: OHHHHHHH i see!!!! that's a great point - i will add a test for handling and a warning in that event!

TheSonOfThomp (approved):

PR #2953, opened 2025-07-14
LG-4867: sticky header interaction test snapshots

charcoalyy:
## ✍️ Proposed changes

Decorate the sticky header interaction test with a fixed height container shorter than its contents, in order to guarantee a "scroll" is possible for Chromatic snapshots. Additionally test for header visibility directly.

🎟 _Jira ticket:_ [LG-4867](https://jira.mongodb.org/browse/LG-4867)

### For new components

### DIFF
--- a/chat/chat-window/src/ChatWindow.stories.tsx
+++ b/chat/chat-window/src/ChatWindow.stories.tsx
@@ -12,7 +12,7 @@ import baseMessages from './utils/baseMessages';
 import { ChatWindow } from '.';
 
 const meta: StoryMetaType<typeof ChatWindow> = {
-  title: 'Composition/ChatWindow',
+  title: 'Composition/Chat/ChatWindow',
   component: ChatWindow,
   args: {
     title: 'LG Chat Demo',
--- a/packages/avatar/src/Avatar/Avatar.tsx
+++ b/packages/avatar/src/Avatar/Avatar.tsx
@@ -8,6 +8,9 @@ import { getAvatarStyles } from './Avatar.styles';
 import { AvatarProps } from './Avatar.types';
 import { AvatarContents } from './AvatarContents';
 
+/**
+ * The Avatar component is a user interface element that represents an individual user or entity within a digital platform or application. Avatars serve as visual identifiers, often depicting a user's profile picture or a symbolic representation, such as initials or an Icon. The primary purpose of Avatars is to provide a visual cue about the identity of the associated user or entity.
+ */
 export const Avatar = forwardRef<HTMLDivElement, AvatarProps>(
   (props, fwdRef) => {
     const {
--- a/packages/button/src/Button/Button.tsx
+++ b/packages/button/src/Button/Button.tsx
@@ -16,7 +16,7 @@ import { BaseButtonProps, Size, Variant } from '../types';
 import { getClassName } from './Button.styles';
 
 /**
- * Buttons allow users to take actions, and make choices, with a single tap.
+ * Buttons allow users to take actions, and make choices, with a single tap. Buttons are a type of call to action (CTA) the user can click or press.
  */
 export const Button = InferredPolymorphic<BaseButtonProps, 'button'>(
   (
--- a/packages/chip/src/Chip/Chip.tsx
+++ b/packages/chip/src/Chip/Chip.tsx
@@ -17,6 +17,9 @@ import {
 } from './Chip.styles';
 import { ChipProps, TruncationLocation, Variant } from './Chip.types';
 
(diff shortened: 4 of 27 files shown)


packages/table/src/Table.Interactions.stories.tsx
```diff
@@ -184,17 +198,24 @@ const Template: StoryFn<StoryTableProps> = args => {
...
 
-// TODO: fix this test https://jira.mongodb.org/browse/LG-4867
 export const StickyHeader = {
   render: (args: StoryTableProps) => <Template {...args} />,
-  play: async ({ canvasElement }) => {
+  play: async ({ canvasElement }: { canvasElement: HTMLElement }) => {
     const canvas = within(canvasElement);
+    const container = await canvas.findByTestId('wrapper-container');
     const table = await canvas.findByTestId('lg-table');
+    const headerCells = await canvas.findAllByTestId('lg-table-header');
```
charcoalyy: i notice that the current implementation applies `data-lgid="lg-table-header"` to each individual header cell, and similarly for each individual body cell. i'm curious as to why we do this instead of on the row itself?

shaneeza: This may have been an oversight on the table header row, but we have testids on each row inside the table body, and inside each row, each cell also has an id. I'm assuming you would have used the header row testid here instead of looping through all the table header row cells? If that's the case, feel free to add that testid.

packages/table/src/utils/getLgIds.ts
```diff
@@ -7,7 +7,9 @@ export const getLgIds = (root: LgIdString = DEFAULT_LGID_ROOT) => {
     root,
     cell: `${root}-cell`,
     row: `${root}-row`,
-    header: `${root}-header`,
+    headerCell: `${root}-header_cell`,
+    headerRow: `${root}-header_row`,
+    header: `${root}-header`, // deprecating in favor of headerRow
```
charcoalyy: can this be safely deleted? would this break anything outside of (possibly) consumer tests?

shaneeza: This would probably break consumer tests. If we do this, then this change should be marked as `major` and we should explain this change in the changeset.

packages/table/src/Cell/HeaderCell/HeaderCell.spec.tsx
```diff
@@ -16,9 +16,9 @@ import { Person } from '../../utils/makeData.testutils';
 
 import HeaderCell, { HeaderCellProps } from '.';
 
-const lgids = getLgIds();
+const lgIds = getLgIds();
```
shaneeza, line 19: 🙏🏽

.changeset/loose-shirts-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/table': major
+---
+
+Patch table header interaction test to scroll in Chromatic snapshots. Introduces breaking change in testing utils: `getLgIds` now returns `headerRow` and `headerCell`. Deprecates the original `header` lgId, which is now `headerCell`.
```
shaneeza: I would format this with bullets so it's easier to scan:

- Patch table header interaction test to scroll in Chromatic snapshots
- Introduces breaking change in testing utils: `getLgIds` now returns `headerRow` and `headerCell`. Deprecates the original `header` lgId, which is now `headerCell`.

I would also explicitly show the before and after changes for the data-testid on `headerCell. Something like:

`headerCell`  **Before**
```jsx
data-testid='lg-table.....'
```
`headerCell` **After**
```jsx
data-testid='lg-table.....'
```

shaneeza: Dev sync has me thinking about this version bump. Stephen was trying to avoid a major bump with LG chat because LG chat had just recently had a major bump. Table just had a major bump as well. Does the sticky header test pass without the header cell check? If so, we can remove that check and avoid a major bump here. We can create a ticket to update the testids later.

charcoalyy: gotcha! i'd like to keep the additive (non-breaking) change of a `headerRow` lgId, and maintain the existing `header` with a note about switching over to `headerCell`. i'll also make a quick ticket.

.changeset/every-apes-return.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/table': patch
```
shaneeza: Since we're adding the new testid, this should be `minor`. `Patch` is for fixes, while `minor` is for additions.

PR #2956, opened 2025-07-15
LG-5335: `compact` menu variant

tsck:
## ✍️ Proposed changes

### For new components


packages/menu/src/Menu.stories.tsx
```diff
@@ -184,9 +191,53 @@ export const InitialOpen = {
...
+    );
+  },
+  parameters: {
+    chromatic: {
+      disableSnapshot: true,
+    },
+  },
+} satisfies StoryObj<typeof Menu>;
+
+export const InitialOpenCompact = {
```
tsck, line 220: Just added this to have some visual regression on the compact menu

packages/menu/src/Menu.stories.tsx
```diff
@@ -109,8 +111,13 @@ export default {
...
       control: 'boolean',
     },
+    variant: {
+      description: 'The variant of the menu',
+      options: MenuVariant,
+      control: { type: 'select' },
+    },
   },
-} satisfies StoryMetaType<typeof Menu>;
+} as StoryMetaType<typeof Menu>;
```
tsck: TS was complaining about this, and this fixed it.

packages/menu/src/Menu.stories.tsx
```diff
@@ -334,8 +385,16 @@ export const DarkModeRightAlign = {
...
+interface InitialLongMenuOpenArgs extends MenuProps {}
+
+interface InitialLongMenuOpenStory {
+  render: (args: InitialLongMenuOpenArgs) => JSX.Element;
+  play: (ctx: { canvasElement: HTMLElement }) => Promise<void>;
+  decorators: [(StoryFn: React.FC, _ctx: any) => JSX.Element];
+}
+
+export const InitialLongMenuOpen: InitialLongMenuOpenStory = {
+  render: (args: InitialLongMenuOpenArgs) => {
```
tsck, line 399: Added types to keep TS from complaining as well

packages/menu/src/Menu/Menu.styles.ts
```diff
@@ -7,17 +7,20 @@ import { color, spacing } from '@leafygreen-ui/tokens';
...
 
-export const getMenuStyles = ({ theme }: MenuStyleArgs) => {
+export const getMenuStyles = ({ theme, variant }: MenuStyleArgs) => {
   return css`
     width: 210px;
     border-radius: ${spacing[300]}px;
     overflow: auto;
-    // FIXME: Should this really be 14px?
-    padding: ${spacing[300] + spacing[50]}px 0;
+    padding: ${variant === MenuVariant.Default ? spacing[300] : spacing[150]}px
```
tsck: Spoke with Sooa who said to just make the spacing 12px instead of 14px.

packages/menu/src/MenuItem/MenuItem.styles.ts
```diff
@@ -45,6 +48,10 @@ export const getMenuItemStyles = ({
       width: 100%;
       min-height: ${spacing[800]}px;
       background-color: ${menuColor[theme].background.default};
+      padding: ${menuVariant === MenuVariant.Default
+          ? spacing[200]
+          : spacing[150]}px
+        ${spacing[300]}px;
```
tsck: On the `default` variant each item had `12px` all around. However, in the Figma designs, this seemed incorrect and it should be `8px 12px`. This also results in all visual regression tests failing due to the slight change in padding.

packages/menu/src/Menu.stories.tsx
tsck: This was resolved in [bebd3b6](https://github.com/mongodb/leafygreen-ui/pull/2956/commits/bebd3b6d89c43a982cb663b0519c81d726cdddb4)

packages/menu/src/MenuItem/InternalMenuItemContent.tsx
```diff
@@ -96,6 +103,7 @@ export const InternalMenuItemContent = React.forwardRef<
             highlighted,
             theme,
             variant,
+            menuVariant: menuVariant || MenuVariant.Default,
```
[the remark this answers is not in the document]
tsck: `menuVariant` is of type `MenuVariant | undefined`. Since `getMenuItemStyles` only accepts `MenuVariant` falling back to `default` is required here

packages/menu/src/MenuItem/InternalMenuItemContent.tsx
```diff
@@ -55,7 +56,13 @@ export const InternalMenuItemContent = React.forwardRef<
   ) => {
     const { as } = useInferredPolymorphic(asProp, rest, 'button');
 
-    const { theme, darkMode, highlight, renderDarkMenu } = useMenuContext();
+    const {
+      theme,
+      darkMode,
+      highlight,
+      renderDarkMenu,
+      variant: menuVariant,
```
stephl3: nit: might set the default here instead
```suggestion
      variant: menuVariant = MenuVariant.Default,
```

tsck: Oh good point! Will update. 

Will also check in with Sooa on the padding. I think you raise a good point about the semver bump though. Even without the default correction, this is "fixing" the overall menu padding as well to align it with our tokens. So should this be a major bump anyway perhaps?

packages/menu/src/MenuItem/MenuItem.styles.ts
tsck: @sooachung Can I get your opinion on this please? Currently a `MenuItem` has 12px top and bottom padding. However, looking at the designs, this seems to be incorrect. The designs have 8px top and bottom padding. Would you prefer we leave it as is, or bring it into alignment with the linked designs?

tsck: Confirmed with @sooachung that this should be fixed to align with designs

packages/menu/src/MenuItem/InternalMenuItemContent.tsx
tsck: Aligned w/ Sooa that we should make this change, and given the above context as well, I've pushed an update to the changeset to make this a major change.

@stephl3 Just want to confirm that we're aligned here before merging.

PR #2958, opened 2025-07-16
Prevent custom `data-testid` from being ignored

shaneeza:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/packages/menu/src/Menu/Menu.tsx
+++ b/packages/menu/src/Menu/Menu.tsx
@@ -253,8 +253,8 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
             {/* Need to stop propagation, otherwise Menu will closed automatically when clicked */}
             {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events*/}
             <ul
-              {...rest}
               data-testid={lgIds.root}
+              {...rest}
               data-lgid={lgIds.root}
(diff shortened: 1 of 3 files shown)


packages/menu/src/Menu/Menu.tsx
```diff
@@ -253,8 +253,8 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
             {/* Need to stop propagation, otherwise Menu will closed automatically when clicked */}
             {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events*/}
             <ul
-              {...rest}
               data-testid={lgIds.root}
+              {...rest}
```
tsck, line 257: Q: Should we just move the spreading of `rest` to after all of props these so that they can all technically be overridden?

shaneeza: I'm not sure if we want consumers to override those props. It would require some investigation, and unfortunately, I don't have the bandwidth to investigate at the moment. This PR mainly addresses the regression from the lgid work I recently worked on, which is creating issues on another team. I can create a ticket to investigate this further.

tsck: That's fair. It's also not important, not even sure if a ticket's necessary. It's easier to give the option to override later if needed than restrict later once it's available, so no pushback by me 🙂

PR #2961, opened 2025-07-17
LG-5357: progress bar customizability modifications

charcoalyy:
## ✍️ Proposed changes

- introduce built-in option to auto-cycle through an array of descriptions
- extend HTML div attributes for additionally styling customizability
- introduce a hook for calculating CSS transition durations based on magnitude of change and desired speed

🎟 _Jira ticket:_ [LG-5357](https://jira.mongodb.org/browse/LG-5357)

### For new components


packages/progress-bar/src/ProgressBar/hooks/useComputedTransitionDuration.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { usePrevious } from '@leafygreen-ui/hooks';
+
+/**
+ * Computes the total transition duration for a CSS animation based on
+ * the change magnitude and a specified speed.
+ *
+ * Note: The previous value is tracked internally via a ref, so it
+ * does not need to be provided as an argument.
+ *
+ * @param param.speed - Desired change in units per second.
```
charcoalyy, line 12: tried using `param.paramName` prefix to differentiate object args used here from positional args elsewhere. is there a typical convention we follow?

packages/progress-bar/src/ProgressBar/hooks/useComputedTransitionDuration.tsx
charcoalyy, line 1: i deliberately wrote both `useComputedTransitionDuration` and `useRotatingItems` hooks to be as general as possible. do we think these are appropriate to move to the wider hooks package, or is it best to keep them in progress-bar until an actual outer use case presents itself?

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -104,8 +133,9 @@ export function ProgressBar(props: ProgressBarProps) {
     <LeafyGreenProvider darkMode={darkMode}>
       <div
         data-lgid={lgIds.root}
-        className={containerStyles}
         aria-disabled={disabled}
+        className={getContainerStyles(className)}
```
charcoalyy, line 140: consumer `classNames` will only ever be passed to the outermost wrapper element, with no control over styling of inner elements. is this appropriate?

PR #2963, opened 2025-07-17
[LG-5362] Add chat/suggestions

joyjwang:
## ✍️ Proposed changes

add SuggestedActions component

🎟 _Jira ticket: [LG-5362](https://jira.mongodb.org/browse/LG-5362)

### For new components

## 🧪 How to test changes

- [ ] review Chats/SuggestionCard/LiveExample

### DIFF
--- a/charts/legend/src/LegendCheckbox/LegendCheckbox.styles.ts
+++ b/charts/legend/src/LegendCheckbox/LegendCheckbox.styles.ts
@@ -31,7 +31,7 @@ const getBaseLegendCheckboxStyles = ({
     .${checkWrapperClassName} {
       border-color: ${checkboxColor || defaultCheckboxColor};
 
-      &:before {
+      &::before {
         background-color: ${checkboxColor || defaultCheckboxColor};
       }
     }
--- a/chat/input-bar/src/InputBar/InputBar.styles.ts
+++ b/chat/input-bar/src/InputBar/InputBar.styles.ts
@@ -157,8 +157,8 @@ const gradientWidth = 3;
 const gradientOffset = 1;
 
 export const gradientAnimationStyles = css`
-  &:before,
-  &:after {
+  &::before,
+  &::after {
     content: '';
     position: absolute;
     top: -${gradientWidth + gradientOffset}px;
@@ -171,11 +171,11 @@ export const gradientAnimationStyles = css`
     background-position: 800% 800%; // set final state of animation
   }
 
-  &:after {
+  &::after {
     animation: 4s animateBg linear;
   }
 
-  &:before {
+  &::before {
     filter: blur(4px) opacity(0.6);
     animation: 4s animateBg, animateShadow linear infinite;
     opacity: 0;
--- a/chat/message/src/Message/Message.styles.ts
+++ b/chat/message/src/Message/Message.styles.ts
@@ -56,7 +56,7 @@ export const messageContainerWrapperStyles = css`
 
 const sharedMessageContainerWedgeStyles = css`
   // Left wedge
-  &:before {
+  &::before {
     content: '';
     position: absolute;
     left: 0;
@@ -70,13 +70,13 @@ const sharedMessageContainerWedgeStyles = css`
 export const messageContainerWedgeStyles = {
   [Theme.Dark]: css`
     ${sharedMessageContainerWedgeStyles}
-    &:before {
+    &::before {
       background-color: ${palette.green.base};
     }
   `,
   [Theme.Light]: css`
     ${sharedMessageContainerWedgeStyles}
-    &:before {
+    &::before {
       background-color: ${palette.green.dark2};
     }
   `,
--- a/chat/message/src/MessageBanner/MessageBanner.styles.ts
+++ b/chat/message/src/MessageBanner/MessageBanner.styles.ts
@@ -6,7 +6,7 @@ import { spacing } from '@leafygreen-ui/tokens';
  */
 export const baseStyles = css`
   // Remove the Banner's left border wedge
-  &:before {
+  &::before {
     content: '';
     background: transparent;
   }
--- a/packages/banner/src/Banner/styles.ts
+++ b/packages/banner/src/Banner/styles.ts
@@ -15,7 +15,7 @@ export const baseBannerStyles = css`
   border-radius: 12px;
   font-family: ${fontFamilies.default};
 
-  &:before {
+  &::before {
     content: '';
     position: absolute;
     width: 13px;
@@ -44,7 +44,7 @@ const darkModeInfoBannerStyles = css`
     }
   }
 
-  &:before {
+  &::before {
     background: linear-gradient(
       to left,
       transparent 6px,
@@ -71,7 +71,7 @@ const darkModeWarningBannerStyles = css`
     }
   }
 
-  &:before {
+  &::before {
     background: linear-gradient(
       to left,
       transparent 6px,
@@ -98,7 +98,7 @@ const darkModeDangerBannerStyles = css`
     }
   }
 
-  &:before {
+  &::before {
     background: linear-gradient(
       to left,
       transparent 6px,
@@ -125,7 +125,7 @@ const darkModeSuccessBannerStyles = css`
     }
   }
 
-  &:before {
+  &::before {
     background: linear-gradient(
       to left,
       transparent 6px,
@@ -152,7 +152,7 @@ const lightModeInfoBannerStyles = css`
     }
   }
 
-  &:before {
+  &::before {
     background: linear-gradient(
       to left,
       transparent 6px,
@@ -179,7 +179,7 @@ const lightModeWarningBannerStyles = css`
     }
   }
 
-  &:before {
+  &::before {
     background: linear-gradient(
       to left,
       transparent 6px,
@@ -206,7 +206,7 @@ const lightModeDangerBannerStyles = css`
     }
   }
 
-  &:before {
+  &::before {
     background: linear-gradient(
       to left,
       transparent 6px,
@@ -233,7 +233,7 @@ const lightModeSuccessBannerStyles = css`
     }
   }
 
-  &:before {
+  &::before {
     background: linear-gradient(
       to left,
       transparent 6px,
--- a/packages/banner/src/BannerDismissButton/styles.ts
+++ b/packages/banner/src/BannerDismissButton/styles.ts
@@ -24,7 +24,7 @@ const darkModeInfoStyles = css`
     box-shadow: 0 0 0 2px ${palette.blue.dark3},
       0 0 0 4px ${palette.blue.light1};
 
-    &:before {
+    &::before {
       background-color: ${palette.blue.dark2};
     }
   }
@@ -39,7 +39,7 @@ const darkModeWarningStyles = css`
     box-shadow: 0 0 0 2px ${palette.yellow.dark3},
       0 0 0 4px ${palette.blue.light1};
 
-    &:before {
+    &::before {
       background-color: ${palette.yellow.dark2};
     }
   }
@@ -54,7 +54,7 @@ const darkModeDangerStyles = css`
     color: ${palette.red.light2};
     box-shadow: 0 0 0 2px ${palette.red.dark3}, 0 0 0 4px ${palette.blue.light1};
 
-    &:before {
+    &::before {
       background-color: ${palette.red.dark2};
     }
   }
@@ -70,7 +70,7 @@ const darkModeSuccessStyles = css`
     box-shadow: 0 0 0 2px ${palette.green.dark3},
       0 0 0 4px ${palette.blue.light1};
 
-    &:before {
+    &::before {
       background-color: ${palette.green.dark2};
     }
   }
@@ -84,7 +84,7 @@ const lightModeInfoStyles = css`
   &:focus-visible {
     color: ${palette.blue.dark2};
 
-    &:before {
+    &::before {
       background-color: ${palette.blue.light2};
     }
   }
@@ -98,7 +98,7 @@ const lightModeWarningStyles = css`
   &:focus-visible {
     color: ${palette.yellow.dark2};
 
-    &:before {
+    &::before {
       background-color: ${palette.yellow.light2};
     }
   }
@@ -112,7 +112,7 @@ const lightModeDangerStyles = css`
   &:focus-visible {
     color: ${palette.red.dark2};
 
-    &:before {
+    &::before {
       background-color: ${palette.red.light2};
     }
   }
@@ -126,7 +126,7 @@ const lightModeSuccessStyles = css`
   &:focus-visible {
     color: ${palette.green.dark2};
 
-    &:before {
+    &::before {
       background-color: ${palette.green.light2};
     }
   }
--- a/packages/callout/src/Callout/Callout.styles.ts
+++ b/packages/callout/src/Callout/Callout.styles.ts
@@ -77,7 +77,7 @@ export const getBaseStyles = (theme: Theme, variant: Variant) =>
     padding-inline-start: ${spacing[300]}px;
     position: relative;
 
-    &:after {
+    &::after {
       content: '';
       position: absolute;
       width: 3px;
--- a/packages/checkbox/src/Check/Check.style.ts
+++ b/packages/checkbox/src/Check/Check.style.ts
@@ -18,7 +18,7 @@ export const checkBorderSizePx = 2;
 
 export const disableAnimation = css`
   &,
-  &:before {
+  &::before {
     transition: unset;
     transition-delay: 0ms;
     transition-duration: 0ms;
@@ -47,7 +47,7 @@ export const wrapperBaseStyle = css`
   /**
    * The animated background circle
    */
-  &:before {
+  &::before {
     content: '';
     position: absolute;
     inset: ${insetPct}%;
@@ -64,14 +64,14 @@ export const wrapperThemeStyle: Record<Theme, string> = {
   [Theme.Light]: css`
     border-color: ${palette.gray.dark2};
 
-    &:before {
+    &::before {
       background-color: ${palette.blue.base};
     }
   `,
   [Theme.Dark]: css`
     border-color: ${palette.gray.base};
 
-    &:before {
+    &::before {
       background-color: ${palette.blue.light1};
     }
   `,
@@ -81,7 +81,7 @@ export const wrapperCheckedBaseStyle = css`
   // Delay background transition in
   transition-delay: 0ms, ${checkAnimationDuration}ms, 0ms;
 
-  &:before {
+  &::before {
     transform: scale(1);
     // No delay on enter
     transition-delay: 0ms;
@@ -105,7 +105,7 @@ export const wrapperDisabledStyle: Record<Theme, string> = {
     background-color: ${palette.gray.light3};
     box-shadow: unset;
 
-    &:before {
+    &::before {
       background-color: ${palette.gray.light3};
     }
   `,
@@ -115,7 +115,7 @@ export const wrapperDisabledStyle: Record<Theme, string> = {
     background-color: ${palette.gray.dark3};
     box-shadow: unset;
 
-    &:before {
+    &::before {
       background-color: ${palette.gray.dark3};
     }
   `,
@@ -125,14 +125,14 @@ export const wrapperCheckedDisabledStyle: Record<Theme, string> = {
   [Theme.Light]: css`
     background-color: ${palette.gray.light2};
 
-    &:before {
+    &::before {
       background-color: ${palette.gray.light2};
     }
   `,
   [Theme.Dark]: css`
     background-color: ${palette.gray.dark2};
 
-    &:before {
+    &::before {
       background-color: ${palette.gray.dark2};
     }
   `,
--- a/packages/code/src/Code/Code.styles.ts
+++ b/packages/code/src/Code/Code.styles.ts
@@ -156,7 +156,7 @@ export const codeWithoutPanelStyles = css`
   // No panel, all code
   grid-template-areas: 'code code';
 
-  &:after {
+  &::after {
     grid-column: -1; // Placed on the right edge
   }
 `;
@@ -167,8 +167,8 @@ export const codeWithPanelStyles = css`
     'code';
   grid-template-columns: unset;
 
-  &:before,
-  &:after {
+  &::before,
+  &::after {
     grid-row: 2; // Placed on the top under the Picker Panel
   }
 `;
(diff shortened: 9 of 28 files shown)


chat/suggestion-card/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "docs": "lg-build docs"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/emotion": "workspace:^",
+    "@leafygreen-ui/lib": "workspace:^",
+    "@lg-tools/test-harnesses": "workspace:^"
+  },
```
stephl3: Need to add the deps used in the source code here. looks like it might just be banner + button?

chat/suggestion-card/tsconfig.json
```diff
@@ -0,0 +1,31 @@
...
+    },
+    {
+      "path": "../../packages/emotion"
+    },
+    {
+      "path": "../../packages/lib"
+    },
+    {
+      "path": "../../packages/table"
+    },
```
stephl3: did you try using LG table and it didn't work? let's drop this if not needed

chat/suggestion-card/src/testing/index.ts
```diff
@@ -0,0 +1,2 @@
+export { getTestUtils } from './getTestUtils';
+export { type TestUtilsReturnType } from './getTestUtils.types';
```
stephl3: We've been creating test utils as a devX testing tool. I think because it might increase scope of this work, let's skip this for now, and I can revisit it towards the end of the project. We should be okay to leave the scaffolding as-is and just drop these 2 export lines

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Options to control the status of the SuggestionCard
+ */
+export const Status = {
+  Available: 'Available',
+  Apply: 'Apply',
+  Success: 'Success',
+  ApplyError: 'ApplyError',
+  GeneralError: 'GeneralError',
```
stephl3: we generally do k-v pairs for enums as:
PascalCase: 'kebab-case'

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
```diff
@@ -0,0 +1,65 @@
+import { Variant as BannerVariant } from '@leafygreen-ui/banner';
+import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
+
+/**
+ * Options to control the status of the SuggestionCard
+ */
+export const Status = {
+  Available: 'Available',
+  Apply: 'Apply',
```
stephl3: do we need both of these? unclear how they differ

if these are intended as default state, can we use an existing pattern of `Unset: 'unset'`?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
stephl3: I'm wondering if we need the general error. I think that is related to an error occuring in the message response and not for this card specifically. I wonder if we can just have `Error: 'error'` for this specific case of an error occurring when applying the suggestions

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Parameters for the suggested configuration
+ */
+export interface SuggestedConfigurationParameters {
+  clusterTier: string;
+  price: string;
+  cloudProvider: string;
+  storage: string;
+  ram: string;
+  vCPUs: string;
```
stephl3: do we know that these are _always_ the suggestions that will be applied or should this be made more generic to accept and type of keys? also are they all required?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/**
+ * Properties for the status banner
+ */
+export interface StatusBannerProps {
+  variant: BannerVariant;
+  label?: string;
+  description: string;
+}
```
stephl3: looks like this might not be getting used? if so, can we remove?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  suggestedConfigurationParameters: SuggestedConfigurationParameters;
+  /**
+   * Callback fired when the user clicks the "Apply" button
+   */
+  handleApply: () => void;
+  /**
+   * Callback fired when the user clicks the "Retry" button
+   */
+  handleRetry: () => void;
```
stephl3: 1. conventionally, these should be `onApply` and `onRetry`
2. for `onApply`, can we use more explicit naming like `onClickApply` or `onAttemptApply`? since it's possible that it may error, `onApply` may mislead consumers to believe that they can hook into this prop to take an action _when_ the suggestions actually apply which this is not being implemented for
3. we might not need `onRetry` if the general error is not needed for this component

chat/suggestion-card/src/SuggestionCard.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+      control: 'object',
+    },
+  },
+};
+
+const Template: StoryFn<SuggestionCardProps> = props => (
+  <SuggestionCard {...props} />
+);
+
+export const Basic = Template.bind({});
```
stephl3: can we replace `Basic` with `LiveExample` and include `Generated` stories? here are a couple examples that will help:
- https://github.com/mongodb/leafygreen-ui/blob/main/packages/context-drawer/src/ContextDrawer.stories.tsx
- https://github.com/mongodb/leafygreen-ui/blob/main/packages/banner/src/Banner.stories.tsx

chat/suggestion-card/src/SuggestionCard/SuggestionCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+  GeneralErrorContainerStyle,
+  ListStyle,
+  SuggestionCardWrapperStyle,
+  TableCellStyle,
+  TableHeaderStyle,
+  TableStyle,
+} from './SuggestionCard.styles';
+import { Status, SuggestionCardProps } from './SuggestionCard.types';
+
+const SuggestionCard = (props: SuggestionCardProps) => {
```
stephl3: can we wrap this in a forwardRef?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <li>
+        <u>Cloud Provider & Region</u>: {cloudProvider}
+      </li>
+      <li>
+        <u>Cluster Tier</u>: {clusterTier}
+      </li>
+    </ul>
+  );
+
+  const StatusBanner = ({ status }: { status: Status }) => {
```
stephl3: can we move this to a separate component?
`chat/suggestion-card/src/StatusBanner/StatusBanner.tsx`

chat/suggestion-card/src/SuggestionCard/SuggestionCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+  BannerWrapperStyle,
+  BoldedTextStyle,
+  DividerStyle,
+  GeneralErrorContainerStyle,
+  ListStyle,
+  SuggestionCardWrapperStyle,
+  TableCellStyle,
+  TableHeaderStyle,
+  TableStyle,
+} from './SuggestionCard.styles';
```
stephl3: we use camelCase for these emotion-generated styles and typically make them plural if more than 1 line of styling. `applyButtonStyles`, `bannerWrapperStyles`, etc

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,66 @@
+import { css } from '@leafygreen-ui/emotion';
+import { palette } from '@leafygreen-ui/palette';
+import { fontWeights, spacing } from '@leafygreen-ui/tokens';
+
+export const DividerStyle = css`
+  width: 353px;
```
stephl3: let's extract this magic number into a constant

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+`;
+
+export const SuggestionCardWrapperStyle = css`
+  background-color: ${palette.gray.light3};
+  border: 1px solid ${palette.gray.light2};
```
stephl3:
```suggestion
export const getSuggetionCardWrapperStyles = (theme: Theme) => css`
  background-color: ${color[theme].background...};
  border: 1px solid ${color[theme].border...};
```
these need to consider `theme` to support dark mode

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+  border-radius: 12px;
```
stephl3:
```suggestion
  border-radius: ${borderRadius[300]}px;
```

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+  width: 100%;
+  padding: ${spacing[300]}px;
+  gap: ${spacing[200]}px;
+  font-size: 13px;
+  line-height: 20px;
```
stephl3:
```suggestion
  font-size: ${typeScales.body1.fontSize}px;
  line-height: ${typeScales.body1.lineHeight}px;
```

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+  width: 100%;
+  margin-top: ${spacing[200]}px;
+  font-weight: ${fontWeights.regular};
+`;
+
+export const BannerWrapperStyle = css`
+  width: 100%;
+  gap: ${spacing[200]}px;
+  font-size: 13px;
+  line-height: 20px;
```
stephl3:
```suggestion
  font-size: ${typeScales.body1.fontSize}px;
  line-height: ${typeScales.body1.lineHeight}px;
```

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+  margin-top: ${spacing[400]}px;
+  justify-content: space-between;
+`;
+
+export const BoldedTextStyle = css`
+  font-weight: 600;
```
stephl3: can you check with design on using an existing font-weight token here?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+`;
+
+export const TableStyle = css`
+  width: 100%;
+  gap: ${spacing[200]}px;
+  line-height: 20px;
```
stephl3:
```suggestion
  line-height: ${typeScales.body1.lineHeight}px;
```

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+`;
+
+export const TableCellStyle = css`
+  text-align: right;
+  padding: ${spacing[50]}px 0;
+`;
+
+export const ListStyle = css`
+  padding-left: ${spacing[500]}px;
+  line-height: 20px;
```
stephl3:
```suggestion
  line-height: ${typeScales.body1.lineHeight}px;
```

chat/suggestion-card/tsconfig.json
joyjwang: mmm i started with LG table but i was having trouble with the built in padding/margins, is there a way I can set those to 0?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
joyjwang: ahh yep the available state is meant to be the default, will update to `Unset: 'unset'`!

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
joyjwang: great point!! i misunderstood the design will update!

chat/suggestion-card/tsconfig.json
stephl3: It depends on what you're looking to change, but you might be able to override some styles with the `className` prop

joyjwang: hmmm i'm trying to override with className but some of the base padding is hard to override maybe we could pair later if we really want to use LG table?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
joyjwang: can def rename!! i saw in the style guide that we should prefix with [handle](https://github.com/mongodb/leafygreen-ui/blob/main/STYLEGUIDE.md#prefer-prefixing-event-handlers-with-handle), is that not relevant to this case?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
joyjwang: ooo i actually have a q about this one, is there a min width for the chat? i mostly added this for local testing, but i think we want this width to be adjustable

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
joyjwang: mmm i had thought about that and generic would be best, but it looks like the banner references specific params, so let me clarify with design!

chat/suggestion-card/tsconfig.json
stephl3: gotcha, if it's not straightforward to do without using additional selectors, it's probably best to avoid that for now

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
stephl3: The mention in the styleguide is to name callbacks in app code as `handle*`. However, `on*` is the convention for prop naming for consumer APIs

This way, when someone needs to hook into the "Apply" button for this card, they can define it like so:
```
<SuggestionCard
  onClickApply={handleClickApply}
/>
```

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
stephl3: ok, in that case, this probably should extend to the full width of the container. probably best to confirm that with design

joyjwang: confirmed with design and looks like we're sticking with fixed size for this MVP! i'll extract this number into a constant 👍

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
joyjwang: chatted with design and looks like we can go with generic key-value pairs! also decided to make the params shown in the status banner dynamic, so added props + support for that functionality!

.changeset/polite-walls-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-chat/suggestion-card': minor
```
stephl3: Because this is a new package, you actually won't need this changeset file. When this PR goes in, the release pipeline will publish v0.1.0

chat/suggestion-card/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@leafygreen-ui/emotion": "workspace:^",
+    "@leafygreen-ui/lib": "workspace:^",
+    "@leafygreen-ui/banner": "workspace:^",
+    "@leafygreen-ui/button": "workspace:^",
+    "@leafygreen-ui/icon": "workspace:^",
+    "@leafygreen-ui/tokens": "workspace:^"
```
stephl3: nit:
```suggestion
    "@leafygreen-ui/banner": "workspace:^",
    "@leafygreen-ui/button": "workspace:^",
    "@leafygreen-ui/emotion": "workspace:^",
    "@leafygreen-ui/icon": "workspace:^",
    "@leafygreen-ui/lib": "workspace:^",
    "@leafygreen-ui/tokens": "workspace:^"
```

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
```diff
@@ -0,0 +1,48 @@
+import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
+
+/**
+ * Options to control the status of the SuggestionCard
+ */
+export const Status = {
+  Unset: 'unset',
+  Apply: 'apply',
```
stephl3: can you share more on how/when this `'unset'` status will be used? seems like all we really need is `'apply'` but curious if there's context I'm missing

chat/suggestion-card/package.json
```diff
@@ -0,0 +1,52 @@
+  
+{
+  "name": "@lg-chat/suggestion-card",
```
stephl3: What do you think about this renamed structure for the package and main component export?
package: `@lg-chat/suggestions` (not limited to the card UI)
main component: `SuggestedActions` (has both the main card but also a status banner attached)

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  suggestedConfigurationParameters: ConfigurationParameters;
+  /**
+   * Parameters that were successfully applied
+   */
+  appliedParameters?: ConfigurationParameters;
+  /**
+   * Parameters that failed to apply
+   */
+  failedParameters?: ConfigurationParameters;
```
stephl3: does this API make sense for how data will flow through this component?

also curious if we might be able to DRY this up by attaching a status to each config param?

chat/suggestion-card/src/SuggestionCard/SuggestionCard.styles.ts
```diff
@@ -0,0 +1,69 @@
...
+  width: ${DIVIDER_WIDTH}px;
+`;
+
+export const getSuggetionCardWrapperStyles = (theme: Theme) => css`
+  background-color: ${color[theme].background.secondary.default};
+  border: 1px solid ${color[theme].border.secondary.default};
+  border-radius: ${borderRadius[300]}px;
+  width: 100%;
+  padding: ${spacing[300]}px;
+  gap: ${spacing[200]}px;
```
stephl3: This `gap` doesn't currently work. Need to add the following:
```suggestion
  display: flex;
  flex-direction: column;
  gap: ${spacing[200]}px;
```

At that point, you can remove the `margin-top` from `applyButtonStyles` and `bannerWrapperStyles`

chat/suggestion-card/src/StatusBanner/StatusBanner.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+
+import Banner, { Variant as BannerVariant } from '@leafygreen-ui/banner';
+
+import {
+  bannerWrapperStyles,
+  boldedTextStyle,
+  listStyles,
+} from '../SuggestionCard/SuggestionCard.styles';
```
stephl3: can you move these styles into a separate `StatusBanner.styles.ts`?

chat/suggestion-card/src/StatusBanner/StatusBanner.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export const StatusBanner = ({
+  status,
+  appliedParameters,
+  failedParameters,
+}: {
+  status: Status;
+  appliedParameters: SuggestionCardProps['appliedParameters'];
+  failedParameters: SuggestionCardProps['failedParameters'];
+}) => {
```
stephl3: There's a circular import between `StatusBanner` + `SuggestionCard` because 
- `StatusBanner` imports types defined from `SuggestionCard`
- `SuggestionCard` imports  `StatusBanner`

Can the `status` + `appliedParameters` + `failedParameters` be defined in a separate `StatusBanner.types.ts` or a separate `src/shared.types.ts` file? `ConfigurationParameters` type can also be defined there

.changeset/polite-walls-add.md
joyjwang: gotcha! will remove then!

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
joyjwang: aha i'm seeing it now great point! i must have intuitively thought there would be a case where we don't want users to apply suggestions but it looks like that's not included in the designs at all so will remove this case!

chat/suggestion-card/package.json
joyjwang: SGTM! will update the naming

chat/suggestion-card/src/SuggestionCard/SuggestionCard.types.ts
joyjwang: mmm gotcha will consolidate the parameters to be filtered by status! 

i think having the status, configurationParameters, and onClickApply make sense, we have separate utils set up in mms to handle the actual application of these configurations and handle errors, so we can dynamically update this component with these props

chat/suggestions/src/SuggestedActions/SuggestedActions.types.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface SuggestedActionsProps
+  extends HTMLElementProps<'div'>,
+    DarkModeProps {
+  /**
+   * Determines rendering of the status banner:
+   * - `'Apply'` will render suggestions and the "Apply" button
+   * - `'Success'` will render success banner with applied suggestions
+   * - `'Error'` will render error banner with instructions to manually apply suggestions
+   */
+  status: Status;
```
stephl3: Sorry, I should have surfaced this in a prior review: can we rename the prop `state`, rename the type `State` and use `Unset: 'unset'` for the default? This would match conventions in the rest of our codebase

PR #2964, opened 2025-07-18
LG-5357: progress bar customizability add-ons

charcoalyy:
## ✍️ Proposed changes

- introduce hook for automatic timed rotation between react nodes
- introduce hook to calculate animation durations based on magnitude of change and desired speed
- extend base HTML div attributes to allow consumer to pass in className
- rename `roleType` to `role` since we have full control over whether it is passed to intrinsic HTML elements, and completely restrict its internal assigned value

🎟 _Jira ticket:_ [LG-5357](https://jira.mongodb.org/browse/LG-5357)

### For new components


packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -104,8 +136,9 @@ export function ProgressBar(props: ProgressBarProps) {
     <LeafyGreenProvider darkMode={darkMode}>
       <div
         data-lgid={lgIds.root}
-        className={containerStyles}
         aria-disabled={disabled}
+        className={getContainerStyles(className)}
```
charcoalyy, line 130: consumer className styles are only applied to the outermost div, keeping the inner divs blackboxed. is this appropriate?

packages/progress-bar/src/ProgressBar/hooks/useComputedTransitionDuration.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { usePrevious } from '@leafygreen-ui/hooks';
+
+/**
+ * Computes the total transition duration for a CSS animation based on
+ * the change magnitude and a specified speed.
+ *
+ * Note: The previous value is tracked internally via a ref, so it
+ * does not need to be provided as an argument.
+ *
+ * @param param.speed - Desired change in units per second.
```
charcoalyy: used `param.paramName `prefix to distinguish between these object args from positional args in the tsdocs comments. do we have a different convention for this?

packages/progress-bar/src/ProgressBar/hooks/useComputedTransitionDuration.tsx
charcoalyy, line 1: both `useComputedTransitionDuration` and `useRotatingItems` were written deliberately to be as general as possible, i.e., not specific to `progress-bar`. does it make sense to move them to the general @hooks package, or should they live in `progress-bar` until a true separate use case presents itself? if the former, i would include separate tests for them.

packages/progress-bar/README.md
```diff
@@ -94,19 +93,19 @@ import ProgressBar from '@leafygreen-ui/progress-bar';
...
+| ------------------ | ---------------------------------------------------------------------------------------------- | --------------- | ---------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `size?`            | `'small'` \| `'default'` \| `'large'`                                                          |                 | all                                      | Optional size (thickness) of the progress bar.                                                                                                                                                              |
+| `description?`     | `React.ReactNode \| Array<React.ReactNode>`                                                    |                 | all                                      | Optional descriptive text below the progress bar. If multiple items are provided in an array, they will be automatically rotated every 2000 milliseconds. Single items are returned as-is without rotation. |
```
charcoalyy, line 99: debated making another prop to control the time interval, but i'm wondering if that's overdoing it for an option people might not even use.

packages/progress-bar/src/ProgressBar/utils/utils.tsx
```diff
@@ -123,6 +123,29 @@ export const resolveProgressBarProps = (
...
 
+/**
+ * Omits resolved props from the original props object.
+ * Useful for spreading the remaining props while excluding resolved ones.
+ *
+ * @param obj - Original props object
+ * @param resolved - Resolved props to omit
+ * @returns New object with resolved props omitted
+ */
+export const omitProps = (
```
charcoalyy: i realize we have a helper method (`pickAndOmit`) that uses the lodash functions. here, i was trying to avoid having to literally write out the array of keys literally and instead programmatically deduce it from the resolved object. any thoughts on if it's preferred to just use `pickAndOmit` anyway?

packages/progress-bar/README.md
TheSonOfThomp: A fixed 2000ms should be fine. Just make sure it's not a magic number in the code (declare it as a const)

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
TheSonOfThomp: Mainly this is for controlling positioning / spacing of the component

packages/progress-bar/src/ProgressBar/hooks/useComputedTransitionDuration.tsx
TheSonOfThomp: Unless there's an obvious case to use right now, let's keep it in this package

packages/progress-bar/src/ProgressBar/hooks/useComputedTransitionDuration.tsx
TheSonOfThomp: Typically I'd extract it to a separate interface
```ts
interface UseComputedTransitionDurationArgs {
  /** Desired change in units per second. */
  speed: number;
  /** ... */
  currentValue?: number;
  /** ... */
  minimumDuration?: number;
  /** ... */
  maximumDuration?: number;
}
```

packages/progress-bar/src/ProgressBar/utils/utils.tsx
TheSonOfThomp: I think the use of this fn could be simplified to do the `resolve` and omit in a single function call

packages/progress-bar/src/ProgressBar/hooks/useRotatingItems.tsx
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Cycles through one or more React nodes at a fixed interval.
+ *
+ * @param input - Single React node or array of nodes to rotate through.
+ * @param [interval=2000] - Time in milliseconds between rotations.
+ * @returns The currently active React node based on the rotation interval.
+ */
+export const useRotatingItems = (
+  input: React.ReactNode | Array<React.ReactNode>,
+  interval = 2000,
```
TheSonOfThomp: this should be a named constant

packages/progress-bar/src/ProgressBar/ProgressBar.types.ts
```diff
@@ -135,19 +141,30 @@ export type ProgressBarProps = BaseProps &
...
+  DeterminatePlain: 'determinate-plain',
+  DeterminateAnimated: 'determinate-animated',
+  Transition: 'indeterminate-to-determinate-transition',
+} as const;
+export type AnimationMode = (typeof AnimationMode)[keyof typeof AnimationMode];
+
+/**
+ * Internal flattened type representing all resolved props for rendering.
+ */
+export type ResolvedProgressBarProps = InheritedProps & {
```
charcoalyy, line 160: git history is a bit confusing here, but essentially, `ResolvedProgressBarProps` now "flattens" all discriminated fields of `ProgressBarProps` for rendering. i don't like how there's code duplication with minor changes (i.e., some fields become mandatory due to defaults being applied. however, i could not determine a reliable way to programmatically flatten `ProgressBarProps` due to its complexity/nested nature.

i tried to improve readability by grouping reusable props where possible and maintaining the same order of appearance in the code. very much open to suggestions here.

packages/progress-bar/src/ProgressBar/utils/utils.tsx
```diff
@@ -273,3 +193,107 @@ export const getHeaderIcon = ({
...
+  obj: ProgressBarProps,
+  omit: any = {},
+): Omit<ProgressBarProps, keyof ProgressBarProps> => {
+  const clone = { ...obj };
+
+  for (const key of Object.keys(omit) as Array<keyof ProgressBarProps>) {
+    delete clone[key as keyof ProgressBarProps];
+  }
+
+  return clone;
```
TheSonOfThomp: I think the Lodash method should work here, no? 
```suggestion
  return _.omit(obj, Object.keys(omit))
```

charcoalyy: *facepalm* yes - thank you!

PR #2965, opened 2025-07-18
refactor: use AssistantAvatar in chat components

stephl3:
## ✍️ Proposed changes

- refactor chat components to use `AssistantAvatar`

## 🧪 How to test changes

- storybook

### DIFF
--- a/.changeset/chat-assistant-avatar.md
+++ b/.changeset/chat-assistant-avatar.md
@@ -1,0 +1,6 @@
+---
+'@lg-chat/fixed-chat-window': patch
+'@lg-chat/input-bar': patch
+---
+
+Replace gray icon with `AssistantAvatar` with gradient


packages/avatar/src/AssistantAvatar/AssistantAvatar.tsx
```diff
@@ -1,36 +1,59 @@
...
-    />
-  </svg>
-));
+const DEFAULT_FILL = 'url(#sparkle-gradient)';
+
+export const AssistantAvatar: React.ComponentType<AssistantAvatarProps> =
+  createGlyphComponent(
+    'Sparkle',
+    ({ darkMode = false, disabled = false, ...rest }: AssistantAvatarProps) => {
+      const theme = getTheme(darkMode);
```
shaneeza, line 15: Any reason why you can't use `useDarkMode` here?

stephl3: I think it could be done either way without much difference. I figured since it's a presentational component, the `darkMode` value will already have been read from the context and could/should be passed down with the `disabled` value

shaneeza: I think to be consistent with our other components, we should have the component read the mode from the provider, so you don't have to explicitly pass it. However, it doesn't look like we do this with other icon components so I'm fine with either approach.

PR #2966, opened 2025-07-21
Adds autodocs support to Storybook

bruugey:
Base PR to get autodocs into the repo. Part one of migrating our docs off of .design :)

### DIFF
--- a/tools/storybook-addon/src/preview/index.tsx
+++ b/tools/storybook-addon/src/preview/index.tsx
@@ -1,12 +1,4 @@
 import React from 'react';
-
-import {
-  Title,
-  Subtitle as StorybookSubtitle,
-  Description,
-  Primary,
-  Controls,
-} from '@storybook/blocks';
 import {
   ComponentPreview,
   PropCombinations,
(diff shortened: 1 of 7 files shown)


tools/storybook-addon/src/preview/index.tsx
```diff
@@ -48,6 +56,15 @@ const parameters = {
...
     source: { type: 'code' },
+    page: () => (
+      <>
+        <Title />
+        <StorybookSubtitle />
+        <Description />
+        <Primary />
+        <Controls />
+      </>
+    ),
```
stephl3, line 67: Screenshot 2025-07-21 at 1 14 54 PM

had a couple questions:
1. is there a way to format the `@param`s?
2. I know we are fairly consistent with tsdocs for props but do we have a standard for including them for component's?some components have them while others don't, and I'm curious how we want to approach that

bruugey: This PR is just to get a baseline into our Storybook. I have a doc for refining everything, but want to get us familiar with how upcoming PRs effect the auto generated docs so we can start to play with this. Think of this as a beta.

PR #2967, opened 2025-07-21
[LG-5368] feat(chat input-bar): add compact variant and deprecate unused export

stephl3:
## ✍️ Proposed changes

- mark `lgInputBarStyles` as `@deprecated`
- refactor sub-components
- add compact variant of `InputBar`

🎟 _Jira ticket:_ [LG-5368](https://jira.mongodb.org/browse/LG-5368)

## 🧪 How to test changes

- unit tests
- storybook

### DIFF
--- a/chat/input-bar/src/InputBar/InputBar.tsx
+++ b/chat/input-bar/src/InputBar/InputBar.tsx
@@ -89,13 +89,10 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
 
     if (
       isCompact &&
-      (shouldRenderHotkeyIndicator ||
-        shouldRenderGradientProp ||
-        badgeText ||
-        dropdownFooterSlot)
+      (shouldRenderHotkeyIndicator || shouldRenderGradientProp || badgeText)
     ) {
       consoleOnce.warn(
-        `@lg-chat/input-bar: The InputBar component's props 'shouldRenderHotkeyIndicator', 'shouldRenderGradient', 'badgeText', and 'dropdownFooterSlot' are only used in the 'spacious' variant. They will not be rendered in the 'compact' variant set by the provider.`,
+        `@lg-chat/input-bar: The InputBar component's props 'shouldRenderHotkeyIndicator', 'shouldRenderGradient', and 'badgeText' are only used in the 'spacious' variant. They will not be rendered in the 'compact' variant set by the provider.`,
       );
     }
 
--- a/chat/input-bar/src/InputBar/InputBar.types.ts
+++ b/chat/input-bar/src/InputBar/InputBar.types.ts
@@ -49,13 +49,11 @@ export type InputBarProps = HTMLElementProps<'form'> &
     disableSend?: boolean;
     /**
      * Footer content displayed when a dropdown is rendered. Only renders when children exist.
-     * @remarks This prop is only considered when the parent `LeafyGreenChatProvider` has `variant="spacious"`.
      */
     dropdownFooterSlot?: ReactElement;
 
     /**
      * Props passed to the Popover that renders the suggested prompts.
-     * @remarks This prop is only considered when the parent `LeafyGreenChatProvider` has `variant="spacious"`.
      */
     dropdownProps?: Omit<
       PopoverRenderModeProps,


chat/input-bar/src/InputBar/InputBarSendButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { ReturnIcon } from './ReturnIcon';
+import { State } from './shared.types';
+
+export const InputBarSendButton = forwardRef<
+  HTMLButtonElement,
+  InputBarSendButtonProps
+>(({ disabled, isCompact, shouldRenderButtonText, state, ...rest }, fwdRef) => {
+  const { theme } = useDarkMode();
+
+  const isLoading = state === State.Loading;
```
TheSonOfThomp, line 20: is there no "loading" state for the compact button?

chat/input-bar/src/InputBar/InputBarSendButton.types.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * Whether the button is a compact or spacious variant
+   */
+  isCompact: boolean;
+
+  /**
+   * Whether the button should render its text
+   * @remarks This prop is only considered when the parent `LeafyGreenChatProvider` has `variant="spacious"`.
+   */
+  shouldRenderButtonText: boolean;
```
TheSonOfThomp, line 19: maybe over-complicating this, but we could turn this into a discriminated union

```ts
{
  isCompact: true,
  shouldRenderButtonText: false
} | {
  isCompact: false,
  shouldRenderButtonText: boolean
}
```

chat/input-bar/src/InputBar/InputBarSendButton.styles.ts
```diff
@@ -0,0 +1,76 @@
...
+const getBaseIconButtonStyles = ({ theme }: { theme: Theme }) => {
+  const darkMode = theme === Theme.Dark;
+  return css`
+    background-color: ${palette.green.dark2};
+    border: 1px solid ${palette.green[darkMode ? 'base' : 'dark2']};
+    color: ${palette.white};
+
+    &:active,
+    &:hover {
+      background-color: #00593f;
```
TheSonOfThomp: if this is intentionally off-palette, can we add a comment indicating why

TheSonOfThomp: also, is dark & light mode hover bg the same?

chat/input-bar/src/InputBar/InputBarFeedback.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <Body className={getLoadingTextStyles(theme)}>
+              MongoDB Assistant is thinking
+            </Body>
+          </div>
+        )}
+
+        {state === State.Error && (
+          <Banner variant="danger" className={bannerStyles}>
+            <div className={bannerContentContainerStyles}>
+              <Error>{errorMessage || 'Oops... Something went wrong.'}</Error>
```
TheSonOfThomp: can we break the default error message out into a constant?

chat/input-bar/src/InputBar/InputBarFeedback.styles.ts
```diff
@@ -0,0 +1,55 @@
...
+  color: ${color[theme].text[Variant.Secondary][InteractionState.Default]};
+`;
+
+const ellipsisAnimation = keyframes`
+  0% { content: ''; }
+  25% { content: '.'; }
+  50% { content: '..'; }
+  75% { content: '...'; }
+  100% { content: '...'; }
+`;
```
TheSonOfThomp, line 34: Nice! 🤘

chat/input-bar/src/InputBar/InputBar.tsx
```diff
@@ -34,55 +35,77 @@ import {
...
+        isCompact &&
+        (shouldRenderHotkeyIndicator ||
+          shouldRenderGradientProp ||
+          badgeText ||
+          dropdownFooterSlot)
+      ) {
+        consoleOnce.warn(
+          `@lg-chat/input-bar: The InputBar component's props 'shouldRenderHotkeyIndicator', 'shouldRenderGradient', 'badgeText', and 'dropdownFooterSlot' are only used in the 'spacious' variant. They will not be rendered in the 'compact' variant set by the provider.`,
+        );
+      }
```
TheSonOfThomp: 👍

chat/input-bar/src/InputBar/InputBar.tsx
```diff
@@ -374,81 +403,94 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
...
+     * so the user can retry sending the message.
+     */
+    useEffect(() => {
+      if (!isError) {
+        return;
+      }
+
+      if (!isControlled) {
+        updateValue(prevMessageBody, textareaRef);
+        setPrevMessageBody('');
```
TheSonOfThomp, line 413: > so the user can retry sending the message

Are we sure we want to reset the message then? This makes me think we should keep it as-is, so they can try the same message again

chat/input-bar/src/InputBar/InputBar.tsx
```diff
@@ -374,81 +403,94 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
...
+              <div
+                className={getContentWrapperStyles({
+                  disabled,
+                  isCompact,
+                  isFocused,
+                  theme,
+                })}
+              >
+                {!isCompact && (
+                  <div className={adornmentContainerStyles}>
```
TheSonOfThomp, line 448: > adornment

So fancy 😅

chat/input-bar/src/InputBar/InputBar.tsx
```diff
@@ -374,81 +403,94 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
...
 
+    /**
+     * When the state is 'error', we want to reset the message body and focus the textarea
+     * so the user can retry sending the message.
+     */
+    useEffect(() => {
+      if (!isError) {
+        return;
+      }
+
```
TheSonOfThomp: Wanted to say "don't use an effect", but state can be set externally so we'll have to do it that way.

We should check `state` against a `prevState` though, otherwise this'll reset every time `prevMessageBody` changes

chat/input-bar/src/InputBar/InputBar.styles.ts
```diff
@@ -1,171 +1,50 @@
...
   transitionDuration,
   typeScales,
+  Variant,
 } from '@leafygreen-ui/tokens';
 
-export const baseStyles = css`
-  width: 100%;
-  position: relative;
-`;
+const ADORNMENT_CONTAINER_HEIGHT = 36;
```
TheSonOfThomp, line 23: can we just define what "adornment" means here 😅 Not opposed to the word, just want to make sure it's understandable to everyone who reads the code

chat/input-bar/src/InputBar/InputBarFeedback.tsx
```diff
@@ -0,0 +1,68 @@
...
+        aria-relevant="all"
+        className={statusContainerStyles}
+        ref={fwdRef}
+        {...rest}
+      >
+        {state === State.Loading && (
+          <div className={loadingContainerStyles}>
+            <AssistantAvatar darkMode={darkMode} size={20} />
+            <Body className={getLoadingTextStyles(theme)}>
+              MongoDB Assistant is thinking
```
TheSonOfThomp: Should prob store this message as constant too

TheSonOfThomp (commented):
Just the `useEffect` and a few stylistic nits

chat/input-bar/src/InputBar/InputBarSendButton.tsx
stephl3: There is for compact but not for spacious. Added a `consoleOnce.warn` for that case

chat/input-bar/src/InputBar/InputBarSendButton.types.ts
stephl3: When I discussed this with design, we opted not to include the variant in the props API to simplify the API so `variant` is always set by the provider and to prevent possible mixing and matching of compact / spacious

chat/input-bar/src/InputBar/InputBarSendButton.styles.ts
stephl3: Oops forgot to refactor this. This is the green value between dark2 and dark3 that is used for the primary variant of LG Button. yes, it's the same for dark & light mode

chat/input-bar/src/InputBar/InputBar.tsx
stephl3: I think this is doing what you're thinking, but maybe my comment led you astray. When the user submits the message, the input clears. This effect resets to the previous message. Updated the comment to make that clearer. Let me know if you're aligned there :)

chat/input-bar/src/InputBar/InputBar.tsx
stephl3: good callout! added

chat/input-bar/src/InputBar/InputBarSendButton.styles.ts
```diff
@@ -0,0 +1,82 @@
+import { css, cx } from '@leafygreen-ui/emotion';
+import { Theme } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+import { color, InteractionState, Variant } from '@leafygreen-ui/tokens';
+
+/**
+ * Off-palette value specific to primary button instances
+ * @see TODO: https://jira.mongodb.org/browse/LG-5388
+ */
+const GREEN_BETWEEN_DARK2_DARK3 = '#00593F';
```
stephl3: This hexcolor is used in the primary button variant: https://github.com/mongodb/leafygreen-ui/blob/main/packages/button/src/Button/Button.styles.ts#L73-L89

The primary button and this component should share the same variable for this color, possibly as a token. I don't want this work to be blocked while we figure out with the broader team where makes the most sense. For now, I've spun up this ticket to address a better place for this color

chat/input-bar/src/InputBar/InputBar.tsx
TheSonOfThomp: Oh I think I misread the code the first time. We're setting the current value to the prev. value, and then resetting the _prev_ message. 

I'm aligned

chat/input-bar/src/InputBar/InputBarFeedback.tsx
```diff
@@ -0,0 +1,72 @@
...
+  bannerContentContainerStyles,
+  bannerStyles,
+  getLoadingTextStyles,
+  loadingContainerStyles,
+  statusContainerStyles,
+} from './InputBarFeedback.styles';
+import { InputBarFeedbackProps } from './InputBarFeedback.types';
+import { State } from './shared.types';
+
+const texts = {
```
TheSonOfThomp: nit: `messages`?

packages/button/src/Button/index.ts
```diff
@@ -1 +1,2 @@
 export { Button } from './Button';
+export { PRIMARY_BUTTON_INTERACTIVE_GREEN } from './Button.styles';
```
TheSonOfThomp, line 2: can you make this var name longer 😜

packages/button/package.json
```diff
@@ -25,7 +25,8 @@
       "require": "./dist/umd/testing/index.js",
       "import": "./dist/esm/testing/index.js",
       "types": "./dist/types/testing/index.d.ts"
-    }
+    },
+    "./Button": "./src/Button/index.ts"
```
TheSonOfThomp: I wonder if we should be explicitly listing the `.styles` file as the export?
```ts
import { PRIMARY_BUTTON_INTERACTIVE_GREEN } from '@leafygreen-ui/button/Button.styles';
```

As it is, I'd be confused why I can't do:
```ts
import { Button } from '@leafygreen-ui/button/Button';
```

TheSonOfThomp (approved):

packages/button/package.json
stephl3: good call, I'm down for that!

PR #2969, opened 2025-07-22
[LG-5376] feat(chat message-rating): add compact variant and deprecate unused exports

stephl3:
## ✍️ Proposed changes

- mark unused exports of `lgMessageRatingStyles`, `lgRadioButtonStyles`, `RadioButton`, and `RadioButtonProps` as `@deprecated`
- refactor sub-components
- add compact variant of `MessageRating`

🎟 _Jira ticket:_ [LG-5376](https://jira.mongodb.org/browse/LG-5376)

## 🧪 How to test changes

- unit tests
- storybook

### DIFF
--- a/chat/message-rating/src/MessageRating.stories.tsx
+++ b/chat/message-rating/src/MessageRating.stories.tsx
@@ -87,6 +87,11 @@ const Template: StoryFn<MessageRatingStoryProps> = ({ variant, ...props }) => (
 
 export const LiveExample: StoryObj<MessageRatingStoryProps> = {
   render: Template,
+  parameters: {
+    chromatic: {
+      disableSnapshot: true,
+    },
+  },
 };
 
 export const Controlled: StoryFn<typeof MessageRating> = ({


chat/message-rating/src/MessageRating.spec.tsx
```diff
@@ -7,92 +11,144 @@ import { MessageRatingProps } from './MessageRating/MessageRating.types';
...
-      render(<MessageRating onChange={onChange} />);
-      const thumbsUp = screen.getByLabelText('Thumbs up this message');
-      fireEvent.click(thumbsUp);
-      expect(onChange).toHaveBeenCalledWith(
-        expect.objectContaining({
-          target: expect.objectContaining({ value: 'liked' }),
-        }),
-      );
+    test('description is not rendered when variant is compact', () => {
+      renderMessageRating({ onChange }, Variant.Compact);
```
shaneeza: Should we pass a description to `renderMessageRating` to demonstrate that even if you pass one, it will not render?

chat/message-rating/src/MessageRating.spec.tsx
```diff
@@ -7,92 +11,144 @@ import { MessageRatingProps } from './MessageRating/MessageRating.types';
...
-      );
-    });
+  describe.each([Variant.Compact, Variant.Spacious])(
+    'in variant: %s',
+    variant => {
+      describe('uncontrolled', () => {
+        test('onChange receives change event when thumbs up button is clicked', () => {
+          renderMessageRating({ onChange }, variant);
+          const thumbsUp = screen.getByLabelText('Thumbs up this message');
+          fireEvent.click(thumbsUp);
```
shaneeza: I know you didn't add this, but should we switch to `userEvent`?

chat/message-rating/src/MessageRating.spec.tsx
```diff
@@ -7,92 +11,144 @@ import { MessageRatingProps } from './MessageRating/MessageRating.types';
 const onChange = jest.fn();
```
shaneeza, line 12: Even though everything is passing as expected in this file, I don't think we should globally define `onChange`. It could lead to some unexpected results because the state might persist across all tests. I remember running into issues with this with the `Select` component. I would recommend defining this in individual tests.

chat/message-rating/src/MessageRating.spec.tsx
```diff
@@ -7,92 +11,144 @@ import { MessageRatingProps } from './MessageRating/MessageRating.types';
...
+          fireEvent.click(thumbsUp);
+          expect(onChange).toHaveBeenCalledWith(
+            expect.objectContaining({
+              target: expect.objectContaining({ value: 'liked' }),
+            }),
+          );
+        });
+      });
+    },
+  );
```
shaneeza, line 170: Should we also test `unselected`?

chat/message-rating/src/MessageRating.spec.tsx
```diff
@@ -7,92 +11,144 @@ import { MessageRatingProps } from './MessageRating/MessageRating.types';
 const onChange = jest.fn();
 const descriptionText = 'How was the response?';
 
+// Mock the ResizeObserver
+global.ResizeObserver = jest.fn().mockImplementation(() => ({
```
shaneeza, line 16: Why do we need this mock for this component?

chat/message-rating/src/MessageRating.stories.tsx
```diff
@@ -54,3 +109,10 @@ export const Controlled: StoryFn<typeof MessageRating> = ({
 Controlled.argTypes = {
   value: { control: 'none' },
 };
+
+export const Generated: StoryObj<MessageRatingStoryProps> = {
+  render: Template,
+  args: {
+    variant: Variant.Compact,
```
shaneeza, line 121: Are there generated stories for the `spacious` variant?

chat/message-rating/src/MessageRating/MessageRating.styles.ts
```diff
@@ -1,35 +1,60 @@
...
+export const getIconFill = ({
+  darkMode,
+  isSelected,
+}: {
+  darkMode: boolean;
+  isSelected: boolean;
+}) => (isSelected ? getSelectedIconFill(darkMode) : getBaseIconFill(darkMode));
+
+const baseActiveStyles = css`
+  &:before {
```
shaneeza:
```suggestion
  &::before {
```

nit: `before` is a pseudo-element, so we should use double brackets `::before`. [Double brackets distinguish pseudo-elements from pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements).

chat/message-rating/src/MessageRating/MessageRating.tsx
```diff
@@ -1,86 +1,181 @@
...
-            </RadioButton>
+        <div className={getContainerStyles(className)} {...rest} ref={ref}>
+          {!isCompact && <Description>{description}</Description>}
+          <div
+            aria-label="Message rating"
+            className={buttonContainerStyles}
+            role="radiogroup"
+          >
+            {isCompact ? (
+              <IconButton
```
shaneeza, line 108: Apologies for asking again if we have already discussed this, but what's the reason for switching to `IconButton` again?

chat/message-rating/src/MessageRating/MessageRating.types.ts
```diff
@@ -12,20 +12,18 @@ export type MessageRatingValue =
...
     DarkModeProps {
-  /**
-   * Determines the currently selected value of the radio buttons.
-   * @default undefined
-   */
-  value?: MessageRatingValue;
   /**
    * Custom description text
    * @default "How was the response?"
+   * @remarks This prop is only considered when the parent `LeafyGreenChatProvider` has `variant="spacious"`.
```
shaneeza, line 18: TIL about `remarks`!

chat/message-rating/src/MessageRating.spec.tsx
stephl3: updated to clear all mocks after each test and added tests to ensure `onChange` is called the expected number of times

chat/message-rating/src/MessageRating.spec.tsx
stephl3: `ResizeObserver` gets used in `LeafyGreenChatProvider`, so it needs to be mocked if I render it in the specs

chat/message-rating/src/MessageRating.stories.tsx
stephl3: No, I only added stories for the net-new variant for now

chat/message-rating/src/MessageRating/MessageRating.tsx
stephl3: No worries, they effectively work the same. `IconButton` already has a lot of the styling done. Much easier to use the existing one to keep consistent rather than custom add to the custom `RadioButton` component

chat/message-rating/src/MessageRating.spec.tsx
shaneeza: @stephl3

stephl3: Added a test for `unselected`

chat/message-rating/src/MessageRating.stories.tsx
```diff
@@ -23,16 +65,29 @@ export default {
...
-const Template: StoryFn<typeof MessageRating> = props => (
-  <MessageRating {...props} />
+const Template: StoryFn<MessageRatingStoryProps> = ({ variant, ...props }) => (
+  <LeafyGreenChatProvider variant={variant}>
+    <MessageRating {...props} />
+  </LeafyGreenChatProvider>
 );
 
-export const Basic: StoryFn<typeof MessageRating> = Template.bind({});
+export const LiveExample: StoryObj<MessageRatingStoryProps> = {
```
shaneeza, line 88: Should we disable snapshots for LiveExample?

PR #2970, opened 2025-07-22
Beta testing IA changes in Storybook

bruugey:
Hailey wants to see changes to IA in Storybook. Reflected in this PR.


packages/drawer/src/Drawer.stories.tsx
```diff
@@ -34,7 +34,7 @@ const snapshotStoryExcludedControlParams = [
 ];
 
 export default {
-  title: 'Components/Drawer',
+  title: 'Sections/Drawer',
```
stephl3, line 37: There are a couple more `*.interactions.stories.tsx` files for drawer that need to be updated as well

PR #2973, opened 2025-07-23
Clean commit  to update descriptions

bruugey:
Clean commit to add descriptions. Sorry for the re-r @tsck


packages/text-area/src/TextArea/TextArea.tsx
```diff
@@ -35,8 +34,6 @@ import { State, TextAreaProps } from './TextArea.types';
  * @param props.baseFontSize Override the global `baseFontSize` set in LeafygreenProvider. This will only change the font size of the input text, not the label or description.
  * @param props.defaultValue The default value of the input field. Unlike value, component will not be controlled if defaultValue is passed.
  */
-
-type TextArea = React.ForwardRefExoticComponent<TextAreaProps>;
```
tsck, line 39: Same confirmation that this was intentional? 🙂

bruugey: Intentional yes we just moved it up in the same file

PR #2979, opened 2025-07-24
LG-5361: Resizable hook

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-5361](https://jira.mongodb.org/browse/LG-5361)

## Not Included
- documentation (will add in a separate PR)

### For new components


packages/hooks/src/index.ts
```diff
@@ -16,6 +16,7 @@ export { default as useMutationObserver } from './useMutationObserver';
 export { default as useObjectDependency } from './useObjectDependency';
 export { default as usePoller } from './usePoller';
 export { default as usePrevious } from './usePrevious';
+export { Position, useResizable } from './useResizable/';
```
stephl3: Hello! First off, amazing work on the `useResizable` hook and the related utils! The way you've separated out logic for `calculateNewSize` and `getResizerAriaAttributes` sets up a really solid foundation.

After continuing research following the initial eng spec and reviewing usage of `useResizable` in `Drawer`, I've uncovered deeper context around best practices and component architecture that makes me think we should evolve the original plan. Curious to get your thoughts on it before we go further.

My main concern with the current "hook-only" or "mixin" approach is the long-term impact it will have on components that consume `useResizable` (`Drawer`, `SideNav`, `ContextDrawer`). A couple architectural challenges:
- prop bloat & unclear API: Components will need to accept props related to configuring resizing (`minWidth`, `maxWidth`, `onResize`, etc.) which will make their APIs more complex
- single responsibility principle: Components forced to take on a second job of managing resize logic and rendering the handle

Instead, I'm wondering if we use industry-standard patterns to go beyond a hook only and provide a dedicated collection of components. These could be packaged in a `@leafygreen-ui/resizable` package and export:
- `ResizablePanelGroup`
- `ResizablePanel`
- `ResizableHandle`

```tsx
<ResizablePanelGroup direction="horizontal">
  <ResizablePanel defaultSize={75}>
    <MainContent />
  </ResizablePanel>
  <ResizableHandle />
  <ResizablePanel defaultSize={25}>
    <Drawer />
  </ResizablePanel>
</ResizablePanelGroup>
```

It's clean/composable, separates concerns, and easier to guarantee accessibility since can be built into those components. You've done a lot of the heavy lifting already, and I think this will set us up to handle resizing in a more robust, scalable, and maintainable way that aligns with modern best practices.

What are your thoughts on this direction? I'm happy to chat through it and sketch out an API for these new components. Thanks again for your great work on this!

shaneeza: Hii, thank you for the helpful detailed feedback. I appreciate your research and I think this is a very valid and important architectural concerns. 

To me, there is a distinction between this internal hook and a public-facing component API. Currently, our immediate needs are for components like Drawer, ContextDrawer, and even SideNav and CloudNav.

Here are my thoughts on why `useResizable` is more suitable for our internal use cases:

- Purpose: I 100% agree that if we are designing a new, public-facing resizable component, then your proposed component-based APi would be the ideal pattern. However, `useResiazable` is more of a low-level, internal utility hook. Its purpose is not to dictate rendering or layout for external consumers but to provide raw size data to our internal components.
- Addressing pop bloat and SRP for internal users: I think these are valid concerns about prop bloat and the single responsibility principle and I think that they are necessary for a public API.
    - The context of an internal hook is a little different. In the case of a drawer, the consumer cannot override the min and max width. I assume this would be the case for similar components. So we’re not adding bloat to the public API; we’re adding internal configurations to a component.
    - And I don’t see this as the component is now forced to manage the resizing logic. The logic is encapsulated within the useResizable hook and component just uses this tool to get the right size.
    - You are, however, correct that we have to render the handle component. I don't see this as a big issue. It's very minimal markup, and it gives drawer full control over where the handle should be placed.
- Maintaining existing and complex layouts and transition: This is I think the most significant challange with a component based approach. Components like `CloudNav`, and `SideNav`, are unique with custom grid layouts, intricate primary and secondary nav structures, and specific overlay/embedded behaviors. They are not generic panels. Introducing a resizable group panel would require a re-architecture of those components. We would have to rip out the existing grid layout logic and force it into a potentially generic layout. Then apply style overrides to try and create the custom behavior. In contrast, the `useResizable` hook is agnostic and allows us to integrate resizing functionality into existing components with minimal disruption. We can keep the current layouts and their existing transition logic and bind the hooks size output to their styles.
- Timeline: Given our current timeline, I don’t think we should move to a resizable component just yet. This would require the creation of the resizable component, re-architecting Drawer and testing and we only have one week left.

Again, I think that a public-facing resiazable component is valuable and important long-term for LeafyGreen. I think that we should proceed with the hook for our immediate needs but consider a resizable component in the future.

Would you like to discuss this more with the team?

packages/hooks/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@leafygreen-ui/lib": "workspace:^",
+    "@leafygreen-ui/palette": "workspace:^",
```
stephl3: when this is moved to another package, we can consume `theme` and use the color token for border instead of the palette value

packages/hooks/src/useResizable/useResizable.constants.ts
```diff
@@ -0,0 +1,28 @@
+import { keyMap } from '@leafygreen-ui/lib';
+
+import { Position, SizeGrowth } from './useResizable.types';
+
+// Mappings for keyboard interactions based on the position
+export const SIZE_GROWTH_KEY_MAPPINGS: Record<
+  Position,
+  { [key: string]: SizeGrowth }
```
stephl3: should we restrict key to being ArrowLeft or ArrowRight?

packages/hooks/src/useResizable/useResizable.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+    enabled: true,
+    initialSize: 300,
+    minSize: 200,
+    maxSize: 600,
+    maxViewportPercentages: 50,
+    position,
+  });
+
+  const resizerProps = getResizerProps();
+  const isVertical = position === 'left' || position === 'right';
```
stephl3: nit:
```suggestion
  const isVertical = position === Position.Left || position === Position.Right;
```

packages/hooks/src/useResizable/useResizable.utils.ts
```diff
@@ -0,0 +1,109 @@
+import { css, cx } from '@leafygreen-ui/emotion';
+import { palette } from '@leafygreen-ui/palette';
+
+import { RESIZER_SIZE } from './useResizable.constants';
+import { Position } from './useResizable.types';
+
+/**
+ * Calculates the new size based on mouse position and constraints
+ */
+export const calculateNewSize = (
```
stephl3: can we add tests for this util?

packages/hooks/src/useResizable/useResizable.types.ts
```diff
@@ -0,0 +1,91 @@
...
+   * This can be used to perform any actions based on the new size.
+   */
+  onResize?: (size: number) => void;
+
+  /**
+   * The percentage of the viewport that the resizable element should occupy at maximum.
+   * This can be used to ensure the element does not exceed a certain size relative to the viewport.
+   * This percentage-based maximum will override the maxSize prop if the calculated pixel value is smaller.
+   */
+  maxViewportPercentages: number;
```
stephl3: had a couple questions on this:
1. can this be consolidated into the `maxSize` prop?
2. unsure about this specifically being scoped to the viewport vs the parent container. why do we bake in the viewport assumption?

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+ * It allows resizing the component using mouse and keyboard interactions.
+ *
+ * @param {ResizableProps} props - The properties for the resizable functionality.
+ * @returns {ResizableReturn} - The state and methods for the resizable component.
+ */
+export const useResizable = <T extends HTMLElement = HTMLDivElement>({
+  enabled = true,
+  initialSize = 0,
+  minSize: minSizeProp = 0,
+  maxSize: maxSizeProp = 0,
```
stephl3: are these the correct default values?

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+  // Ref to hold the current value of isResizing to prevent stale closures in event handlers
+  const isResizingRef = useRef<boolean>(isResizing);
+  const [size, setSize] = useState<number>(initialSize);
+  const minSize = minSizeProp ?? initialSize;
+  const maxSize = maxSizeProp ?? initialSize;
+  const isVertical = position === Position.Left || position === Position.Right;
+
+  // Keeps track of all the sizes that can be used for resizing with keyboard
+  const keyboardSizes = [initialSize, minSize, maxSize];
+  const sortedKeyboardSizes = [...keyboardSizes].sort((a, b) => a - b);
```
stephl3: how were these determined? curious if there should be more granular steps? also, should this be able to reach 0 for cases like closing a drawer?

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+          event.code === keyMap.ArrowLeft ||
+          event.code === keyMap.ArrowRight
+        ) {
+          event.preventDefault();
+        }
+      } else {
+        if (event.code === keyMap.ArrowUp || event.code === keyMap.ArrowDown) {
+          event.preventDefault();
+        }
+      }
```
stephl3: This looks more appropriately placed directly in `handleKeyDown` before `getNextKeyboardSize` is called

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+      const getNextSize = (sizeGrowth: SizeGrowth) => {
+        const currentSize = size;
+        const sizes = sortedKeyboardSizes;
+
+        if (sizeGrowth === SizeGrowth.Increase) {
+          return sizes.find(size => size > currentSize);
+        } else {
+          return [...sizes].reverse().find(size => size < currentSize);
+        }
+      };
```
stephl3: can we move this to utils and add tests?

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+
+      const handleSizeChange = (nextSize: number | undefined) => {
+        if (nextSize !== undefined) {
+          setSize(nextSize);
+          onResize?.(nextSize);
+        }
+      };
```
stephl3: can we define this function outside of `getNextKeyboardSize`?

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+    (event: React.KeyboardEvent | KeyboardEvent) => {
+      getNextKeyboardSize(event, position);
+    },
+    [getNextKeyboardSize, position],
+  );
+
+  /**
+   * Handles mouse down event for resizing
+   */
+  const handleOnMouseDown = useCallback(
```
stephl3:
```suggestion
  const handleMouseDown = useCallback(
```

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+          : resizableRef.current.offsetHeight;
+      }
+    },
+    [enabled, isVertical],
+  );
+
+  /**
+   * Handle focus event for the resizer
+   */
+  const handleOnFocus = useCallback(() => {
```
stephl3:
```suggestion
  const handleFocus = useCallback(() => {
```

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+    setIsFocused(true);
+  }, []);
+
+  /**
+   * Handle blur event for the resizer
+   */
+  const handleOnBlur = useCallback(() => {
```
stephl3:
```suggestion
  const handleBlur = useCallback(() => {
```

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+    handleOnMouseDown,
+    handleOnFocus,
+    handleOnBlur,
+    isResizing,
+    isVertical,
+  ]);
+
+  // Effect hook to add and remove global mouse event listeners
+  // These listeners are added to 'window' to ensure dragging works even if the mouse
+  // moves off the resizer handle during the drag.
```
stephl3: can we keep the comment pattern used above consistent for the below effects?
```suggestion
  /**
   * Effect hook to add and remove global mouse event listeners
   * These listeners are added to 'window' to ensure dragging works even if the mouse
   * moves off the resizer handle during the drag.
   */
```

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,293 @@
...
+      window.addEventListener('mousemove', handleMouseMove);
+      window.addEventListener('mouseup', handleMouseUp);
+    } else {
+      cleanupListeners();
+    }
+
+    return () => {
+      cleanupListeners();
+    };
+  }, [isResizing, handleMouseMove, handleMouseUp, enabled]);
```
stephl3: I think this is a cleaner/more idiomatic way of writing this effect. Similar could be applied to the below effects
```suggestion
  useEffect(() => {
    if (!enabled || !isResizing) {
    return;
    }
    
    window.addEventListener('mousemove', handleMouseMove);
    window.addEventListener('mouseup', handleMouseUp);

    return () => {
      window.removeEventListener('mousemove', handleMouseMove);
      window.removeEventListener('mouseup', handleMouseUp);
    };
  }, [enabled, isResizing, handleMouseMove, handleMouseUp]);
```

packages/hooks/src/index.ts
shaneeza: Update: We discussed as a team, and we're going to continue with the hook approach. The main reason is that the component approach would require a component re-architecture for existing components. 

A con to this approach is that this is in the hooks package, and a major change to one hook affects all hooks.

packages/hooks/src/useResizable/useResizable.types.ts
shaneeza: 1. Could you elaborate on your first question? 
2. This is a good callout. Currently, the resizing is based on the window as the parent. I think it should be based on a specified parent container with the window as the default. That means there would need to be another prop for the container and `calculateNewSize` would need a little refactoring to ensure the correct size is returned.

packages/hooks/src/useResizable/useResizable.ts
shaneeza: No, the defaults should be the initial size. Will update.

packages/hooks/src/useResizable/useResizable.ts
shaneeza: There aren't any specific guidelines on what the steps should be. [This WCAG article on keyboard touches on this a little](https://www.w3.org/WAI/WCAG20/Understanding/keyboard). As long as a change is visible then that should be good so I decided to go with the min, max, and initial numbers for now. 

In the case of drawer, the min width is around 320, so you shouldn't be able to resize the drawer to 0. If you want the element to resize all the way to 0 then the minSize should be set to 0.

That said, maybe only allowing these 3 values is not enough and does not offer enough fine-tuned control. This is how gemini interprets the article above:

The "Too Large" Problem (Precision & Control):

WCAG Connection: While not explicitly a single success criterion, the overall goal of operability implies that users should have sufficient control over interactive elements. For controls that adjust a range of values (like a size), precision is often needed.

Interpretation: If the steps are very large (e.g., 100px at a time), a user might easily jump past their desired size. They might want a width of 450px, but the control only allows 400px or 500px.

Impact: This leads to:

- Lack of Fine-Grained Control: The user cannot reach specific, intermediate values.

- Frustration: They can't achieve their exact desired state.

- Reduced Usability: The control becomes less effective for tasks requiring accuracy.

Perhaps this should change to maybe a 50px increment/decrement within the min and max width.

packages/hooks/src/useResizable/useResizable.types.ts
stephl3: for the first question, I'm curious why we have a separate prop to handle resizing relative to the parent or viewport. I think `maxSize` supporting both cases would be a simpler/cleaner API

packages/hooks/src/useResizable/useResizable.ts
stephl3: Yes, I agree that equal, incremental values might provide a better experience for users although I'm unsure what the increments should be. If it hasn't been discussed with design already, I think it'd be valuable to discuss with them.

Regarding closing the drawer, I'm not sure I agree that `minSize` determines if you can close the drawer. As a user, I would expect `minSize` to be the last step before closing to match what can be done with the mouse

shaneeza: You can't resize the drawer closed; you can only resize it down to the min size with a mouse. This was confirmed with Sooa, but since we don't support resizing a closed drawer open, you also shouldn't be able to resize an open drawer closed. So in this case, setting the keyboard min size as `minSize` matches the mouse behavior.

packages/hooks/src/useResizable/useResizable.types.ts
shaneeza: `maxSize` and `minSize` are expecting pixel values since we're expecting the resizable element to be a fixed value and not a percentage value. 

If the max size of the drawer is 500px, that could potentially take up more than 50% of the container, which, in the case of the drawer, we don't want the drawer to take up more than 50% of the viewport. We could avoid the percentage prop by setting a `max-width` directly in the element. So in drawer that would be `max-width: 50vw`

stephl3: Gotcha, what do you think about moving the `max-width: 50vw` out into `Drawer` directly while preserving the current `maxSize`/`minSize` expectation of pixel values? This way, `useResizable` is simpler and doesn't require tracking its state in relation to the viewport which might be specific to `Drawer` but not as much to other resizable components

packages/hooks/src/useResizable/useResizable.ts
stephl3: Gotcha, I wasn't aware resizing could not close. I guess we can start with that for now and find out more from user behavior

Would still be curious about using more fine-grained control for resizing

packages/hooks/src/useResizable/useResizable.types.ts
shaneeza: I think this is a good solution! I'll remove `maxViewportPercentages` from the hook. Thank you.

packages/hooks/src/useResizable/useResizable.ts
shaneeza: I spoke with Sooa, and she's okay with the current keyboard approach. However, she's also OK with us making it more fine-grained. I suggested a 50px increment/decrement, and she was on board.

packages/hooks/src/index.ts
stephl3: is this still getting moved into its own package? I thought that was what we landed on in the last discussion with the team?

packages/hooks/src/useResizable/useResizable.utils.ts
```diff
@@ -0,0 +1,54 @@
+import { css, cx } from '@leafygreen-ui/emotion';
```
stephl3: at this point, I would think it's fine to move these utils into the `utils` subdir

packages/hooks/src/useResizable/useResizable.ts
```diff
@@ -0,0 +1,300 @@
...
+   * For example:
+   * - If position is 'left' and the left arrow key is pressed, it decreases the size.
+   * - If position is 'right' and the left arrow key is pressed, it increases the size.
+   */
+  const setNextKeyboardSize = useCallback(
+    (event: React.KeyboardEvent | KeyboardEvent, position: Position | null) => {
+      if (position && event.code in SIZE_GROWTH_KEY_MAPPINGS[position]) {
+        const sizeGrowth =
+          SIZE_GROWTH_KEY_MAPPINGS[position][event.code as Arrow];
+        const nextSize = getNextKeyboardSize({
```
stephl3: Seeing some strange behavior when using arrow keys to resize on smaller screens. If you resize with arrow keys on smaller viewports in a shrinking direction, it seems to overshoot at the end and then continuing 1 tick further in the same direction will push the size back to what I think is the `minSize`

https://github.com/user-attachments/assets/2b0aa68a-9439-44ed-a9a0-3de64994688f

packages/hooks/src/index.ts
shaneeza: I don't believe so. I thought we landed on keeping this as an internal hook inside the hooks package. I can bring this up with the team again to confirm.

shaneeza: Update: A new package will be created, `leafygreen-ui/resizable` because it's cleaner for future maintenance to have separate packages for specific functionality like this. We also have other "internal" hooks that are their own packages(`descendants`, `input-item`).

PR #2980, opened 2025-07-24
LG-5375: Resizable drawer

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-5375](https://jira.mongodb.org/browse/LG-5375)

## Changes 
- Adds `resizable` prop which supports a resizable embedded drawer
- Updates `DrawerLayout` to accept `drawer` prop
- Adds `DrawerLayoutContext`, which stores the values for`isDrawerOpen`, `displayMode`, `resizable`, and `onClose`. These values are read inside `Drawer`.

## TODO
- Update docs and tests (separate PR)
- Fix flaky toolbar story (separate PR)

### For new components


packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -129,20 +131,38 @@ const getOverlayStyles = ({
...
   );
 
+const getEmbeddedStyles = ({ open, size }: { open: boolean; size: number }) =>
+  cx(
+    css`
+      transition: width ${drawerTransitionDuration}ms linear;
+    `,
+    {
+      [css`
+        width: ${size}px;
```
shaneeza: I'm considering reading this value from a CSS variable over a size prop so a new class is not created every time the size prop is updated.

packages/drawer/src/Drawer.stories.tsx
```diff
@@ -117,9 +119,8 @@ const TemplateComponent: StoryFn<DrawerProps> = ({
   const renderDrawer = () => (
     <Drawer
       {...rest}
-      displayMode={displayMode}
-      open={open}
       onClose={() => setOpen(false)}
+      open={undefined} // Prevent open from passing since it is passed to DrawerLayout
```
TheSonOfThomp: do we need to do this explicitly, or can we just leave it unset? AFAI can see it's not being spread into the component by `...rest`, so we don't need to override it

packages/drawer/src/DrawerLayout/DrawerLayout.types.ts
```diff
@@ -2,45 +2,90 @@ import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
...
+   *
+   */
+  displayMode?: 'overlay';
+
+  /**
+   * Determines if the drawer is resizable. This is only recommended for individual drawers, not stacked drawers.
+   *
+   * @defaultValue false
+   */
+  resizable?: never;
```
TheSonOfThomp, line 48: do we want `never` or `false` here? 1. Wouldn't the `?` mean this resolves to `undefined | never` => `undefined`? 2. If we _do_ want `never` we should change the `@default` annotation

packages/drawer/src/DrawerLayout/DrawerLayout.types.ts
```diff
@@ -2,45 +2,90 @@ import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
...
+export type DrawerLayoutPropsWithToolbar = {
+  /**
+   * An array of data that will be used to render the toolbar items and the drawer content.
+   */
+  toolbarData: DrawerToolbarLayoutProps['toolbarData'];
+
   /**
    * Determines if the Drawer is open. This is only needed if using the Drawer without a toolbar. This will shift the layout to the right by the width of the drawer if `displayMode` is set to 'embedded'.
    */
   isDrawerOpen?: never;
```
TheSonOfThomp, line 82: same `never` comment here (I do think `never` is correct in this case though)

packages/drawer/src/LayoutComponent/LayoutComponent.types.ts
```diff
@@ -1,24 +1,17 @@
...
 
-// This interface is used to define the common properties for OverlayDrawerLayout and EmbeddedDrawerLayout
-export interface BaseLayoutComponentProps
-  extends Omit<HTMLElementProps<'div'>, 'children'> {
   /**
-   * Determines if the Toolbar is present in the layout
+   * Slot prop for the drawer
    */
-  hasToolbar?: boolean;
+  drawer?: React.ReactNode;
```
TheSonOfThomp: should these be inherited from somewhere?

packages/drawer/src/OverlayDrawerLayout/OverlayDrawerLayout.types.ts
```diff
@@ -1,3 +1,5 @@
-import { BaseLayoutComponentProps } from '../LayoutComponent/LayoutComponent.types';
+import { LayoutComponentProps } from '../LayoutComponent/LayoutComponent.types';
 
-export type OverlayDrawerLayoutProps = BaseLayoutComponentProps;
+export type OverlayDrawerLayoutProps = Omit<LayoutComponentProps, 'drawer'> & {
+  isDrawerOpen?: never;
+};
```
TheSonOfThomp: what's the benefit to `Omit`ing the `drawer` prop, but then setting `isDrawerOpen` to `never`? Should the both either be `Omit`ed or `never`ed?

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
TheSonOfThomp, line 1: Can we use a different name other than `Container` 😅

TheSonOfThomp: Maybe just `Content`?

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -58,18 +57,9 @@ export const DrawerToolbarLayoutContainer = forwardRef<
...
-        hasToolbar={hasData}
-        isDrawerOpen={isDrawerOpen}
-      >
+      <LayoutComponent {...rest} ref={forwardRef} isDrawerOpen={isDrawerOpen}>
         <div className={contentStyles}>{children}</div>
-        <DrawerWithToolbarWrapper
-          displayMode={displayMode}
-          isDrawerOpen={isDrawerOpen}
-        >
+        <DrawerWithToolbarWrapper isDrawerOpen={isDrawerOpen}>
```
TheSonOfThomp: Why do both `LayoutComponent` and `DrawerWithToolbarWrapper` need the `isDrawerOpen` prop? Can those just both read from context?

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout.stories.tsx
TheSonOfThomp, line 1: super nit: if there's a lot of boilerplate/constants it might be easier/cleaner to create a 'testUtils` file

packages/drawer/src/DrawerWithToolbarWrapper/DrawerWithToolbarWrapper.tsx
TheSonOfThomp, line 1: For components that are exclusively used in other components, can we nest the directories to make this easier to parse?

packages/drawer/src/DrawerWithToolbarWrapper/DrawerWithToolbarWrapper.tsx
```diff
@@ -17,16 +19,12 @@ export const DrawerWithToolbarWrapper = forwardRef<
   DrawerWithToolbarWrapperProps
 >(
   (
-    {
-      children,
-      className,
-      isDrawerOpen,
-      displayMode,
-    }: DrawerWithToolbarWrapperProps,
+    { children, className, isDrawerOpen }: DrawerWithToolbarWrapperProps,
```
TheSonOfThomp: can we pull `isDrawerOpen` from context instead?

packages/drawer/src/DrawerWithToolbarWrapper/DrawerWithToolbarWrapper.styles.ts
```diff
@@ -94,7 +117,7 @@ const getDrawerShadowStyles = ({ theme }: { theme: Theme }) => css`
   &::before {
     transition-property: opacity;
     transition-duration: ${drawerTransitionDuration}ms;
-    transition-timing-function: ease-in-out;
+    transition-timing-function: linear;
```
TheSonOfThomp: Most of our animations are `eased`. Why is this one linear?

packages/drawer/src/DrawerWithToolbarWrapper/DrawerWithToolbarWrapper.styles.ts
```diff
@@ -153,19 +182,26 @@ export const getDrawerWithToolbarWrapperStyles = ({
...
-  isDrawerOpen: boolean;
+  isDrawerOpen?: boolean;
   shouldAnimate?: boolean;
-  displayMode: DisplayMode;
+  displayMode?: DisplayMode;
   theme: Theme;
-}) =>
-  cx(
+}) => {
+  const isOverLay = displayMode === DisplayMode.Overlay;
```
TheSonOfThomp:
```suggestion
  const isOverlay = displayMode === DisplayMode.Overlay;
```

packages/drawer/src/constants.ts
```diff
@@ -1,3 +1,4 @@
+import { TOOLBAR_WIDTH } from '@leafygreen-ui/toolbar';
```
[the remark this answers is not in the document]
TheSonOfThomp: Do we need to re-export this. Copilot is wrong, but I don't think this needs to be re-exported, does it?

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -38,24 +50,50 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+    } = useDrawerLayoutContext();
     const [shouldAnimate, setShouldAnimate] = useState(false);
     const ref = useRef<HTMLDialogElement | HTMLDivElement>(null);
-    const drawerRef = useMergeRefs([fwdRef, ref]);
+
+    // Prop overrides
+    const displayMode =
+      displayModeProp ?? displayModeContextProp ?? DisplayMode.Overlay;
+    const open = openProp ?? isDrawerOpen ?? false;
+    const onClose = onCloseProp ?? onCloseContextProp;
```
TheSonOfThomp: could these be passed into the `useContext` hook and have the prop resolution happen there?

TheSonOfThomp: or create a new hook `useResolvedDrawerProps` that merges the component props, context props, and defaults

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -114,6 +152,31 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+      position: Position.Right,
+    });
+
+    // Create merged ref after resizableRef is defined
+    const refsToMerge = [fwdRef, ref];
+
+    // Use a conditional to ensure resizableRef is only included when it's needed
+    if (isEmbedded && resizable) {
+      refsToMerge.push(resizableRef);
+    }
```
TheSonOfThomp: is it wrong if we just add the undefined `resizableRef` to the array without this check?

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -126,6 +189,7 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
             className,
             displayMode,
             zIndex: 1000 + drawerIndex,
+            size: resizable ? drawerSize : open ? initialSize : 0,
```
TheSonOfThomp: nit: maybe make this a small function?

packages/drawer/src/DrawerLayout/DrawerLayout.types.ts
shaneeza: I think it should be `never`. Currently, only an embedded drawer is resizable, and the resizable prop changes nothing with an overlay drawer, so I don't think you should be able to pass a boolean. In the case of an overlay drawer, you have to exclude the prop. Do you see anything wrong with that pattern?

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: I can change it, but what's bad about `Container`?

packages/drawer/src/DrawerWithToolbarWrapper/DrawerWithToolbarWrapper.styles.ts
shaneeza: Ooops, forgot to revert this. I changed it while testing.

packages/drawer/src/constants.ts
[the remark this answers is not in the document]
shaneeza: removed

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: `isDrawerOpen` is internally set when there is a toolbar. So the value returned from the context would not be correct.

However, I think I found a good workaround: return a setter function from the context so I can update `isDrawerOpen` anywhere, read the updated value from the context, and avoid prop drilling.

packages/drawer/src/DrawerWithToolbarWrapper/DrawerWithToolbarWrapper.tsx
shaneeza: I'll do this in a separate PR

packages/drawer/src/OverlayDrawerLayout/OverlayDrawerLayout.types.ts
shaneeza: refactored to remove `isDrawerOpen`

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout.stories.tsx
shaneeza: i'll update this in a separate PR

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
shaneeza: I'll address this in a separate PR

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
```diff
@@ -40,8 +38,9 @@ export const DrawerToolbarLayoutContainer = forwardRef<
     const { openDrawer, closeDrawer, getActiveDrawerContent, isDrawerOpen } =
       useDrawerToolbarContext();
     const { id, title, content } = getActiveDrawerContent() || {};
+    const { onClose, displayMode } = useDrawerLayoutContext();
+
```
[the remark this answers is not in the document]
shaneeza: ?

shaneeza:
@TheSonOfThomp I ended up refactoring Drawer so that the updated size is passed to the layout context and read from the layout component. This way, grid-template-column is updated rather than the drawer width.

packages/drawer/src/DrawerLayout/DrawerLayout.types.ts
TheSonOfThomp: thinking about this again, it's probably fine to leave it as `never`. There might be issues though if someone is dynamically changing the `displayMode`, and the string literal isn't know

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayoutContainer.tsx
TheSonOfThomp: Having `Wrapper` and `Container` is a bit confusing

PR #2984, opened 2025-07-28
LG-5389: progress bar QA improvements

charcoalyy:
## ✍️ Proposed changes

* use CSS variables instead of directly passing in variable to emotion, to prevent rebuilding styles on every change
* avoid coupling tests with implementation details
* explicitly guard against non-animated variants being passed into indeterminate bars

🎟 _Jira ticket:_ [LG-5389](https://jira.mongodb.org/browse/LG-5389)

### For new components


tools/install/src/ALL_PACKAGES.ts
```diff
@@ -49,11 +50,13 @@ export const ALL_PACKAGES = [
   '@leafygreen-ui/polymorphic',
   '@leafygreen-ui/popover',
   '@leafygreen-ui/portal',
+  '@leafygreen-ui/preview-card',
   '@leafygreen-ui/progress-bar',
   '@leafygreen-ui/radio-box-group',
   '@leafygreen-ui/radio-group',
   '@leafygreen-ui/ripple',
   '@leafygreen-ui/search-input',
+  '@leafygreen-ui/section-nav',
```
TheSonOfThomp: something you might be able to tackle once you're done here... ensure these get added in the correct "version packages" commit, not here

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -179,12 +179,19 @@ export function ProgressBar(props: ProgressBarProps) {
           >
             <div
               data-lgid={lgIds.fill}
+              style={
+                {
+                  // to avoid rebuilding styles on every width change, use CSS variables
```
TheSonOfThomp: technically it's to avoid adding additional emotion style classes

TheSonOfThomp: > use CSS variables

why? to do what?
(just a documentation nit)

packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -179,12 +179,19 @@ export function ProgressBar(props: ProgressBarProps) {
...
+                  ...(!isIndeterminate && {
+                    '--width': `${displayWidth}%`,
+                    '--width-animation-duration': `${widthAnimationDuration}ms`,
+                  }),
+                } as React.CSSProperties
```
TheSonOfThomp: nit: I feel like this might be cleaner if we defined the styles as a var outside the render block

packages/progress-bar/src/ProgressBar/ProgressBar.styles.ts
```diff
@@ -166,18 +165,14 @@ const getDeterminateFillStyles = ({
...
   variant: Variant;
   disabled?: boolean;
-  width: number;
-  widthAnimationDuration?: number;
 }) => css`
-  width: ${width}%;
-  transition: width
-    ${widthAnimationDuration || DEFAULT_WIDTH_ANIMATION_DURATION}ms ease;
+  width: var(--width);
+  transition: var(--width-animation-duration) ease;
```
TheSonOfThomp, line 175: Nit, since CSS vars are hard to trace, can we document here what element they're being defined on

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -148,50 +140,6 @@ describe('packages/progress-bar', () => {
         expect(queryIcon()).toBeNull();
       });
     });
-
-    describe('with unexpected input', () => {
-      test('renders width 0% when value is less than 0', () => {
```
TheSonOfThomp, line 153: should we add some of these back as stories?

charcoalyy: since these edge cases are util-specific (i.e., resolved at props util, then propagated to rendering), i figured the util tests were sufficiently appropriate and avoids story clutter.

PR #2989, opened 2025-07-29
[LG-5391] feat(chat-window): add compact variant

stephl3:
## ✍️ Proposed changes

- add compact variant of `ChatWindow`

🎟 _Jira ticket:_ [LG-5391](https://jira.mongodb.org/browse/LG-5391)

## 🧪 How to test changes

- storybook

### DIFF
--- a/.changeset/chat-compact-chat-window.md
+++ b/.changeset/chat-compact-chat-window.md
@@ -1,0 +1,5 @@
+---
+'@lg-chat/chat-window': minor
+---
+
+[LG-5391](https://jira.mongodb.org/browse/LG-5391): add variant support for when the parent `LeafyGreenChatProvider` has `variant="compact"`


chat/chat-window/src/ChatWindow/ChatWindowContents.tsx
```diff
@@ -33,19 +33,22 @@ export const ChatWindowContents = forwardRef(
     ref: ForwardedRef<HTMLDivElement>,
   ) => {
     const { theme } = useDarkMode(darkModeProp);
-    const { containerWidth } = useLeafyGreenChatContext();
+    const { containerWidth, variant } = useLeafyGreenChatContext();
+
+    const isCompact = variant === Variant.Compact;
+    const isMobile = !!containerWidth && containerWidth < breakpoints.Tablet;
```
TheSonOfThomp, line 39: Should we just return `isMobileSize` from the context? 
(Also nit that got called out in the nav a while back: technically it's `isMobileSize` not `isMobile` since we're not checking the UserAgent or input devices)

stephl3: That's a good callout, and I wish I made this change with the addition of `variant`. Adding `isMobileSize` to the context now will trigger a major change in chat components that rely on it, and I'd like to avoid that for now. I think we can revisit once we determine with design how we want to approach mobile responsiveness in the compact set of components

PR #2991, opened 2025-07-30
LG-3077: compare icon checksums against prior builds

charcoalyy:
## ✍️ Proposed changes

- uses checksums generated during prebuild to detect source code changes and _only_ rebuild stale icons, shrinking the best case build time for `@leafygreen-ui/icon` to <30s
- introduces GitHub action steps to restore old icon builds _even_ if source code has changed, since icon's build script can detect changes, shaving ~60s off CI

🎟 _Jira ticket:_ [LG-3077](https://jira.mongodb.org/browse/LG-3077)


.github/workflows/pr.yml
```diff
@@ -36,6 +36,19 @@ jobs:
...
+      # @leafygreen-ui/icon has an additional built-in cache layer
+      - uses: actions/cache/restore@v4
+        name: Restore previous icon build
+        if: ${{ steps.build-cache.outputs.cache-hit != 'true' }}
+        id: icon-build-cache
+        with:
+          path: |
+            packages/icon/dist
+          key: icon-dist-${{ github.ref }}-${{ github.run_id }}
+          # using the latest build in this branch, icon build script detects changes and only rebuilds what's stale
```
charcoalyy, line 48: how do we feel about this from a CI perspective? essentially, icon is _able_ to work with stale/outdated build files because of the new checksum logic, but this means we don't start off a clean "slate" every time. 

alternatively, we could have CI run the same as before and exclude this change, which means checksums will only help build time in local development.

packages/icon/scripts/compare-checksum.ts
```diff
@@ -0,0 +1,70 @@
...
+/** Gets list of changed icons by comparing checksums. */
+export function getChangedChecksums(): Array<string> {
+  const prevChecksums = getPrevChecksums();
+  const newChecksums: Record<string, string> = {};
+
+  const changes: Array<string> = [];
+
+  for (const iconName of iconNames) {
+    const filePath = path.join(ICON_DIR, `${iconName}.tsx`);
+    const checksum = extractChecksumFromFile(filePath);
```
TheSonOfThomp: Could probably speed this up a bit if we read the file first and then check the string. Right now we're reading the file each iteration

packages/icon/scripts/build-parallel.ts
```diff
@@ -98,8 +91,10 @@ async function buildAllBatches(
 
 async function main(): Promise<void> {
```
TheSonOfThomp: I _believe_ this would be better called as a callback inside the `Commander` program than as a main function when the file is imported

packages/icon/scripts/compare-checksum.ts
```diff
@@ -0,0 +1,70 @@
...
+      changes.push(iconName);
+    }
+  }
+
+  if (changes.length > 0) {
+    console.log(
+      'Checksum comparison complete. Changes include:',
+      changes.join(', '),
+    );
+  }
```
TheSonOfThomp: Don't think we need to log here. I think logging in the build step would be a bit more useful (e.g. "Building icons: X, Y, Z...")

.github/workflows/pr.yml
```diff
@@ -36,6 +36,19 @@ jobs:
...
 
+      # @leafygreen-ui/icon has an additional built-in cache layer
+      - uses: actions/cache/restore@v4
+        name: Restore previous icon build
+        if: ${{ steps.build-cache.outputs.cache-hit != 'true' }}
+        id: icon-build-cache
+        with:
+          path: |
+            packages/icon/dist
+          key: icon-dist-${{ github.ref }}-${{ github.run_id }}
```
TheSonOfThomp, line 47: I don't think this cache key will work. AFAIK the `github.run_id` changes each time, no? (unless I'm confusing it with something else)

TheSonOfThomp (commented):
Couple code-organization notes:
- I noticed `DELIMITER` is defined in 2 separate files. We should DRY that and have a constants file
- I wonder if it'd make sense to have `prebuild` `build` and `postbuild` directories in scripts to keep things a bit more manageable

.github/workflows/pr.yml
charcoalyy: correct - `restore-keys` below will find the next closest key (i.e., by matching prefix). so theoretically, the previous build.

packages/icon/scripts/compare-checksum.ts
charcoalyy: hmm think i'm a little confused as to what this means - per file per icon is needed because the checksums are stored per file.

packages/icon/package.json
```diff
@@ -7,9 +7,9 @@
   "types": "./dist/types/index.d.ts",
   "typesVersions": {},
   "scripts": {
-    "prebuild": "ts-node ./scripts/prebuild.ts",
-    "postbuild": "ts-node ./scripts/postbuild.ts",
-    "build": "ts-node ./scripts/build-parallel.ts",
+    "prebuild": "ts-node ./scripts/prebuild/prebuild.ts",
+    "postbuild": "ts-node ./scripts/postbuild/postbuild.ts",
+    "build": "ts-node ./scripts/build/build-parallel.ts",
```
TheSonOfThomp: technically could make them `index.ts` files at this point. As-is works though—makes it easier to understand what file you have open

packages/icon/scripts/build/build-parallel.ts
```diff
@@ -80,31 +72,39 @@ async function buildBatch(batch: Array<string>): Promise<void> {
...
     console.error('Build failed:', err);
     process.exit(1);
   }
 }
 
-main();
+new Command()
+  .description('Split icon files into batches for bundling in parallel')
+  .option('-v, --verbose', 'Enable verbose output', false)
+  .action(main)
```
TheSonOfThomp: nit: I might rename `main` to something a bit more clear

packages/icon/scripts/build/build-parallel.ts
```diff
@@ -80,31 +72,39 @@ async function buildBatch(batch: Array<string>): Promise<void> {
  * build all batches in parallel. Waits until all batches have finished.
  */
 async function buildAllBatches(
+  fullBatch: Array<string>,
   batchSize = 10,
   numWorkers = 4,
-  fullBatch: Array<string>,
+  verbose = false,
 ): Promise<void> {
```
TheSonOfThomp: My general rule-of-thumb is if a function takes more than 3 args, turn it into an `options` object arg.

```suggestion
async function buildAllBatches({
  fullBatch,
  batchSize = 10,
  numWorkers = 4,
  verbose = false,
}: {
  fullBatch: Array<string>,
  batchSize: number,
  numWorkers: number,
  verbose?: boolean,
}
): Promise<void> {
```

packages/icon/package.json
charcoalyy: ooh, sweet. i think i do actually prefer the `index.ts` as it makes clear what's the "main" script per folder.

packages/icon/scripts/build/compare-checksum.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Loads checksums from the .checksums.json file as an object.
+ * If it doesn't exist, returns an empty object.
+ */
+function getPrevChecksums(): Record<string, string> {
+  try {
+    return JSON.parse(fs.readFileSync(CHECKSUMS_FILE, 'utf-8'));
+  } catch {
+    return {};
```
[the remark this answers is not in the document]
TheSonOfThomp: ^ This could be helpful for debugging?

PR #2992, opened 2025-07-30
[LG-5379] perf(CodeEditor): Reduce bundle size by 31% with lazy module loading

tsck:
### Background
During initial planning, we implemented lazy loading for language modules to reduce bundle size. However, subsequent analysis revealed that even with lazy language loading, the overall bundle remained too large due to eager loading of the core `@uiw/react-codemirror` wrapper and all its dependencies.

### Details
This PR achieves a **31% reduction in parsed bundle size**, decreasing from **~610 KB to ~422 KB**. This optimization includes the addition of `@leafygreen-ui/typography` for improved loading state presentation, which increases the size slightly.

#### Key Changes

**Architecture Improvements:**
- **New lazy loading system**: Introduced `useModuleLoaders()` and `useLazyModules()` hooks that create a unified, type-safe mechanism for dynamically importing only the modules needed based on editor configuration
- **Migration to native CodeMirror**: Replaced `@uiw/react-codemirror` wrapper with direct CodeMirror v6 packages for granular control over module loading
- **Extension hooks library**: Created reusable hooks for CodeMirror extensions (`useLanguageExtension`, `useThemeExtension`, etc.) that encapsulate functionality and can be used by consumers

**Bundle Optimization:**
- Language-specific modules now load only when that language is selected
- Core extensions (autocomplete, lint, etc.) load conditionally based on enabled features
- Eliminated unnecessary wrapper dependencies

**Testing & Documentation:**
- Updated test utilities to handle asynchronous editor initialization
- Enhanced documentation with lazy loading behavior details

🎟 _Jira ticket:_ [LG-5379](https://jira.mongodb.org/browse/LG-5379)

### For new components
- [ ] I have verified the Live Example looks as intended on the design website

## 🧪 How to test changes
- **Manual QA**: Test all language modes and features in Storybook
- **Functionality**: Ensure all unit and regression tests pass
- **Integration**: Test lazy loading behavior with different prop combinations


packages/code-editor/src/CodeEditor/CodeEditor.types.ts
```diff
@@ -108,6 +98,13 @@ export interface CodeEditorTooltip {
 }
 
 export interface CodeEditorProps extends DarkModeProps {
+  /**
+   * Font size of text in the editor.
+   *
+   * @default 14
+   */
+  baseFontSize?: BaseFontSize;
```
tsck: Drive by: I had forgotten to add this prop the first time around so added it in passing

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
```diff
@@ -216,9 +224,23 @@ export interface CodeEditorProps extends DarkModeProps {
    */
   tooltips?: Array<CodeEditorTooltip>;
 
+  /**
+   * Controlled value of the editor. If set, the editor will not be editable
+   * and will not update its value on change. Use `onChange` to update the
+   * value externally.
+   */
+  value?: string;
```
tsck, line 238: Drive by: with the React wrapper we were unable to have a true controlled value. Without it, we can, so I added this in passing.

packages/code-editor/src/CodeEditor/CodeEditor.styles.ts
```diff
@@ -9,43 +16,103 @@ export const getEditorStyles = ({
...
+    min-width: ${minWidth || 'none'};
+    height: ${height || '100%'};
+    max-height: ${maxHeight || 'none'};
+    /**
+     * The editor being rendered depends on a lazy loaded module, so it has no 
+     * height until it loads. By default, its height expands to fit its 
+     * content, therefore we won't know its actual height until it loads.
+     * To ensure the loader is visible, we need to set an arbitrary min height.
+     */
+    min-height: ${minHeight || '234px'};
```
TheSonOfThomp, line 109: should this use a token?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -1,250 +1,278 @@
...
+    });
 
-    const createIndentExtension = useCallback(
-      (unit: IndentUnits, size: number) => {
-        let indentString: string;
+    const codeFoldingExtension = useCodeFoldingExtension({
+      editorViewInstance: editorViewRef.current,
+      props,
+      modules,
+    });
```
TheSonOfThomp: nit: Would it make sense to have a `useExtensions` hook that returns an array of `Extension`? Really just away to abstract these ~80 lines into a single hook

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -1,250 +1,278 @@
...
+      forceParsingProp,
+      autoCompleteExtension,
+      lineNumbersExtension,
+      readOnlyExtension,
+      placeholderExtension,
+    ]);
+
+    useImperativeHandle(forwardedRef, () => ({
+      getEditorViewInstance: () => editorViewRef.current,
+    }));
```
TheSonOfThomp, line 137: Nice! Never actually seen `useImperativeHandle` in the wild

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -1,250 +1,278 @@
...
+              minWidth,
+              maxWidth,
+              height,
+              minHeight,
+              maxHeight,
+            })}
+          >
+            <Body className={getLoadingTextStyles(theme)}>
+              Loading code editor...
+            </Body>
```
TheSonOfThomp, line 167: I don't remember where we landed—should this use Skeleton or LoadingIndicator?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -1,250 +1,278 @@
...
-         * list because it automates updating on theme change.
-         */
-        theme={[
-          createThemeExtension(theme, baseFontSize),
-          createHighlightExtension(theme),
-        ]}
-        extensions={[
+      const domNode = editorContainerRef.current;
+
+      const editor = (editorViewRef.current = new EditorView.EditorView({
```
TheSonOfThomp: This `const x = (y = z)` pattern is a bit tough to parse (tbh I forgot you could do this). Can we simplify this into 2 declarations/assignments?

TheSonOfThomp: (or just reference `editorViewRef.current` below instead of `editor`?)

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -1,250 +1,278 @@
...
+        const Language = modules?.['@codemirror/language'];
+        const docLength = editor.state.doc.length;
+
+        if (Language && Language.forceParsing && docLength > 0) {
+          Language.forceParsing(editor, docLength, 150);
+        }
+      }
+
+      return () => {
+        delete (domNode as any)._cm;
```
TheSonOfThomp: I'm not really clear on what this is doing

packages/code-editor/src/index.ts
```diff
@@ -1,17 +1 @@
...
-  type CodeMirrorState,
-  type CodeMirrorView,
-  createCodeMirrorHighlightExtension,
-  createCodeMirrorLanguageExtension,
-  createCodeMirrorThemeExtension,
-  createCodeMirrorTooltipsExtension,
-  IndentUnits,
-  LanguageName,
-} from './CodeEditor';
+export * from './CodeEditor';
```
TheSonOfThomp: just a note on this, I find it easier to debug when I explicitly name all the exports (in case I forgot to export something from the `index` one level down)

packages/code-editor/src/CodeEditor/testing/testUtils.tsx
```diff
@@ -24,7 +64,8 @@ function getBySelector(
   selector: CodeEditorSelectors,
   options?: { text?: string },
 ) {
-  const elements = editorView.dom.querySelectorAll(selector);
+  const view = ensureEditorView();
+  const elements = view.dom.querySelectorAll(selector);
```
TheSonOfThomp, line 68: Future consideration: Might be useful to have `find...` utils that use `waitForEditorView` under the hood

packages/code-editor/src/CodeEditor/hooks/useLazyModules.ts
```diff
@@ -0,0 +1,132 @@
...
+        }, {} as Partial<T>);
+
+        setModules(prev => ({ ...prev, ...newModules }));
+      } catch (err: any) {
+        console.error(`Failed to load module:`, err);
+      } finally {
+        setIsLoading(false);
+      }
+    })();
+  }, [loaders, modules]);
```
TheSonOfThomp: This effect makes a call to `setModules` which will trigger it to run again. The early return will prevent the infinite loop, but this is a bit of a code-smell.

I'd recommend replacing the `modules` state with a reducer, and moving the bulk of this pre-computation logic into that reducer. Then all the effect needs to do is dispatch an action when `loaders` changes

packages/code-editor/src/CodeEditor/hooks/extensions/useExtension.ts
```diff
@@ -0,0 +1,60 @@
...
+  useEffect(() => {
+    if (compartment && editorViewInstance) {
+      /**
+       * This will reconfigure the compartment with the new extension
+       * whenever the value changes.
+       */
+      editorViewInstance.dispatch({
+        effects: compartment.reconfigure(extension),
+      });
+    }
```
TheSonOfThomp: do we need to remove anything from `editorViewInstance` when the effect returns?

TheSonOfThomp (commented):
Overall really great improvement!

packages/code-editor/src/CodeEditor/CodeEditor.styles.ts
tsck: Hmm I'm not sure. I actually need to speak with Sandy about this as well because the 234px is arbitrary and honestly pretty random. Is there a specific token you had in mind here?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: I did consider this as well. I think it would make sense, I'll throw that change up in a bit

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: This is deleting the CodeMirror node. I'll be honest, I only added this because I saw that they do it in Compass. I'm not actually sure it's necessary but I added a comment explaining what this is.

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: This is a good point. Not actually sure this is adding any value. Will update.

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: We decided not to use a skeleton because the editor could be empty. Designs do still show a loading indicator but I'm hesitant to add it do to the Lottie issues.

packages/code-editor/src/CodeEditor/hooks/useLazyModules.ts
tsck: Great call out, will update

packages/code-editor/src/CodeEditor/hooks/extensions/useExtension.ts
tsck: This is a good point. I'm not sure it's necessary, but we could reset any extension configurations. I'll add that because I don't think it hurts.

packages/code-editor/README.md
```diff
@@ -383,6 +383,46 @@ const myExtension = useExtension({
...
+const allExtensions = useExtensions({
+  editorViewInstance,
+  props: {
+    language: LanguageName.javascript,
+    enableLineNumbers: true,
+    enableCodeFolding: true,
+    readOnly: false,
+    // ... other props
+  },
+  modules: lazyLoadedModules,
```
TheSonOfThomp: nit: is it worth showing how to get `lazyLoadedModules` in this example?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
TheSonOfThomp: Is there a way of explicitly typing `domNode` so that TS knows it will have a `._cm` attribute?

packages/code-editor/README.md
tsck: Can't hurt, added it!

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Added a type for this!

packages/code-editor/src/CodeEditor/hooks/useLazyModules.ts
```diff
@@ -51,38 +106,29 @@ export type LoadersMap<T> = { [K in keyof T]: LoaderFn<T[K]> };
 export const useLazyModules = <T extends object>(
   loaders: LoadersMap<Partial<T>>,
 ) => {
-  const [modules, setModules] = useState<Partial<T>>({});
-  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [state, dispatch] = useReducer(lazyModulesReducer<T>, {
```
TheSonOfThomp, line 109: nit: I usually destructure the state here for readability 
```suggestion
  const [{
    isLoading,
    keysToLoad,
    modules
  }, dispatch] = useReducer(lazyModulesReducer<T>, {
```

packages/code-editor/src/index.ts
```diff
@@ -1 +1,28 @@
...
+  CodeEditorSelectors,
+  type CodeEditorTooltip,
+  CodeEditorTooltipSeverity,
+  type CodeMirrorExtension,
+  type CodeMirrorState,
+  type CodeMirrorView,
+  IndentUnits,
+  LanguageName,
+  useAutoCompleteExtension,
+  useCodeFoldingExtension,
```
TheSonOfThomp, line 14: do we need to export all these hooks?

packages/code-editor/src/CodeEditor/CodeEditor.styles.ts
TheSonOfThomp: as you said, `234` is pretty arbitrary. IMO it should be a sum of some `spacing` tokens (maybe `256` or `192`?)

packages/code-editor/src/index.ts
tsck: I think we do. The only ones I would consider unnecessary potentially would be `useLazyModules`, `useModuleLoaders`, and possible `useExtensions`. The remainder are required in order for a consumer to also use the extensions we are using under the hood. 

Are there any in particular that you're concerned about?

tsck: Since this is going into the integration branch, I'm going to merge and will followup with another PR if we decide to remove some of these

PR #2993, opened 2025-07-31
[LG-5372] feat(message-actions): MessageActions component part 1

stephl3:
## ✍️ Proposed changes

This is part 1 of the `MessageActions` component. Most of the core features are implemented here.

Part 2 will cover the copy button functionality. PR for part 2 is here: https://github.com/mongodb/leafygreen-ui/pull/3001

🎟 _Jira ticket:_ [LG-5372](https://jira.mongodb.org/browse/LG-5372)

### For new components

## 🧪 How to test changes

- unit tests
- storybook

### DIFF
--- a/chat/message-actions/README.md
+++ b/chat/message-actions/README.md
@@ -30,6 +30,7 @@ import {
   LeafyGreenChatProvider,
   Variant,
 } from '@lg-chat/leafygreen-chat-provider';
+import { Message } from '@lg-chat/message';
 import { MessageActions } from '@lg-chat/message-actions';
 import { MessageRatingValue } from '@lg-chat/message-rating';
 
@@ -62,12 +63,14 @@ const Example = () => {
 
   return (
     <LeafyGreenChatProvider variant={Variant.Compact}>
-      <MessageActions
-        onClickCopy={handleCopy}
-        onClickRetry={handleRetry}
-        onRatingChange={handleRatingChange}
-        onSubmitFeedback={handleSubmitFeedback}
-      />
+      <Message messageBody="Test message">
+        <MessageActions
+          onClickCopy={handleCopy}
+          onClickRetry={handleRetry}
+          onRatingChange={handleRatingChange}
+          onSubmitFeedback={handleSubmitFeedback}
+        />
+      </Message>
     </LeafyGreenChatProvider>
   );
 };
@@ -97,6 +100,10 @@ The `MessageActions` component provides a comprehensive set of actions for chat 
   - **Retry Button**: Renders when `onClickRetry` is provided
   - **Rating Buttons**: Renders when `onRatingChange` is provided
   - **Feedback Form**: Shows when a rating is selected and `onSubmitFeedback` is provided
+
+### Context Integration
+
+The component relies on `MessageContext` to access the message body for copy functionality.
 
 ### Rating and Feedback Flow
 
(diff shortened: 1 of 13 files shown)


chat/message-actions/src/MessageActions/MessageActions.tsx
```diff
@@ -0,0 +1,166 @@
...
+      setFeedback(e.target.value);
+    },
+    [isSubmitted],
+  );
+
+  const handleFeedbackSubmit = useCallback(
+    (e: FormEvent<HTMLFormElement>) => {
+      if (rating === MessageRatingValue.Unselected || !feedback) {
+        return;
+      }
```
tsck, line 112: Should we surface this? Right now it fails silently but curious if we should be indicating why submission failed to the user

chat/message-actions/src/MessageActions/MessageActions.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+      onSubmitFeedback?.(e, { rating, feedback });
+      setIsSubmitted(true);
+    },
+    [feedback, onSubmitFeedback, rating],
+  );
+
+  const handleCloseFeedback = useCallback(
+    (e: MouseEvent<HTMLButtonElement>) => {
+      setRating(MessageRatingValue.Unselected);
```
tsck, line 122: Q: Any reason why we don't reset all state props and just reset the rating on close?

chat/message-actions/src/MessageActions/MessageActions.styles.ts
```diff
@@ -0,0 +1,58 @@
...
+import {
+  color,
+  InteractionState,
+  spacing,
+  Variant,
+} from '@leafygreen-ui/tokens';
+
+const dividerSizes = {
+  height: 23,
+  width: 1,
```
tsck: Q: where do these values come from? The answer might just be "by design" but curious if we should add a comment here just to clarify

chat/message-actions/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Examples
+
+```tsx
+import {
+  LeafyGreenChatProvider,
+  Variant,
+} from '@lg-chat/leafygreen-chat-provider';
+import { MessageActions } from '@lg-chat/message-actions';
+
```
tsck, line 36: Nit: This may be intentional but just wanted to not that there are a few missing imports 

```tsx
import { FormEvent } from 'react';
import { MessageRatingValue } from '@lg-chat/message-rating';
```

chat/message-actions/src/MessageActions/MessageActions.tsx
tsck: Actually just reading the README, it states "All state is reset when the feedback form is closed" so this may be incorrect currently?

chat/message-actions/src/MessageActions/MessageActions.tsx
stephl3: This intentionally fails silently so as not to overload the user with required action on a low-importance task. I will add a comment about this here

chat/message-actions/src/MessageActions/MessageActions.tsx
stephl3: good catch! updating

chat/message-actions/src/MessageActions/MessageActions.tsx
tsck: Gotcha. It still seems bad to me that a user could submit a form, it could fail, and they'd have no idea, even on an unimportant task. I won't be a hard block on this, but I'm not sure I agree with failing silently here.

stephl3: I've shared this feedback with design. Will revisit and iterate on this if there's an alternative approach we land on

tsck (approved):

PR #2994, opened 2025-07-31
LG-5401 drawer size prop

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-5401](https://jira.mongodb.org/browse/LG-5401)

This PR adds the `size` prop to `Drawer`:
- `Default` is `432px`. 
- `Large` is `520px`

This PR also partially updates the README, but full README updates will be included in a separate PR.

## Not included
- Changeset (All drawer changesets and README updates will be added in a separate PR)

### For new components


packages/drawer/src/DrawerLayout/DrawerLayout.types.ts
charcoalyy, line 1: (out of curiosity) where is this used?

packages/drawer/src/Drawer.stories.tsx
charcoalyy, line 1: (also out of curiosity) is there a reason why we separate these into individual stories instead of using `generate`?

packages/drawer/src/index.ts
```diff
@@ -3,6 +3,7 @@ export {
   Drawer,
   drawerClassName,
   type DrawerProps,
+  type Size,
```
stephl3:
```suggestion
 Size,
```

packages/drawer/src/constants.ts
```diff
@@ -1,15 +1,20 @@
...
+import { Size } from './Drawer/Drawer.types';
+
 export const GRID_AREA = {
   drawer: 'drawer',
   content: 'content',
   toolbar: 'toolbar',
   innerDrawer: 'inner-drawer',
 };
 
+export const DRAWER_WIDTH: Record<Size, number> = {
```
stephl3: nit
```suggestion
export const drawerWidths: Record<Size, number> = {
```

packages/drawer/src/Drawer/Drawer.utils.ts
```diff
@@ -1,12 +1,24 @@
-import { DRAWER_WIDTH, DRAWER_WITH_TOOLBAR_WIDTH } from '../constants';
+import { DRAWER_TOOLBAR_WIDTH, DRAWER_WIDTH } from '../constants';
```
stephl3: curious why there is a constants.ts and a Drawer/Drawer.constants.ts?

packages/drawer/src/Drawer.stories.tsx
stephl3: `Drawer` takes up a lot of real estate on the screen, so it's helpful to check for visual regression with 1 drawer per snapshot

packages/drawer/README.md
```diff
@@ -355,6 +355,7 @@ const MultipleDrawersComponent = () => {
...
+| `size`                     | `'default'` \| `'large'`                     | The size of the `Drawer`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | `'default'` |
```
stephl3: worth specifying what widths the size maps to?

packages/drawer/src/DrawerLayout/DrawerLayout.types.ts
shaneeza: `DrawerLayoutProps` is used in `DrawerLayout`, and `DrawerLayout` provides the appropriate layout for a drawer. The layout is determined by the displayMode and whether or not there is toolbarData.

packages/drawer/src/Drawer/Drawer.utils.ts
shaneeza: I think we initially had Drawer.constants.ts, but I added .constant.ts since these constants were used outside of the Drawer directory. It will probably be clearer to add all the constants to the global constants.ts file

packages/drawer/src/constants.ts
shaneeza: I used `DRAWER_WIDTH` because the values are all static. This matches what we have in the STYLEGUIDE:

> Use UPPERCASE for static constants, and refactor to a shared file if needed.

Does that not apply to this?

sooachung (approved):

packages/hooks/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@leafygreen-ui/lib": "workspace:^",
+    "@leafygreen-ui/palette": "workspace:^",
```
stephl3: is this needed in this PR?

packages/drawer/src/constants.ts
stephl3: I threw this in here because I previously received feedback in a PR that we should use camelCase for objects/arrays, but that may have been wrong.

After asking Gemini about it, it had this to say which I think makes more sense:
```
## Objects and Arrays
For objects and arrays, the convention is more nuanced:

If the object or array is truly immutable (meaning its contents will never be changed), use SCREAMING_SNAKE_CASE. This is often paired with Object.freeze() to enforce immutability.

If the object or array's contents might be changed, use camelCase.

Immutable Object/Array (Constant)
Use SCREAMING_SNAKE_CASE when the collection itself and its values are meant to be constant.

JavaScript

const VALID_ROLES = Object.freeze(['admin', 'editor', 'viewer']);

const DEFAULT_CONFIG = Object.freeze({
  host: 'localhost',
  port: 8080,
});
Mutable Object/Array (Static, but not Constant)
Use camelCase if the object or array is defined at the top level but its contents are expected to be modified.

JavaScript

const defaultHeaders = {
  'Content-Type': 'application/json',
};

// This is allowed because the variable is camelCase,
// signaling that it might be mutated.
if (user.token) {
  defaultHeaders['Authorization'] = `Bearer ${user.token}`;
}
```

stephl3: so I think `DRAWER_WIDTHS` makes sense and we can take it a step further and wrap static objects/arrays in an `Object.freeze()`

shaneeza: I'm not sure we need `Object.freeze()`. It definitely won't hurt to add it but ~~I think in this case we can use `as const`. From Gemini:~~

```
The as const assertion instantly tells any developer working on the design system that this object is a constant and should not be changed.

If you are only worried about compile-time safety and not runtime guarantees, you can omit Object.freeze() and just use as const
```

NVM we lose type safety with `as const` but we can use `as const satisfies`. This checks that the objects has the correct shape and value types and is readonly.

shaneeza: I'm gonna merge this in, but if it's incorrect, I can always update it in another PR

PR #3003, opened 2025-08-04
LG-5412: patch `darkMode` propagation in ProgressBar

charcoalyy:
## ✍️ Proposed changes

- properly propagate `darkMode` downwards from any provider/context wrapped around the component
- disable `darkMode` as a control in generated stories, which automatically apply `darkMode` themselves

🎟 _Jira ticket:_ [LG-5412](https://jira.mongodb.org/browse/LG-5412)

### DIFF
--- a/packages/progress-bar/src/ProgressBar.stories.tsx
+++ b/packages/progress-bar/src/ProgressBar.stories.tsx
@@ -167,9 +167,9 @@ export const WithProviderDarkMode: StoryObj<typeof ProgressBar> = {
       ],
     },
   },
-  render: _args => (
-    <LeafyGreenProvider darkMode={true}>
-      <ProgressBar {...sharedDeterminateArgs} />
+  render: () => (
+    <LeafyGreenProvider darkMode>
+      <ProgressBar label="Label" isIndeterminate />
     </LeafyGreenProvider>
   ),
   ...disableDarkModeControl,


packages/progress-bar/src/ProgressBar/ProgressBar.tsx
```diff
@@ -59,7 +59,7 @@ export function ProgressBar(props: ProgressBarProps) {
     ...rest
   } = resolveProgressBarProps(props);
```
TheSonOfThomp, line 60: Should we remove the default darkMode value in this resolution function as well?

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -48,16 +58,22 @@ const meta: StoryMetaType<typeof ProgressBar> = {
...
+          </LeafyGreenProvider>
+        </div>
+      ),
     },
   },
+  decorators: [
+    (InstanceFn, context) => (
+      <LeafyGreenProvider darkMode={context?.args.darkMode}>
+        <InstanceFn />
+      </LeafyGreenProvider>
```
TheSonOfThomp: This is still just testing when `darkMode` is passed to the Provider _and_ the component. 
This story won't work: 

```tsx
export const WithProvider: StoryObj<typeof ProgressBar> = {
  ...disabledDarkModeArgTypes,
  args: {
    label: 'Label',
    description: 'Helper text',
  },
  render: args => {
    // eslint-disable-next-line react-hooks/rules-of-hooks
    const [darkMode, setDarkMode] = useState(false);

    return (
        <div
          className={css`
            min-width: 768px;
            padding: 16px;
            background-color: ${color[darkMode ? 'dark' : 'light'].background
              .primary.default};
          `}
        >
          <Label htmlFor="toggle">
            Toggle Dark Mode
            <input
              id="toggle"
              type="checkbox"
              checked={darkMode}
              onChange={e => setDarkMode(e.target.checked)}
            />
            <br />
          </Label>
          <ProgressBar {...args} />
        </div>
      </LeafyGreenProvider>
    );
  },
};
```

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -150,6 +158,23 @@ export const WithHeaderTruncation: StoryObj<typeof ProgressBar> = {
   },
 };
 
+export const WithProviderDarkMode: StoryObj<typeof ProgressBar> = {
+  parameters: {
+    backgrounds: {
+      default: 'dark',
+      values: [
+        { name: 'dark', value: color.dark.background.secondary.default },
```
charcoalyy, line 166: this is pretty ugly. i tried to find a way to manually get the background we use for darkMode _without_ passing darkMode args into the component itself (which negates what is being tested). any suggestions?

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -150,6 +158,23 @@ export const WithHeaderTruncation: StoryObj<typeof ProgressBar> = {
...
+      ],
+    },
+  },
+  render: _args => (
```
TheSonOfThomp: don't even need to declare args `() => `

packages/progress-bar/src/ProgressBar.stories.tsx
TheSonOfThomp: Not sure this is entirely necessary here, but this looks like it works! (it'd be better to do this storybook-wide, but that's a whole different project)

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -150,6 +158,23 @@ export const WithHeaderTruncation: StoryObj<typeof ProgressBar> = {
...
+    <LeafyGreenProvider darkMode={true}>
```
TheSonOfThomp: nit:
```suggestion
    <LeafyGreenProvider darkMode>
```

packages/progress-bar/src/ProgressBar.stories.tsx
```diff
@@ -150,6 +158,23 @@ export const WithHeaderTruncation: StoryObj<typeof ProgressBar> = {
...
+      <ProgressBar {...sharedDeterminateArgs} />
```
TheSonOfThomp: can we add a label to this (to make sure those render dark mode too)?

packages/progress-bar/src/ProgressBar.stories.tsx
TheSonOfThomp: actually we should probably pass in these args. Props like `variant` etc. don't work in Storybook (we should either make them work, or remove them)

PR #3004, opened 2025-08-04
refactor: misc. `Drawer` and `Resizable` updates

shaneeza:
## ✍️ Proposed changes

### `Resizable`
- Moves `onKeyDown` from the window to the resizer element. This simplifies the handler because it does not rely on a useEffect.
- Removes `mousemove` and `mouseup` listeners if the hook is not enabled or resizing

### `Drawer`
- Only enables the `useResizable` hook if the drawer is open.
- Passes the drawer's open size to `initialState` in the `useResizable` hook since it is only enabled when the drawer is open. 
- Adds padding to the embedded layout wrapper so that the focus outline of the toolbar is fully visible.

### For new components


packages/hooks/package.json
```diff
@@ -17,7 +17,6 @@
   },
   "dependencies": {
     "@leafygreen-ui/lib": "workspace:^",
-    "@leafygreen-ui/palette": "workspace:^",
```
shaneeza: This was added when `useResizable` was included with `hooks`, but should be removed since `useResizable` is in its own package

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -148,32 +148,51 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
     const {
       resizableRef,
       size: drawerSize,
       getResizerProps,
       isResizing,
     } = useResizable<HTMLDialogElement | HTMLDivElement>({
-      enabled: resizable && isEmbedded,
-      initialSize: open ? initialSize : 0,
+      enabled: resizable && isEmbedded && open,
+      initialSize: initialSize,
```
TheSonOfThomp:
```suggestion
      initialSize,
```

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -148,32 +148,51 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+    // Enables resizable functionality if the drawer is resizable, embedded and open.
     const {
       resizableRef,
       size: drawerSize,
       getResizerProps,
       isResizing,
     } = useResizable<HTMLDialogElement | HTMLDivElement>({
-      enabled: resizable && isEmbedded,
-      initialSize: open ? initialSize : 0,
+      enabled: resizable && isEmbedded && open,
```
TheSonOfThomp: nit: might be more readable to extract into `const isEnabled = ... `

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -148,32 +148,51 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+      }
+
+      // If the drawer is not resizable, we manually set the drawer width to the size prop when the drawer is open or closed.
+      if (isEmbedded && !resizable) {
+        if (open) {
+          setDrawerWidth(initialSize);
+        } else {
+          setDrawerWidth(0);
+        }
       }
```
TheSonOfThomp: IMO this logic flow is a bit easier to follow: 
```suggestion
      if (!isEmbedded) return;

      if (open) {
        // If the drawer is not resizable, we manually set the drawer width to the size prop when the drawer is open or closed.
        const size = resizable ? drawerSize : initialSize;
        setDrawerWidth(size);
        if (resizable) {
          setIsDrawerResizing(isResizing);
        }
      } else {
        // If the drawer is closed, we set the drawer width to 0.
        setDrawerWidth(0);
        if (resizable) {
          setIsDrawerResizing(false);

        }
      }
```

TheSonOfThomp: Also, should we only be triggering this when `open` changes (i.e. comparing `open` to a `prevOpen`?)

packages/drawer/src/constants.ts
```diff
@@ -19,6 +19,7 @@ export const DRAWER_WIDTHS = {
 export const DRAWER_TOOLBAR_WIDTH = TOOLBAR_WIDTH;
 export const TRANSITION_TIMING_FUNCTION = 'ease-in-out';
 export const TRANSITION_DURATION = transitionDuration.slower;
+export const EMBEDDED_TOOLBAR_OVERFLOW_PADDING = 4;
 
 export const HEADER_HEIGHT = 48;
 export const MOBILE_BREAKPOINT = 390;
```
TheSonOfThomp, line 25: should these use tokens?

packages/resizable/src/useResizable/useResizable.ts
```diff
@@ -240,30 +224,23 @@ export const useResizable = <T extends HTMLElement = HTMLDivElement>({
...
-  useEffect(() => {
-    if (!isFocused && !enabled) return;
-
-    window.addEventListener('keydown', handleKeyDown);
+    if (isResizing && enabled) {
+      window.addEventListener('mousemove', handleMouseMove);
+      window.addEventListener('mouseup', handleMouseUp);
+    } else {
+      // If not enabled, remove the listeners
+      cleanupListeners();
```
TheSonOfThomp, line 237: :+1:

packages/resizable/src/useResizable/useResizable.spec.tsx
```diff
@@ -270,24 +269,30 @@ describe('useResizable', () => {
...
 
-    // Press right arrow to increase size
+    // Trigger keyDown directly since the hook handles keyboard events directly
     act(() => {
-      fireEvent.keyDown(window, { code: keyMap.ArrowRight });
+      resizerProps?.onKeyDown({
+        code: 'ArrowRight',
+        preventDefault: jest.fn(),
+      } as unknown as React.KeyboardEvent);
     });
```
TheSonOfThomp, line 281: Was initially confused by this. We're testing that the handler triggers the resize callback, right? 
Feels strange since this _feels_ like we should be using `userEvent`, but that would technically be testing an implementation detail (did the consumer pass `onKeyDown` to the right element)

TheSonOfThomp (approved):
Nothing blocking, but a few stylistic nits

packages/resizable/src/useResizable/useResizable.spec.tsx
shaneeza: Yeah, this is correct. It's testing that it's calling the resize callback and that the correct size is also returned.

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: Not in the case of the drawer resizing. The size should keep updating while the drawer is open.

packages/drawer/src/constants.ts
shaneeza: The mobile breakpoint is a little too small for drawer so this is a custom size

PR #3005, opened 2025-08-04
Update `Toolbar` active colors

shaneeza:
## ✍️ Proposed changes

- There was a background-color inherited from `IconButton` which was darkening the active background color. These changes override that background-color
- The active icon color was also changing on hover. These changes override the active color

### For new components


packages/toolbar/src/ToolbarIconButton/ToolbarIconButton.styles.ts
```diff
@@ -12,18 +12,36 @@ export const baseIconButtonStyles = css`
...
     border-radius: ${borderRadius[150]}px;
   }
+
+  &::before {
+    background-color: transparent;
+  }
 `;
 
 export const getIconButtonActiveStyles = ({ theme }: { theme: Theme }) =>
   cx(
```
tsck: I know this is not being added in this PR but Is this `cx()` call doing anything here?

shaneeza: nope, i can remove

PR #2612, opened 2024-12-24
LG-4650: migrate from yarn to pnpm

stephl3:
## ✍️ Proposed changes

- migrates from `yarn` to `pnpm`

🎟 _Jira ticket:_ [LG-4650](https://jira.mongodb.org/browse/LG-4650)

### For bug fixes, new features & breaking changes

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `yarn changeset` and documented my changes

## 🧪 How to test changes

- [ ] spot check storybook

PR #3008, opened 2025-08-05
fix: pnpm slackbot

stephl3:
## ✍️ Proposed changes

After merging PR https://github.com/mongodb/leafygreen-ui/pull/2612, the "Notify Slack & Website" step of our release pipeline began failing. PNPM handles arguments differently than yarn so "release" was inadvertently being passed as the arg for updates data instead of the actual release outputs. `lg slackbot` already lists `release` as a subcommand, so it is now dropped from the root `package.json` alias. `release.yml` is also updated to use the fixed alias. This change has been tested in the #design-systems-testing channel.

### For new components

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -239,7 +239,7 @@ jobs:
 
       - name: Notify Slack channel of new releases
         id: slackbot
-        run: pnpm slackbot '${{ needs.release.outputs.publishedPackages }}'
+        run: pnpm slackbot release '${{ needs.release.outputs.publishedPackages }}'
         env:
           SLACK_BOT_TOKEN: '${{ secrets.SLACK_BOT_TOKEN }}'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -239,7 +239,7 @@ jobs:
 
       - name: Notify Slack channel of new releases
         id: slackbot
-        run: pnpm slackbot '${{ needs.release.outputs.publishedPackages }}'
+        run: pnpm slackbot release '${{ needs.release.outputs.publishedPackages }}'
```
TheSonOfThomp, line 242: that's it...!?

TheSonOfThomp (approved):

.github/workflows/release.yml
stephl3: 🤣 😭

PR #3016, opened 2025-08-06
[LG-5417] fix(drawer): stale toolbar data

shaneeza:
…state handling

## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-5417](https://jira.mongodb.org/browse/LG-5417)

This PR fixes a bug where changes to `toolbarData` were not reflected in an open drawer. For example, [if the state for the drawer content lived outside of the content, and the state was updated, those changes were not reflected in the opened drawer](https://codesandbox.io/p/sandbox/drawer-with-toolbar-example-forked-lkml5p). This happens because when a drawer opens, the active drawer content is cached, and that data is rendered in the drawer. If changes are made to the data while the drawer is open, the drawer will still render the stale cached data. To fix this, the active data is no longer cached. Only the active data id is cached, and the active data is always fetched from the fresh data.

## 🧪 How to test changes

Check out the dynamic drawer stories:
- [Overlay Dynamic](https://642700aa3e49e32bdbf0b0fc-xdyxpydgzu.chromatic.com/?path=/story/sections-drawer-toolbar--overlay-dynamic)
- [Embedded Dynamic](https://642700aa3e49e32bdbf0b0fc-xdyxpydgzu.chromatic.com/?path=/story/sections-drawer-toolbar--embedded-dynamic)


packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -236,6 +236,7 @@ const getInnerOpenContainerStyles = css`
   transition-property: opacity;
   transition-duration: ${transitionDuration.slowest}ms;
   transition-timing-function: ${TRANSITION_TIMING_FUNCTION};
+  transition-delay: ${transitionDuration.default}ms;
```
shaneeza, line 239: add note in cs

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarContext/DrawerToolbarContext.tsx
```diff
@@ -22,9 +21,14 @@ export const DrawerToolbarProvider = ({
   children,
   data,
 }: DrawerToolbarProviderProps) => {
-  const [content, setContent] = useState<ContextData>(undefined);
+  const [activeDrawerId, setActiveDrawerId] = useState<DataId | null>(null);
   const [isDrawerOpen, setIsDrawerOpen] = useState<boolean>(false);
```
TheSonOfThomp, line 25: Since `activeDrawerId` and `isDrawerOpen` are almost always changed together, it might be worth converting this to a reducer to avoid unnecessary rerenders

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarContext/DrawerToolbarContext.tsx
```diff
@@ -33,30 +37,37 @@ export const DrawerToolbarProvider = ({
...
+   * Closes the drawer and sets the active drawer id to null.
+   * The content will be removed after a delay to allow the drawer to close first.
+   */
   const closeDrawer = useCallback(() => {
     // Delay the removal of the content to allow the drawer to close before removing the content
     setTimeout(() => {
-      setContent(undefined);
+      setActiveDrawerId(null);
     }, TRANSITION_DURATION);
     setIsDrawerOpen(false);
```
TheSonOfThomp: nit: imo calling `setIsDrawerOpen` before the timeout would be more readable (since it happens first)

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout/DrawerToolbarLayout.spec.tsx
```diff
@@ -78,4 +90,20 @@ describe('packages/DrawerToolbarLayout', () => {
 
     expect(onCloseMock).toHaveBeenCalled();
   });
+
+  test('allows for dynamic toolbar data', () => {
```
TheSonOfThomp: nit: this test works, so I don't think you need to change it. but IMO testing `render(<DrawerLayout toolbarData={DATA} />)` and then `rerender(<DrawerLayout toolbarData={NEW_DATA} />)` would be more directly testing what you want to test

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarContext/DrawerToolbarContext.tsx
shaneeza: In React 18, [state updates are batched](https://react.dev/blog/2022/03/29/react-v18#new-feature-automatic-batching), but this will probably cause an extra rerender in React 17, which I think is okay.

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout/DrawerToolbarLayout.spec.tsx
shaneeza: I think I prefer this

PR #3022, opened 2025-08-08
[LG-5427] fix(drawer): Drawer title width

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-5427](https://jira.mongodb.org/browse/[LG-5427)

This PR updates the drawer title so that it takes up all of the available free space in the container.

- [x] I have added tests/story that prove my fix is effective or that my feature works

## 🧪 How to test changes

View the following story:


packages/drawer/src/Drawer.stories.tsx
```diff
@@ -35,6 +38,29 @@ const snapshotStoryExcludedControlParams = [
...
+      `}
+    >
+      <div>Drawer title</div>
+      <div
+        className={css`
+          display: flex;
+        `}
+      >
+        <IconButton onClick={() => console.log('click')}>
+          <CloudIcon aria-label="Cloud" />
```
stephl3: does this `aria-label` need to be moved to the `IconButton` instance? I think that story page has console errors because it's missing an `aria-label` or `aria-labelledby`

packages/drawer/src/Drawer.stories.tsx
```diff
@@ -35,6 +38,29 @@ const snapshotStoryExcludedControlParams = [
...
+        justify-content: space-between;
+        align-items: center;
+      `}
+    >
+      <div>Drawer title</div>
+      <div
+        className={css`
+          display: flex;
+        `}
+      >
```
stephl3, line 67: is this extra wrapper div doing anything or can it be removed?

packages/drawer/src/Drawer.stories.tsx
```diff
@@ -35,6 +38,29 @@ const snapshotStoryExcludedControlParams = [
...
+const DrawerCustomTitle = () => {
+  return (
+    <div
+      className={css`
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+      `}
+    >
+      <div>Drawer title</div>
```
stephl3: I know this is just a story, but because consumers refer to these at times for their implementations, can we use the correct semantic typography to wrap this title?

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -240,6 +241,7 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
                   as={typeof title === 'string' ? 'h2' : 'div'}
                   baseFontSize={BaseFontSize.Body2}
                   id={titleId}
+                  className={titleStyles}
```
stephl3, line 244: unrelated to the current ticket, but curious about this: should the `<strong>` tag be wrapping a `ReactNode` or should it only be used if `title` is a string?

packages/drawer/src/Drawer.stories.tsx
shaneeza: removed

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: From researching on Claude, it seems like it should be okay to wrap a `ReactNode` in a strong tag. This indicates that the entire title content has strong importance in the drawer's context. 

I'm also thinking that, for now, if we remove the `<strong>` tag, then this will be considered a breaking change.

If we don't want to keep as-is then we would either have the consumer wrap the title in a `<strong>` tag or we can create a new `<DrawerTitle>` component with all the styles built in:

`DrawerTitle` would look like:
```jsx
<Body
  as='h2'
  weight={FontWeight.SemiBold}
>
  {children}
</Body>
```
And the story above would become:
```jsx
const DrawerCustomTitle = () => {
  return (
    <div
      className={css`
        display: flex;
        align-items: center;
      `}
    >
      <DrawerTitle>
        Drawer title
      </DrawerTitle>
      <IconButton
        aria-label="Cloud"
        onClick={() => console.log('cloud clicked')}
      >
        <CloudIcon />
      </IconButton>
    </div>
  );
};
```

Inside Drawer:
```jsx
<div
    id={titleId}
    className={titleStyles}
  >
    {typeof title === 'string' ? <DrawerTitle>{title}</DrawerTitle> : title}
  </div>
  {showCloseButton && (
    <IconButton
      aria-label="Close drawer"
      data-lgid={lgIds.closeButton}
      data-testid={lgIds.closeButton}
      onClick={onClose}
    >
      <XIcon />
    </IconButton>
  )}
</div>
```

Thoughts?

.changeset/pretty-places-visit.md
```diff
@@ -0,0 +1,6 @@
+---
+'@leafygreen-ui/drawer': minor
+---
+
+- Adds `inlineChildrenInTitle` prop to `Drawer` and `DrawerToolbarLayout` components. This prop allows for inline children to be rendered to the right of the title.
```
shaneeza: Not crazy about `inlineChildrenInTitle`, but `inlineChildren` to me is not obvious that the inline children will be in the title. When Terrence is back, I'll see what prop name he's using for the toolbar

packages/drawer/src/Drawer/Drawer.tsx
stephl3: This is what I got back from Gemini on `<strong>` vs `font-weight`:
```
To minimize mental overhead, adopt this clear rule:
1. Use CSS font-weight for all visual styling where you want text to appear bold.
2. Use the <strong> tag only when you need to semantically mark text as important, serious, or urgent.
```

It doesn't look like we consistently apply `<strong>` tags across the system, so I'd be curious to develop better understanding of when it should be used or if we already have docs on this

In this case, I think it might be better to rely on the proper heading tag (`<h2>`) for semantic meaning while applying font-weight with visual styling. If consumers really needed to apply additional emphasis, they could use the `<strong>` tag, but I'm not convinced it's definitive for this case

Happy to discuss this further during a dev sync and don't need to block on it right now since it's already in prod

.changeset/pretty-places-visit.md
stephl3: Agreed, `inlineChildren` naming is specific to the `Tabs` implementation. Aligning with the `CodeEditor` API sounds more appropriate for this case

packages/drawer/src/Drawer/Drawer.types.ts
```diff
@@ -54,6 +54,15 @@ export interface DrawerProps
...
 
+  /**
+   * Inline children in the title of the Drawer
+   */
+  inlineChildrenInTitle?: React.ReactNode;
+
+  /**
+   * Custom action buttons in the title of the Drawer
+   */
+  customActionButtons?: React.ReactNode;
```
stephl3: `headerActions`, `customHeaderActions`, `inlineHeaderActions`?

packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -314,3 +314,11 @@ export const getScrollContainerStyles = ({
...
     [baseScrollContainerStyles]: scrollable,
   });
+
+export const titleStyles = css`
+  flex-grow: 1;
+`;
+
+export const inlineChildrenStyles = css`
+  display: flex;
+`;
```
stephl3: should these be vertically aligned and any gap?

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -240,9 +243,15 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
                 >
                   <strong>{title}</strong>
                 </Body>
+                {inlineChildrenInTitle && (
+                  <div className={inlineChildrenStyles}>
+                    {inlineChildrenInTitle}
+                  </div>
+                )}
```
stephl3: wonder if maybe these should all be in the same container with the close button? probably would make sense if they have uniform gaps

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: I forgot to update this comment, but I agree with using `font-weight`, and I've already updated the title to use it.

packages/drawer/src/Drawer/Drawer.styles.ts
shaneeza: The specs have no gap, and they should already be vertically aligned. Did they look off-center?

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: I can update this

.changeset/pretty-places-visit.md
shaneeza: Terrence is using `children` because the slot can technically accept anything, like icons or a select component. In this case, we just want icons. So we can align with `Code`, which uses `customActionButtons` or we can go with one of your suggestions. I like `customHeaderActions` and `inlineHeaderActions`. I can bring this up during dev sync so we're all aligned.

shaneeza: Update: During dev sync, we agreed to remove this new prop and have only the `title` prop. If a string is passed to `title`, we will render the title with the correct styles and semantic tag. If a React node is passed, the consumer is responsible for the styles and adding the correct semantic tag. An recipe/example will be added to the README.

packages/drawer/src/Drawer.stories.tsx
shaneeza: Added back since I added more icons

PR #3023, opened 2025-08-08
Allows Select to be a controlled component

bruugey:
## ✍️ Proposed changes

### For new components


packages/icon/src/generated/AIModel.tsx
TheSonOfThomp, line 1: does this need to be synced with `main`?

packages/select/src/Select/Select.tsx
```diff
@@ -121,7 +123,11 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>(
     const descriptionId = `${id}-description`;
     const menuId = `${id}-menu`;
 
-    const [open, setOpen] = useState(false);
+    const [uncontrolledOpen, setUncontrolledOpen] = useState(false);
+
+    // Use controlled open state if provided, otherwise use internal state
+    const open = openProp !== undefined ? openProp : uncontrolledOpen;
+    const setOpen = openProp !== undefined ? setOpenProp! : setUncontrolledOpen;
```
TheSonOfThomp: We use a similar pattern across a bunch of components (e.g. https://github.com/mongodb/leafygreen-ui/blob/92c38b1c187e80ed26a564dfeba18523d9310d51/packages/tooltip/src/Tooltip/Tooltip.tsx). We should consider making this a hook

packages/select/src/Select/Select.tsx
```diff
@@ -143,17 +149,27 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>(
...
      * Open / close state
      */
     const onOpen = useCallback(() => {
       setOpen(true);
-    }, []);
+    }, [setOpen]);
 
     const onClose = useCallback(() => {
       setOpen(false);
       menuButtonRef.current!.focus();
```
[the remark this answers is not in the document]
TheSonOfThomp: ^ Copilot is correct. If for some reason `menuButtonRef.current` doesn't exist, this will throw an error

packages/icon/src/generated/AIModel.tsx
bruugey: Rebasing :)

.changeset/famous-ties-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/select': minor
+---
+
+Adds controlled way to use Select component
```
TheSonOfThomp: Can we clarify what this means?

packages/select/src/Select/Select.tsx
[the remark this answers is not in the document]
TheSonOfThomp: Copilot is right here. `setOpenProp` might not be passed in even when `openProp` is defined. In that case we'll get errors when calling `setOpen()`

TheSonOfThomp: This still isn't correct though, since it'll use the internal setter with the external state value
```suggestion
    const open = isDefined(openProp) ? openProp : uncontrolledOpen;
    const setOpen = isDefined(openProp) && isDefined(setOpenProp) ? setOpenProp : setUncontrolledOpen;
```

TheSonOfThomp (changes requested):

.changeset/famous-ties-arrive.md
bruugey: Updated

packages/select/src/Select/Select.spec.tsx
```diff
@@ -1163,4 +1163,161 @@ describe('packages/select', () => {
...
+      render(<ControlledOpenSelect initialOpen={true} />);
+
+      const { getPopover } = getTestUtils();
+
+      await waitFor(() => {
+        expect(getPopover()).toBeInTheDocument();
+      });
+    });
+
+    test('backwards compatibility - uncontrolled mode still works', async () => {
```
TheSonOfThomp, line 1271: this test is unnecessary, since this is what the existing tests should be testing

packages/select/src/Select/Select.tsx
```diff
@@ -121,7 +123,14 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>(
...
 
-    const [open, setOpen] = useState(false);
+    const [uncontrolledOpen, setUncontrolledOpen] = useState(false);
+
+    // Use controlled open state if provided, otherwise use internal state
+    const open = isDefined(openProp) ? openProp : uncontrolledOpen;
+    const setOpen =
+      isDefined(openProp) && isDefined(setOpenProp)
+        ? setOpenProp
+        : setUncontrolledOpen;
```
TheSonOfThomp, line 133: If `setOpenProp` is defined, but `openProp` _is_, then nothing will happen when a user clicks the Select/backdrop to close it (since the wrong callback gets fired)

(We have this same bug in [`Menu`](https://github.com/mongodb/leafygreen-ui/blob/main/packages/menu/src/Menu/Menu.tsx) as well)

packages/select/src/Select.stories.tsx
```diff
@@ -219,3 +236,260 @@ export const InitialLongSelectOpen = {
...
+      >
+        <Button onClick={() => setIsOpen(true)} size="small">
+          Open Select
+        </Button>
+        <Button onClick={() => setIsOpen(false)} size="small">
+          Close Select
+        </Button>
+        <Button onClick={() => setIsOpen(!isOpen)} size="small">
+          Toggle Select
+        </Button>
```
TheSonOfThomp, line 268: nit: the toggle doesn't work to close the menu (backdrop click will fire, and then button click to open again)

packages/select/src/Select/Select.tsx
TheSonOfThomp: Just noticed that's what the useEffect below it checking for

packages/select/src/Select/Select.tsx
```diff
@@ -143,17 +152,27 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>(
       }
     }, [onChange, readOnly, value]);
 
+    useEffect(() => {
+      if (openProp !== undefined && setOpenProp === undefined) {
+        console.warn(
```
TheSonOfThomp, line 157: I think this should be an error, since the component won't behave as expected
```suggestion
        consoleOnce.error(
```

packages/select/src/Select/Select.tsx
```diff
@@ -18,7 +18,7 @@ import LeafyGreenProvider, {
   PopoverPropsProvider,
   useDarkMode,
 } from '@leafygreen-ui/leafygreen-provider';
-import { keyMap } from '@leafygreen-ui/lib';
+import { isDefined, keyMap } from '@leafygreen-ui/lib';
```
TheSonOfThomp, line 21:
```suggestion
import { consoleOnce, isDefined, keyMap } from '@leafygreen-ui/lib';
```

packages/select/src/Select/Select.tsx
```diff
@@ -143,17 +152,27 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>(
...
+    }, [setOpen]);
 
     const onClose = useCallback(() => {
       setOpen(false);
-      menuButtonRef.current!.focus();
-    }, [menuButtonRef]);
+      menuButtonRef.current?.focus();
+    }, [setOpen, menuButtonRef]);
 
     useEffect(() => {
```
TheSonOfThomp, line 177: (out-of-scope) can this be converted to use `useBackdropClick`?

PR #3033, opened 2025-08-12
[LG-5432] feat(suggestions): add apply state and improve component flexibility

stephl3:
## ✍️ Proposed changes

This PR enhances the SuggestedActions component with improved flexibility and updated state management system. The main changes include:

1. **New 'apply' state**: Adds a dedicated `State.Apply` value that conditionally renders the apply button, while `State.Unset` no longer shows the button
2. **Enhanced prop spreading**: Enables all HTMLDivElement props to spread to the root container node for better customization
3. **Responsive width**: Removes the fixed 353px width constraint, allowing the component to expand to its parent container
4. **Improved styling**: Adds `className` prop support for custom styling while maintaining the existing design system

These changes provide developers with more control over the component's appearance and behavior, making it more suitable for various use cases while maintaining backward compatibility.

🎟️ _Jira ticket:_ [LG-5432](https://jira.mongodb.org/browse/LG-5432)

## ✅ Checklist

For bug fixes, new features & breaking changes:

### Testing the new Apply state:
1. Render `SuggestedActions` with `state={State.Apply}` - verify the apply button is visible
2. Render `SuggestedActions` with `state={State.Unset}` - verify the apply button is NOT visible
3. Test both states with `onClickApply` callback - verify the button functions correctly

### Testing rest props spreading:
1. Pass custom props like `data-testid="test"`, `aria-label="test"`, or `style={{ border: '1px solid red' }}`
2. Verify these props are applied to the root div element
3. Test with `className="custom-class"` - verify custom styles are merged with base styles

### Testing responsive width:
1. Place the component in containers of different widths
2. Verify the component expands to fill the parent container instead of being fixed at 353px
3. Test in responsive layouts to ensure proper behavior

### Testing backward compatibility:
1. Verify existing `State.Success` and `State.Error` states work as before
2. Ensure all existing props continue to function
3. Verify the component renders identically for existing use cases

### DIFF
--- a/chat/suggestions/src/SuggestedActions/SuggestedActions.spec.tsx
+++ b/chat/suggestions/src/SuggestedActions/SuggestedActions.spec.tsx
@@ -527,40 +527,6 @@ describe('chat/suggestions', () => {
 
       expect(ref.current).toBeInstanceOf(HTMLDivElement);
     });
-
-    test('spreads rest props to root div', () => {
-      const { container } = renderSuggestedActions({
-        'data-testid': 'suggestions-card',
-        'aria-label': 'Configuration suggestions',
-        style: { border: '2px solid red' },
-      });
-
-      const rootDiv = container.firstChild as HTMLElement;
-      expect(rootDiv).toHaveAttribute('data-testid', 'suggestions-card');
-      expect(rootDiv).toHaveAttribute(
-        'aria-label',
-        'Configuration suggestions',
-      );
-      expect(rootDiv).toHaveStyle('border: 2px solid red');
-    });
-
-    test('applies custom className to root div', () => {
-      const customClassName = 'custom-suggestions-class';
-      const { container } = renderSuggestedActions({
-        className: customClassName,
-      });
-
-      const rootDiv = container.firstChild as HTMLElement;
-      expect(rootDiv).toHaveClass(customClassName);
-    });
-
-    test('expands to parent container width', () => {
-      const { container } = renderSuggestedActions();
-      const rootDiv = container.firstChild as HTMLElement;
-
-      // The component should have width: 100% instead of a fixed width
-      expect(rootDiv).toHaveStyle('width: 100%');
-    });
   });
 
   describe('props validation', () => {


chat/suggestions/README.md
```diff
@@ -37,13 +37,20 @@ const configurationParameters = [
...
   onClickApply={() => console.log('Apply clicked')}
 />
 
+// Apply state with apply button
+<SuggestedActions
+  state={State.Apply}
+  configurationParameters={configurationParameters}
+  onClickApply={() => console.log('Apply clicked')}
+/>
+
```
shaneeza, line 53: This isn't related to this PR, but do the examples below also need `onClickApply`?

chat/suggestions/src/SuggestedActions/SuggestedActions.spec.tsx
```diff
@@ -530,6 +527,40 @@ describe('chat/suggestions', () => {
 
       expect(ref.current).toBeInstanceOf(HTMLDivElement);
     });
+
+    test('spreads rest props to root div', () => {
```
shaneeza: These tests test implementation rather than behavior. I'm curious, what's the reason for including them?

chat/suggestions/README.md
shaneeza: I guess it makes sense to keep for whenever the state changes to `apply`

chat/suggestions/src/SuggestedActions/SuggestedActions.types.ts
```diff
@@ -5,20 +5,23 @@ import { ConfigurationParameters, State } from '../shared.types';
...
-   * Determines rendering of the SuggestedActions:
-   * - `'Unset'` will render suggestions and the "Apply" button
-   * - `'Success'` will render success banner with applied suggestions
-   * - `'Error'` will render error banner with instructions to manually apply suggestions
-   */
-  state: State;
   /**
    * Configuration parameters with their individual state values.
-   * Each parameter includes a key, value, and state (unset/success/error).
+   * Each parameter includes a key, value, and state (unset/apply/success/error).
```
shaneeza, line 10: What happens if the state is `apply` or `unset`? I only see a filter for `success` and `error` within `configurationParameters` in `SuggestedActions.tsx`

chat/suggestions/README.md
stephl3: yeah I think it makes sense. It would need to have been used to reach a success or error state

chat/suggestions/src/SuggestedActions/SuggestedActions.spec.tsx
stephl3: LOL oops. AI doing AI things 🙃 removed

chat/suggestions/src/SuggestedActions/SuggestedActions.types.ts
stephl3: Unset: only the suggestions card
Apply: suggestions card + apply button
Success: suggestions card + success banner
Error: suggestions card + error banner

shaneeza: This comment you're updating is for `configurationParameters` but inside `SuggestedActions.tsx`, there is only a filter for `success` and `error`.

https://github.com/mongodb/leafygreen-ui/blob/5fd75db2cd3b81e781554999a7dbe02cff20a4fb/chat/suggestions/src/SuggestedActions/SuggestedActions.tsx#L37-L43

if you pass the following to configurationParameters, what will happen?
```tsx
configurationParameters: [
      {
        key: 'Cluster Tier',
        value: 'M200 ($14.59/hour)',
        state: State.Apply,
      },
    ],
```

stephl3: `successParameters` and `errorParameters` are both used in the `StatusBanner` that is only rendered for the case when state is `'success'` or `'error'`
```
{([State.Success, State.Error] as Array<State>).includes(state) && (
  <StatusBanner
    state={state}
    appliedParameters={successParameters}
    failedParameters={errorParameters}
  />
)}
```

In that case, they would render in the card, and a `StatusBanner` instance would not render

shaneeza: I got it. So you don't need to pass `apply` or `unset` to `configurationParameters` because it won't do anything. This sort of threw me off because it seems like, since you can pass `Apply` or `Unset`, something would change. Should we narrow the state to only `success` or `error`?

stephl3: I'm hesitant to restrict this because the eng from Atlas Growth who initially implemented this API might need this. I made this ticket to revisit and can clean this up if not needed: https://jira.mongodb.org/browse/LG-5457

PR #3037, opened 2025-08-13
test(popover): enable top-layer popover tests with polyfill and improve test coverage

stephl3:
## ✍️ Proposed changes

This PR fixes the Popover component's top-layer rendering tests by adding the `@oddbird/popover-polyfill` dependency and enabling previously skipped tests. The changes include:

1. **Test infrastructure**: Added polyfill dependency to support Popover API testing in JSDOM
2. **Enabled skipped tests**: Uncommented and fixed tests for auto-dismiss behavior (click outside, escape key)
3. **Improved test coverage**: Enhanced test helpers with proper state management and event handling
4. **Code cleanup**: Removed deprecated `usePortal` prop usage and improved type safety

These changes ensure the Popover component's dismiss behavior is properly tested across all render modes and dismiss modes, improving reliability and catching potential regressions.

��️ _Jira ticket:_ No JIRA ticket associated with this change

## ✅ Checklist

For bug fixes, new features & breaking changes:

### Testing the polyfill integration:
1. Run the Popover test suite to verify all previously skipped tests now pass
2. Verify that `DismissMode.Auto` tests work correctly for both click outside and escape key scenarios
3. Confirm that `DismissMode.Manual` tests continue to work as expected

### Testing top-layer popover behavior:
1. Test auto-dismiss when clicking outside the popover (should close automatically)
2. Test auto-dismiss when pressing Escape key (should close automatically)
3. Test manual dismiss mode (should not close on outside clicks or escape key)

### Testing render modes:
1. Verify inline render mode works correctly with the updated test helpers
2. Verify portal render mode works correctly with the updated test helpers
3. Verify top-layer render mode works correctly with the polyfill

### Testing backward compatibility:
1. Ensure existing popover functionality remains unchanged
2. Verify that the `onToggle` callback is properly called in all scenarios
3. Confirm that state management works correctly across all render modes

### DIFF
--- a/packages/popover/package.json
+++ b/packages/popover/package.json
@@ -29,7 +29,8 @@
   "devDependencies": {
     "@leafygreen-ui/button": "workspace:^",
     "@leafygreen-ui/palette": "workspace:^",
-    "@lg-tools/build": "workspace:^"
+    "@lg-tools/build": "workspace:^",
+    "@oddbird/popover-polyfill": "^0.5.2"
   },
   "peerDependencies": {
     "@leafygreen-ui/leafygreen-provider": "workspace:^"
--- a/packages/popover/src/Popover/Popover.spec.tsx
+++ b/packages/popover/src/Popover/Popover.spec.tsx
@@ -6,6 +6,12 @@ import { axe } from 'jest-axe';
 import Button from '@leafygreen-ui/button';
 import { PopoverContext } from '@leafygreen-ui/leafygreen-provider';
 
+/**
+ * JSDOM does not support the popover API, so we polyfill it
+ * https://github.com/jsdom/jsdom/issues/3721
+ */
+import '@oddbird/popover-polyfill';
+
 import { Popover } from './Popover';
 import {
   DismissMode,
--- a/.changeset/eager-pigs-build.md
+++ b/.changeset/eager-pigs-build.md
@@ -1,5 +1,0 @@
----
-'@lg-tools/test': patch
----
-
-Add `@oddbird/popover-polyfill` dep for polyfilling popover API testing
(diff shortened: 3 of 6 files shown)


packages/popover/package.json
```diff
@@ -29,7 +29,8 @@
   "devDependencies": {
     "@leafygreen-ui/button": "workspace:^",
     "@leafygreen-ui/palette": "workspace:^",
-    "@lg-tools/build": "workspace:^"
+    "@lg-tools/build": "workspace:^",
+    "@oddbird/popover-polyfill": "^0.5.2"
```
stephl3, line 33: heads up @TheSonOfThomp I changed it up and updated to only using the polyfill for the popover package because it was causing significant testing regressions in other packages

TheSonOfThomp: Hmm, can you open a ticket to go back and address those?

packages/popover/src/Popover/Popover.spec.tsx
```diff
@@ -249,43 +270,57 @@ describe('packages/popover', () => {
...
+         * `@oddbird/popover-polyfill` there isn't a way to click outside of the
+         * popover to close it. This is currently tested in the Live Example story
+         */
+        fireEvent(
+          popover,
+          new ToggleEvent('toggle', {
+            bubbles: true,
+            oldState: 'open',
+            newState: 'closed',
+          }),
```
TheSonOfThomp, line 296: is this worth testing then, since we're just testing that the toggle function is called when we trigger a toggle

stephl3: synced offline; leaving in and filling ticket with polyfill library

PR #3038, opened 2025-08-13
[LG-5429] fix(Table): Dynamic table data

shaneeza:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-5429](https://jira.mongodb.org/browse/LG-5429).

This PR fixes a bug that prevented rows and cells from re-rendering when changes were made inside the column definition. Currently, rows are memorized using a custom check, but because of how deeply nested row children are, it's not straightforward to check if a child cell has dynamically updated and if it should re-render. In this PR, I am checking if changes have been made to the column definition. If there are changes, that means that the cell content may have changed, and we should re-render all rows.

## 🧪 How to test changes

Check out the following stories and click on the toggle button to see the cell update 

[regular react table](https://642700aa3e49e32bdbf0b0fc-pkvifjkwqj.chromatic.com/?path=/story/composition-data-display-table--dynamic-data)
[virtual react table](https://642700aa3e49e32bdbf0b0fc-pkvifjkwqj.chromatic.com/?path=/story/composition-data-display-table-with-virtualized-scrolling--dynamic-data)


packages/table/src/Row/Row.types.ts
```diff
@@ -13,17 +13,14 @@ import { Theme } from '@leafygreen-ui/lib';
 
 import { LeafyGreenTableRow, LGRowData } from '../useLeafyGreenTable';
 
-export interface InternalRowBaseProps extends ComponentPropsWithoutRef<'tr'> {
+export interface RowBaseProps extends ComponentPropsWithoutRef<'tr'> {
```
shaneeza, line 16: This is a drive-by update. These props were marked as internal, but they are used for the consumer-facing row component.

packages/table/src/Table/TableWithVS.stories.tsx
```diff
@@ -193,6 +193,122 @@ export const Basic: StoryFn<StoryTableProps> = args => {
...
+        </TableBody>
+      </Table>
+    </>
+  );
+};
+DynamicData.parameters = {
+  chromatic: {
+    disableSnapshot: true,
+  },
+};
```
TheSonOfThomp, line 310: Should we have a chromatic test with a `play` function to validate that interaction causes the table to change?

packages/table/src/TableContext/TableContext.types.ts
```diff
@@ -27,6 +27,11 @@ interface BaseTableContextValue<T extends LGRowData> {
    * LGIDs for table components
    */
   lgIds: GetLgIdsReturnType;
+
+  /**
+   * Whether the rows should be memoized.
+   */
+  shouldMemoizeRows?: boolean;
```
TheSonOfThomp, line 34: can we add a note that this is extended from the hook props?

TheSonOfThomp, line 34: should this be required or optional (since we're controlling when the context gets rendered, right? )

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.tsx
```diff
@@ -34,6 +36,16 @@ function useLeafyGreenTable<T extends LGRowData, V extends unknown = unknown>({
     (rest.initialState?.expanded as ExpandedState) ?? {},
   );
 
+  const prevColumns = usePrevious(columnsProp);
+  /**
+   * This is used to determine if table rows should re-render. If the column definitions are the same, table rows will not re-render. If they are different, all rows will re-render.
```
TheSonOfThomp:
```suggestion
   * This is used to determine if table rows should re-render when column definitions change. If the column definitions are the same, table rows will not re-render. If they are different, all rows will re-render.
```

TheSonOfThomp: Also, can you be more explicit about what happens when this is true, and when it's false? I think I understand but, it's still a bit unclear

packages/table/src/useLeafyGreenTable/useLeafyGreenTable.types.ts
```diff
@@ -81,5 +81,14 @@ export interface LeafyGreenTable<T extends LGRowData>
...
+
+  /**
+   * Available [properties and methods](https://tanstack.com/virtual/latest/docs/api/virtualizer#virtualizer-instance) return from the Virtualizer instance.
+   */
   virtual: never;
+
+  /**
+   * Whether the rows should be memoized.
+   */
+  shouldMemoizeRows: boolean;
```
TheSonOfThomp: Since this is the user-facing type, can you add more TSDoc here explaining what happens when this is true, and when it's false, and how its value is determined?

packages/table/src/Row/InternalRowWithRT.tsx
```diff
@@ -77,17 +78,19 @@ const InternalRowWithRT = <T extends LGRowData>({
...
 
-    const propsAreEqual =
-      isEqual(restPrevProps, restNextProps) &&
-      prevChildrenCount === nextChildrenCount;
+    if (prevChildrenCount !== nextChildrenCount) return false;
 
-    return propsAreEqual;
+    // Compare all props except children and shouldMemoizeRows
+    const { children: _1, shouldMemoizeRows: _2, ...prevRest } = prevProps;
+    const { children: _3, shouldMemoizeRows: _4, ...nextRest } = nextProps;
```
TheSonOfThomp: Just a nit, but if we're omitting more than one prop, I'd use `_.omit`. Not sure if it's more performant, but it's easier to read imo

```suggestion
    const prevRest = omit(prevProps, ['children', 'shouldMemoizeRows']);
    const nextRest = omit(nextProps, ['children', 'shouldMemoizeRows']);
```

TheSonOfThomp (approved):
Can you add more consumer-facing docs on what this does and why it's needed? 
Otherwise LGTM!

packages/table/src/TableContext/TableContext.types.ts
shaneeza: The context is used for both a react table and a non react table. If using a react table then the hook should return `shouldMemoizeRows`, else that value is undefined. Because of that I have it set to optional.

PR #3041, opened 2025-08-14
[LG-5095] feat(CodeEditor): adds `CodeEditorCopyButton`

tsck:
## ✍️ Proposed changes

- Implemented `CodeEditorCopyButton` to allow users to copy text content to the clipboard. This will be used in a future PR internally by `CodeEditor`.
- Added visual feedback for copy success with tooltip support.
- Integrated accessibility features including aria-labels and screen reader announcements.
- Enhanced styling for different button variants (Button and IconButton).
- Included tests for rendering, copy functionality, and keyboard interactions.
- Updated package.json to include new dependencies for the component.

🎟 _Jira ticket:_ [LG-5095](https://jira.mongodb.org/browse/LG-5095)

## 🧪 How to test changes
- All added tests passing


packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.types.ts
```diff
@@ -0,0 +1,41 @@
...
+   * Callback function called after successful copy operation.
+   * Optional - if not provided, only the visual feedback will be shown.
+   */
+  onCopy?: () => void;
+
+  /**
+   * Function that returns the content to be copied to the clipboard.
+   * This function is called when the copy button is clicked.
+   */
+  getContents: () => string;
```
stephl3: thoughts on making this `onCopyClick`? or maybe `getContentsToCopy`?

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.types.ts
```diff
@@ -0,0 +1,41 @@
...
+  variant?: CopyButtonVariant;
+
+  /**
+   * Whether to render the button as an icon-only variant for use in panels.
+   * When true, renders as an IconButton; when false, renders as a Button with text.
+   *
+   * @default false
+   */
+  isPanelVariant?: boolean;
+}
```
stephl3: looks like this might not be used in favor of using `variant`? if so, can we rm here and the mention in the tsdocs?

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+            leftGlyph={copied ? <CheckmarkIcon /> : <CopyIcon />}
+            size="xsmall"
+            {...sharedButtonProps}
+          >
+            {copied && (
+              <VisuallyHidden role="alert">{COPIED_TEXT}</VisuallyHidden>
+            )}
+          </Button>
+        )
+      }
```
stephl3: nit: looks like it could be a separate subcomponent and opportunity to DRY some of this up

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+  /**
+   * Handles keyboard interactions for the copy button.
+   * Supports Enter and Space keys to trigger the copy action.
+   */
+  const handleKeyDown = (e: React.KeyboardEvent) => {
+    switch (e.key) {
+      case keyMap.Escape:
+      case keyMap.Tab: {
+        break;
+      }
```
stephl3: do we no longer want this to handle closing the tooltip?

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+        const textArea = document.createElement('textarea');
+        textArea.value = getContents();
+        textArea.style.position = 'fixed';
+        textArea.style.left = '-999999px';
+        textArea.style.top = '-999999px';
+        document.body.appendChild(textArea);
+        textArea.focus();
+        textArea.select();
+        document.execCommand('copy');
+        textArea.remove();
```
stephl3: curious why we need to support this? at that point, I wonder if we should use the `clipboard` package 😅

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+    }),
+    onClick: handleClick,
+    onKeyDown: handleKeyDown,
+    ref: buttonRef,
+    disabled,
+    ...rest,
+  } as const;
+
+  return (
+    <Tooltip
```
stephl3, line 161: We previously controlled this tooltip instance in `Code` because `trigger` is a prop of `Tooltip`. When the `copied` state toggles and the `trigger` needs to re-render, it would also remount the `Tooltip` instance itself

I wonder if we might need to keep that behavior here for both variants

Here's the PR fixing that issue in `Code`: https://github.com/mongodb/leafygreen-ui/pull/2232

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.styles.ts
```diff
@@ -0,0 +1,133 @@
+import { css, cx } from '@leafygreen-ui/emotion';
+import { Theme } from '@leafygreen-ui/lib';
+import { palette } from '@leafygreen-ui/palette';
+import { color, transitionDuration } from '@leafygreen-ui/tokens';
```
stephl3, line 4: nit: observing some opportunities to evolve to tokens in place of style objects and palette

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.styles.ts
```diff
@@ -0,0 +1,133 @@
...
+  );
+
+/**
+ * Theme-specific styles applied when the button is in the "copied" state.
+ * Shows green background and white text to indicate successful copy operation.
+ */
+export const copiedThemeStyle: Record<Theme, string> = {
+  [Theme.Light]: css`
+    &,
+    & > div > svg {
```
stephl3, line 64: I know these are from the `Code` component, but I'm not a fan of these nested selectors because it's coupled to the structure of `Tooltip`. Is it possible to use the icon `fill` instead for the `svg` selectors?

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.types.ts
tsck: Oh good catch! I missed this, will remove

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.types.ts
tsck: I think I like the latter. It's more descriptive than what I have. The former seems less clear to me since it sounds almost like just a click handler.

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.tsx
tsck: Honestly, I think this was AI just being extra careful. I left it because I figured why not have a fallback, but looking at the support, it seems safe to remove it.

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.tsx
tsck: Oh no, not sure why I removed that. Was an accident, will add it back in.

tsck: Ah, because I removed the open and close states and just relied on the tooltip to open and close. It seemed unnecessary but if there was a reason, I can add that all back.

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.tsx
tsck: No issue with me. Will add this all back as per https://github.com/mongodb/leafygreen-ui/pull/3041#discussion_r2283406243

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.styles.ts
tsck: I'll look at these on the Panel PR so that I can see the button being rendered and make sure it's correct. https://github.com/mongodb/leafygreen-ui/pull/3047

packages/code-editor/src/CodeEditorCopyButton/CodeEditorCopyButton.styles.ts
tsck: I'll address this on the Panel PR as well since I'm scared to update this without rendering it https://github.com/mongodb/leafygreen-ui/pull/3047

PR #3043, opened 2025-08-14
LG-5444 - Toolbar pt 2: adds copy button to editor

tsck:
## ✍️ Proposed changes
- Introduces `copyButtonAppearance` prop to control the visibility of the copy button in the CodeEditor.
- Implements styles for different copy button appearances: 'hover', 'persist', and 'none'.
- Updates tests to verify the rendering behavior of the copy button based on the new prop.
- Refactors CodeEditor component to integrate the new copy button functionality and ensure proper rendering based on user interactions.

Drive by:
- Updated the `useCodeFoldingExtension` to use dynamic spacing for icon margins and set custom dimensions for icons. This is just to align with styles so I through it in here.
- Enhanced the `useThemeExtension` by adding padding to improve layout consistency.

🎟 _Jira ticket:_ [LG-5444](https://jira.mongodb.org/browse/LG-5444)

## 🧪 How to test changes
- Previous unit tests passing
- Added unit tests passing
- Manual QA in storybook


tsck:
@TheSonOfThomp Just a heads up that this currently based on another branch, not the integration branch. It's a chain of branches I broke up from one larger one that I intend on merging consecutively.

packages/code-editor/src/CodeEditor/CodeEditor.styles.ts
```diff
@@ -1,13 +1,24 @@
...
+  spacing,
+  transitionDuration,
   Variant,
 } from '@leafygreen-ui/tokens';
 
-import { CodeEditorSelectors } from './CodeEditor.types';
+import { CodeEditorSelectors, CopyButtonAppearance } from './CodeEditor.types';
+
+export const copyButtonClassName = createUniqueClassName(
+  'code_editor_copy_button',
```
stephl3: was going to comment about the formatting of this but tbh not sure what is the correct pattern. I tend to follow the "bem-ish" approach described for `data-lgid` and `data-testid`: https://github.com/mongodb/leafygreen-ui/blob/main/STYLEGUIDE.md#follow-bem-ish-patterns-when-hard-coding-a-data-testid-or-data-lgid

I also see some cases where we use straight kebab-case in the repo

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -147,9 +160,20 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
           minHeight,
           maxHeight,
           className,
+          copyButtonAppearance,
         })}
         {...rest}
       >
+        {(copyButtonAppearance === CopyButtonAppearance.Hover ||
+          copyButtonAppearance === CopyButtonAppearance.Persist) && (
```
stephl3, line 168: might we want to only have this show up if there is contents to be copied?

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
```diff
@@ -34,6 +34,19 @@ export interface HTMLElementWithCodeMirror extends HTMLDivElement {
...
+ */
+export const CopyButtonAppearance = {
+  Hover: 'hover',
+  Persist: 'persist',
+  None: 'none',
+} as const;
+export type CopyButtonAppearance =
+  (typeof CopyButtonAppearance)[keyof typeof CopyButtonAppearance];
+
+export const CopyButtonLgId = 'lg-code_editor-code_editor_copy_button';
```
stephl3, line 48:
```suggestion
export const CopyButtonLgId = 'lg-code_editor-copy_button';
```

packages/code-editor/src/CodeEditor/hooks/extensions/useCodeFoldingExtension.tsx
```diff
@@ -55,15 +56,19 @@ export function useCodeFoldingExtension({
                 glyph="ChevronDown"
                 size="small"
                 className={css`
-                  margin-top: 2px;
+                  margin-top: ${spacing[100]}px;
+                  height: 10px; // Custom icon height
+                  width: 10px; // Custom icon width
```
stephl3: nit: could turn these into constants and reuse for both cases

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
tsck: This was somewhat intentional. Technically the component name is `CodeEditorCopyButton` and not just `CopyButton`. I don't feel strongly about what this is, but I think the current form is _technically_ correct

packages/code-editor/src/CodeEditor/CodeEditor.styles.ts
tsck: I can update this to whatever the lgid ends up being below. Does that make sense?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: I'm not sure. Technically, one can copy "" which an empty editor. The copy button that ends up on the panel is not disabled when there's no content, so I'm not sure we should hide this.

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
stephl3: If `CopyButton` were imported from a separate package, would the assigned `data-lgid` still be `lg-code_editor-code_editor_copy_button` vs just `lg-code_editor-copy_button`?

packages/code-editor/src/CodeEditor/hooks/extensions/useCodeFoldingExtension.tsx
```diff
@@ -55,15 +58,19 @@ export function useCodeFoldingExtension({
                 glyph="ChevronDown"
                 size="small"
                 className={css`
-                  margin-top: 2px;
+                  margin-top: ${spacing[100]}px;
+                  height: ${CUSTOM_ICON_SIZE}px;
+                  width: ${CUSTOM_ICON_SIZE}px;
```
stephl3: could we pass this value to `size` instead?
`size={CUSTOM_ICON_SIZE}`

packages/code-editor/src/CodeEditor/hooks/extensions/useCodeFoldingExtension.tsx
```diff
@@ -55,15 +58,19 @@ export function useCodeFoldingExtension({
...
               />
             ) : (
               <Icon
                 glyph="ChevronRight"
                 size="small"
                 className={css`
-                  margin-top: 2px;
+                  margin-top: ${spacing[100]}px;
+                  height: 10px;
+                  width: 10px;
```
stephl3: probably should keep this consistent with what we do for ChevronDown

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
tsck: My thinking was just that it was basically `lg-package_name-component_name`. So even if `CodeEditorCopyButton` was just renamed to `CopyButton` I would think the lgid would be `lg-code_editor-copy_button`. If this thinking is not in line with what we normally do please do correct me though. I haven't had to add many LGIDs.

packages/code-editor/src/CodeEditor/hooks/extensions/useCodeFoldingExtension.tsx
tsck: Oops, totally missed this, good catch!

packages/code-editor/src/CodeEditor/hooks/extensions/useCodeFoldingExtension.tsx
tsck: I admittedly did not realize that we could just pass in an arbitrary number 😄 makes sense

PR #3047, opened 2025-08-15
[LG-5094] feat(code-editor): add panel component for toolbar functionality

tsck:
## ✍️ Proposed changes

This PR introduces a new **Panel component** to the `@leafygreen-ui/code-editor` package, providing a comprehensive toolbar interface for the CodeEditor with formatting, copying, and custom action buttons.

The Panel component includes:
- A configurable title display for code language or content description
- Format button for code prettification (when formatting is available)
- Copy button for copying code to clipboard using the existing CodeEditorCopyButton
- Secondary menu with additional actions (undo, redo, download, view shortcuts)
- Support for custom secondary buttons for application-specific actions
- Custom children content support in the left slot
- Full accessibility support with proper ARIA labels
- Integration with CodeEditor context for seamless copy functionality

This enhancement significantly improves the user experience by providing easy access to common editor actions through a clean, organized toolbar interface.

🎟️ _Jira ticket:_ N/A

- [x] I have added/updated tests for this change
- [x] I have added/updated TypeScript types for this change  
- [x] I have added/updated documentation (Storybook) for this change
- [x] I have added a changeset for this change
- [x] I have tested this change in isolation

## 🧪 How to test changes

1. **Basic Panel Rendering:**
   ```tsx
   import { CodeEditor, Panel } from '@leafygreen-ui/code-editor';
   
   <CodeEditor
     defaultValue="const greeting = 'Hello World';"
     language="javascript"
     panel={
       <Panel 
         title="JavaScript"
         showFormatButton
         showCopyButton
         showSecondaryMenuButton
       />
     }
   />
   ```

2. **Test Copy Functionality:**
   - Render a CodeEditor with Panel and `showCopyButton={true}`
   - Click the copy button
   - Verify content is copied to clipboard

3. **Test Secondary Menu:**
   - Render Panel with `showSecondaryMenuButton={true}`
   - Click the ellipsis button to open the menu
   - Verify all menu items (Undo, Redo, Download, View shortcuts) are present and accessible

4. **Test Custom Secondary Buttons:**
   ```tsx
   <Panel
     showSecondaryMenuButton
     customSecondaryButtons={[
       {
         label: 'Custom Action',
         glyph: <CloudIcon />,
         onClick: () => console.log('custom clicked'),
         'aria-label': 'Perform custom action'
       }
     ]}
   />
   ```

5. **Test Accessibility:**
   - Navigate the Panel using keyboard only
   - Verify all buttons and menu items have proper ARIA labels
   - Test with screen reader to ensure proper announcements

6. **Test Integration:**
   - Verify Panel integrates properly with CodeEditor context
   - Test that the Panel styling matches the CodeEditor theme
   - Ensure responsive behavior across different screen sizes

### DIFF
--- a/packages/code-editor/src/Panel/Panel.styles.ts
+++ b/packages/code-editor/src/Panel/Panel.styles.ts
@@ -42,7 +42,7 @@ export const getPanelTitleStyles = (theme: Theme, baseFontSize: number) => {
 
 export const getPanelInnerContentStyles = () => {
   return css`
-    grid-area: innerContent;
+    grid-area: inner-content;
   `;
 };
 
--- a/packages/code-editor/src/Panel/Panel.types.ts
+++ b/packages/code-editor/src/Panel/Panel.types.ts
@@ -1,3 +1,5 @@
+import { ReactElement, ReactNode } from 'react';
+
 import { DarkModeProps } from '@leafygreen-ui/lib';
 
 /**
@@ -12,7 +14,7 @@ export interface SecondaryButtonConfig {
   /**
    * Optional icon element to display in the button.
    */
-  glyph?: React.ReactElement;
+  glyph?: ReactElement;
 
   /**
    * Optional callback fired when the button is clicked.
@@ -55,7 +57,7 @@ export interface PanelProps extends DarkModeProps {
    * Optional React node to render between the title and the buttons.
    * Can be used to add custom controls to the panel.
    */
-  innerContent?: React.ReactNode;
+  innerContent?: ReactNode;
 
   /**
    * Optional callback fired when the copy button is clicked.
@@ -116,5 +118,5 @@ export interface PanelProps extends DarkModeProps {
    * Optional title text to display in the panel header.
    * Typically used to show the current language or content description.
    */
-  title?: string;
+  title?: ReactNode;
 }


tsck:
@stephl3 This is the `Panel` addition that uses the copy button. Note that this is based off of `LG-5444/ce-add-copy-to-editor` and not `main` as a part of this long chain of PR's to ultimately go into the integration branch. I'll rebase as needed as things get merged.

packages/code-editor/README.md
```diff
@@ -68,12 +68,67 @@ console.log(greet('MongoDB user'));`;
...
 | `width` _(optional)_                | Sets the editor's width. If not set, the editor will be 100% width of its parent container.                                                                                                                                                                                                                                                                                                                                                                                                                                    | `string`                     | `undefined` |
 
+### `<Panel>`
+
+The Panel component provides a toolbar interface for the CodeEditor with formatting, copying, and custom action buttons. It displays at the top of the CodeEditor and can include a title, action buttons, and custom content.
```
stephl3, line 80: what do you think about adding a note about what functionality is provided out-of-the-box vs what consumers will need to implement?

stephl3: actually, will all of these actions be supported out-of-the-box? (I might be incorrectly assuming based off of having only reviewed copy)

packages/code-editor/src/Panel/Panel.types.ts
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Configuration object for custom secondary buttons that appear in the Panel's secondary menu.
+ * Extends MenuItemProps to inherit all MenuItem functionality while adding required label.
+ */
+type SecondaryButtonConfig = MenuItemProps & {
+  /**
+   * Text label for the button that will be displayed in the menu item.
+   */
+  label: string;
```
stephl3, line 12: what's the reason for using a separate prop vs using `children` which is on `MenuItemProps`? If not using `children`, I think we should `Omit` it

packages/code-editor/src/Panel/Panel.types.ts
```diff
@@ -0,0 +1,118 @@
+import { DarkModeProps } from '@leafygreen-ui/lib';
+import { type MenuItemProps } from '@leafygreen-ui/menu';
+
+/**
+ * Configuration object for custom secondary buttons that appear in the Panel's secondary menu.
+ * Extends MenuItemProps to inherit all MenuItem functionality while adding required label.
+ */
+type SecondaryButtonConfig = MenuItemProps & {
```
stephl3: is it necessary to include some of these props? looks like some are already part of `MenuItemProps`

packages/code-editor/src/Panel/Panel.types.ts
```diff
@@ -0,0 +1,118 @@
...
+   * Determines whether to show the secondary menu button (ellipsis icon) in the panel.
+   * When enabled, displays a menu with additional actions like undo, redo, download, and view shortcuts.
+   */
+  showSecondaryMenuButton?: boolean;
+
+  /**
+   * Optional title text to display in the panel header.
+   * Typically used to show the current language or content description.
+   */
+  title?: string;
```
stephl3: any reason not to use `ReactNode`?

packages/code-editor/src/Panel/Panel.styles.ts
```diff
@@ -0,0 +1,50 @@
...
+  InteractionState,
+  Variant,
+} from '@leafygreen-ui/tokens';
+
+export const getPanelStyles = (theme: Theme) => {
+  return css`
+    background-color: ${color[theme].background[Variant.Secondary][
+      InteractionState.Default
+    ]};
+    height: 36px;
```
stephl3: nit: looks like a magic number worth turning into constant

packages/code-editor/src/Panel/Panel.styles.ts
```diff
@@ -0,0 +1,50 @@
...
+    width: 100%;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    padding: 0 16px 0 10px;
```
stephl3: tokens?

packages/code-editor/src/Panel/Panel.styles.ts
```diff
@@ -0,0 +1,50 @@
...
+  return css`
+    grid-area: title;
+    font-size: ${baseFontSize}px;
+    color: ${color[theme].text[Variant.Secondary][InteractionState.Default]};
+  `;
+};
+
+export const getPanelInnerContentStyles = () => {
+  return css`
+    grid-area: children;
```
stephl3: hmm I see that this `grid-area` isn't necessary anyways, but I'd still hesitate on naming this `children` to avoid creating confusion between `innerContent` vs `children`

packages/code-editor/README.md
tsck: Yup, all of these actions will be provided! Only added secondary actions won't be.

packages/code-editor/src/Panel/Panel.styles.ts
tsck: Admittedly just copied over from `code/Panel`, will update

packages/code-editor/src/Panel/Panel.styles.ts
tsck: Ah good catch. This prop originally didn't exist and I was relying on `children` and neglected to rename this. Will update

packages/code-editor/src/Panel/Panel.types.ts
tsck: I didn't use `children` because this is a config object and not a component. We could have that prop on the object, but it seems funny to me to have an object prop called `children`.  If you feel otherwise let me know. Happy to omit it though!

packages/code-editor/src/Panel/Panel.types.ts
tsck: Nope, just a miss on my end I think. Will update

packages/code-editor/src/Panel/Panel.types.ts
tsck: Given the `innerContent` slot, this felt like it was explicitly for when a consumer wants to add a pre-formatted title to the panel. I don't have a hard issue with making it a node, but `innerContent` is already a node and will take up the entire left side if `title` is omitted.

packages/code-editor/src/Panel/Panel.types.ts
tsck: Ok, this was a bit weird. Since `MenuItemProps` is polymorphic, it made using both `href` and `onClick` a bit wonky, causing TS to get upset when passing them down to `MenuItem`. Not sure if there's a better way to handle this, but I just decided to not extend `MenuItemProps` at all, and instead define a subset of props that can be passed to `MenuItem`. I feel like this is fine, but am also open to ideas.

packages/code-editor/src/Panel/Panel.types.ts
stephl3: Gotcha, I might just make this a `ReactNode` anyways. If it's not and a team needs to pass in a node, they may end up ignoring `title` and abusing `innerContent`

packages/code-editor/src/Panel/Panel.styles.ts
```diff
@@ -0,0 +1,53 @@
...
+  return css`
+    grid-area: title;
+    font-size: ${baseFontSize}px;
+    color: ${color[theme].text[Variant.Secondary][InteractionState.Default]};
+  `;
+};
+
+export const getPanelInnerContentStyles = () => {
+  return css`
+    grid-area: innerContent;
```
stephl3: nit: naming convention from mdn https://developer.mozilla.org/en-US/docs/Web/CSS/grid-area#custom-ident
```suggestion
    grid-area: inner-content;
```

packages/code-editor/src/Panel/Panel.types.ts
tsck: Why not. I'm not inherently opposed so I'll just make it a `ReactNode`

tsck:
I noticed that I forgot to update `innerContent` to `inner-content` in the `grid-template-areas` prop. Will update on forthcoming branch `LG-5128/ce-add-formatting`

PR #3058, opened 2025-08-19
[LG-5128] feat(code-editor): add code formatting functionality with multi-language support

tsck:
## ✍️ Proposed changes

This PR adds comprehensive code formatting functionality to the CodeEditor component, enabling users to format their code using industry-standard formatters for multiple programming languages. The implementation introduces a new lazy-loading architecture for formatting modules to minimize bundle size impact while supporting a wide range of languages through Prettier and WASM-based formatters.

Key additions include:
- **New formatting system**: Introduced `useFormattingModuleLoaders` and `useCodeFormatter` hooks for on-demand loading of formatting libraries
- **Multi-language support**: Added support for JavaScript/TypeScript, CSS, HTML, Java, Kotlin, PHP, Ruby, Rust, C++, C#, Go, and Python formatting
- **Panel integration**: Enhanced the Panel component with formatting functionality, including proper disabled states and error handling
- **Performance optimization**: Formatters are loaded only when needed, preventing unnecessary bundle bloat
- **Comprehensive testing**: Added extensive test coverage for formatting module loaders and error handling scenarios

The solution provides a clean API for consumers while maintaining compatibility with existing CodeEditor functionality.

- [x] I have added a changeset
  - [ ] If this is a visual change, I have added a screenshot or video to the PR
  - [ ] If this change affects layouts or text, I have tested in all supported browsers at mobile, tablet, and desktop screen sizes
- [x] If this is a new feature, I have added proper documentation (JSDocs for components, Storybook for visual components)
- [x] I have checked that all existing tests are still passing
- [x] I have written new tests for the new functionality
- [x] If this is a breaking change, I have added a note in the changeset and in the README
- [x] If this change should be part of a coordinated release with other packages, I have added the appropriate changeset format

## 🧪 How to test changes

1. **Test basic formatting functionality**:
   - Open CodeEditor with JavaScript/TypeScript code
   - Verify the format button appears in the Panel
   - Click the format button and confirm code is properly formatted using Prettier

2. **Test multi-language support**:
   - Switch between different supported languages (JS, TS, CSS, HTML, Java, etc.)
   - Verify formatting works correctly for each language
   - Test that unsupported languages gracefully disable the format button

3. **Test error handling**:
   - Try formatting malformed code to ensure graceful error handling
   - Verify console warnings appear for missing optional formatter plugins
   - Confirm the format button properly disables when formatting is not available

4. **Test performance**:
   - Verify that formatting modules are only loaded when the format button is clicked
   - Check that initial CodeEditor load time is not significantly impacted
   - Confirm that switching languages doesn't cause unnecessary re-renders

5. **Test accessibility**:
   - Verify the format button has proper ARIA labels
   - Test keyboard navigation for formatting functionality
   - Ensure disabled states are properly communicated to screen readers


packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
```diff
@@ -0,0 +1,279 @@
...
+              )
+            : createClangFormatConfig(
+                'Microsoft',
+                options,
+                {
+                  BreakBeforeBraces: 'Allman',
+                },
+                editorTabWidth,
+                editorUseTabs,
+              );
```
tsck: These settings were primarily set by Cursor, with some tweaks if something looked off when I formatted that language. I've left them as is because they seem "correct" to me visually.

packages/code-editor/src/CodeEditor/hooks/formatting/types.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface FormattingOptions {
+  semi?: boolean;
+  singleQuote?: boolean;
+  printWidth?: number;
+  trailingComma?: 'none' | 'es5' | 'all';
+  bracketSpacing?: boolean;
+  jsxBracketSameLine?: boolean;
+  arrowParens?: 'avoid' | 'always';
+  // Note: tabWidth and useTabs are derived from CodeEditor props (indentSize, indentUnit)
+}
```
tsck, line 13: These were again chosen by Cursor. Most of them only apply to Prettier formatter languages but I made sure that was added in the docs.

packages/code-editor/src/CodeEditor/hooks/useModuleLoaders.ts
tsck, line 1: This was a bit of a drive-by. In `useFormattingModuleLoaders`, defining the dynamic import functions inside of the memo triggered an infinite loop downstream since `loaders` then was always deemed new, which indicated changes, which then thought it needed new `loaders` etc. I didn't see this behavior with `useModuleLoaders` but this seemed safer so that that doesn't occur.

packages/code-editor/src/CodeEditor/testing/codeSnippets.ts
tsck, line 1: Another small drive-by here. I noticed that `tsx` and `jsx` snippets were not added, but technically they are separate supported languages so I thought they should be.

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -150,10 +215,15 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
     useImperativeHandle(forwardedRef, () => ({
       getEditorViewInstance: () => editorViewRef.current,
       getContents,
+      formatCode: handleFormatCode,
+      isFormattingAvailable,
     }));
```
tsck, line 232: I'm continuing to pass functionality down for imperative use. It seems like this is helpful so that consumers can trigger things in the editor imperatively. I don't know if this is necessary, but I figured it adds flexibility on how one can use the component. Definitely open to feedback here though.

packages/code-editor/src/Panel/Panel.styles.ts
```diff
@@ -28,7 +28,7 @@ export const getPanelStyles = (theme: Theme) => {
     border-top-right-radius: ${borderRadius[300]}px;
     display: grid;
     grid-template-columns: auto 1fr auto;
-    grid-template-areas: 'title innerContent buttons';
+    grid-template-areas: 'title inner-content buttons';
```
tsck, line 31: Drive-by fix from https://github.com/mongodb/leafygreen-ui/pull/3047 ([Explaining comment](https://github.com/mongodb/leafygreen-ui/pull/3047#issuecomment-3210847797))

packages/code-editor/src/CodeEditor.stories.tsx
```diff
@@ -320,6 +321,12 @@ Json.args = {
   defaultValue: codeSnippets.json,
 };
 
+export const Jsx = Template.bind({});
+Jsx.args = {
+  language: 'jsx',
+  defaultValue: codeSnippets.jsx,
+};
+
```
tsck, line 329: Drive-by: Adding JSX story

packages/code-editor/src/CodeEditor.stories.tsx
```diff
@@ -350,6 +357,12 @@ Rust.args = {
   defaultValue: codeSnippets.rust,
 };
 
+export const Tsx = Template.bind({});
+Tsx.args = {
+  language: 'tsx',
+  defaultValue: codeSnippets.tsx,
+};
+
```
tsck, line 365: Drive-by: Adding TSX story

.changeset/violet-queens-train.md
```diff
@@ -0,0 +1,39 @@
+---
+'@leafygreen-ui/code-editor': minor
```
stephl3, line 2: hmm something I'm only just realizing: are these changesets necessary / relevant to consumers since the package is not yet released to consumers? would the README be sufficient for the initial release? (excuse me if I've asked this before 😅 you may have had a reason for this that I forget about)

packages/code-editor/README.md
```diff
@@ -140,21 +140,197 @@ import CloudIcon from '@leafygreen-ui/icon';
 | `label`                   | Text label for the button.                                             | `string`             | —           |
 | `onClick` _(optional)_    | Callback fired when the button is clicked.                             | `() => void`         | `undefined` |
 
+## Code Formatting
```
stephl3, line 143: I get the importance of "Code Formatting" having its own section, but I feel like the IA of this README is a bit disjointed since there's a section on `useCodeFormatter` under the "Feature Extension Hooks" section. Is it worth consolidating these or maybe even link to that section to connect the 2?

packages/code-editor/README.md
```diff
@@ -140,21 +140,197 @@ import CloudIcon from '@leafygreen-ui/icon';
...
+- **JSON**
+- **Java**
+- **C++**
+- **C#**
+- **Go**
+- **Python**
+
+**Note:** Kotlin, PHP, Ruby, and Rust are not supported due to current browser compatibility limitations.
+
+### Module Loading Requirements
```
stephl3: This is specific to the case of the consumer using the `useCodeFormatter` hook directly, correct? I'm wondering if it might make more sense to house this as a subsection/note under "#### 3. Via useCodeFormatter Hook (Standalone)"

packages/code-editor/README.md
```diff
@@ -140,21 +140,197 @@ import CloudIcon from '@leafygreen-ui/icon';
...
+interface FormattingOptions {
+  semi?: boolean; // Add semicolons
+  singleQuote?: boolean; // Use single quotes
+  printWidth?: number; // Line length
+  trailingComma?: 'none' | 'es5' | 'all'; // Trailing commas
+  bracketSpacing?: boolean; // Spaces around object brackets
+  jsxBracketSameLine?: boolean; // JSX bracket placement
+  arrowParens?: 'avoid' | 'always'; // Arrow function parentheses
+}
+```
```
stephl3: I know we were talking about this during dev sync earlier this week. Is it worth moving this to the below "## Types and Variables" section or linking between the 2?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -69,19 +76,90 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
...
+    // Load modules
+    const coreModuleLoaders = useModuleLoaders(props);
+    const formattingModuleLoaders = useFormattingModuleLoaders(language);
+    const allModuleLoaders = useMemo(
+      () => ({
+        ...coreModuleLoaders,
+        ...formattingModuleLoaders,
+      }),
+      [coreModuleLoaders, formattingModuleLoaders],
+    );
```
stephl3: is there some benefit to having 2 separate hooks for getting these moduleLoaders?  what's the reason stopping us from doing:
```
const moduleLoaders = useModuleLoaders(props);
```

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -69,19 +76,90 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
...
+    const { isLoading, modules } = useLazyModules(allModuleLoaders);
```
stephl3: oh wait I may have just answered my prior question: should this be calling `useLazyModules` in 2 separate chunks and then we only pass formatting modules to `useCodeFormatter`?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -69,19 +76,90 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
...
+    const handleFormatCode = useCallback(async (): Promise<string> => {
+      const currentContent = getContents();
+
+      if (!isFormattingAvailable) {
+        console.warn('Formatting is not available for the current language');
+        return currentContent;
+      }
+
+      try {
+        const formattedContent = await formatCode(currentContent);
```
stephl3, line 129: Is this `formatCode` invocation supposed to have a 2nd arg for the formatting options? are we not allowing that for now following our earlier discussion?

stephl3: Another idea I had (which is totally scope creep but may be worth thinking about): what does it look like for the end user to be able to alter their formatting settings in the more actions or at an app level? 🤔

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -69,19 +76,90 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
...
+              to: editorViewRef.current.state.doc.length,
+              insert: formattedContent,
+            },
+          });
+          editorViewRef.current.dispatch(transaction);
+
+          // Update controlled value if in controlled mode
+          if (isControlled) {
+            setControlledValue(formattedContent);
+            onChangeProp?.(formattedContent);
```
stephl3: Should `onChangeProp` only be called when the component is controlled? This contradicts the pattern I've seen elsewhere in LG and think it should be getting called regardless of controlled or not

packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
```diff
@@ -0,0 +1,279 @@
...
+  createPrettierConfig,
+} from './utils';
+
+/**
+ * Helper function to format code using Prettier
+ */
+const formatWithPrettier = (
+  code: string,
+  parserModuleName: keyof CodeEditorModules,
+  config: any,
```
stephl3: is it worth having more restrictive types here? maybe a comment if not?

packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * Helper function to format code using WASM formatters
+ */
+const formatWithWasm = async <T extends Array<any>>(
+  code: string,
+  wasmModuleName: keyof CodeEditorModules,
+  modules: Partial<CodeEditorModules>,
+  warningMessage: string,
+  formatFn: (formatter: any, code: string, ...args: T) => string,
```
stephl3: similar question about the `any` type here: can we make it more restrictive or add a comment?

packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
```diff
@@ -0,0 +1,279 @@
...
+import {
+  createClangFormatConfig,
+  createJavaScriptConfig,
+  createPrettierConfig,
+} from './utils';
+
+/**
+ * Helper function to format code using Prettier
+ */
+const formatWithPrettier = (
```
stephl3: nit: for functions with more than two or three arguments, I find an options object really improves readability

packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
```diff
@@ -0,0 +1,279 @@
...
+    return code;
+  }
+
+  return prettier.format(code, config);
+};
+
+/**
+ * Helper function to format code using WASM formatters
+ */
+const formatWithWasm = async <T extends Array<any>>(
```
stephl3: nit: single object arg for readability

packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
```diff
@@ -0,0 +1,279 @@
...
+ *
+ * @param code - The code string to format
+ * @param language - The programming language of the code
+ * @param options - Optional formatting options
+ * @param editorTabWidth - Tab width from editor configuration
+ * @param editorUseTabs - Whether to use tabs from editor configuration
+ * @param modules - Module dependencies containing formatting modules
+ * @returns Promise resolving to formatted code, or original code if formatting fails/unavailable
+ */
+export const formatCode = async (
```
stephl3: nit: single object arg for readability

packages/code-editor/src/CodeEditor/hooks/formatting/utils.ts
```diff
@@ -0,0 +1,130 @@
+import { LanguageName } from '../extensions/useLanguageExtension';
+import { CodeEditorModules } from '../moduleLoaders.types';
+
+import { FormattingOptions } from './types';
+
+/**
+ * Creates base Prettier configuration with common settings
+ */
+export const createPrettierConfig = (
```
stephl3: nit: single object arg for readability

packages/code-editor/src/CodeEditor/hooks/formatting/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+  useTabs: editorUseTabs,
+  printWidth: options.printWidth ?? 80,
+  ...options,
+  ...overrides,
+});
+
+/**
+ * Creates shared JavaScript/TypeScript configuration
+ */
+export const createJavaScriptConfig = (
```
stephl3: nit: single object arg for readability

packages/code-editor/src/CodeEditor/hooks/formatting/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+      arrowParens: options.arrowParens ?? 'avoid',
+    },
+    editorTabWidth,
+    editorUseTabs,
+  );
+
+/**
+ * Creates base clang-format configuration for C++/Java/C# languages
+ */
+export const createClangFormatConfig = (
```
stephl3: nit: single object arg for readability

packages/code-editor/src/Panel/Panel.tsx
```diff
@@ -104,7 +115,7 @@ export function Panel({
             align="top"
             justify="middle"
             trigger={
-              <IconButton onClick={onFormatClick} aria-label="Format code">
+              <IconButton onClick={handleFormatClick} aria-label="Format code">
```
stephl3, line 118: More design feedback but wondering what you think about including more visual feedback that it successfully formatted? The contrast between this and the copy button succeeding makes this feel more drastic

.changeset/violet-queens-train.md
tsck: Haha no, it's a good point. I think this was something I discussed with Adam, maybe? And have just been adding each changeset as I go. I guess it doesn't hurt. They will all be added into the integration branch which will ultimately have a major bump and just end up at v1 anyway. Do you see any harm in doing it that way?

packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
tsck: I would completely agree! Will update

packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
tsck: I believe so. I'll see what I can do here.

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: I did think about this. I ultimately just went with the two in order to not make `useModuleLoaders` too big. Since there was no overlap as in, none of the modules needed for formatting are needed elsewhere, I just separated the logic out. Do you feel there would be a benefit to combining them? I'm definitely open to it

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: That was the original design! I actually changed it to this last minute because I thought combining them in one call to `useLazyModules` didn't have any real downside and was a bit cleaner. Is there any advantage to calling it twice as opposed to just combining all of the modules?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Good catch here. Currently there's no API to adjust formatting options at the component level. It seemed odd to have a `formattingOptions` prop on the `CodeEditor` when the formatting button is on the `Panel`. Maybe this speaks a bit more to what we spoke about offline, potentially not allowing formatting options to be adjusted at all. How do you feel about that?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Good catch, I totally agree. I'll update this

packages/code-editor/src/Panel/Panel.tsx
tsck: I think this is a good point. I'll chat with Sandy (probably Monday) and see what she thinks here

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: I'm going to make the change to prevent formatting options from being passed in. We can discuss further after I re-request if you disagree.

packages/code-editor/README.md
tsck: I think that's a good point, will update

packages/code-editor/README.md
tsck: It definitely should be listed down there, I agree, will update accordingly

packages/code-editor/README.md
tsck: I think this is a great point. It's also in line with another change I made in a future PR where I create an "Imperative API" section and just link to that ([change](https://github.com/mongodb/leafygreen-ui/pull/3063/files#diff-c84937ab77adcf09e406aa3aeb30a7be80c5b744c84debb78dccbd8d22cbe852R188)). I'll move the hook definition completely under the hooks section and just link to it there.

packages/code-editor/README.md
tsck: This is removed now that I removed the ability to override options

tsck:
@stephl3 Sorry for having to rebase, but I think/hope things are back in order, and should be ready for another look. The commits that came after your last review start at [b20b1f4](https://github.com/mongodb/leafygreen-ui/pull/3058/commits/b20b1f46cdec7d20ff79c5c6291b9e4508ccc7b2).

.changeset/violet-queens-train.md
stephl3: No real harm, but it does create a duplicate release

1. When integration branch merges into main, v0.1.0 publishes
2. Version packages PR is created consolidating all changesets
3. When version packages PR merges into main, v1.0.0 publishes with no real differences from v0.1.0

packages/code-editor/src/CodeEditor/CodeEditor.tsx
stephl3: Fine with either. Since `useModuleLoaders` isn't called anywhere else, it seems like it could be consolidated, but not a big deal to keep them separate as well

packages/code-editor/src/CodeEditor/CodeEditor.tsx
stephl3: This one does seem like it could make sense to split into separate chunks. I'd be curious:
1. does formatting functionality depend on core modules being loaded?
2. do we care about formatting modules blocking core modules being loaded? no difference in load time?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
stephl3: that sounds good to me! starting restrictive and then opening up for more flexibility

packages/code-editor/src/CodeEditor/hooks/formatting/formatters.ts
```diff
@@ -0,0 +1,338 @@
...
+const formatWithPrettier = ({
+  code,
+  parserModuleName,
+  config,
+  modules,
+  warningMessage,
+}: {
+  code: string;
+  parserModuleName: keyof CodeEditorModules;
+  config: PrettierConfig;
```
stephl3, line 66: 🥇

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: I don't think there would be an impact to load time because we have to wait until everything is loaded and wired before we render anything anyway. I think this is more of a micro-optimization, but I'm happy to revert back to the way it was and just keep both loads totally separate. Will update.

.changeset/violet-queens-train.md
tsck: In order to actually change this, I'd need to merge a change into the base branch and get rid of all the change logs. I'll make sure this is considered before merging into main

PR #3061, opened 2025-08-20
[LG-5414] feat(message-feedback): enable async form submission and support error message customization

stephl3:
## ✍️ Proposed changes

This PR enhances the chat message feedback system by adding robust async form submission capabilities and comprehensive error handling. The main changes include replacing the boolean `isSubmitted` prop with a more flexible `state` enum in `InlineMessageFeedback`, and adding error handling support to `MessageActions` with customizable error messages.

Key improvements:
- **Breaking Change**: `InlineMessageFeedback` now uses a `state` enum (`'unset'`, `'submitting'`, `'error'`, `'submitted'`) instead of the `isSubmitted` boolean
- **Async Support**: Both components now handle Promise-based form submissions
- **Error Handling**: Automatic error state management with customizable error messages
- **Loading States**: Form elements properly disable during submission
- **Accessibility**: Fixed label/textarea relationship and improved ARIA attributes

🎟️ _Jira ticket:_ [LG-5414](https://jira.mongodb.org/browse/LG-5414)

## ✅ Checklist

### Testing InlineMessageFeedback (Breaking Change)
1. Update any existing usage from `isSubmitted={true}` to `state="submitted"`
2. Test async form submission by providing an async `onSubmit` handler
3. Test error handling by throwing an error in the `onSubmit` handler
4. Verify error message displays and form remains interactive for retry

### Testing MessageActions (New Features)
1. Provide an async `onSubmitFeedback` callback that can throw errors
2. Add an `errorMessage` prop to customize error text
3. Test the complete flow: rating selection → feedback input → submission → error/success states
4. Verify form disables during submission and shows loading indicators
5. Test retry functionality after submission failures

### Storybook Testing
- Check the new error handling stories in both components
- Verify success and error state transitions work correctly
- Test both light and dark mode variants

### DIFF
--- a/chat/message-actions/src/MessageActions/MessageActions.tsx
+++ b/chat/message-actions/src/MessageActions/MessageActions.tsx
@@ -55,7 +55,7 @@ export function MessageActions({
   const [rating, setRating] = useState<MessageRatingValue>(
     MessageRatingValue.Unselected,
   );
-  const [feedback, setFeedback] = useState<string | undefined>(undefined);
+  const [feedback, setFeedback] = useState<string>('');
   const [feedbackFormState, setFeedbackFormState] = useState<FormState>(
     FormState.Unset,
   );


chat/message-actions/src/MessageActions/MessageActions.tsx
```diff
@@ -55,7 +56,19 @@ export function MessageActions({
     MessageRatingValue.Unselected,
   );
   const [feedback, setFeedback] = useState<string | undefined>(undefined);
-  const [isSubmitted, setIsSubmitted] = useState(false);
+  const [feedbackFormState, setFeedbackFormState] = useState<FormState>(
+    FormState.Unset,
+  );
+  const [showFeedbackForm, setShowFeedbackForm] = useState(false);
```
tsck, line 62: Q: Does this have to be stateful and set separately? Can it just be determined by other state such as `feedbackFormState !== FormState.Submitted`?

chat/message-actions/src/MessageActions/MessageActions.tsx
```diff
@@ -55,7 +56,19 @@ export function MessageActions({
...
+
+  const isSubmitState =
+    feedbackFormState === FormState.Submitted ||
+    feedbackFormState === FormState.Submitting;
+
+  const resetFeedback = useCallback(() => {
+    setFeedback('');
```
tsck, line 69: This is initialized as `undefined`. Wonder if we should be consistent and either initialize this a `''` or set it back to `undefined`. Think I would prefer the former but curious what you think.

chat/message-actions/src/MessageActions/MessageActions.types.ts
```diff
@@ -7,6 +7,12 @@ import { DarkModeProps, HTMLElementProps } from '@leafygreen-ui/lib';
 export interface MessageActionsProps
   extends DarkModeProps,
     HTMLElementProps<'div'> {
+  /**
+   * Optional error message to display when feedback submission fails.
+   * @default 'Oops, please try again.'
```
tsck, line 12: Just confirming that this copy is what product/design wants? I recall once a desire to stop using copy that was too polite. Sounds strange, I know, but it was a thing. So instead of something like 'Oops, please try again.' it would be something like 'Something went wrong, try again.' So just want to verify.

.changeset/cruel-areas-care.md
```diff
@@ -0,0 +1,40 @@
+---
+'@lg-chat/message-feedback': major
+---
+
+[LG-5414](https://jira.mongodb.org/browse/LG-5414)
+
+#### Breaking Changes
+
+- Replaced `isSubmitted` boolean prop with `state` enum prop with default of `'unset'`
```
tsck, line 9: Q: Just curious what driving force behind this change was, since it is a major breaking change?

chat/message-actions/src/MessageActions/MessageActions.tsx
stephl3: Originally, it could be. However, I was told there is a separate endpoint for submitting a rating vs submitting a rating + feedback

In some cases, an end user may attempt to submit feedback, only for the submission to fail. If the user closes the form, we don't want to deselect the rating, but we will need to hide the form at that point despite the user not having submitted the form

chat/message-actions/src/MessageActions/MessageActions.tsx
stephl3: Good call, let me update to empty string

.changeset/cruel-areas-care.md
stephl3: hah yeah the driving force was primarily me wanting to avoid:
```
isSubmitted?: boolean;

isError?: boolean;

isLoading?: boolean;
```

I know we had earlier discussions about being mindful of major changes. I assessed it would be appropriate in this case because `InlineMessageFeedback` has only 1 external usage, and it needs to be replaced by `MessageActions` anyways

chat/message-actions/src/MessageActions/MessageActions.types.ts
stephl3: Yes, this is the copy design has in their specs. I'm not sure how much thought went into the default language. Is it okay to ship as-is for now? I can check with Sooa on this when she's back unless you feel strongly about altering it now

tsck: I think so. I think the wider docs team was pushing to be more involved with copy at one point. This might be a larger conversation we have with them and Hailey going forth to see if they want to be a final check on components as well. If this is something they still want, I think they should be. But good for now!

.changeset/cruel-areas-care.md
tsck: Ok cool. Just wanted to make sure it was considered. If you think it's warranted, I'm on board!

PR #3062, opened 2025-08-20
[LG-5453] feat(chat-disclaimer): update DisclaimerText with theming and deprecate DisclaimerModal

stephl3:
## ✍️ Proposed changes

This PR updates the `@lg-chat/chat-disclaimer` package with theming improvements for the `DisclaimerText` component and officially deprecates the `DisclaimerModal` component in favor of `@leafygreen-ui/marketing-modal`. The `DisclaimerText` component now supports dark mode theming, improved styling with proper spacing and alignment, and follows the latest LeafyGreen design patterns.

🎟️ _Jira ticket:_ [LG-5453](https://jira.mongodb.org/browse/LG-5453)

## ✅ Checklist

## 🧪 How to test changes

1. **Test DisclaimerText theming:**
   - Import and render `DisclaimerText` in both light and dark modes
   - Verify the title text uses the correct secondary text color for each theme
   - Confirm the component properly centers content and applies correct spacing

2. **Test deprecated DisclaimerModal:**
   - Import `DisclaimerModal` and verify the deprecation warning appears in TypeScript
   - Confirm the component still functions but shows deprecation notice

3. **Test new DarkModeProps support:**
   - Pass `darkMode={true}` and `darkMode={false}` to `DisclaimerText`
   - Verify the component respects the explicit dark mode setting

### DIFF
--- a/.changeset/light-years-cough.md
+++ b/.changeset/light-years-cough.md
@@ -1,0 +1,8 @@
+---
+'@lg-chat/chat-disclaimer': patch
+---
+
+[LG-5453](https://jira.mongodb.org/browse/LG-5453)
+
+- Update `DisclaimerText` component with theming and styling improvements
+- Mark `DisclaimerModal` component deprecated. Use `@leafygreen-ui/marketing-modal` instead


chat/chat-disclaimer/src/index.ts
```diff
@@ -1,2 +1,5 @@
+/**
+ * @deprecated - use `@leafygreen-ui/marketing-modal` instead
```
shaneeza, line 2: Is there also a component that we can add this to?

stephl3: can you clarify this? I think this should be sufficient since it's the only external-facing export but curious where else you might add

shaneeza: I meant, should we also add the following in the `DisclaimerModal` component, similar to what's done in `ChatAvatar` :
```
/**
 * @deprecated Use `MarketingModal` instead.
 */
```

stephl3: Ahh, I see. I guess it can be helpful for internal documentation. For this case, I will be pushing a follow-up major version to remove the export altogether since it doesn't see any usage anyways

shaneeza: When you say, remove the export, do you mean delete that package?

stephl3: Package will remain intact. `DisclaimerText` is used and will remain in the package, but `DisclaimerModal` will be removed

shaneeza: Should we keep it and archive it? I remember we spoke about archiving deprecated components in dev sync. However, this is such a tiny component, it probably doesn't make sense to archive it.

PR #3063, opened 2025-08-21
[LG-51290 feat(code-editor): add undo/redo functionality

tsck:
## ✍️ Proposed changes

This PR adds comprehensive undo/redo functionality to the CodeEditor component, providing users with full editor history management capabilities. The implementation includes both imperative API methods accessible via ref and functional UI integration through the existing Panel component.

Key enhancements include:
- **Imperative API**: New `undo()` and `redo()` methods on the CodeEditor ref handle that return boolean success indicators
- **Panel Integration**: Existing undo/redo buttons in the Panel secondary menu now perform actual operations instead of just triggering callbacks
- **Context Support**: Undo/redo functions available to child components via `CodeEditorContext` for custom UI implementations
- **Enhanced Testing**: New test utilities and functional tests that verify actual content changes rather than mock function calls
- **Comprehensive Documentation**: New "Imperative API" section with reorganized structure and practical examples

The implementation leverages CodeMirror's built-in `@codemirror/commands` module with proper error handling and maintains full backward compatibility while extending functionality.

## ✅ Checklist

- [x] I have tested my changes locally and they work as expected
- [x] I have added or updated tests to cover my changes
- [x] I have updated the documentation (README, JSDoc comments, etc.)
- [x] My changes follow the existing code style and conventions
- [x] I have considered the impact on bundle size and performance
- [x] I have verified that my changes don't introduce any accessibility regressions
- [x] I have added appropriate TypeScript types and interfaces
- [ ] I have tested my changes across different browsers (if applicable)
- [ ] I have considered backwards compatibility and migration path (if applicable)

### Testing Imperative API
1. **Functional Testing**:
   ```tsx
   const editorRef = useRef<CodeEditorHandle>(null);
   
   // Make some changes to the editor content
   // Call undo
   const undoSuccess = editorRef.current?.undo();
   // Verify content reverted
   
   // Call redo  
   const redoSuccess = editorRef.current?.redo();
   // Verify content restored
   ```

2. **Panel UI Testing**:
   - Render CodeEditor with `<Panel showSecondaryMenuButton />` 
   - Make changes to editor content
   - Click the ellipsis menu → "Undo" and verify content changes
   - Click "Redo" and verify content restores

3. **Context Integration Testing**:
   - Create custom child component that uses `useCodeEditorContext()`
   - Access `undo` and `redo` functions from context
   - Verify they work correctly when called from child components

4. **Edge Cases**:
   - Test undo when no history exists (should return `false`)
   - Test redo when no redo history exists (should return `false`)
   - Test with empty editor content
   - Test with large content changes

### Running Tests
```bash
# Run CodeEditor component tests
npm test -- packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx

# Run Panel component tests  
npm test -- packages/code-editor/src/Panel/Panel.spec.tsx
```
```

### CODE REVIEW

File: packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx [hunk 1]
```diff
@@ -350,4 +350,72 @@ describe('packages/code-editor', () => {
...
+      const { editor } = renderCodeEditor();
+
+      await editor.waitForEditorView();
+
+      const handle = editor.getHandle();
+      expect(typeof handle.undo).toBe('function');
+      expect(typeof handle.redo).toBe('function');
+      expect(typeof handle.getContents).toBe('function');
+      expect(typeof handle.formatCode).toBe('function');
+      expect(typeof handle.getEditorViewInstance).toBe('function');
```
tsck, line 366: Drive-by: these weren't previously tested so just added them along with this

packages/code-editor/src/Panel/Panel.spec.tsx
```diff
@@ -246,6 +257,106 @@ describe('Panel', () => {
       expect(onRedoClick).toHaveBeenCalledTimes(1);
     });
 
+    it('calls context undo function when undo menu item is clicked', async () => {
```
tsck, line 202: I just realized that a lot of the `code-editor` tests are using `it()` instead of `test()`. Will open a followup PR to clean this all up

packages/code-editor/src/CodeEditor/testing/testUtils.tsx
```diff
@@ -233,18 +234,45 @@ function isLineWrappingEnabled() {
...
+ * @returns The current text content of the editor
+ * @throws Error if editor view is not initialized
+ */
+function getContent(): string {
+  const view = ensureEditorView();
+  return view.state.doc.toString();
+}
+
+/**
+ * Gets the editor handle instance for testing imperative methods
```
shaneeza, line 248: ~~What is the handle instance?~~ nvm i get it!

packages/code-editor/src/Panel/Panel.spec.tsx
```diff
@@ -246,6 +257,106 @@ describe('Panel', () => {
...
+      renderPanel({ showSecondaryMenuButton: true });
+
+      const menuButton = screen.getByLabelText('Show more actions');
```
shaneeza: Is the menu button part of the panel, or does the consumer render it? If it's always part of the panel, maybe we should include test utils so it's easier to target it in tests?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -149,6 +149,36 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
...
+     */
+    const handleUndo = useCallback((): boolean => {
+      const commands = modules?.['@codemirror/commands'];
+
+      if (!editorViewRef.current || !commands) {
+        console.warn('Undo is not available - editor or commands not loaded');
+        return false;
+      }
+
+      return commands.undo(editorViewRef.current);
```
shaneeza, line 176: I was testing this behavior against VSCode's undo and redo. In VSCode, if you type a long string, pressing undo will undo each word until the string is gone, but in this case, pressing undo removes the entire string. Is that the correct behaviour?

tsck: By default, CodeMirror (and I think vscode too) group transactions into undo steps based on:
- Time intervals - transactions close together in time are grouped
- User activity - certain actions naturally break undo groups

I notice that if I just type `const` for example in vscode, then hit cmd+z, it does undo the whole word, same as in this editor. The way they group does seem somewhat different though. If I type `const x = 1;` vscode seems to undo piece by piece, and this editor undos the whole thing.

We do have the ability to override and customize the logic that handles grouping, but given that it's _close_ and there's no real requirement to have it match vscode completely, I'd lean towards not doing that for now. I think this is always something that could be tweaked later if we feel we need to.

packages/code-editor/src/Panel/Panel.spec.tsx
tsck: Yup it is, good point. I'll push a change for this shortly

tsck: I added an entire suite of testing utils and updated `Panel.spec` to use them instead.

tsck: I also updated the location of `testing/` since it should be at the `src` root I think

.changeset/sixty-turtles-tell.md
```diff
@@ -0,0 +1,15 @@
...
+'@leafygreen-ui/code-editor': minor
+---
+
+**Feature**: Add comprehensive undo/redo functionality to CodeEditor component
+
+- **Imperative API**: Added `undo()` and `redo()` methods to CodeEditor ref handle that return boolean success indicators
+- **Panel integration**: Existing undo/redo buttons in Panel secondary menu now perform actual operations instead of just triggering callbacks
+- **Context support**: Undo/redo functions available to child components via `CodeEditorContext` for custom UI implementations
+- **Enhanced TypeScript**: Complete `CodeEditorHandle` interface documentation with all imperative methods (`getContents`, `formatCode`, `undo`, `redo`, `getEditorViewInstance`)
+- **Test utilities**: New testing helpers including `editor.getContent()`, `editor.getHandle()`, `editor.interactions.undo()`, and `editor.interactions.redo()`
```
shaneeza: Should this also mention panel test utils? Also, are these test utilities only for internal testing?

tsck: The test utils are not just meant for internal testing. I can add a line about the utils here.

tsck: @shaneeza I've pushed this change

packages/code-editor/src/testing/panelTestUtils.tsx
shaneeza, line 1: I noticed that these external utils follow a different pattern compared to other components. Usually, our harnesses only return a single element or an array of elements that the consumer then uses however they want. So, for example, they can use the returned element to perform a click or read the content. We don't have utils that render the component. 

1. What's the reason for including rendering utils? 
2. Also, would these utils work as expected if there are multiple code editors being rendered at the same time?

tsck: In a code editor, rendering has to happen before an `editorView` is available. To perform specific actions on an editor, we need access to that view instance. So just exposing something like `getIndentUnit()` or `getContent()` isn't possible without having a view instance, because it isn't a simple selector that allows that.

However, I think you make a really good point. For external consumers, this will be of no value since they're already rendering editors on their pages and want to use our utils to help interact with them, as opposed to testing an instance in isolation. This makes me feel that all of the harnesses are not really going to be of value, honestly. 

I have a ticket coming up about test harnesses: https://jira.mongodb.org/browse/LG-5100. I think instead I should repurpose that into refactoring test utils to work better with consumer implementations. And I can take care of refactoring all of the test utils in a way that makes sense to the consumer then. 

How does that sounds on your end? I'd rather not do that in this ticket because it's a larger refactor and this PR is first in a line of 3 more PRs that are based off it, which would cause a cascade and honestly kind of a mess.

shaneeza: Sounds good to me!

shaneeza (approved):

PR #3072, opened 2025-08-23
[LG-5437] feat(rich-links): update RichLink styling and remove imageUrl prop

stephl3:
## ✍️ Proposed changes

This PR updates the `RichLink` component with more compact styling and removes the deprecated `imageUrl` prop as part of a design refresh. The changes include removing background image support, updating the layout to use flexbox for better alignment, and adding `forwardRef` support to `RichLinksArea` for improved component composition.

## ✅ Checklist

## 🧪 How to test changes

1. Navigate to the RichLinks Storybook stories
2. Verify that RichLink components display with the new compact styling (reduced padding and height)
3. Confirm that the `imageUrl` prop no longer exists in the component interface
4. Test that RichLinksArea now uses flexbox layout instead of CSS grid
5. Verify that badge positioning is now inline rather than absolutely positioned
6. Check that the `title` attribute is properly set on RichLink cards for accessibility

### DIFF
--- a/.changeset/curly-eyes-drum.md
+++ b/.changeset/curly-eyes-drum.md
@@ -1,0 +1,5 @@
+---
+'@lg-chat/message': minor
+---
+
+[LG-5437](https://jira.mongodb.org/browse/LG-5437): add expand/collapse functionality for `MessageLinks` and update to latest `@lg-chat/rich-links`
--- a/chat/message/package.json
+++ b/chat/message/package.json
@@ -19,6 +19,8 @@
     "@leafygreen-ui/banner": "workspace:^",
     "@leafygreen-ui/emotion": "workspace:^",
     "@leafygreen-ui/hooks": "workspace:^",
+    "@leafygreen-ui/icon": "workspace:^",
+    "@leafygreen-ui/icon-button": "workspace:^",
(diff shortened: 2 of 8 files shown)


.changeset/solid-rice-love.md
```diff
@@ -0,0 +1,15 @@
+---
+'@lg-chat/rich-links': major
+---
+
+[LG-5437](https://jira.mongodb.org/browse/LG-5437)
+
+#### Breaking Changes
+
+- Removed `imageUrl` prop from `RichLink` component
```
TheSonOfThomp, line 9: What was this used for? ~~As far as I can see, we never tested it in Chromatic~~ Oh I see it there now ("Video Variant")— doesn't seem like it was rendering correctly in the first place

TheSonOfThomp (approved):

TheSonOfThomp:
@stephl3 Any ideas why the Banner stories are changing? They shouldn't be using anything that changed in this PR, right?

.changeset/solid-rice-love.md
stephl3: yes, for showing a thumbnail of the video although it's dropped in the latest

PR #3073, opened 2025-08-23
[LG-5437] feat(message): add expand/collapse functionality for MessageLinks

stephl3:
## ✍️ Proposed changes

This PR adds expand/collapse functionality to the `MessageLinks` component, providing users with an interactive way to show/hide related resources. The implementation includes smooth CSS transitions, proper accessibility support with ARIA attributes, and follows established patterns from other LeafyGreen components like `ExpandableCard` and `ContextDrawer`. The component defaults to a collapsed state and uses dynamic height calculation for smooth animations.

_Note:  Consideration was put into abstracting the collapse/expand pattern, but it was deprioritized given the current timeline. See https://jira.mongodb.org/browse/LG-5468_

## ✅ Checklist

## 🧪 How to test changes

1. Navigate to MessageLinks in Storybook or test environment with links data
2. Verify the component renders in collapsed state by default (RichLinksArea hidden)
3. Click the chevron icon button to expand and confirm smooth height animation
4. Verify the chevron rotates 180 degrees when expanded/collapsed
5. Test keyboard navigation and confirm proper focus management
6. Verify accessibility features: `aria-expanded`, `aria-controls`, `role="region"`
7. Confirm the expand/collapse animation uses 300ms ease-in-out timing
8. Test that the component works with empty links array

### DIFF
--- a/chat/message/src/MessageLinks/MessageLinks.spec.tsx
+++ b/chat/message/src/MessageLinks/MessageLinks.spec.tsx
@@ -248,25 +248,5 @@ describe('MessageLinks', () => {
         expect(toggleButton).toHaveAttribute('aria-expanded', 'true');
       });
     });
-
-    test('updates height when links change', async () => {
-      const { rerender } = renderMessageLinks();
-
-      const newLinks = [
-        ...mockLinks,
-        { children: 'Link 4', href: 'https://example.com/4' },
-        { children: 'Link 5', href: 'https://example.com/5' },
-      ];
-
-      rerender(
-        <LeafyGreenChatProvider variant={Variant.Compact}>
-          <MessageLinks {...defaultProps} links={newLinks} />
-        </LeafyGreenChatProvider>,
-      );
-
-      // Should still work with new links
-      expect(screen.getByText('Link 4')).toBeInTheDocument();
-      expect(screen.getByText('Link 5')).toBeInTheDocument();
-    });
   });
 });
--- a/chat/message/src/MessageLinks/MessageLinks.styles.ts
+++ b/chat/message/src/MessageLinks/MessageLinks.styles.ts
@@ -5,7 +5,7 @@ import { spacing, transitionDuration } from '@leafygreen-ui/tokens';
  * The length of the box shadow for the RichLink component that
  * renders when a link is hovered.
  */
-const LINK_BOX_SHADOW_LENGTH = 3;
+const LINK_BOX_SHADOW_LENGTH = 4;
 const TRANSITION_DURATION = transitionDuration.slower;
 
 export const containerStyles = css`
@@ -35,7 +35,7 @@ export const getIconStyles = (isExpanded: boolean) =>
 
 const baseLinksWrapperStyles = css`
   display: grid;
-  padding-top: ${spacing[50]}px;
+  padding-top: ${spacing[100]}px;
   transition-property: height, grid-template-rows, opacity, visibility;
   transition-duration: ${TRANSITION_DURATION}ms;
   transition-timing-function: ease-in-out;
--- a/chat/message/src/MessageLinks/MessageLinks.tsx
+++ b/chat/message/src/MessageLinks/MessageLinks.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useState } from 'react';
 import { RichLinksArea } from '@lg-chat/rich-links';
 
 import { useIdAllocator } from '@leafygreen-ui/hooks';
@@ -29,9 +29,6 @@ export function MessageLinks({
 
   const [isExpanded, setIsExpanded] = useState(false);
 
-  const contentRef = useRef<HTMLDivElement>(null);
-  const linksAreaRef = useRef<HTMLDivElement>(null);
-
   const headingId = useIdAllocator({ prefix: 'message-links-heading' });
   const contentId = useIdAllocator({ prefix: 'message-links-content' });
 
@@ -59,7 +56,6 @@ export function MessageLinks({
           </IconButton>
         </div>
         <div
-          ref={contentRef}
           id={contentId}
           role="region"
           aria-labelledby={headingId}
@@ -69,7 +65,6 @@ export function MessageLinks({
             className={linksInnerWrapperStyles}
             links={links}
             onLinkClick={onLinkClick}
-            ref={linksAreaRef}
           />
         </div>
       </div>


chat/message/src/MessageLinks/MessageLinks.tsx
```diff
@@ -18,13 +24,83 @@ export function MessageLinks({
   onLinkClick,
   ...divProps
 }: MessageLinksProps) {
-  const { theme } = useDarkMode(darkModeProp);
+  const { darkMode } = useDarkMode(darkModeProp);
+
+  const [isExpanded, setIsExpanded] = useState(false);
```
shaneeza, line 30: There is a CSS-only way to animate a dynamic height that I used for vertical stepper: https://keithjgrant.com/posts/2023/04/transitioning-to-height-auto/#with-grid. Maybe this could be beneficial here? With this approach, you wouldn't need to measure the height of the links on load and when the content changes.

chat/message/src/Message/Message.stories.tsx
```diff
@@ -178,45 +178,49 @@ export const VerifiedAnswer: StoryObj<MessageStoryProps> = {
...
+    variant: 'Website',
+  },
+];
+const getWithRichLinksChildren = () => (
+  <>
+    <MessageFeedback />
+    <MessageLinks links={links} />
+  </>
+);
 export const WithRichLinks: StoryObj<MessageStoryProps> = {
```
shaneeza: nit: I know this was here before, but this story's name is a little misleading. The `How to Test Changes` section in the description says to "Navigate to MessageLinks in Storybook"  but there is no story with that name. Could this be changed to `withMessageLinks`?

chat/message/src/MessageLinks/MessageLinks.spec.tsx
```diff
@@ -0,0 +1,334 @@
...
+      expect(toggleButton).toHaveAttribute('aria-expanded', 'false');
+
+      // Click to expand
+      await userEvent.click(toggleButton);
+
+      // Should now be expanded with different styling
+      await waitFor(() => {
+        expect(toggleButton).toHaveAttribute('aria-expanded', 'true');
+        // The icon should have different classes applied for rotation
+        expect(toggleButton.className).toMatch(/leafygreen-ui/);
```
shaneeza: Why do we need this test? I also don't think this test is correct since the className always starts with `leafygreen-ui`

chat/message/src/MessageLinks/MessageLinks.spec.tsx
```diff
@@ -0,0 +1,334 @@
...
+      });
+    });
+  });
+
+  describe('height calculation', () => {
```
shaneeza: Do we need this test? It seems repetitive since you're testing `expand/collapse behavior` above.

chat/message/src/MessageLinks/MessageLinks.spec.tsx
```diff
@@ -0,0 +1,334 @@
...
+      toggleButton.focus();
+
+      await userEvent.keyboard(' ');
+
+      expect(toggleButton).toHaveAttribute('aria-expanded', 'true');
+    });
+  });
+
+  describe('edge cases', () => {
+    test('handles empty links array', () => {
```
shaneeza, line 232: Should this component not render if there are no links?

chat/message/src/MessageLinks/MessageLinks.spec.tsx
```diff
@@ -0,0 +1,334 @@
...
+      const toggleButton = screen.getByRole('button');
+      await userEvent.click(toggleButton);
+
+      // Should still expand without errors
+      await waitFor(() => {
+        expect(toggleButton).toHaveAttribute('aria-expanded', 'true');
+      });
+    });
+
+    test('updates height when links change', async () => {
```
shaneeza: I don't think this is testing the correct behavior. Links are collapsed by default but still in the DOM, so even if you add more links, they will always be in the DOM.

chat/message/src/MessageLinks/MessageLinks.spec.tsx
```diff
@@ -0,0 +1,334 @@
...
+        </LeafyGreenChatProvider>,
+      );
+
+      // Should still work with new links
+      expect(screen.getByText('Link 4')).toBeInTheDocument();
+      expect(screen.getByText('Link 5')).toBeInTheDocument();
+    });
+  });
+
+  describe('dark mode', () => {
```
shaneeza: Do we need this test? It's the same as the rendering tests above, but only the darkMode prop is different.

chat/message/src/MessageLinks/MessageLinks.styles.ts
```diff
@@ -1,22 +1,57 @@
...
+  transform: rotate(180deg);
+`;
+
+export const getIconStyles = (isExpanded: boolean) =>
+  cx(baseIconStyles, {
+    [expandedIconStyles]: isExpanded,
+  });
+
+const baseLinksWrapperStyles = css`
+  /* overflow: hidden; */
```
shaneeza: rm

chat/message/src/Message/Message.stories.tsx
shaneeza: Also, do you have any thoughts on adding content below the richlinks story so we can see how the transition affects the content below it?

stephl3: I'll leave it off for now since we don't expect consumers to render children below the links and don't want to encourage it by including it in our stories code

shaneeza (commented):
Just a few small things

chat/message/src/MessageLinks/MessageLinks.tsx
```diff
@@ -18,13 +25,55 @@ export function MessageLinks({
...
+
+  const contentRef = useRef<HTMLDivElement>(null);
+  const linksAreaRef = useRef<HTMLDivElement>(null);
```
shaneeza: Do you still need these refs?

chat/message/src/MessageLinks/MessageLinks.spec.tsx
shaneeza: Do we still need this test?

chat/message/src/MessageLinks/MessageLinks.styles.ts
```diff
@@ -1,22 +1,62 @@
...
-  Variant,
-} from '@leafygreen-ui/tokens';
+import { css, cx } from '@leafygreen-ui/emotion';
+import { spacing, transitionDuration } from '@leafygreen-ui/tokens';
+
+/**
+ * The length of the box shadow for the RichLink component that
+ * renders when a link is hovered.
+ */
+const LINK_BOX_SHADOW_LENGTH = 3;
```
shaneeza: I think this needs to be increased a little more to include the focus outline.
Image

shaneeza (approved):

PR #3076, opened 2025-08-26
chore(TS) Include `.spec` files in tsconfig

TheSonOfThomp:
Adding `.spec` files to the tsconfig ensures that any type tests get type checked at build-time.

Updates spec files to resolve TS issues where the solution is simple. 
Adds `FIXME` comments where types in spec files were already broken

**Next steps:** Type check spec & story files only on `test` run (not every `build`) 

This _will_ slow down builds somewhat in the short term


packages/icon-button/src/IconButton/IconButton.spec.tsx
```diff
@@ -100,7 +100,7 @@ describe('packages/icon-button', () => {
     });
 
     test('requires either aria-label or aria-labelledby', () => {
-      /// @ts-expect-error
+      // FIXME:  // @ts-expect-error - aria-label or aria-labelledby is required
```
TheSonOfThomp, line 103: https://jira.mongodb.org/browse/LG-5475

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.spec.tsx
```diff
@@ -104,7 +104,7 @@ describe('packages/info-sprinkle', () => {
   // eslint-disable-next-line jest/no-disabled-tests
   test.skip('types behave as expected', () => {
     <>
-      {/* @ts-expect-error - Missing children */}
+      {/* FIXME:  add expect error @ts-expect-error - Missing children */}
```
TheSonOfThomp, line 107: https://jira.mongodb.org/browse/LG-5476

packages/typography/src/Link/Link/Link.spec.tsx
```diff
@@ -6,11 +6,7 @@ import { InferredPolymorphicProps } from '@leafygreen-ui/polymorphic';
 import Link from './Link';
 import { BaseLinkProps } from './Link.types';
 
-type SpanLikeProps = InferredPolymorphicProps<'span', BaseLinkProps>;
-
-type AnchorLikeProps = InferredPolymorphicProps<'a', BaseLinkProps>;
-
-type LinkRenderProps = SpanLikeProps | AnchorLikeProps;
+type LinkRenderProps = InferredPolymorphicProps<'span', BaseLinkProps>;
```
shaneeza, line 9: Can we get away with this, because if `href` is defined, the component props are set to `a` element props, and `span` is ignored?

charts/core/src/ChartContext/ChartContext.spec.tsx
```diff
@@ -12,6 +12,8 @@ const addChartSeries = jest.fn();
 const ChartProviderMock = ({ children }: PropsWithChildren<{}>) => {
   return (
     <ChartContext.Provider
+      //  TODO: Fix this test context @tsck
```
tsck, line 15: 👍 Will fix this once it's merged - https://jira.mongodb.org/browse/LG-5478

packages/typography/src/Link/Link/Link.spec.tsx
TheSonOfThomp: That's by design. If we define `href` we want to use `a` props not `span` props

PR #3078, opened 2025-08-27
[LG-5473] feat(drawer): add hasPadding prop to decouple padding from scrollable behavior

stephl3:
## ✍️ Proposed changes

This PR adds a new `hasPadding` prop to the `Drawer` component to allow independent control of padding and scrollable behavior. Previously, padding could only be removed by setting `scrollable` to `false`, which tightly coupled these two concerns. Now developers can control padding and scroll behavior independently, providing more flexibility for full-width/height content scenarios.

The changes include:
- New `hasPadding` prop (defaults to `true`) that controls content padding independently
- Updated `scrollable` prop to only control scroll behavior 
- Comprehensive documentation updates with behavior matrix
- Enhanced Storybook stories with play tests for better visual regression testing
- Support for `hasPadding` in `DrawerToolbarLayout` components

🎟️ _Jira ticket:_ [LG-5473](https://jira.mongodb.org/browse/LG-5473)

## ✅ Checklist

## 🧪 How to test changes

1. **Test independent padding control:**
   - Navigate to the Drawer Storybook stories
   - Try combinations of `hasPadding={true/false}` and `scrollable={true/false}` 
   - Verify all four combinations render correctly: padded+scrollable, padded+not-scrollable, not-padded+scrollable, not-padded+not-scrollable

2. **Test visual regression with play tests:**
   - Run the "ScrollableAndPadded" and "ScrollableAndNotPadded" stories
   - Verify the play tests automatically scroll the drawer content before taking snapshots
   - Confirm padding differences are visually apparent between the two stories

3. **Test DrawerToolbarLayout integration:**
   - Test that `hasPadding` prop is properly exposed in toolbar data objects
   - Verify backward compatibility with existing `scrollable` usage

### DIFF
--- a/.changeset/wise-queens-take.md
+++ b/.changeset/wise-queens-take.md
@@ -1,0 +1,7 @@
+---
+'@leafygreen-ui/drawer': minor
+---
+
+[LG-5473](https://jira.mongodb.org/browse/LG-5473) Add `hasPadding` prop to `Drawer` to allow removing padding styles from container wrapping `children`. Also expose `hasPadding` prop in `toolbarData` objects.
+
+Note: previously, padding styles could be removed by setting `scrollable` to `false`. This has been de-coupled from `scrollable`, and `hasPadding` and `scrollable` work independently. [Learn more here about padding and scroll behavior](https://github.com/mongodb/leafygreen-ui/blob/main/packages/drawer/README.md#padding-and-scroll-behavior)


packages/drawer/src/Drawer/Drawer.spec.tsx
```diff
@@ -67,44 +67,6 @@ describe('packages/drawer', () => {
     });
   });
 
-  describe('scrollable prop', () => {
```
stephl3, line 70: nothing to see here folks... just removing some AI-generated tests that were definitely doing something 🤡

packages/drawer/src/Drawer.stories.tsx
```diff
@@ -404,43 +407,93 @@ export const Default: StoryObj<DrawerProps> = {
...
-  );
-  display: flex;
-  align-items: center;
-  justify-content: center;
-`;
-const FullWidthHeightContent = () => (
-  <div className={fullWidthHeightContentStyles}>
-    <Subtitle>Full Width/Height Content</Subtitle>
-  </div>
-);
```
stephl3, line 423: Decided it was better to consistently render `LongContent` as children in these stories and use play tests to show the scroll behavior

packages/drawer/src/Drawer/Drawer.styles.ts
```diff
@@ -198,6 +198,26 @@ export const getDrawerStyles = ({
     drawerClassName,
   );
 
+export const getResizerStyles = ({
```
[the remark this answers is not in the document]
stephl3: I prefer ordering these in order of usage  🤷

tsck (commented):
No concerns technically. Curious why we went with a prop that controls styling here instead of exposing a class name constant that consumers can use to style things themselves? I just know I've seen us doing the latter a bit in the recent past.

tsck:
> > No concerns technically. Curious why we went with a prop that controls styling here instead of exposing a class name constant that consumers can use to style things themselves? I just know I've seen us doing the latter a bit in the recent past.
> 
> 2 reasons for this:
> 
> 1. we already have `scrollable` and preserving that parallel pattern can be less mental overhead for consumers
> 2. we're opinionated about styling the `Drawer` and don't want to give too much flexibility here. Consumers _should_ either have the standard padding or use the full-width/height

I feel like when we want to lock down style, a more common approach I've seen lately is to use variants. Such as the default vs compact `Menu`, where primarily one has less padding. Does this component not lend itself to that well?

I'm just concerned that we're starting to make it so that some components export child classes, while others have style props, without very clear justification as to why. This seems like more mental overhead for consumers in general when using LG. This is something we should codify more in our future architecture/philosophy doc.

I'm not a hard blocker, just want to make sure we're looking at all options since I feel its a bit of a different pattern than I've seen lately.

tsck (approved):

PR #3087, opened 2025-08-28
[LG-5437] refactor(message): migrate MessageActions into Message package

stephl3:
## ✍️ Proposed changes

Move `MessageActions` code from `@lg-chat/message-actions` into `@lg-chat/message` package to consolidate related functionality. This is part of a larger effort to streamline our chat component architecture by co-locating tightly coupled components.

🎟️ _Jira ticket:_ [LG-5437](https://jira.mongodb.org/browse/LG-5437)

## 🔄 Migration Details

- Copied over `MessageActions` code and documentation from `@lg-chat/message-actions`
- Preserved test IDs and constants for backwards compatibility
- No functional changes to the component behavior

## ✅ Checklist

- [x] I have added stories/tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `pnpm changeset` and documented my changes

## 🧪 How to test changes

1. Verify code migration
   - Confirm all `MessageActions` files are copied over correctly
   - Check that constants like `FEEDBACK_TEXTAREA_TEST_ID` are preserved
   - Ensure documentation is migrated and up to date

2. Run tests
   - `npm test chat/message/src/MessageActions`
   - Verify all existing tests pass without modification

3. Visual verification
   - Run Storybook
   - Check MessageActions stories render correctly in new location
   - Verify all interactions work as before

## 📝 Notes

This is part 1 of 3 PRs:
1. Copy over `MessageActions` code: this PR
2. Add helper utils for subcomponent composition: https://github.com/mongodb/leafygreen-ui/pull/3088
3. Handle `Message` subcomponent composition: https://github.com/mongodb/leafygreen-ui/pull/3089

PR #3088, opened 2025-08-28
[LG-5437] feat(lib): add findChild and filterChildren utils for compound components

stephl3:
## ✍️ Proposed changes

Add new `findChild` and `filterChildren` utils to `@leafygreen-ui/lib` to support compound component patterns in the design system. These utils enable finding and filtering specific subcomponents by their static properties, with additional support for Emotion's `styled()` HOCs and React fragments.

## ✅ Checklist

## 🧪 How to test changes

1. Run the comprehensive test suite and verify test cases pass
   ```bash
   pnpm run test packages/lib/src/findChild
   ```
2. Run the comprehensive test suite and verify test cases pass
   ```bash
   pnpm run test packages/lib/src/filterChildren
   ```

## 📝 Notes

This is part 2 of 3 PRs:
1. Copy over `MessageActions` code: https://github.com/mongodb/leafygreen-ui/pull/3087
2. Add helper utils for subcomponent composition: this PR
3. Handle `Message` subcomponent composition: https://github.com/mongodb/leafygreen-ui/pull/3089


packages/lib/src/filterChildren/filterChildren.tsx
```diff
@@ -0,0 +1,94 @@
...
+ * <Foo isFoo={true} />
+ *
+ * // ✅ Will filter out: Children inside fragments with matching properties
+ * <><Bar isBar={true} /></>
+ *
+ * // ❌ Will NOT filter: Nested fragments
+ * <><><Baz isBaz={true} /></></>
+ *
+ * // ❌ Will NOT filter: Deeply nested
+ * <div><Foo isFoo={true} /></div>
```
TheSonOfThomp, line 28: a future improvement could be recursive with a `maxDepth` param, but this lgtm for now

packages/lib/src/filterChildren/filterChildren.tsx
```diff
@@ -0,0 +1,94 @@
...
+      filteredChildren.push(child);
+      continue;
+    }
+
+    // Handle fragments by filtering their children
+    if (child.type === Fragment) {
+      const fragmentChildren = Children.toArray(child.props.children);
+      const filteredFragmentChildren: Array<ReactNode> = [];
+
+      for (const fragmentChild of fragmentChildren) {
```
TheSonOfThomp: I have a hunch that this might have some optimizations to be more performant (especially since we're calling `Children.toArray`)? Could be wrong though

```suggestion
      React.Children(fragmentChildren, (fragmentChild) => {
```

packages/lib/src/filterChildren/filterChildren.tsx
```diff
@@ -0,0 +1,94 @@
...
+ */
+function hasAnyStaticProperty(
+  component: any,
+  staticProperties: Array<string>,
+): boolean {
+  return staticProperties.some(
+    property =>
+      component[property] || // Direct property on component
+      component.target?.[property] || // Standard styled() components store original in target
+      component.__emotion_base?.[property], // Exotic forward ref styled components store original in __emotion_base
```
TheSonOfThomp: I assume this is for `styled`? Can you add docs for when this might be used?

packages/lib/src/filterChildren/filterChildren.tsx
```diff
@@ -0,0 +1,94 @@
+import { Children, Fragment, isValidElement, ReactNode } from 'react';
+
+/**
+ * Check if a component has any of the specified static properties.
+ * Handles regular components, standard styled() components, and exotic forward ref styled components.
+ */
+function hasAnyStaticProperty(
+  component: any,
```
TheSonOfThomp: this should be more strictly typed. Should it be the full `React.Component`, or just the `component.type` parameter

packages/lib/src/filterChildren/filterChildren.tsx
```diff
@@ -0,0 +1,94 @@
...
+        if (!isValidElement(fragmentChild)) {
+          filteredFragmentChildren.push(fragmentChild);
+          continue;
+        }
+
+        if (
+          !hasAnyStaticProperty(fragmentChild.type, staticPropertiesToExclude)
+        ) {
+          filteredFragmentChildren.push(fragmentChild);
+        }
```
TheSonOfThomp: Might be misunderstanding something here, but this only checks for the existence of a property, not whether it equals some expected value.

If this is correct, could you clarify in a comment what the check is actually doing

packages/lib/src/findChild/findChild.spec.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+const Baz = React.forwardRef<HTMLDivElement, { text: string }>(
+  ({ text }, ref) => <div ref={ref}>{text}</div>,
+);
+Baz.displayName = 'Baz';
+
+// Add static properties to test components with type assertion
+(Foo as any).isFoo = true;
+(Bar as any).isBar = true;
+(Baz as any).isBaz = true;
```
TheSonOfThomp, line 25: To clarify, for any compound components we'll have to explicitly define e.g. `isMessageBanner`?

packages/lib/src/findChild/findChild.tsx
```diff
@@ -0,0 +1,81 @@
...
+ * Check if a component has the specified static property.
+ * Handles regular components, standard styled() components, and exotic forward ref styled components.
+ */
+function hasStaticProperty(component: any, staticProperty: string): boolean {
+  return (
+    component[staticProperty] || // Direct property on component
+    component.target?.[staticProperty] || // Standard styled() components store original in target
+    component.__emotion_base?.[staticProperty] // Exotic forward ref styled components store original in __emotion_base
+  );
+}
```
TheSonOfThomp: You're declaring this function in 2 places. Can we extract this to its own utility?

TheSonOfThomp (changes requested):
Just the comment on moving (and more strictly-typing) the `hasAnyStaticProperty` function.

Otherwise just a couple nits, and LGTM

packages/lib/src/filterChildren/filterChildren.tsx
stephl3: Agreed! I would be curious how deep or if we should even support that level of depth vs keep it restricted to the current depth check, but we can discuss further as this pattern develops further

packages/lib/src/filterChildren/filterChildren.tsx
stephl3: I actually updated this util to use `Children.map()` which has better readability and optimized for react :)

packages/lib/src/findChild/findChild.spec.tsx
stephl3: yessir. here's what that looks like for `Message`: https://github.com/mongodb/leafygreen-ui/pull/3089/files#diff-ac866ab27d482537aa7ad55635c182afc0d285819a4be8249e3072c0d7c99bfcR101-R114

packages/lib/src/filterChildren/filterChildren.tsx
stephl3: now that we use `Children.map()` and the logic is flipped, it should read more clearly in the latest

packages/lib/src/filterChildren/filterChildren.tsx
stephl3: I did still need to include `Record<string, any>`. Let me know if there is something more specific you think we can use

TheSonOfThomp (approved):

PR #3089, opened 2025-08-28
[LG-5437] feat(message): implement compound components pattern

stephl3:
## ✍️ Proposed changes

This PR implements a compound components pattern for the Message component, allowing developers to compose messages using subcomponents like `Message.Actions`, `Message.Links`, and `Message.VerifiedBanner`. The changes improve the component's flexibility and modularity while maintaining backward compatibility.

Key improvements:
- Enhanced `Message.VerifiedBanner` to accept `className` and HTML props for `@emotion/styled` support
- Updated TypeScript interfaces for better type safety with compound components
- Comprehensive README documentation with usage examples and behavior explanations
- Established compound components patterns

## ✅ Checklist

## 🧪 How to test changes

1. **Compound Components Pattern**: Test that `Message.Actions`, `Message.Links`, and `Message.VerifiedBanner` can be used as children of the `Message` component
2. **Styled Components Support**: Verify that `styled(Message.VerifiedBanner)` now properly applies custom styles with `className` forwarding
3. **Variant Behavior**: Confirm that subcomponents only render in compact variant
4. **Optional Props**: Test that components conditionally render based on provided props (e.g., retry button only shows when `onClickRetry` is provided)
5. **Documentation**: Review updated README examples to ensure all compound component patterns work as documented

## 📝 Notes

This is part 1 of 3 PRs:
1. Copy over `MessageActions` code: https://github.com/mongodb/leafygreen-ui/pull/3087
2. Add helper utils for subcomponent composition: https://github.com/mongodb/leafygreen-ui/pull/3088
3. Handle `Message` subcomponent composition: this PR

### DIFF
--- a/chat/message/README.md
+++ b/chat/message/README.md
@@ -63,13 +63,6 @@ return (
 
 The `Message` component uses a compound component pattern, allowing you to compose different parts of a message using subcomponents like `Message.Actions`, `Message.Links`, and `Message.VerifiedBanner`.
 
-This pattern provides several benefits:
-
-- **Modularity**: Each subcomponent has its own props and encapsulated behavior.
-- **Flexibility**: Mix and match only the components you need for your use case.
-- **Type Safety**: Each compound component exposes its own TypeScript interface.
-- **Conditional Rendering**: Subcomponents only render when their required props are provided.
-
 **Note 1:** All compound components only render in the `compact` variant.  
 **Note 2:** The layout and order of compound components are enforced by the `Message` component itself. Even if you change the order of subcomponents in your JSX, they will be rendered in the correct, intended order within the message bubble. This ensures consistent UI and accessibility regardless of how you compose your message children.
 
--- a/chat/message/src/Message/CompactMessage.tsx
+++ b/chat/message/src/Message/CompactMessage.tsx
@@ -7,6 +7,7 @@ import { filterChildren, findChild } from '@leafygreen-ui/lib';
 import { BaseFontSize } from '@leafygreen-ui/tokens';
 import { Body } from '@leafygreen-ui/typography';
 
+import { MessageSubcomponentProperty } from '../constants';
 import {
   MessageContainer,
   Variant as MessageContainerVariant,
@@ -36,16 +37,18 @@ export const CompactMessage = forwardRef<HTMLDivElement, MessageProps>(
     const { assistantName } = useLeafyGreenChatContext();
 
     // Find subcomponents
-    const actions = findChild(children, 'isLGMessageActions');
-    const verifiedBanner = findChild(children, 'isLGMessageVerifiedBanner');
-    const links = findChild(children, 'isLGMessageLinks');
+    const actions = findChild(children, MessageSubcomponentProperty.Actions);
+    const verifiedBanner = findChild(
+      children,
+      MessageSubcomponentProperty.VerifiedBanner,
+    );
+    const links = findChild(children, MessageSubcomponentProperty.Links);
 
     // Filter out subcomponents from children
-    const remainingChildren = filterChildren(children, [
-      'isLGMessageActions',
-      'isLGMessageVerifiedBanner',
-      'isLGMessageLinks',
-    ]);
+    const remainingChildren = filterChildren(
+      children,
+      Object.values(MessageSubcomponentProperty),
+    );
 
     return (
       <div
--- a/chat/message/src/Message/Message.tsx
+++ b/chat/message/src/Message/Message.tsx
@@ -9,6 +9,7 @@ import LeafyGreenProvider, {
 } from '@leafygreen-ui/leafygreen-provider';
 import { consoleOnce } from '@leafygreen-ui/lib';
 
+import { MessageSubcomponentProperty } from '../constants';
 import { MessageActions } from '../MessageActions';
 import { MessageVerifiedBanner } from '../MessageBanner';
 import { MessageContext } from '../MessageContext';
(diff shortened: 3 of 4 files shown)


chat/message/src/Message/Message.tsx
```diff
@@ -9,13 +9,21 @@ import LeafyGreenProvider, {
...
+import {
+  ActionsType,
+  LinksType,
+  type MessageProps,
+  type VerifiedBannerType,
+} from './Message.types';
 import { SpaciousMessage } from './SpaciousMessage';
 
-export const Message = forwardRef<HTMLDivElement, MessageProps>(
+export const BaseMessage = forwardRef<HTMLDivElement, MessageProps>(
```
tsck: Does this need to be exported now? It doesn't get used outside of this file atm.

chat/message/src/Message/CompactMessage.tsx
```diff
@@ -34,6 +35,18 @@ export const CompactMessage = forwardRef<HTMLDivElement, MessageProps>(
...
+    const actions = findChild(children, 'isLGMessageActions');
+    const verifiedBanner = findChild(children, 'isLGMessageVerifiedBanner');
+    const links = findChild(children, 'isLGMessageLinks');
+
+    // Filter out subcomponents from children
+    const remainingChildren = filterChildren(children, [
+      'isLGMessageActions',
+      'isLGMessageVerifiedBanner',
+      'isLGMessageLinks',
+    ]);
```
tsck: I know these are ultimately property names, but it feels like they'd be better off being constants rather than re-used strings.

chat/message/src/Message/Message.spec.tsx
```diff
@@ -96,4 +96,45 @@ describe('Message', () => {
 
     expect(consoleOnce.warn).not.toHaveBeenCalled();
   });
+
+  test('renders subcomponents and filters them from remaining children', () => {
```
tsck, line 100: Q: What filtering is being done and tested here?

chat/message/README.md
```diff
@@ -57,7 +59,21 @@ return (
...
+### Compound Components
+
+The `Message` component uses a compound component pattern, allowing you to compose different parts of a message using subcomponents like `Message.Actions`, `Message.Links`, and `Message.VerifiedBanner`.
+
+This pattern provides several benefits:
+
+- **Modularity**: Each subcomponent has its own props and encapsulated behavior.
+- **Flexibility**: Mix and match only the components you need for your use case.
+- **Type Safety**: Each compound component exposes its own TypeScript interface.
+- **Conditional Rendering**: Subcomponents only render when their required props are provided.
```
tsck: Q: Are these actually the benefits of just subcomponents? I feel like these are true of all components

chat/message/README.md
tsck, line 1: Q: Is there any value to making note regarding the deprecated way of doing things?

stephl3: For the compact variant, these are all backwards-compatible additions. I think because there's only 1 current case of the old props-based approach that is also for the spacious variant, it's better to leave off documentation about the alternative way of approaching the subcomponents and will prevent confusion about which approach to use

chat/message/src/Message/Message.spec.tsx
stephl3: This spec is a happy path test that `Message` uses `filterChildren` to remove found subcomponents from `children` and renders them only once to ensure they are not duplicated among the remaining children. More testing for `findChild` and `filterChildren` in PR https://github.com/mongodb/leafygreen-ui/pull/3088

chat/message/src/Message/CompactMessage.tsx
stephl3: good feedback! please let me know what you think of the updated :)

chat/message/README.md
stephl3: haha you're right. this is a bit over-the-top AI doc gen. removed!

chat/message/README.md
tsck: Makes sense to me!

chat/message/src/Message/Message.spec.tsx
tsck: Are there actually duplicates that are being filtered out in this test though?

chat/message/src/Message/CompactMessage.tsx
tsck: LGTM!

chat/message/src/Message/Message.spec.tsx
stephl3: @tsck gotcha, so it's not testing that duplicate instances are only rendered once. It's testing that we don't render a single instance twice

However, the `findChild` util does have a spec for ensuring that only the first found instance is returned
```
test('should find first match even with multiple matches', () => {
    const children = (
      <>
        <Foo text="first-match" />
        <Foo text="second-match" />
        <Bar text="different" />
      </>
    );

    const found = findChild(children, 'isFoo');
    expect(found).toBeDefined();
    expect((found as React.ReactElement).props.text).toBe('first-match');
  });
  ```

tsck: I won't block on this, it just seems like either an odd test, or minimally a misdescribed test to me.

stephl3: I wanted to include _something_ that tests the integration of using `findChild` and `filterChildren` in the `Message `component. Open to feedback if you'd describe it differently or drop it altogether

PR #3096, opened 2025-08-29
chore(Types): Enables TS downleveling

TheSonOfThomp:
Enables TS downleveling for all consumer-facing packages from TS5 to TS4.9


charts/chart-card/package.json
```diff
@@ -37,5 +37,20 @@
   },
   "devDependencies": {
     "@lg-tools/build": "workspace:^"
+  },
+  "typesVersions": {
+    "*": {
+      ".": [
+        "././dist/types/index.d.ts"
```
shaneeza: `typesVersions` is a fallback if `exports` is not supported, is that correct?

chat/message-feedback/package.json
```diff
@@ -5,7 +5,6 @@
   "main": "./dist/umd/index.js",
   "module": "./dist/esm/index.js",
   "types": "./dist/types/index.d.ts",
```
TheSonOfThomp, line 7: This is a bandaid. We'll need to come up with a more rigorous solution if we want to export and share stories between components.

charts/core/package.json
```diff
@@ -44,5 +44,15 @@
...
     "url": "https://jira.mongodb.org/projects/PD/summary"
-  }
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/types/index.d.ts",
+      "types@<=5.0": "./dist/types/ts4.9/index.d.ts",
+      "import": "./dist/esm/index.js",
+      "require": "./dist/umd/index.js",
+      "types@<5.0": "./dist/types/ts4.9/index.d.ts"
```
shaneeza: Do we need this? It seems redundant given the presence of `types@<=5.0`

TheSonOfThomp: nope. Thought I had removed all these

PR #3098, opened 2025-08-29
[LG-5445] feat(code-editor): add keyboard shortcuts help modal

tsck:
## ✍️ Proposed changes

This PR adds a keyboard shortcuts help modal to the Code Editor component to improve user experience and discoverability. Users can now access a comprehensive list of available keyboard shortcuts directly from the editor's menu, eliminating the need to search through documentation or guess at available shortcuts.

The implementation includes a new `ShortcutMenu` component that displays shortcuts in a clean, organized table format, and integrates seamlessly with the existing Code Editor interface through a new "View shortcuts" menu item.

Note: The `ESC` shortcut differs slightly from the designs. This was discussed and decided upon with Sandy.

🎟️ _Jira ticket:_ [LG-5445](https://jira.mongodb.org/browse/LG-5445)

## ✅ Checklist

## 🧪 How to test changes

1. **Access the shortcuts modal:**
   - Open any Code Editor instance
   - Click the three-dot menu (⋯) in the editor toolbar
   - Click "View shortcuts" menu item
   - Verify the modal opens with keyboard shortcuts displayed

2. **Verify shortcuts content:**
   - Confirm the modal shows all expected shortcuts:
     - `TAB` - to indent
     - `ESC` - to unfocus editor
     - `⌘/CTRL + F` - to find
     - `⌘/CTRL + Z` - to undo
     - `⌘/CTRL + SHIFT + Z` - to redo
   - Check that keyboard shortcuts are properly styled with badges
   - Verify the modal can be closed by clicking outside or the close button

3. **Test functionality:**
   - Test that the actual keyboard shortcuts work as documented in the modal
   - Verify the feature works across different browsers and operating systems
   - Ensure the modal doesn't interfere with existing editor functionality


packages/code-editor/src/Panel/Panel.tsx
```diff
@@ -125,100 +129,120 @@ export function Panel({
...
+                Redo
+              </MenuItem>
+              <MenuItem
+                glyph={<DownloadIcon />}
+                onClick={handleDownloadClick}
+                aria-label="Download code"
+              >
+                Download
+              </MenuItem>
+              <MenuItem
```
shaneeza, line 212: If I click on `View shortcuts`, the modal appears, and the menu remains open. Once I close the modal, the menu also closes. Is that behavior correct?

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
```diff
@@ -669,4 +670,135 @@ describe('packages/code-editor', () => {
       }
     });
   });
+
+  describe('Keybindings', () => {
+    test('Pressing ESC key unfocuses the editor', async () => {
```
shaneeza, line 675: What is the use case for unfocusing the editor with the ESC key? I'm also curious: If the editor is no longer in focus, what happens if you press `tab`? What will be in focus?

tsck: This is to provide a way for a keyboard user to navigate away from the editor via `tab`. Without this, `tab` would only insert tabs in the editor and the keyboard user would be stuck in the editor. "unfocusing" the editor actually results to moving focus from inside the editor, to outside the editor, so that tab navigation works as expected again. 

I have more detail on this in the technical spec under the [Avoiding Trapping Keyboard-Only Users](https://docs.google.com/document/d/18SX9deSWeEdcjp4VMtUAEDcQ9L9LlCFpoKFNt8bEEag/edit?tab=t.clfropts70z8#heading=h.e4e1zbal9wv9) section

packages/code-editor/src/Panel/Panel.styles.ts
```diff
@@ -51,3 +51,7 @@ export const getPanelButtonsStyles = () => {
     grid-area: buttons;
   `;
 };
+
+export const ModalStyles = css`
+  height: 354px;
```
shaneeza: Why this value, and can we move it to a constant?

packages/code-editor/src/Panel/Panel.tsx
```diff
@@ -125,100 +129,120 @@ export function Panel({
...
+              trigger={
+                <IconButton
+                  onClick={handleFormatClick}
+                  aria-label="Format code"
+                >
+                  <FormatIcon />
+                </IconButton>
+              }
+              triggerEvent="hover"
+              darkMode={darkMode}
```
shaneeza: nit: I think the default for `Tooltip` is `hover`, so you don't need to specify it. Also, is `CodeEditor` wrapped in a `LeafyGreenProvider`? If so, you probably don't need to pass `darkMode` to `Tooltip`.

packages/code-editor/src/Panel/Panel.tsx
tsck: That's a good question, and I'm not sure. I'll look into this and see what other similar things do, and possibly reach out to design to see if there's a preference.

packages/code-editor/src/Panel/Panel.styles.ts
tsck: That's just the height of the modal in the designs 🤷‍♂️ https://www.figma.com/design/QsOruUtjXisdkdZHJh9CsG/Code-Editor?node-id=515-8714&t=cEGDKYZaLkmNzYfg-4

But yeah, will definitely move it to a constant!

packages/code-editor/src/ShortcutMenu/ShortcutMenu.styles.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const TableStyles = css`
+  width: ${TABLE_WIDTH};
+  border-spacing: 0;
+
+  tr {
+    margin-bottom: 12px;
+    display: block;
+
+    &:not(:last-child) {
```
shaneeza: Are you trying to add padding-bottom to all the rows except for the last row?

shaneeza: You can probably do something like this, which only adds padding to any cell immediately following a row. This way, you can avoid changing the display property of the `tr` and `td`.



```
tr {
    & + & > td {
       padding-top: 12px;
    }
}

File: packages/code-editor/src/ShortcutMenu/ShortcutMenu.tsx [hunk 6]
```diff
@@ -0,0 +1,71 @@
...
+import { Disclaimer, H3 } from '@leafygreen-ui/typography';
+
+import {
+  HeadingStyles,
+  PlusSignStyles,
+  TableContainerStyles,
+  TableStyles,
+} from './ShortcutMenu.styles';
+
+export function ShortcutMenu({ className }: { className?: string }) {
```
shaneeza: I'm curious about the naming of this. It made me think that it was a menu like our menu component. Why `ShortcutMenu` vs something like `ShortcutTable`?

packages/code-editor/src/ShortcutMenu/ShortcutMenu.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <div className={className}>
+      <H3 className={HeadingStyles}>Code Editor Shortcuts</H3>
+      <div className={TableContainerStyles}>
+        <table className={TableStyles}>
+          <tr>
```
shaneeza: We should wrap these `tr`s in a `tbody`:

```
<tbody>
     <tr></tr>
</tbody>
```

packages/code-editor/src/ShortcutMenu/ShortcutMenu.styles.ts
tsck: That is what I was trying to do, but honestly omitting the last isn't necessary, I don't think. I'm going to remove that.

The alternative didn't seem to work. Curious why the latter would be a better solution?

packages/code-editor/src/ShortcutMenu/ShortcutMenu.tsx
tsck: I like `ShortcutTable`! Will update

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
shaneeza: My concern is where the focus is moved outside the DOM. Is it to the next element in the DOM, or does the focus move all the way to the top of the DOM? Is there any way we can test this to see where the focus ends up?

packages/code-editor/src/Panel/Panel.tsx
tsck: I've updated this so that the menu closes first on click before the modal opens. I also duplicated this logic in the other menu handlers.

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
tsck: Yup, it ends up on the container, and tab brings you to the next element. Best way to test this is by trying it in storybook. If you hit ESC, and then SHIFT+TAB, I see the focus move to the previous panel button. If ESC moved the focus to the first item in the DOM, I wouldn't expect this. I also to be safe, added a random button to the story locally, and can confirm that ESC, then SHIFT, moved correctly to the button.

shaneeza: Thanks for the detailed example. I was able to confirm this!

packages/code-editor/src/ShortcutMenu/ShortcutMenu.styles.ts
shaneeza: Sorry about that, i'm not sure why `& + &` is not working, but this does:

```
tr + tr td {
    padding-top: 12px;
  }
```
Also, I believe it's better to add style to the `<td>` element. From Gemini: 

> In a table, you add padding to the `<td>` (table data) elements and not the `<tr>` (table row) because of the CSS box model and the specific behavior of table elements. The <tr> element is designed to be a container for cells, and it does not behave like a standard block-level element that can have padding applied to it.

With the solution above, you also avoid removing the default display property of a row and cell.

packages/code-editor/src/Panel/Panel.tsx
shaneeza: nit: You could also remove `darkMode` from `Menu`

packages/code-editor/src/Panel/Panel.tsx
```diff
@@ -105,120 +110,145 @@ export function Panel({
...
+                    {label}
+                  </MenuItem>
+                ),
+              )}
+            </Menu>
+          )}
+        </div>
       </div>
-    </div>
+      <Modal
```
shaneeza, line 244: This is not related to your work, but it doesn't appear that the focus is trapped within the modal when it's open. Should I create a ticket for this, or do you want to?

tsck: You can create it since you found it!

packages/code-editor/src/ShortcutTable/ShortcutTable.styles.ts
```diff
@@ -0,0 +1,45 @@
...
+  width: ${TABLE_WIDTH};
+  border-spacing: 0;
+
+  tr + tr td {
+    padding-top: ${ROW_PADDING}px;
+  }
+
+  td {
+    display: inline-block;
+    vertical-align: top;
```
shaneeza, line 24: nit: You probably don't need display and vertical-align anymore

PR #3103, opened 2025-09-03
[LG-5100] feat(code-editor): add test harnesses

tsck:
## ✍️ Proposed changes
- Moves and makes internal the following:
  - `src/testing/testUtils.tsx` -> `src/CodeEditor/CodeEditor.testUtils.tsx`
  - `src/testing/panelTestUtils.tsx` -> `src/Panel/Panel.testUtils.tsx`
  - `src/testing/extensionTestUtils.ts` -> `src/CodeEditor/hooks/hooks.testUtils.ts`
- Add test harnesses `src/testing/getTestUtils`
- Updates docs to remove old utils and document the new ones
- Updates `CodeEditor` and `Panel` as needed to work with LGID's

🎟 _Jira ticket:_ [LG-5100](https://jira.mongodb.org/browse/LG-5100)

## ✅ Checklist

## 🧪 How to test changes

1. **Basic test utility usage**:
   ```typescript
   const utils = getTestUtils('your-editor-lgid');
   expect(utils.getEditor()).toBeInTheDocument();
   expect(utils.getLanguage()).toBe('javascript');
   ```

2. **Async content testing**:
   ```typescript
   await utils.waitForInitialization();
   const content = await utils.getContent();
   expect(content).toBe('console.log("hello");');
   ```

3. **Panel functionality testing**:
   ```typescript
   const panelUtils = utils.getPanelUtils();
   expect(panelUtils?.getPanelTitle()).toBe('My Panel');
   expect(panelUtils?.getCopyButton()).toBeInTheDocument();
   ```

4. **Run the comprehensive test suite**: `pnpm test packages/code-editor/src/testing/getTestUtils.spec.tsx` to see all usage examples and verify functionality

### DIFF
--- a/packages/code-editor/src/testing/index.ts
+++ b/packages/code-editor/src/testing/index.ts
@@ -1,6 +1,6 @@
 export { codeSnippets } from './codeSnippets';
 export { getTestUtils } from './getTestUtils';
 export type {
+  GetTestUtilsReturnType,
   PanelTestUtilsReturnType,
-  TestUtilsReturnType,
 } from './getTestUtils.types';


packages/code-editor/src/CodeEditor/CodeEditor.testUtils.tsx
```diff
@@ -96,43 +96,6 @@ function getBySelector(
...
 
-/**
- * Returns all elements matching a specific CodeEditor selector
- * @param selector - The CSS selector to look for in the editor
- * @param options - Optional filtering options
- * @param options.text - Optional text content filter
- * @returns All DOM elements matching the selector and optional text filter
- * @throws Error if no elements are found
- */
-function getAllBySelector(
```
tsck, line 107: This wasn't being used, and since this is just an internal tool now, it didn't make sense to keep it

packages/code-editor/src/CodeEditor/CodeEditor.testUtils.tsx
```diff
@@ -174,40 +137,6 @@ function queryBySelector(
...
 }
 
-/**
- * Returns all elements matching a specific CodeEditor selector or null if none are found
- * @param selector - The CSS selector to look for in the editor
- * @param options - Optional filtering options
- * @param options.text - Optional text content filter
- * @returns All DOM elements matching the selector and optional text filter, or null if none found
- */
-function queryAllBySelector(
```
tsck, line 184: Same on not being used

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
```diff
@@ -117,161 +117,162 @@ export interface CodeEditorTooltip {
   severity?: CodeEditorTooltipSeverity;
 }
 
-export type CodeEditorProps = DarkModeProps & {
```
tsck, line 123: Only thing that changes here is adding `LgIdProps`

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    // Try to get content from rendered CodeMirror content element
+    const contentElement = element.querySelector(CodeEditorSelectors.Content);
+
+    if (contentElement?.textContent) {
+      return contentElement.textContent;
+    }
+
+    // In test environments or when CodeMirror isn't available,
```
shaneeza: What's the difference between the actual CodeMirror instance and a test environment?

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,479 @@
...
+import { LanguageName } from '../CodeEditor/hooks/extensions/useLanguageExtension';
+import { Panel } from '../Panel';
+
+import { getTestUtils } from './getTestUtils';
+
+// For testing purposes, create a simplified test environment that focuses on
+// demonstrating the getTestUtils API rather than fully mocking CodeMirror
+//
+// Note: In real applications, consumers would use getTestUtils with actual
+// CodeEditor components that don't need this level of mocking
```
shaneeza: why do we need this level of mocking but consumers dont?

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,479 @@
...
+
+import '@testing-library/jest-dom';
+
+import { CodeEditor } from '../CodeEditor';
+import { LanguageName } from '../CodeEditor/hooks/extensions/useLanguageExtension';
+import { Panel } from '../Panel';
+
+import { getTestUtils } from './getTestUtils';
+
+// For testing purposes, create a simplified test environment that focuses on
```
shaneeza: Similar to what Adam mentioned in dev sync yesterday, should we be testing the actual CodeEditor component and not a mock?

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,479 @@
...
+            />
+          }
+          data-lgid="lg-panel-example"
+        />,
+      );
+
+      const utils = getTestUtils('lg-panel-example');
+
+      // Wait for component to be fully rendered
+      await new Promise(resolve => setTimeout(resolve, 100));
```
shaneeza: Why do we have to wait?

shaneeza: Would this work? You could probably avoid the timeout this way.

```js
await waitFor(() => {
  expect(utils.queryPanel()).toBeInTheDocument();
});
```

packages/code-editor/src/Panel/Panel.tsx
```diff
@@ -74,6 +75,7 @@ import { PanelProps } from './Panel.types';
 export function Panel({
   baseFontSize: baseFontSizeProp,
   customSecondaryButtons,
+  'data-lgid': dataLgId,
```
shaneeza: If a consumer passes a custom data-lgid to `CodeEditor`, it won't propagate down to the panel. So, for example, 

```js
<CodeEditor data-lgid="lg-custom" panel={<Panel/>}>...</CodeEditor>
```

In this case, any component that uses the lgids instantiated in `CodeEditor` will have a data-lgid prefixed with `lg-custom.....`. But Panel and all of its children data-lgids will still be prefixed with`lg-code_editor...`. 

What I've done in instances like this is store the lgids instantiated in the parent component, which is `CodeEditor`, in context and read those values from context in child component. [The `Panel` component in `Code` does this ](https://github.com/mongodb/leafygreen-ui/blob/main/packages/code/src/Panel/Panel.tsx#L32).

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,479 @@
...
+      // Example: Testing panel functionality
+      render(
+        <CodeEditor
+          defaultValue="SELECT * FROM users;"
+          panel={
+            <Panel
+              title="SQL Query"
+              showCopyButton={true}
+              showFormatButton={true}
+              data-lgid="lg-panel-example"
```
shaneeza: Oh, I see. You're passing the same data-lgid to `CodeEditor` and `Panel`, so `Panel` has the correct prefix. You can avoid this if you use the context approach I mentioned above.

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,479 @@
...
+          enableLineNumbers={true}
+          enableCodeFolding={true}
+          data-lgid="lg-features-example"
+        />,
+      );
+
+      const utils = getTestUtils('lg-features-example');
+
+      // Test feature detection methods
+      expect(typeof utils.getHasLineNumbers()).toBe('boolean');
```
shaneeza: Whats the benefit of testing this over the actual boolean value?

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,479 @@
...
+        );
+      };
+
+      render(<MyForm />);
+
+      // Consumer can test the code editor within their form
+      const utils = getTestUtils('lg-form-code-editor');
+
+      // Wait for component to be fully rendered
+      await new Promise(resolve => setTimeout(resolve, 100));
```
shaneeza: Why 100?

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,479 @@
...
+      expect(panelUtils?.getPanelTitle()).toBe('JavaScript Editor');
+      expect(panelUtils?.getPanelCopyButton()).toBeInTheDocument();
+      expect(panelUtils?.getFormatButton()).toBeInTheDocument();
+
+      // Features are enabled as expected - wait for initialization first
+      try {
+        await utils.waitForInitialization(2000);
+      } catch {
+        // If initialization fails, still proceed with basic checks
+      }
```
shaneeza: It's a little unclear what's happening here. Can you clarify?

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,479 @@
...
+
+      const utils = getTestUtils('lg-async-example');
+
+      // The recommended pattern for content testing:
+      try {
+        await utils.waitForInitialization(1000); // 1 second timeout
+        const content = await utils.getContent();
+
+        // In a real browser environment, this would return the actual content
+        // In test environments, it falls back to data attributes
```
shaneeza: Which data attributes?

shaneeza: Also, why is there a fallback in test environments?

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,351 @@
...
+
+  /**
+   * Gets the content that was passed to the CodeEditor component.
+   * This reads the actual content from CodeMirror when available, falling back to props.
+   */
+  const getContent = async (): Promise<string | null> => {
+    // First try to get content from actual CodeMirror instance if available
+    const containerElement = element as any;
+
+    if (containerElement._cm?.state?.sliceDoc) {
```
shaneeza: What's the benefit of using the CodeMirror instance over getting the content from the rendered CodeMirror content element?

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,351 @@
...
+      getSecondaryMenuButton,
+      getSecondaryMenu,
+      isSecondaryMenuOpen,
+    };
+  };
+
+  /**
+   * Gets all tooltip elements currently visible in the editor
+   */
+  const getTooltips = (): Array<HTMLElement> => {
```
shaneeza: Can more than one tooltip be opened at a time?

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,351 @@
...
+   */
+  const getIsReadOnly = (): boolean => {
+    const editorElement = element.querySelector(CodeEditorSelectors.Editor);
+    return editorElement?.classList.contains('cm-readonly') || false;
+  };
+
+  /**
+   * Checks if line numbers are enabled
+   */
+  const getHasLineNumbers = (): boolean => {
```
shaneeza: What is the benefit of adding feature detection utilities?

packages/code-editor/README.md
```diff
@@ -317,545 +317,223 @@ editorRef.current.downloadContent(); // No filename provided
...
+// Now test CodeMirror-specific functionality
 ```
 
-#### `editor.getAllBySelector(selector, options?)`
+#### Basic Editor Utilities
 
-Returns all elements matching a specific CodeEditor selector. Throws an error if no elements are found.
+- `getEditor()` - Returns the CodeEditor root element
+- `getContent()` - Gets the current text content from the editor (returns `Promise<string | null>`)
+- `getLanguage()` - Gets the programming language currently set on the editor (from CSS classes)
```
shaneeza: what does from CSS classes mean?

packages/code-editor/README.md
```diff
@@ -317,545 +317,223 @@ editorRef.current.downloadContent(); // No filename provided
...
-import { PanelSelectors } from '@leafygreen-ui/code-editor';
-
-// Use instead of hardcoded strings
-panel.getMenuItem(PanelSelectors.UndoMenuItem);
-panel.getMenuItem(PanelSelectors.RedoMenuItem);
-panel.getMenuItem(PanelSelectors.DownloadMenuItem);
-panel.getMenuItem(PanelSelectors.ViewShortcutsMenuItem);
-```
+  // Test configuration detection
+  expect(typeof utils.getIsLoading()).toBe('boolean');
```
shaneeza: What's the use case for consumers wanting to test this?

packages/code-editor/README.md
```diff
@@ -317,545 +317,223 @@ editorRef.current.downloadContent(); // No filename provided
...
-    contextConfig: {
-      formatCode: mockFormatCode,
-    },
-  });
-
-  await panel.interactions.clickFormatButton();
-
-  expect(onFormatClick).toHaveBeenCalled();
-  expect(mockFormatCode).toHaveBeenCalled();
+  // Test actual content and language (works in all environments!)
```
shaneeza: What do you mean by works in all environments?

packages/code-editor/README.md
```diff
@@ -317,545 +317,223 @@ editorRef.current.downloadContent(); // No filename provided
...
 
-// Override default functions with jest mocks in your test files
-const mockUndo = jest.fn(() => true);
-const mockFormatCode = jest.fn();
+  // Wait for loading to complete (useful for testing loading state changes)
+  try {
+    await utils.waitForLoadingToComplete(1000); // 1 second timeout
+    expect(utils.getIsLoading()).toBe(false);
+  } catch {
+    // Timeout is acceptable in test environments
```
shaneeza: Can you clarify what this comment means?

packages/code-editor/README.md
```diff
@@ -317,545 +317,223 @@ editorRef.current.downloadContent(); // No filename provided
...
-const { panel } = renderPanel({
-  contextConfig: {
-    undo: mockUndo,
-    formatCode: mockFormatCode,
-  },
+  // For content testing, wait for initialization first
+  try {
+    await utils.waitForInitialization(1000);
+    const content = await utils.getContent();
+    expect(typeof content).toBe('string');
```
shaneeza: Should this example test the actual content?

packages/code-editor/src/testing/getTestUtils.spec.tsx
shaneeza: OH, I think I get it now. data attributes, as in using querySelector.

packages/code-editor/src/Panel/Panel.tsx
tsck: My thinking was just that they are technically separate components so having separate ID's made sense. But since this is already a pattern that's been established, I'll definitely update accordingly!

packages/code-editor/src/testing/getTestUtils.ts
tsck: I had them in the internal tools so I just kept them as a way for a consumer to assert that a feature they wanted enabled was in fact enabled. But I am comfortable getting rid of them. It will clean up the API and make things simpler. I suppose we can always add them back later if it's something consumers ask of us.

tsck: Actually might ping you on the side. I'd be curious what you think actually is beneficial here for consumers. Having not done much with our test utils, I think it would be beneficial just to expose what's inline with what we usually do

packages/code-editor/src/testing/getTestUtils.ts
tsck: That's a really good point. They technically can stack, but really they shouldn't, and just checking they still render in the same `.cm-tooltip` component. Multiple "lint ranges" can show (the little squiggly line under the text indicating that a tooltip is there) but this doesn't get them. Will sync with you about this when we sync about what this should actually contain to see if this functionality is even valuable

packages/code-editor/src/testing/getTestUtils.ts
tsck: I think this was just AI doing what AI does and over-complicating things, and me not doing my due diligence.

In theory, AI claims, the internal editor state is the most accurate value. However, this is bad testing philosophy imo. What a user cares about is what's rendered and not what the internal state is. That's testing implementation and not reality. I'm going to rewrite this function and simplify it significantly. 

Thanks for catching!

packages/code-editor/src/testing/getTestUtils.ts
tsck: When I was building this out with AI, it claimed that certain envs can't render out the editor fully. So data values were added to represent certain props. I'm questioning this now and am going to look into removing these completely.

packages/code-editor/README.md
tsck: JSDom and actual DOM envs. AI just added this to be sales-y when I made it remove some mocks and test actual content lol

packages/code-editor/src/testing/getTestUtils.ts
tsck: I've removed this completely. I'm not sure if there was really value here.

packages/code-editor/src/testing/getTestUtils.ts
tsck: I did go ahead and remove these

packages/code-editor/README.md
tsck: I was thinking if/when they're using it as a controlled prop. They may be controlling the loading state themselves and want to test if their logic is working as expected.

packages/code-editor/README.md
tsck: This is gone now, so will just resolve

packages/code-editor/README.md
tsck: Not sure where this was but it's gone now 🙂

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
```diff
@@ -320,24 +371,34 @@ describe('packages/code-editor', () => {
...
 
     await editor.waitForEditorView();
 
+    // Wait a bit for copy button to be rendered
+    await new Promise(resolve => setTimeout(resolve, 100));
+
+    // The copy button selector looks for the specific lgid structure
     expect(
-      container.querySelector(CodeEditorSelectors.CopyButton),
+      container.querySelector('[data-lgid="lg-test-copy-hover-copy_button"]'),
```
shaneeza: nit: could you use the test util here? I see it as another way of testing that our utils work.

packages/code-editor/src/CodeEditor/CodeEditor.spec.tsx
```diff
@@ -320,24 +371,34 @@ describe('packages/code-editor', () => {
   test('renders copy button when copyButtonAppearance is "hover"', async () => {
     const { container, editor } = renderCodeEditor({
       copyButtonAppearance: CopyButtonAppearance.Hover,
+      'data-lgid': 'lg-test-copy-hover',
     });
 
     await editor.waitForEditorView();
 
+    // Wait a bit for copy button to be rendered
+    await new Promise(resolve => setTimeout(resolve, 100));
```
shaneeza: nit: instead of the timeout, would this work?
```js
await waitFor(() => {
  expect(...).toBeInTheDocument();
});
```

packages/code-editor/src/Panel/Panel.tsx
```diff
@@ -93,7 +93,7 @@ export function Panel({
   const { theme } = useDarkMode(darkMode);
   const baseFontSize = useBaseFontSize();
 
-  const { getContents, formatCode, undo, redo, downloadContent } =
+  const { getContents, formatCode, undo, redo, downloadContent, lgIds } =
```
shaneeza, line 96: 👍🏽

packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -0,0 +1,246 @@
...
+// MutationObserver not supported in JSDOM
+global.MutationObserver = jest.fn().mockImplementation(() => ({
+  observe: jest.fn(),
+  unobserve: jest.fn(),
+  disconnect: jest.fn(),
+  takeRecords: jest.fn().mockReturnValue([]),
+}));
+
+// Mock Modal component because HTMLDialogElement not supported in JSDOM
+jest.mock('@leafygreen-ui/modal', () => {
```
shaneeza: nit: I'm wondering, if instead of mocking modal, if you could get away with what we do in `Drawer` :

https://github.com/mongodb/leafygreen-ui/blob/1f334cbb324515ba3d210cb554e57602aeaab951/packages/drawer/src/Drawer/Drawer.spec.tsx#L30-L42

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+  const getContent = async (): Promise<string | null> => {
+    await waitForLoadingToComplete();
+    const contentElement = getContentElement();
+    return contentElement?.textContent || null;
+  };
+
+  /**
+   * Types the text into the editor
+   */
+  const typeContent = async (text: string): Promise<void> => {
```
shaneeza: I'm not sure if we should offer this as a util since we usually leave user events to the consumer. As long as they have access to the content element, that should be good enough.

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+ * // Get the actual content from CodeMirror
+ * const content = await utils.getContent();
+ *
+ * // Check loading state
+ * expect(utils.isLoading()).toBe(false);
+ * ```
+ */
+
+// Helper functions for DOM queries
+const getByLgId = (lgId: string): HTMLElement => {
```
shaneeza: I'm sorry I missed this earlier. Is there a reason you are not using `getByLgId` from the test-harnesses package? 

`import { findByLgId, getByLgId, queryByLgId } from '@lg-tools/test-harnesses';`

Our test harnesses use `@testing-library/dom` under the hood so we don't have to manually write them ourselves.

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+    waitForLoadingToComplete,
+    getContent,
+    getContentElement,
+    typeContent,
+    isLoading,
+    isReadOnly,
+    getCopyButton,
+    getPanelUtils,
+  };
+};
```
shaneeza, line 117: Overall, I think these are much simpler and helpful!

packages/code-editor/src/testing/getTestUtils.types.ts
```diff
@@ -0,0 +1,55 @@
+export interface PanelTestUtilsReturnType {
+  getPanelElement: () => HTMLElement | null;
```
shaneeza: Can you add tsdocs for these?

packages/code-editor/README.md
```diff
@@ -317,546 +317,97 @@ editorRef.current.downloadContent(); // No filename provided
...
-  await panel.interactions.clickUndoMenuItem();
-});
-```
 
-#### Panel Test Utilities Object
+  const utils = getTestUtils('lg-my-editor');
 
-The `panel` object provides access to Panel elements and interactions:
+  // Test editor presence and basic structure
+  expect(utils.getEditorElement()).toBeInTheDocument();
```
shaneeza: Can we also add examples for the basic editor utilities?

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+      getPanelTitle,
+      getFormatButton,
+      getPanelCopyButton,
+      getSecondaryMenuButton,
+      getSecondaryMenu,
+    };
+  };
+
+  return {
+    getEditorElement: () => element,
```
shaneeza: We've also started returning `find`, `query`, and `get`, so consumers have a few more options. I think adding them would be beneficial.
https://github.com/mongodb/leafygreen-ui/blob/1f334cbb324515ba3d210cb554e57602aeaab951/packages/context-drawer/src/testing/getTestUtils.tsx#L14-L16

packages/code-editor/src/testing/getTestUtils.spec.tsx
tsck: Will update but I don't really know if I see much of a difference here since this is meant to test the utils and both end up mocking the modal functionality anyway.

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -1,65 +1,83 @@
...
- * await utils.waitForLoadingToComplete();
- *
- * // Get the actual content from CodeMirror
- * const content = await utils.getContent();
- *
- * // Check loading state
- * expect(utils.isLoading()).toBe(false);
- * ```
+ * Returns a set of utility functions to query and get parts of a progress bar component for testing.
+ * @param lgId - The base LeafyGreen ID prefix for the progress bar. Defaults to `DEFAULT_LGID_ROOT`.
```
shaneeza:
```suggestion
 * @param lgId - The base LeafyGreen ID prefix for the code editor. Defaults to `DEFAULT_LGID_ROOT`.
```

tsck: Lol

packages/code-editor/src/testing/getTestUtils.ts
```diff
@@ -0,0 +1,117 @@
...
+): GetTestUtilsReturnType<T> => {
+  const lgIds = getLgIds(lgId);
+
+  /** Editor element utils */
+  const getEditor = () => getByLgId!<T>(lgIds.root);
+  const findEditor = () => findByLgId!<T>(lgIds.root);
+  const queryEditor = () => queryByLgId!<T>(lgIds.root);
+
+  /** Content element utils */
+  const getContentContainer = () => getByLgId!<T>(lgIds.content);
```
shaneeza, line 22: Curious, why the switch from `getContent` to `getContentContainer`?

tsck: Good question. `getContent` to me reads "get the content value" and not "get the content element". Before I had `getContentElement` and `getContent` where `getContent` returned the value and not the element. In the spirit of your comment before regarding not providing a util like `typeContent` when they have access to `content`. I removed both `typeContent` and the original `getContent`, leaving just `getContentElement`. I didn't like that this was the only util that had "element" in it, but I still thought `getContent` was confusing since it doesn't return the content of the editor, but the container of that content. So I updated it to `getContentContainer`.

PR #3105, opened 2025-09-03
chore(Icon): Update icon build script & package exports

TheSonOfThomp:
## ✍️ Proposed changes

'@leafygreen-ui/icon': minor
---

### New Features 
- Adds "exports" field  to package json
- Adds direct icon import pattern (e.g. `import CloudIcon from "@leafygreen-ui/icon/Cloud"`)
- Ensures backwards compatibility using "dist/*" imports

### Tooling
- Isolates build script from component TS project
- Updates individual icon build script to use the Rollup API directly
- Removes `postbuild` script. It's now redundant with the `"./dist/*"` exports field

Why: 
- Having icon build script code & component code combined in the same ts project caused ts build conflicts
- Using "package.exports` required updating the build script
- Enabling TS downlevelling requires "package.exports"


packages/icon/scripts/prebuild/index.ts
```diff
@@ -6,7 +6,8 @@ import { Command } from 'commander';
 import fs from 'fs';
 import path from 'path';
 
-import svgrrc from '../../.svgrrc';
+// @ts-ignore
+import svgrrc from '../.svgrrc';
```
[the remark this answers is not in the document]
TheSonOfThomp: no

packages/icon/scripts/build/build.ts
```diff
@@ -1,14 +1,20 @@
 /* eslint-disable no-console */
 import { exec } from 'child_process';
 import { Command } from 'commander';
+import path from 'path';
+
+import { DELIMITER } from '../constants';
```
TheSonOfThomp: wrong

tools/link/src/utils/install.spec.ts
```diff
@@ -9,7 +9,7 @@ describe('tools/link/utils/install', () => {
   let spawnSpy: jest.SpyInstance<ChildProcess>;
 
   beforeAll(() => {
-    fsx.mkdirSync('./tmp/');
+    fsx.ensureDirSync('./tmp/');
```
TheSonOfThomp, line 12: drive-by

tools/meta/src/getPackageJson/getPackageJson.spec.ts
```diff
@@ -4,7 +4,7 @@ import { getPackageJson } from './getPackageJson';
 
 describe('tools/meta/getPackageJson', () => {
   beforeAll(() => {
-    fsx.mkdirSync('./tmp/');
+    fsx.ensureDirSync('./tmp/');
```
TheSonOfThomp, line 7: drive-by

packages/icon/scripts/build/compare-checksum.ts
```diff
@@ -1,34 +1,48 @@
...
+import fsx from 'fs-extra';
 import path from 'path';
 
 const ICON_DIR = path.resolve(process.cwd(), 'src/generated');
 
-const iconNames = fs
-  .readdirSync(ICON_DIR)
-  .filter(f => /\.tsx?$/.test(f))
-  .map(f => path.basename(f, path.extname(f)));
+export function getAllIcons(): Array<string> {
```
shaneeza, line 12: Can you add TSDocs for these new functions?

packages/icon/src/Icon.spec.tsx
```diff
@@ -277,43 +275,6 @@ describe('Generated glyphs', () => {
     });
   });
 
-  describe('are up-to-date and not been modified', () => {
```
shaneeza, line 280: Why is this being removed?

packages/icon/tsconfig.json
```diff
@@ -1,18 +1,14 @@
 {
   "extends": "@lg-tools/build/config/package.tsconfig.json",
   "compilerOptions": {
-    "rootDir": "./",
```
shaneeza, line 4: What was this doing that's no longer needed?

packages/icon/package.json
```diff
@@ -4,12 +4,22 @@
...
-  "types": "./dist/types/src/index.d.ts",
-  "typesVersions": {},
+  "types": "./dist/types/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/esm/index.js",
+      "require": "./dist/umd/index.js",
+      "types": "./dist/types/index.d.ts"
+    },
+    "./dist/*": {
```
shaneeza, line 14: As a refresher from yesterday's dev sync, this is what ensures backwards compatibility when using "dist/*" imports?

shaneeza (approved):

packages/icon/package.json
TheSonOfThomp: correct, yeah

packages/icon/tsconfig.json
TheSonOfThomp: previously we needed to set the rootdir to the package root (not `src`) so that the scripts would be included in the ts build. 
I moved around the files so that's no longer necessary

packages/icon/src/Icon.spec.tsx
TheSonOfThomp: This wasn't checking whether the component itself works, but is checking that the build scripts worked. I'm not convinced it's necessary, and this type of check would be better as either a `validate` step, or unit tests in the build scripts

PR #3106, opened 2025-09-04
[LG-5096] feat(code-editor): add context menu functionality with cut/copy/paste operations

tsck:
## ✍️ Proposed changes

Adds context menu functionality to the `CodeEditor `component, providing standard text editing operations (Cut, Copy, Paste) via right-click context menu.

The implementation includes two new modular components:
- **ContextMenu**: Base context menu wrapper with automatic text selection detection
- **CodeEditorContextMenu**: Pre-configured context menu specifically for `CodeEditor`

🎟 _Jira ticket:_ [LG-5096](https://jira.mongodb.org/browse/LG-5096)

## ✅ Checklist

## 🧪 How to test changes

1. **Basic Context Menu**: Right-click on any text in the CodeEditor to see the context menu with Cut/Copy/Paste options
2. **Text Selection**: Select text in the editor, right-click, and verify the Cut/Copy operations work with the selected text
3. **Paste Functionality**: Copy text to clipboard, position cursor in editor, right-click and select Paste to verify content insertion
4. **Custom Menu Items**: Test `CodeEditorContextMenu` with custom menu items to ensure they appear below the separator


packages/code-editor/src/CodeEditor.stories.tsx
```diff
@@ -371,6 +322,114 @@ Typescript.args = {
   defaultValue: codeSnippets.typescript,
 };
 
+export const TooltipOnHover: StoryObj<{}> = {
```
TheSonOfThomp, line 325: this is still unstyled, right?

packages/code-editor/src/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,170 @@
...
+ *   <div>
+ *     Custom menu will show
+ *   </div>
+ *   <div data-no-context-menu="true">
+ *     Custom menu won't show
+ *   </div>
+ * </ContextMenu>
+ * ```
+ */
+export const ContextMenu = ({
```
TheSonOfThomp, line 40: The idea here is to eventually move to something like `@leafygreen-ui/context-menu`?

packages/code-editor/src/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { getLgIds } from '../utils';
+
+import { containerStyles, getMenuContainerStyles } from './ContextMenu.styles';
+import type { ContextMenuProps } from './ContextMenu.types';
+
+/**
+ * Context menu that adds custom right-click functionality to child elements.
+ *
+ * Elements with `data-no-context-menu="true"` will not trigger the custom
+ * context menu, allowing the default browser context menu to appear instead.
```
TheSonOfThomp, line 19: This approach feels a bit hacky to me, but I can't think of a better all-inclusive api. Curious if you have research backing this up?

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
```diff
@@ -0,0 +1,39 @@
...
+  label: string;
+
+  /** Whether the item is disabled */
+  disabled?: boolean;
+
+  /** Action handler - receives the selected text */
+  action?: (selectedText: string) => void;
+
+  /** Whether this is a separator */
+  isSeparator?: boolean;
```
TheSonOfThomp, line 20: I wonder if we should align with how we do this in CloudNav?

```ts
export const SideNavItemType = {
  Item: 'item',
  Divider: 'divider',
} as const;

export interface SideNavOptionInterface{
  type?: SideNavItemType;
  ...
}
```

TheSonOfThomp: ^ Doesn't have to be exactly that, and we can change the nav api if we need to, but IMO we should do the same thing in similar ways across the system.

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export interface ContextMenuProps extends LgIdProps {
+  /** The content to provide context menu functionality to */
+  children: ReactNode;
+
+  /** Menu items to show on right-click */
+  menuItems?: Array<ContextMenuItem>;
+
+  /** Whether to prevent the default context menu. @default true */
+  preventDefaultContextMenu?: boolean;
```
TheSonOfThomp: How does the menu work if this is `false`?

packages/code-editor/src/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,170 @@
...
+  );
+
+  /**
+   * Handle hiding menu on click outside of the menu.
+   *
+   * This is typically handled inside of the Menu component. However, on close
+   * the Menu component tries to focus on the next focusable element. This is
+   * not desirable for a context menu. Instead we just hide the entire thing
+   * before the refocus occurs by setting the open state to false and only
+   * rendering the Menu when the open state is true.
```
TheSonOfThomp, line 95: I wonder if we should add an escape-hatch prop to Menu to control this behavior? Could be abused though and break accessibility

packages/code-editor/src/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,170 @@
...
+              return (
+                <MenuItem
+                  key={index}
+                  disabled={item.disabled}
+                  onClick={() => {
+                    if (item.action && !item.disabled) {
+                      item.action(selectedText);
+                    }
+                    setIsOpen(false);
+                  }}
```
TheSonOfThomp: Should this be extracted to `handleItemClick(item)`?

packages/code-editor/src/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,170 @@
...
+       */
+      const target = e.target as Element;
+      if (target.closest(`[data-lgid="${lgIds.contextMenu}"]`)) return;
+
+      setIsOpen(false);
+    },
+    [lgIds.contextMenu],
+  );
+
+  useEffect(() => {
```
TheSonOfThomp, line 129: can we add TSDoc for what this is doing?

packages/code-editor/src/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,170 @@
...
+    if (isOpen) {
+      document.addEventListener('contextmenu', handleGlobalContextMenu);
+      /**
+       * Must capture click to prevent default Menu handling of clicks.
+       * See {@link handleClick} comment for more details.
+       */
+      document.addEventListener('click', handleClick, { capture: true });
+      return () => {
+        document.removeEventListener('contextmenu', handleGlobalContextMenu);
+        document.addEventListener('click', handleClick, { capture: true });
```
TheSonOfThomp: Should this be `removeEventListener`?
```suggestion
        document.removeEventListener('click', handleClick, { capture: true });
```

packages/code-editor/src/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,170 @@
...
+ *     },
+ *   ]}
+ * >
+ *   <div>
+ *     Custom menu will show
+ *   </div>
+ *   <div data-no-context-menu="true">
+ *     Custom menu won't show
+ *   </div>
+ * </ContextMenu>
```
TheSonOfThomp, line 37: Not 100% sold on this API, but it should be fine for now and is in line with some other components.

I wonder if this could be a case for Compound Components?
```tsx
<ContextMenu.Container>
  <ContextMenu.Menu>
    <ContextMenu.Item label="Copy" />
  <ContextMenu.Menu>
  <div>
    Custom menu will show
   </div>
   <div data-no-context-menu="true">
     Custom menu won't show
   </div>
</ContextMenu.Container>
```

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
TheSonOfThomp: Doesn't look like you're actually using `preventDefaultContextMenu` or `disabled` in the `ContextMenu` component

packages/code-editor/src/ContextMenu/ContextMenu.styles.ts
```diff
@@ -0,0 +1,15 @@
...
+export const containerStyles = css`
+  display: contents;
+`;
+
+export const getMenuContainerStyles = (position: {
+  x: number;
+  y: number;
+}) => css`
+  position: absolute;
+  z-index: 50;
```
TheSonOfThomp: 😬 Can we avoid this?

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <ContextMenu menuItems={[{ label: MENU_LABEL, action: actionMock }]}>
+      <div data-testid={CHILD_TEST_ID}>Test</div>
+      {props.children}
+    </ContextMenu>
+  );
+};
+
+describe('ContextMenu', () => {
+  test('renders when inner element is right clicked', () => {
```
TheSonOfThomp, line 29: Is this menu accessible to keyboard-only users? (not sure what the WCAG guidelines are for context menus)

TheSonOfThomp, line 29: Should add a test to assert it does _not_ render on left click (I envision see this accidental regression)

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+    userEvent.click(screen.getByTestId(CHILD_TEST_ID), { button: 2 });
+    expect(screen.queryByText(MENU_LABEL)).toBeInTheDocument();
+    userEvent.keyboard('{escape}');
+
+    await waitFor(() => {
+      expect(screen.queryByText(MENU_LABEL)).not.toBeInTheDocument();
+    });
+  });
+
+  test('hides when clicking outside of the menu', async () => {
```
TheSonOfThomp, line 65: Should we test that a second right click opens the menu in a new location?

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+    render(
+      <TestComponent>
+        <div data-no-context-menu="true" data-testid={NO_CONTEXT_MENU_ID} />
+      </TestComponent>,
+    );
+    userEvent.click(screen.getByTestId(NO_CONTEXT_MENU_ID), { button: 2 });
+    expect(screen.queryByText(MENU_LABEL)).not.toBeInTheDocument();
+  });
+
+  test('hides when Escape key is pressed', async () => {
```
TheSonOfThomp, line 54: What do Arrow keys and Enter do?

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.types.ts
```diff
@@ -0,0 +1,11 @@
+import { LgIdProps } from '@leafygreen-ui/lib';
+
+import { ContextMenuItem } from '../ContextMenu';
+
+export interface CodeEditorContextMenuProps extends LgIdProps {
```
TheSonOfThomp: Should this extend `ContextMenuProps`?

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+describe('CodeEditorContextMenu', () => {
+  test('renders default menu items in menu on right click', () => {
+    render(
+      <CodeEditorContextMenu>
+        <div data-testid="code-editor-content">content</div>
+      </CodeEditorContextMenu>,
+    );
+
+    userEvent.click(screen.getByTestId('code-editor-content'), { button: 2 });
```
TheSonOfThomp, line 75: nit: Copilot loves using `screen.getXYZ` lol. We've typically used the scoped version `const { getXYZ } = render(...)`. IMO it makes tests cleaner

TheSonOfThomp: Although, we need to use `screen.` for anything in Popovers, so this is prob. fine

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+    renderEditableContentWithContextMenu({
+      customMenuItems: [{ label: 'Custom Action', action: () => {} }],
+    });
+    userEvent.click(screen.getByTestId('code-editor-content'), { button: 2 });
+    expect(
+      screen.getByRole('menuitem', { name: 'Custom Action' }),
+    ).toBeInTheDocument();
+  });
+
+  test('calls writeText() with the selected text when copy is clicked', () => {
```
TheSonOfThomp, line 91: nit: I'd ideally like to see the `copy`/`cut`/`paste` functionality isolated and tested outside of the menu interaction. That way we can determine whether any issue is in React or just the DOM

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -350,56 +352,65 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
...
+            minHeight,
+            maxHeight,
+            className,
+            copyButtonAppearance,
+          })}
+          data-lgid={lgIds.root}
+          {...rest}
+        >
+          {panel && (
+            <div data-no-context-menu="true">
```
TheSonOfThomp, line 375: should we test this? (We're already testing `data-no-context-menu` in the ContextMenu itself, so could go either way

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
tsck: Lol all of these tests are actually handcoded now, so this was admittedly a ME choice. For some reason I thought I remembered this being the preference, that's the only reason I used it. Maybe I've been hanging out with copilot too much lol

Definitely okay with changing this across the tests in this PR.

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
tsck: I'm okay with that. That is essentially all this test suite is really testing though since the interactions are tested in the base `ContextMenu` tests. I can create a utils with just the handlers and test them in iso, but then I'm not sure this suite is really needed for anything?

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
tsck: This and disabled were removed, but these were forgotten about. Something copilot did when it went rogue early on that I removed. I'll remove it

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.types.ts
tsck: Oh yeah, good call

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: We don't actually have logic to determine which click triggers it since it triggers when the `contextmenu` event fires. I don't think we actually not have to test this

tsck: I think this is tricky. I did find [this on webaccessibility.com](https://www.webaccessibility.com/resource-library/best-practice/?bestPractice=557). Since all we provide natively are copy, paste, and cut, those are definitely redundant methods. So to that extent, I think this is accessible. Custom actions gets trickier. 

We could provide a shortcut, but that gets strange too because from what I gather there's no standard between mac and windows (I could be wrong there but I couldn't find one).

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: Arrows up an down navigate the menu, but that functionality is built into menu.

Arrows sideways and enter do nothing.

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: Yeah, I think this is valid since there is some logic making that work. I'll add a test before merging.

packages/code-editor/src/ContextMenu/ContextMenu.styles.ts
tsck: Actually I think I must of added this at the beginning and forgot about it. I don't even think it's needed. WIll remove

packages/code-editor/src/ContextMenu/ContextMenu.tsx
tsck: This is an internal component so it felt to me like doing something more complex would just lead to over-engineering. This was actually only added to make sure that the panel wasn't given the custom menu. I could have probably restructured the HTML to prevent this as well, but honestly this just seemed like the simplest path. 

What type of research did you have in mind?

packages/code-editor/src/ContextMenu/ContextMenu.tsx
tsck: Actually probably! Given that this is an internal component, and this seems like a large lift at this point, I'd rather not here. But if this was ever pulled out to be used by consumers, I think this makes sense!

packages/code-editor/src/ContextMenu/ContextMenu.tsx
tsck: I wouldn't say that's the idea exactly. I just extracted this as a good practice to encapsulate the logic, but I was mindful that that way if another context menu was ever needed, or this was ever made an external component, it would be a good jumping off point. But I didn't build this with the idea that this was going to be moved as is. I think that would warrant an RFC.

packages/code-editor/src/CodeEditor.stories.tsx
tsck: Yup, there's another ticket coming to get this right. The designs didn't exist when this was implemented so no style was added then. I just moved this in this ticket.

packages/code-editor/src/ContextMenu/ContextMenu.tsx
tsck: I did consider this because the downside of my approach is that it loses the close animation. I couldn't think of a way around this and the refocus behavior was just too jarring. 

That said, I think you make a good point about accessibility.

packages/code-editor/src/ContextMenu/ContextMenu.tsx
tsck: It definitely should! lol good catch

packages/code-editor/src/ContextMenu/ContextMenu.tsx
tsck: Why not, could be cleaner

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
tsck: I'd have to dig into exactly what's done there. Since this is an internal component atm, I think I'd rather leave it for now. I think we can always refactor this if we ever decide to expose something like this

tsck:
> @tsck Does this need a changeset?

@TheSonOfThomp I originally was adding changesets to the integration branch for each feature, but stopped doing that with the idea that I'm going to open a PR to actually remove all the currently changesets, and ultimately add one changeset for all the functionality that's on the branch prior to the v0.1.0 release. Do you see any issue with this. (hopefully not because I definitely already left out a few 😅)

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Yeah, that was more or less why I left it out. Might still be worth a small test in case a regression occurs where the attribute is removed

packages/code-editor/src/ContextMenu/ContextMenu.tsx
tsck: On second thought, the tooltip will ultimately need this as well so restructuring the HTML like I had mentioned would have been short sighted on my part. This or another solution would have become necessary

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: Actually, I'm struggling to figure out how this would be done. Seems like boundingClientRect in the testing env continuously returns 0 for all coordinates. I even tried styling the container to verify it had a height and width, and using `fireEvent` with set coordinates instead, but it still showed 0 for all coordinated on the element. 

Skipping this for now but open to suggestions here.

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
tsck: On second thought I'm not sure here. Though this would give us more fine tuned control, I feel it might be getting more into the "true unit test" philosophy. A big part of the reason for separating `CodeEditorContextMenu` and `ContextMenu` was the separation of concerns. `ContextMenu` only cares about making sure a custom context menu appears as expected. `CodeEditorContextMenu` only really cares about its specific functionality - copy/cut/paste. So I don't think having this more "integration-y" is such a bad thing. 

Of course open to pushback if you feel strongly here, but probably going to skip this for now otherwise.

packages/code-editor/src/ContextMenu/ContextMenu.tsx
TheSonOfThomp: Definitely would warrant more thought & RFC. I think it'd probably be something we consider though

packages/code-editor/src/ContextMenu/ContextMenu.tsx
TheSonOfThomp: I guess within the context (heh) of CodeEditor it's internal. I was imagining there'd be some external api that might need this, but maybe not. 

By research I just mean finding whether there's a standard for doing this kind of thing

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
TheSonOfThomp: Won't consumers need to pass in the custom menu items in this api?

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
TheSonOfThomp: I think we should still test this since we could still silently break this down the line by changing the triggering event

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
TheSonOfThomp: Might be a better case for SB interaction test?

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
TheSonOfThomp: I think we should add tests for those to avoid regressions and assert that arrow keys actually work

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
TheSonOfThomp: In think that might be easier to test. One suite checks the functionality and the suite for the context just asserts that the function was called. 

We're implicitly combining two assertions here. Not the end of the world, but breaking it out might make it easier to debug down the line

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
tsck: That's a good point, but actually imo demonstrates a bug. Consumers should not be able to pass in a separator. So from a `CodeEditor` API standpoint, all they should be able to pass in is the label, action, and disabled states

packages/code-editor/src/ContextMenu/ContextMenu.tsx
tsck: Gotcha, yeah at the moment I don't believe so

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
tsck: Pushing up a fix for this now

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: Fair, adding now

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: That was my thought but would regression actually catch that? Wouldn't it just take a screenshot of the 2nd menu rendered?

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: Ok, my thought was that this is already tested in the Menu component, but happy to assert the integration is still working.

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
tsck: Actually ended up not updating this for these tests based on the popover stuff. Will keep this in mind going forth

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
tsck: We are but isn't that sort of the philosophy we've been speaking about regarding not mocking and getting too unit focused? I think if these functions were used elsewhere or exported independently I'd agree. From the viewpoint of testing actual user functionality, this is the only place they're actually used. If the functionality broke for the user, these tests should fail and we should know.

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Added

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: Pushed

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: Pushed

packages/code-editor/src/CodeEditorContextMenu/CodeEditorContextMenu.spec.tsx
TheSonOfThomp: Fair enough. As long as the functionality gets tested somewhere, I'm cool with this

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
TheSonOfThomp: yeah, my instinct is that b/c this is triggered differently from a normal menu, we should test this behavior

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
TheSonOfThomp: IMO the SB test would look something like
- open menu
- `expect` menu is open
- click somewhere else to move menu
- `expect` menu is still open
- snapshot 2nd menu

packages/code-editor/src/ContextMenu/ContextMenu.types.ts
TheSonOfThomp: Would a consumer not be allowed to create multiple sections?

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
```diff
@@ -8,11 +8,17 @@ import { ContextMenu } from './ContextMenu';
...
+const MENU_LABEL_2 = 'Menu Label 2';
 const actionMock = jest.fn();
 
 const TestComponent = (props: PropsWithChildren) => {
   return (
-    <ContextMenu menuItems={[{ label: MENU_LABEL, action: actionMock }]}>
+    <ContextMenu
+      menuItems={[
+        { label: MENU_LABEL, action: actionMock },
+        { label: MENU_LABEL_2, action: actionMock },
```
TheSonOfThomp: nit: these should probably be separate action mocks

packages/code-editor/src/ContextMenu/ContextMenu.styles.ts
```diff
@@ -9,7 +9,7 @@ export const getMenuContainerStyles = (position: {
   y: number;
 }) => css`
   position: absolute;
-  z-index: 50;
+  /* z-index: 50; */
```
TheSonOfThomp: rm

TheSonOfThomp (approved):

packages/code-editor/src/ContextMenu/ContextMenu.spec.tsx
tsck: Good point

PR #3110, opened 2025-09-08
[LG-5480] chore: Enables React 17 testing 

TheSonOfThomp:
## ✍️ Proposed changes

Creates `scripts/react17/init.mjs` and re-enables React17 CI steps

Note for reviewer: 
**React 18** builds & test should pass
**React 17** builds will fail. Fixes in https://github.com/mongodb/leafygreen-ui/pull/3111 
**React 17** tests will fail. Addressed in https://github.com/mongodb/leafygreen-ui/pull/3112

### DIFF
--- a/.changeset/poly-types.md
+++ b/.changeset/poly-types.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/polymorphic': minor
+---
+
+Updates `PolymorphicRenderFunctionReturnType` to reflect the discrepancy between React 17 & React 18 function component return types - [LG-5354](https://jira.mongodb.org/browse/LG-5354)
--- a/.changeset/progress-bar-strict.md
+++ b/.changeset/progress-bar-strict.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/progress-bar': patch
+---
+
+Updates ProgressBar `useScreenReaderAnnouncer` to deterministically return a consistent status message (primarily a concern in React 17 strict mode)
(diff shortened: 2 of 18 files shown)


tsck:
@TheSonOfThomp I was mid-review on this, but then when I went to submit it I realized I couldn't because it was merged 😅. I'll try to in short summarize my review here:

---

package.json
Q: Was having two `init:react17` scripts intentional? I didn't think that was valid

---

scripts/react17/init.mjs
Correct me if I'm wrong but it doesn't appear `mergePackageVersions` and `mergeTsConfigs` are doing anything async, but are marked and treated as such.

---
scripts/react17/init.mjs
I could be following the recursive logic incorrectly, but would setting the result key when undefined prevent the comment `later objects taking precedence over earlier ones` from being accurate?
```
        } else if (result[key] === undefined) {
          result[key] = obj[key];
        }
```

TheSonOfThomp:
@tsck Lol sorry I saw the green button and just merged into the integration branch 😅

TheSonOfThomp:
> package.json
> Q: Was having two `init:react17` scripts intentional? I didn't think that was valid

No, that's fixed in the integration branch

---

> scripts/react17/init.mjs
> Correct me if I'm wrong but it doesn't appear `mergePackageVersions` and `mergeTsConfigs` are doing anything async, but are marked and treated as such.

Good catch. I was using the async versions previously, but it's easier to just use the  `XyzSync` methods

---

> scripts/react17/init.mjs
> I could be following the recursive logic incorrectly, but would setting the result key when undefined prevent the comment `later objects taking precedence over earlier ones` from being accurate?
> ```
>         } else if (result[key] === undefined) {
>           result[key] = obj[key];
>         }
> ```

It would seem so, yeah... this seems to be working though

PR #3112, opened 2025-09-09
fix: Fixes tests failing in React 17

TheSonOfThomp:

### DIFF
--- a/packages/section-nav/src/utils/transformToNestedData/transformToNestedData.spec.ts
+++ b/packages/section-nav/src/utils/transformToNestedData/transformToNestedData.spec.ts
@@ -34,14 +34,12 @@ describe('transformToNestedData', () => {
     expect(getFlatLevelDataToNestedData).toHaveBeenCalledWith(flatLevelInput);
   });
 
-  test('throws error for unsupported data type', () => {
-    const renderWithInvalidType = () => {
-      renderHook(() =>
-        // @ts-expect-error Testing with invalid type
-        transformToNestedData({ type: 'invalidType', data: {} }),
-      );
-    };
+  test('throws error for unsupported data type', async () => {
+    const { result } = renderHook(() => {
+      // @ts-expect-error Testing with invalid type
+      transformToNestedData({ type: 'invalidType', data: {} });
+    });
 
-    expect(renderWithInvalidType).toThrow('Unsupported data type: invalidType');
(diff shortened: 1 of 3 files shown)


.changeset/progress-bar-strict.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/progress-bar': patch
+---
+
+Updates ProgressBar `useScreenReaderAnnouncer` to deterministically return a consistent status message (primarily an concern in React 17 strict mode)
```
TheSonOfThomp:
```suggestion
Updates ProgressBar `useScreenReaderAnnouncer` to deterministically return a consistent status message (primarily a concern in React 17 strict mode)
```

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
```diff
@@ -246,17 +248,36 @@ describe('packages/progress-bar', () => {
       expect(screen.queryByRole('status')).toBeNull();
     });
 
-    test('updates live region text for initial value; ignores further changes until next threshold met', () => {
-      const { rerender } = render(
+    test('updates live region text for initial value', () => {
+      const { queryByRole } = render(
```
tsck: In this file sometimes we use `screen.*` and other times we're using the method returned from `render`. Is there intent behind the mismatch?

packages/resizable/src/useResizable/useResizable.spec.tsx
```diff
@@ -11,16 +11,20 @@ import { Position } from './useResizable.types';
...
       offsetWidth: 300,
       offsetHeight: 300,
       style: {
+        ...mockDiv.style,
         setProperty: jest.fn(),
         removeProperty: jest.fn(),
       },
-    },
+    } as HTMLDivElement,
   };
```
tsck, line 28: Just curious why this was needed?

packages/progress-bar/src/ProgressBar/ProgressBar.spec.tsx
TheSonOfThomp: I blame Copilot

packages/resizable/src/useResizable/useResizable.spec.tsx
TheSonOfThomp: Started getting flaky TS errors otherwise

PR #3115, opened 2025-09-11
chore: adds React17 reset script

TheSonOfThomp:
## ✍️ Proposed changes

Creates `scripts/react17/reset.mjs` to undo the changes made by `scripts/react17/init.mjs`

### DIFF
--- a/scripts/react17/reset.mjs
+++ b/scripts/react17/reset.mjs
@@ -122,7 +122,6 @@ function getModifiedTsConfigFiles() {
  * Find and reset all tsconfig.json files in package directories
  */
 async function resetTsConfigFiles() {
-  const allTsConfigPaths = findTsConfigFiles(ROOT_DIR);
   const modifiedTsConfigFiles = getModifiedTsConfigFiles();
 
   console.log(


scripts/react17/reset.mjs
```diff
@@ -0,0 +1,210 @@
...
+    if (isVerbose) {
+      console.log(`\t ✓ ${fileName} (no changes)`);
+    }
+    return;
+  }
+
+  try {
+    const cmd = `git checkout HEAD -- ${fileName}`;
+    if (isVerbose) {
+      console.log(`\t \x1b[90m${cmd}\x1b[0m`);
```
shaneeza, line 59: nit: why this vs using chalk for the color?

scripts/react17/reset.mjs
```diff
@@ -0,0 +1,210 @@
...
+  } catch (error) {
+    return [];
+  }
+}
+
+/**
+ * Find and reset all tsconfig.json files in package directories
+ */
+async function resetTsConfigFiles() {
+  const allTsConfigPaths = findTsConfigFiles(ROOT_DIR);
```
shaneeza: It doesn't look like you're using this

scripts/react17/reset.mjs
TheSonOfThomp: b/c I don't want to have to install dependencies before running this script

PR #3119, opened 2025-09-12
React 17 Compatibility Integration

TheSonOfThomp:


scripts/react17/init.mjs
```diff
@@ -0,0 +1,213 @@
...
+    console.log('pnpm-lock.yaml not found, skipping');
+  }
+
+  // Step 2: Update package.json versions
+  console.log('2️⃣ Updating package.json...');
+  await mergePackageVersions();
+
+  // Step 3: Update tsconfig files
+  console.log('3️⃣ Updating TypeScript configurations...');
+  await mergeTsConfigs();
```
tsck: Nit: these aren't async and don't need to be awaited

TheSonOfThomp:
```suggestion
  mergePackageVersions();

  // Step 3: Update tsconfig files
  console.log('3️⃣ Updating TypeScript configurations...');
  mergeTsConfigs();
```

PR #3129, opened 2025-09-17
LG-5477: Updates Menu Props

bruugey:
## ✍️ Proposed changes

Previously: Needed to add a @ts-expect-error when adding a `data-testid` to Menu
Now: MenuProps extend `React.ComponentProps<'ul'>`

### For new components

### DIFF
--- a/.changeset/three-toys-count.md
+++ b/.changeset/three-toys-count.md
@@ -2,4 +2,4 @@
 '@leafygreen-ui/menu': patch
 ---
 
-Improves MenuProps types to extend `React.ComponentProps<'div'>`
+Improves MenuProps types to extend `React.ComponentProps<'ul'>`


packages/menu/src/Menu/Menu.tsx
```diff
@@ -81,6 +81,8 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
     portalRef,
     scrollContainer,
     popoverZIndex,
+    // Extract only ul-appropriate props from rest
+    id,
     ...rest
```
TheSonOfThomp, line 87: where is `rest` used now?

packages/menu/src/Menu/Menu.tsx
```diff
@@ -264,8 +266,8 @@ export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(function Menu(
             {/* Need to stop propagation, otherwise Menu will closed automatically when clicked */}
             {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events*/}
             <ul
-              data-testid={lgIds.root}
-              {...rest}
+              data-testid={(rest as any)['data-testid'] || lgIds.root}
```
TheSonOfThomp: We should extract `testIdProp` from `rest` in the component props

```suggestion
              data-testid={testIdProp || lgIds.root}
```

TheSonOfThomp: Why do we need to cast `as any`?

packages/menu/src/Menu/Menu.tsx
bruugey: rest didnt change

packages/menu/src/Menu/Menu.tsx
bruugey: fixed!

packages/menu/src/Menu/Menu.tsx
TheSonOfThomp: previously it was spread onto the `ul` but now it's not used at all

packages/menu/src/Menu/Menu.types.ts
```diff
@@ -20,14 +20,20 @@ export const MenuVariant = {
...
   /**
    * The menu items, or submenus
    * @type `<MenuItem />` | `<SubMenu />` | `<MenuGroup />` | `<MenuSeparator />`
    */
   children: ReactNode;
 
+  /**
+   * Test id for the menu element
+   */
+  'data-testid'?: string;
```
TheSonOfThomp, line 35: Surprised this isn't valid on the ComponentProps<ul>

TheSonOfThomp (approved):

PR #3130, opened 2025-09-17
LG-5476: Updates InfoSprinkleProps

bruugey:
## ✍️ Proposed changes

InfoSprinkle now requires `children` in props

🎟 _Jira ticket:_ [LG-5476](https://jira.mongodb.org/browse/LG-5476)

### For new components

### DIFF
--- a/.changeset/cyan-walls-sell.md
+++ b/.changeset/cyan-walls-sell.md
@@ -1,5 +1,5 @@
 ---
-'@leafygreen-ui/info-sprinkle': major
+'@leafygreen-ui/info-sprinkle': patch
 ---
 
 `children` is now a required property


.changeset/cyan-walls-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/info-sprinkle': major
```
TheSonOfThomp: Re:Slack thread, since this was broken before, this can be considered a bug fix
Screenshot 2025-09-17 at 11 07 50 AM

packages/info-sprinkle/src/InfoSprinkle/InfoSprinkle.types.ts
```diff
@@ -6,6 +6,7 @@ export { Align, Justify };
 
 type ModifiedTooltipProps = Omit<
   TooltipProps,
+  | 'children'
```
TheSonOfThomp, line 9: I don't think `children` is even a valid prop on `TooltipProps`, is it?

TheSonOfThomp: Either way, I don't think listing this in the `Omit` keys is necessary, since we're overriding it below

bruugey: we're overriding it as required below. it was optional before

.changeset/cyan-walls-sell.md
bruugey: I updated - it broke one of our tests so i'm not entirely sure this is correct but happy to go forward with it as a patch if you think that's best

TheSonOfThomp (approved):

PR #3133, opened 2025-09-17
[LG-5450] refactor(CodeEditor): updates tooltip API and styling to match designs

tsck:
## ✍️ Proposed changes
- Adds `CodeEditorTooltip` component to render in a tooltip
- Updates `CodeEditor` API to work with new underlying component
- Increases visual regression coverage with addition tooltip tests
- _Drive-by:_ Moves `CodeEditor` stories under `Inputs` category in storybook

🎟 _Jira ticket:_ [LG-5450](https://jira.mongodb.org/browse/LG-5450)

## 🧪 How to test changes
- Check tooltip related stories in storybook
- All previous and added tests are passing


packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -94,8 +99,13 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
...
         isLoading: isLoadingProp,
         extensions: consumerExtensions,
-        darkMode: darkModeProp,
-        baseFontSize: baseFontSizeProp,
+        /**
+         * CodeEditorTooltip in particular renders outside of the LeafyGreenProvider
+         * so it won't be able to access the darkMode from the provider. When
+         * there's no override, we need to pass the theme from the provider.
+         */
+        darkMode: darkModeProp || theme === Theme.Dark,
```
shaneeza: Instead of `darkModeProp`, should this be `darkMode` that is returned from the `useDarkMode` hook?

```js
const { darkMode, theme } = useDarkMode(darkModeProp);
```

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * Tooltip to be used by the CodeEditor component.
+ */
+export function CodeEditorTooltip({
+  messages,
+  links,
+  darkMode,
+  baseFontSize: baseFontSizeProp,
+}: CodeEditorTooltipProps) {
+  const { theme } = useDarkMode(darkMode);
+  const baseFontSize = useBaseFontSize();
```
shaneeza: If the tooltip renders outside of leafygreenprovider, will either of these return the correct value?

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
```diff
@@ -0,0 +1,61 @@
...
+            {message}
+          </p>
+        ))}
+      </div>
+      {links?.length && links.length > 0 && (
+        <div className={getTooltipLinksContainerStyles(theme)}>
+          <ul className={tooltipLinksListStyles}>
+            {links?.map(link => (
+              <li key={link.label} className={tooltipLinksListItemStyles}>
+                <Link href={link.href} hideExternalIcon>
```
shaneeza, line 48: Could these be internal links within a NextJS app?

packages/code-editor/src/CodeEditor.stories.tsx
```diff
@@ -340,7 +298,47 @@ Typescript.args = {
   defaultValue: codeSnippets.typescript,
 };
 
-export const TooltipOnHover: StoryObj<{}> = {
+const tooltipPlayFunction = async ({
+  canvasElement,
+}: {
+  canvasElement: HTMLElement;
+}) => {
+  // Wait for diagnostic to be applied
```
shaneeza: What does this mean?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Oh I didn't know it returned that. I actually think that could replace the whole line.
```ts
 darkMode,
 ```
 Will update

File: packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx (hunk 2 above)
Comment (tsck) [reply packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx]: Yes and no. It will probably always render outside of the LG provider. That's why the prop logic is so important. The only reason I'm actually calling these are to get sensible fallbacks, but in theory since it's an internal component, I don't feel those should ever be necessary.

Given that `darkMode` is passed into `useDarkMode` that will work as an override. So as long as that's being passed correctly down from `CodeEditor`, that should return with the correct value. 

`useBaseFontSize` will return "some default". That's why the styling on L39 is prioritizing the prop and only using that response when the prop is undefined.

Let me know if I'm missing something there!

File: packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx (hunk 3 above)
Comment (tsck) [reply packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx:48]: Do you have an example of a case where we accept those?

File: packages/code-editor/src/CodeEditor.stories.tsx (hunk 4 above)
Comment (tsck) [reply packages/code-editor/src/CodeEditor.stories.tsx]: A "tooltip" is actually a [CodeMirror Diagnostic](https://codemirror.net/docs/ref/#lint.Diagnostic).

File: packages/code-editor/README.md [hunk 5]
```diff
@@ -144,6 +144,36 @@ import CloudIcon from '@leafygreen-ui/icon';
 | `label`                   | Text label for the button.                                             | `string`             | —           |
 | `onClick` _(optional)_    | Callback fired when the button is clicked.                             | `() => void`         | `undefined` |
 
+## Types and Variables
```
shaneeza: Thoughts on separating types and variables?

packages/code-editor/README.md
```diff
@@ -144,6 +144,36 @@ import CloudIcon from '@leafygreen-ui/icon';
...
+#### `CodeEditorTooltip`
+
+TypeScript interface defining the structure for tooltips displayed on hover in the editor:
+
+| Name                        | Description                                                 | Type                                      | Default     |
+| --------------------------- | ----------------------------------------------------------- | ----------------------------------------- | ----------- |
+| `baseFontSize` _(optional)_ | Font size of text in the tooltip.                           | `14 \| 16`                                | `14`        |
+| `messages` _(optional)_     | Messages to display in the tooltip.                         | `Array<string>`                           | `undefined` |
+| `links` _(optional)_        | Links to display in a section at the bottom of the tooltip. | `Array<{ label: string; href: string; }>` | `undefined` |
+| `darkMode` _(optional)_     | Determines if the tooltip appears in dark mode.             | `boolean`                                 | `undefined` |
```
shaneeza: Are we following the current tooltip pattern, where the tooltip is light in dark mode and dark in light mode?

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
shaneeza: I found this in `Menu`, https://github.com/mongodb/leafygreen-ui/tree/main/packages/menu#usage-with-nextjs-link-components. I believe that internal links within a nextjs app need to use `Link` from `next/link`. https://nextjs.org/docs/pages/api-reference/components/link

packages/code-editor/README.md
tsck: Oh, good catch! No we definitely shouldn't. Let me fix the type and omit that for the prop

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
tsck: Is the suggestion to accept an array of links as opposed to just a config?

packages/code-editor/README.md
tsck: I think that makes sense!

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -94,8 +99,13 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
         onChange: onChangeProp,
         isLoading: isLoadingProp,
         extensions: consumerExtensions,
-        darkMode: darkModeProp,
-        baseFontSize: baseFontSizeProp,
+        baseFontSize: baseFontSizeProp || baseFontSize,
```
shaneeza: I'm not blocking, but I'm just curious. I was going to suggest using `useUpdatedBaseFontSize` instead of `useBaseFontSize` like `Code`, but I just noticed that `Code` uses `13` and `16` while `CodeEditor` uses `14` and `16`. Is that intentional?

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
shaneeza: Maybe, if we need to support links from nextjs. Or maybe you could also add `as` to the config.

packages/code-editor/src/CodeEditor.stories.tsx
```diff
@@ -340,7 +298,47 @@ Typescript.args = {
   defaultValue: codeSnippets.typescript,
 };
 
-export const TooltipOnHover: StoryObj<{}> = {
+const tooltipPlayFunction = async ({
```
shaneeza: Thoughts on moving this to an interaction story test file since it's testing user interactions, such as hovering over text? We do this in `Table`, `Drawer`, `ProgressBar`, and `Toolbar`.

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
shaneeza: I see what you're saying. Is this beneficial over setting defaults where you destructure `darkMode` and `baseFontSize`?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: No, this should actually use 13 but since `useBaseFontSize` defaults to 14 and 16, and `useUpdatedBaseFontSize` is deprecated, my assumption was that we now wanted 14 when it says 13.

~~Actually `useUpdatedBaseFontSize` defaults to 14 too, no? https://github.com/mongodb/leafygreen-ui/blob/main/packages/typography/src/utils/useUpdatedBaseFontSize.ts. Oh but it takes an override, maybe that's the difference.~~ This is incorrect. Very weird situation lol
 
~~Going to reach out to the team in #b-team and see what should be done here.~~ Going to look at this a bit and see what makes sense. It should be 13 one way or another.

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
tsck: I'm not sure I'm following. Can you further explain what you mean?

The only benefit these have is basically grabbing defaults in cases where props aren't passed and it's outside of a provider. Are you seeing a risk in doing this?

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
tsck: It's not an explicit requirement that was laid out in advance but if we do otherwise, I guess we should. I'll look and see how this is done in `Menu` and see what makes sense

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
shaneeza: no, i don't think there is a risk. I was just wondering if there was a benefit over something like this:

```js
export function CodeEditorTooltip({
  messages,
  links,
  darkMode = false,
  baseFontSize: baseFontSizeProp = 13,
}: CodeEditorTooltipProps) {
```

BTW, this isn't blocking; I was just curious.

But I do have a follow-up about the component rendering outside of the `LeafyGreenProvider` because I think I overlooked what it means. Why does this render outside the provider?

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
tsck: Actually, I'm unsure here. The design eng spec specifically marks these as `External Link`'s. Perhaps that actually indicates that this should always be just that. `MenuItem`'s `as` prop provides full polymorphic functionality beyond Next `Link`'s, and they default to a `button` as opposed to a `a`. So I'm wondering if the use case there is a bit different.

I'm also thinking that polymorphic behavior is additive - we could always add it later, but once it's there, we have to support it.

Curious what you think here?

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: For future context - I did update this to use `useUpdateBaseFontSize` and reached out to the team about why this is deprecated.

packages/code-editor/src/CodeEditor.stories.tsx
tsck: Sure, I guess we could. Will update.

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
tsck: > no, i don't think there is a risk. I was just wondering if there was a benefit over something like this: 

Oh I see. The benefit in my eyes is that it keeps up with the global source of truth. If for example we chose to use a default `baseFontSize` of 15 tomorrow, the benefit of having a global hook is that it can be changed in one place and be immediately propagated everywhere. By relying on the hook default, the component would automatically get the new default, whereas Inlining would break the font size of the component.

> But I do have a follow-up about the component rendering outside of the LeafyGreenProvider because I think I overlooked what it means. Why does this render outside the provider?

Yeah, good question. This component is actually rendered inline as a string and injected into a dom element that will later be rendered into a tooltip by CodeMirror. By rendering it inline in isolation, it doesn't appear to have access to the context of the parent that's calling the hook where it's rendered (anecdotally, based on my testing, but also it makes sense to me that it would work like this). https://github.com/mongodb/leafygreen-ui/blob/LG-5450/cs-tooltip-refactor/packages/code-editor/src/CodeEditor/hooks/extensions/useTooltipExtension.ts#L61-L72

packages/code-editor/src/CodeEditorTooltip/CodeEditorTooltip.tsx
shaneeza: Since the specs say these are external links, we should be ok, and as you've mentioned, we can always add these in later.

packages/code-editor/src/Panel/Panel.testUtils.tsx
```diff
@@ -329,18 +329,15 @@ async function hoverCopyButton() {
...
-  if (expectedTitle === '') {
-    const titleElements = document.querySelectorAll('.leafygreen-ui-12f55h');
-    return Array.from(titleElements).some(el => el.textContent === '');
-  }
-
-  return screen.queryByText(expectedTitle) !== null;
+function getTitleText(): string {
+  // Query by the LGID used in Panel.tsx for the panel title
+  const titleElement = document.querySelector(
+    '[data-lgid="lg-code_editor-panel_title"]',
```
shaneeza, line 338: I don't think this will work if the consumer is using a custom data-lgid. In `code`, in order to use the `select` test utils, we have to pass in the dynamic lgid and I think you'll have to do something similar here.
https://github.com/mongodb/leafygreen-ui/blob/73d25ee63f9d006db4581d593084051ada56e9c3/packages/code/src/testing/getTestUtils.ts#L37-L38

shaneeza: discussed over slack that this is for internal testing only

PR #3139, opened 2025-09-19
feat(drawer): add focus management to embedded drawers

shaneeza:
## ✍️ Proposed changes

This PR enhances the drawer component by adding comprehensive focus management for embedded drawers. The changes ensure that when an embedded drawer is opened, focus automatically moves to the first focusable element within the drawer, and when closed, focus is restored to the previously focused element (typically the trigger button). This improves keyboard navigation and accessibility compliance.

🎟️ _Jira ticket:_ No Jira ticket found

## ✅ Checklist

## 🧪 How to test changes

1. **Test embedded drawer focus management:**
   - Open an embedded drawer using a toolbar button or main content button
   - Verify focus moves to the first focusable element in the drawer
   - Close the drawer and verify focus returns to the original trigger element

2. **Test overlay drawer behavior:**
   - Open an overlay drawer and verify existing focus behavior is maintained
   - Close the drawer and verify focus returns correctly

3. **Test accessibility with screen readers:**
   - Open a drawer and verify the live region announces the drawer title
   - Navigate using keyboard only to ensure all focus transitions work smoothly
   
4. **Run the new interaction stories:**
   - `EmbeddedToolbarIsFocusedOnClose` and `EmbeddedButtonIsFocusedOnClose`
   - `OverlayToolbarIsFocusedOnClose` and `OverlayButtonIsFocusedOnClose`


packages/drawer/src/testing/getTestUtils.tsx
```diff
@@ -38,19 +38,14 @@ export const getTestUtils = <T extends HTMLElement = HTMLElement>(
...
-   * properties for `display`, `opacity`, and `visibility`
+   * Checks the `aria-hidden` attribute
    */
   const isOpen = () => {
     const element = getDrawer();
 
     const isAriaVisible = element.getAttribute('aria-hidden') === 'false';
-    const { display, opacity, visibility } = window.getComputedStyle(element);
 
-    const isCSSVisible =
```
shaneeza, line 50: I'm removing this check because Drawer no longer updates these properties directly in the component, meaning the values were always the same regardless of whether the drawer was open or closed.

packages/drawer/src/DrawerToolbarLayout/DrawerToolbarLayout/DrawerToolbarLayout.interactions.stories.tsx
```diff
@@ -274,6 +274,83 @@ const playClosesDrawerWhenActiveItemIsRemovedFromToolbarData = async ({
   });
 };
 
+// Reusable play function for testing focus management with toolbar buttons
```
stephl3, line 281: looks like there is a flaky snapshot because it's taking before animations/repositioning completes. can we add snapshot delays to ensure UI is stable before capturing? https://www.chromatic.com/docs/delay/

packages/drawer/src/utils/getLgIds.ts
```diff
@@ -8,4 +8,5 @@ export const getLgIds = (root: LgIdString = DEFAULT_LGID_ROOT) =>
     closeButton: `${root}-close_button`,
     scrollContainer: `${root}-scroll_container`,
     toolbar: `${root}-toolbar`,
+    resizer: `${root}-resizer`,
```
stephl3, line 11: curious what is the reason for including this in this PR?

packages/drawer/src/testing/getTestUtils.tsx
stephl3: noted, is there something else we should check in addition to the aria attribute?

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -134,27 +136,58 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
 
+    const previouslyFocusedRef = useRef<HTMLElement | null>(null);
+    const hasHandledFocusRef = useRef<boolean>(false);
+
     /**
-     * Focuses the first focusable element in the drawer when the animation ends. We have to manually handle this because we are hiding the drawer with visibility: hidden, which breaks the default focus behavior of dialog element.
+     * Focuses the first focusable element in the drawer when the drawer is opened.
+     * Also handles restoring focus when the drawer is closed.
      *
+     * This is only necessary for embedded drawers. Overlay drawers use the native focus behavior of the dialog element.
```
stephl3: `queryFirstFocusableElement` has [a set list of focusable selectors](https://github.com/mongodb/leafygreen-ui/blob/6c511bc7cbb5066bc68c278a0ca0718cfb26b167/packages/lib/src/queryFocusableElements/queryFocusableElements.ts#L13-L20) it considers and "the native focus behavior of the dialog element" is to look for children with the `autofocus` attribute. However, these don't actually cover all cases that consumers may want to focus, such as a code editor.

After receiving feedback about this use case, it led to [this change reintroducing the `initialFocus` prop in `Modal`.](https://github.com/mongodb/leafygreen-ui/pull/3202) We don't use `Modal` and `Drawer` in all the same ways, but seeing this change makes me think they should follow the same focus management pattern. We may be able to use a similar method as  `focusModalChildElement` for the first part of this effect

packages/drawer/src/Drawer/Drawer.tsx
```diff
@@ -134,27 +136,58 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+      } else if (!open && hasHandledFocusRef.current) {
+        // Check if the current focus is not in the drawer
+        // This means the user has navigated away from the drawer, like the toolbar, and we should not restore focus.
+        if (!ref.current?.contains(document.activeElement)) {
+          hasHandledFocusRef.current = false;
+          previouslyFocusedRef.current = null;
+          return;
+        }
+
+        // Restore focus when closing (only if we had handled focus during this session)
```
stephl3: This is starting to feel like a re-implementation of the `<dialog>` element's focus patterns. Is there an explicit ticket requesting this functionality? If so, I'm curious what do you think about using the `<dialog>` under the hood for `displayMode="embedded"`? Since we're using a non-modal `<dialog>` element, we already end up having to apply the overlay styles for `displayMode="overlay"`. The benefit would be that we get focus management for free; trade-off is that we'd be _incorrectly_ using a dialog element

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: I like this pattern. However, I think if we go with this approach, then we might have to override the default dialog behavior since `initialFocus` does not work by default. 

So the pattern would be in a native dialog and the embedded div:
1) Drawer opens
2) We check the initialFocus prop
    a) if auto, we find the autofocus attribute or the first focusable item
    b) if provided, we focus that provided element
    c) if null, we don't focus anything
    
We'll end up losing the native dialog focus management this way, but it does become more customizable. 

However, I am curious about the modal focus management. Why do we need a ref option when you can add `autofocus` to the element?

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: There was no ticket requesting this. I just observed that the embedded drawer was not accessible when it was opened. And you're correct, I tried to mimic the native `dialog` behavior. I like that if we use a non-modal `dialog`, then we get this behavior for free, but I don't know if using `dialog` incorrectly is worth it. We're trying to be more intentional, and this seems semantically incorrect. It's not really a dialog, and screen readers would announce it as a dialog as well. 

Also, If we end up adding the same focus management as the modal, then we probably won't even be using the free focus management from dialog.

packages/drawer/src/utils/getLgIds.ts
shaneeza: I can remove this. I thought I was going to end up needing it.

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: Curious, why do you consider using a dialog incorrect in this case? Is it because a dialog is supposed to float on top of content?

packages/drawer/src/Drawer/Drawer.tsx
stephl3: What is the native dialog focus management you're concerned about losing? In `Modal`, `initialFocus` is mostly additive and doesn't conflict

`autofocus` can only be applied to the typical focusable elements but can't be used for something like a code editor. Although the initial implementation of `Modal` used a css selector like `"#my-code-editor"`, I added a ref option because it's more type-safe and robust than a string

packages/drawer/src/Drawer/Drawer.tsx
stephl3: Yes, my understanding is that a dialog is intended for modal or non-modal overlays. What makes the embedded drawer not accessible? I'm curious what determines focus jumping to the embedded drawer as accessible vs not doing that?

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: > What is the native dialog focus management you're concerned about losing?

Not necessarily losing, but having to re-implement it. I noticed here that if any child has `autofocus`, we let the browser focus it automatically, but we manually focus the first focusable element. So in this case, aren't we manually overriding the default behavior of finding the first focusable element and re-implementing it?  
https://github.com/mongodb/leafygreen-ui/blob/f6058dffbca9c2a5d64118351b2101bee4c0f632/packages/modal/src/utils/focusModalChildElement.ts#L44-L54

> autofocus can only be applied to the typical focusable elements but can't be used for something like a code editor. Although the initial implementation of Modal used a css selector like "#my-code-editor", I added a ref option because it's more type-safe and robust than a string

Thanks for clarifying.

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: My concern about the embedded drawer is that, when it is opened, the focus does not shift to the drawer. It's hard to find out what the correct accessibility behavior of the embedded drawer should be since this is a div and not a dialog. Since the overlay drawer is sitting on top of the content, it makes sense that the focus should automatically move to the drawer. For the embedded drawer, the drawer opens and shifts the content, but the focus remains on the trigger (if there is a trigger). The consumer would have to do a lot of tabbing to get to the drawer.

The research I've done on this hasn't been that helpful, but I still think I should do some more investigating before moving forward with this change.

packages/drawer/src/Drawer/Drawer.tsx
stephl3: > So in this case, aren't we manually overriding the default behavior of finding the first focusable element and re-implementing it?

`focusModalChildElement` is the logic we use to determine how a child element should be focused. The consumer has full control using either the `initialFocus` prop or the `autoFocus` attribute on a child element. If `initialFocus` is not specified and a child element does not have `autoFocus` specified, then we query for the first focusable element. Happy to huddle on this if further clarification is needed!

packages/drawer/src/Drawer/Drawer.tsx
stephl3: I see, this sounds like it may vary depending on product usage. Can we pull design in and maybe discuss usage with product engineers before applying to all cases?

Alternatively, we could put this behind a discriminated union prop and make this opt-in. i.e.
- if `overlay`, rely on `autoFocus` being set (although this also may require `initialFocus`)
- if `embedded`, check prop to determine focus management. Could see this working in a few ways; a couple ideas are:
  - default focus first focusable (breaking change)
  - default no focus (minor change) and allow users to specify `"auto"` which would focus first focusable or specify an element ref

packages/drawer/src/Drawer/Drawer.tsx
shaneeza: Yes, actually would love to huddle on this, I'll Slack you

PR #3140, opened 2025-09-19
[LG-5526] feat(ESLint plugin) Adds 'no-render-hook-from-rtl` rule

TheSonOfThomp:
## ✍️ Proposed changes
[LG-5526](https://jira.mongodb.org/browse/LG-5526)

Since we need to run our tests in both React 17 & React 18 environments, we need to use custom import logic for `renderHook`. This means we need to import `renderHook` from `@leafygreen-ui/testing-lib` instead of RTL or RHTL, otherwise tests will fail in one env. or the other.


tools/eslint-plugin/src/tests/no-render-hook-from-rtl.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      code: `import { renderHook, type RenderHookResult } from '@testing-library/react-hooks'`,
+      output: `import { renderHook, type RenderHookResult } from '@leafygreen-ui/testing-lib'`,
+      errors: [
+        {
+          messageId: 'issue:noRenderHookFromRTLHooks',
+        },
+      ],
+    },
+  ],
+});
```
shaneeza, line 89: Do we also need to test `import type { RenderHook } ....`?

tools/eslint-plugin/README.md
```diff
@@ -46,15 +46,12 @@ Optionally configure the rules you want to use under the rules section.
...
-| [standard-testid](docs/rules/standard-testid.md)         | Enforce a consistent prefix for hard-coded `data-testid` attributes      |                     | ![badge-internal][] | 🔧 |    |
+🔧 Automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/user-guide/command-line-interface#--fix).
+
+| Name                                                             | Description                                                                                                       | 💼                  | 🔧 |
+| :--------------------------------------------------------------- | :---------------------------------------------------------------------------------------------------------------- | :------------------ | :- |
+| [no-indirect-imports](docs/rules/no-indirect-imports.md)         | Forbid indirect imports from `src/` or `packages/`                                                                | ![badge-internal][] | 🔧 |
+| [no-render-hook-from-rtl](docs/rules/no-render-hook-from-rtl.md) | Disallow importing `renderHook` or related types from `@testing-library/react` or `@testing-library/react-hooks`. |                     | 🔧 |
```
shaneeza: Why are we removing the other rules from the readme?

tools/eslint-plugin/src/tests/no-render-hook-from-rtl.spec.ts
TheSonOfThomp: Wouldn't hurt, yeah

tools/eslint-plugin/README.md
TheSonOfThomp: Only 2 exist now: `no-indirect-imports` and `no-render-hook-from-rtl`

tools/eslint-plugin/src/tests/no-render-hook-from-rtl.spec.ts
TheSonOfThomp: It won't autofix to `import type {}`, but added it to the valid code tests

PR #3145, opened 2025-09-23
[LG-5525] refactor(Panel): Convert `Panel` into compound component `CodeEditor.Panel`

tsck:
## ✍️ Proposed changes
Convert `Panel` into compound component `CodeEditor.Panel`

🎟 _Jira ticket:_ [LG-5525](https://jira.mongodb.org/browse/LG-5525)

## 🧪 How to test changes
- "Panel" story still works as expected in storybook
- All tests are passing


packages/code-editor/src/testing/getTestUtils.spec.tsx
```diff
@@ -154,7 +158,11 @@ describe('getTestUtils', () => {
   describe('getPanelUtils', () => {
     test('`getPanelElement` returns correct element when panel is present', async () => {
       await act(() => {
-        render(<TestComponent panel={<Panel />} />);
+        render(
+          <TestComponent>
+            <CodeEditor.Panel />
+          </TestComponent>,
+        );
```
stephl3, line 165: nit: could wrap some of these in a `describe` block with a `beforeEach` to DRY up these tests

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
```diff
@@ -6,10 +7,29 @@ import { BaseFontSize } from '@leafygreen-ui/tokens';
...
+
+/**
+ * Type representing the possible static property names for CodeEditor subcomponents.
+ */
+export type CodeEditorSubcomponentProperty =
+  (typeof CodeEditorSubcomponentProperty)[keyof typeof CodeEditorSubcomponentProperty];
+
+export type PanelType = ForwardRefExoticComponent<PanelProps> & {
+  isLGPanel?: boolean;
+};
```
stephl3, line 31: nice! I like organizing these together in the *.types.ts file. Not sure why I made a separate constants file for the `*SubcomponentProperty` enum. I will move those in message

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
```diff
@@ -6,10 +7,29 @@ import { BaseFontSize } from '@leafygreen-ui/tokens';
...
+} as const;
+
+/**
+ * Type representing the possible static property names for CodeEditor subcomponents.
+ */
+export type CodeEditorSubcomponentProperty =
+  (typeof CodeEditorSubcomponentProperty)[keyof typeof CodeEditorSubcomponentProperty];
+
+export type PanelType = ForwardRefExoticComponent<PanelProps> & {
+  isLGPanel?: boolean;
```
stephl3: I'm updating this in `Message.types.ts` as well

```suggestion
  [CodeEditorSubcomponentProperty.Panel]?: boolean;
```

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -299,7 +311,12 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
         const Language = modules?.['@codemirror/language'];
         const docLength = editorViewRef.current?.state.doc.length;
 
-        if (Language && Language.forceParsing && docLength > 0) {
+        if (
+          Language &&
+          Language.forceParsing &&
+          docLength !== undefined &&
+          docLength > 0
```
stephl3: nit: unless `docLength` can actually be `undefined` in some cases, I wonder if it might make sense to update `docLength` to always be a number

```
const docLength = editorViewRef.current?.state.doc.length ?? 0;
```

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
```diff
@@ -137,15 +157,15 @@ export type CodeEditorProps = DarkModeProps &
...
+     * Determines the appearance of the copy button when no panel is present. The copy button allows the code block to be copied to the user's clipboard by clicking the button.
      *
      * If `hover`, the copy button will only appear when the user hovers over the code block. On mobile devices, the copy button will always be visible.
      *
      * If `persist`, the copy button will always be visible.
      *
      * If `none`, the copy button will not be rendered.
      *
-     * Note: 'panel' cannot be used with `copyButtonAppearance`. Either use `copyButtonAppearance` or `panel`, not both.
+     * Note: When a `<CodeEditor.Panel>` child component is present, this prop is ignored as the panel provides its own copy button.
```
stephl3, line 168: I think this was why I was thinking `CodeEditor.CopyButton` could make sense. What I'm curious about is if there are consumers who might want to use the hover/persist copy button with a panel instance that has `showCopyButton={false}`.

Is there good reasoning for why that's not allowed? It seems like something that will get asked about

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: I like that! Will update

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
tsck: Makes sense!

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
tsck: This is a good question. The only reasons I have it restricted is because we have this restriction in `Code` and because it is basically defined to not have the panel and the hover button at the same time in the design spec: https://www.figma.com/design/QsOruUtjXisdkdZHJh9CsG/Code-Editor?node-id=415-5528&t=Lsi8QEW0lOz4BMIb-4. 

From a technical standpoint, I don't have any inherent problem removing the restriction though.

packages/code-editor/src/testing/getTestUtils.spec.tsx
tsck: I'm not a big fan of layered nesting. I find that over time nested blocks become very difficult to maintain, so I admittedly try to avoid them. I have a similar mindset to `beforeEach` hooks when I can avoid them. This is a personal take, but I don't like when something that's not explicitly called in the test is called in the background. Again, I just find that overtime this makes test suites more difficult to reason with. 

Kent C. Dodds has a good article that speaks to this a bit here: https://kentcdodds.com/blog/avoid-nesting-when-youre-testing

stephl3: Appreciate the read! These test cases still seem like appropriate `beforeEach` cases to me, but I can see the case for erring on the side of avoiding to reduce maintenance burden

packages/code-editor/src/CodeEditor/CodeEditor.types.ts
stephl3: Gotcha, that sounds good to me!

packages/code-editor/src/testing/getTestUtils.spec.tsx
tsck: I would disagree here. I honestly find very few cases where I think the implicit behavior is justifiable in a test. Even in this case, if we did feel an extraction was necessary, I would prefer that we extract the repeatable logic into a function that's called explicitly in each test. This isn't something where the benefit is immediately noticeable, but I have found it make a very big difference when maintaining large test suites over time.

PR #3148, opened 2025-09-24
refactor(input-bar): use assistantName from LeafyGreenChatContext for loading message

stephl3:
## ✍️ Proposed changes

This PR refactors the `InputBar` component to use the `assistantName` value from `LeafyGreenChatContext` instead of a separate `loadingMessage` prop. This change simplifies the API by removing the need for consumers to pass a custom loading message, while making the loading text consistent with the assistant name configured in the chat provider.

The key changes include:
- Removed `loadingMessage` prop from `InputBar`, `InputBarFeedback`, and shared types
- Updated `InputBarFeedback` to use `useLeafyGreenChatContext()` to get the `assistantName`
- Changed loading message from static text to dynamic `"${assistantName} is thinking"`
- Updated tests and stories to remove references to the removed prop
- Updated README documentation to reflect the API changes

This ensures that the loading message always matches the configured assistant name in the chat provider, providing a more consistent user experience.

## ✅ Checklist

## 🧪 How to test changes

1. **Test default behavior:**
   - Render an `InputBar` with `state="loading"` inside a `LeafyGreenChatProvider`
   - Verify the loading message displays "MongoDB Assistant is thinking" (default assistant name)

2. **Test custom assistant name:**
   - Wrap `InputBar` in a `LeafyGreenChatProvider` with a custom `assistantName` prop (e.g., "Claude")
   - Set `state="loading"` on the InputBar
   - Verify the loading message displays "[Custom Name] is thinking"

3. **Verify API changes:**
   - Confirm that the `loadingMessage` prop is no longer accepted by `InputBar`
   - Check that TypeScript compilation succeeds without the removed prop
   - Verify that existing functionality (error states, send functionality) remains unchanged

### DIFF
--- a/chat/input-bar/src/InputBar/InputBar.spec.tsx
+++ b/chat/input-bar/src/InputBar/InputBar.spec.tsx
@@ -302,11 +302,26 @@ describe('packages/input-bar', () => {
   });
 
   describe('status states', () => {
-    test('renders loading state when state is "loading"', () => {
+    test('renders loading state with default assistantName when state is "loading"', () => {
       renderInputBar({ state: State.Loading });
 
       expect(
         screen.getByText(/MongoDB Assistant is thinking/i),
+      ).toBeInTheDocument();
+    });
+
+    test('renders loading state with custom assistantName when state is "loading"', () => {
+      render(
+        <LeafyGreenChatProvider
+          variant={Variant.Compact}
+          assistantName="Custom Assistant"
+        >
+          <InputBar state={State.Loading} />
+        </LeafyGreenChatProvider>,
+      );
+
+      expect(
+        screen.getByText(/Custom Assistant is thinking/i),
       ).toBeInTheDocument();
     });
 


chat/input-bar/src/InputBar/InputBarFeedback.tsx
```diff
@@ -52,7 +52,7 @@ export const InputBarFeedback = forwardRef<
           <div className={loadingContainerStyles}>
             <AssistantAvatar darkMode={darkMode} size={20} />
             <Body className={getLoadingTextStyles(theme)}>
-              {loadingMessage || messages.loading}
+              {`${assistantName} is thinking`}
```
TheSonOfThomp, line 55: Will folks still want to customize the "is thinking" part of the string? 

OTOH, this locks down the copy, which might be a good thing

chat/input-bar/src/InputBar/InputBar.spec.tsx
```diff
@@ -302,23 +302,14 @@ describe('packages/input-bar', () => {
   });
 
   describe('status states', () => {
-    test('renders loading state with default message when state is "loading"', () => {
+    test('renders loading state when state is "loading"', () => {
```
TheSonOfThomp: Should we add a unit test that verifies the prop/context value is consumed here?

TheSonOfThomp (approved):

shaneeza (commented):
If you haven't already, I would give Mohamed a heads-up, since he added this feature last week.

chat/input-bar/src/InputBar/InputBarFeedback.tsx
stephl3: I checked with design, and we landed on this vs full customization for now. May revisit in the future

PR #3143, opened 2025-09-22
Add message promotions to lg-chat

hschawe:
## ✍️ Proposed changes

Adds a promoted content section to lg-chat messages. This will be used by the MongoDB AI Assistant to promote skills during conversations.

🎟 _Jira ticket:_ [EAI-1318](https://jira.mongodb.org/browse/EAI-1318])

[Figma](https://www.figma.com/design/qZxxWKLt7kWMcWbZfvt6vK/EXPO-5323-Skills-in-Atlas-and-Chat?node-id=2726-10013) 

### For bug fixes, new features & breaking changes

- [x] I have added stories/tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `pnpm changeset` and documented my changes

## 🧪 How to test changes

1. Links open in new tab
      - Using storybook, click promotion links and verify they open in new tab
2. Keyboard navigation
    - Again on storybook, test keyboard navigation highlights the promotion's "Learn more" link using tab button

PR #3149, opened 2025-09-24
Add message promotions to lg-chat 

hschawe:
## ✍️ Proposed changes

Adds a promoted content section to lg-chat messages. This will be used by the MongoDB AI Assistant to promote skills during conversations.

Duplicate of #3143 - needed to raise PR from this repo.

🎟 _Jira ticket:_ [EAI-1318](https://jira.mongodb.org/browse/EAI-1318])

[Figma](https://www.figma.com/design/qZxxWKLt7kWMcWbZfvt6vK/EXPO-5323-Skills-in-Atlas-and-Chat?node-id=2726-10013) 

## 🧪 How to test changes

1. Links open in new tab
      - Using storybook, click promotion links and verify they open in new tab
2. Keyboard navigation
    - Again on storybook, test keyboard navigation highlights the promotion's "Learn more" link using tab button

### DIFF
--- a/chat/message/README.md
+++ b/chat/message/README.md
@@ -155,31 +155,6 @@ const MessageWithLinks = () => {
 };
 ```
 
-### Message.VerifiedBanner
-
-```tsx
-import React from 'react';
-import {
-  LeafyGreenChatProvider,
-  Variant,
-} from '@lg-chat/leafygreen-chat-provider';
-import { Message } from '@lg-chat/message';
-
-const MessageWithVerifiedBanner = () => {
-  return (
-    <LeafyGreenChatProvider variant={Variant.Compact}>
-      <Message isSender={false} messageBody="Test message">
-        <Message.VerifiedBanner
-          verifier="MongoDB Staff"
-          verifiedAt={new Date('2024-03-24T16:20:00Z')}
-          learnMoreUrl="https://mongodb.com/docs"
-        />
-      </Message>
-    </LeafyGreenChatProvider>
-  );
-};
-```
-
 ### Message.Promotion
 
 ```tsx
@@ -200,6 +175,31 @@ const MessageWithPromotion = () => {
           promotionText="Go learn more about this skill!"
           promotionUrl="https://learn.mongodb.com/skills"
           onPromotionClick={handlePromotionClick}
+        />
+      </Message>
+    </LeafyGreenChatProvider>
+  );
+};
+```
+
+### Message.VerifiedBanner
+
+```tsx
+import React from 'react';
+import {
+  LeafyGreenChatProvider,
+  Variant,
+} from '@lg-chat/leafygreen-chat-provider';
+import { Message } from '@lg-chat/message';
+
+const MessageWithVerifiedBanner = () => {
+  return (
+    <LeafyGreenChatProvider variant={Variant.Compact}>
+      <Message isSender={false} messageBody="Test message">
+        <Message.VerifiedBanner
+          verifier="MongoDB Staff"
+          verifiedAt={new Date('2024-03-24T16:20:00Z')}
+          learnMoreUrl="https://mongodb.com/docs"
         />
       </Message>
     </LeafyGreenChatProvider>
@@ -250,20 +250,21 @@ const Example = () => {
   ];
 
   const handleLinkClick = () => console.log('Link clicked');
+  const handlePromotionClick = () => console.log('Promotion clicked');
 
   return (
     <LeafyGreenChatProvider variant={Variant.Compact}>
       <Message isSender={false} messageBody="Test message">
+        <Message.Promotion
+          promotionText="Go learn more about this skill!"
+          promotionUrl="https://learn.mongodb.com/skills"
+          onPromotionClick={handlePromotionClick}
+        />
         <Message.Actions
           onClickCopy={handleCopy}
           onClickRetry={handleRetry}
           onRatingChange={handleRatingChange}
           onSubmitFeedback={handleSubmitFeedback}
-        />
-        <Message.Promotion
-          promotionText="Go learn more about this skill!"
-          promotionUrl="https://learn.mongodb.com/skills"
-          onPromotionClick={() => console.log('Promotion clicked')}
         />
         <Message.VerifiedBanner
           verifier="MongoDB Staff"
--- a/chat/message/README.md
+++ b/chat/message/README.md
@@ -414,7 +414,6 @@ The `MessagePromotion` component displays promotional content with an award icon
 #### Rendering Behavior
 
 - If `promotionText` is empty or undefined, the component returns `null` and does not render anything
-- The component supports markdown content by default
 - If `promotionUrl` is provided, a "Learn More" link is displayed that opens in a new tab
 - If `promotionUrl` is not provided or is an empty string, no "Learn More" link is rendered
 - The component only renders in the `compact` variant when used as a compound component
--- a/chat/message/README.md
+++ b/chat/message/README.md
@@ -292,9 +292,9 @@ const Example = () => {
 | `markdownProps`                   | [`LGMarkdownProps`](https://github.com/mongodb/leafygreen-ui/tree/main/chat/lg-markdown#properties) | Props passed to the internal ReactMarkdown instance                               |                                                                                                                           |
 | `messageBody`                     | `string`                                                                                            | Message body text passed to LGMarkdown                                            |                                                                                                                           |
 | `onLinkClick`                     | `({ children: string; imageUrl?: string }) => void`                                                 | A callback function that is called when the link is clicked.                      |                                                                                                                           |
-| `onPromotionClick`                | `() => void`                                                                                        | Callback function for when promotional content is clicked           |                                                                                                                           |
-| `promotion`                       | `string`                                                                                            | Text to render as promotional content on the message            |                                                                                                                           |
-| `promotionUrl`                    | `string`                                                                                            | URL for the promotion to link the "Learn more" to                |                                                                                                                           |
+| `onPromotionClick`                | `() => void`                                                                                        | Callback function for when promotional content is clicked                         |                                                                                                                           |
+| `promotion`                       | `string`                                                                                            | Text to render as promotional content on the message                              |                                                                                                                           |
+| `promotionUrl`                    | `string`                                                                                            | URL for the promotion to link the "Learn more" to                                 |                                                                                                                           |
 | `sourceType`                      | `'markdown' \| 'text'`                                                                              | Determines the rendering method of the message                                    |                                                                                                                           |
 | `verified`                        | `{ verifier?: string; verifiedAt?: Date; learnMoreUrl?: string; }`                                  | Sets if an answer is "verified" and controls the content of the message banner.   |                                                                                                                           |
 | `...`                             | `HTMLElementProps<'div'>`                                                                           | Props spread on the root element                                                  |                                                                                                                           |
@@ -323,14 +323,14 @@ const Example = () => {
 
 ### Message.Promotion
 
-| Prop                                 | Type                              | Description                                                   | Default     |
-| ------------------------------------ | --------------------------------- | ------------------------------------------------------------- | ----------- |
-| `promotionText`                      | `string`                          | Promotion text content.                                       |             |
-| `promotionUrl` _(optional)_          | `string`                          | Promotion URL for the "Learn More" link.                      |             |
-| `baseFontSize` _(optional)_          | `BaseFontSize`                    | Base font size.                                               |             |
-| `onPromotionClick` _(optional)_      | `() => void`                      | Promotion onClick callback handler.                           |             |
-| `markdownProps` _(optional)_         | `Omit<LGMarkdownProps, 'children'>` | Props passed to the internal ReactMarkdown instance.         |             |
-| `...`                                | `HTMLElementProps<'div'>`         | Props spread on the root element                              |             |
+| Prop                            | Type                                | Description                                          | Default |
+| ------------------------------- | ----------------------------------- | ---------------------------------------------------- | ------- |
+| `promotionText`                 | `string`                            | Promotion text content.                              |         |
+| `promotionUrl` _(optional)_     | `string`                            | Promotion URL for the "Learn More" link.             |         |
+| `baseFontSize` _(optional)_     | `BaseFontSize`                      | Base font size.                                      |         |
+| `onPromotionClick` _(optional)_ | `() => void`                        | Promotion onClick callback handler.                  |         |
+| `markdownProps` _(optional)_    | `Omit<LGMarkdownProps, 'children'>` | Props passed to the internal ReactMarkdown instance. |         |
+| `...`                           | `HTMLElementProps<'div'>`           | Props spread on the root element                     |         |
 
 ### Message.VerifiedBanner
 
--- a/chat/message/README.md
+++ b/chat/message/README.md
@@ -61,7 +61,7 @@ return (
 
 ### Compound Components
 
-The `Message` component uses a compound component pattern, allowing you to compose different parts of a message using subcomponents like `Message.Actions`, `Message.Links`, and `Message.VerifiedBanner`.
+The `Message` component uses a compound component pattern, allowing you to compose different parts of a message using subcomponents like `Message.Actions`, `Message.Links`, `Message.Promotion`, and `Message.VerifiedBanner`.
 
 **Note 1:** All compound components only render in the `compact` variant.  
 **Note 2:** The layout and order of compound components are enforced by the `Message` component itself. Even if you change the order of subcomponents in your JSX, they will be rendered in the correct, intended order within the message bubble. This ensures consistent UI and accessibility regardless of how you compose your message children.
@@ -180,6 +180,33 @@ const MessageWithVerifiedBanner = () => {
 };
 ```
 
+### Message.Promotion
+
+```tsx
+import React from 'react';
+import {
+  LeafyGreenChatProvider,
+  Variant,
+} from '@lg-chat/leafygreen-chat-provider';
+import { Message } from '@lg-chat/message';
+
+const MessageWithPromotion = () => {
+  const handlePromotionClick = () => console.log('Promotion clicked');
+
+  return (
+    <LeafyGreenChatProvider variant={Variant.Compact}>
+      <Message isSender={false} messageBody="Test message">
+        <Message.Promotion
+          promotionText="Go learn more about this skill!"
+          promotionUrl="https://learn.mongodb.com/skills"
+          onPromotionClick={handlePromotionClick}
+        />
+      </Message>
+    </LeafyGreenChatProvider>
+  );
+};
+```
+
 ### Complete Example with All Subcomponents
 
 ```tsx
@@ -233,6 +260,11 @@ const Example = () => {
           onRatingChange={handleRatingChange}
           onSubmitFeedback={handleSubmitFeedback}
         />
+        <Message.Promotion
+          promotionText="Go learn more about this skill!"
+          promotionUrl="https://learn.mongodb.com/skills"
+          onPromotionClick={() => console.log('Promotion clicked')}
+        />
         <Message.VerifiedBanner
           verifier="MongoDB Staff"
           verifiedAt={new Date('2024-03-24T16:20:00Z')}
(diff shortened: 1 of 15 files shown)


chat/message/src/index.ts
```diff
@@ -32,3 +32,5 @@ export {
   MessageLinks,
   type MessageLinksProps,
 } from './MessageLinks';
+export { MessagePromotion } from './MessagePromotion/MessagePromotion';
+export { type MessagePromotionProps } from './MessagePromotion/MessagePromotion.types';
```
stephl3:
```suggestion
export {
  MessagePromotion,
  type MessagePromotionProps,
} from './MessagePromotion';
```

chat/message/src/MessagePromotion/MessagePromotion.tsx
```diff
@@ -0,0 +1,56 @@
...
+            {' '}
+            <Link
+              href={promotionUrl}
+              onClick={onPromotionClick}
+              baseFontSize={baseFontSize}
+            >
+              Learn More
+            </Link>
+          </>
+        )}
```
stephl3: Is there an updated figma I can refer to? This design that I'm referring to looks like we might want to be less structured so the link can be interpolated more flexibly
Screenshot 2025-09-24 at 2 43 35 PM

chat/message/src/MessagePromotion/MessagePromotion.types.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Promotion URL.
+   */
+  promotionUrl?: string;
+
+  /**
+   * Base font size.
+   */
+  baseFontSize?: BaseFontSize;
```
stephl3: We can remove this for now since all font sizes for chat are the same

chat/message/src/MessagePromotion/MessagePromotion.types.ts
```diff
@@ -0,0 +1,32 @@
+import { LGMarkdownProps } from '@lg-chat/lg-markdown';
+
+import { HTMLElementProps } from '@leafygreen-ui/lib';
+import { BaseFontSize } from '@leafygreen-ui/tokens';
+
+export interface MessagePromotionProps
+  extends Omit<HTMLElementProps<'div'>, 'children'> {
```
stephl3: We actually now use `React.ComponentPropsWithoutRef` instead of `HTMLElementProps`

chat/message/src/MessagePromotion/MessagePromotion.types.ts
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Promotion text content.
+   */
+  promotionText: string;
+
+  /**
+   * Promotion URL.
+   */
+  promotionUrl?: string;
```
stephl3: Shared some earlier feedback about making this component less structured to support more flexibility with link placement. If that isn't actually the case and we do want this more structured approach, what is the reasoning for not making this a required prop?

chat/message/src/MessagePromotion/MessagePromotion.types.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Promotion onClick callback handler
+   */
+  onPromotionClick?: () => void;
+
+  /**
+   * Props passed to the internal ReactMarkdown instance
+   */
+  markdownProps?: Omit<LGMarkdownProps, 'children'>;
```
stephl3: Is this necessary for this component? It doesn't look like it's getting used

chat/message/src/MessagePromotion/MessagePromotion.tsx
```diff
@@ -0,0 +1,56 @@
...
+  baseFontSize,
+  promotionText,
+  promotionUrl,
+  onPromotionClick,
+  markdownProps,
+  ...rest
+}: MessagePromotionProps) {
+  if (!promotionText) {
+    return null;
+  }
```
stephl3: Probably can remove this since `promotionText` is a required prop

chat/message/src/MessagePromotion/MessagePromotion.tsx
```diff
@@ -0,0 +1,56 @@
...
+  promotionContainerStyles,
+  promotionIconStyles,
+  promotionLearnMoreStyles,
+} from './MessagePromotion.styles';
+import { type MessagePromotionProps } from './MessagePromotion.types';
+
+/**
+ * Renders promotional content with an award icon and optional "Learn More" link.
+ *
+ * @returns The rendered promotional message component, or null if no promotionText is provided
```
stephl3: would also update this since `promotionText` is required

chat/message/src/MessagePromotion/MessagePromotion.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return null;
+  }
+
+  return (
+    <div className={promotionContainerStyles}>
+      <div className={promotionIconStyles}>
+        <Icon glyph="Award" />
+      </div>
+      <Body
+        className={promotionLearnMoreStyles}
```
stephl3: I believe you can drop the `promotionLearnMoreStyles` and instead render this `Body` instance as a span by setting `as="span"`

I don't believe you'll need the `color: inherit;` style

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
+import { css } from '@leafygreen-ui/emotion';
+import { palette } from '@leafygreen-ui/palette';
+import { spacing } from '@leafygreen-ui/tokens';
+
+export const promotionContainerStyles = css`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  margin: ${spacing[200]}px 0 0 0;
```
stephl3: The `Message` component itself already has spacing handled here: https://github.com/mongodb/leafygreen-ui/blob/main/chat/message/src/MessageContainer/MessageContainer.styles.ts#L8-L15

So it's better to drop this `margin`

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+`;
+
+export const promotionIconStyles = css`
+  box-sizing: border-box;
```
stephl3: Based off this design, I think we'd want to continue using the default `box-sizing` value of `content-box`

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+  display: flex;
+  flex-direction: row;
```
stephl3, line 16: nit: I think it's fine to leave this off since it's the default value, but no harm including it

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+  align-items: center;
+  justify-content: center;
+
+  width: ${spacing[900]}px;
+  height: ${spacing[600]}px;
+  margin-right: ${spacing[200]}px;
```
stephl3: Can we instead use `gap` in `promotionContainerStyles`? It's the more modern way of handling spacing like this case

More here: https://blog.logrocket.com/css-gap-vs-margin/

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  border-radius: ${spacing[600]}px;
```
stephl3:
```suggestion
  border-radius: ${borderRadius[600]}px;
```

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+  border: 1px solid ${palette.green.light2};
+  background-color: ${palette.green.light3};
+
+  & svg {
+    color: ${palette.green.dark2};
+  }
```
stephl3: I was about to say that these need dark mode support as well. And then it hit me that a lot of these styles are already handled in the `Badge` from `@leafygreen-ui/badge`. It may be easier to start with that and pass in additional styles (width, height, padding) via the `className` prop
```
<Badge variant={Variant.Green} className={badgeStyles}><Icon ... /></Badge>
```

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const promotionIconStyles = css`
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+
+  width: ${spacing[900]}px;
+  height: ${spacing[600]}px;
```
stephl3: can we instead assign these as constant values?
```suggestion
  width: ${BADGE_WIDTH}px;
  height: ${BADGE_HEIGHT}px;
```

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  & svg {
+    color: ${palette.green.dark2};
+  }
+`;
+
+export const promotionLearnMoreStyles = css`
+  display: inline;
+  color: inherit;
+`;
```
stephl3: Probably can be removed after updating the `Body` instance to render `as="span"`

chat/message/src/MessagePromotion/MessagePromotion.spec.tsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { axe } from 'jest-axe';
+
+import { MessagePromotion, type MessagePromotionProps } from '.';
+
+jest.mock('@lg-chat/lg-markdown', () => ({
+  LGMarkdown: jest.fn(({ children }) => <div>{children}</div>),
+}));
```
stephl3: Necessary?

chat/message/src/MessagePromotion/MessagePromotion.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+describe('MessagePromotion', () => {
+  describe('a11y', () => {
+    test('does not have basic accessibility issues', async () => {
+      const { container } = renderMessagePromotion();
+      const results = await axe(container);
+      expect(results).toHaveNoViolations();
+    });
+  });
+
+  describe('rendering', () => {
```
stephl3, line 33: Several of these can probably be removed since we have snapshot testing from Chromatic. For rendering, we'll usually only test conditional prop-based rendering

chat/message/src/MessagePromotion/MessagePromotions.stories.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { StoryMetaType } from '@lg-tools/storybook-utils';
+import { StoryFn } from '@storybook/react';
+
+import { MessagePromotion, MessagePromotionProps } from '.';
+
+const meta: StoryMetaType<typeof MessagePromotion> = {
```
stephl3, line 7: Instead of this format, can you update to leverage the `generate` functionality for the different prop combinations? Here are a couple examples:
- https://github.com/mongodb/leafygreen-ui/blob/main/packages/badge/src/Badge.stories.tsx
- https://github.com/mongodb/leafygreen-ui/blob/main/chat/message-actions/src/MessageActions.stories.tsx

Probably just need a `LiveExample` story and `Generated` story

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,35 @@
+import { css } from '@leafygreen-ui/emotion';
+import { palette } from '@leafygreen-ui/palette';
+import { spacing } from '@leafygreen-ui/tokens';
+
+export const promotionContainerStyles = css`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
```
stephl3, line 10: forgot to note that this should be `flex-start` based off designs

chat/message/src/MessagePromotion/MessagePromotion.tsx
hschawe: we had a meeting with design last Friday and decided we'd change the promotion link design to instead use a "Learn more" external link. that's also why i made the promotionUrl optional, in case we don't want to append that link at all. 
(& the figma has been updated!)

chat/message/src/MessagePromotion/MessagePromotion.types.ts
hschawe: see - https://github.com/mongodb/leafygreen-ui/pull/3149#discussion_r2377150587 
basically, it's optional so users have more control over including "Learn more" component. we can change this to be required if you think this is unnecessary

chat/message/src/MessagePromotion/MessagePromotion.tsx
stephl3: Great! Can we validate that we actually want to support cases where a link is not provided? It seems like it almost should _always_ be included because the text by itself doesn't seem to serve much purpose and feels like clutter

chat/message/src/MessagePromotion/MessagePromotion.types.ts
stephl3: Yes I think this should be required unless explicitly requested by your PM or design. Or I guess if the API sometimes doesn't provide a link although it seems like it always should

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
hschawe: we may need to have some extra top spacing but it's a bit unclear, working on getting some clarification

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
hschawe: interesting, i didn't know about this... it would've saved me a lot of headache in the past haha. thanks!

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
hschawe: (as you know,) we've decided to use the standard line height instead of adding extra top spacing

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
hschawe: with the recent [chat figma updates](https://www.figma.com/design/L41rvEwQmJJ8VzO4hCPc7Y/LeafyGreen-Chat-Library?node-id=4667-16281&t=1RTtxNBKtWMSlfO2-0), looks like it is `center` now

chat/message/src/Message.stories.tsx
```diff
@@ -118,6 +118,15 @@ const getLinksChild = () => (
   />
 );
 
+const getPromotionsChild = () => (
```
stephl3: nit
```suggestion
const getPromotionChild = () => (
```

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,22 @@
...
+`;
+
+export const badgeStyles = css`
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+
+  width: ${BADGE_WIDTH}px;
+  height: ${BADGE_HEIGHT}px;
+  border-radius: ${borderRadius[600]}px;
```
stephl3: nit: this border-radius is probably already set in `Badge.styles.ts` if you wanted to drop

chat/message/src/MessagePromotion/MessagePromotion.types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * Promotion URL.
+   */
+  promotionUrl: string;
+
+  /**
+   * Promotion onClick callback handler
+   */
+  onPromotionClick?: () => void;
```
stephl3: missed this earlier; what do you think about naming this more explicitly? maybe something like `onPromotionLinkClick`?

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -0,0 +1,22 @@
+import { css } from '@leafygreen-ui/emotion';
+import { borderRadius, spacing } from '@leafygreen-ui/tokens';
+
+const BADGE_HEIGHT = spacing[600];
+const BADGE_WIDTH = spacing[900];
+
+export const promotionContainerStyles = css`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
```
stephl3, line 10: last thing I'd check for is that design confirms this value

comment thread here: https://www.figma.com/design/L41rvEwQmJJ8VzO4hCPc7Y?node-id=4667-16281#1443084964

chat/message/src/MessagePromotion/MessagePromotions.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+  title: 'Composition/Chat/MessagePromotion',
+  component: MessagePromotion,
+  parameters: {
+    default: 'WithPromotionTextAndUrl',
+    generate: {
```
stephl3, line 12: fantastic!! 🚀

chat/message/README.md
```diff
@@ -286,6 +322,17 @@ const Example = () => {
 | `onLinkClick` _(optional)_ | `({ children: string }) => void` | A callback function that is called when any link is clicked. |                       |
 | `...`                      | `HTMLElementProps<'div'>`        | Props spread on the root element                             |                       |
 
+### Message.Promotion
+
+| Prop                            | Type                                | Description                                          | Default |
+| ------------------------------- | ----------------------------------- | ---------------------------------------------------- | ------- |
+| `promotionText`                 | `string`                            | Promotion text content.                              |         |
+| `promotionUrl` _(optional)_     | `string`                            | Promotion URL for the "Learn More" link.             |         |
```
stephl3: no longer optional

chat/message/README.md
```diff
@@ -286,6 +322,17 @@ const Example = () => {
...
+| `baseFontSize` _(optional)_     | `BaseFontSize`                      | Base font size.                                      |         |
```
stephl3: rm?

chat/message/README.md
```diff
@@ -286,6 +322,17 @@ const Example = () => {
...
+| `onPromotionClick` _(optional)_ | `() => void`                        | Promotion onClick callback handler.                  |         |
+| `markdownProps` _(optional)_    | `Omit<LGMarkdownProps, 'children'>` | Props passed to the internal ReactMarkdown instance. |         |
```
stephl3: rm?

chat/message/README.md
```diff
@@ -361,6 +408,21 @@ The component manages its own internal state for:
 
 - Expansion state: Controls whether the links section is expanded or collapsed
 
+### Message.Promotion
+
+The `MessagePromotion` component displays promotional content with an award icon and optional "Learn More" link.
```
stephl3: no longer optional

chat/message/src/MessagePromotion/MessagePromotion.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Body, Link } from '@leafygreen-ui/typography';
+
+import {
+  badgeStyles,
+  promotionContainerStyles,
+} from './MessagePromotion.styles';
+import { type MessagePromotionProps } from './MessagePromotion.types';
+
+/**
+ * Renders promotional content with an award icon and optional "Learn More" link.
```
stephl3: no longer optional

chat/message/src/MessagePromotion/MessagePromotion.types.ts
hschawe: i generally like more explicit names like `onPromotionLinkClick` and think it's preferable here, the only problem is that it could be slightly confused with `onLinkClick` for the references.

chat/message/README.md
hschawe: I also updated the promotion Rendering behavior section

chat/message/README.md
```diff
@@ -259,7 +292,7 @@ const Example = () => {
...
 | `messageBody`                     | `string`                                                                                            | Message body text passed to LGMarkdown                                            |                                                                                                                           |
-| `onLinkClick`                     | `({ children: string; imageUrl?: string }) => void`                                                 | A callback function that is called when the link is clicked.                      |                                                                                                                           |
+| `onLinkClick`                     | `({ children: string; imageUrl?: string }) => void`                                                 | A callback function that is called when the link is clicked.                      |
```
stephl3: is this supposed to be changed?

chat/message/README.md
```diff
@@ -361,6 +403,20 @@ The component manages its own internal state for:
...
+
+#### Rendering Behavior
+
+- Both `promotionText` and `promotionUrl` are required for this component to render.
+- The promotion component renders after the Message body with an `Award` badge to distinguish it from the main body text.
+- The "Learn more" link is displayed inline with the promotion text.
+
+#### Callback Behavior
+
+- The `onPromotionLinkClick` callback is triggered when clicking the "Learn More" link
```
stephl3: Can we drop these subsections? don't think they are necessary

hschawe: both of them? i am find to drop the callback behavior but all the other compound components document their rendering behavior

chat/message/README.md
hschawe: nope, this was accidental

PR #3153, opened 2025-09-24
[LG-5553] feat(hooks): add useControlled hook for generic controlled/uncontrolled state management

shaneeza:
## ✍️ Proposed changes

This PR introduces a new `useControlled` hook that provides a more generic version of controlled/uncontrolled state management. The hook enables components to seamlessly work in both controlled and uncontrolled modes, handling the state management logic internally. Additionally, the existing `useControlledValue` hook has been refactored to use `useControlled` under the hood, promoting code reuse and consistency.

🎟️ _Jira ticket:_ [LG-5553](https://jira.mongodb.org/browse/LG-5553)

## ✅ Checklist

## 🧪 How to test changes

1. Import and use the new `useControlled` hook in a React component
2. Test controlled mode by passing a value and onChange handler
3. Test uncontrolled mode by omitting the value prop and providing only an initialValue
4. Verify that the hook correctly manages state in both modes
5. Run the existing test suite to ensure the refactored `useControlledValue` hook still works as expected
6. Check that the hook properly handles different data types (string, number, boolean, array, object, Date)

### DIFF
--- a/packages/hooks/src/useControlled/useControlled.ts
+++ b/packages/hooks/src/useControlled/useControlled.ts
@@ -58,7 +58,7 @@ export const useControlled = <T extends any>(
   useEffect(() => {
     if (isUndefined(controlledValue) && isUndefined(initialValue)) {
       consoleOnce.error(
-        `Warning: \`useControlled\` hook is being used without a value or initialValue. If using an input, this will cause a React warning when an input changes. Please decide between using a controlled or uncontrolled input element, and provide either a controlledValue or initialValue to \`useControlled\``,
+        `Warning: \`useControlled\` hook is being used without a value or initialValue. If using an input, this will cause a React warning when an input changes. Please decide between using a controlled or uncontrolled element, and provide either a controlledValue or initialValue to \`useControlled\``,
       );
     }
   }, [controlledValue, initialValue]);


packages/preview-card/src/PreviewCard/PreviewCard.tsx
```diff
@@ -37,7 +37,7 @@ export const PreviewCard = forwardRef<HTMLDivElement, PreviewCardProps>(
     fwdRef,
   ) => {
     const { darkMode, theme } = useDarkMode(darkModeProp);
-    const { value: isOpen, updateValue } = useControlledValue(
+    const { value: isOpen, updateValue } = useControlledInputValue(
       controlledIsOpen,
       onOpenChange,
       defaultOpen,
```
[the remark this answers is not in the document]
shaneeza: Will open a ticket to update this since the onOpenChange type will need to be updated

packages/context-drawer/src/ContextDrawer/ContextDrawer.tsx
```diff
@@ -40,7 +40,7 @@ export const ContextDrawer = forwardRef<HTMLDivElement, ContextDrawerProps>(
     fwdRef,
   ) => {
     const { darkMode, theme } = useDarkMode(darkModeProp);
-    const { value: isOpen, updateValue } = useControlledValue(
+    const { value: isOpen, updateValue } = useControlledInputValue(
       controlledIsOpen,
       onOpenChange,
       defaultOpen,
```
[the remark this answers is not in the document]
shaneeza: Will open a ticket to update this since the onOpenChange type will need to be updated

chat/message-rating/src/MessageRating/MessageRating.tsx
```diff
@@ -55,7 +55,7 @@ export const MessageRating = forwardRef(
     const likeButtonRef = useRef<HTMLButtonElement | null>(null);
     const dislikeButtonRef = useRef<HTMLButtonElement | null>(null);
 
-    const { value, handleChange, updateValue } = useControlledValue<
+    const { value, handleChange, updateValue } = useControlledInputValue<
       MessageRatingProps['value']
     >(valueProp, onChange, MessageRatingValue.Unselected);
```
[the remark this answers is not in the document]
shaneeza: this is an input

.changeset/new-dryers-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/hooks': major
+---
+
+Refactors `useControlledValue` to be more generic and adds input-specific logic to new `useControlledInputValue` hook.
```
stephl3: Have some concerns about external impact as it pertains to versioning fatigue since this is an internal-driven change and not a product team request.

Thankfully, it doesn't look like any external teams currently leverage `useControlledValue`: https://github.com/search?q=(org:10gen+OR+org:evergreen-ci+OR+org:mongodb+OR+org:mongodb-js+OR+org:mongodb-labs+OR+org:wiredtiger)+AND+useControlledValue+AND+NOT+(repo:mongodb/leafygreen-ui+OR+repo:10gen/lg-chat+OR+repo:10gen/leafygreen-ui-private+OR+repo:mongodb-js/leafygreen-ui)&type=code

However, a considerable number of teams _do_ leverage `@leafygreen-ui/hooks`: https://github.com/search?q=(org:10gen+OR+org:evergreen-ci+OR+org:mongodb+OR+org:mongodb-js+OR+org:mongodb-labs+OR+org:wiredtiger)+AND+@leafygreen-ui/hooks%22:+AND+NOT+(repo:mongodb/leafygreen-ui+OR+repo:10gen/lg-chat+OR+repo:10gen/leafygreen-ui-private+OR+repo:mongodb-js/leafygreen-ui)&type=code

This major version upgrade looks avoidable by:
1. adding the net-new hook and have it duplicate the refactored logic in `useControlledValue`, and
2. marking `useControlledValue` as deprecated

It also looks like `useControlledInputValue` is intended to fully replace `useControlledValue`. Is that right? If so, what do you think about this alternative approach, so we can make this a minor version upgrade to reduce the migration overhead for the consuming teams?

shaneeza: > It also looks like useControlledInputValue is intended to fully replace useControlledValue. Is that right?

Not quite. The current `useControlledValue` hook is very input-specific. It requires a ref to the input, creates a synthetic event whenever `updateValue` is called, and `handleChange` expects a `ChangeEventHandler`. This works fine when working directly with inputs, but it's not really ideal for non-inputs, like Tooltip, that have an `open` state that can be controlled or uncontrolled.

I need a non-input version of this hook for `TimeInput`. Yes, it's technically an input, but it won't be used as a traditional input. I believe that Adam also needs this new hook for the wizard he's working on.

A compromise might be to keep the `useControlledValue` hook as-is with the input-specific logic and create a new hook without input logic, something like `useControlled`.

stephl3: If there's a way to avoid making a breaking change while still accomplishing what we need to do, that sounds like the move. Having an explicitly different name might be helpful to identify which is used when. If there's enough logic the 2 share, you could make a base hook for the duplicated logic.

TheSonOfThomp: I agree that we should avoid a `major` change if we can

> A compromise might be to keep the `useControlledValue` hook as-is with the input-specific logic and create a new hook without input logic, something like `useControlled`

I like that idea, and the name `useControlled`! 👍

shaneeza: Sounds good, i'll update!

packages/hooks/src/useControlledValue/useControlledValue.spec.tsx
```diff
@@ -282,80 +125,17 @@ describe('packages/hooks/useControlledValue', () => {
...
-        const button = result.getByTestId('test-button');
-        userEvent.click(button);
-        expect(input).toHaveValue('carrot');
-      });
+      // Click button to trigger updateValue with ref
+      userEvent.click(button);
+      expect(input).toHaveValue('carrot');
     });
   });
 });
```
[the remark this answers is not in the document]
shaneeza: r u ok?

packages/hooks/src/useControlled/useControlled.ts
```diff
@@ -0,0 +1,71 @@
...
+    onChange?.(newVal);
+  };
+
+  /**
+   * Log a warning if neither controlled value or initialValue is provided
+   */
+  useEffect(() => {
+    if (isUndefined(controlledValue) && isUndefined(initialValue)) {
+      consoleOnce.error(
+        `Warning: \`useControlled\` hook is being used without a value or initialValue. If using an input, this will cause a React warning when an input changes. Please decide between using a controlled or uncontrolled input element, and provide either a controlledValue or initialValue to \`useControlled\``,
```
shaneeza: React only shows a warning when form inputs switch between controlled and uncontrolled, so I'm thinking this message should be in the input-specific hook, but then this should also apply to input-like components using this hook. And I think we should still warn if value or initialValue is not passed in.

PR #3156, opened 2025-09-25
feat(lib): Adds `findChildren` & restructure child query utils

TheSonOfThomp:
## ✍️ Proposed changes

1. Restructures `findChild` related utils under `lib/childQueries`
2. Creates `isChildWithProperty` utility, used by both `findChild` & `findChildren`
3. Creates `findChildren` function to find all children of a given type
4. Updates `findChild` and `findChildren` tests to use `Iterable<ReactNode>` as the test argument (to avoid confusion around nested Fragments)


packages/lib/src/childQueries/isChildWithProperty/isChildWithProperty.ts
```diff
@@ -0,0 +1,21 @@
...
+import { hasStaticProperty } from '../hasStaticProperty';
+
+export const isChildWithProperty = <S extends string>(
+  child: ReactNode,
+  staticProperty: S,
+): child is ReactElement<any, ComponentType & { [K in S]: any }> => {
+  if (!isValidElement(child)) return false;
+
+  // child must be a single node
+  if (Children.count(child) !== 1) return false;
```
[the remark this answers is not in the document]
TheSonOfThomp: No, since `child` is a `ReactNode`, it could potentially be `Iterable<ReactNode`, which would make `Children.count(child)` be greater than one

tools/install/src/ALL_PACKAGES.ts
```diff
@@ -76,6 +76,7 @@ export const ALL_PACKAGES = [
   '@leafygreen-ui/toolbar',
   '@leafygreen-ui/tooltip',
   '@leafygreen-ui/typography',
+  '@leafygreen-ui/wizard',
```
stephl3: rm?

packages/lib/src/childQueries/unwrapRootFragment/unwrapRootFragment.ts
```diff
@@ -0,0 +1,39 @@
...
+ * ```tsx
+ * unwrapRootFragment(
+ *  <Fragment>
+ *    <Foo />
+ *    <Bar />
+ *  </Fragment>
+ * ) // [<Foo />, <Bar />]
+ * ```
+ */
+export const unwrapRootFragment = (
```
stephl3, line 19: is it intentional that this one doesn't have an index file?

packages/lib/src/childQueries/unwrapRootFragment/unwrapRootFragment.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { Fragment } from 'react';
+
+import { unwrapRootFragment } from './unwrapRootFragment';
+
+describe('packages/lib/unwrapRootFragment', () => {
```
stephl3, line 5: what do you think about including a test case for a nested Fragment to show it doesn't currently unwrap past the root?

packages/lib/src/childQueries/isChildWithProperty/isChildWithProperty.ts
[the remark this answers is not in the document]
stephl3: isn't `isChildWIthProperty` only ever getting called on a single child node even if it has multiple children nested in that node? what is an example of a case with `Iterable<ReactNode>` that would get processed by this function?

packages/lib/src/childQueries/isChildWithProperty/isChildWithProperty.spec.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+      const element = <ComponentWithProps text="multi-prop" />;
+
+      expect(isChildWithProperty(element, 'isPrimary')).toBe(true);
+      expect(isChildWithProperty(element, 'isSecondary')).toBe(true);
+      expect(isChildWithProperty(element, 'isSpecial')).toBe(true);
+      expect(isChildWithProperty(element, 'nonExistent')).toBe(false);
+    });
+
+    // TODO: How should we handle `isFoo = false`?
```
stephl3: can you elaborate on this?

packages/lib/src/childQueries/findChild/findChild.tsx
```diff
@@ -0,0 +1,49 @@
+import { Children, ReactElement, ReactNode } from 'react';
+
+import { isChildWithProperty } from '../isChildWithProperty/isChildWithProperty';
+import { unwrapRootFragment } from '../unwrapRootFragment/unwrapRootFragment';
```
stephl3: nit
```suggestion
import { isChildWithProperty } from '../isChildWithProperty';
import { unwrapRootFragment } from '../unwrapRootFragment';
```

packages/lib/src/childQueries/findChildren/findChildren.ts
```diff
@@ -0,0 +1,62 @@
+import { ReactElement, ReactNode } from 'react';
+
+import { isChildWithProperty } from '../isChildWithProperty/isChildWithProperty';
+import { unwrapRootFragment } from '../unwrapRootFragment/unwrapRootFragment';
```
stephl3: nit
```suggestion
import { isChildWithProperty } from '../isChildWithProperty';
import { unwrapRootFragment } from '../unwrapRootFragment';
```

packages/lib/src/childQueries/isChildWithProperty/isChildWithProperty.spec.tsx
TheSonOfThomp: Since we check for existence of the property (not the value), this case is weird. Should this component be "found" with `findChild`? 
```tsx
const Foo = () => <div>foo</div>
Foo.isFoo = false;
```

packages/lib/src/childQueries/isChildWithProperty/isChildWithProperty.ts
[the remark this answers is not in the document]
TheSonOfThomp: In the current usage, yes. 

But as-typed,  `child` could potentially be an array of `ReactNode`

e.g.

```tsx
const renderedDiv = (
  <div>
    <Foo />
    <Bar />
  </div>
);

isChildWithProperty(
  renderedDiv.props.children, // the `child` arg here is the Iterable `[<Foo />, <Bar />]`
  'isFoo'
)
```

packages/lib/src/childQueries/isChildWithProperty/isChildWithProperty.spec.tsx
stephl3: If the static property evaluates to false, I believe `hasStaticProperty` should fail, or at least it _should_. If this test was showing otherwise, can we patch that?

If it's a matter of the name suggesting existence vs evaluation, we can rename it more explicitly

TheSonOfThomp: Oh true, if the value is "falsy" then it'll return `false`

packages/lib/src/childQueries/isChildWithProperty/isChildWithProperty.spec.tsx
```diff
@@ -0,0 +1,148 @@
...
+      const element = <ComponentWithProps text="multi-prop" />;
+
+      expect(isChildWithProperty(element, 'isPrimary')).toBe(true);
+      expect(isChildWithProperty(element, 'isSecondary')).toBe(true);
+      expect(isChildWithProperty(element, 'isSpecial')).toBe(true);
+      expect(isChildWithProperty(element, 'nonExistent')).toBe(false);
+    });
+
+    // eslint-disable-next-line jest/no-disabled-tests
+    test.skip('should handle falsy static property values', () => {
```
stephl3: still need to skip this?

PR #3157, opened 2025-09-25
feat(lib) Add deepOmit utility function

TheSonOfThomp:
Introduce a new utility function `deepOmit` that allows for deeply omitting properties from objects based on specified paths. This addition includes comprehensive tests to ensure functionality and edge case handling.


packages/lib/src/helpers/deepOmit/deepOmit.spec.ts
```diff
@@ -0,0 +1,314 @@
+import { deepOmit } from './deepOmit';
+
+describe('deepOmit', () => {
```
stephl3, line 3: have you considered adding a test case for duplicate keys that are adjacent or nested?

TheSonOfThomp: No, since that wouldn't be allowed as an input object in JS 
```js
const x = { a: 1, a: 2 } // { a: 2 }
```

stephl3: Ahh misspoke, I meant like this case:
```
const x = { a: { b: 1 }, b: 2 };
```

PR #3158, opened 2025-09-26
[LG-5570] feat(lg-markdown): expand markdown support and improve styling system

stephl3:
## ✍️ Proposed changes

This PR enhances the LGMarkdown component with expanded markdown element support and a more robust theming system. The changes improve the visual consistency and functionality of markdown rendering in chat interfaces by adding support for blockquotes, horizontal rules, and additional heading levels, while also implementing theme-aware styling for light and dark modes.

Additionally, fixes text wrapping issues in the Message component to prevent overflow problems with long content.

🎟️ _Jira ticket:_ [LG-5570](https://jira.mongodb.org/browse/LG-5570)

## ✅ Checklist

### LGMarkdown enhancements:
1. Navigate to the LGMarkdown Storybook story
2. Verify that blockquotes render with a left border wedge that changes color based on theme
3. Test horizontal rules (`---`) render as styled dividers
4. Confirm h4, h5, and h6 headings display with appropriate typography
5. Check that nested lists have proper indentation and spacing
6. Toggle between light and dark modes to verify theme-aware styling works correctly

### Message text wrapping fix:
1. Navigate to the Message component story
2. Test with long text content or URLs that exceed container width
3. Verify text wraps properly without causing horizontal overflow
4. Confirm `pre-line` preserves line breaks while allowing wrapping

### DIFF
--- a/chat/lg-markdown/src/LGMarkdown/LGMarkdown.styles.ts
+++ b/chat/lg-markdown/src/LGMarkdown/LGMarkdown.styles.ts
@@ -57,14 +57,17 @@ const getBaseStyles = (theme: Theme) => css`
     ${getHrStyles(theme)}
   }
 
-  li > :first-child {
-    padding-top: ${VERTICAL_SPACING}px;
+  li,
+  ol,
+  ul {
+    white-space: normal;
   }
 
   ol,
   ul {
+    margin: 0;
     padding-inline-start: ${LIST_INDENT_SPACING}px;
-    display: inline-flex;
+    display: flex;
     flex-direction: column;
     gap: ${VERTICAL_SPACING}px;
   }


chat/lg-markdown/src/LGMarkdown/componentsMap.tsx
```diff
@@ -31,21 +45,69 @@ const componentsMap: ReactMarkdownComponents = {
...
+  }: ComponentPropsWithoutRef<'h5'> & ExtraProps) => {
+    return <Body {...rest}>##### {children}</Body>;
+  },
+  h6: ({
+    children,
+    node: _node,
+    ...rest
+  }: ComponentPropsWithoutRef<'h6'> & ExtraProps) => {
+    return <Body {...rest}>###### {children}</Body>;
+  },
```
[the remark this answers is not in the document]
stephl3: added comment that these are intentionally not supported at this time

chat/lg-markdown/src/LGMarkdown.stories.tsx
```diff
@@ -4,13 +4,21 @@ import { StoryFn } from '@storybook/react';
...
 
 ## Heading 2
 
 ### Heading 3
 
+#### Heading 4
+
+##### Heading 5 (not supported)
+
+###### Heading 6 (not supported)
```
tsck, line 18: Q: Just curious what the value is of adding unsupported functionality?

chat/lg-markdown/src/LGMarkdown.stories.tsx
```diff
@@ -30,16 +42,33 @@ function helloWorld() {
...
+   2. Nested ordered list item 2.2
+3. Ordered list item 3
+
+Here's an unordered list:
+
+- Unordered list item 1
+- Unordered list item 2
+  - Nested unordered list item 2.1
+  - Nested unordered list item 2.2
+- Unordered list item 3
```
tsck, line 64: These don't appear to rendering correctly in storybook: https://642700aa3e49e32bdbf0b0fc-wxezzezmaw.chromatic.com/?path=/docs/composition-chat-lgmarkdown--docs

Image

chat/lg-markdown/src/LGMarkdown.stories.tsx
tsck: Oh nevermind, its because we have logic that renders the hashmarks for these.

chat/lg-markdown/src/LGMarkdown/componentsMap.tsx
```diff
@@ -31,21 +45,71 @@ const componentsMap: ReactMarkdownComponents = {
...
+    children,
+    node: _node,
+    ...rest
+  }: ComponentPropsWithoutRef<'h4'> & ExtraProps) => {
+    return (
+      <Body as="h4" weight={FontWeight.SemiBold} {...rest}>
+        {children}
+      </Body>
+    );
+  },
```
tsck, line 91: Q: Why do we render different typography elements "as" the semantically correct element?

stephl3: This is intended to match the CommonMark spec which is used by `react-markdown`: https://commonmark.org/help/

tsck: Like stylistically match it?

chat/lg-markdown/src/LGMarkdown.stories.tsx
stephl3: for future reference for ourselves!

chat/lg-markdown/src/LGMarkdown.stories.tsx
stephl3: ahh yeah it only worked on smaller breakpoints which is what I had visible previously. should be fixed now!

chat/lg-markdown/src/LGMarkdown/componentsMap.tsx
stephl3: Styles are customized for MongoDB, but the header hierarchy is intended to match commonmark. This is similarly done in README and CHANGELOG files. Here's the button changelog as an example: https://github.com/mongodb/leafygreen-ui/tree/main/packages/button

tsck: I think I'm missing something. I'm not sure how this explains why a `Body` element is used for an `H3`, for example? You're saying that it matches commonmark? In what way?

tsck (approved):

chat/lg-markdown/src/LGMarkdown/componentsMap.tsx
stephl3: Ahh I understand better what you're confused about (very reasonable because we should revisit this to make it less confusing... 😅 ). Most, if not all, of our typography components are polymorphic. They can be rendered `as` any other tag. Ideally, we should avoid doing this too freely. I consider the naming of some, like `H1`, `H2`, and `H3`, as tech debt since they should be named more generally to describe the styles that they provide or use cases that they serve. e.g. `Disclaimer`, `Overline`, or `Subtitle`

In the case of `LGMarkdown`, we are moving away from using `H1`, `H2`, and `H3` for the markdown used for `<h1>`, `<h2>`, and `<h3>` because the styles are way too big. Instead, we rely on `H3`, `Subtitle`, and `Body` for providing the right-sized styles for the markdown elements.

stephl3: @tsck

tsck: Gotcha! Thanks for the explanation. Makes much more sense now.

PR #3161, opened 2025-09-26
[LG-5563] feat(Wizard) Adds Wizard

TheSonOfThomp:


packages/wizard/src/Wizard.stories.tsx
```diff
@@ -1,17 +1,111 @@
...
+import { StoryObj } from '@storybook/react';
+
+import Card from '@leafygreen-ui/card';
 
 import { Wizard } from '.';
 
+faker.seed(0);
+
 export default {
   title: 'Components/Wizard',
```
stephl3: maybe?
```suggestion
  title: 'Composition/Data Display/Wizard',
```

packages/wizard/src/index.ts
```diff
@@ -1 +1,3 @@
-export { Wizard, type WizardProps } from './Wizard';
\ No newline at end of file
+export { Wizard, type WizardProps } from './Wizard';
+export { WizardFooter, type WizardFooterProps } from './WizardFooter';
+export { WizardStep, type WizardStepProps } from './WizardStep';
```
stephl3: are you using compound components for this? if so, I don't think you'll want to export `WizardFooter` or `WizardStep`

packages/wizard/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const WIZARD_STEP_KEY = 'isWizardStep';
+export const WIZARD_FOOTER_KEY = 'isWizardFooter';
```
stephl3: For consistency, can we use the same pattern used in `CodeEditor.types.ts` and `Message.types.ts` for storing subcomponent properties?

packages/wizard/src/WizardStep/WizardStep.tsx
```diff
@@ -1,8 +1,18 @@
...
+
+import { stepStyles } from './WizardStep.styles';
 import { WizardStepProps } from './WizardStep.types';
 
-export function WizardStep({}: WizardStepProps) {
-  return <div>your content here</div>;
+export function WizardStep({ title, description, children }: WizardStepProps) {
+  return (
+    <div className={stepStyles}>
+      <H3>{title}</H3>
```
stephl3: The `H3` component doesn't currently have the same check that `Description` does when determining what to render the final element `as`. This can currently lead to an `<h3>` tag wrapping another typography element. We should probably update the type of `title` or add that logic to the `H3` component.

packages/wizard/src/WizardStep/WizardStep.styles.ts
```diff
@@ -1,4 +1,6 @@
-
 import { css } from '@leafygreen-ui/emotion';
+import { spacing } from '@leafygreen-ui/tokens';
 
-export const baseStyles = css``;
+export const stepStyles = css`
+  padding: 0 ${spacing[1800]}px;
```
stephl3, line 5: whoa my first time seeing a `spacing[1800]`! 😮

packages/wizard/src/WizardFooter/WizardFooter.types.ts
```diff
@@ -1 +1,18 @@
...
+
+  /**
+   * Props for the cancel button (center button)
+   */
+  cancelButtonProps?: FormFooterProps['cancelButtonProps'];
+
+  /**
+   * Props for the primary button (right-most button)
+   */
+  primaryButtonProps: FormFooterProps['primaryButtonProps'];
```
stephl3, line 17: What do you think about exposing the `activeStep` in the click handlers for these buttons? I think it would be beneficial for integration of analytics. An example would be funnel metrics to determine where users are bailing from the flow

packages/wizard/src/Wizard/Wizard.spec.tsx
```diff
@@ -1,11 +1,8 @@
...
 import { render } from '@testing-library/react';
 
 import { Wizard } from '.';
 
 describe('packages/wizard', () => {
-  test('condition', () => {
-
-  })
-})
+  test('condition', () => {});
```
stephl3: Missing tests commit?

packages/wizard/src/Wizard/Wizard.styles.ts
```diff
@@ -1,4 +1,15 @@
-
 import { css } from '@leafygreen-ui/emotion';
 
-export const baseStyles = css``;
+export const wizardContainerStyles = css`
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  gap: 24px;
```
stephl3: nit
```suggestion
  gap: ${spacing[600]}px;
```

packages/wizard/src/Wizard/Wizard.styles.ts
```diff
@@ -1,4 +1,15 @@
...
+  outline: 1px solid red;
```
stephl3: rm? looks like it was for debugging

packages/wizard/src/Wizard/Wizard.tsx
```diff
@@ -1,8 +1,76 @@
...
+  } = useWizardControlledValue<number>(activeStepProp, undefined, 0);
+
+  const stepChildren = Children.toArray(children).filter(child => {
+    if (isValidElement(child)) {
+      const displayName = (child.type as any)?.displayName;
+      return displayName && displayName.includes('Step');
+    }
+
+    return false;
+  });
```
stephl3: 1. this variable name is kinda funny 😂 
2. is this goign to get refactored to use `findChildren`?

packages/wizard/src/Wizard/Wizard.tsx
```diff
@@ -1,8 +1,76 @@
...
+      value={{
+        activeStep,
+        updateStep,
+      }}
+    >
+      <div className={wizardContainerStyles}>
+        <div className={stepContentStyles}>{currentStep}</div>
+        {/* Render footer */}
+        {footerChild}
+      </div>
```
stephl3: Do you think users may ever want to render additional content between `currentStep` and `footerChild`? I wonder if it'd be valuable to preemptively add some consideration to that case if we are not too rigid about how this should be used

packages/wizard/src/Wizard/Wizard.tsx
TheSonOfThomp: 1. Better than `evilStepMother` 😂
2. Yeah, once that's in I'll use findChildren

packages/wizard/src/Wizard/Wizard.tsx
TheSonOfThomp: I think that content could just be rendered inside the `Step`. Rather not open this up too much to start

packages/wizard/src/constants.ts
TheSonOfThomp: Sure thing!

packages/wizard/src/Wizard/Wizard.tsx
stephl3: I guess it could also be added as its own subcomponent later!

packages/wizard/src/index.ts
TheSonOfThomp: Oh good point

packages/wizard/src/WizardFooter/WizardFooter.types.ts
TheSonOfThomp: We could, though that complicates the props a lot. Also, just a hunch, but I expect in cases this will be "controlled" so consumers will have access to the `activeStep` in local state

PR #3163, opened 2025-09-29
Bumps svgr from v5 to v8

bruugey:
## ✍️ Proposed changes

### For new components


packages/icon/scripts/prebuild/index.ts
```diff
@@ -117,32 +115,108 @@ function makeFileProcessor(outputDir: string, options?: PrebuildOptions) {
...
+            ],
+          },
+        },
       },
+      { componentName: file.name },
+    );
+
+    // Post-process to add custom LeafyGreen wrapper
+    const customizedSVGR = processedSVGR.replace(
+      /import \* as React from "react";\nimport type \{ SVGProps \} from "react";\nconst (\w+) = \(props: SVGProps<SVGSVGElement>\) => (.*?);\nexport default \1;/s,
```
TheSonOfThomp: Why are we removing `svgrTemplate.ts` and hard-coding it inline here?

packages/icon/scripts/prebuild/index.ts
```diff
@@ -117,32 +115,108 @@ function makeFileProcessor(outputDir: string, options?: PrebuildOptions) {
       encoding: 'utf8',
     });
 
-    // Note: must use `require` since svgrrc is a CommonJS module
-    // @ts-ignore - svgrrc is not typed
-    const svgrrc = await require('../../.svgrrc.js');
```
TheSonOfThomp, line 122: should we rm `svgrrc.js` as well?

packages/icon/scripts/prebuild/index.ts
TheSonOfThomp: SVGR 8 still takes a `template` prop. (https://react-svgr.com/docs/options/#template) IMO we should be using this instead of hard-coding a complex regex replacer

bruugey: Updated

packages/icon/scripts/prebuild/index.ts
```diff
@@ -117,32 +116,52 @@ function makeFileProcessor(outputDir: string, options?: PrebuildOptions) {
...
       svgContent,
       {
-        ...svgrrc,
-        template: svgrTemplate,
-      },
-      {
-        componentName: file.name,
+        plugins: ['@svgr/plugin-jsx'],
+        typescript: true,
+        jsx: {
```
TheSonOfThomp, line 125: can we extract this config into either a separate file, or at least a constant?

packages/icon/scripts/prebuild/index.ts
```diff
@@ -141,71 +142,15 @@ function makeFileProcessor(outputDir: string, options?: PrebuildOptions) {
...
-};
-
-$1.displayName = '$1';
-$1.isGlyph = true;
-
-export default $1;`,
-    );
-
+    // Apply custom LeafyGreen wrapper using template
+    const customizedSVGR = applyLeafyGreenTemplate(processedSVGR, file.name);
```
TheSonOfThomp, line 146: this just moves the brittle string replacement to a separate file. We should use the transform `Config.template` property

TheSonOfThomp: I noticed Cursor has difficulty doing this. There's a weird bug/quirk in the `tpl` function where it doesn't like when you concat words after interpolated var names. 

If you tell cursor to do that it works:
e.g. instead of `${variables.componentName}Props)` use `${variables.componentName + "Props)"}`

PR #3165, opened 2025-09-29
refactor(vertical-stepper): move vertical stepper package from private to leafygreen-ui

stephl3:
Moves `@lg-private/vertical-stepper` to `@leafygreen-ui/vertical-stepper`

🎟️ _Jira ticket:_ [LG-5578](https://jira.mongodb.org/browse/LG-5578)

## ✅ Checklist

- [x] I have verified the Live Example will look as intended on the design website.

### DIFF
--- a/packages/vertical-stepper/src/VerticalStepper.stories.tsx
+++ b/packages/vertical-stepper/src/VerticalStepper.stories.tsx
@@ -102,6 +102,11 @@ export default {
 
 export const LiveExample = {
   render: ({ ...args }) => <VerticalStepper {...args}></VerticalStepper>,
+  parameters: {
+    chromatic: {
+      disableSnapshot: true,
+    },
+  },
 } satisfies StoryObj<typeof VerticalStepper>;
 
 export const Generated = {


.changeset/mighty-actors-tie.md
TheSonOfThomp, line 1: Do we need a changeset? We can just merge it in and publish the latest on the next release

packages/vertical-stepper/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@leafygreen-ui/vertical-stepper",
+  "version": "0.0.1",
```
TheSonOfThomp: IMO we should probably release this initially at the same v3.0.0

.changeset/mighty-actors-tie.md
stephl3: oh hello! LOL I dropped a question on slack. since we are publishing it under the `@leafygreen-ui/*` scope, I was wondering if we should start from 1.0.0 instead

packages/vertical-stepper/CHANGELOG.md
```diff
@@ -0,0 +1,43 @@
+# @leafygreen-ui/vertical-stepper
+
+# @lg-private/vertical-stepper (deprecated)
```
TheSonOfThomp: Maybe just "Previously published as..."?

PR #3166, opened 2025-09-30
LG-5586: Update MenuItem to prevent form submission

bruugey:
## ✍️ Proposed changes

### For new components

### DIFF
--- a/packages/menu/src/MenuItem/InternalMenuItemContent.tsx
+++ b/packages/menu/src/MenuItem/InternalMenuItemContent.tsx
@@ -83,6 +83,13 @@ export const InternalMenuItemContent = React.forwardRef<
           }
         : {};
 
+    const defaultButtonProps =
+      as === 'button'
+        ? {
+            type: 'button',
+          }
+        : {};
+
     return (
       <InputOption
         ref={fwdRef}
@@ -133,6 +140,7 @@ export const InternalMenuItemContent = React.forwardRef<
           },
           className,
         )}
+        {...defaultButtonProps}
         {...defaultAnchorProps}
         {...rest}
       >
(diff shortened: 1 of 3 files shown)


packages/menu/src/MenuItem/InternalMenuItemContent.tsx
```diff
@@ -83,6 +83,13 @@ export const InternalMenuItemContent = React.forwardRef<
           }
         : {};
 
+    const defaultButtonProps =
+      as === 'button'
+        ? {
+            type: 'button',
+          }
+        : {};
+
```
TheSonOfThomp, line 92: there's probably a cleaner way to do this, but should be fine for now

packages/menu/src/MenuItem/InternalMenuItemContent.tsx
```diff
@@ -83,6 +83,13 @@ export const InternalMenuItemContent = React.forwardRef<
           }
         : {};
 
+    const defaultButtonProps =
+      as === 'button'
+        ? {
+            type: 'button',
+          }
```
TheSonOfThomp, line 90: should we also add `as` here? (or should we omit from anchor props on L80?)

packages/menu/src/MenuItem/InternalMenuItemContent.tsx
bruugey: this was the most type safe

TheSonOfThomp (approved):

packages/menu/src/MenuItem/InternalMenuItemContent.tsx
bruugey: as is already being passed in

PR #3172, opened 2025-09-30
Create TextNode component

TheSonOfThomp:
## ✍️ Proposed changes

'@leafygreen-ui/typography': minor
---

Adds `TextNode` component. 

Wraps a string in the provided `as` component,
or renders the provided `ReactNode`.

Useful when rendering `children` props that can be any react node

```tsx
<TextNode as={h1}>Hello!</TextNode> 
// Renders: <h1>Hello!</h1>
```

```tsx
<TextNode>
  <h2>Hello!</h2>
</TextNode> 
// Renders: <h2>Hello!</h2>
```


packages/typography/src/TextNode/TextNode.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const TextNode = ({
+  children,
+  as,
+}: PropsWithChildren<{ as?: PolymorphicAs }>) => {
+  return typeof children === 'string' || typeof children === 'number' ? (
+    <Polymorph as={as}>{children}</Polymorph>
+  ) : (
+    children
+  );
+};
```
stephl3, line 29: In what cases will this be used? Is it intended to be a wrapper that replaces logic elsewhere like in `Description`? If so, I'm wondering if we should add an explicit wrapper `<div>` around `children` on L27 so attributes/props can be passed to the component

shaneeza: A use case for this could be `Drawer`.  We currently have this:
https://github.com/mongodb/leafygreen-ui/blob/bda06c5d9901ebb06c71ebb078d4c9c33cf3a729/packages/drawer/src/Drawer/Drawer.tsx#L263-L270

but that could be replaced with `TextNode` like this:
```js
const Wrapper = ({ children }: PropsWithChildren<{}>) => ( 
    <Body
      as={'h2'}
      id={titleId}
      className={titleStyles}
    >
      {title}
    </Body>
);

  <TextNode as={Wrapper}>{title}</TextNode>
```

However, in this case, the consumer will lose the `id `attribute if using a `ReactNode` so being able to pass that `id` to the `ReactNode` could be beneficial.

stephl3: Good example! Yes, it seems like being able to pass attributes/props to the `TextNode` would be important

TheSonOfThomp: I can see the value in spreading props into the block that uses `Polymorph`, since that avoids the need for a Wrapper component.



```tsx
<Drawer title="Some String" />

// renders
<TextNode as={Wrapper}>{title}</TextNode>
// or
<TextNode as='h2' className="..." id="abc">{title}</TextNode>

/// renders 
<Wrapper>{children}</Wrapper>
// or
<h2 {...props}>{children}</h2>
```

But I think an internal Wrapper might be preferable to potentially overriding props that the user has provided.   

```tsx
<Drawer title={<span id="xyz">My Title</span>} />

/// renders 
<TextNode as={H2} id="abc">{title}</TextNode>

/// renders
<span id="???">My title</span>
```

TheSonOfThomp: Also, nesting in an extra `div` could make styling & layout brittle

How about this:

```suggestion
export const TextNode = ({
  children,
  as,
  ...props
}: PropsWithChildren<{ as?: PolymorphicAs }>) => { // TODO: make props Polymorphic
  return typeof children === 'string' || typeof children === 'number' ? (
    <Polymorph as={as}>{children}</Polymorph>
  ) : (
    React.cloneElement(children, {
      ...props,
      ...children.props
    })
  );
};
```

PR #3180, opened 2025-10-01
fix(message): remove standalone MessagePromotion export

stephl3:
## ✍️ Proposed changes

Removes the standalone `MessagePromotion` export from `@lg-chat/message` while keeping the `MessagePromotionProps` type export. This change aligns with the compound component pattern introduced in v8.2.0, where `MessagePromotion` should only be accessed as `Message.Promotion`, not as a standalone export.

The component remains fully functional as `Message.Promotion`, and TypeScript users can still import the `MessagePromotionProps` type for type checking.

🎟️ _Jira ticket:_ No JIRA ticket found in changeset files

## ✅ Checklist

## 🧪 How to test changes

1. Verify that `Message.Promotion` continues to work as expected when used as a compound component
2. Confirm that `MessagePromotionProps` can still be imported for TypeScript type checking
3. Verify that attempting to import `MessagePromotion` as a standalone component results in an error (expected behavior)
4. Check existing stories and tests to ensure `Message.Promotion` functionality is unchanged

### DIFF
--- a/.changeset/rude-rockets-crash.md
+++ b/.changeset/rude-rockets-crash.md
@@ -1,0 +1,5 @@
+---
+'@lg-chat/message': patch
+---
+
+Remove `MessagePromotion` export


.changeset/rude-rockets-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-chat/message': patch
```
shaneeza, line 2: Why is this only a patch?

stephl3: @shaneeza it was inadvertently exported in the previous release. I was treating this as a fix and going to deprecate the version that initially exported it since it doesn't yet have usage. Do you think that would be fine or does this need to be a major version?

shaneeza: Honestly, I'm not sure. I would bring it up with the team.

stephl3: Synced offline and we deemed patch was okay in this case despite this technically being a major

PR #3183, opened 2025-10-02
[LG-5593] fix(modal) revert change in order of `data-testid` and `...rest` in `ModalView`

kraenhansen:
## ✍️ Proposed changes

Merging this PR will:
- Propagate `data-testid` through the `ConfirmationModal` and `Modal` components.

🎟 _Jira ticket:_ [LG-5593](https://jira.mongodb.org/browse/LG-5593)

### For new components

- [ ] I have added my new package to the global tsconfig
- [ ] I have added my new package to the Table of Contents on the global README
- [ ] I have verified the Live Example will look as intended on the design website.

### For bug fixes, new features & breaking changes

- [ ] I have added stories/tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] I have run `pnpm changeset` and documented my changes

PR #3184, opened 2025-10-02
chore(packages) avoid `data-lgid` and `data-testid` attributes being duplicated in components

kraenhansen:
## ✍️ Proposed changes

Merging this PR will:
- Create and add a new (internal) eslint rule which warns when two `data-lgid` or `data-testid` attributes are referencing the same id.
- Fix existing occurrences of the issue.

### For new components

### DIFF
--- a/packages/typography/src/Description/Description.spec.tsx
+++ b/packages/typography/src/Description/Description.spec.tsx
@@ -1,18 +1,10 @@
-import React, { ReactNode } from 'react';
+import React from 'react';
 import { render } from '@testing-library/react';
-
-import { PolymorphicAs } from '@leafygreen-ui/polymorphic';
 
 import Description from './Description';
 
-const renderDescription = ({
-  as,
-  children = 'Test description',
-}: {
-  as?: PolymorphicAs;
-  children?: ReactNode;
-}) => {
-  return render(<Description as={as}>{children}</Description>);
+const renderDescription = (props: React.ComponentProps<typeof Description>) => {
+  return render(<Description {...props} />);
 };
 
 describe('Description Component', () => {
@@ -36,5 +28,24 @@ describe('Description Component', () => {
       children: <span>Test Description</span>,
     });
     expect(container.querySelector('div')).toBeInTheDocument();
+  });
+
+  test('renders with a default data-lgid', () => {
+    const { container } = renderDescription({ children: 'Test description' });
+    expect(container.querySelector('p')).toHaveAttribute(
+      'data-lgid',
+      'lg-typography-description',
+    );
+  });
+
+  test('renders with a custom data-lgid', () => {
+    const { container } = renderDescription({
+      children: 'Test description',
+      'data-lgid': 'lg-custom-lgid',
+    });
+    expect(container.querySelector('p')).toHaveAttribute(
+      'data-lgid',
+      'lg-custom-lgid',
+    );
   });
 });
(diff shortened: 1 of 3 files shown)


packages/checkbox/src/Checkbox/Checkbox.tsx
```diff
@@ -137,8 +137,8 @@ const Checkbox = React.forwardRef(
             className={cx(labelStyle, labelHoverStyle[theme], {
               [disabledLabelStyle]: disabled,
             })}
-            data-lgid={lgIds.root}
-            data-testid={lgIds.root}
+            data-lgid={lgIds.label}
```
shaneeza, line 140: One of the reasons we pass the root to label and description is because `<Label>` and `<Description>` already generate their own specific lgids internally. 

https://github.com/mongodb/leafygreen-ui/blob/9b2d1563782de4142141bbe4da4262e9cf11d2cc/packages/typography/src/Label/Label.tsx#L38-L39

So if you pass `lgids.root`, the root is `lg-checkbox` and the id generated in `<Label>` will be `lg-checkbox-label`. 

However, if we pass `lgIds.label`, that will pass `lg-checkbox-label` to  `<Label>`, which will generate `lg-checkbox-label-label`.

The first approach allows us to avoid duplication in the id string

kraenhansen: Ahh I see. Wouldn't it make more sense (and in alignment with the style-guide / BEM-like) for the `Label` component to have its own `getLgIds` function? 🤔 I mean, `Label` is not a "child element" of some "typography block".

I've [pushed a suggestion](https://github.com/mongodb/leafygreen-ui/pull/3184/commits/148ee521a904951036c5a6add92519daac717f4a) to work around this in the simplest possible way I could imagine.

https://github.com/mongodb/leafygreen-ui/blob/148ee521a904951036c5a6add92519daac717f4a/packages/typography/src/Description/Description.tsx#L36-L41

shaneeza: That is a good point. If we take this approach, we'll probably have to update some of our tests. I want to look into this a little deeper. Is this currently blocking you?

shaneeza: Also, I noticed this PR #3183 was merged in last week and it looks like some of those changes are duplicated in this PR. Where the changes in #3183 the ones that were blocking you?

packages/select/src/Select/Select.tsx
```diff
@@ -574,8 +574,8 @@ export const Select = forwardRef<HTMLDivElement, SelectProps>(
 
               {description && (
                 <Description
-                  data-lgid={lgIds.root}
-                  data-testid={lgIds.root}
+                  data-lgid={lgIds.description}
```
shaneeza, line 577: If we make this change, then we should update the test utils. Currently, in a few test utils, we are calling the typography getLgIds function with the root, but with this change, we'll no longer need to do that. 

https://github.com/mongodb/leafygreen-ui/blob/b9fb557268ed34f8d2f06563ce69117203de2d45/packages/select/src/testing/utils/getTestUtils.ts#L22

packages/checkbox/src/Checkbox/Checkbox.tsx
kraenhansen: As far as I remember, the changes in this PR are not blocking us.
It was indeed stacked on the #3183 branch and I've force-pushed a rebase 👍

Feel free to close this PR if this is not the direction you want to go 👍
I was just pretty sure I found a bug with those duplications 🙂

PR #3185, opened 2025-10-03
Fix default style (svg width/height) & css prop on promotions

hschawe:
## ✍️ Proposed changes

Fixes default style on Message.Promotion component. The style `box-sizing: border-box` was missing, causing the promotion badge size to be wrong when implemented on the MongoDB Assistant. On Storybook, `border-box` is set by default on all elements under the Story root element, so this issue was not caught previously.

I've also made it possible for users to pass in custom styles on the promotions with `className`, as it could be a helpful workaround for users if any similar issues arise.

I also updated the stories:
- Shortened the text in the promotion stories to potentially catch any future wrong-size issues.
- Added a story for the new `className` prop, passing different `box-sizing` on the promotion container.

- [*] I have added stories/tests that prove my fix is effective or that my feature works
- [*] I have added necessary documentation (if appropriate) _[unnecessary,  `HTMLElementProps<'div'>` is already documented in the readme]_

## 🧪 How to test changes

* Review custom style generated story for MessagePromotion on Storybook for blue badge color, increased gap size.
* Review promotion badge elements in stories for Message & MessagePromotion on Storybook all have correct element size (36x24)

### DIFF
--- a/.changeset/curvy-lines-tie.md
+++ b/.changeset/curvy-lines-tie.md
@@ -1,0 +1,5 @@
+---
+'@lg-chat/message': minor
+---
+
+Add `className` prop to `Message.Promotion` for custom styles
--- a/chat/message/src/MessagePromotion/MessagePromotion.styles.ts
+++ b/chat/message/src/MessagePromotion/MessagePromotion.styles.ts
@@ -9,6 +9,9 @@ export const promotionContainerStyles = css`
   flex-direction: row;
   align-items: center;
   gap: 0px ${spacing[200]}px;
+  & div {
+    box-sizing: border-box;
+  }
 `;
 
 export const badgeStyles = css`
@@ -16,6 +19,6 @@ export const badgeStyles = css`
   flex-direction: row;
   justify-content: center;
 
-  width: ${BADGE_WIDTH}px;
-  height: ${BADGE_HEIGHT}px;
+  min-width: ${BADGE_WIDTH}px;
+  min-height: ${BADGE_HEIGHT}px;
 `;
--- a/chat/message/src/MessagePromotion/MessagePromotion.tsx
+++ b/chat/message/src/MessagePromotion/MessagePromotion.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import Badge, { Variant } from '@leafygreen-ui/badge';
+import { cx } from '@leafygreen-ui/emotion';
 import Icon from '@leafygreen-ui/icon';
 import LeafyGreenProvider, {
   useDarkMode,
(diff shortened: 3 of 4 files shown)


chat/message/src/MessagePromotion/MessagePromotion.tsx
```diff
@@ -23,17 +24,16 @@ export function MessagePromotion({
...
-        <div>
-          <Badge variant={Variant.Green} className={badgeStyles}>
-            <Icon glyph="Award" />
-          </Badge>
-        </div>
+      <div className={cx(promotionContainerStyles, className)}>
+        <Badge variant={Variant.Green} className={badgeStyles}>
+          <Icon glyph="Award" />
+        </Badge>
         <Body as="span" {...rest}>
```
stephl3: can we almost move this `{...rest}` to the outer div?

chat/message/src/MessagePromotion/MessagePromotion.styles.ts
```diff
@@ -9,13 +9,16 @@ export const promotionContainerStyles = css`
   flex-direction: row;
   align-items: center;
   gap: 0px ${spacing[200]}px;
+  & div {
+    box-sizing: border-box;
+  }
```
stephl3: Is this specifically for the badge element? If so, wondering if we can add directly to `badgeStyles` below

.changeset/curvy-lines-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lg-chat/message': minor
```
stephl3: Since this is a backwards-compatible bugfix, this can be marked as a patch https://semver.org/#summary

chat/message/src/MessagePromotion/MessagePromotions.stories.tsx
```diff
@@ -13,10 +16,22 @@ const meta: StoryMetaType<typeof MessagePromotion> = {
       combineArgs: {
         darkMode: [false, true],
         promotionText: [
-          'Challenge your knowledge by earning the Advanced Schema Design skill!',
-          'Challenge your knowledge by earning the Advanced Schema Design skill! This is a really really really really really really really really really really really really really really really really really  really really really really really really really really long copy text to test how the component handles long text content.',
+          'Challenge your knowledge and earn the Schema Design badge!',
+          'Challenge your knowledge by earning the Advanced Schema Design skill! This is a really really really really really really really really really long copy text to test how the component handles multiline text content.',
         ],
         promotionUrl: ['https://learn.mongodb.com/skills'],
+        className: [
```
stephl3: We don't typically generate story instances for demonstrating `className` overrides. Although we'll expose the `className` prop for components, it's typically an exception rather than a promoted feature to keep UI consistent

.changeset/curvy-lines-tie.md
hschawe: was thinking this could be considered a minor change due to the addition of `className`

chat/message/src/MessagePromotion/MessagePromotions.stories.tsx
hschawe: fair! I thought it'd be useful to keep a `content-box` test case, but it might be more appropriate to test that on the `Badge` component anyway

.changeset/curvy-lines-tie.md
stephl3: I would mark it minor if it was net-new, but in this case, it was already added and getting passed to the `Body` instance instead of the outermost div

PR #3186, opened 2025-10-06
[LG-5098] feat(CodeEditor): adds custom search panel

tsck:
## ✍️ Proposed changes
Adds a custom search panel to the `CodeEditor` component. Contains all of the same functionality that was in the built in panel but matches the LG design language.

🎟 _Jira ticket:_ [LG-5098](https://jira.mongodb.org/browse/LG-5098)

## 🧪 How to test changes
- Pull up using CMD/CTRL+F in storybook
- All previous tests are passing
- New tests are passing


packages/code-editor/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,415 @@
...
+            placeholder="Find"
+            aria-labelledby="find"
+            onChange={handleSearchQueryChange}
+            onKeyDown={handleFindInputKeyDown}
+            className={findInputStyles}
+            value={searchString}
+            baseFontSize={baseFontSizeProp || baseFontSize}
+            darkMode={darkMode}
+            // CodeMirror looks for this attribute to refocus when CMD+F is pressed and the panel is already open
+            main-field="true"
```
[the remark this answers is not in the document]
tsck: This is a specific attribute used by CodeMirror, I don't believe adding `data` would work

packages/code-editor/src/CodeEditor/hooks/useModuleLoaders.spec.ts
```diff
@@ -95,7 +106,7 @@ describe('useModuleLoaders', () => {
         {
           line: 1,
           column: 1,
-          content: 'Error message',
+          messages: ['Error message'],
```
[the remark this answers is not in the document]
tsck: This is just a drive-by from a previous change that wasn't caught

packages/code-editor/README.md
```diff
@@ -57,6 +57,7 @@ console.log(greet('MongoDB user'));`;
...
+| `enableSearchPanel` _(optional)_      | Enables the find and replace search panel in the editor.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           | `boolean`                    | `true`      |
```
stephl3: What do you think about including a note here or elsewhere in the README about how to open/close this?

packages/code-editor/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,415 @@
...
+  baseFontSize: baseFontSizeProp,
+  hasPanel,
+}: SearchPanelProps) {
+  const [isOpen, setIsOpen] = useState(false);
+  const [searchString, setSearchString] = useState('');
+  const [replaceString, setReplaceString] = useState('');
+  const [isCaseSensitive, setIsCaseSensitive] = useState(false);
+  const [isWholeWord, setIsWholeWord] = useState(false);
+  const [isRegex, setIsRegex] = useState(false);
+  const [query, setQuery] = useState<SearchQuery>(
```
stephl3, line 66: Not sure where is the best to report this, but do we have configurability of the highlight color in light mode? The contrast is quite difficult to make out
Screenshot 2025-10-08 at 11 44 12 AM

packages/code-editor/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,415 @@
...
+    new SearchQuery({
+      search: searchString,
+      caseSensitive: isCaseSensitive,
+      regexp: isRegex,
+      wholeWord: isWholeWord,
+      replace: replaceString,
+    }),
+  );
+  const [findCount, setFindCount] = useState(0);
+  const { theme } = useDarkMode(darkMode);
```
stephl3, line 76: It looks like dark mode implementation may need another look

hovering `IconButton` instance turns svg black
Screenshot 2025-10-08 at 11 45 45 AM

focusing `IconButton` instance turns svg black
Screenshot 2025-10-08 at 11 45 51 AM

disabled/enabled `IconButton` instances look reversed; I'm not sure if that's because the background, but can we check that out? Also can compare how the format and copy buttons appear vs the ones in the `SearchPanel`
Screenshot 2025-10-08 at 11 45 59 AM

packages/code-editor/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,415 @@
...
+  const baseFontSize = useUpdatedBaseFontSize();
+  const [selectedIndex, setSelectedIndex] = useState<number | null>(null);
+  const findOptions = {
+    isCaseSensitive: 'Match case',
+    isRegex: 'Regexp',
+    isWholeWord: 'By word',
+  } as const;
```
stephl3: nit: could define these outside the component

packages/code-editor/src/CodeEditor/CodeEditor.tsx
```diff
@@ -271,7 +269,75 @@ const BaseCodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(
           ),
 
           commands.history(),
-          searchModule.search(),
+
+          enableSearchPanel && searchModule
```
stephl3: would it improve readability/maintainability to break up these code editor chunks?

packages/code-editor/src/SearchPanel/SearchPanel.styles.ts
```diff
@@ -0,0 +1,169 @@
...
+  theme: Theme;
+  isOpen: boolean;
+  baseFontSize: BaseFontSize;
+  hasPanel: boolean;
+}) =>
+  cx(getBaseContainerStyles(theme, baseFontSize), {
+    [openContainerStyles]: isOpen,
+    [css`
+      border-top-right-radius: ${borderRadius[300]}px;
+    `]: !hasPanel,
```
stephl3, line 84: This or the outer container in `CodeEditor` might need another look
Screenshot 2025-10-08 at 11 41 41 AM

packages/code-editor/src/SearchPanel/SearchPanel.styles.ts
```diff
@@ -0,0 +1,169 @@
...
+`;
+
+/**
+ * Inner section used for padding and border so that the outer section can
+ * fully close to 0px when set to 0fr.
+ */
+export const getReplaceInnerSectionStyles = (theme: Theme) => css`
+  display: flex;
+  align-items: center;
+  padding: 8px 10px 8px 44px;
```
stephl3: nit: tokens?

packages/code-editor/README.md
tsck: Can't hurt, will do!

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: Break up the whole effect? I broke this effect up a bit in the QA PR that's not merged yet. Not quite as much as we could, but I did separate the initialization from the configuration https://github.com/mongodb/leafygreen-ui/compare/LG-5446/ce-qa?expand=1#diff-85df5eb3676e4f9ebec8364e075627e7b4aa40362faa78ab3294256d336cfd91R250

packages/code-editor/src/SearchPanel/SearchPanel.tsx
tsck: Yeah, I would agree. We can make them whatever we want, these are just the defaults in CodeMirrors. @sandynguyenn this may be more a question for you - is there another color you'd prefer us to set these yellow highlights to?

packages/code-editor/src/SearchPanel/SearchPanel.styles.ts
tsck: Oof, I was trying to be cute about this and take a bit of a shortcut, but I clearly broke it again at some point. Will take a look

packages/code-editor/src/SearchPanel/SearchPanel.tsx
tsck: Spoke with Sandy who didn't feel strongly, but recommended aligning with the `Code` highlight color: https://www.figma.com/design/4h2VwjCuJUbeZ7hzD2J1rq/LeafyGreen-Design-System?m=auto&node-id=29198-325819&t=S3xLTLPhXkq6XQ3w-1

I've pushed up commit [6bf2b88](https://github.com/mongodb/leafygreen-ui/pull/3186/commits/6bf2b88be02f955e2b1701fc3cc9d68952a31ca1) to address this

tsck: Oops, need to update dark mode

packages/code-editor/src/CodeEditor/CodeEditor.tsx
stephl3: I think I was thinking along the lines of breaking out the hooks into separate pieces with explicit, readable names to quickly identify what each one is doing. This way, `CodeEditor` file is more easily parseable. Just an idea though; you can decide if that's helpful or overkill

packages/code-editor/src/SearchPanel/SearchPanel.tsx
tsck: Reverting these changes until I sync with Sandy today

packages/code-editor/src/CodeEditor/CodeEditor.tsx
tsck: I'm not opposed to this, I'd rather just consider that in the QA branch since I'm already refactoring this effect there, might be easier on the conflicts 🙂

tsck: I've actually backtracked on this. For the search panel logic, I'm creating a hook that just encapsulates this logic. Since it returns an extension, it fits nicely into the use*Extension hooks, and will actually make the conflicts easier to deal with because they'll be less changes in the `CodeEditor` component itself. Pushing shortly

packages/code-editor/src/SearchPanel/SearchPanel.tsx
tsck: Sandy and I went with 2 palette colors and made the text the light theme base font color for the highlights. This should be updated now

packages/code-editor/src/SearchPanel/SearchPanel.tsx
```diff
@@ -0,0 +1,398 @@
...
+              value={searchString}
+              // CodeMirror looks for this attribute to refocus when CMD+F is pressed and the panel is already open
+              main-field="true"
+              // @ts-expect-error - The TextInput component forwards refs, but the types do not explicitly include the `ref` prop.
+              ref={inputRef}
+            />
+            <div className={findOptionsContainerStyles}>
+              {searchString && (
+                <Body>
+                  {selectedIndex ?? '?'}/{findCount}
```
stephl3, line 306: I'm picking up some unexpected behavior with `selectedIndex` selection/rendering. A couple paths to try:
1. When I type in a query, I see a `"?"` as the numerator which makes me think we're missing a `setSelectedIndex` call somewhere. This could be expected based on the initial specs, but I wanted to point out how I see Cursor handle this which is to use the last position of the text cursor to find the next result and use its index as the numerator
2. When I type in a query with a match and click down arrow, the `selectedIndex` correctly identifies a result. When I then update the query without a match, the `selectedIndex` does not update which will show something like `"1/0"` or `"2/0"`. This latter one seems like something we'd want to patch before shipping

packages/code-editor/src/SearchPanel/SearchPanel.styles.ts
```diff
@@ -0,0 +1,166 @@
...
+  Variant,
+} from '@leafygreen-ui/tokens';
+import { color } from '@leafygreen-ui/tokens';
+
+const CONTAINER_MAX_WIDTH = 500;
+const SECTION_HEIGHT = 52;
+const INPUT_WIDTH = 240;
+const INPUT_MIN_WIDTH = 120;
+
+const getBaseContainerStyles = (
```
stephl3, line 19: There's an edge case 🥁 for when a user toggles on the `SearchPanel` when the `CodeEditor` is at a single line
Screenshot 2025-10-10 at 8 14 39 AM

stephl3: maybe hard to tell; check out the right border

tsck: I agree that it looks weird, but it is _technically_ correct. The border is on the input and not on the search panel. The search panel has box shadows on the left and bottom, but not on the right. So what we're seeing there is the the border stopping since that's on the input, and the search panel overflowing the input.

stephl3: Makes sense it's technically correct, but do we want to add some additional border for that stretch? It may never be observable since users are unlikely to use the search panel when there's only 1 line of code, but I figured I would mention as design feedback in case we want to add styling

packages/code-editor/src/SearchPanel/SearchPanel.tsx
tsck: > 1. When I type in a query, I see a "?" as the numerator which makes me think we're missing a setSelectedIndex call somewhere. This could be expected based on the initial specs, but I wanted to point out how I see Cursor handle this which is to use the last position of the text cursor to find the next result and use its index as the numerator

This is because when the input is updated, we update the query in state, which triggers an update on the find count, but we don't implicitly call `findNext(view)` to trigger a selection of any of the found items. So at the point after you type, they're all highlighted but you technically have none selected, thus the "?". We could call `findNext(view)` implicitly at this point and update the selected index, but I'm not sure that creates a great experience either because the selected item will be jumping around as the user types. FWIW, it doesn't have a count so there's no "?" but how ours works currently is how the native codemirror one works as well. 

> 2. When I type in a query with a match and click down arrow, the selectedIndex correctly identifies a result. When I then update the query without a match, the selectedIndex does not update which will show something like "1/0" or "2/0". This latter one seems like something we'd want to patch before shipping

Good catch! Agreed, this definitely is a bug. Will look into pushing a fix for this shortly.

packages/code-editor/src/SearchPanel/SearchPanel.styles.ts
tsck: I can see how it looks if I just add a border to the right of the search panel, but if that doesn't look right I would probably suggest we just leave it off. To your point, its probably not a very realistic use case anyway.

tsck: I was able to fix this by adding a right border to the panel and rendering it 1 px to the right so that when its open the border is actually showing on the panel.

packages/code-editor/src/SearchPanel/SearchPanel.tsx
tsck: Point 2 should now be fixed

PR #3190, opened 2025-10-07
chore(select, code): remove prod dependency on @lg-tools/build; rename postinstall script file to prepare

gribnoysup:
## ✍️ Proposed changes

Currently there are two connected issues in select and code components:

- `@leafygreen-ui/select` has build toolchain (`@lg-tools/build`) in the peer dependencies making it so that when you install the code component, it installs a lot of unnedeed dependencies (in our update attempt package-lock has almost 5k LoC additions caused by just this). This patch removes the dep from peer dependencies
- `@leafygreen-ui/code` component has a postintall script (prepare runs on postinstall hook) that just breaks the postinstall process completely if `@lg-tools/build` is not in the dependencies. This currently goes unnoticed because `code` depends on `select` and `select` depends on `build`. If we just remove `build` from `select` peer dependencies, `code` will start to throw when externally installed. This patch adjusts the postinstall script in a way that makes it exit if `@lg-tools/build` module can't be resolved

🎟 _Jira ticket:_ [LG-5472](https://jira.mongodb.org/browse/LG-5472) [LG-5599](https://jira.mongodb.org/browse/LG-5599)

### For new components

## 🧪 How to test changes

Testing the actual effect when installing is tricky: you need to npm pack both these packages and install them to an external repo. For testing that this still creates the config locally you can just run install command and check that file exists


packages/code/scripts/postinstall.ts
```diff
@@ -19,6 +18,30 @@ const tsConfigPath = path.join(highlightJsPath, 'tsconfig.json');
...
+        'utf-8',
+      ),
+    );
+  } catch {
+    console.warn(
+      "⚠️ Can't resolve @lg-tools/build, probably running post install hook outside of the monorepo",
+    );
+    return;
+  }
+
```
TheSonOfThomp: This will be resolved on the next publish. We've moved the script from `postinstall` to `prepare` https://github.com/mongodb/leafygreen-ui/pull/3179/files

gribnoysup: Removed

gribnoysup: There's no changelog in the patch you linked, will it actually be published?

TheSonOfThomp: We forgot a changelog in that commit. It's is in a separate PR https://github.com/mongodb/leafygreen-ui/pull/3191

PR #3194, opened 2025-10-07
[LG-5595] fix(input-bar): change shouldRenderGradient default to false

stephl3:
## ✍️ Proposed changes

~~This PR updates the default value of the `shouldRenderGradient` prop from `true` to `false` in the `InputBar` component. This change prevents console warnings from being logged when the compact variant is used, as the gradient is only applicable to the spacious variant.~~

Alternatively, this PR will instead take the approach of removing the logs for when `shouldRenderGradient` is true (which happens by default). This will prevent making this a major change since the prop will eventually be removed along with other spacious variant code.

Additionally, the props in `InputBar.types.ts` have been reorganized alphabetically for better code organization and maintainability.

~~This is a breaking change as it alters the default behavior of the component, though most users of the compact variant (which doesn't support gradients) will see an improvement with fewer console warnings.~~

🎟️ _Jira ticket:_ N/A

## ✅ Checklist

## 🧪 How to test changes

1. Render an `InputBar` component in compact variant without explicitly setting `shouldRenderGradient`
2. Verify that no console warnings appear about unused props

### DIFF
--- a/.changeset/large-lands-worry.md
+++ b/.changeset/large-lands-worry.md
@@ -1,0 +1,8 @@
+---
+'@lg-chat/input-bar': minor
+---
+
+[LG-5600](https://jira.mongodb.org/browse/LG-5600)
+Fix send button disabled logic: the send button now remains enabled during loading state (even with empty message body) to allow users to stop the request. The `disabled` and `disableSend` props still take precedence.
+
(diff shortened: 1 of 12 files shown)


tsck (commented):
Just curious what led to this change? Since it is a major change and all it seems to do is remove console warnings. Would it make sense to just remove those warnings resulting in a patch (if anything).

PR #3195, opened 2025-10-07
[LG-5600] feat(input-bar): add onClickStop prop and update button disabled logic

stephl3:
## ✍️ Proposed changes

This PR improves the send button disabled state during loading. Key changes include:

- Added `onClickStop` prop that fires when the stop button is clicked during loading state
- When stop is triggered, the previous message body is automatically restored to the input field
- Introduced shared `restorePreviousMessage` helper function used by both stop and error state handling
- Fixed send button to remain enabled during loading state (even with empty message body) to allow users to stop the request. The `disabled` and `disableSend` props still take precedence and will disable the button regardless of loading state

🎟️ _Jira ticket:_ [LG-5600](https://jira.mongodb.org/browse/LG-5600)

## ✅ Checklist

## 🧪 How to test changes

1. **Test basic stop functionality:**
   - Render an `InputBar` with `state={State.Loading}` and `onClickStop` handler
   - Verify the send button is enabled even with empty message
   - Click the button and verify `onClickStop` is called

2. **Test message restoration:**
   - Type a message and submit it to trigger loading state
   - While loading, click the stop button
   - Verify the previous message is restored to the input field

3. **Test disabled state precedence:**
   - Set `disabled={true}` or `disableSend={true}` with `state={State.Loading}`
   - Verify the button remains disabled

4. **Test error state behavior:**
   - Submit a message and transition to error state
   - Verify the previous message is restored (existing behavior maintained)

5. **Run tests:** `pnpm test -- --testPathPattern=input-bar` - all 28 tests should pass

6. **Visual testing:** Check Storybook to verify no console warnings in compact variant

### DIFF
--- a/chat/input-bar/src/InputBar/InputBar.spec.tsx
+++ b/chat/input-bar/src/InputBar/InputBar.spec.tsx
@@ -385,31 +385,39 @@ describe('packages/input-bar', () => {
     });
   });
 
-  describe('onClickStop', () => {
-    test('enables send button during loading state even with empty message', () => {
+  describe('onClickStopButton', () => {
+    test('renders stop button during loading state', () => {
       renderInputBar({ state: State.Loading });
 
-      const sendButton = screen.getByRole('button', { name: 'Send message' });
-      expect(sendButton).not.toHaveAttribute('aria-disabled', 'true');
+      const stopButton = screen.getByRole('button', { name: 'Stop message' });
+      expect(stopButton).toBeInTheDocument();
+      expect(stopButton).not.toHaveAttribute('aria-disabled', 'true');
     });
 
-    test('calls onClickStop when button is clicked during loading state', () => {
-      const onClickStop = jest.fn();
-      renderInputBar({ state: State.Loading, onClickStop });
+    test('does not render send button during loading state', () => {
+      renderInputBar({ state: State.Loading });
 
-      const sendButton = screen.getByRole('button', { name: 'Send message' });
-      userEvent.click(sendButton);
+      const sendButton = screen.queryByRole('button', { name: 'Send message' });
+      expect(sendButton).not.toBeInTheDocument();
+    });
 
-      expect(onClickStop).toHaveBeenCalledTimes(1);
+    test('calls onClickStopButton when button is clicked during loading state', () => {
+      const onClickStopButton = jest.fn();
+      renderInputBar({ state: State.Loading, onClickStopButton });
+
+      const stopButton = screen.getByRole('button', { name: 'Stop message' });
+      userEvent.click(stopButton);
+
+      expect(onClickStopButton).toHaveBeenCalledTimes(1);
     });
 
     test('restores previous message when stop is clicked in uncontrolled mode', () => {
-      const onClickStop = jest.fn();
+      const onClickStopButton = jest.fn();
       const onMessageSend = jest.fn();
-      const { rerender } = renderInputBar({ onClickStop, onMessageSend });
+      const { rerender } = renderInputBar({ onClickStopButton, onMessageSend });
 
       const textarea = screen.getByRole('textbox');
-      let sendButton = screen.getByRole('button', { name: 'Send message' });
+      const sendButton = screen.getByRole('button', { name: 'Send message' });
 
       userEvent.type(textarea, TEST_INPUT_TEXT);
       expect(textarea).toHaveValue(TEST_INPUT_TEXT);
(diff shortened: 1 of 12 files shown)


chat/input-bar/src/InputBar/InputBar.types.ts
```diff
@@ -46,7 +46,13 @@ export type InputBarProps = React.ComponentPropsWithoutRef<'form'> &
     >;
 
     /**
-     * Submit event handler.
+     * Callback fired when the stop button is clicked during a loading state.
+     * When triggered, the message input will be restored to the previous message body.
+     */
+    onClickStop?: () => void;
```
TheSonOfThomp: `onStopButtonClick`?

`onClickStop` seems like something that fires if the click event is stopped


```suggestion
    onStopButtonClick?: () => void;
```

chat/input-bar/src/InputBar/InputBar.tsx
```diff
@@ -366,6 +368,12 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
         return;
       }
 
+      if (isLoading && onClickStop) {
```
TheSonOfThomp: why would the submit handler be fired if `isLoading` is true? Shouldn't those be separate events?

TheSonOfThomp: Oh I see, `isLoading` is a proxy for "submit event is actually a stop event"

IMO this feels brittle, and we should either have separate events for "submit" and "stop",
_or_ inspect the `event` object for a payload to determine the intent of this click event

chat/input-bar/src/InputBar/InputBar.tsx
```diff
@@ -142,8 +143,9 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
     const [shouldRenderButtonText, setShouldRenderButtonText] =
       useState<boolean>(false);
 
+    const isLoading = state === State.Loading;
     const isSendButtonDisabled =
-      disableSend || disabled || messageBody?.trim() === '';
+      disabled || disableSend || (!isLoading && messageBody?.trim() === '');
```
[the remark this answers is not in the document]
TheSonOfThomp: Agree. We should probably break out the "send" and "stop" button into separate components

chat/input-bar/src/InputBar/InputBar.spec.tsx
```diff
@@ -384,4 +384,107 @@ describe('packages/input-bar', () => {
...
+
+  describe('onClickStop', () => {
+    test('enables send button during loading state even with empty message', () => {
+      renderInputBar({ state: State.Loading });
+
+      const sendButton = screen.getByRole('button', { name: 'Send message' });
+      expect(sendButton).not.toHaveAttribute('aria-disabled', 'true');
+    });
+
+    test('calls onClickStop when button is clicked during loading state', () => {
```
TheSonOfThomp: do we have visual tests that assert the stop button icon & label?

chat/input-bar/src/InputBar/InputBar.types.ts
stephl3: I wanted to preserve a similar pattern used for `MessageActions`. what about `onClickStopButton`?

chat/input-bar/src/InputBar/InputBar.tsx
stephl3: I will split up the buttons and handlers :)

chat/input-bar/src/InputBar/InputBar.spec.tsx
stephl3: We have the generated stories showing different `IconButton` instances, but there is no visible label rendered with it

chat/input-bar/src/InputBar/InputBar.tsx
```diff
@@ -490,12 +508,18 @@ export const InputBar = forwardRef<HTMLFormElement, InputBarProps>(
                       /
                     </div>
                   )}
-                  <InputBarSendButton
-                    disabled={isSendButtonDisabled}
-                    isCompact={isCompact}
-                    shouldRenderButtonText={shouldRenderButtonText}
-                    state={state}
-                  />
+                  {isLoading && isCompact ? (
```
TheSonOfThomp, line 511: nit: not a huge deal, but maybe a more explicit `shouldRenderStopButton`? 
Or even a utility component with a `switch` if this ever gets more complex

stephl3: Going to leave as-is. I don't foresee it getting more complicated than this, and it will be further simplified with [this ticket to remove spacious variant code](https://jira.mongodb.org/browse/LG-5575)

PR #3197, opened 2025-10-08
Fixes TextInput component types

bruugey:
## ✍️ Proposed changes

### For new components


tools/install/src/ALL_PACKAGES.ts
```diff
@@ -6,6 +6,7 @@ export const ALL_PACKAGES = [
   '@leafygreen-ui/avatar',
   '@leafygreen-ui/badge',
   '@leafygreen-ui/banner',
+  '@leafygreen-ui/box',
```
TheSonOfThomp: need to rm the `/box` dir from your file system in order to prevent this

packages/text-area/src/TextArea/TextArea.spec.tsx
```diff
@@ -182,9 +182,14 @@ describe('packages/text-area', () => {
     });
   });
 
-  /* eslint-disable jest/no-disabled-tests */
-  describe.skip('types behave as expected', () => {
-    test('TextArea throws error when neither aria-labelledby or label is supplied', () => {
+  describe('types behave as expected', () => {
```
kraenhansen, line 185: Curious why the `skip` was moved here and removed entirely from `packages/text-input/src/TextInput/TextInput.spec.tsx`. Seems inconsistent and perhaps a mistake?

TheSonOfThomp: Technically should still be skipped, since there's no `expect` statement

PR #3417, opened 2025-12-23
LG-5835: Rm dupe aria-* attributes in FormField

bruugey:
## ✍️ Proposed changes

🎟 _Jira ticket:_ [LG-5835](https://jira.mongodb.org/browse/LG-5835)

### For new components

### DIFF
--- a/.changeset/plain-houses-lose.md
+++ b/.changeset/plain-houses-lose.md
@@ -1,0 +1,5 @@
+---
+'@leafygreen-ui/form-field': patch
+---
+
+Removes duplicate `aria-*` attributes
--- a/packages/form-field/src/FormField/FormField.tsx
+++ b/packages/form-field/src/FormField/FormField.tsx
@@ -47,6 +47,10 @@ export const FormField = forwardRef<HTMLDivElement, FormFieldProps>(
       optional,
       id,
       'data-lgid': dataLgId,
+      'aria-label': ariaLabelProp,
+      'aria-labelledby': ariaLabelledbyProp,
+      'aria-invalid': ariaInvalidProp,
+      readOnly,
       ...rest
     }: FormFieldProps,
     fwdRef,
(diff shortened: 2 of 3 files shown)


.changeset/plain-houses-lose.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/form-field': patch
+---
+
+Removes duplicate `aria-*` attributes
```
tsck, line 5: Nit: also removes duplicate `readonly` attribute

tsck (approved):

packages/form-field/src/FormField/FormField.tsx
```diff
@@ -47,6 +47,10 @@ export const FormField = forwardRef<HTMLDivElement, FormFieldProps>(
       optional,
       id,
       'data-lgid': dataLgId,
+      'aria-label': ariaLabelProp,
+      'aria-labelledby': ariaLabelledbyProp,
+      'aria-invalid': ariaInvalidProp,
+      readOnly,
```
shaneeza, line 53: I think we should match everything thats returned with input props:

https://github.com/mongodb/leafygreen-ui/blob/c2e7029733f62c868bb369c7a5e66cac8fecb6ff/packages/form-field/src/FormField/useFormFieldProps.ts#L7-L14

.changeset/plain-houses-lose.md
```diff
@@ -0,0 +1,5 @@
+---
+'@leafygreen-ui/form-field': patch
```
stephl3, line 2: I believe if this is a `patch` bump, it won't propagate to downstream dependencies unless they are explicitly listed here

packages/form-field/src/FormField/FormField.tsx
stephl3: can we also add test specs for this fix?

.changeset/plain-houses-lose.md
```diff
@@ -0,0 +1,5 @@
+---
```
stephl3, line 1: Screenshot 2025-12-23 at 11 08 58 AM

When checking the `FormField` storybook example, I see this change has addressed the initial issue, but it looks like there is a separate issue with duplicate usages of `data-lgid` and `data-testid` on different nodes. Can we batch those fixes with this change?

shaneeza (changes requested):
The comments left by Stephen and me should be addressed before merging in.

PR #3500, opened 2026-02-03
fix(mcp): add origin validation to useRenderData hook

RinZ27:
## ✍️ Proposed changes

I've implemented origin validation in the `useRenderData` hook to enhance security for iframe-based communication in the MCP integration. Previously, the hook's message handler accepted data from any source origin without verification. This update introduces an optional `allowedOrigins` configuration, allowing developers to restrict data updates to trusted origins only.

🎟 _Jira ticket:_ N/A (Security enhancement for MCP hooks)

## 🧪 How to test changes

1. Configure the `useRenderData` hook with an `allowedOrigins` array (e.g., `['https://mongodb.com']`).
2. Simulate a cross-document message using `window.dispatchEvent(new MessageEvent('message', { ... }))` from an origin not in the list. Observe that the state remains unchanged.
3. Simulate a message from an origin in the allowlist and verify that `renderData` is updated as expected.
4. Existing unit tests in `useRenderData.spec.tsx` have been updated to cover these scenarios.


mcp/hooks/src/useRenderData/useRenderData.ts
```diff
@@ -109,7 +119,7 @@ export function useRenderData<T = unknown>(): UseRenderDataResult<T> {
     window.parent.postMessage({ type: MESSAGE_TYPE_IFRAME_READY }, '*');
 
     return () => window.removeEventListener('message', handleMessage);
-  }, []);
+  }, [options]);
```
TheSonOfThomp: **Issue 1 — Effect dep regression: re-fires the parent handshake on every render**

Changing the dep from `[]` to `[options]` means the effect tears down and re-adds the listener on every render where `options` is an inline object literal — its identity changes each render. The critical side-effect: `window.parent.postMessage({ type: MESSAGE_TYPE_IFRAME_READY }, '*')` re-fires each time. Any parent that treats the ready message as a one-time handshake (debouncing, state machines, capability negotiation) will see a flood of events. No test enforces the "effect runs once" contract, so this slipped through.

Suggested fix — depend on a stable serialization rather than the object reference:
```ts
const allowedOriginsKey = JSON.stringify(options?.allowedOrigins);
useEffect(function handleMessages() {
  // ...
}, [allowedOriginsKey]);
```
Alternatively, hold `options` in a ref and read it inside the effect, keeping the dep array as `[]`.

mcp/hooks/src/useRenderData/useRenderData.ts
```diff
@@ -16,7 +17,12 @@ const MESSAGE_TYPE_IFRAME_READY = 'ui-lifecycle-iframe-ready';
  */
 function validateRenderData<T>(
   event: MessageEvent<RenderDataMessage>,
+  options?: UseRenderDataOptions,
 ): ValidationResult<T> | null {
+  if (options?.allowedOrigins && !options.allowedOrigins.includes(event.origin)) {
```
TheSonOfThomp: **Issues 2 & 4 — Origin filter precedes message-type check; empty array silently blocks everything**

**Issue 2 (ordering):** Returning `null` here — the same sentinel as "unrelated postMessage, ignore" — *before* checking `MESSAGE_TYPE_RENDER_DATA` makes origin rejection indistinguishable from noise. A developer who fat-fingers `allowedOrigins` ends up with a hook stuck in `isLoading: true` forever with no diagnostic signal. Consider checking message type first, then origin:
```ts
if (!isRenderDataMessage) return null; // unrelated traffic
if (options?.allowedOrigins?.length && !options.allowedOrigins.includes(event.origin)) {
  console.warn(`[useRenderData] Blocked message from disallowed origin: ${event.origin}`);
  return null;
}
```

**Issue 4 (empty array footgun):** `options.allowedOrigins &&` is truthy for `[]`, so `allowedOrigins: []` blocks *every* origin silently. A developer passing an empty default state hits the same stuck-loading behavior. Either treat `[]` as "no restriction" (i.e. `allowedOrigins?.length && ...`) or add a JSDoc callout on the type making the behavior explicit.

mcp/hooks/src/useRenderData/useRenderData.ts
```diff
@@ -109,7 +119,7 @@ export function useRenderData<T = unknown>(): UseRenderDataResult<T> {
     window.parent.postMessage({ type: MESSAGE_TYPE_IFRAME_READY }, '*');
```
TheSonOfThomp: **Issue 6 — Outbound handshake still broadcasts to `'*'`**

This PR restricts which origins can *send* data to the hook, but the outbound `iframe-ready` signal is still broadcast to any embedder. An attacker frame that embeds the component still learns the iframe is alive and ready, which partially undermines the security intent.

If `allowedOrigins` is set, consider scoping the target:
```ts
const targetOrigin = options?.allowedOrigins?.[0] ?? '*';
window.parent.postMessage({ type: MESSAGE_TYPE_IFRAME_READY }, targetOrigin);
```
Or iterate and post to each allowed origin. At minimum, a comment explaining why the outbound side is intentionally left open (if that is the decision) would help reviewers.

mcp/hooks/src/useRenderData/useRenderData.spec.tsx
```diff
@@ -133,6 +133,50 @@ describe('mcp/hooks/useRenderData', () => {
...
+            type: 'ui-lifecycle-iframe-render-data',
+            payload: { renderData: { name: 'attacker' } },
+          },
+          origin: 'https://attacker.com',
+        }),
+      );
+    });
+
+    expect(result.current.isLoading).toBe(true);
+    expect(result.current.data).toBe(null);
```
TheSonOfThomp: **Issue 3 — No test for the default (no-options) path**

Both new tests pass `allowedOrigins`. The implicit back-compat contract — "omitting `allowedOrigins` continues to accept messages from any origin" — is untested. A regression here would be silent.

Suggested addition after this block:
```ts
test('accepts messages from any origin when allowedOrigins is not specified', async () => {
  const { result } = renderHook(() => useRenderData()); // no options
  act(() => {
    window.dispatchEvent(new MessageEvent('message', {
      data: {
        type: 'ui-lifecycle-iframe-render-data',
        payload: { renderData: { name: 'test' } },
      },
      origin: 'https://any-origin.example.com',
    }));
  });
  await waitFor(() => {
    expect(result.current.isLoading).toBe(false);
    expect(result.current.data).toEqual({ name: 'test' });
  });
});
```

mcp/hooks/src/useRenderData/useRenderData.types.ts
```diff
@@ -9,6 +9,14 @@ export interface BaseRenderData {
   darkMode?: boolean;
 }
 
+export interface UseRenderDataOptions {
+  /**
+   * List of allowed origins for postMessage communication.
+   * If provided, messages from other origins will be ignored.
+   */
+  allowedOrigins?: Array<string>;
```
TheSonOfThomp, line 20: **Issue 5 — Sandboxed / opaque origins have no escape hatch, and this is undocumented**

Sandboxed iframes (`<iframe sandbox>` without `allow-same-origin`) and `file://` parents deliver `event.origin === 'null'` (the string literal, not JS `null`). `Array.includes` does exact string matching, so there is no way to allow these contexts, and a developer won't know why their `allowedOrigins` setup silently fails.

Suggested JSDoc addition:
```ts
/**
 * List of allowed origins for postMessage communication.
 * If provided, messages from other origins will be ignored.
 *
 * Note: sandboxed iframes (without `allow-same-origin`) and `file://` parents
 * report `event.origin` as the string `'null'`. To allow them, include `'null'`
 * in this array explicitly.
 */
```

mcp/hooks/src/useRenderData/useRenderData.types.ts
```diff
@@ -9,6 +9,14 @@ export interface BaseRenderData {
   darkMode?: boolean;
 }
 
+export interface UseRenderDataOptions {
```
TheSonOfThomp, line 12: **Issue 7 — `UseRenderDataOptions` may be missing from the package barrel export**

This type is defined here but does not appear in the diff for `mcp/hooks/src/index.ts`. Consumers who want to declare `const opts: UseRenderDataOptions` as a named import won't find it. Please verify the barrel and add the re-export if it's missing.

TheSonOfThomp (commented):
Seven issues found during review — one behavioral regression, several test coverage gaps, and security hardening notes. Detailed inline comments on each below.

RinZ27:
Addressed the feedback by refactoring the validation logic and fixing the dependency regression. Here's a summary of the updates:

*   **Logic fix:** Swapped the validation order to check message type before origin. This prevents the hook from getting stuck in a loading state silently. Added a `console.warn` for blocked origins.
*   **Performance:** Wrapped options in a stable memo to prevent redundant handshake messages on every re-render.
*   **Security:** Tightened the handshake by restricting `targetOrigin` to the first allowed origin when provided, instead of using `*`.
*   **DX & Docs:** Added JSDoc notes regarding sandboxed/opaque origins (`"null"`) and exported the missing types.
*   **Test Coverage:** Added new test cases covering target origin restriction, empty array behavior, and effect stability.

PR #3526, opened 2026-07-16
UXE-477: pack once, publish the same tarball to every registry

TheSonOfThomp:
Fixes the release pipeline so a given version's tarball is byte-identical across npmjs and CodeArtifact, instead of each registry getting an independently-packed (and drifting) artifact.

`release-aws` previously ran `pnpm publish -r --no-git-checks`, which rebuilds and re-packs every package from source, independently of the `release` job's npmjs publish moments earlier — see UXE-477 / LG-6236 for evidence of the same version producing three different tarball sha512s across npmjs/Artifactory/CodeArtifact. This blocks Bazel checksum verification during the MMS Artifactory→CodeArtifact migration (DEVPROD-37252 / DEVPROD-31845).

`release-aws` now fetches the exact tarball `release` already published to npmjs (`npm pack <name>@<version>` on a registry spec downloads the published bytes verbatim — it does not rebuild) and republishes those same bytes to CodeArtifact via the new `scripts/republish-to-registry.mjs`.

Implements the plan from UXE-477 (formerly LG-6236).

## Test Results

No existing automated test target covers `.github/workflows/` or `scripts/` in this repo (e.g. `scripts/login-codeartifact.sh` and `scripts/validate-changelogs.mjs` have no spec files either), and this diff touches neither `packages/*` nor `tools/*` source, so nothing to scope a package-level test run to.

Verified instead:
- `release.yml` re-validated with `yaml.safe_load`; `bash -n` on the shell script; `node --check` on the new script; `prettier --check` clean on the yml/mjs files.
- Ran `republish-to-registry.mjs` standalone against the real npmjs registry, fetching `@leafygreen-ui/a11y@3.0.5` — the fetched tarball's sha512 matched the ticket's own cited npmjs sha512 exactly.
- Confirmed (by hand, against a poisoned `@scope:registry` config) that an explicit `--registry` flag on `npm pack`/`npm publish` loses to an ambient `@scope:registry` entry — which is why the npmjs fetch uses an isolated `--userconfig`, and why publish deliberately relies on the existing scope config (with a safety check) instead of an explicit `--registry` flag.
- Ran the script with a registry argument matching this machine's real ambient scope config — confirmed it reaches the real `npm publish` call (fails only on auth/already-published, as expected in this sandbox).
- Ran the script with a mismatched registry argument — confirmed the safety check throws before attempting to publish.

## Review notes

A fresh-context review pass flagged a few non-blocking judgment calls, left as-is per this ticket's scope (byte-identical artifacts) rather than expanded into general error-handling hardening:
- A mid-loop failure (`npm pack`/`npm publish` for package N of M) aborts the whole script, leaving 1..N-1 already republished; a naive retry would re-attempt those and fail (registries reject re-publishing an existing version). Same idempotency characteristics as the old `pnpm publish -r`, just worth knowing going in.
- No retry/backoff around `npm pack` for a transient propagation-lag right after `release` just published — untested in practice since the standalone test used a long-published version, not a just-published one.
- The isolated `--userconfig` used for the npmjs fetch isolates against *user-level* npm config (e.g. `~/.npmrc`, which is what `login-codeartifact.sh` writes to) but not a hypothetical *project-level* `.npmrc` scope override — not exploitable today since this repo's `.npmrc` sets no scope/registry keys, but worth knowing if that ever changes.


.github/workflows/test-republish-to-registry.yml
```diff
@@ -0,0 +1,48 @@
...
+        id: login-codeartifact
+        run: scripts/login-codeartifact.sh
+
+      - name: Publish to CodeArtifact
+        env:
+          PACKAGE_NAME: ${{ inputs.package_name }}
+          PACKAGE_VERSION: ${{ inputs.package_version }}
+        run: |
+          PACKAGES=$(node -e "console.log(JSON.stringify([{name: process.env.PACKAGE_NAME, version: process.env.PACKAGE_VERSION}]))")
+          node scripts/republish-to-registry.mjs "${{ steps.login-codeartifact.outputs.registry }}" "$PACKAGES"
```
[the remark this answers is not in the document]
TheSonOfThomp: Added an explicit `permissions: contents: read` block to the job.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/test-republish-to-registry.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Use Node 22
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
```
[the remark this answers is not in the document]
TheSonOfThomp: Pinned to the commit SHA `aws-actions/configure-aws-credentials@v4` currently resolves to (verified via `git ls-remote --tags`), with the version as a comment.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/test-republish-to-registry.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  test-republish:
+    name: Republish a single package
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Use Node 22
+        uses: actions/setup-node@v4
```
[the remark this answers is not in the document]
TheSonOfThomp: Pinned to the commit SHA `actions/setup-node@v4` currently resolves to (verified via `git ls-remote --tags`), with the version as a comment.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/test-republish-to-registry.yml
```diff
@@ -0,0 +1,48 @@
...
+      package_version:
+        description: 'Version to test with, e.g. 3.0.5'
+        required: true
+
+jobs:
+  test-republish:
+    name: Republish a single package
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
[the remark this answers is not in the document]
TheSonOfThomp: Pinned to the commit SHA `actions/checkout@v4` currently resolves to (verified via `git ls-remote --tags`), with the version as a comment.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

scripts/republish-to-registry.mjs
```diff
@@ -0,0 +1,90 @@
...
+if (!Array.isArray(packages) || packages.length === 0) {
+  console.log('No published packages to republish.');
+  process.exit(0);
+}
+
+const workDir = mkdtempSync(path.join(tmpdir(), 'leafygreen-republish-'));
+
+// No scope overrides here, so this can't be redirected toward `registry`.
+const npmjsConfigPath = path.join(workDir, '.npmrc-npmjs-only');
+writeFileSync(npmjsConfigPath, 'registry=https://registry.npmjs.org\n');
```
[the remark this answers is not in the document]
TheSonOfThomp: Good catch — wrapped the fetch/publish loop in try/finally and now `rmSync` the temp work directory afterward (verified it's actually removed in both the success and the thrown-error path).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

scripts/republish-to-registry.mjs
```diff
@@ -0,0 +1,94 @@
...
+        '--json',
+      ],
+      { encoding: 'utf-8', stdio: ['ignore', 'pipe', 'inherit'] },
+    );
+    const [{ filename }] = JSON.parse(packOutput);
+    const tarballPath = path.join(workDir, filename);
+
+    // Catch a scope login-codeartifact.sh forgot to configure, instead of silently
+    // publishing it somewhere else.
+    const scope = name.startsWith('@') ? name.split('/')[0] : null;
```
stephl3: Unscoped packages bypass the safety check *and* get mis-routed. The `scope` guard (`name.startsWith('@') ? … : null`) skips the `configuredRegistry !== registry` verification for any unscoped name — and since the real `npm publish <tarball>` relies on ambient `@scope:registry` config to reach CodeArtifact, an unscoped package (no scope override) would publish to the default registry (npmjs) instead. All LG packages are scoped today so this is latent, but the failure mode is silent mis-publishing rather than an error. Consider failing loudly on an unscoped name instead of skipping the check.

scripts/republish-to-registry.mjs
```diff
@@ -0,0 +1,94 @@
...
+        throw new Error(
+          `${scope} is configured to publish to "${configuredRegistry}", not the ` +
+            `expected "${registry}". Add ${scope} to login-codeartifact.sh's SCOPES ` +
+            `list before publishing ${spec}.`,
+        );
+      }
+    }
+
+    console.log(`Publishing ${spec} to ${registry}...`);
+    execFileSync('npm', ['publish', tarballPath], { stdio: 'inherit' });
```
stephl3: Mid-loop failure leaves a partial, non-idempotent publish: if `npm publish` throws on package N (throttle, propagation lag, network), packages 1..N-1 are already in CodeArtifact and a plain job re-run then fails because those versions already exist — blocking clean recovery. This is acknowledged in the PR's Review notes as accepted scope; flagging inline so it's discoverable. A cheap improvement would be to treat "version already exists" as a no-op so re-runs are idempotent.

.github/workflows/test-republish-to-registry.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_CODEARTIFACT_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_CODEARTIFACT_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+
+      - name: Login to CodeArtifact
+        id: login-codeartifact
+        run: scripts/login-codeartifact.sh
+
+      - name: Publish to CodeArtifact
```
stephl3: This smoke test republishes an already-published version, which npm rejects with a non-zero exit — and the script's `npm publish` throws on that, so the step goes red even on the happy path. With no `continue-on-error` and nothing distinguishing the *expected* conflict from a real failure (bad auth, pack 404, scope-check throw), a passing run and a genuine regression look identical, which undercuts the workflow's stated "fails harmlessly" premise. Consider `continue-on-error` plus an assertion that the failure was specifically the version conflict, or have the script treat "already exists" as success.

stephl3 (commented):
This is a well-motivated, correct fix for cross-registry sha512 drift: downloading the exact npmjs tarball and republishing those bytes to CodeArtifact is the right approach, and the `needs: [build, release] -> [release]` plus build-cache removal is correct since no local build is needed. The design is sound — shell-less `execFileSync`, an isolated `--userconfig` for the fetch, and an explicit scope-registry safety check.

A few non-blocking robustness notes below. Both `major`s are latent (the partial-publish one is already acknowledged in your Review notes as accepted scope). The one worth a second look is the smoke-test workflow: because its premise is republishing an already-published version, `npm publish` exits non-zero and the step likely goes red even on a successful run, so a real regression and a happy-path run look identical.

scripts/republish-to-registry.mjs
TheSonOfThomp: Fixed — unscoped names now throw immediately instead of skipping the check (verified with an unscoped package: fails loudly with a clear message before any publish attempt).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

scripts/republish-to-registry.mjs
TheSonOfThomp: Took the suggested approach — `npm publish` failures matching "cannot publish over the previously published version" are now treated as a no-op instead of an error, so a re-run after a mid-loop failure skips the already-published packages and continues. Verified the regex against real npm 403 conflict text and confirmed it doesn't swallow 404/401 errors.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/test-republish-to-registry.yml
TheSonOfThomp: Fixed via the same idempotency change in the script rather than `continue-on-error` — republishing an already-published version is now a success (not just a harmless failure), so the smoke test's happy path is actually green, and a genuine failure (bad auth, pack 404, scope mismatch) still fails the step. Updated the workflow comment to match.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

stephl3 (approved):

Bod1eF (approved):

PR #3528, opened 2026-07-21
fix(descendants): register descendants rendered inside an iframe

tsck:
## ✍️ Proposed changes

`Tabs`/`Tab` and any component built on `@leafygreen-ui/descendants` misbehaved inside an `iframe`. Clicking one tab selected every tab and rendered all panels. The registration effect in `useDescendant` checked `document.contains(ref.current)`, which returned `false` for elements in an iframe's document and left every item with a fallback index. The fix uses the element's own document: `ref.current?.ownerDocument?.contains(ref.current) ?? false`.

🎟 _Jira ticket:_ [UXE-495](https://jira.mongodb.org/browse/UXE-495)

## 🧪 How to test changes

1. Render `Tabs` with several `Tab`s inside an `iframe`.
2. Click a single tab.
3. Before this fix all tabs appear selected and every panel renders. After, only the clicked tab is selected.

Automated coverage: `packages/descendants/src/Descendants/Descendants.spec.tsx`, "registers descendants rendered inside an iframe" (fails on `main`, passes with this change).

### Before
before-buggy

### After
after-fixed

### DIFF
--- a/packages/descendants/src/Descendants/Descendants.spec.tsx
+++ b/packages/descendants/src/Descendants/Descendants.spec.tsx
@@ -180,7 +180,7 @@ describe('packages/descendants', () => {
       const container = iframeDocument.createElement('div');
       iframeDocument.body.appendChild(container);
 
-      const { queryByText, unmount } = render(
+      const { queryByText } = render(
         <TestParent>
           <TestDescendant>Apple</TestDescendant>
           <TestDescendant>Banana</TestDescendant>
@@ -189,14 +189,11 @@ describe('packages/descendants', () => {
         { container },
       );
 
-      try {
-        expect(queryByText('Apple')).toHaveAttribute('data-index', '0');
-        expect(queryByText('Banana')).toHaveAttribute('data-index', '1');
-        expect(queryByText('Carrot')).toHaveAttribute('data-index', '2');
-      } finally {
-        unmount();
-        document.body.removeChild(iframe);
-      }
+      expect(queryByText('Apple')).toHaveAttribute('data-index', '0');
+      expect(queryByText('Banana')).toHaveAttribute('data-index', '1');
+      expect(queryByText('Carrot')).toHaveAttribute('data-index', '2');
+
+      document.body.removeChild(iframe);
     });
   });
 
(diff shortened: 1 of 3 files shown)


packages/descendants/src/Descendants/Descendants.spec.tsx
```diff
@@ -172,6 +172,32 @@ describe('packages/descendants', () => {
...
+        </TestParent>,
+        { container },
+      );
+
+      expect(queryByText('Apple')).toHaveAttribute('data-index', '0');
+      expect(queryByText('Banana')).toHaveAttribute('data-index', '1');
+      expect(queryByText('Carrot')).toHaveAttribute('data-index', '2');
+
+      document.body.removeChild(iframe);
+    });
```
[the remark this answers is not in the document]
tsck: Fixed in f93068f71: wrapped the assertions in try/finally and added unmount() so the iframe is removed even if an assertion fails. Kept the contentDocument non-null assertion since an attached iframe always has one in jsdom, consistent with other tests in this file.

packages/descendants/src/Descendants/Descendants.spec.tsx
```diff
@@ -172,6 +172,32 @@ describe('packages/descendants', () => {
...
+      );
+
+      try {
+        expect(queryByText('Apple')).toHaveAttribute('data-index', '0');
+        expect(queryByText('Banana')).toHaveAttribute('data-index', '1');
+        expect(queryByText('Carrot')).toHaveAttribute('data-index', '2');
+      } finally {
+        unmount();
+        document.body.removeChild(iframe);
+      }
```
TheSonOfThomp: is the `try/finally` necessary here?

TheSonOfThomp (approved):

packages/descendants/src/Descendants/Descendants.spec.tsx
tsck: Nope, not necessary — reverted it. Jest fails the test on a throw anyway and RTL cleanup handles the unmount, so it was just defensive noise from the Copilot nit. Back to the simple version.

